PR #124, opened 2024-05-15
Restors register and hcpeopnshiftcluster_methods files while fixing dependabot issues with typespec/autorest

s-amann:
### What this PR does
**Before this PR:**

**After this PR:**

Jira: 
Link to demo recording: 

### Checklist


internal/api/v20240610preview/hcpopenshiftclusters_methods.go
```diff
@@ -0,0 +1,599 @@
...
+	"github.com/Azure/ARO-HCP/internal/api/v20240610preview/generated"
+)
+
+type HcpOpenShiftClusterResource struct {
+	generated.HcpOpenShiftClusterResource
+}
+
+type VersionProfile struct {
+	generated.VersionProfile
+}
```
mbarnes, line 22: We don't strictly need these wrapper types.  I was being a little too cute with all the `Normalize` methods I slapped on everything.  Like in this case we could get by with just the `normalizeVersion` function and drop the method.  But we can defer that to a follow-up pull request.

mbarnes (dismissed):
LGTM.  Had a comment below about how to further simplify the `Normalize` logic but it doesn't need to be addressed here.

internal/api/v20240610preview/hcpopenshiftclusters_methods.go
s-amann: +1 for a follow up, this PR is already large

package-lock.json
mjlshen, line 1: Do we want to undo these changes?

PR #128, opened 2024-05-20
updates db client setup with new config creation

tonytheleg:
### What this PR does
Follow-up to !125 based on Mathew's comments:
* adds DBConfig type to decouple the dbconfig and credentials from the client setup
* adds DBconfig to client for use in later calls and client instantiaion
* This will allows to define more explicit credential types and options later by only modifying the DBconfig and NewDBConfig objects, leaving client agnostic to it

Jira: 
Link to demo recording: 

### Checklist


petrkotas (approved):
LGTM, thanks.

mjlshen (approved):

frontend/database.go
```diff
@@ -17,30 +17,50 @@ const (
...
+func NewDatabaseClient(config *DBConfig) *DBClient {
 	d := &DBClient{
-		DBName: os.Getenv("DB_NAME"),
-		DBUrl:  os.Getenv("DB_URL"),
+		config: config,
 	}
-	client, _ := azcosmos.NewClient(d.DBUrl, cred, nil)
-	d.client = client
 
+	client, _ := azcosmos.NewClient(d.config.DBUrl, d.config.Cred, nil)
```
s-amann, line 51: why not check this error?

s-amann (approved):

frontend/database.go
tonytheleg: good call, think i missed this because  client is defined in the struct which would't allow for error handling further up. I can move this out like how database config is done now so we can have error handling

```
func NewFrontend(logger *slog.Logger, listener net.Listener, emitter metrics.Emitter, dbConfig *DBConfig) *Frontend {
	f := &Frontend{
		logger:   logger,
		listener: listener,
		metrics:  emitter,
		server: http.Server{
			ErrorLog: slog.NewLogLogger(logger.Handler(), slog.LevelError),
			BaseContext: func(net.Listener) context.Context {
				return ContextWithLogger(context.Background(), logger)
			},
		},
		cache:    *NewCache(),
		dbClient: *NewDatabaseClient(dbConfig),  <---
		done:     make(chan struct{}),
	}
```

tonytheleg: ok added the error check and moved the client setup to main so it can be passed to Frontend and includes the DB config with it. Validated changes still work in dev environment.

mjlshen (dismissed):

mbarnes (changes requested):
One suggestion.  I'm making it a requested change so I don't merge prematurely again.  :slightly_smiling_face:

PR #137, opened 2024-05-22
add subscription state metrics

katherinelc321:
### What this PR does
**Before this PR:**
There are no metrics that tracks subscription state
**After this PR:**
The subscription validator middleware emits metrics for the state
Added env var containing the region where the RP is deployed to the metric contains the region label

Jira: ARO-6006
Link to demo recording: 

### Checklist

### DIFF
--- a/frontend/main.go
+++ b/frontend/main.go
@@ -29,6 +29,13 @@ func main() {
 
 	logger.Info(fmt.Sprintf("%s (%s) started", ProgramName, version))
 
+	// Fetch the region from the env variable
+	region := os.Getenv("REGION")
+	if region == "" {
+		logger.Error("REGION env variable is not set.")
+	}
+	logger.Info(fmt.Sprintf("Application running in region: %s", region))
+
 	ctx := context.Background()
 	stop := make(chan struct{})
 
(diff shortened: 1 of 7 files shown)


frontend/main.go
```diff
@@ -29,6 +29,13 @@ func main() {
 
 	logger.Info(fmt.Sprintf("%s (%s) started", ProgramName, version))
 
+	// Fetch the region from the env variable
+	region := os.Getenv("REGION")
```
s-amann, line 33: I think this is fine for now. Eventually we will have more external configuration for the service when running in a hosted environment and we may want to reconsider pulling from the env in an ad hoc manner

frontend/middleware_validatesubscription.go
```diff
@@ -49,6 +54,13 @@ func (s *SubscriptionStateMuxValidator) MiddlewareValidateSubscriptionState(w ht
 		return
 	}
 
+	// Emit the subscription state metric
+	s.metrics.EmitGauge("subscription_lifecycle", 1, map[string]string{
```
s-amann, line 58: the name `subscription_lifecycle` implies this is emitted when the Subscription Lifecycle endpoint is called. However, this middleware is not invoked at that time. Rather, this middleware is invoked for every call a user would make on the resources provided.

perhaps a name like `resource_request_subscription_state`

frontend/middleware_validatesubscription_test.go
```diff
@@ -16,13 +16,29 @@ import (
 	"github.com/google/go-cmp/cmp"
 
 	"github.com/Azure/ARO-HCP/internal/api/arm"
+	"github.com/Azure/ARO-HCP/internal/metrics"
 )
 
+type MockEmitter struct{}
```
s-amann, line 22: should this get extracted to a test/util package for importing in other test files using the `emitter`?

s-amann (commented):
Left some suggestions for review

PR #140, opened 2024-05-22
Fix up deployment what-if github action

bennerv:
### What this PR does
**Before this PR:**
- broken github action

**After this PR:**
- probably still broken


bennerv:
Closing in favour of #141

PR #145, opened 2024-05-23
adds db calls for subscriptions

tonytheleg:
### What this PR does
**Before this PR:**
* Subscription data was only stored in cache and not cosmos

**After this PR:**
* Subscription data is put into Cosmos DB
* Subscription data can be fetched and updated
* Document types are combined into single `documents.go`

Jira: [ARO-7410](https://issues.redhat.com/browse/ARO-7410)
Link to demo recording: 

Validation of all changes available in Jira comments ([LINK](https://issues.redhat.com/browse/ARO-7410?focusedId=24786971&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24786971)

### Checklist


frontend/database.go
```diff
@@ -153,3 +155,58 @@ func (d *DBClient) DeleteClusterDoc(ctx context.Context, resourceID string, part
...
+			err = json.Unmarshal(item, &doc)
+			if err != nil {
+				return nil, false, err
+			}
+		}
+	}
+	if doc != nil {
+		return doc, true, nil
+	}
+	return nil, false, nil
```
mjlshen, line 192: Thinking out loud, for this and `GetClusterDoc` should "not found" always return an error? Do we need to boolean return to identify if an object was found or not?

mjlshen (approved):
/lgtm, non-blocking question

frontend/database.go
mjlshen: Oh, yeah, we could do https://go.dev/blog/go1.13-errors instead of string comparison if we wanted to get rid of the boolean return value.

tonytheleg: Sounds like a plan, ill work on getting this updated to define an error and check the error type instead.

tonytheleg: @mjlshen updated, lemme know what you think, its better IMO. Cleaner.

frontend/frontend.go
```diff
@@ -311,13 +313,13 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
...
 	var doc *HCPOpenShiftClusterDocument
-	doc, found, err := f.dbClient.GetClusterDoc(ctx, resourceID, parsed.SubscriptionID)
+	doc, err = f.dbClient.GetClusterDoc(ctx, resourceID, parsed.SubscriptionID)
 	if err != nil {
 		f.logger.Error("failed to fetch document for %s: %v", resourceID, err)
 		arm.WriteInternalServerError(writer)
 		return
 	}
-	if !found {
+	if errors.Is(err, ErrNotFound) {
```
mbarnes, line 322: This check needs to come before `if err != nil` or else we'll end up treating `ErrNotFound` like any other error.

frontend/database.go
mbarnes: +1 on eliminating the boolean, but as I noted elsewhere be careful in error handling to check specifically for `NotFoundError` _before_ `err != nil`.

frontend/frontend.go
tonytheleg: good catch, thanks! Change coming

tonytheleg: ok should be updated everywhere to 
1) Check for the ErrNotFound first when it make sense to
2) Finishes implementing the change from bool to error for Subscription docs too

tonytheleg:
Throwing this in draft as there is one other bit of logic i think needs updating and want to make sure its all tested.

mjlshen (approved):

frontend/frontend.go
```diff
@@ -387,9 +391,15 @@ func (f *Frontend) ArmResourceDelete(writer http.ResponseWriter, request *http.R
 	parsed, _ := azure.ParseResourceID(resourceID)
 	err = f.dbClient.DeleteClusterDoc(ctx, resourceID, parsed.SubscriptionID)
 	if err != nil {
-		f.logger.Error(err.Error())
-		arm.WriteInternalServerError(writer)
-		return
+		if errors.Is(err, ErrNotFound) {
+			f.logger.Info(fmt.Sprintf("cluster document cannot be deleted -- document not found for %s", resourceID))
+			writer.WriteHeader(http.StatusNotFound)
```
mbarnes, line 396: Quoth the RP Contract for resource deletion:
> The resource provider can return 200 (OK) or 204 (NoContent) to indicate that the operation completed successfully. A 200 (OK) should be returned if the object exists and was deleted successfully; and a 204 (NoContent) should be used if the resource does not exist and the request is well formed.

So I think we want `http.StatusNoContent` here.

frontend/frontend.go
```diff
@@ -432,6 +473,31 @@ func (f *Frontend) ArmSubscriptionAction(writer http.ResponseWriter, request *ht
 	subId := request.PathValue(PathSegmentSubscriptionID)
 	f.cache.SetSubscription(subId, &subscription)
 
+	var doc *SubscriptionDocument
+	doc, err = f.dbClient.GetSubscriptionDoc(ctx, subId)
+	if err != nil && !errors.Is(err, ErrNotFound) {
```
mbarnes, line 478: This error handling logic flow seems a little clunky to me.  I get that you're trying to handle actual error conditions first, but I think it would read better if you structured it as you've done elsewhere:

```
if err != nil {
        if errors.Is(err, ErrNotFound) {
                // initialize new SubscriptionDocument
        } else {
                // write error and return
        }
} else {
        // use found SubscriptionDocument
}
```

Just my opinion though.  If you disagree then disregard.

mbarnes (changes requested):
Couple more nits.  Requested change is for the status code, the other comment is optional.

frontend/frontend.go
tonytheleg: Good ole RP requirements haha. Easy enough fix though, thanks for catching it!

frontend/frontend.go
tonytheleg: Don't have strong opinions here, your ask makes sense, happy to add the change while im making the other one.

tonytheleg: I remember what my thinking was here. Since this call is being used to create or update, i was trying to avoid setting the `doc.Subscription` value before checking for all possible errors. I think what you've suggested does the same so that works for me, i think i just missed this way of getting to the same end. Thanks!

tonytheleg:
Requested changes added

mbarnes (approved):

mbarnes (dismissed):
Approved once the merge conflict is resolved.

mbarnes (approved):

PR #149, opened 2024-05-29
Refactor Frontend to Support CLI Flags for CS integration

mjlshen:
### What this PR does
Bug Fixes:
* MetricsMiddleware was not initializing its cache from frontend
* `frontend_counter` had mismatched labels
* Makefile deploy/undeploy bugfixes after the new `REGION` environment variable was introduced

Changes:
* Frontend code is restructured
* The addition of CLI flags to configure communication to Clusters Service to pave the way for future integration
* Addition of CLI flags to configure CosmosDB, existing environment variables still work

Jira: [ARO-7828](https://issues.redhat.com/browse/ARO-7828)
Link to demo recording: TBD - available if it will be helpful

Tested existing Frontend --> Cosmos functionality to create/update subscriptions and create/delete hostedclusters
Tested new Frontend --> local CS functionality to validate a successful connection


tonytheleg (dismissed):

tonytheleg (approved):

frontend/pkg/frontend/frontend.go
```diff
@@ -36,14 +36,15 @@ const (
 )
 
 type Frontend struct {
+	conn     *sdk.Connection
 	logger   *slog.Logger
 	listener net.Listener
 	server   http.Server
 	cache    Cache
-	dbClient DBClient
+	DbClient database.DBClient
```
mbarnes, line 44: nit: Any reason why this is being made public?  I see in **cmd.go** it's used to call `DBConnectionTest` but that could also be done with the local `dbClient` variable being passed to `NewFrontend`.  That's the only place I see it being externally accessed at the moment.

mbarnes (approved):
It's a little hard to isolate the significant changes from all the package shuffling.  I wonder if most of the package shuffling could have been isolated to a separate commit, before introducing the cobra dependency.  That aside, looks pretty good to me.  I'm not real familiar with cobra so I'm taking it on faith that the `main.go` changes work ok.

Had one nit comment but it's not a blocker.

frontend/pkg/frontend/frontend.go
mjlshen: Ah it's used in all of the `ArmResource*` functions as well - like anytime we need to retrieve or set something in the database.

tonytheleg (approved):

frontend/pkg/frontend/frontend.go
mbarnes: Yeah but those are all methods of `Frontend`, so they have access to private members.

mjlshen: Ah good point, I can make that change

mbarnes (approved):

tonytheleg (approved):

PR #154, opened 2024-05-30
Fix bugs with subscription state metrics

mjlshen:
### What this PR does
* MetricsMiddleware was not initializing its cache from frontend,
  resulting in dereferencing a nil pointer
* Fixes bugs around not being able to deploy/undeploy frontend as
  expected with the new REGION environment variable
* frontend_count was not consistent with the number of labels (4 or 5),
  I addressed this by setting the subscription state label to "Unknown"
  when the subscription ID is not present in the request. Open to suggestions
* `go mod tidy`

Originally implemented in #137, broke these changes out from #149 to make both both PRs easier to review.


tonytheleg (dismissed):

mbarnes (approved):

frontend/metrics.go
```diff
@@ -98,16 +98,16 @@ func (mm MetricsMiddleware) Metrics() MiddlewareFunc {
...
 		}
 
-		// Emit metrics
-		mm.Emitter.EmitCounter("frontend_count", 1.0, map[string]string{
-			"verb":        r.Method,
-			"api_version": r.URL.Query().Get(APIVersionKey),
-			"code":        strconv.Itoa(lrw.statusCode),
-			"route":       routePattern,
-		})
-
```
AldoFusterTurpin, line 110: I think we can simplify that a bit if we remove the duplication of code about the Emitter part and just assign the appropriate value to the subscriptionState, what about something like this:
```go
                subscriptionState := "Unknown"
		if subscriptionId != "" {
			sub, exists := mm.cache.GetSubscription(subscriptionId)

			if !exists {
				arm.WriteError(
					w, http.StatusBadRequest,
					arm.CloudErrorInvalidSubscriptionState, "",
					UnregisteredSubscriptionStateMessage,
					subscriptionId)
				return
			}
			subscriptionState = string(sub.State)
		}
		
		mm.Emitter.EmitCounter("frontend_count", 1.0, map[string]string{
			"verb":        r.Method,
			"api_version": r.URL.Query().Get(APIVersionKey),
			"code":        strconv.Itoa(lrw.statusCode),
			"route":       routePattern,
			"state":       subscriptionState,
		})
```

AldoFusterTurpin (commented):
LGTM. I just made an optional suggestion to remove some duplicate code and avoid the else for better readability. Thanks!

frontend/metrics.go
mjlshen: Sure, adjusted!

s-amann (approved):

tonytheleg (approved):

PR #158, opened 2024-05-30
Validate database connectivity in health probe

mjlshen:
### What this PR does
* Makes /healthz/ready fail if the frontend cannot reach the database
* The binary doesn't crash or panic, but this would allow Kubernetes to get good signal from the configured liveness/readiness Probes https://github.com/Azure/ARO-HCP/blob/bb1805bd8917c0bb4df9bece4b35097085014a1a/frontend/deploy/aro-hcp-frontend.yml#L91-L103

```bash
❯ make frontend
❯ ./aro-hcp-frontend --clusters-service-url "http://localhost:8000" --use-cache # Don't need to have CS running locally, we're just health-checking the database

# In another terminal
❯ curl -I -X GET "http://localhost:8443/healthz/ready"
HTTP/1.1 200 OK
Date: Thu, 30 May 2024 20:22:37 GMT
Content-Length: 0
```

```bash
❯ ./aro-hcp-frontend --clusters-service-url "http://localhost:8000" --cosmos-name test --cosmos-url test # Simulating unable to reach CosmosDB

# In another terminal
❯ curl -I -X GET "http://localhost:8443/healthz/ready"
HTTP/1.1 500 Internal Server Error
Date: Thu, 30 May 2024 20:23:14 GMT
Content-Length: 0
```

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -151,11 +151,9 @@ func (f *Frontend) Run(ctx context.Context, stop <-chan struct{}) {
 	}
 
 	f.logger.Info(fmt.Sprintf("listening on %s", f.listener.Addr().String()))
-
 	f.ready.Store(true)
 
-	err := f.server.Serve(f.listener)
-	if err != http.ErrServerClosed {
+	if err := f.server.Serve(f.listener); !errors.Is(err, http.ErrServerClosed) {
 		f.logger.Error(err.Error())
 		os.Exit(1)
 	}
@@ -167,7 +165,15 @@ func (f *Frontend) Join() {
 	<-f.done
 }
 
-func (f *Frontend) CheckReady() bool {
+func (f *Frontend) CheckReady(ctx context.Context) bool {
+	// Verify the DB is available and accessible
+	result, err := f.dbClient.DBConnectionTest(ctx)
+	if err != nil {
+		f.logger.Error(fmt.Sprintf("Database test failed to fetch properties: %v", err))
+		return false
+	}
+	f.logger.Debug(fmt.Sprintf("Database check completed - %s", result))
+
 	return f.ready.Load().(bool)
 }
 
@@ -180,7 +186,8 @@ func (f *Frontend) NotFound(writer http.ResponseWriter, request *http.Request) {
 
 func (f *Frontend) HealthzReady(writer http.ResponseWriter, request *http.Request) {
 	var healthStatus float64
-	if f.CheckReady() {
+
+	if f.CheckReady(request.Context()) {
 		writer.WriteHeader(http.StatusOK)
 		healthStatus = 1.0
 	} else {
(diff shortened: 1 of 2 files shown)


tonytheleg (approved):

frontend/pkg/frontend/frontend.go
```diff
@@ -168,6 +166,14 @@ func (f *Frontend) Join() {
 }
 
 func (f *Frontend) CheckReady() bool {
+	// Verify the DB is available and accessible
+	result, err := f.dbClient.DBConnectionTest(context.Background())
```
bennerv, line 170: Should we put a context timeout here, or will the overarching liveness/readiness probes cancel the http context should they exceed the threshold?

mjlshen: A timeout makes sense, I'm not sure if they get canceled, so I'll try reading up

mbarnes: I suggest have the function take a `context.Context` argument and decide the timeout in the endpoint handler.  For now we could just plumb the request's context into here.

s-amann (approved):

mbarnes (approved):

frontend/pkg/frontend/frontend.go
```diff
@@ -167,7 +165,15 @@ func (f *Frontend) Join() {
...
 }
 
-func (f *Frontend) CheckReady() bool {
+func (f *Frontend) CheckReady(ctx context.Context) bool {
+	// Verify the DB is available and accessible
+	result, err := f.dbClient.DBConnectionTest(ctx)
+	if err != nil {
+		f.logger.Error(fmt.Sprintf("Database test failed to fetch properties: %v", err))
+		return false
+	}
```
AldoFusterTurpin, line 174: Would it make more sense for DBConnectionTest() to just return an error or boolean instead of a string and a boolean? Will that string be never useful?

AldoFusterTurpin (commented):
This was just merged while I was reviewing it. Not a blocker, just a minor question. Thanks!

frontend/pkg/frontend/frontend.go
mbarnes: This is a good point, sorry for merging prematurely.

Looking over what the string return is for, I'd suggest keep the return signature but change `"No database configured, skipping"` to an empty string.  That would let us more easily distinguish a live database from no database, which _might_ be useful in the future.

AldoFusterTurpin: @mbarnes No worries 🙂 

For the case of differentiating between 
- not having the database live vs 
- not having the DB configured (or any other case)
I would rather use 2 centinel errors that we can compare against, as this string that is returned as part of the DB test call has no specific meaning. 

As I see it, the interface provides a contract and in that case, the contract includes a field that can mean or be used for anything which can lead to strange uses of it. Just my two-cents on this, not super important.

mjlshen: https://github.com/Azure/ARO-HCP/pull/162

PR #160, opened 2024-06-03
Add logging for subs changes

katherinelc321:
### What this PR does
**Before this PR:**
There is no logging for subscription doc changes
**After this PR:**
Added logging to console when the subscription documents changes. This will need to be modified later for a different logging mechanism

Jira:  ARO-6311
Link to demo recording: 

### Checklist


frontend/pkg/frontend/frontend.go
```diff
@@ -641,3 +644,50 @@ func (f *Frontend) ArmDeploymentPreflight(writer http.ResponseWriter, request *h
 
 	arm.WriteDeploymentPreflightResponse(writer, preflightErrors)
 }
+
+func (f *Frontend) logSubscriptionDifferences(oldSub, newSub *arm.Subscription) {
```
AldoFusterTurpin, line 648: Maybe we could pass a *slog.Logger directly as a parameter and do not define the method in the frontend struct as it has a lot of methods already and we do not use anything here rather than f.logger, so it could make sense to just pass what we need to simplify it a bit. We could also group the 
// Compare tenantId
and 
// Compare registeredFeatures code blocks
under the same parent "if" as it is repeated right now. Lastly, we could merge some split if in the same using &&.

Maybe something like this:
```go
func logSubscriptionDifferences(logger *slog.Logger, oldSub, newSub *arm.Subscription) {
	// Compare states
	if oldSub.State != newSub.State {
		logger.Info(fmt.Sprintf("Subscription state changed from %s to %s", oldSub.State, newSub.State))
	}

	if oldSub.Properties != nil && newSub.Properties != nil {
		// Compare tenantId
		if oldSub.Properties.TenantId != nil && newSub.Properties.TenantId != nil &&
			*oldSub.Properties.TenantId != *newSub.Properties.TenantId {
			logger.Info(fmt.Sprintf("Subscription tenantId changed from %s to %s", *oldSub.Properties.TenantId, *newSub.Properties.TenantId))
		}

		// Compare registeredFeatures
		if oldSub.Properties.RegisteredFeatures != nil && newSub.Properties.RegisteredFeatures != nil {
			oldFeatures := featuresMap(oldSub.Properties.RegisteredFeatures)
			newFeatures := featuresMap(newSub.Properties.RegisteredFeatures)

			for featureName, oldState := range oldFeatures {
				newState, exists := newFeatures[featureName]
				if !exists {
					logger.Info(fmt.Sprintf("Feature %s removed", featureName))
				} else if oldState != newState {
					logger.Info(fmt.Sprintf("Feature %s state changed from %s to %s", featureName, oldState, newState))
				}
			}
			for featureName, newState := range newFeatures {
				if _, exists := oldFeatures[featureName]; !exists {
					logger.Info(fmt.Sprintf("Feature %s added with state %s", featureName, newState))
				}
			}
		}
	}
}
```
Just an idea while reviewing it.

frontend/pkg/frontend/frontend.go
```diff
@@ -511,11 +504,21 @@ func (f *Frontend) ArmSubscriptionPut(writer http.ResponseWriter, request *http.
 		doc.Subscription = &subscription
 	}
 
+	// Log differences
+	f.logSubscriptionDifferences(doc.Subscription, &subscription)
```
AldoFusterTurpin, line 508: A small suggestion: maybe this line could be 
```go
logSubscriptionDifferences(f.logger, doc.Subscription, &subscription)
```
as we don't need anything other than the logger inside the function. Also, the frontend has a lot of methods defined right now.

frontend/pkg/frontend/frontend.go
```diff
@@ -641,3 +644,50 @@ func (f *Frontend) ArmDeploymentPreflight(writer http.ResponseWriter, request *h
...
+				if _, exists := oldFeatures[featureName]; !exists {
+					f.logger.Info(fmt.Sprintf("Feature %s added with state %s", featureName, newState))
+				}
+			}
+		}
+	}
+}
+
+func featuresMap(features *[]arm.Feature) map[string]string {
+	featureMap := make(map[string]string)
```
AldoFusterTurpin, line 686: I think if we modify a bit the code, we can pass to "make" the capacity of the map (len(*features))as a second argument as it is known beforehand (we should check first that it is not nil, though, as the range now protects against nil cases).

frontend/pkg/frontend/frontend.go
```diff
@@ -511,11 +504,21 @@ func (f *Frontend) ArmSubscriptionPut(writer http.ResponseWriter, request *http.
...
+	// Log differences
+	f.logSubscriptionDifferences(doc.Subscription, &subscription)
+
 	err = f.dbClient.SetSubscriptionDoc(ctx, doc)
 	if err != nil {
 		f.logger.Error("failed to create document for subscription %s: %v", subscriptionID, err)
 	}
 
+	// Emit the subscription state metric
+	f.metrics.EmitGauge("subscription_lifecycle", 1, map[string]string{
```
AldoFusterTurpin, line 516: I know it was done like this before, but at some point it could be a good idea to extract all the "subscription_lifecycle" and other metric names to constants on top of the file (or in a different file).

AldoFusterTurpin (commented):
Thank you for this. I just left some optional (no blocking merge) suggestions in case any makes sense to you. Thanks!

frontend/pkg/frontend/frontend.go
mbarnes: Another possibility is to wrapper `EmitGauge` with more specific use cases.  Like this case might be:
```
func (pe *PrometheusEmitter) EmitSubscriptionLifecycle(region, subscriptionID string, state arm.RegistrationState) {
        pe.EmitGauge("subscription_lifecycle", 1, map[string]string{
                "region":         region,
                "subscriptionid": subscriptionID,
                "state":          string(state),
        })
}
```

frontend/pkg/frontend/frontend.go
mbarnes: Another possibility, since all the messages are Info-level, is return a slice of message strings and let the caller forward each string to `logger.Info`.  Then the function wouldn't need a `slog.Logger` argument at all, and also wouldn't necessarily need to live in `frontend.go`.

frontend/pkg/frontend/frontend.go
```diff
@@ -511,11 +504,21 @@ func (f *Frontend) ArmSubscriptionPut(writer http.ResponseWriter, request *http.
 		doc.Subscription = &subscription
 	}
 
+	// Log differences
```
AldoFusterTurpin, line 507: IMHO this is redundant.

frontend/pkg/frontend/frontend.go
```diff
@@ -511,11 +504,21 @@ func (f *Frontend) ArmSubscriptionPut(writer http.ResponseWriter, request *http.
...
+	logSubscriptionDifferences(f.logger, doc.Subscription, &subscription)
+
 	err = f.dbClient.SetSubscriptionDoc(ctx, doc)
 	if err != nil {
 		f.logger.Error("failed to create document for subscription %s: %v", subscriptionID, err)
 	}
 
+	// Emit the subscription state metric
```
AldoFusterTurpin, line 515: IMHO this is redundant.

frontend/pkg/frontend/frontend.go
```diff
@@ -641,3 +644,51 @@ func (f *Frontend) ArmDeploymentPreflight(writer http.ResponseWriter, request *h
 
 	arm.WriteDeploymentPreflightResponse(writer, preflightErrors)
 }
+
+func logSubscriptionDifferences(logger *slog.Logger, oldSub, newSub *arm.Subscription) {
+	// Compare states
```
AldoFusterTurpin, line 649: IMHO this is redundant.

frontend/pkg/frontend/frontend.go
```diff
@@ -641,3 +644,51 @@ func (f *Frontend) ArmDeploymentPreflight(writer http.ResponseWriter, request *h
...
+	}
+
+	if oldSub.Properties != nil && newSub.Properties != nil {
+		// Compare tenantId
+		if oldSub.Properties.TenantId != nil && newSub.Properties.TenantId != nil &&
+			*oldSub.Properties.TenantId != *newSub.Properties.TenantId {
+			logger.Info(fmt.Sprintf("Subscription tenantId changed from %s to %s", *oldSub.Properties.TenantId, *newSub.Properties.TenantId))
+		}
+
+		// Compare registeredFeatures
```
AldoFusterTurpin, line 661: IMHO this is redundant.

AldoFusterTurpin (dismissed):
@katherinelc321 Great 🚀 Thank you for this and for applying the optional suggestions. Just left a few notes about some redundant comments, but not a blocker at all and LGTM! Thanks!

frontend/pkg/frontend/frontend.go
```diff
@@ -645,39 +647,41 @@ func (f *Frontend) ArmDeploymentPreflight(writer http.ResponseWriter, request *h
 	arm.WriteDeploymentPreflightResponse(writer, preflightErrors)
 }
 
-func logSubscriptionDifferences(logger *slog.Logger, oldSub, newSub *arm.Subscription) {
+func logSubscriptionDifferences(oldSub, newSub *arm.Subscription) []string {
```
AldoFusterTurpin, line 650: Last optional suggestion: change this function name to getSubscriptionsDifferences as we are no loner logging inside the function.

AldoFusterTurpin (dismissed):
LGTM! One last minor optional suggestion. Thanks!

tonytheleg (approved):

AldoFusterTurpin (dismissed):
LGTM. Thanks!

AldoFusterTurpin (approved):

PR #164, opened 2024-06-04
begins building out frontend to cs integration

tonytheleg:
### What this PR does
* adds logic to get/create/delete clusters using a local running CS
* removes any calls to frontend/cache.go in all CRUD ops to leverage CS instead
* adds ocm pkg for translating and instantiating CS Cluster objects and HCP Cluster objects
* Makefile and dev doc updates for changes
* Minor bug fixes:
  * Static validation was not accounting for operations on subscriptions like registration
  * utils/create.go updated for proper cluster input needed for CS cluster creation
  * database cache fix for SetClusterDoc as it was using wrong value

Jira: 
Link to demo recording: 

### Checklist

### DIFF
--- a/frontend/Makefile
+++ b/frontend/Makefile
@@ -13,6 +13,9 @@ REGION ?= eastus
 frontend:
 	go build -o aro-hcp-frontend .
 
+run:
+	./aro-hcp-frontend --use-cache --region eastus --clusters-service-url http://localhost:8000
+
 clean:
 	rm -f aro-hcp-frontend
 
--- a/frontend/Makefile
+++ b/frontend/Makefile
@@ -14,7 +14,7 @@ frontend:
 	go build -o aro-hcp-frontend .
 
 run:
-	./aro-hcp-frontend --use-cache --region eastus --clusters-service-url http://localhost:8000
+	./aro-hcp-frontend --use-cache --region ${REGION} --clusters-service-url http://localhost:8000
 
 clean:
 	rm -f aro-hcp-frontend
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -16,11 +16,14 @@ import (
 	"strings"
 	"sync/atomic"
 
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+
 	"github.com/google/uuid"
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
 	"github.com/Azure/ARO-HCP/frontend/pkg/database"
+	"github.com/Azure/ARO-HCP/frontend/pkg/ocm"
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 )
@@ -258,12 +261,24 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
 
 	// URL path is already lowercased by middleware.
 	resourceID := request.URL.Path
-	cluster, found := f.cache.GetCluster(resourceID)
-	if !found {
-		writer.WriteHeader(http.StatusNotFound)
-		return
+	subscriptionID := request.PathValue(PathSegmentSubscriptionID)
+	doc, err := f.dbClient.GetClusterDoc(ctx, resourceID, subscriptionID)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			f.logger.Error(fmt.Sprintf("existing document not found for cluster: %s", resourceID))
+			writer.WriteHeader(http.StatusNoContent)
+			return
+		} else {
+			f.logger.Error(err.Error())
+			writer.WriteHeader(http.StatusInternalServerError)
+			return
+		}
 	}
-	versionedResource := versionedInterface.NewHCPOpenShiftCluster(cluster)
+
+	cluster, _ := f.conn.ClustersMgmt().V1().Clusters().Cluster(doc.ClusterID).Get().Send()
+	hcpCluster, _ := ocm.ConvertCStoFrontend(*cluster.Body())
+
+	versionedResource := versionedInterface.NewHCPOpenShiftCluster(hcpCluster)
 	resp, err := json.Marshal(versionedResource)
 	if err != nil {
 		f.logger.Error(err.Error())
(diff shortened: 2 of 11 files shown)


mjlshen (dismissed):
Suggested changes that got implemented - was able to run Frontend/CS locally and confirm that I can create a subscription, create/get/delete an hcpopenshiftcluster through frontend while seeing it make the correct calls and enact the expected behavior in CS

frontend/pkg/frontend/frontend.go
```diff
@@ -258,12 +261,36 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
 
 	// URL path is already lowercased by middleware.
 	resourceID := request.URL.Path
-	cluster, found := f.cache.GetCluster(resourceID)
-	if !found {
-		writer.WriteHeader(http.StatusNotFound)
+	subscriptionID := request.PathValue(PathSegmentSubscriptionID)
+	doc, err := f.dbClient.GetClusterDoc(ctx, resourceID, subscriptionID)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
```
mbarnes, line 267: Our cluster database is still technically a cache and I'm wondering if we should account for cache misses in case the database is somehow incomplete.  Instead of failing here we could ask Cluster Service to search for the cluster with [ClustersListRequest.Search](https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go@v0.1.422/clustersmgmt/v1#ClustersListRequest.Search), and update the cache if we get a hit.  (Probably would want to break this out into a new function since multiple endpoints would need to do this.)

frontend/pkg/frontend/frontend.go
```diff
@@ -296,17 +323,55 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
...
+	resourceGroup := request.PathValue(PathSegmentResourceGroupName)
 	subscriptionID := request.PathValue(PathSegmentSubscriptionID)
 
-	cluster, updating := f.cache.GetCluster(resourceID)
-	versionedCurrentCluster := versionedInterface.NewHCPOpenShiftCluster(cluster)
+	var doc *database.HCPOpenShiftClusterDocument
+	var updating bool = true
+	doc, err = f.dbClient.GetClusterDoc(ctx, resourceID, subscriptionID)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
```
mbarnes, line 333: Similar comment as above -- maybe we want to ask Cluster Service to search to make sure the cluster _really_ doesn't exist before initiating a cluster creation.

frontend/pkg/frontend/frontend.go
```diff
@@ -399,12 +460,28 @@ func (f *Frontend) ArmResourceDelete(writer http.ResponseWriter, request *http.R
...
 	subscriptionID := request.PathValue(PathSegmentSubscriptionID)
 
-	_, found := f.cache.GetCluster(resourceID)
-	if !found {
-		writer.WriteHeader(http.StatusNotFound)
-		return
+	var doc *database.HCPOpenShiftClusterDocument
+	doc, err = f.dbClient.GetClusterDoc(ctx, resourceID, subscriptionID)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
```
mbarnes, line 466: Similar comment as above -- maybe we want to ask Cluster Service to search first.

frontend/pkg/ocm/ocm.go
```diff
@@ -0,0 +1,148 @@
...
+					ManagedResourceGroup:   cluster.Azure().ManagedResourceGroupName(),
+					SubnetID:               cluster.Azure().SubnetResourceID(),
+					OutboundType:           "",
+					NetworkSecurityGroupID: cluster.Azure().NetworkSecurityGroupResourceID(),
+					EtcdEncryptionSetID:    "",
+				},
+				IssuerURL: "",
+				ExternalAuth: api.ExternalAuthConfigProfile{
+					Enabled:       false,
+					ExternalAuths: []*v1.OIDCProvider{},
```
mbarnes, line 84: Was `ExternalAuth` intentionally skipped?  I think it's fine to skip it on the 1st pass because, honestly, it's a pain.  But we should leave a TO-DO comment or something similar here, and maybe even add a JIRA story to the backlog so it's not forgotten.

frontend/pkg/ocm/ocm.go
```diff
@@ -0,0 +1,148 @@
...
+				},
+				Ingress: []*api.IngressProfile{},
```
mbarnes, line 86: Similar question: was `Ingress` intentionally skipped?
Ingress profiles are simple enough that we should just do it now.

mbarnes (changes requested):
Had a suggestion about maybe handling cache misses in our database, so we're not mistakenly responding with "Not Found".  i.e. Check the cache for a quick answer, but still treat Cluster Service as authoritative.

Also it looks like the translation to and from Cluster Service's cluster representation has some gaps.  That's fine for now but I'd like to see a follow-up JIRA story to make sure we don't lose it.

frontend/pkg/ocm/ocm.go
```diff
@@ -0,0 +1,148 @@
...
+const (
+	csCloudProvider    string = "azure"
+	csProductId        string = "aro"
+	csHypershifEnabled bool   = true
+	csMultiAzEnabled   bool   = true
+	csCCSEnabled       bool   = true
+)
+
+// ConvertCStoFrontend converts a CS Cluster object into HCPOpenShiftCluster object
+func ConvertCStoFrontend(cluster *cmv1.Cluster) (*api.HCPOpenShiftCluster, error) {
```
AldoFusterTurpin, line 20: Minor optional suggestion: it could be interesting to rename this to _ConvertCStoHCPOpenShiftCluster_ as this could potentially be used in different places than just the frontend, and we don't need to couple here the concept as the function is not coupled with anything about the frontend, it is just called from there.

frontend/pkg/frontend/frontend.go
```diff
@@ -258,12 +261,36 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
...
+			writer.WriteHeader(http.StatusNoContent)
+			return
+		} else {
+			f.logger.Error(err.Error())
+			writer.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+	}
+
+	cluster, err := f.conn.ClustersMgmt().V1().Clusters().Cluster(doc.ClusterID).Get().Send()
```
AldoFusterTurpin, line 278: Maybe not for this PR and maybe not useful for the team, but I think it could be a good idea to put an interface (like ClustersController or something similar) between frontend and CS to provide a contract where the type the interface understands is _api.HCPOpenShiftCluster_.

That way, we could run the frontend without CS if we want with a simple in-memory struct that implements the interface. By doing so, we can run the frontend and try it without explicitly depending on CS details. To do that, we would move all the CS specifics that are in frontend.go to its package, and in NewFrontend() pass an interface as a parameter that gets assigned to a field of the frontend. 

Then we could use an env variable (like LOCAL_CS="true" with a default value of "false") that would instantiate a fake CS instead of the "real one", and we would control that in frontend/cmd/cmd.go where we are calling NewFrontend(). That way we can decouple frontend and CS and just focus on the contract in the frontend code.

AldoFusterTurpin (commented):
I just left a minor suggestion about a function name and the idea to decouple frontend and CS moving some frontend code to a different package and creating an interface.Nothing blocking this to get merged, though. Thank you!

frontend/pkg/frontend/frontend.go
tonytheleg: Talking about this with @mjlshen  and @bennerv  and @miguelsorianod  on the side. Cosmos is acting as a source of truth here because hitting CS directly without a cluster id could return multiple results back. Using something like the resource name of the cluster may not be unique enough to only get one cluster back. Sounds like CS has new support for using subscription ID and resource group name for queries that might improve that using the search method but even that may not be unique enough if customers spam create the same cluster name in same sub and rg. If they delete the cluster and recreate, it should be removed from CS but might not be right away maybe?

Maybe we leave this for now but i can create a card to make sure we revist this because its for sure something we want to get right but not sure we have all the pieces 100% today? Thoughts?

frontend/pkg/ocm/ocm.go
tonytheleg: Honestly some of these were skipped because they are a bunch of work and we don't have any values just yet so i left things out. 100% we need to track updating these to ensure all objects are set. I can add a comment and create a card if that works for you

frontend/pkg/ocm/ocm.go
tonytheleg: I can get this added

frontend/pkg/frontend/frontend.go
tonytheleg: > That way, we could run the frontend without CS if we want with a simple in-memory struct that implements the interface.

I was saying this exact thing to Michael yesterday about this, not requiring CS would be nice. Follow up PR/Card seems like a good place, i can get one created for backlog and see what folks think

frontend/pkg/frontend/frontend.go
mbarnes: I'm ok with deferring this.  Cluster Service might have their own identification scheme that's name-independent but the cluster still has to be unique to ARM, which identifies resources by subscription ID, resource group name and resource name.  So I think customers can still get name collisions at the ARM level if they're spam-creating clusters like you said.  But yes, we can revisit this after Cluster Service implements their [DDR](https://docs.google.com/document/d/19Yuu-k5IjGW1dac-tSpHDD0KTjPe3gAt_NUvHaTV9Yk/edit).

frontend/Makefile
```diff
@@ -13,6 +13,9 @@ REGION ?= eastus
 frontend:
 	go build -o aro-hcp-frontend .
 
+run:
+	./aro-hcp-frontend --use-cache --region eastus --clusters-service-url http://localhost:8000
```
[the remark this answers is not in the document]
tonytheleg: Updated

frontend/pkg/frontend/frontend.go
tonytheleg: Will get a card created today

frontend/pkg/ocm/ocm.go
mbarnes: Agree, and another thing to keep in mind is we'll need a similar function for node pools.

frontend/pkg/ocm/ocm.go
tonytheleg: Ive updated this to empty values right now as it may not be as one-to-one as i hoped. in CS its a list of Ingresses that we'd have to loop through, but URL and IP dont appear to be available values, only Visibility (in the form of ListeningMethod) so this might need to be revisted. Unless i totally missed something!

frontend/pkg/ocm/ocm.go
tonytheleg: Updated

frontend/pkg/frontend/middleware_validatestatic.go
mjlshen, line 1: I think we can revert the changes in this file since this got split out into #167

mjlshen (changes requested):

mjlshen (approved):

PR #166, opened 2024-06-05
Improving documentation and clarity

mjlshen:
### What this PR does
* Documentation and clarity updates in `frontend/pkg/database`
* typo fix in `internal/api/arm/subscription.go`
* One small bugfix in `DeleteClusterDoc` to not return an error if a cluster is not found during deletion.

### DIFF
--- a/frontend/pkg/database/database.go
+++ b/frontend/pkg/database/database.go
@@ -25,11 +25,17 @@ type DBClient interface {
 	// to be used, an error should be returned.
 	DBConnectionTest(ctx context.Context) error
 
-	GetClusterDoc(ctx context.Context, resourceID string, partitionKey string) (*HCPOpenShiftClusterDocument, error)
+	// GetClusterDoc retrieves an HCPOpenShiftClusterDocument from the database given its resourceID and containing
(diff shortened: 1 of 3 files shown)


tonytheleg (approved):

frontend/pkg/database/database.go
```diff
@@ -94,7 +100,7 @@ func (d *CosmosDBClient) DBConnectionTest(ctx context.Context) error {
 }
 
 // GetClusterDoc retrieves a cluster document from async DB using resource ID
-func (d *CosmosDBClient) GetClusterDoc(ctx context.Context, resourceID string, partitionKey string) (*HCPOpenShiftClusterDocument, error) {
+func (d *CosmosDBClient) GetClusterDoc(ctx context.Context, resourceID string, subscriptionID string) (*HCPOpenShiftClusterDocument, error) {
```
mbarnes, line 103: +1 - This is clearer.  Yes we're using subscription ID as the partition key but that should remain an internal implementation detail.

mbarnes (approved):

PR #167, opened 2024-06-05
Fix bug when validating subscription resources

mjlshen:
### What this PR does
Fixes a bug introduced in #159 where we were mistakenly trying to validate subscription resource with our HcpOpenShiftCluster resource name regex.

Before:
```bash
❯ curl -X PUT "http://localhost:8443/subscriptions/00000000-0000-0000-0000-000000000000?api-version=2.0" --json '{"state":"Registered", "properties": {"tenantId": "00000000-0000-0000-0000-000000000000"}}'
{
    "error": {
        "code": "InvalidResourceName",
        "message": "The Resource 'Microsoft.Resources/subscriptions/00000000-0000-0000-0000-000000000000' under resource group '' is invalid.",
        "target": "/subscriptions/00000000-0000-0000-0000-000000000000"
    }
}
```

After:
```bash
❯ curl -X PUT "http://localhost:8443/subscriptions/00000000-0000-0000-0000-000000000000?api-version=2.0" --json '{"state":"Registered", "properties": {"tenantId": "00000000-0000-0000-0000-000000000000"}}'
{"state":"Registered","properties":{"tenantId":"00000000-0000-0000-0000-000000000000"}}
```

### DIFF
--- a/frontend/pkg/frontend/middleware_validatestatic.go
+++ b/frontend/pkg/frontend/middleware_validatestatic.go
@@ -8,7 +8,7 @@ import (
 	"regexp"
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
-	uuid "github.com/google/uuid"
+	"github.com/google/uuid"
 
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 )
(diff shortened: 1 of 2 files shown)


mbarnes (approved):
LGTM, but I wonder if maybe we should be more selective about which endpoints we apply this particular middleware to.  Does it really make sense for the subscription endpoint?

mjlshen:
Yeah, I'm down to take out this middleware for subscription resources as well

PR #169, opened 2024-06-05
Remove cache from SubscriptionStateMuxValidator

mjlshen:
### What this PR does
* Leverage DBClient instead
* Allows us to start deprecating this cache
* Allows us to add tenantID to our request context

### DIFF
--- a/frontend/pkg/frontend/cache.go
+++ b/frontend/pkg/frontend/cache.go
@@ -38,11 +38,3 @@ func (c *Cache) GetSubscription(id string) (*arm.Subscription, bool) {
 	subscription, found := c.subscription[id]
 	return subscription, found
 }
-
-func (c *Cache) SetSubscription(id string, subscription *arm.Subscription) {
-	c.subscription[id] = subscription
-}
-
-func (c *Cache) DeleteSubscription(id string) {
-	delete(c.subscription, id)
-}
--- a/frontend/pkg/frontend/context.go
+++ b/frontend/pkg/frontend/context.go
@@ -30,7 +30,7 @@ const (
 	contextKeyVersion
 	contextKeyCorrelationData
 	contextKeySystemData
-	contextKeySubscriptionState
+	contextKeySubscription
 )
 
 func ContextWithOriginalPath(ctx context.Context, originalPath string) context.Context {
@@ -123,17 +123,33 @@ func SystemDataFromContext(ctx context.Context) (*arm.SystemData, error) {
 	return systemData, nil
 }
 
-func ContextWithSubscriptionState(ctx context.Context, subscriptionState arm.RegistrationState) context.Context {
-	return context.WithValue(ctx, contextKeySubscriptionState, subscriptionState)
+func ContextWithSubscription(ctx context.Context, subscription arm.Subscription) context.Context {
+	return context.WithValue(ctx, contextKeySubscription, subscription)
 }
 
-func SubscriptionStateFromContext(ctx context.Context) (arm.RegistrationState, error) {
-	subscriptionState, ok := ctx.Value(contextKeySubscriptionState).(arm.RegistrationState)
+func SubscriptionFromContext(ctx context.Context) (arm.Subscription, error) {
+	sub, ok := ctx.Value(contextKeySubscription).(arm.Subscription)
 	if !ok {
-		err := &ContextError{
-			got: subscriptionState,
+		return arm.Subscription{}, &ContextError{
+			got: sub,
 		}
-		return subscriptionState, err
 	}
-	return subscriptionState, nil
+	return sub, nil
+}
+
+func TenantIDFromContext(ctx context.Context) (string, error) {
+	sub, ok := ctx.Value(contextKeySubscription).(arm.Subscription)
+	if !ok {
+		return "", &ContextError{
+			got: sub,
+		}
+	}
+
+	if sub.Properties == nil || sub.Properties.TenantId == nil {
+		return "", &ContextError{
+			got: sub,
+		}
+	}
+
+	return *sub.Properties.TenantId, nil
 }
--- a/frontend/pkg/frontend/context_test.go
+++ b/frontend/pkg/frontend/context_test.go
@@ -1,0 +1,55 @@
+package frontend
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+)
+
+func stringPtr(s string) *string {
+	return &s
+}
+
+func TestTenantIDFromContext(t *testing.T) {
+	tests := []struct {
+		name      string
+		sub       arm.Subscription
+		expected  string
+		expectErr bool
+	}{
+		{
+			name: "Valid",
+			sub: arm.Subscription{
+				Properties: &arm.Properties{
+					TenantId: stringPtr("tenant-id"),
+				},
+			},
+			expected:  "tenant-id",
+			expectErr: false,
+		},
+		{
+			name:      "Missing tenantId",
+			sub:       arm.Subscription{},
+			expectErr: true,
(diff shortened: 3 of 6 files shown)


frontend/pkg/frontend/cache.go
mjlshen, line 1: After this PR, these two methods are no longer used, eventually we'll want to clean up the rest!

frontend/pkg/frontend/context.go
```diff
@@ -137,3 +138,17 @@ func SubscriptionStateFromContext(ctx context.Context) (arm.RegistrationState, e
 	}
 	return subscriptionState, nil
 }
+
+func ContextWithTenantID(ctx context.Context, sub arm.Subscription) context.Context {
```
mbarnes, line 142: Why are we passing the whole `arm.Subscription` struct here just to pluck out the `TenantId` string?

Alternatively, would it make more sense to stash the entire `arm.Subscription` struct in the context?  That could replace both `ContextWithTenantID` and `ContextWithSubscriptionState`.

mbarnes (commented):
Question and suggestion below.

frontend/pkg/frontend/context.go
mjlshen: Just for typing/prevent user error sending the wrong string. `SubscriptionState` has its own unique type already to mitigate the issue, but `TenantID` is just a vanilla string. Definitely no harm just folding the two together though, I'll do that.

frontend/pkg/frontend/middleware_validatesubscription.go
```diff
@@ -49,10 +51,10 @@ func (s *SubscriptionStateMuxValidator) MiddlewareValidateSubscriptionState(w ht
 		return
 	}
 
-	// the subscription exists, store its current state as context
-	ctx := ContextWithSubscriptionState(r.Context(), sub.State)
+	ctx := ContextWithTenantID(r.Context(), *sub.Subscription)
+	ctx = ContextWithSubscriptionState(ctx, sub.Subscription.State)
```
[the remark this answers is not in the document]
mjlshen: I had them separate to get some typing assistance/avoid passing the wrong string, but I'll fold these two methods into one to store the whole subscription in the context instead.

frontend/pkg/frontend/context.go
mbarnes: To your point about typing, I wonder if a type alias in the `api/arm` package would be useful.
```
type TenantID string
```

frontend/pkg/frontend/context.go
```diff
@@ -123,17 +123,33 @@ func SystemDataFromContext(ctx context.Context) (*arm.SystemData, error) {
 	return systemData, nil
 }
 
-func ContextWithSubscriptionState(ctx context.Context, subscriptionState arm.RegistrationState) context.Context {
-	return context.WithValue(ctx, contextKeySubscriptionState, subscriptionState)
+func ContextWithSubscription(ctx context.Context, subscription arm.Subscription) context.Context {
+	return context.WithValue(ctx, contextKeySubscription, subscription)
 }
 
 func SubscriptionStateFromContext(ctx context.Context) (arm.RegistrationState, error) {
```
mbarnes, line 130: Sorry, I should have been clearer.  Implicit in my suggestion was to also replace this and `TenantIDFromContext` with `SubscriptionFromContext`.  Unless you disagree?

mbarnes (dismissed):
One more change requested.  My bad for expecting you to read my mind.

frontend/pkg/frontend/context.go
mjlshen: No qualms with adding `SubscriptionFromContext` - I can change it

mjlshen: I felt like keeping `TenantIDFromContext` could make sense to centralize the error handling around the optional fields, what do you think?

mbarnes: SGTM  :+1:

tonytheleg (approved):

frontend/pkg/frontend/context_test.go
```diff
@@ -0,0 +1,55 @@
...
+		} else {
+			if test.expectErr {
+				t.Error("expected err to be non-nil")
+			}
+
+			if actual != test.expected {
+				t.Errorf("expected %s, got %s", test.expected, actual)
+			}
+		}
+	}
```
AldoFusterTurpin, line 54: Minor optional suggestion: as this is a unit test, every time we find an error we could do a t.Fatalf and we can remove the elses and nesting (we also fail sooner, which is good IMHO)

frontend/pkg/frontend/middleware_validatesubscription_test.go
```diff
@@ -155,9 +155,21 @@ func TestMiddlewareValidateSubscription(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			dbClient := database.NewCache()
+			middleware := NewSubscriptionStateMuxValidator(dbClient)
 
 			if tt.cachedState != "" {
-				cache.SetSubscription(subscriptionId, &arm.Subscription{State: tt.cachedState})
+				if err := dbClient.SetSubscriptionDoc(context.Background(), &database.SubscriptionDocument{
```
AldoFusterTurpin, line 162: Maybe we could move &database.SubscriptionDocument to a variable to improve readability.

frontend/pkg/frontend/middleware_validatesubscription_test.go
```diff
@@ -178,25 +190,22 @@ func TestMiddlewareValidateSubscription(t *testing.T) {
...
-				if !reflect.DeepEqual(result, tt.expectedState) {
-					t.Error(cmp.Diff(result, tt.expectedState))
+			sub, err := SubscriptionFromContext(request.Context())
+			if err != nil {
+				if tt.expectedError != nil {
+					var actualError *arm.CloudError
+					body, _ := io.ReadAll(http.MaxBytesReader(writer, writer.Result().Body, 4*megabyte))
+					_ = json.Unmarshal(body, &actualError)
+					if (writer.Result().StatusCode != tt.expectedError.StatusCode) || actualError.Code != tt.expectedError.Code || actualError.Message != tt.expectedError.Message {
+						t.Errorf("unexpected CloudError, wanted %v, got %v", tt.expectedError, actualError)
```
AldoFusterTurpin, line 200: Same as before, I think we can avoid the else if we do t.Fatalf.

AldoFusterTurpin (approved):
LGTM. Thanks! Just a couple of optional suggestions.

frontend/pkg/frontend/context_test.go
mjlshen: I think I get what you're saying, but I think the elses are necessary for a passing test. Like

```go
// Let's say err == test.expectErr from a test case
if err != nil {
  if !test.expectErr {}
  }
}
// If no else, it would keep checking, with the else these checks are skipped
if test.expectErr {}
if actual != test.expected {}
```

AldoFusterTurpin: But we can always add a "return" inside the if (just at the end, right), like we usually do in Go to remove else's ? 🙂

mjlshen: I don't think `return` would work since there are multiple test cases, `continue` might in this case because I noticed I forgot to start each test case with
```go
t.Run(test.name, func(t *testing.) { ... }
```
, but if I remembered to do that `continue` wouldn't work either!

AldoFusterTurpin: You are right 🙂
I have opened a small PR with a refactor to use subtests and remove else: https://github.com/Azure/ARO-HCP/pull/191

PR #173, opened 2024-06-06
postgres flexible server bicep template

geoberle:
### What this PR does

this PR introduces several bicep modules to
* create a basic postgres flexible server instance and database - `postgres.bicep`
* create a database user based on a managed identity and grant access to the database - `postgres-access.bicep`
* orchestrate the creation of the DB, the user and the placement of configmap to an AKS cluster for CS usage

Jira: https://issues.redhat.com/browse/ARO-7985

the user creation on the database is not supported via ARM, therefore a deploymentscript is currently used to implement this. after discussing the usage of deploymentscripts with MSFT, we will investigate how this can be refactored so that the access management to the database will be done with shell scripts after the bicep run in a way that is compatible with
* production rollouts via EV2
* development environments provisioning via Makefile

cc @whober0521 @nwnt

### DIFF
--- a/dev-infrastructure/configurations/svc-cluster.bicepparam
+++ b/dev-infrastructure/configurations/svc-cluster.bicepparam
@@ -15,7 +15,7 @@ param maestroNamespace = 'maestro'
 param maestroKeyVaultName = take('maestro-kv-${uniqueString(currentUserId)}', 24)
 param maestroEventGridNamespacesName = '${maestroInfraResourceGroup}-eventgrid'
 param maestroCertDomain = 'selfsigned.maestro.keyvault.aro-int.azure.com'
-param deployCsInfra = true
+param deployCsInfra = false
 param csNamespace = 'cluster-service'
 param workloadIdentities = items({
   frontend_wi: {
--- a/dev-infrastructure/configurations/svc-cluster.bicepparam
+++ b/dev-infrastructure/configurations/svc-cluster.bicepparam
@@ -17,6 +17,7 @@ param maestroEventGridNamespacesName = '${maestroInfraResourceGroup}-eventgrid'
 param maestroCertDomain = 'selfsigned.maestro.keyvault.aro-int.azure.com'
 param deployCsInfra = false
 param csNamespace = 'cluster-service'
+param csPostgresServerName = take('cs-pg-${uniqueString(currentUserId)}', 60)
 param workloadIdentities = items({
   frontend_wi: {
     uamiName: 'frontend'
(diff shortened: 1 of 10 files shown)


dev-infrastructure/docs/development-setup.md
```diff
@@ -346,4 +346,28 @@ To tear down your CS setup:
...
+psql
+```
+
+> The password is a temporary access token that is valid only for 1h
+
+### Access the database from inside of the AKS cluster
+
+* the `cs` namespace of the svc cluster contains
+  * a `ConfigMap` named `cs` with the connection parameters
+  * a `ServiceAccount` named `cs` that is annotated for workload identity usage
```
markturansky, line 370: can we name these `cs` resources something else?  `cs` is terribly generic

markturansky: whatever the current maps/secrets are named in commercial is probably easiest to keep in sync across multiple templates

dev-infrastructure/modules/cs.bicep
```diff
@@ -0,0 +1,138 @@
...
+    name: postgresServerName
+    databaseAdministrators: [
+      // add the dedicated admin managed identity as administrator
+      // this one is going to be used to manage DB access
+      {
+        principalId: postgresAdminManagedIdentity.properties.principalId
+        principalName: postgresAdminManagedIdentity.name
+        principalType: 'ServicePrincipal'
+      }
+      // use the current user as DB admin if defined - for dev purposes
```
weinong, line 44: can we add a // TODO: clean up when we release to prod?

geoberle: i removed it entirely from the bicep files.
admin access to the database for dev purposes can now be aquired on-demand via the makefile target `cs-out-of-cluster-db-access`

dev-infrastructure/docs/development-setup.md
geoberle: adapted

dev-infrastructure/configurations/svc-cluster.bicepparam
```diff
@@ -11,6 +11,8 @@ param disableLocalAuth = false
 param deployFrontendCosmos = false
 param deployMaestroInfra = false
 param maestroNamespace = 'maestro'
+param deployCsInfra = true
```
mjlshen, line 14:
```suggestion
param deployCsInfra = false
```
I think we want this to default to false by convention

mjlshen (commented):
Looks good overall, small nit

dev-infrastructure/configurations/svc-cluster.bicepparam
geoberle: good point. fixed

mjlshen (dismissed):

tonytheleg (approved):

tonytheleg (approved):

PR #182, opened 2024-06-06
Update frontend and internal gomod with dependabot

mjlshen:
Putting dependabot to work 💪


mbarnes (approved):
Not familiar with the config schema here, but +1 on the intent.

tonytheleg (approved):

PR #188, opened 2024-06-07
Pin OutboundType to LoadBalancer as it is the only valid option

mjlshen:
### What this PR does
Currently this is the only valid value for this field: https://github.com/Azure/ARO-HCP/blob/683f939a2b23282796b2f91065e4ea20463e7c07/internal/api/enums.go#L14-L19

Jira: [ARO-7958](https://issues.redhat.com/browse/ARO-7958)


bennerv (approved):
good for now; will change later

PR #209, opened 2024-06-12
adds sub get/put integration tests

tonytheleg:
### What this PR does
**Before this PR:**
* no e2e tests for subscriptions calls on frontend

**After this PR:**
* adds test for subscription get/put
* minor bug fix where incorrect status codes were returned on some calls (thanks test for finding those!)

Jira: 
Link to demo recording: 

### Checklist

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -463,7 +463,7 @@ func (f *Frontend) ArmSubscriptionGet(writer http.ResponseWriter, request *http.
 	if err != nil {
 		if errors.Is(err, database.ErrNotFound) {
 			f.logger.Error(fmt.Sprintf("document not found for subscription %s", subscriptionID))
-			writer.WriteHeader(http.StatusNoContent)
+			writer.WriteHeader(http.StatusNotFound)
 			return
 		} else {
 			f.logger.Error(err.Error())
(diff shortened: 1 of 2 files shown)


frontend/pkg/frontend/frontend.go
```diff
@@ -463,7 +463,7 @@ func (f *Frontend) ArmSubscriptionGet(writer http.ResponseWriter, request *http.
 	if err != nil {
 		if errors.Is(err, database.ErrNotFound) {
 			f.logger.Error(fmt.Sprintf("document not found for subscription %s", subscriptionID))
-			writer.WriteHeader(http.StatusNotFound)
+			writer.WriteHeader(http.StatusNoContent)
```
mjlshen, line 466: I'm not 100% sure if we're supposed to return a 404 or a 204 if we attempt to get a subscription that doesn't exist

mjlshen (commented):
lgtm, just one spot I'm not sure about

frontend/pkg/frontend/frontend.go
tonytheleg: My understanding is based on a comment @mbarnes left me on another PR but it was specifically for a Delete call so maybe its different when its a GET: 

>The resource provider can return 200 (OK) or 204 (NoContent) to indicate that the operation completed successfully. A 200 (OK) should be returned if the object exists and was deleted successfully; and a 204 (NoContent) should be used if the resource does not exist and the request is well formed.

mjlshen: I think this is the relevant documentation: https://armwiki.azurewebsites.net/rpaas/rpaasextensionapi.html

I'm just learning, so take this with a grain of salt, but if I'm understanding this doc correctly, there are certain cases where a 404 seems relevant:

> 404: returned by MetaRP if resource group or subscription not found. ResourceCreationBegin will not be called in this case.

tonytheleg: I reverted this back since i explicitly changed it and at some point someone purposefully put this, its probably accurate haha.

mjlshen (approved):

PR #210, opened 2024-06-12
rename diskSizeGB to diskSizeGiB

mjlshen:
### What this PR does
Azure disks are measured in GiB
https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types

This also aligns us with ocm-sdk-go https://github.com/openshift-online/ocm-sdk-go/blob/5991978bab9cd8093d06e3d83bc90c29322e11f1/clustersmgmt/v1/azure_node_pool_type.go#L27 and gets rid of a TODO from #201 

Jira: [ARO-7592](https://issues.redhat.com/browse/ARO-7592)


tonytheleg (approved):

mbarnes (approved):
Approved but also cc @petrkotas as this change needs to be propagated to https://github.com/Azure/azure-rest-api-specs-pr/pull/18316

PR #215, opened 2024-06-13
Fix typo

mjlshen:
* There was a missing trailing slash in #207 
* Aligned the bicep-what-if and build-dev-infra actions to ensure they're trying to apply the same thing


tonytheleg (dismissed):
Assuming the error is related to bad apply? Approving if so to unblock fixing it if needed

mjlshen:
> Assuming the error is related to bad apply? Approving if so to unblock fixing it if needed

The what-if and apply actions weren't doing the exact same thing, so the failure was valid - should be fixed now

tonytheleg (approved):

PR #226, opened 2024-06-14
Ensure GitHub Action can authenticate to the dev service cluster

mjlshen:
### What this PR does
Depends on #227

Lays the groundwork for GitHub Actions deploying frontend to a service cluster

PR #227, opened 2024-06-14
Grant GitHub Action Service Principal AKS access 

mjlshen:
### What this PR does
* This will allow us to create a GitHub Action to deploy resources to
  relevant AKS clusters (#226)

Jira: [ARO-8285](https://issues.redhat.com/browse/ARO-8285)

### DIFF
--- a/.github/workflows/build-dev-infra.yml
+++ b/.github/workflows/build-dev-infra.yml
@@ -62,10 +62,11 @@ jobs:
               --parameters persist="true"
 
             SVC_CLUSTER_NAME=$(az deployment group show --resource-group "${RESOURCEGROUP}" --name "svc-cluster-${GITHUB_RUN_ID}" --output tsv --query properties.outputs.aksClusterName.value)
+
             # service cluster rolebindings
             az deployment group create \
               --name "svc-rolebindings-${GITHUB_RUN_ID}" \
               --resource-group "${RESOURCEGROUP}" \
               --template-file templates/dev-aks-roleassignments.bicep \
               --parameters aksClusterName=${SVC_CLUSTER_NAME} \
-              --parameters location="eastus"
+              --parameters githubActionsPrincipalID=${{ secrets.GHA_PRINCIPAL_ID }}
--- a/.github/workflows/build-dev-infra.yml
+++ b/.github/workflows/build-dev-infra.yml
@@ -61,9 +61,11 @@ jobs:
               --parameters deployFrontendCosmos="true" \
               --parameters persist="true"
 
-#            # management cluster
-#            az deployment group create \
-#              --name "mgmt-cluster-${GITHUB_RUN_ID}" \
-#              --resource-group "${RESOURCEGROUP}" \
-#              --template-file templates/mgmt-cluster.bicep \
-#              --parameters configurations/mgmt-cluster.bicepparam \
+            SVC_CLUSTER_NAME=$(az deployment group show --resource-group "${RESOURCEGROUP}" --name "svc-cluster-${GITHUB_RUN_ID}" --output tsv --query properties.outputs.aksClusterName.value)
+            # service cluster rolebindings
+            az deployment group create \
+              --name "svc-rolebindings-${GITHUB_RUN_ID}" \
+              --resource-group "${RESOURCEGROUP}" \
+              --template-file templates/dev-aks-roleassignments.bicep \
+              --parameters aksClusterName=${SVC_CLUSTER_NAME} \
+              --parameters location="eastus"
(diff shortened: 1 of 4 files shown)


dev-infrastructure/templates/dev-aks-roleassignments.bicep
```diff
@@ -0,0 +1,28 @@
+// This is used to grant CI the ability to deploy resources into
+// dev AKS clusters. It should not be used in higher environments.
+param aksClusterName string
+param location string
+
+var githubActionsAppID = '3aa9b94b-acfc-4e7e-b952-158a174cb1bb'
```
bennerv, line 6: Is this the application id, or the principal id of the service principal?  RBAC is defined on the latter.  

We should probably drop it as a secret into the GH repo as well, or we can query it using MSGraph using the application id.

.github/workflows/build-dev-infra.yml
```diff
@@ -61,9 +61,11 @@ jobs:
...
-#              --template-file templates/mgmt-cluster.bicep \
-#              --parameters configurations/mgmt-cluster.bicepparam \
+            SVC_CLUSTER_NAME=$(az deployment group show --resource-group "${RESOURCEGROUP}" --name "svc-cluster-${GITHUB_RUN_ID}" --output tsv --query properties.outputs.aksClusterName.value)
+            # service cluster rolebindings
+            az deployment group create \
+              --name "svc-rolebindings-${GITHUB_RUN_ID}" \
+              --resource-group "${RESOURCEGROUP}" \
+              --template-file templates/dev-aks-roleassignments.bicep \
+              --parameters aksClusterName=${SVC_CLUSTER_NAME} \
+              --parameters location="eastus"
```
bennerv, line 71: can we pull location into a var used for GH actions, or set it higher up?  

It looks like we're using it in a couple places and we may want to change this at some point to deploy to a region that doesn't have 3 azs.

dev-infrastructure/templates/dev-aks-roleassignments.bicep
mjlshen: ack, I tested the RBAC grant works, so it must be the principal ID. Bicep has a preview feature for MSGraph, so we can try that as well: https://learn.microsoft.com/en-us/graph/templates/overview-bicep-templates-for-graph

tonytheleg: Based on [THIS](https://learn.microsoft.com/en-us/graph/templates/quickstart-create-bicep-interactive-mode?tabs=CLI) using Graph with bicep is a preview feature, assuming we cant control that with github actions? For now ill update the variable name so it reflects the principal and add definitions to manifest, but someone would need to add the secret for me

tonytheleg: changed to a param called `githubActionsPrincipalID`, this way we can define it using `--parameters` and the github secret value

.github/workflows/build-dev-infra.yml
tonytheleg: everywhere else we define `location` based on the resource group location
`param location string = resourceGroup().location`

updated the bicep template to do the same so that the `--parameters location="eastus"` flag could be removed and location will always be based on the resource group, allowing for us to target other regions

.github/workflows/build-dev-infra.yml
```diff
@@ -61,9 +61,12 @@ jobs:
...
-#              --parameters configurations/mgmt-cluster.bicepparam \
+            SVC_CLUSTER_NAME=$(az deployment group show --resource-group "${RESOURCEGROUP}" --name "svc-cluster-${GITHUB_RUN_ID}" --output tsv --query properties.outputs.aksClusterName.value)
+
+            # service cluster rolebindings
+            az deployment group create \
+              --name "svc-rolebindings-${GITHUB_RUN_ID}" \
+              --resource-group "${RESOURCEGROUP}" \
+              --template-file templates/dev-aks-roleassignments.bicep \
+              --parameters aksClusterName=${SVC_CLUSTER_NAME} \
+              --parameters githubActionsPrincipalID=${{ secrets.GITHUB_ACTIONS_PRINCIPAL_ID }}
```
mjlshen, line 72:
```suggestion
              --parameters githubActionsPrincipalID=${{ secrets.GHA_PRINCIPAL_ID }}
```
Just because `Secret names must not start with GITHUB_.`

tonytheleg: Updated!

mjlshen (commented):
/lgtm

tonytheleg (approved):

PR #234, opened 2024-06-17
adds alternative deployment method using kustomize

tonytheleg:
### What this PR does
* Updates `ARO_HCP_BASE_IMAGE` in makefile to target new dev ACR
* Adds kustomize manifests for deploying frontend to AKS using Github actions
  * per [ARO-8285](https://issues.redhat.com/browse/ARO-8285?focusedId=24934691&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24934691) using oc template is not well supported with actions, the current `ubuntu-latest` runner has both kubectl and kustomize installed, making this easier to use
* updates actual deployment to use configmap for key variables since we are going to make one anyway (more details later)
* adds new make targets for updating values used in kustomize manifests and targets for deploying/deleting the frontend deployment using kustomize (kubectl -k)

Jira: [ARO-8285](https://issues.redhat.com/browse/ARO-8285)

I've tested that this all works as expected, validation info available in the [card](https://issues.redhat.com/browse/ARO-8285?focusedId=24945623&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24945623)

Some design decisions that may be useful to know:
* I left the original deployment template for now so that folks testing with personal clusters can still use the old `deploy` and `undeploy` make targets without needing to install `kustomize` cli -- baby steps
* Kustomize doesnt really support variables that  change (DB name is specific to user or dev, image name changes every time, etc), it works better with hardcoded values
  * The concept of using a configmap generator and replacement definitions is a workaround many have used for things like CI/CD to allow for dynamic values
  * using a configmap generator, allows us to use the kustomize CLI to update the values in file automatically using env vars we're already setting using Makefile making it easier to update in CI/CD without hardcoding those values in git (changes are never commited by pipeline)
  * to facilitate all the configurations we need, there are values in the config map we don't really use in the deployment -- if we can find a better way thats great but for now this is a working MVP


mjlshen (approved):

frontend/Makefile
```diff
@@ -31,6 +31,28 @@ endif
 push: image
 	docker push ${ARO_HCP_FRONTEND_IMAGE}
 
+
+kustomize-update:
+	@test "${RESOURCEGROUP}" != "" || (echo "RESOURCEGROUP must be defined" && exit 1)
```
mbarnes, line 36: For convenience, can we have a default resource group that matches dev-infrastructure?

mbarnes (approved):
Minor question, but this looks good to me too!

PR #235, opened 2024-06-17
added user agent pool

mukrishn:
### What this PR does

Included additional agentpool for application pods, these userpools are meant for application(HCP) pods and existing system pools are for hosting system critical pods only.  we can treat the system pools as Infra nodes in Openshift cluster.

This PR adds an additional agentpool to the cluster.

As recommended in [this](https://learn.microsoft.com/en-us/azure/aks/use-system-pools?tabs=azure-cli#system-and-user-node-pools) doc, system agent should be present as it has its benefits to host system critical pods.

They run operator pods on system pools,  ex - on installing and configuring Azure Managed Prometheus to collect metrics, `ama-metrics-operator-targets-6878d6665d-lb6rv` runs on system pool and leave the userpool for user application.

### DIFF
--- a/dev-infrastructure/modules/aks-cluster-base.bicep
+++ b/dev-infrastructure/modules/aks-cluster-base.bicep
@@ -1,8 +1,16 @@
 // Constants
 param aksClusterName string = take('aro-hcp-${clusterType}-${uniqueString(clusterType)}', 63)
-param agentMinCount int = 2
-param agentMaxCount int = 3
-param agentVMSize string = 'Standard_D2s_v3'
+
+// System agentpool spec(Infra)
+param systemAgentMinCount int = 2
+param systemAgentMaxCount int = 3
+param systemAgentVMSize string = 'Standard_D2s_v3'
+
+// User agentpool spec (Worker)
+param userAgentMinCount int = 2
+param userAgentMaxCount int = 3
+param userAgentVMSize string = 'Standard_D2s_v3'
+
 param serviceCidr string = '10.130.0.0/16'
 param dnsServiceIP string = '10.130.0.10'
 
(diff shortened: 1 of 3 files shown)


mukrishn:
@mjlshen @tonytheleg I am adding extra agent pool to the cluster using the template, it works on my local, not sure about this error. could you please take a look ? 
```
ERROR: InvalidTemplateDeployment - The template deployment '9553501288' is not valid according to the validation procedure. The tracking id is '56289969-55ca-4d85-9048-46d658bf7b4e'. See inner errors for details.
BadRequest - Preflight validation check for resource(s) for container service aro-hcp-svc-cluster-zw5yioe62vtuu in resource group aro-hcp-dev failed. Message: A new agent pool was introduced. Adding agent pools to an existing cluster is not allowed through managed cluster operations. For agent pool specific change, please use per agent pool operations: https://aka.ms/agent-pool-rest-api. Details:
```

tonytheleg:
> @mjlshen @tonytheleg I am adding extra agent pool to the cluster using the template, it works on my local, not sure about this error. could you please take a look ?
> 
> ```
> ERROR: InvalidTemplateDeployment - The template deployment '9553501288' is not valid according to the validation procedure. The tracking id is '56289969-55ca-4d85-9048-46d658bf7b4e'. See inner errors for details.
> BadRequest - Preflight validation check for resource(s) for container service aro-hcp-svc-cluster-zw5yioe62vtuu in resource group aro-hcp-dev failed. Message: A new agent pool was introduced. Adding agent pools to an existing cluster is not allowed through managed cluster operations. For agent pool specific change, please use per agent pool operations: https://aka.ms/agent-pool-rest-api. Details:
> ```

Have you tested updating an existing cluster locally?:
1) Create a cluster without agent pool
2) Add agent pool config to your bicep
3) Run bicep again to see if it can add agent pool to existing cluster

Now that we have long-lived infra, including a cluster, this bicep what-if test checks against our existing cluster to make sure it can run before allowing merge.  If updating an existing cluster works locally, possibly this is a bicep version issue with github actions?

I think you can also define the agent pool separately in same bicep and mark the parent of it as the AKS cluster
https://learn.microsoft.com/en-us/azure/templates/microsoft.containerservice/managedclusters/agentpools?pivots=deployment-language-bicep

```
resource symbolicname 'Microsoft.ContainerService/managedClusters/agentPools@2024-03-02-preview' = {
  name: 'AgentPoolName'
  parent: aksCluster
  properties: {...}
```

mukrishn:
Thanks @tonytheleg, I have not tried update scenario, it works on a fresh cluster install with 2 agentpools. I will try with recommended changes.

mukrishn:
It worked after making it using resource types, PTAL when you get a chance.
Thanks!!

mjlshen (approved):
@mukrishn I added a commit to add and wire a `deployUserAgentPool` param so that the user pool only gets added to the management cluster. If that sounds good to you I'm happy to merge, thanks!

PR #252, opened 2024-06-20
deploy maestro server to dev service cluster

geoberle:
### What this PR does

as part of MVP we need the maestro-server running on the service cluster

Jira: 
Link to demo recording:

### DIFF
--- a/maestro/deploy/openshift-templates/namespace-template.yml
+++ b/maestro/deploy/openshift-templates/namespace-template.yml
@@ -1,0 +1,26 @@
+---
+apiVersion: v1
+kind: Template
+metadata:
+  name: maestro-ns
+parameters:
+
+  - name: NAMESPACE
+    description: The OpenShift Namespace where the resources will be created.
+    displayName: Namespace
+    required: true
+    value: maestro
+
+  - name: ISTIO_VERSION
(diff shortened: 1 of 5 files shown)


tonytheleg (approved):

maestro/deploy/openshift-templates/namespace-template.yml
```diff
@@ -0,0 +1,26 @@
...
+    required: true
+
+objects:
+
+  - apiVersion: v1
+    kind: Namespace
+    metadata:
+      name: ${NAMESPACE}
+      labels:
+        istio.io/rev: ${ISTIO_VERSION}
```
mjlshen, line 26: I'm not sure if we want to "manage" this istio version label on namespaces in the future because during an istio upgrade we would want to make sure this isn't getting applied to the old version, but lgtm for now

mjlshen (approved):

maestro/deploy/openshift-templates/namespace-template.yml
geoberle: i agree. istio upgrads and sidecar injection annotations need to be coordinated eventually if we decide to go with istio.

PR #253, opened 2024-06-20
ARO-8287 add smoke tests

tonytheleg:
### What this PR does
* adds smoke tests that are intended to be run against a live frontend deployment with cosmos
  * intention is to add a github action that will run this against DEV to ensure changes don't break frontend functionality
  * build tag added to smoketest so that it doesnt run with other tests since it requires live infra
* github action to follow in a subsequent PR

Jira: ARO-8287
Link to demo recording: Coming soon!


frontend/utils/frontend_smoke_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"net/http"
+	"os"
+	"strings"
+	"testing"
+)
+
+const (
+	apiURL           = "http://localhost:8443"
+	testSubscription = "00000000-0000-0000-0000-000000000000"
+	baseResourceID   = "/subscriptions/" + testSubscription + "?api-version=2.0"
```
tonytheleg, line 22: Realizing this is too specific to subscriptions, when we do clusters this wont be as universal....going to tweak this

tonytheleg: fixed

mjlshen (approved):
The tests look good. We're just testing subscription actions which do not interact with any downstream components.

PR #255, opened 2024-06-21
refactors CS calls to simplify their use

tonytheleg:
### What this PR does
* minor refactor to just clean up CS calls and make it easier to use the same calls more than once when needed (likely will)

Jira: 
Link to demo recording:


frontend/pkg/frontend/ocm.go
```diff
@@ -254,3 +258,30 @@ func (f *Frontend) BuildCSNodepool(ctx context.Context, nodepool *api.HCPOpenShi
 
 	return npBuilder.Build()
 }
+
+// GetCSCluster creates and sends a GET request to fetch a cluster from Clusters Service
+func (f *Frontend) GetCSCluster(doc *database.HCPOpenShiftClusterDocument) (*cmv1.ClusterGetResponse, error) {
```
mjlshen, line 263:
```suggestion
func (f *Frontend) GetCSCluster(clusterID string) (*cmv1.ClusterGetResponse, error) {
```
I think this makes more sense/aligns with `DeleteCSCluster` as well - though you'll also need to update wherever this function is used.

mjlshen (changes requested):

frontend/pkg/frontend/ocm.go
tonytheleg: yeah good call, that makes sense to me too. Fix coming.

tonytheleg: Updated

mjlshen (approved):

PR #269, opened 2024-06-25
reverts workflow triggering, adds smoketests

tonytheleg:
### What this PR does
* workflow triggers are adding too much complexity, plus it appears to require secrets defined per branch when using AZ cli
* reverting back to last working workflows but adding pr check job

Jira: 
Link to demo recording:


tonytheleg:
While i was able to get the PR Check pipeline to work, the tests fail due to frontend not being able to talk to Cosmos. This is because workload identities are namespace locked, and targeting a new namespace each run would require updating that value for the namespace. We will revisit this later in the project, when we move towards a fully automated test of spinning up infra and running tests on expected namespaces/DB's

mjlshen (approved):

PR #277, opened 2024-06-26
feat: get/put/delete node_pools in frontend

AldoFusterTurpin:
### What this PR does
- It creates handlers and routes to GET, PUT and DELETE node pools of clusters.

 JIRA [issue](https://issues.redhat.com/browse/ARO-7592)

I have tested all the endpoints using cURL. There will be a follow-up work to refactor it to improve the design and testability of this and the other endpoints.

I have added the option to create a NodePool in the create.go utility.

I have also added a small script to teardown + configure + start CS and updated the README of the frontend as we need a couple of config files that Miguel (from CS) shared with me.

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"sync/atomic"
 
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/google/uuid"
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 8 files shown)


frontend/pkg/frontend/frontend.go
```diff
@@ -287,7 +361,7 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
 
 	var hcpCluster *api.HCPOpenShiftCluster
 	var csResp *cmv1.ClusterGetResponse
-	if doc.ClusterID != "" {
+	if doc.ClusterID != "" { // redundant if check ? at this point doc.ClusterID can't be "", otherwise error raised.
```
tonytheleg, line 364: This is likely old code before a lot of the error handling above it was defined. +1 that it can go away

tonytheleg: Actually i take that back -- the reason is there is if we dont find a cluter document, it means we don't have a cluster and clusterID would be empty. So we dont want to call CS to get the cluster if it doesnt exist yet. Its logic to skip execution which would fail

frontend/pkg/frontend/frontend.go
```diff
@@ -237,6 +237,80 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
...
+		if errors.Is(err, database.ErrNotFound) {
+			f.logger.Error(fmt.Sprintf("existing cluster document not found for cluster: %s on GET node pool by name", resourceID))
+			writer.WriteHeader(http.StatusNoContent)
+			return
+		} else {
+			f.logger.Error(err.Error())
+			arm.WriteInternalServerError(writer)
+			return
+		}
+	}
```
mbarnes, line 266: I think it would be worth adding the Cluster ID to the node pool document so only one CosmosDB query is required to retrieve node pool data from Cluster Service.  Would simplify the logic both here and in the DELETE handler.

frontend/pkg/frontend/frontend.go
```diff
@@ -237,6 +237,80 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
...
+		}
+
+		f.logger.Error(err.Error())
+		arm.WriteInternalServerError(writer)
+		return
+	}
+
+	targetNodePoolName := nodePoolDoc.ID
+	// nodePoolGetResponse, err := f.conn.ClustersMgmt().V1().Clusters().Cluster(clusterDoc.ClusterID).NodePools().NodePool(targetNodePoolName).Get().Send()
+	nodePoolGetResponse, err := f.conn.ClustersMgmt().V1().Clusters().Cluster(clusterDoc.ClusterID).NodePools().NodePool(targetNodePoolName).Get().Send()
```
mbarnes, line 283: Not an ask for this pull request, but I wonder if there would be benefit in defining a CSClient interface similar to our DBClient interface, so that Cluster Service could be mocked for testing.

mbarnes (changes requested):
Tell me if you disagree, but I think the handler logic here could be simplified by including the Cluster Service cluster ID in `NodePoolDocument`.

frontend/pkg/frontend/frontend.go
tonytheleg: +1 to adding cluster id to node pool doc

I dont know if there is an ADR/DDR i missed but is it possibly more beneficial to leverage the cluster docs to store node pool information and forgo node pool docs altogether? Since node pools are tied to clusters, getting the cluster doc would provide us the ID and potentially all node pools tied to it? 

maybe not for MVP but a thought.

tonytheleg:
A general comment, you mentioned a refactor following this, just want to +1 that, it would be good to stick with just the ArmACTIONNAME functions defined in frontend.go and migrate any functions for node pool work to respective files (database.go, document.go, ocm.go)

frontend/pkg/frontend/frontend.go
mbarnes: I'm not sure I would combine multiple resources into one document, since that runs a higher risk of database commit conflicts.  But we could potentially combine the cluster and node pool database containers into a more generalized "resource" container.  The only significant schema change would be capturing how to uniquely identify any Cluster Service resource.  My suggestion there is to use their own [REST API](https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id__node_pools__node_pool_id_) as a guide.

e.g.
```
ExternalID: {azure_resource_id}
InternalID: "/clusters/{cluster_id}/node_pools/{node_pool_id}"
```

frontend/pkg/frontend/frontend.go
AldoFusterTurpin: True. Thanks! IMHO The fact that we use the same functions for both create and update is making the code a bit more complex than it could be as we need to think about both cases at the same time.

frontend/pkg/frontend/frontend.go
AldoFusterTurpin: Yes! We agreed in a previous [comment](https://github.com/Azure/ARO-HCP/pull/164#discussion_r1628086699) of a PR that it would be beneficial 🙂

frontend/pkg/frontend/frontend.go
AldoFusterTurpin: I understand the point of trying to avoid one DB call, but I am not completely sure about that because that will imply that when we want to query a NodePool, we skip the retrieval of the ClusterDoc from Cosmos, assuming it is already there and it is okay, something I am not 100% sure we should do. I am just trying to think about that in terms of consistency, to check if there is something strange that could happen by not checking the existence of the cluster doc before the nodepool doc. 🤔

frontend/pkg/frontend/frontend.go
```diff
@@ -431,6 +509,227 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
 	}
 }
 
+// clusterResourceIdFromNodePoolResourceID strips the nodepool part of the full resource ID
+// returning the full cluster resourceID.
+// given a full resource ID like
```
mjlshen, line 514:
```suggestion
// given a full nodepool resource ID like
```

mjlshen (dismissed):
lgtm

tonytheleg (approved):

frontend/pkg/frontend/frontend.go
```diff
@@ -284,6 +284,84 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
...
+		}
+
+		f.logger.Error(err.Error())
+		arm.WriteInternalServerError(writer)
+		return
+	}
+
+	targetNodePoolName := request.PathValue(PathSegmentNodepoolName)
+
+	f.logger.Error(fmt.Sprintf("targetNodePoolName is : %v", targetNodePoolName))
```
bennerv, line 332:
```suggestion
	f.logger.Info(fmt.Sprintf("targetNodePoolName is : %v", targetNodePoolName))

File: frontend/pkg/frontend/frontend.go [hunk 6]
```diff
@@ -431,6 +509,227 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
 	}
 }
 
+// clusterResourceIdFromNodePoolResourceID strips the nodepool part of the full resource ID
+// returning the full cluster resourceID.
+// given a full resource ID like
+// /subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/dev-test-rg/providers/microsoft.redhatopenshift/hcpopenshiftclusters/aldo-cluster/nodepools/aldo-nodepool
+// it returns the resourceID of the cluster
+// /subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/dev-test-rg/providers/microsoft.redhatopenshift/hcpopenshiftclusters/aldo-cluster
+func clusterResourceIdFromNodePoolResourceID(resourceID string) string {
```
bennerv, line 518: Use https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm#ParseResourceID to parse resource IDs.

frontend/pkg/frontend/frontend.go
```diff
@@ -431,6 +509,227 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
...
+	i := strings.Index(resourceID, "/nodepools")
+	clusterResourceId := resourceID[0:i]
+	return clusterResourceId
+}
+
+func (f *Frontend) CreateNodePool(writer http.ResponseWriter, request *http.Request) {
+	var err error
```
bennerv, line 525: if we're assigning err near the top, why declare it as a var here?

frontend/pkg/frontend/frontend.go
```diff
@@ -431,6 +509,227 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
...
+			}
+		} else {
+			f.logger.Error(fmt.Sprintf("failed to fetch node pool document for %s: %v", nodePoolResourceID, err))
+			arm.WriteInternalServerError(writer)
+			return
+		}
+	}
+
+	if clusterDoc.ClusterID == "" {
+		f.logger.Error("unexpected error: clusterID of clusterDoc is empty.")
```
bennerv, line 582: can we log out clusterID?

frontend/pkg/frontend/frontend.go
bennerv: We're inconsistent in the functions in this regard.  Delete doesn't have it.

frontend/pkg/frontend/frontend.go
```diff
@@ -431,6 +509,227 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
...
+	// URL path is already lowercased by middleware.
+	nodePoolResourceID := request.URL.Path
+	clusterResourceID := clusterResourceIdFromNodePoolResourceID(nodePoolResourceID)
+
+	subscriptionID := request.PathValue(PathSegmentSubscriptionID)
+	clusterDoc, err := f.dbClient.GetClusterDoc(ctx, clusterResourceID, subscriptionID)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			f.logger.Info(fmt.Sprintf("existing document not found for cluster %s when deleting node pool", clusterResourceID))
+			arm.WriteInternalServerError(writer)
```
bennerv, line 681: I think this should be a 404 error on delete.  Resource doesn't exist sort of thing, not internal server error.

bennerv (commented):
we, as follow-up, need to adhere to error guidelines by common api contract.  

If we return an error to the customer, it should be in odata format.  So wherever we're writing out a non-200 response, we should have an accompanying error adhering to guidelines.

https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/common-api-details.md#error-response-content

Doesn't have to be part of this PR, maybe someone decides to follow-up on it?

bennerv:
> we, as follow-up, need to adhere to error guidelines by common api contract.

Created https://issues.redhat.com/browse/ARO-8958

frontend/pkg/frontend/frontend.go
AldoFusterTurpin: Sure. Thanks.

frontend/pkg/frontend/frontend.go
AldoFusterTurpin: Agree. Left over from a previous block copy.

frontend/pkg/frontend/frontend.go
AldoFusterTurpin: Sure. That was the idea. Good catch 🙂 Thanks!

frontend/pkg/frontend/frontend.go
AldoFusterTurpin: Sure. I forgot to modify the Copy/pasted block.Thanks.

frontend/pkg/frontend/frontend.go
AldoFusterTurpin: Sure. Right. Thanks!

frontend/pkg/frontend/frontend.go
bennerv: ack this is an essentially: "should never happen" sort of thing.  My bad.

frontend/pkg/frontend/frontend.go
AldoFusterTurpin: After the conversation in the Slack Thread, I am leaving the CosmosDB access as it is, as we will need to check the Provisioning state of the cluster and check that it is different than _ProvisioningStateDeleting_ once we have this info available in database.HCPOpenShiftClusterDocument (not ready yet).

mjlshen (approved):

tonytheleg (approved):

PR #283, opened 2024-06-27
AKS cluster name parameter

geoberle:
### What this PR does

the new mandatory paramter `aksClusterName` can be used to define the name of the AKS cluster to be created. works on both SC and MC templates.

our bicepparam files for MVP keep backwards compatibility for now so the names of existing clusters don't change.

Jira: 
Link to demo recording:

### DIFF
--- a/dev-infrastructure/configurations/mvp-mgmt-cluster.bicepparam
+++ b/dev-infrastructure/configurations/mvp-mgmt-cluster.bicepparam
@@ -6,6 +6,7 @@ param vnetAddressPrefix = '10.132.0.0/14'
 param subnetPrefix = '10.132.8.0/21'
 param podSubnetPrefix = '10.132.64.0/18'
 param enablePrivateCluster = false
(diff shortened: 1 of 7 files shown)


dev-infrastructure/configurations/mvp-mgmt-cluster.bicepparam
```diff
@@ -6,6 +6,7 @@ param vnetAddressPrefix = '10.132.0.0/14'
 param subnetPrefix = '10.132.8.0/21'
 param podSubnetPrefix = '10.132.64.0/18'
 param enablePrivateCluster = false
+param aksClusterName = take('aro-hcp-mgmt-cluster-${uniqueString('mgmt-cluster')}', 63)
```
geoberle, line 9: i would like to keep the names backwards compatible for the MVP SC and MC for now. we can beautify the name next time we may need to reinstall them.

mjlshen (approved):

PR #295, opened 2024-07-02
Fix Management cluster parallelism issue

jonathan34c:
… group

### What this PR does
Currently, the management cluster can't deploy parallelly. The main reason was due to the overlapping of the eventgrid certificate name. The current code uses the aks cluster name as the certificate name, we suggest using the Resource group name due to it's unique nature(```hcp-underlay-${region}-mgmt-${stamp number}```), and we would like to keep the aks cluster name the same throughout all resource group. 

Explanation: 
Screenshot 2024-07-01 160714
When the aks cluster name was unique, I was able to deploy parallel, but not when all cluster share the same name. 

Jira: 
Link to demo recording:

### DIFF
--- a/dev-infrastructure/templates/mgmt-cluster.bicep
+++ b/dev-infrastructure/templates/mgmt-cluster.bicep
@@ -60,6 +60,8 @@ param maestroInfraResourceGroup string
 @description('The name of the eventgrid namespace for Maestro.')
 param maestroEventGridNamespacesName string
 
+func isValidMaestroConsumerName(input string) bool => length(input) <= 90 && contains(input, '[^a-zA-Z0-9_-]') == false
+
 module mgmtCluster '../modules/aks-cluster-base.bicep' = {
   name: 'aks_base_cluster'
   scope: resourceGroup()
@@ -101,7 +103,7 @@ module maestroConsumer '../modules/maestro/maestro-consumer.bicep' = if (deployM
     )[0].uamiClientID
     namespace: maestroNamespace
     maestroInfraResourceGroup: maestroInfraResourceGroup
-    maestroConsumerName: resourceGroup().name
+    maestroConsumerName: isValidMaestroConsumerName(resourceGroup().name)?resourceGroup().name:''
     maestroEventGridNamespaceName: maestroEventGridNamespacesName
     maestroKeyVaultName: maestroKeyVaultName
     maestroKeyVaultOfficerManagedIdentityName: maestroKeyVaultCertOfficerMSIName
--- a/dev-infrastructure/templates/mgmt-cluster.bicep
+++ b/dev-infrastructure/templates/mgmt-cluster.bicep
@@ -8,7 +8,7 @@ param persist bool = false
 param currentUserId string
 
 @description('AKS cluster name')
-param aksClusterName string = 'aro-hcp-mgmt-aks1'
+param aksClusterName string = 'aro-hcp-mgmt-1'
 
 @description('Name of the resource group for the AKS nodes')
 param aksNodeResourceGroupName string = '${resourceGroup().name}-aks1'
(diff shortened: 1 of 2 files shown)


dev-infrastructure/templates/mgmt-cluster.bicep
```diff
@@ -8,7 +8,7 @@ param persist bool = false
 param currentUserId string
 
 @description('AKS cluster name')
-param aksClusterName string
+param aksClusterName string = 'aro-hcp-mgmt-aks1'
```
weinong, line 11: why aks1?

dev-infrastructure/templates/mgmt-cluster.bicep
```diff
@@ -101,7 +101,7 @@ module maestroConsumer '../modules/maestro/maestro-consumer.bicep' = if (deployM
     )[0].uamiClientID
     namespace: maestroNamespace
     maestroInfraResourceGroup: maestroInfraResourceGroup
-    maestroConsumerName: mgmtCluster.outputs.aksClusterName
+    maestroConsumerName: resourceGroup().name
```
geoberle, line 104: i'm ok with this. since this names is passed to several others systems (eventgrid, maestro), we would need to make sure that the resourcegroup name consists only of certain characeters. the following seems to be safe

* max 90 chars long
* contains only characters, numbers, underscores and hyphens

is this going to compatible with what you have in mind RG naming wise?
in any case, we should have some validation in place

dev-infrastructure/templates/mgmt-cluster.bicep
weinong: since the resource group already denotes the nature of the cluster, mgmt  or svc, just call it `aro-hcp-aks`?

jonathan34c: sorry it should be aro-hcp-mgmt-1. Currently we only have 1 aks cluster, but we would like to have the number at the end incase of future expansion (when we have more than 1 aks cluster)

dev-infrastructure/templates/mgmt-cluster.bicep
jonathan34c: yes, the naming convention for the RG is ```hcp-underlay-${region}-mgmt-${stamp number}```. so it should be within the boundary.

geoberle: do you want to add some validation for this in this PR? if not, we can also add it in an upcoming PR

dev-infrastructure/templates/mgmt-cluster.bicep
whober0521: I love `aro-hcp-aks`

geoberle: i was wondering about the `1` suffix as well. i was under the impression we would expand MCs via additional resourcegroup, each containing a new MC, no?

mjlshen: > I like aro-hcp-aks and we can use aro-classic-aks for new classic aks cluster

Not an issue for this thread, but to clarify, we won't have aro-classic-aks in the same resource group either

dev-infrastructure/templates/mgmt-cluster.bicep
jonathan34c: just added the validation

dev-infrastructure/templates/mgmt-cluster.bicep
jonathan34c: thanks for the inputs, I have updated the naming in the shared [google doc](https://microsoft-my.sharepoint.com/:w:/p/gniranjan/ERWrqFehqZZDl3jRjaS0vrIB1MxGERKjhe-jByWIFT_PcA?e=Mb727L)

mjlshen (approved):

PR #300, opened 2024-07-03
grant the cluster-service managed identity access to the svc kv

geoberle:
### What this PR does

Jira: https://issues.redhat.com/browse/ARO-7387
Link to demo recording:

### DIFF
--- a/dev-infrastructure/modules/keyvault/keyvault-secret-access.bicep
+++ b/dev-infrastructure/modules/keyvault/keyvault-secret-access.bicep
@@ -1,0 +1,57 @@
+@description('The name of the keyvault where access is managed')
+param keyVaultName string
+
+@description('Optional name of the secret in the keyvault where access is managed. If not provided, access will be granted to all secrets in the keyvault.')
+param secretName string = ''
+
+@description('The name of the role that will be assigned to the managed identity for the secret in KV')
+@allowed([
+  'Key Vault Secrets Officer'
+  'Key Vault Secrets User'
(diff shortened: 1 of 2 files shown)


bennerv (approved):

dev-infrastructure/modules/keyvault/keyvault-secret-access.bicep
```diff
@@ -0,0 +1,57 @@
+@description('The name of the keyvault where access is managed')
+param keyVaultName string
+
+@description('Optional name of the secret in the keyvault where access is managed. If not provided, access will be granted to all secrets in the keyvault.')
+param secretName string = ''
+
+@description('The name of the role that will be assigned to the managed identity for the secret in KV')
+@allowed([
+  'Key Vault Secrets Officer'
+  'Key Vault Secrets User'
```
JameelB, line 10: Don't we also need Key Vaults Certificate User? What's the reason for granting the identity the officer role? 🤔

bennerv: We'll need that eventually.  

But for now, when you create a certificate, you download the secret (which includes the private half of the certificate).  So you technically don't need certificate user I believe.

JameelB: thanks for the explanation @bennerv, looks good to me!

mjlshen (approved):

PR #301, opened 2024-07-03
Use CBL-mariner base images for frontend

mjlshen:
### What this PR does
* Allows passing TARGETPLATFORM to build on other platforms (e.g. ARM) for local development
* Always attempt to build a docker image from the make target, defer to docker if builds are cached

Jira: [ARO-8662](https://issues.redhat.com/browse/ARO-8662)


frontend/Makefile
mjlshen, line 1: No need to skip builds at a Makefile level if an image already exists - if so docker/podman perform cached builds

tonytheleg (approved):

bennerv (approved):

PR #314, opened 2024-07-05
aro-8600 deploy olm on MC

ritmun:
Currently, ACM requires olm to be deployed. Installing it as step 1 for ACM install.

### What this PR does

Jira:  https://issues.redhat.com/browse/ARO-8600 
Link to demo recording:


acm/scripts/prepare-olm.sh
mjlshen, line 1: Should we just use the `install.sh` bundled with every release instead? https://github.com/operator-framework/operator-lifecycle-manager/releases

acm/scripts/prepare-olm.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+#
+
+set -e
+
+release="v0.27.0"
```
mjlshen, line 6:
```suggestion
release="v0.28.0"
```
Just curious if we chose this version intentionally when compared to latest

mjlshen (commented):
All my questions are not blockers, I am not familiar with deploying OLM either and asking mostly out of curiosity

mjlshen (approved):

geoberle (approved):

PR #319, opened 2024-07-08
provide a dns subdomain default for personal dev envs

geoberle:
### What this PR does

reintroduce a default for regionalDNSSubdomain in svc-cluster.bicepparam for conflict-free personal dev environment setups.

Jira: 
Link to demo recording:


mjlshen (approved):
Yeah, you're right about devs interfering with MVP's zone delegation - I did not consider that since the `${region}.hcp.osadev.cloud` zone is in the service cluster resource group, not the global one.

tonytheleg (approved):

PR #330, opened 2024-07-09
Break apart armResourceCreateorUpdate

katherinelc321:
### What this PR does
Created a func that handles PATCH request sorely since updates can come as both PUT/PATCH 

Jira: https://issues.redhat.com/browse/ARO-8660
Link to demo recording:

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -347,7 +347,7 @@ func (f *Frontend) ArmResourceCreate(writer http.ResponseWriter, request *http.R
 	doc.ClusterID = req.Body().ID()
 	err = f.dbClient.SetClusterDoc(ctx, doc)
 	if err != nil {
-		f.logger.Error(fmt.Sprintf("failed to create document for resource %s: %v", doc.Key, err))
+		f.logger.Error(fmt.Sprintf("failed to create document for resource %s: %v", doc.ClusterID, err))
 	}
 
 	resp, err := json.Marshal(versionedRequestCluster)
@@ -435,7 +435,7 @@ func (f *Frontend) ArmResourceUpdate(writer http.ResponseWriter, request *http.R
 	doc.ClusterID = req.Body().ID()
 	err = f.dbClient.SetClusterDoc(ctx, doc)
 	if err != nil {
-		f.logger.Error(fmt.Sprintf("failed to update document for resource %s: %v", doc.Key, err))
+		f.logger.Error(fmt.Sprintf("failed to update document for resource %s: %v", doc.ClusterID, err))
 	}
 
 	resp, err := json.Marshal(versionedRequestCluster)
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -302,42 +302,19 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
 }
 
 func (f *Frontend) ArmResourceCreate(writer http.ResponseWriter, request *http.Request) {
-	var err error
-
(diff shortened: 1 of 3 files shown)


frontend/pkg/frontend/frontend.go
```diff
@@ -284,92 +284,102 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
...
+	if err != nil {
+		return
+	}
 
-	versionedInterface, err := VersionFromContext(ctx)
+	if updating {
+		f.logger.Info(fmt.Sprintf("document already exists for %s, use PATCH instead", doc.Key))
+		arm.WriteInternalServerError(writer)
+		return
+	}
```
tonytheleg, line 298: with regards to cluster docs, im not sure there is a scenario where we would actually update the doc to where it mattered if we were updating or not that we need to check it here?

The ArmResourceCreate call would only ever get routed to on PUT, in which if the cluster already exists i think the CS logic later on would handle errors trying to PUT a cluster that already exists and we can let that handle the error?

Thoughts @mbarnes @mjlshen ?

If we do want to explicitly check, we probably need to return the proper status code instead of a server error

frontend/pkg/frontend/frontend.go
```diff
@@ -284,92 +284,102 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
...
-			return
-		}
+		return
+	}
+
+	if !updating {
+		originalPath, _ := OriginalPathFromContext(ctx)
+		f.logger.Error("Resource not found")
+		arm.WriteError(writer, http.StatusNotFound, arm.CloudErrorCodeNotFound, originalPath, "Resource not found")
+		return
```
tonytheleg, line 360: Similar to above, but in this scenario if the doc is not found its likely the cluster doesnt exist and patching just fails so the error check just above this may handle that for us

frontend/pkg/frontend/frontend.go
mbarnes: Updating with PUT is a valid ARM operation, even if it's not a valid Cluster Service operation.   We shouldn't be erroring out here.  If the resource exists and the PUT request passes static validation, pass it to the update handler.

mbarnes (changes requested):

frontend/pkg/frontend/frontend.go
```diff
@@ -301,7 +301,7 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
 	writer.WriteHeader(http.StatusOK)
 }
 
-func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request *http.Request) {
+func (f *Frontend) ArmResourceCreateorUpdate(writer http.ResponseWriter, request *http.Request) {
```
mjlshen, line 304:
```suggestion
func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request *http.Request) {
```
Any reason we renamed this?

mjlshen (commented):
Looks good to me, minor nit to un-rename `ArmResourceCreateorUpdate`

frontend/pkg/frontend/frontend.go
katherinelc321: ah the capital O is a typo
but updates requests can come as both PUT or Patch requests so i'm splitting the func into one that handles only PATCH and one that handles both PUT/PATCH

mjlshen: Yup, agreed, just keeping the old one the same :D

mjlshen (approved):

tonytheleg (approved):

PR #333, opened 2024-07-10
Init repository sync for HCP

janboll:
### What this PR does
This adds tooling designed to automate the process of synchronizing container images from specified external repositories to a designated Azure Container Registry, focusing on the latest few versions of each image.

https://issues.redhat.com/browse/ARO-8595


tooling/image-sync/sync.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return &secret, nil
+}
+
+func filterTagsToSync(src, target []string) []string {
+	var tagsToSync []string
+
+	for _, srcTag := range src {
+		found := false
+		for _, targetTag := range target {
```
geoberle, line 102: can we get rid of the nested loop
e.g. wrap target into a `map[string]whatever` and use it for existance lookups?

tooling/image-sync/repository.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	var tags []string
+
+	for _, tag := range tagsResponse.Tags {
+		if tag.Name == "latest" {
+			continue
+		}
+		tags = append(tags, tag.Name)
+		if len(tags) >= q.numberOftags {
```
geoberle, line 107: the quay API endpoint to list tags supports a paramter `limit`
https://docs.quay.io/api/swagger/#!/tag/listRepoTags

using this would make `ReadAll` and `Unmarshal` run faster as well.

geoberle: also if we are interested in more than 100 images, we need paging as the max value for `limit` is 100

tooling/image-sync/repository.go
```diff
@@ -0,0 +1,363 @@
...
+			break
+		}
+		returnTags = append(returnTags, uploadedTagAt[k]...)
+	}
+
+	return returnTags, nil
+}
+
+// GetTags returns the tags in the given repository
+func (o *OCIRegistry) GetTags(ctx context.Context, image string) ([]string, error) {
```
geoberle, line 331: any reason to now use `docker.GetRepositoryTags()` from github.com/containers/image as a basis for the implementation?

geoberle (commented):
looks good. some nits

tooling/image-sync/repository.go
janboll: I considered it, but it returns a flat list of tags ([]string), ref: https://pkg.go.dev/github.com/containers/image/v5/docker#GetRepositoryTags

The basis for this sync however is to only sync N tags, ordered by update/creation. Thus, we need our own implementation here.

tooling/image-sync/sync.go
janboll: Great idea, will do it.

tooling/image-sync/repository.go
janboll: I wanted to overhaul this later, but I can also do it now to avoid broken/unfinished code.

tooling/image-sync/go.mod
```diff
@@ -0,0 +1,88 @@
...
+go 1.22.2
+
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.12.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.7.0
+	github.com/Azure/azure-sdk-for-go/sdk/containers/azcontainerregistry v0.2.1
+	github.com/containers/image v3.0.2+incompatible
+	github.com/spf13/cobra v1.8.1
+	github.com/spf13/viper v1.19.0
+	go.uber.org/zap v1.27.0
```
cgiradkar, line 12: according to docs [here](https://github.com/uber-go/zap), it is stated that

> Note that zap only supports the two most recent minor versions of Go.

this is a vague statement, but do we anticipate anything breaking due to this?

janboll: I Don't think it's an issue, cause we can pin the version.

WDYT?

tooling/image-sync/repository.go
janboll: I Added pagination, but still keeping the limit of 100 here. The reason is, that _latest_ can occur a couple of times in the response. The API will return one entry for _latest_, every time it is pushed. Probably page size could be `numberOftags * 2`. But adding this complexity is not worth vs. having todo extra requests. Which will undoubtedly be slower than reading a couple of more bytes.

tooling/image-sync/go.mod
cgiradkar: > I Don't think it's an issue, cause we can pin the version.
> 
> WDYT?

fine with me then

cgiradkar:
lgtm

geoberle (dismissed):

tooling/image-sync/internal/sync.go
```diff
@@ -0,0 +1,196 @@
...
+			}
+			Log().Debugw("Got tags from quay", "tags", srcTags)
+		} else {
+			oci := NewOCIRegistry(cfg, baseURL)
+			srcTags, err = oci.GetTags(ctx, repoName)
+			if err != nil {
+				return fmt.Errorf("error getting oci tags: %w", err)
+			}
+			Log().Debugw(fmt.Sprintf("Got tags from %s", baseURL), "repo", repoName, "tags", srcTags)
+		}
```
mjlshen, line 162: This is a little brittle, we are assuming here that a quay pull secret is needed for `quay.io` images and no secret is needed for any other registry right?

tooling/image-sync/Makefile
```diff
@@ -0,0 +1,23 @@
+SHELL = /bin/bash
+COMMIT = $(shell git rev-parse --short=7 HEAD)$(shell [[ $$(git status --porcelain) = "" ]] || echo -dirty)
+ARO_HCP_BASE_IMAGE ?= devarohcp.azurecr.io
+ARO_HCP_IMAGE_SYNC_IMAGE ?= $(ARO_HCP_BASE_IMAGE)/image-sync/image-sync:$(COMMIT)
+
+image-sync:
+	go build -o image-sync .
```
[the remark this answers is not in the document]
mjlshen: Optional nit, since the package name is `image-sync` I don't think we need to specify `-o image-sync`

tooling/image-sync/internal/repository.go
```diff
@@ -0,0 +1,379 @@
...
+			Transport: &AuthedTransport{
+				Key:     "Bearer " + bearerToken,
+				Wrapped: http.DefaultTransport,
+			},
+		},
+		baseUrl:      "https://quay.io",
+		numberOftags: cfg.NumberOfTags,
+	}
+	return q
+}
```
mjlshen, line 57: What do you think about mounting our pull secrets into standard locations https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md#description to avoid having to handle referencing/storing the secrets in our code?

mjlshen (changes requested):

tooling/image-sync/Makefile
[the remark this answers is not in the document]
janboll: Great suggestion, let me add this.

tooling/image-sync/go.mod
```diff
@@ -0,0 +1,90 @@
...
+
+go 1.22.0
+
+toolchain go1.22.2
+
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.12.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.7.0
+	github.com/Azure/azure-sdk-for-go/sdk/containers/azcontainerregistry v0.2.1
+	github.com/containers/image v3.0.2+incompatible
```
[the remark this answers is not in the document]
janboll: That was a mistake, using the latest now

tooling/image-sync/internal/repository.go
janboll: That's great. I'll remove the source auth and replace it with this. It will also implicitly enable auth for other registries.

tooling/image-sync/internal/sync.go
janboll: See below comment, it will work using passed in containers auth file

tooling/image-sync/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM --platform=${TARGETPLATFORM:-linux/amd64} mcr.microsoft.com/oss/go/microsoft/golang:1.22-fips-cbl-mariner2.0@sha256:d024591e53909d3f5d9f27cde4e1fea45f32963d97711042db36a499779630f8 as builder
+
+WORKDIR /app
+ADD . .
+# https://github.com/microsoft/go/tree/microsoft/main/eng/doc/fips#build-option-to-require-fips-mode
+ENV CGO_ENABLED=1 GOFLAGS='-tags=requirefips'
+
+RUN tdnf install -y gpgme-devel; \
+    tdnf clean all; \
+    make image-sync
```
mjlshen, line 11: We don't need to setup this dependency with the `containers_image_openpgp` build tag
```suggestion
RUN make image-sync
```

tooling/image-sync/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+FROM --platform=${TARGETPLATFORM:-linux/amd64} mcr.microsoft.com/cbl-mariner/base/core:2.0-nonroot@sha256:fac93debd6c88c9935fdb85e08c3a57e601124018b87015635f50549b6527f5a
```
mjlshen, line 13:
```suggestion
FROM --platform=${TARGETPLATFORM:-linux/amd64} mcr.microsoft.com/cbl-mariner/distroless/base:2.0-nonroot@sha256:ef0dc582fc2a8dd34fbb41341a3a9a1aaa70d4542ff04ce4e33a641e52e4807e
```
We can use the distroless container I think with the `containers_image_openpgp` build tag

.github/workflows/ci-go.yml
```diff
@@ -19,6 +19,9 @@ jobs:
         with:
           fetch-depth: 1
 
+      - name: 'Install gpgme'
+        run: sudo apt-get update && sudo apt-get install -y libgpgme-dev
+
```
mjlshen, line 24:
```suggestion
```
We don't need to setup this dependency with the `containers_image_openpgp` build tag

.github/workflows/ci-go.yml
```diff
@@ -36,6 +39,9 @@ jobs:
...
```
mjlshen, line 44:
```suggestion
```
We don't need to setup this dependency with the `containers_image_openpgp` build tag

mjlshen (changes requested):

tooling/image-sync/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Base and builder image will need to be replaced by Fips compliant one
```
mjlshen, line 1:
```suggestion
```

mjlshen (approved):

geoberle (approved):

PR #339, opened 2024-07-11
rename CS MI and SA to clusters-service

geoberle:
### What this PR does

this is for consistency reasons with non-ARO-HCP CS deployments

Jira: https://issues.redhat.com/browse/ARO-8949
Link to demo recording:

### DIFF
--- a/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
+++ b/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
@@ -214,6 +214,9 @@ parameters:
 - name: CPU_LIMIT
   description: CPU limit.
   value: "1"
+- name: AZURE_CS_MI_CLIENT_ID
+  description: Client ID for the clusters-service manage identity.
+  required: true
 
 objects:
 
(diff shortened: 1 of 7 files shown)


cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -214,6 +214,9 @@ parameters:
 - name: CPU_LIMIT
   description: CPU limit.
   value: "1"
+- name: CS_MI_CLIENT_ID
```
JameelB, line 217: Should we prefix this with `AZURE`? For cloud specific parameters, I think it's prefixed with the associated cloud, e.g. `AWS_STS_POLICY_DIRECTORY`.  

Can also be done in a follow up PR if we do decide if we want to add this.

JameelB (approved):
lgtm, thanks @geoberle. 

Have one minor comment around naming of the managed identity client id parameter but that can be followed up in a separate pr.

tonytheleg (approved):

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
geoberle: sure. fixed

bennerv (approved):

jmelis (approved):
LGTM

PR #351, opened 2024-07-15
postgres managed-identity auth fix

geoberle:
### What this PR does

* drop the region suffix on managed identity names (simplifies the entra DB auth setup)
* rename the CS database to clusters-service (adding an s)
* don't place the service account manifests on cluters anymore (all service component deployments template their own SAs with the proper MI client id)
* add make targets to connect to the postgres databases for CS and maestro via managed identity (to test and verify the same authn/z setup as on the AKS cluster)
* grant schema permissions to managed identity on postgres

Jira: 
Link to demo recording:


dev-infrastructure/modules/cluster-service.bicep
```diff
@@ -8,7 +8,7 @@ param clusterServiceManagedIdentityName string
 param clusterServiceManagedIdentityPrincipalId string
 
 @description('The name of the database to create for CS')
-param csDatabaseName string = 'cluster-service'
+param csDatabaseName string = 'clusters-service'
```
mjlshen, line 11: Unusure about this rename, but trusting that it works for CS

tonytheleg, line 11: Doesnt block merge, but if the CS app expects a different name like `cluster-service` this might cause you some head scratching later :)

tonytheleg (approved):

dev-infrastructure/modules/cluster-service.bicep
geoberle: the name is configurable in CS and Mark preferred it like this because it is closer to other setups

PR #352, opened 2024-07-15
frontend: Add comment to explain MiddlewareLowercase

mbarnes:
Just adding a comment to try to clarify the purpose of `MiddlewareLowercase`, since I saw some recent confusion around it.


mbarnes:
I don't know why CI thinks this is coming from a fork when it clearly isn't.  Being as this is just a code comment, can we override CI this time?

PR #355, opened 2024-07-15
Explicitly write out ARO HCP paths

mjlshen:
### What this PR does
Removes `MuxPattern()` in favor of explicitly writing out our paths

Pros:
* Readability

Cons:
* Need to manually ensure future added paths are all lowercase
* Easier to have typos

Both of the cons can be mitigated by integration tests, which we do not have yet at this point in the project.

### DIFF
--- a/frontend/pkg/frontend/const.go
+++ b/frontend/pkg/frontend/const.go
@@ -8,13 +8,4 @@ const (
 
 	// APIVersionKey is the request parameter name for the API version.
 	APIVersionKey = "api-version"
-
-	// Wildcard path segment names for request multiplexing, must be lowercase as we lowercase the request URL pattern when registering handlers
-	PageSegmentLocation          = "location"
-	PathSegmentSubscriptionID    = "subscriptionid"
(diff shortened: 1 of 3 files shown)


mjlshen:
fyi @AldoFusterTurpin if this is what you had in mind

mbarnes (changes requested):
Not really thrilled about this change, but I'll go with the team consensus.

Can we at least use the `PathSegment` constants for the wildcard parts of the routes, so those stay consistent for use with `request.PathValue`?

i.e. either `fmt.Sprintf("/subscriptions/{%s}/...", PathSegmentSubscriptionID)` or `"/subscriptions/{" + PathSegmentSubscriptionID + "}/..."` so that `request.PathValue(PathSegmentSubscriptionID)` will always work

Whether that further improves readability I'll defer to others on.

AldoFusterTurpin (dismissed):
LGTM! IMHO Explicit is better than implicit, and with this change we can directly see the full path.
Also +1 to what Matthew suggests about the constants.

PR #387, opened 2024-07-23
Prepare the api for ARM

petrkotas:
### What this PR does
Make the API design compliant with upstream repository.
Upstream PR to check https://github.com/Azure/azure-rest-api-specs-pr/pull/18316 

Jira: [](https://issues.redhat.com/browse/ARO-6591)
Link to demo recording: 

Implements the changes to make the API design compliant with azure-rest-api-specs.
Changes here are all done to make the linters and checks in the upstream repo happy.


mbarnes (dismissed):
Autorest appears to be deleting files it shouldn't again.

These files need to be restored and any compilation errors fixed:

- `internal/api/v20240610preview/hcpopenshiftclusters_methods.go`
- `internal/api/v20240610preview/nodepools_methods.go`

bennerv (commented):
general comment.  There were some not obvious but significant differences between getting the API "upstreamed" vs. what we had in the repo for a while.  What can we do to reconcile that in the future?  Are there linters we should pull into the ARO-HCP repo in order to prevent more churn in the future?

petrkotas:
> general comment. There were some not obvious but significant differences between getting the API "upstreamed" vs. what we had in the repo for a while. What can we do to reconcile that in the future? Are there linters we should pull into the ARO-HCP repo in order to prevent more churn in the future?

As with everything in azure repo, there is custom code doing the validation. Extracting it and moving here does not make sense, as we would easily miss an update. AFAIK the recommended way is doing regular PRs to the https://github.com/Azure/azure-rest-api-specs-pr and check the output for complains.

api/package-lock.json
mjlshen, line 1: Sorry, I'm npm illiterate so I don't have a good suggestion, but is there some way to get the `^0.44.0` changes propagated into here as well (maybe manual is good enough)?

bennerv: needs an `npm i` to update `package-lock.json`

mjlshen (approved):

mjlshen:
I rebased the PR

mjlshen (approved):

mbarnes (approved):

PR #402, opened 2024-07-25
Create regional resource group

bennerv:
* DNS is deployed into the regional resource group
* This removes a reliance on the service resource group in order to create mgmt cluster
* Establishes a pattern for 'shared' resources within a region


.github/workflows/aro-hcp-dev-env-cd.yml
```diff
@@ -56,18 +56,30 @@
...
+                --template-file templates/dev-dns.bicep \
+                --parameters \
+                  configurations/dev-dns.bicepparam \
+                --parameters \
+                  currentUserId=$(CURRENTUSER)
 
               # dev Azure Container Registry (ACR)
               az deployment group create \
                 --name "dev-acr-${GITHUB_RUN_ID}" \
-                --resource-group "${SC_RESOURCEGROUP}" \
```
bennerv, line 70: Moves ACR into regional resource group. 

Next steps would be to move this to a "global" resource group and leverage it across all regional deployments.

dev-infrastructure/Makefile
```diff
@@ -3,11 +3,11 @@ SHELL = /bin/bash
 CURRENTUSER=$(shell az ad signed-in-user show | jq -r '.id')
 LOCATION?=westus3
 RESOURCEGROUP?=aro-hcp-${AKSCONFIG}-$(USER)
+REGIONAL_RESOURCEGROUP?=aro-hcp-$(LOCATION)-$(USER)
 MAESTRO_INFRA_RESOURCEGROUP?=aro-hcp-svc-cluster-$(USER)
```
bennerv, line 7: we should rename all the resource groups to have the region names or short region names in them so if we wanted to create duplicate dev infra in multiple regions within a single sub it would work.  (plus they're easier to search with region in the name)

dev-infrastructure/Makefile
```diff
@@ -31,29 +31,49 @@ lint:
...
+feature-registration: # hardcoded to eastus as this is a subscription deployment, not a resource group
 	@az deployment sub create \
-		--name "feature-registration"
+		--name "feature-registration" \
+		--location eastus \
 		--template-file templates/feature-registration.bicep \
 		--confirm-with-what-if
 .PHONY: feature-registration
 
-rg: feature-registration
```
bennerv, line 41: Puts the AFEC feature flags into "Registering" state which makes the AKS deployment fail.  

Just do this once per sub.

bennerv:
@geoberle @mjlshen what-if is going to fail because I moved the ACR into the regional resource group and there would be a naming conflict.  

I also moved the DNS zone which will cause problems as well.  

Do we repave infra to make what-if pass, or delete the DNS zone + ACR and then we will probably have to redeploy external-dns operator as well in order to get it to pick up the new zone configuration?

Going to move it out of draft, but will be out until next Tuesday to resolve.  Feel free to edit or push to the PR yourselves

.github/workflows/aro-hcp-dev-env-cd.yml
```diff
@@ -1,5 +1,11 @@
 ---
   name: ARO HCP Dev Environment Continuous Deployment
+  env:
+    REGION: westus3
+    REGIONAL_RESOURCEGROUP: aro-hcp-dev-westus3
+    SC_RESOURCEGROUP: aro-hcp-dev-westus3-sc
+    MC_RESOURCEGROUP: aro-hcp-dev-westus3-mc-1
```
weinong, line 7: i suppose `sc` is for service cluster and `mc` is for management cluster?
We should have a naming convention already. @nwnt can provide the doc.
Nonetheless, even we will only have one svc cluster, it's possible we will have multiple of them for various migration reason. so keeping the numbering suffix will make it consistent and future proof.

mjlshen (approved):

dev-infrastructure/templates/region.bicep
mjlshen, line 1: Discussed folding this into `region.bicep`, not a blocker for this PR but more of a follow-up TODO

.github/workflows/aro-hcp-dev-env-cd.yml
geoberle: these are just the names for our DEV env. we can adopt the naming pattern of microsoft though. @bennerv @mjlshen

PR #413, opened 2024-07-29
Parameterize disableLocalAccounts field

jonathan34c:
### What this PR does

I need this field to be an option for me to run istio upgrade to restart deployment 
Jira: 
Link to demo recording:


mjlshen (changes requested):
Is this allowed? Will take this discussion out of GitHub, but my understanding was that we must disable local accounts

PR #425, opened 2024-07-31
frontend: Prefer the term "location" instead of "region"

mbarnes:
### What this PR does

Even Microsoft is inconsistent about this in their documentation, but for consistency with the "az" tool and resource provider URIs, use the term "location" at least within the RP source code.

Jira: _Very_ loosely related to [ARO-5737 : Implement Async API ARM Requirements for Long-Running Operations](https://issues.redhat.com/browse/ARO-5737)
Link to demo recording: 

This has been bugging me for awhile, but decided to fix it when I had to build a URI that had `/locations/{$region}`.

This change only targets the RP.  I'm leaving other components like Cluster Service alone, since they're still mostly based on AWS and AWS uses "region".


frontend/Makefile
```diff
@@ -16,7 +16,7 @@ frontend:
 	go build -o aro-hcp-frontend .
 
 run:
-	./aro-hcp-frontend --use-cache --region ${REGION} \
```
geoberle, line 19: the REGION variabel for personal DEV environments comes from dev-infrastructure/configuration/dev.mk, which is shared by basically all services (CS, maestro, hypershift-operator, ...)

if you want to change this then you need to change it throughout the entire repo

mbarnes: I don't want the change to be that widespread, at least not in this PR.  I'll add this to the RP's Makefile:

```
# dev-infrastructure defines this as REGION
LOCATION ?= ${REGION}
```

.github/workflows/aro-hcp-dev-env-cd.yml
```diff
@@ -1,7 +1,7 @@
 ---
   name: ARO HCP Dev Environment Continuous Deployment
   env:
-    REGION: westus3
```
geoberle, line 4: most of the makefiles still require REGION over LOCATION
so please leave the REGION var here and add LOCATION on top. we can clean up once we adjusted all Makefiles

mbarnes: Ok, sorry, I didn't realize the impact on other components.
I'm trying to contain the scope of this to the RP, so I'll back this change out and leave it as REGION.
The workaround I added earlier for the RP (`LOCATION ?= ${REGION}`) should suffice for now.

bennerv (approved):
lgtm; can swap all the dev infrastructure in another PR.

geoberle (approved):

PR #431, opened 2024-07-31
The ACRs exist in the global resource group now

mjlshen:
### What this PR does
No ACR's currently exist in the regional resource group. Open to not merging if we intend to have ACR's in the regional resource group


bennerv (approved):
lgtm; what-if is failing because of an active deployment.  Did we limit concurrency?

mjlshen:
> what-if is failing because of an active deployment. Did we limit concurrency?
It happened because we're not blocking a what-if if a deployment is in progress (should fix that if possible). The actual deployments are concurrency limited.

geoberle (approved):

PR #434, opened 2024-08-01
database: Pass resource IDs as structs instead of strings

mbarnes:
### What this PR does

Simplify the `DBClient` interface by passing the resource ID as a struct instead of a string.

This has the advantage of easily extracting the subscription ID from the resource ID for the partition key, so it doesn't have to be passed separately.

Jira: Loosely related to [ARO-5737 : Implement Async API ARM Requirements for Long-Running Operations](https://issues.redhat.com/browse/ARO-5737)
Link to demo recording:


AldoFusterTurpin (approved):
LGTM. This change is nice. Thanks!

mociarain (approved):
LGTM

PR #437, opened 2024-08-02
Align RP returns to the ARM API contract

mjlshen:
### What this PR does

* We were returning 204's when GET-ing a resource that did not exist instead of a 404
* We were returning 202's when DELETE-ing a resource, which is an asynchronous deletion. This is future state, but not yet implemented. For now, we should return 200's for a synchronous DELETE
* Send ARM-formatted errors when resources are not found

```
#  curl -i -X GET "localhost:8443/subscriptions/00000-00000-00000/resourceGroups/mshen-p1-temp/providers/Microsoft.RedHatOpenshift/hCPOpenShiftClusters/mshEn2?api-version=2024-06-10-preview" 
HTTP/1.1 404 Not Found
X-Ms-Error-Code: ResourceGroupNotFound
X-Ms-Request-Id: 28a5090f-c5bc-411e-8a03-6a633f110061
Date: Fri, 02 Aug 2024 19:41:33 GMT
Content-Length: 182

{
    "error": {
        "code": "ResourceGroupNotFound",
        "message": "The resource 'hCPOpenShiftClusters/mshEn2' under resource group 'mshen-p1-temp' was not found."
    }
}
```

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -207,6 +207,9 @@ func (f *Frontend) ArmResourceList(writer http.ResponseWriter, request *http.Req
 	writer.WriteHeader(http.StatusOK)
 }
 
+// ArmResourceRead implements the GET single resource API contract for ARM
+// * 200 If the resource exists
+// * 404 If the resource does not exist
 func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Request) {
 	ctx := request.Context()
 
@@ -230,7 +233,9 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
 	if err != nil {
 		if errors.Is(err, database.ErrNotFound) {
 			f.logger.Error(fmt.Sprintf("existing document not found for cluster: %s", resourceID))
-			writer.WriteHeader(http.StatusNoContent)
+
+			originalPath, _ := OriginalPathFromContext(ctx)
+			arm.WriteResourceNotFoundError(writer, originalPath)
 			return
 		} else {
 			f.logger.Error(err.Error())
@@ -242,7 +247,9 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
 	csCluster, err := f.clusterServiceConfig.GetCSCluster(doc.ClusterID)
 	if err != nil {
 		f.logger.Error(fmt.Sprintf("cluster not found in clusters-service: %v", err))
-		writer.WriteHeader(http.StatusNoContent)
+
+		originalPath, _ := OriginalPathFromContext(ctx)
+		arm.WriteResourceNotFoundError(writer, originalPath)
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


frontend/pkg/frontend/frontend.go
```diff
@@ -230,7 +233,7 @@ func (f *Frontend) ArmResourceRead(writer http.ResponseWriter, request *http.Req
 	if err != nil {
 		if errors.Is(err, database.ErrNotFound) {
 			f.logger.Error(fmt.Sprintf("existing document not found for cluster: %s", resourceID))
-			writer.WriteHeader(http.StatusNoContent)
+			writer.WriteHeader(http.StatusNotFound)
```
mjlshen, line 236: These need to return ARM-formatted errors/CloudErrors

frontend/pkg/frontend/frontend.go
```diff
@@ -572,7 +579,8 @@ func (f *Frontend) ArmResourceDelete(writer http.ResponseWriter, request *http.R
 	}
 	f.logger.Info(fmt.Sprintf("document deleted for resource %s", resourceID))
 
-	writer.WriteHeader(http.StatusAccepted)
+	// TODO: Eventually this will be an asynchronous delete and need to return a 202
+	writer.WriteHeader(http.StatusOK)
```
mbarnes, line 583: Not sure this is really necessary.  We're never going to deploy to production with synchronous deletion.  We can leave it in, but I'm working on the async part of this now so it will be short-lived.

frontend/pkg/frontend/frontend.go
mbarnes: Were you going to make it a cloud error in this PR or are you just focused on status codes here?  I can pick it up if not.

mbarnes (commented):
Tentative approval with a couple questions.  Withholding approval for the moment only because it would trigger a merge.

frontend/pkg/frontend/frontend.go
mjlshen: Oh sorry - I'll update it for cloud errors for this PR

frontend/pkg/frontend/frontend.go
mjlshen: 100% short-lived, this is just for QE testing in P1. When they get to QE-ing async deletion, the different status code will be one part of their validation

bennerv (approved):

mbarnes:
I didn't get a chance to comment before this got merged, but one simplification I'd like to see made to `WriteResourceNotFoundError` is to have it take a resource ID struct instead of a path string.  I've been trying to nudge the frontend code into using resource ID structs more since I think they're easier to work with than strings that would have to be re-parsed (and the error handling that goes with it).

mjlshen:
No qualms, currently resource ID is the one that needs to be re-parsed from this originalPath string to maintain casing https://github.com/Azure/ARO-HCP/pull/437/files#diff-4e6d42d6c7f64263ba205c4f395ea62c7e99d7f223e73ad80bafcff8e2afa5b7R115, but if we refactor these two calls away so that it's easier to pass a resource ID instead I'm all for it!

mbarnes:
Yeah, the resource ID struct I recently added to the request context is built from originalPath (with original casing) so it should be a super simple refactor.

PR #438, opened 2024-08-02
Update Resource Naming Restrictions

bennerv:
Updates resource naming restrictions for `hcpOpenShiftClusters` and `hcpOpenShiftClusters/nodePools`


frontend/pkg/frontend/const.go
```diff
@@ -17,4 +17,7 @@ const (
 	PathSegmentDeploymentName    = "deploymentname"
 	PathSegmentActionName        = "actionname"
 	PathSegmentNodepoolName      = "nodepoolname"
+
+	HCPClusterResourceType = "hcpopenshiftclusters"
+	NodePoolResourceType   = "hcpopenshiftclusters/nodepools"
```
mbarnes, line 22: Can you move the node pool constant to `internal/api/registry.go`?
We already have a constant for cluster resource type defined there.

mbarnes (dismissed):
Just a small nit about where constants are defined.

frontend/pkg/frontend/const.go
bennerv: Done

mbarnes (dismissed):

bennerv:
There's going to be a problem with nesting node pool resource types as the resource type name.  Need to fix it quick.  Moving to draft.

mjlshen (approved):

PR #441, opened 2024-08-02
Fix header writes after writing body

bennerv:
### What this PR does

Fixes error messages in writing cloud errors.  
```
2024/08/02 18:54:23 http: superfluous response.WriteHeader call from github.com/Azure/ARO-HCP/internal/api/arm.WriteCloudError (error.go:106)
```

You're supposed to write the header status code, otherwise golang will default jto a 200 status.

```
Status Code
The resource provider must return 200 (OK) to indicate that the operation completed successfully. If the resource does not exist, 404 (NotFound) should be returned. If the resource group does not exist, 404 (NotFound) will be returned by the proxy *without* reaching the resource provider.
If the subscription does not exist, 404 (NotFound) will be returned by the proxy *without* reaching the resource provider. If the subscription is not registered with the resource provider, it should return a 404 (NotFound) or an empty collection.
If there are no types and all parent entities (resource group and subscription) exist, it should return an empty collection. It must not return a 5xx status code or 403 for this specific condition.
For other errors use the appropriate HTTP error code (e.g. 500 for Internal server error).


stevekuznetsov (commented):
Good catch - it looks like other handlers in this file are making the same mistake, could we fix them all please?

mjlshen (approved):

PR #451, opened 2024-08-06
ARO-9365 acrContributorPermission for cs managed identity

ritmun:
### What this PR does

Gives contributor role to clusters-service managed identity in arohcpdev ACR

Jira: 
[ARO-9365](https://issues.redhat.com/browse/ARO-9365)
Link to demo recording:


dev-infrastructure/templates/svc-cluster.bicep
```diff
@@ -277,3 +280,22 @@ module acrPushRole '../modules/acr-permissions.bicep' = [
     }
   }
 ]
+
+resource clustersServiceAcrResourceGroups 'Microsoft.Resources/resourceGroups@2023-07-01' existing = [
```
mjlshen, line 284: We could potentially combine `clustersServiceAcrResourceGroupNames` with `imageSyncAcrResourceGroupNames` if we expect that these will always be the same. I am not sure if this is true or not since we may have multiple ACRs

mjlshen (approved):
LGTM

but open question for others to confirm/deny

PR #453, opened 2024-08-07
Refactor ARO HCP cluster ingress API 

mjlshen:
### What this PR does
* Only the default ingress is configurable, previously a slice of
  ingresses could be provided.
* The IP field has been removed. It may be added back in the future if
  there is a compelling use-case.

* The devcontainer was no longer working since the top-level package-lock.json was moved under `api`


bennerv (approved):
lgtm

PR #461, opened 2024-08-08
make managed grafana name configurable

geoberle:
### What this PR does

the name for managed grafana needs to be unique, so making it configurable allows us to adapt in different environments. for personal dev environments our defaults ensure uniqueness

```
NameNotUnique - The Grafana Workspace name 'aro-hcp-grafana' is already being used. Please choose another name.
```

Jira: 
Link to demo recording:


geoberle:
close in favor of #460 460

PR #473, opened 2024-08-08
Adds Entra auth for CS postgres

markturansky:
Adds Entra authentication to CS deployment


bennerv:
Since we don't have prow, no /hold label applies so moved into draft to prevent accidental merge.

markturansky:
@geoberle @bennerv I've modified this and the deployment template to use the Workload ID param for postgres.  I'm specifically setting this workload ID as a new param to avoid conflict with the single secret creds we have today.

miguelsorianod:
We can close this as it's been obsoleted by other changes.

We have rolled out changes to the P1 environment and we are currently running CS against postgresql Entra DB

PR #478, opened 2024-08-09
Add Component sync job

janboll:
### What this PR does

Jira: 
Link to demo recording:


mjlshen (changes requested):
Since this is a new component, can we refactor its deployment to either kustomize or helm? We have agreed that we will not use oc templates in higher environments. If it's urgent to roll this out, I am happy to be overruled for now

mjlshen (approved):

PR #480, opened 2024-08-09
add admin api

katherinelc321:
### What this PR does
Adding an admin api for geneva actions

Jira: https://issues.redhat.com/browse/ARO-6238
Link to demo recording:


mbarnes (changes requested):
Code looks okay overall but can you please move the `internal/adminapi` directory to `internal/api/admin`.  (Generally the pattern is `internal/api/{version}` where `{version}` is a value that would be passed as an `api-version` URL parameter.)

mociarain:
> Code looks okay overall but can you please move the `internal/adminapi` directory to `internal/api/admin`. (Generally the pattern is `internal/api/{version}` where `{version}` is a value that would be passed as an `api-version` URL parameter.)

For reference this is how it looks in "Classic" after a few years: https://github.com/Azure/ARO-RP/tree/master/pkg/api

mbarnes (commented):
Question for @bennerv or @mjlshen :
Do we need a typespec definition for the admin API?  I see that this RP is lacking any generated API structs.  For now it would just be a copy of the **v20240610preview** API, but this might be a good opportunity to work out how we're going to manage multiple API versions under `/api` going forward and reorganize files as needed.

bennerv:
> Do we need a typespec definition for the admin API
No.

PR #492, opened 2024-08-12
Add static validation to subscription registration (and related refactoring)

mbarnes:
### What this PR does

Fixes a bug reported during P1 testing (see Jira for details).

Jira: [ARO-9619 - ARO-HCP RP: subscription handling doesn't follow RP Contract v1.0](https://issues.redhat.com/browse/ARO-9619)
Link to demo recording: 

The RPC states that `properties.additionalProperties.resourceProviderProperties.resourceProviderNamespace` is required in a subscription registration payload, but I'd like to find a second citation before marking it required it in the code.  ARO-RP doesn't validate this field at all.


internal/api/validate.go
```diff
@@ -0,0 +1,184 @@
...
+			if strings.HasPrefix(tag, "enum_") {
+				message += fmt.Sprintf(" (must be one of: %s)", fieldErr.Param())
+			} else {
+				switch tag {
+				case "api_version": // custom tag
+					message = fmt.Sprintf("Unrecognized API version '%s'", fieldErr.Value())
+				case "required", "required_for_put": // custom tag
+					message = fmt.Sprintf("Missing required field '%s'", fieldErr.Field())
+				case "cidrv4":
+					message += " (must be a v4 CIDR address)"
```
mjlshen, line 136:
```suggestion
					message += " (must be a v4 CIDR range)"
```

internal/api/validate.go
```diff
@@ -0,0 +1,184 @@
...
+	// There is no PATCH method for subscriptions, so assume PUT.
+	errorDetails := ValidateRequest(validate, http.MethodPut, subscription)
+	if errorDetails != nil {
+		cloudError.Details = append(cloudError.Details, errorDetails...)
+	}
+
+	switch len(cloudError.Details) {
+	case 0:
+		cloudError = nil
+	case 1:
```
mjlshen, line 178:
```suggestion
	default:
```
This might be my own ignorance, but is this field guaranteed to have a length of either 0 or 1?

mjlshen (commented):
LGTM overall, just two pieces of nit-level feedback

internal/api/validate.go
mbarnes: The default case (len>1) would leave the error as is, so it's omitted.

In the case of len=1 we rearrange the error structure to avoid ending up with:
```
{
  code: "MultipleErrorsOccurred",
  message: "Content validation failed on multiple fields"
  details: [
    ... only one error is listed, not multiple ...
  ]
}
```

mjlshen: Oh, thanks. I wasn't catching that before entering this switch statement, it was already formatted for len>1. This switch statement is solely for modifying the format for len == 0 or 1

PR #496, opened 2024-08-13
Rename "api" directory to "typespec"

mbarnes:
I've been wanting to rename this directory since the earliest days of ARO-HCP but I was holding off so as not to disrupt Petr's API specification work.

In my opinion, "typespec" is a more precise name for the content of the directory than just "api".

Compare with [ARO Classic's "swagger" directory](https://github.com/Azure/ARO-RP/tree/master/swagger) which holds _its_ API specifications.

I tried my best to update all the directory name references in our scripts and documentation.


mociarain:
*bikesheding alert* FWIW I think `api` is a better name. `typespec` is an implementation detail IMO. I'm still getting to grips with HCP so feel free to disregard this

mbarnes:
We also have an [internal/api](https://github.com/Azure/ARO-HCP/tree/main/internal/api) directory containing Go code, so an additional goal with this is to eliminate the ambiguity when we refer to the "api" directory.

mbarnes:
Got :-1: from multiple team members so closing.

PR #502, opened 2024-08-14
Remove subnet delegate

geoberle:
### What this PR does

do not merge. is use this to test the cluster creation without subnet delegate 

Jira: 
Link to demo recording:


bennerv:
AKS is fixing a bug on their side.  Don't need this as we're doing thet correct thing.  Closing :)

PR #513, opened 2024-08-15
Implement Get/Create/Delete CosmosDB Nodepool operations

mjlshen:
Jira: [ARO-9686](https://issues.redhat.com/browse/ARO-9686)

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -202,16 +202,89 @@ func (d *CosmosDBClient) DeleteClusterDoc(ctx context.Context, resourceID *azcor
 }
 
 func (d *CosmosDBClient) GetNodePoolDoc(ctx context.Context, resourceID *azcorearm.ResourceID) (*NodePoolDocument, error) {
-	panic("implement me")
+	// Make sure lookup keys are lowercase.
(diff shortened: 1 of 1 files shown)


internal/database/database.go
```diff
@@ -202,16 +202,89 @@ func (d *CosmosDBClient) DeleteClusterDoc(ctx context.Context, resourceID *azcor
...
+	if err != nil {
+		return err
+	}
+
+	container, err := d.client.NewContainer(nodePoolsContainer)
+	if err != nil {
+		return err
+	}
+
+	_, err = container.UpsertItem(ctx, azcosmos.NewPartitionKeyString(doc.PartitionKey), data, nil)
```
bennerv, line 257: todo: should make sure we're taking into account concurrency when we run multiple creates with the same name.

bennerv (approved):

internal/database/database.go
bennerv: There's a card for this somewhere.  Hence we can ignore it for now

mjlshen: Yup, [ARO-9674](https://issues.redhat.com/browse/ARO-9674)

PR #526, opened 2024-08-20
Make managedResourceGroup optional

mociarain:
## What:

This is a draft MR for https://issues.redhat.com/browse/ARO-9525. After discussing this with some members of the HCP team there was a weak preference to place the fallback logic in the CS rather than in the RP. Open to changing this but if this looks correct I'll add tests and we can go from there.


mbarnes (dismissed):
Can you also please remove the `validate:"required_for_put"` struct tag from this field in `/internal/api/hcpopenshiftcluster.go`?

mjlshen (approved):

PR #527, opened 2024-08-20
refactor metrics infrastructure

tony-schndr:
### What this PR does
Align dev metrics infrastructure with upstream bicep modules published from ARO-Pipelines.
Decouple metrics infrastructure from regional bicep templates, metrics infrastructure will be deployed separately in higher environments.
Use az CLI to enable aks cluster metrics to avoid needing to maintain additional bicep templates. (This aligns with the vision for higher environments)

Jira: https://issues.redhat.com/browse/ARO-9621


.github/workflows/aro-hcp-dev-env-cd.yml
```diff
@@ -161,6 +168,14 @@
                 --parameters kvNames="['${SVC_KV_NAME}']" \
                 --parameters githubActionsPrincipalID=${{ secrets.GHA_PRINCIPAL_ID }}
 
+              # enable aks metrics
+              AZ_MONITOR_RESOURCE_ID=$(az deployment group show --resource-group "${REGIONAL_RESOURCEGROUP}" --name "metrics-infra-${GITHUB_RUN_ID} --output tsv --query properties.outputs.monitorId.value)
```
janboll, line 172: How quick is `az deployment group show`? I think it could be refactored to one call if it is slow

.github/workflows/aro-hcp-dev-env-cd.yml
```diff
@@ -209,6 +223,14 @@
                 --parameters grantCosmosAccess=false \
                 --parameters githubActionsPrincipalID=${{ secrets.GHA_PRINCIPAL_ID }}
 
+              # enable aks metrics
+              AZ_MONITOR_RESOURCE_ID=$(az deployment group show --resource-group "${REGIONAL_RESOURCEGROUP}" --name "metrics-infra-${GITHUB_RUN_ID} --output tsv --query properties.outputs.monitorId.value)
```
janboll, line 227: See above comment

janboll (approved):

.github/workflows/aro-hcp-dev-env-cd.yml
tony-schndr: It is not slow but I agree it is redundant.  I'll need to dig into how it can be set once as a global so both steps have access to it.

mjlshen (approved):

PR #528, opened 2024-08-20
Switch from clustersmgmt v1 to v2alpha1

mbarnes:
### What this PR does

Self-explanatory, I think.  We should start using the newer OCM API in the RP.

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -17,7 +17,7 @@ import (
 	"sync/atomic"
 
 	"github.com/google/uuid"
-	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	cmv2alpha1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v2alpha1"
 
 	"github.com/Azure/ARO-HCP/internal/api"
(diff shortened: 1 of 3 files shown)


mjlshen (commented):
Do we know whether cmv2alpha1 is ready for use?

mbarnes:
I don't think v2alpha1 is finalized, but the RP can currently build against it.  If API changes break the RP, I'd rather adapt quickly than defer certain parts of async ops until everything's set in stone.

mjlshen:
> I don't think v2alpha1 is finalized, but the RP can currently build against it. As I understood, v2alpha1 is initially just v1 and they're introducing changes gradually. If an API change breaks the RP, I'd rather adapt quickly than defer certain parts of async ops until everything's set in stone.

Oh, I'm with you on that. I'm just not sure whether CS themselves thinks this is functional/ready for us to use yet or not. I can start a thread to check

mbarnes:
This was answered in your [Slack thread](https://redhat-internal.slack.com/archives/C05Q232L29E/p1724213935375989):

> As far as I know the v2alpha1 uses the same endpoint handlers than the v1. So it is buildable but it will require further implementation
> So it is ready and functional in a sense that it will behave the same as v1

mjlshen (approved):

PR #529, opened 2024-08-21
Frontend health should not depend on Database

janboll:
### What this PR does

- Remove dependecy to database from healthchecks
- Use global prometheus registry

Jira: 
Link to demo recording:

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -83,14 +83,7 @@ func (f *Frontend) Join() {
 	<-f.done
 }
 
-func (f *Frontend) CheckReady(ctx context.Context) bool {
-	// Verify the DB is available and accessible
-	if err := f.dbClient.DBConnectionTest(ctx); err != nil {
-		f.logger.Error(fmt.Sprintf("Database test failed: %v", err))
-		return false
-	}
-	f.logger.Debug("Database check completed")
-
+func (f *Frontend) CheckReady() bool {
 	return f.ready.Load().(bool)
 }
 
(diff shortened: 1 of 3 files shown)


frontend/pkg/frontend/frontend.go
mjlshen, line 1: Could you help me understand the motivation behind the changes in this file? Not strongly opposed, just curious about the background.

janboll: Just wanted to learn more how the frontend is implemented and stumbled over that part. 

Not depending on databases and other external services comes from experience. It can cause cascading failures which can be even harder to recover from.

mjlshen: I can see that in terms of multiple microservices, but I think a ReadinessProbe defined against a database connection test is a normal usage: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes

We don't want this pod to accept any requests if it cannot connect to the database

janboll: It's personal preference then. I prefer pods to keep running and handle errors in my application rather than have them restart once the database connections got issues.

Feel free to discard and close this PR.

geoberle: this health check is used for readiness and lifeness probe alike, right?

lifeness: including a DB check into a lifeness probe is not the right thing to do in my opinion. it will rarely solve the issue if the DB has a problem and can have negative impact on the DB.

readiness: what is the expectation of ARM towards an RP? is it preferrable to have the RP answering with an error or not answering at all? if the DB is not working it would affect all pods, and a readiness probe including a DB check would empty the endpoint list of the service. i prefer a service to remain accessible and answer with proper error message. but i'm not going to die on that hill.

tldr: i think Jans change is the right thing to do. we can revisite if lifeness / readiness probes can/should check different things

geoberle: also: not sure how we are going to define scrape targets for prometheus. in general prometheus will not scrape non-ready pods when using a `ServiceMonitor` as they will not show up in the endpoint list of a service.

@tony-schndr what scrape-config approach are we going to leverage?

tony-schndr: @geoberle I'm not certain at this time, I have only had success with Prometheus annotations using Istio's metric merging feature.  `ServiceMonitor` can't connect due to Azure Managed Prometheus not having the certificate to satisfy strict mTLS.  I'm going to raise this with the team after recharge/f2f, maybe there is something I'm missing.

mjlshen: I'm ok to move the DB check into a startup probe and definitely agree the liveness probe has no need to check the database.

I can see how the RP responding with a 500/Internal Server Error is preferable and we can alert on that.

SudoBrendan: Agree with other reviewers here, we should not validate dependencies ever to serve `/healthz` endpoints. Customers should get 5xx, not `tcp i/o connection timeout` when our database is down, since the liveness will determine if we are an Endpoint in the Service (if I am understanding our architecture correctly?)

PR #541, opened 2024-08-28
First party application service principal

janboll:
Add biceps files and scripts for first party application service principle

### What this PR does

Jira: 
Link to demo recording:


dev-infrastructure/configurations/mvp-first-party.bicepparam
```diff
@@ -0,0 +1,5 @@
+using '../templates/first-party-identity.bicep'
+
+param kvCertOfficerManagedIdentityName = 'aro-hcp-dev-first-party'
+param globalKeyVaultName = 'aro-hcp-dev-global-kv'
```
bennerv, line 4: Why a new key vault?  Why not the service key vault?

dev-infrastructure/modules/first-party-identity.bicep
```diff
@@ -0,0 +1,128 @@
...
+          'Microsoft.Resources/subscriptions/resourceGroups/read'
+          'Microsoft.Resources/subscriptions/resourceGroups/write'
+          'Microsoft.Authorization/*/action'
+        ]
+        notActions: []
+      }
+    ]
+    assignableScopes: [
+      subscription().id
+      '/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/resourcegroups/global'
```
bennerv, line 79: Shouldn't be hardcoding subscriptions in our bicep.  `subscription().id` should be sufficient.

dev-infrastructure/modules/first-party-identity.bicep
```diff
@@ -0,0 +1,128 @@
...
+    ]
+  }
+}
+
+// This is not supported as of now, since service principal creation is not supported in deployment scripts
```
bennerv, line 84: There's a Microsoft Graph bicep extension we can probably use to create an application.

bennerv (commented):
Can we remove the creation of the mock 1p app in `dev-infrastructure/scripts` 
- `dev-appplications.sh`
- `mock-dev-role-definition.json`

dev-infrastructure/modules/first-party-identity.bicep
janboll: :facepalm: I overlooked that one, will fix it

dev-infrastructure/modules/first-party-identity.bicep
janboll: The upside of using az cli here is, there is lots of code around retrieving the certificate from keyvault and using it to create the Service Principal. 

I'd rather reuse that existing code.

janboll:
Isn't there still a usecase for the script, like if an OCM dev wants to create it's own principal for local testing?

dev-infrastructure/configurations/mvp-first-party.bicepparam
bennerv: The service key vault will contain: 
- 1p certificate
- MSI 1p certificate
- RP TLS certificate

And all will be mounted using secrets CSI driver.  

What would be stored within the "global" key vault?  The 1p certificate is provisioned per region along with the tls certificates.

bennerv:
> Isn't there still a usecase for the script, like if an OCM dev wants to create it's own principal for local testing?

I guess we should've outlined use cases first and what we're trying to achieve.

1. Have a single mock 1p service principal used for all dev instances
- have a single custom role definition, if we need to update it, update the custom role definition

1. Provision a mock 1p service principal per dev instance
- need to clean up 1p service principal on deletion, otherwise we'll max out graph quota on apps for our tenant
- share or reuse custom role definition; there's a limit for these per tenant as well, so ideally share

We'll have to apply the same logic to the ARM helper identity.

dev-infrastructure/configurations/mvp-first-party.bicepparam
bennerv: why do we provision one key vault per region for 1p apps?  For redundancy. If a region goes down, it doesn't affect other regions.

janboll:
Todo: 

- ensure create-for-rbac script is idempotent 
- ensure user-assigned identity name does not collide with sp name

geoberle (approved):

geoberle (approved):

PR #551, opened 2024-09-02
Messin' with constants

mbarnes:
### What this PR does

Cherry picks from my async operations branch with the theme of tweaking constants (mostly route constants).


mociarain (approved):
LGTM

PR #554, opened 2024-09-03
ARO-8948 Update ARO HCP template with CSI Driver

ahitacat:
### What this PR does

This PR adds a SecretClassProvider that will store the first-party credential for CS deployment.
Additionally, a new make target and documentation have been added.

Jira:[ ARO-8948](https://issues.redhat.com/browse/ARO-8948)
Link to demo recording:

### DIFF
--- a/cluster-service/Makefile
+++ b/cluster-service/Makefile
@@ -5,6 +5,8 @@ CONFIG_PROFILE ?= dev
 include ../dev-infrastructure/configurations/$(CONFIG_PROFILE).mk
 
 CONSUMER_NAME ?= $(shell az aks list --query "[?tags.clusterType == 'mgmt-cluster' && starts_with(resourceGroup, '$(REGIONAL_RESOURCEGROUP)')].resourceGroup" -o tsv)
+KEYVAULT_NAME ?= $(shell az keyvault list --query "[?starts_with(name, 'service-kv')].name" -g ${RESOURCEGROUP} --output tsv)
+FPA_CERT_NAME ?= firstPartyMock
 
 deploy:
 	ZONE_RESOURCE_ID=$(shell az network dns zone list -g ${REGIONAL_RESOURCEGROUP} --query "[?zoneType=='Public'].id" -o tsv) && \
@@ -20,12 +22,17 @@ deploy:
 			-n clusters-service \
 			--query clientId) && \
 	CS_SERVICE_PRINCIPAL_CREDS_BASE64='$(shell az keyvault secret show --vault-name "service-kv-aro-hcp-dev" --name "aro-hcp-dev-sp-cs" | jq .value -r | base64 | tr -d '\n')' && \
+	TENANT_ID=$(shell az account show --query tenantId --output tsv) && \
+	KEYVAULT_NAME=$(shell az keyvault list --query "[?starts_with(name, 'service-kv')].name" -g ${REGIONAL_RESOURCEGROUP} --output tsv) && \
 	oc process --local -f deploy/openshift-templates/arohcp-service-template.yml \
 	  -p AZURE_CS_MI_CLIENT_ID=$${AZURE_CS_MI_CLIENT_ID} \
+	  -p TENANT_ID=$${TENANT_ID} \
 	  -p REGION=${REGION} \
+	  -p SERVICE_KEYVAULT_NAME=${KEYVAULT_NAME} \
 	  -p CS_SERVICE_PRINCIPAL_CREDS_BASE64=$${CS_SERVICE_PRINCIPAL_CREDS_BASE64} \
 	  -p IMAGE_REGISTRY=${ARO_HCP_IMAGE_ACR}.azurecr.io \
 	  -p IMAGE_REPOSITORY=app-sre/uhc-clusters-service \
+	  -p FPA_CERT_NAME=${FPA_CERT_NAME} \
 	  -p IMAGE_TAG=2256c75 | oc apply -f -
 
 deploy-integ:
--- a/cluster-service/Makefile
+++ b/cluster-service/Makefile
@@ -23,7 +23,6 @@ deploy:
 			--query clientId) && \
 	CS_SERVICE_PRINCIPAL_CREDS_BASE64='$(shell az keyvault secret show --vault-name "service-kv-aro-hcp-dev" --name "aro-hcp-dev-sp-cs" | jq .value -r | base64 | tr -d '\n')' && \
 	TENANT_ID=$(shell az account show --query tenantId --output tsv) && \
-	KEYVAULT_NAME=$(shell az keyvault list --query "[?starts_with(name, 'service-kv')].name" -g ${REGIONAL_RESOURCEGROUP} --output tsv) && \
 	oc process --local -f deploy/openshift-templates/arohcp-service-template.yml \
 	  -p AZURE_CS_MI_CLIENT_ID=$${AZURE_CS_MI_CLIENT_ID} \
 	  -p TENANT_ID=$${TENANT_ID} \
@@ -44,3 +43,8 @@ provision-shard:
 	sed -e "s#ZONE_RESOURCE_ID#$${ZONE_RESOURCE_ID}#g" -e "s/REGION/${REGION}/g" -e "s/CONSUMER_NAME/${CONSUMER_NAME}/g" deploy/dev-provisioning-shards.yml
 
 .PHONY: deploy deploy-integ provision-shard
+
+copy-fpc:
+	@FPC_SECRET=$(shell az keyvault secret show --vault-name "service-kv-aro-hcp-dev" --name ${FPA_CERT_NAME} --query "value" -o tsv) && \
+ 	../dev-infrastructure/scripts/kv-add-secret.sh ${KEYVAULT_NAME} ${RESOURCEGROUP} ${FPA_CERT_NAME} $${FPC_SECRET}
+.PHONY: copy-fpc
--- a/cluster-service/Makefile
+++ b/cluster-service/Makefile
@@ -32,7 +32,7 @@ deploy:
 	  -p IMAGE_REGISTRY=${ARO_HCP_IMAGE_ACR}.azurecr.io \
 	  -p IMAGE_REPOSITORY=app-sre/uhc-clusters-service \
 	  -p FPA_CERT_NAME=${FPA_CERT_NAME} \
-	  -p IMAGE_TAG=2256c75 | oc apply -f -
+	  -p IMAGE_TAG=1c6873c | oc apply -f -
 
 deploy-integ:
 	oc process --local -f deploy/integration/cluster-service-namespace.yaml | oc apply -f -
--- a/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
+++ b/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
@@ -551,6 +551,7 @@ objects:
           - --cloud-region-config=/configs/cloud-resources/cloud-regions.yaml
           - --cloud-region-constraints-config=/configs/cloud-resource-constraints/cloud-region-constraints.yaml
           - --azure-auth-config-path=/secrets/azure-credentials/azure-auth-config
+          - --azure-first-party-application-certificate-bundle-path=/secrets/keyvault/firstPartyApplicationCertificateBundle
           livenessProbe:
             httpGet:
               path: /api/clusters_mgmt/v1
--- a/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
+++ b/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
@@ -222,7 +222,36 @@ parameters:
 - name: CS_SERVICE_PRINCIPAL_CREDS_BASE64
   description: Base64 encoded service principal credentials for CS. This is temporary for P1
   required: true
+- name: SERVICE_KEYVAULT_NAME
+  description: The name of the Key Vault containing keys, secrets and certificates used to support services running on the cluster service.
+  value: ""
+- name: TENANT_ID
+  description: The Tenant ID of clusters service.
+  value: ""
+- name: FPA_CERT_NAME
+  description: The name of the secret that contains the first party application certificate bundle
+
 objects:
+
+- apiVersion: secrets-store.csi.x-k8s.io/v1
+  kind: SecretProviderClass
+  metadata:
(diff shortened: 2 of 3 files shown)


ahitacat:

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -222,8 +222,35 @@ parameters:
 - name: CS_SERVICE_PRINCIPAL_CREDS_BASE64
   description: Base64 encoded service principal credentials for CS. This is temporary for P1
   required: true
+- name: KEYVAULT_NAME
```
JameelB, line 225: @bennerv will there be a case where these types of credentials/certificates are moved to their own key vault? 

I'd like to make the name of this key vault specific. For example we also have the arm helper identity credentials. Are these going to be stored in the same key vault? cc @miguelsorianod 

Maybe something like `AZURE_FIRST_PARTY_APPLICATION_CREDENTIALS_KEYVAULT_NAME` or `AZURE_IDENTITY_CREDENTIALS_KEYVAULT_NAME`?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -222,8 +222,35 @@ parameters:
...
+  description: The name of the Key Vault where the first party credentials are stored.
```
JameelB, line 226:
```suggestion
  description: The name of the Key Vault where the first party application credentials are stored.
```

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -222,8 +222,35 @@ parameters:
...
+  value: "service-kv-aro-hcp-dev"
+- name: TENANT_ID
```
JameelB, line 228: Let's specify that this is the keyvault tenant id  in the name. 

Similar to my comment above, are all of our key vaults going to live in the same tenant? Or should we make this specific per environment. Maybe something like `AZURE_FIRST_PARTY_APPLICATION_KEYVAULT_TENANT_ID` or `AZURE_IDENTITY_CREDENTIALS_KEYVAULT_TENANT_ID` (where the arm helper identities can also use this tenant id)?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -222,8 +222,35 @@ parameters:
...
+  description: The Tenant ID of Key Vault.
+  value: ""
```
JameelB, line 230: Since we have a default value for the keyvault name, I think we should then specify the tenant id of that keyvault here as default value.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -222,8 +222,35 @@ parameters:
...
+
 objects:
 
+- apiVersion: secrets-store.csi.x-k8s.io/v1
+  kind: SecretProviderClass
+  metadata:
+    name: cs-keyvault
```
JameelB, line 237: name may need to change based on how we're organising keyvault identities. It may be the case we'll need multiple SecretProviderClass if we have multiple keyvaults to integrate with cs.

JameelB (commented):
Thanks for this @ahitacat! lgtm overall, some comments about naming. Maybe @bennerv, @miguelsorianod or @janboll can help here to give context on how these identity credentials may be organized.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -465,6 +498,9 @@ objects:
             subPath: cloud-region-constraints.yaml
           - name: azure-credentials
             mountPath: /secrets/azure-credentials
+          - name: keyvault
+            mountPath: "/secrets/keyvault"
+            readOnly: true
           env:
```
JameelB, line 504: Can we also add the flag `azure-first-party-application-certificate-bundle-path` since the file path is now known? 🙏  You may need to update the image tag of cs to an image tag that has this change, similar to what's done in this mr: https://github.com/Azure/ARO-HCP/pull/510/files

The latest tag in that repo `app-sre/uhc-clusters-service` is currently `1c6873c`, which matches the latest commit on cs master. Any reason not to use the latest tag @miguelsorianod @bennerv @janboll ?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
ahitacat: Would we like to create different key vaults to store different secrets? 
If so I'm ok with changing it, but this PR stores it in the private service-kv. Thus, I will propose (at this moment) to change it to `SERVICE_KEYVAULT_NAME`. What do you think?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
ahitacat: Similar question from above. In addition, this tenant is the same as the one required for maestro and hypershift. Note that for maestro we create a new key vault to store its related secrets.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
ahitacat: Related to the same questions above.

ahitacat:
## Latest Changes:

- Added a new commit that adds the `azure-first-party-application-certificate-bundle-path` and updates clusters-service tag to the latest.
- Added Tenant ID default value, and fixed a minor description.

I'm waiting for the names of the Keyvault and Tenant ID parameters questions to be resolved.
This PR will need to be closed and I need to make the changes in a branch of this project to make the pipelines run. But I prefer to do this once the review has finished.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
JameelB: I think we should make the name meaningful to CS as the current kv name may change. It may also not have the same name for other environments. I think `SERVICE_KEYVAULT_NAME` is a bit ambiguous though, what does `SERVICE` mean? Is it related to cluster service, the service cluster, or something else? 🤔 

As for multiple kv vs one kv, I think @bennerv or @janboll is best to answer that 🙏

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
JameelB: Note that we also provide other tenant ids for other purposes. For example, in the current aro auth config, there is a tenant id there used for authenticating clients within CS. I think it should be clear in the name as to what the tenant will be used for to avoid confusion and prevent people using this for different purposes other than what it's intended for.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
ahitacat: Make sense, I will wait for the answer of multiple kv vs one, so we can select one of your suggestions. As one describes a general kv than the other.

bennerv: service comes from the name of the key vault used to support the service running the resource provider.  There will be other secrets in the service key vault which contain TLS certificates, MDSD certificates (potentially) etc.  

It might make sense to align on naming across the components so we're all talking about the same thing, but it's up to CS to determine what they want to do with the name.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
bennerv: TenantID should always be the same no matter what key vault you're accessing.  The only difference would be if you're using the 1p credential to auth into the customer's tenant, but that's not a configuration that's set as a runtime env var or command-line flag.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
bennerv: We'll have multiple key vaults, but I believe this is the only one which will be used to mount a secret in the cluster for CS usage.  The others CS will access the data plane of the key vaults directly, rather than leveraging secrets CSI driver.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -222,8 +222,35 @@ parameters:
...
 objects:
 
+- apiVersion: secrets-store.csi.x-k8s.io/v1
+  kind: SecretProviderClass
+  metadata:
+    name: cs-keyvault
+    namespace: ${NAMESPACE}
+  spec:
+    parameters:
+      clientID: ${AZURE_CS_MI_CLIENT_ID}
```
bennerv, line 241: @geoberle should this be CS' identity used here? Shouldn't we have a dedicated identity for secrets CSI storage driver to mount the secret into the pod, rather than give it broad permissions CS has?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
bennerv: It looks added here.  I don't see a reason not to use it if CS has functionality for it.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
ahitacat: Then I'll opt for AZURE_IDENTITY_CREDENTIALS_KEYVAULT_NAME as it seems it will potentially contain other certs.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
JameelB: Then in that case, maybe just `KEYVAULT_TENANT_ID` would suffice? or to keep it in line with the service keyvault, `SERVICE_KEYVAULT_TENANT_ID`? 

There is another TenantID that we'd need to set as an env var, its the tenant id of where CS is running on and will be added/documented as part of https://issues.redhat.com/browse/ARO-7425

Will the tenant where CS is running on and where the service keyvault is located always be the same?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
JameelB: i'm fine with leaving it with this name then, feel free to close thread as resolved 👍

ahitacat: I can use something like `cs-azure-id-credentials-keyvault` to specify what the keyvault is for.

cluster-service/Makefile
```diff
@@ -20,13 +21,16 @@ deploy:
 			-n clusters-service \
 			--query clientId) && \
 	CS_SERVICE_PRINCIPAL_CREDS_BASE64='$(shell az keyvault secret show --vault-name "service-kv-aro-hcp-dev" --name "aro-hcp-dev-sp-cs" | jq .value -r | base64 | tr -d '\n')' && \
+	TENANT_ID=$(shell az account show --query tenantId --output tsv) && \
 	oc process --local -f deploy/openshift-templates/arohcp-service-template.yml \
 	  -p AZURE_CS_MI_CLIENT_ID=$${AZURE_CS_MI_CLIENT_ID} \
+	  -p TENANT_ID=$${TENANT_ID} \
 	  -p REGION=${REGION} \
+	  -p KEYVAULT_NAME=${KEYVAULT_NAME} \
```
geoberle, line 29: should this be named `SERVICE_KEYVAULT_NAME` to match the parameter name expected by the template?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -222,8 +222,35 @@ parameters:
 - name: CS_SERVICE_PRINCIPAL_CREDS_BASE64
   description: Base64 encoded service principal credentials for CS. This is temporary for P1
   required: true
+- name: SERVICE_KEYVAULT_NAME
+  description: The name of the Key Vault containing keys, secrets and certificates used to support services running on the cluster service.
+  value: "service-kv-aro-hcp-dev"
+- name: TENANT_ID
+  description: The Tenant ID of Key Vault.
+  value: "64dc69e4-d083-49fc-9569-ebece1dd1408"
```
geoberle, line 230: i would prefer to not provide a default here targeting a specific tenant

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
geoberle: good point. i would not hold off this PR from being merged because of this though. this is something we should do for other services as well, e.g. maestro-server

i've created some cards to track this work
https://issues.redhat.com/browse/ARO-10155

cluster-service/Makefile
ahitacat: good catch

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
ahitacat: I would prefer that too, @JameelB wdyt?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
ahitacat: Once those identities are created they can be changed here. Resolving the conversation.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -222,8 +222,35 @@ parameters:
...
+    parameters:
+      clientID: ${AZURE_CS_MI_CLIENT_ID}
+      cloudName: AzurePublicCloud
+      keyvaultName: ${SERVICE_KEYVAULT_NAME}
+      objects: |-
+        array:
+          - |
+            objectName: firstPartyMock
+            objectType: secret
+            objectAlias: "firstPartyMock"
```
JameelB, line 249: Been thinking about this, is it possible to parameterise it like the keyvault name? I assume for the different environments, the name isn't going to have 'mock' on them 🤔 wdyt?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
bennerv: Tenant will always be the same in which the service keyvault exists and CS runs.  No need to make it specific to the key vault instance.

ahitacat:
Superseded by https://github.com/Azure/ARO-HCP/pull/597

PR #558, opened 2024-09-04
Add Service Account and roles for integration

janboll:
### What this PR does
https://issues.redhat.com/browse/ARO-9988

Jira: 
Link to demo recording:

### DIFF
--- a/cluster-service/deploy/integration/cluster-service-namespace.yaml
+++ b/cluster-service/deploy/integration/cluster-service-namespace.yaml
@@ -39,7 +39,7 @@ objects:
   - apiVersion: rbac.authorization.k8s.io/v1
     kind: ClusterRoleBinding
     metadata:
-      name: cluster-service-naemspace-admin
+      name: cluster-service-namespace-admin
     roleRef:
       apiGroup: rbac.authorization.k8s.io
       kind: ClusterRole
--- a/cluster-service/deploy/integration/cluster-service-namespace.yaml
+++ b/cluster-service/deploy/integration/cluster-service-namespace.yaml
@@ -1,0 +1,50 @@
+---
+apiVersion: v1
+kind: Template
+metadata:
+  name: cluster-service-admin
+objects:
+  - apiVersion: v1
+    kind: Namespace
+    metadata:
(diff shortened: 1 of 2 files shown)


cluster-service/deploy/integration/cluster-service-namespace.yaml
```diff
@@ -0,0 +1,50 @@
...
+      namespace: cluster-service-admin
+  - apiVersion: rbac.authorization.k8s.io/v1
+    kind: ClusterRole
+    metadata:
+      name: namespace-admin
+    rules:
+    - apiGroups:
+      - ""
+      resources:
+      - namespace
```
oriAdler, line 24: Does the cluster role need to include: serviceaccounts and tokens?

cluster-service/deploy/integration/cluster-service-namespace.yaml
```diff
@@ -0,0 +1,50 @@
...
+      kind: ClusterRole
+      name: admin
+    subjects:
+    - kind: ServiceAccount
+      name: cluster-service-mgmt
+      namespace: cluster-service-admin
+  - apiVersion: rbac.authorization.k8s.io/v1
+    kind: ClusterRoleBinding
+    metadata:
+      name: cluster-service-naemspace-admin
```
oriAdler, line 42:
```suggestion
      name: cluster-service-namespace-admin
```

cluster-service/deploy/integration/cluster-service-namespace.yaml
oriAdler: I guess `cluster-admin` would be too much for our case, or given that it's a test environment, is it reasonable?

janboll: Tried it locally using minikube with 1.31, there it was not included in admin

janboll: I think clusteradmin would be to wide

oriAdler: Thanks for the clarification:)

PR #562, opened 2024-09-04
Split metrics endpoint to a different port, default 8081 

mjlshen:
### What this PR does
* This will allow metrics to be scraped over HTTP while the application itself behind port 8443 is secured by Istio mTLS
* `go work sync`

Jira: [ARO-9999](https://issues.redhat.com/browse/ARO-9999)


tony-schndr (approved):
lgtm, thanks for the quick turn around!

bennerv (approved):

PR #569, opened 2024-09-05
add admin component

katherinelc321:
### What this PR does
Adds a new folder for the admin api
Compiling and testing locally works with this:
`curl -X GET "0.0.0.0:8443/v1/ocm/clusters?api-version=2024-06-10-preview"`

Jira: https://issues.redhat.com/browse/ARO-9997
Link to demo recording:


admin/main.go
```diff
@@ -0,0 +1,31 @@
...
+	"time"
+
+	"github.com/Azure/ARO-HCP/admin/pkg/admin"
+)
+
+func main() {
+
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
+
+	listener, _ := net.Listen("tcp", ":8446")
```
mjlshen, line 17:
```suggestion
	listener, err := net.Listen("tcp", ":8443")
	if err != nil {
	  panic(err)
	}
```
I know the port doesn't really "matter", but I think we should use 8080 or 8443 for consistency

admin/main.go
```diff
@@ -0,0 +1,31 @@
...
+	// Initialize the Admin server
+	admin := admin.NewAdmin(logger, listener, "us-east")
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go func() {
+		time.Sleep(10 * time.Second)
+		cancel()
+	}()
```
mjlshen, line 28: What is this for?

admin/pkg/admin/routes.go
mjlshen, line 1: No need for most of this logic from frontend imo, this file can look similar to https://github.com/Azure/ARO-HCP/blob/59f266445c2ca27f55fbc8348753b971e10fd967/frontend/pkg/frontend/routes.go#L149-L154

admin/pkg/admin/adminResourceList.go
```diff
@@ -0,0 +1,156 @@
...
+	"net/http"
+	"net/url"
+	"strconv"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+)
+
+func (a *Admin) AdminResourceList(writer http.ResponseWriter, request *http.Request) {
```
mjlshen, line 16:
```suggestion
func (a *Admin) AdminClustersListFromCS(writer http.ResponseWriter, request *http.Request) {
```

admin/pkg/admin/adminResourceList.go
mjlshen, line 1: Up to you to decide (because the goal can totally change), but let's say that the goal of this Geneva Action is to get a list of all clusters from OCM. This will probably look something like:

```
GET /v1/ocm/clusters
```

With room for a future Geneva Action to provide cluster_id or cluster_name, like

```
GET /v1/ocm/clusters/id/${ID}
GET /v1/ocm/clusters?name=mycluster
```

I think this workload's "API" will be radically different from the RP/frontend

mjlshen (changes requested):

geoberle:
i'm closing this as it is 9 months old
@katherinelc321 feel free to reopen and rebase

PR #586, opened 2024-09-09
make multiple agentpool for usergroup

mukrishn:
AKS autoscaler profile does not enable this by default, it can be enabled by  `--balance-similar-node-group`  to maintain the node distribution.

### DIFF
--- a/dev-infrastructure/configurations/cs-integ-mgmt-cluster.bicepparam
+++ b/dev-infrastructure/configurations/cs-integ-mgmt-cluster.bicepparam
@@ -9,9 +9,10 @@ param aksKeyVaultName = 'aks-kv-cs-integ-mc-1'
 param systemAgentMinCount = 2
 param systemAgentMaxCount = 3
 param systemAgentVMSize = 'Standard_D2s_v3'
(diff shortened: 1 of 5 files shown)


weinong:
what does etcd's anti affinity rule look like?

mukrishn:
This is etcd pod antiaffinity spec in HCP. 

- Don't use the same node as other replicas 
- Don't use the same zone as other replicas

```
  affinity:                                                                                                           
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            app: etcd
        topologyKey: topology.kubernetes.io/zone
      - labelSelector:
          matchLabels:
            app: etcd
        topologyKey: kubernetes.io/hostname
```

stevekuznetsov (approved):

mjlshen:
@mukrishn Could you re-create this PR from a branch of Azure/ARO-HCP instead of from your fork, thanks! (That's why CI is failing :D)

mukrishn:
this is the new PR https://github.com/Azure/ARO-HCP/pull/590.

PR #590, opened 2024-09-10
multiple agentpool for usergroup

mukrishn:
AKS autoscaler profile does not enable this by default, it can be enabled by --balance-similar-node-group to maintain the node distribution.

PR: #586


mukrishn:
cc: @weinong @stevekuznetsov @mjlshen

weinong (approved):
LGTM

mukrishn:
https://github.com/Azure/ARO-HCP/pull/593

PR #593, opened 2024-09-10
multiple agentpool for usergroup

mukrishn:
AKS autoscaler profile does not enable this by default, it can be enabled by `--balance-similar-node-group` to maintain the node distribution.

PR: #590

### DIFF
--- a/dev-infrastructure/modules/aks-cluster-base.bicep
+++ b/dev-infrastructure/modules/aks-cluster-base.bicep
@@ -9,10 +9,10 @@ param systemAgentMaxCount int = 3
 param systemAgentVMSize string = 'Standard_D2s_v3'
 
 // User agentpool spec (Worker)
-param deployUserAgentPool bool = false
-param userAgentMinCount int = 2
+param userAgentMinCount int = 1
 param userAgentMaxCount int = 3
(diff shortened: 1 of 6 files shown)


mukrishn:
cc: @weinong @mjlshen @stevekuznetsov 
Thanks.

mjlshen:
I can temporarily remove all the agent pools from our dev AKS cluster and re-run CI

dev-infrastructure/modules/aks-cluster-base.bicep
```diff
@@ -103,9 +104,8 @@ var systemAgentPool = [
   }
 ]
 
-var userAgentPool = [
-  {
-    name: 'user'
+var userAgentPool = [for i in range(0, userAgentPoolAZCount): {
+    name: 'user-${(i + 1)}'
```
mjlshen, line 108:
```suggestion
    name: 'user${(i + 1)}'
```
The name of a node pool may only contain lowercase alphanumeric characters and must begin with a lowercase letter. For Linux node pools, the length must be between 1-12 characters. (Hyphen is not allowed)

mjlshen (approved):

PR #606, opened 2024-09-13
Refactor nodepools

mociarain:
### What this PR does

- Splits out the node pool code
- Makes some refactors to make the NodePools more testable
- Adds a basic test for NodePool creation to demonstrate it all working

The follow up for this will be several PRs to add more tests for the Cluster/resource methods and the other parts of the NodePool


frontend/apiversions.go
```diff
@@ -5,6 +5,7 @@ package main
 
 // This will invoke the init() function in each
 // API version package so it can register itself.
+// HERE
```
mociarain, line 8: Tests are failing because the apiVersions aren't getting populated because this part isn't being imported when testing. 

What's the right way to deal with this? AFAIK, in Classic we explicitly pass the valid versions when creating the NewFrontend

frontend/pkg/frontend/node_pool.go
```diff
@@ -0,0 +1,404 @@
...
+	if err != nil {
+		f.logger.Error(fmt.Sprintf("failed to write operation document: %v", err))
+		arm.WriteInternalServerError(writer)
+		return
+	}
+
+	writer.WriteHeader(http.StatusAccepted)
+}
+
+func (f *Frontend) OperationResult(writer http.ResponseWriter, request *http.Request) {
```
mbarnes, line 317: This function shouldn't be in node_pools.go.  It's an asynchronous operation endpoint, not node pool specific.

mbarnes (changes requested):

frontend/apiversions.go
mbarnes: Can't you copy the below import into your test?

mociarain: I could just import that in the test file or frontend.go but that seems wrong. We'll need to (potentially, so even worse) update those files every time we want to add a version.

Hmmm

frontend/pkg/frontend/ocm.go
```diff
@@ -140,13 +140,13 @@ func (f *Frontend) BuildCSCluster(resourceID *arm.ResourceID, tenantID string, h
...
-		additionalProperties["provision_shard_id"] = *f.clusterServiceConfig.ProvisionShardID
+	if f.clusterServiceClient.GetProvisionShardID() != nil {
+		additionalProperties["provision_shard_id"] = *f.clusterServiceClient.GetProvisionShardID()
 	}
-	if f.clusterServiceConfig.ProvisionerNoOpProvision {
+	if f.clusterServiceClient.GetProvisionerNoOpProvision() {
 		additionalProperties["provisioner_noop_provision"] = "true"
 	}
-	if f.clusterServiceConfig.ProvisionerNoOpDeprovision {
+	if f.clusterServiceClient.GetProvisionerNoOpDeprovision() {
```
mbarnes, line 149: This is totally optional for this pull request but I was thinking since `ClusterServiceClient` already has the optional cluster flags it would be cool to hide all this `additionalProperties` logic (including the hard-coded ones above) behind a `ClusterServiceClient` method that looks something like
```
func (csc *ClusterServiceClient) AddProperties(builder *cmv1.ClusterBuilder) *cmv1.ClusterBuilder
```
and then further down where we use `additionalProperties` just have
```
clusterBuilder = csc.AddProperties(clusterBuilder)
```

mbarnes (dismissed):
I like it, nice work.  I had one suggestion below but it can be done separately from this pull request (or not at all if you disagree).

frontend/pkg/frontend/ocm.go
mociarain: Oh I like this. I'll do it in a separate PR for clarity

mociarain: Done here: https://github.com/Azure/ARO-HCP/pull/720

mbarnes (approved):
LGTM!

mjlshen (approved):

PR #612, opened 2024-09-16
add helm linter github action

tony-schndr:
### What this PR does

Adds a Helm chart linter to CI checks.  The step to run the lint is conditional, running only if there are changes to any of the detected charts.

This uses default installed lint / schema configuration, if/when Charts become more defined we can update.
Jira: https://issues.redhat.com/browse/ARO-10414

Link to demo recording: 

Testing involved making a change to an existing chart and ensuring that the lint step actually ran and output was similar to what I experienced running ct locally.

### DIFF
--- a/.github/workflows/helm-lint.yml
+++ b/.github/workflows/helm-lint.yml
@@ -1,0 +1,55 @@
+---
+# Helm chart lint
+# More info at https://github.com/helm/chart-testing-action
+name: Lint Helm Charts
+
+on: pull_request
+
+jobs:
+  lint-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v4.2.0
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/helm-lint.yml
mjlshen, line 1: Could we version these by SHA with the most specific version commented if possible? Like https://github.com/Azure/ARO-HCP/blob/main/.github/workflows/ci-go.yml

This lets dependabot keep these up-to-date without getting any unexpected updates

mjlshen (commented):
functionally LGTM

.github/workflows/helm-lint.yml
tony-schndr: Yes, that would be ideal.  Where does the SHA come from?

mjlshen: Oh it's the git commit SHA that lines up with the tag

tony-schndr: I updated the pipeline I added here and other pipelines that were not using the sha.

geoberle (approved):
LGTM

PR #613, opened 2024-09-16
Add unit tests for "validate" struct tags

mbarnes:
### What this PR does

Adds unit tests for the various "validate" struct tags in use, which leverage the [package validator](https://github.com/go-playground/validator) module.  I also added some tweaks and corrections to existing "validate" tags based on this testing.

Another benefit to this test framework is it allows us to easily try out new validators, or combinations of validators, without having to stand up a full test environment.

Jira: none, was inspired by a [Slack thread](https://redhat-external.slack.com/archives/C075PHEFZKQ/p1726249159120609)
Link to demo recording:


mjlshen (approved):
This is great - thanks for working on it!

PR #616, opened 2024-09-17
Support overwrite for secrets

janboll:
Various quay orgs might require different secrets, thus allow chaning secret identifier

### What this PR does

Jira: 
Link to demo recording:


geoberle (commented):
i observed concurrency issues on multiple ARM RPs when creating things in parallel of the same kind. would you add @\batch annotations to the resourcelists to prevent that?

janboll:
I tried it with a new ACR, no issues. So I'd not add the batch annotation

geoberle (approved):

PR #618, opened 2024-09-17
frontend: Get tenant ID from x-ms-home-tenant-id header 

mbarnes:
### What this PR does

The JIRA bug below made me aware of the fact that the frontend assumes a tenant ID is always available from the subscription data in Cosmos, even though the tenant ID field is optional when registering a subscription.  (So states the [Subscription Lifecycle API Reference](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/subscription-lifecycle-api-reference.md).)

Instead, get the tenant ID from a [request header supplied by ARM](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/common-api-details.md):

> **x-ms-home-tenant-id**
>
> Added for requests at or below subscription scopes. Set to the tenant ID of the subscription.

For the convenience of test environments this also adds a hack to the subscription validation middleware, which already fetches subscription data from Cosmos.  If the **x-ms-home-tenant-id** request header is not present and we have a tenant ID in our own subscription data, then inject the header into the request before invoking the endpoint handler.

Jira: Related to [ARO-9925](https://issues.redhat.com/browse/ARO-9925) even though @bennerv already closed it; there's still an outstanding issue to address


frontend/pkg/frontend/middleware_validatesubscription.go
```diff
@@ -51,8 +51,20 @@ func (s *SubscriptionStateMuxValidator) MiddlewareValidateSubscriptionState(w ht
...
+	// from the SubscriptionDocument.
+	if r.Header.Get(arm.HeaderNameHomeTenantID) == "" {
+		if sub.Subscription != nil &&
+			sub.Subscription.Properties != nil &&
+			sub.Subscription.Properties.TenantId != nil {
+			r.Header.Set(
+				arm.HeaderNameHomeTenantID,
+				*sub.Subscription.Properties.TenantId)
+		}
+	}
```
mjlshen, line 66: No technical concerns/I think this will work. I'm trying to weigh whether or not we want to fail anyway (basically force non-prod environment clients to provide this header)

Pros:
- We don't unexpectedly use this logic in prod, admittedly this isn't a big deal because it should never happen, the header should always exist in prod.

Cons:
- I'm not sure if it's providing any value to force non-prod clients to do something differently than prod clients
- On the off chance that this header is actually missing in prod, this fallback logic helps avoid returning a 500

So I agree with this direction of keeping the code as-is

mjlshen (approved):

frontend/pkg/frontend/middleware_validatesubscription.go
mbarnes: We could mitigate your concern if we had a flag or environment variable or other means to know for certain whether we're deployed to prod (i.e. requests are actually coming from ARM).  At the moment we can only infer this from request headers.

PR #621, opened 2024-09-18
feat: add azure runtime config flag to cs service template

JameelB:
### What this PR does
Adds a configmap containing the azure runtime configuration in the ARO HCP CS service template. This is used as value to the flag `azure-runtime-config` which has been added in https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/8478

Current configuration available to be set here is the cloud environment where CS is running on. For now, only the cloud environment is added there. Additional configuration will be added at a later time. 

CS image tag has been updated to [9da687c](https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/tree/9da687cd0053da08b98c1a373080f122b9b99a66) which contains the changes in the MR linked above.

Jira: https://issues.redhat.com/browse/ARO-7653
Link to demo recording: N/A

Please note that this depends on https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/8478. The cs image in the makefile will be updated as soon as that's merged.

### DIFF
--- a/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
+++ b/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
@@ -360,6 +360,16 @@ objects:
               min_version: 4.11.0
 
 - apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: azure-runtime-config
+  data:
+    config.json: |
+      {
+        "cloudEnvironment": "AzurePublicCloud"
+      }
+
+- apiVersion: v1
   kind: ServiceAccount
   metadata:
     name: clusters-service
@@ -434,6 +444,9 @@ objects:
         - name: config
           configMap:
             name: clusters-service
+        - name: azure-runtime-config
+          configMap:
+            name: azure-runtime-config
         - name: mixin-pull-secret
           secret:
             secretName: hive-ci-global-pull-secret
@@ -506,6 +519,8 @@ objects:
           - name: keyvault
             mountPath: "/secrets/keyvault"
             readOnly: true
+          - name: azure-runtime-config
+            mountPath: /configs/azure-runtime-config
           env:
           - name: NAMESPACE
             valueFrom:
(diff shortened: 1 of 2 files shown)


cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -359,6 +359,16 @@ objects:
...
 
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: azure-runtime-config
+  data:
+    config.json: |
+      {
+        "cloudEnvironment": "AzurePublicCloud"
+      }
```
JameelB, line 370: In CS, we have a parameter that takes a json obj and set it here (see [MR#8478](https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/8478/diffs#24149fd14e0b7c80e0bcb139f89e767ed43c806f_332_340))
```
config.json: ${AZURE_RUNTIME_CONFIG}
```

I see that the rest of the configmap values are hardcoded here so i've kept it the same. Is there a preference to parameterise this? Additional configuration that would be added here for example would be ACR configuration that's used in CS (see related jira issue [ARO-9331](https://issues.redhat.com/browse/ARO-9331)).

geoberle: once additional dynamic configuration is required for an env, you can do something like this here 

https://github.com/Azure/ARO-HCP/blob/129323a66958d8122ecacecd1350af3e493967f2/maestro/deploy/helm/agent/templates/maestro.secret.yaml

so basically templating individual inner fields of the json value.

JameelB: ah perfect! I think we can do this as a follow on task, for now we can start with the hardcoded configmap 👍 I'll create a jira issue to track this.

mjlshen (approved):

PR #622, opened 2024-09-18
Add shellcheck linter for Makefiles

ahitacat:
### What this PR does

Add shellcheck linter for makefiles. Only use the code [SC1101](https://github.com/koalaman/shellcheck/wiki/SC1101), which will be triggered if space is found after backlash. This can cause several problems, i.e. https://github.com/Azure/ARO-HCP/pull/620 where the addition of that space causes the Makefile to fail silently.

The command options used are:
-i 1101: only look for these spaces
-f gcc: formats the output in a clear manner
`cluster-service/Makefile:27:15: error: Delete trailing spaces after \ to break line (or use quotes for literal space). [SC1101]`

### Why use shellcheck?
Shellcheck is part of the mega-linter, so that we can take advantage of it. 
Megalinter already has a Makefile check  `makecheck` but this linter doesn't check for shell code inside of it.

### How it looks
The new linter should have marked the trailing space.
image

### Notes
As it is using shellcheck that is meant for shell scripts, we need to be cautious when adding other checks. We are looking into Makefiles to check for these errors.


geoberle:
great idea 😍

janboll (approved):

geoberle (approved):

PR #626, opened 2024-09-20
reducing user agentpool maxpod limit to 225

mukrishn:
As we often hit resource crisis on the worker nodes due to HCP burst, leads to containerd and kubelet being OOM killed when running 250 pods per node irrespective of total available resources(noticed in D16 as well as E16 types)

Limiting the pod space on worker would help this situation,  at least burst caused by HCP pods were not affecting those daemon during 50 HCP test on perfscale environments.


mukrishn:
cc: @weinong @stevekuznetsov

mjlshen (approved):

PR #627, opened 2024-09-20
Fix 1st party application credentials in dev

mjlshen:
### What this PR does
This commit hardcodes the application ID of the mock first-party credentials for the dev environment. This is because the GitHub Actions entity does not have permissions to list applications at the tenant scope.

Fixes CD errors deploying CS, like https://github.com/Azure/ARO-HCP/actions/runs/10917480196/job/30436069484


bennerv (approved):
lgtm, not sensitive and we don't have admin consent right now so we'll not query it.

PR #628, opened 2024-09-20
Revert "reducing user agentpool maxpod limit to 225"

mjlshen:
Reverts Azure/ARO-HCP#626

This is to temporarily clean up the service/management clusters before recreating


mjlshen:
fyi @mukrishn , will re-revert this afterwards

bennerv (approved):

PR #563, opened 2024-09-04
Update frontend to Go 1.23

mjlshen:
### What this PR does
* We're not using any Go 1.23 features yet, but why not I guess?
* Also update golangci-lint to v1.60.3 and fix new linter findings

### Special notes for your reviewer

To double-check I'm not making up a SHA :D
```bash
❯ skopeo inspect --format "{{ .Digest }}" "docker://mcr.microsoft.com/oss/go/microsoft/golang:1.23.0-fips-cbl-mariner2.0" --override-arch amd64 --override-os linux
sha256:f182cc18fc741e51cb131d21a7f023a374c120e0ff49627238cefb719acc813d
```

PR #643, opened 2024-09-23
Update frontend to Go 1.23

mjlshen:
### What this PR does
* Recreates #563 after disabling CodeQL because it doesn't support Go 1.23 yet
* Also update golangci-lint to v1.61.0

To double-check I'm not making up a SHA :D
```bash
❯ skopeo inspect --format "{{ .Digest }}" "docker://mcr.microsoft.com/oss/go/microsoft/golang:1.23.0-fips-cbl-mariner2.0" --override-arch amd64 --override-os linux
sha256:f182cc18fc741e51cb131d21a7f023a374c120e0ff49627238cefb719acc813d
```

### DIFF
--- a/.devcontainer/postCreate.sh
+++ b/.devcontainer/postCreate.sh
@@ -21,7 +21,7 @@ npm install -g oav@3.3.4
 
 # Install the golang-lint
 # binary will be $(go env GOPATH)/bin/golangci-lint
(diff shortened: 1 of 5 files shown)


mbarnes (approved):
I actually had something come up recently where I thought "hey a generator function (or whatever Go calls it) would work well here", so :+1:

bennerv (approved):

PR #650, opened 2024-09-24
Ensure local golangci-lint version mirrors CI

mjlshen:
### What this PR does
This commit sets up a repo-local `tooling/bin/golangci-lint` that will use the same version as CI instead of using whatever local version of golangci-lint is available on a developer's device.

### DIFF
--- a/.devcontainer/postCreate.sh
+++ b/.devcontainer/postCreate.sh
@@ -19,12 +19,6 @@ cd api && npm ci
 # TODO: if we need to, we should move this out of here and into `api/package.json`
 npm install -g oav@3.3.4
 
-# Install the golang-lint
-# binary will be $(go env GOPATH)/bin/golangci-lint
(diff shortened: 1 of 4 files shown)


mbarnes (approved):
This looks good, but maybe we need a discussion about the purpose of our dev container if everyone is still just using their own local tools.  I'm guilty, but mostly because I don't typically use VSCode.

mjlshen:
I'm open to it, personally I use the devcontainer anytime I need to do TypeSpec/ARM API changes because I don't maintain my local npm setup at all

mbarnes:
>personally I use the devcontainer anytime I need to do TypeSpec/ARM API changes because I don't maintain my local npm setup at all - but never use it otherwise

Same.  Definitely prefer keeping all the Typespec and JavaScript stuff containerized, but I'm not sure the container really needs to be bound to VSCode.

PR #654, opened 2024-09-25
Add global keyvault

janboll:
- Moving private endpoint out, to reduce complexity for setting up new key vaults. Enables setting different scope for key vaults and deploying them to different resource groups.
- Create global service keyvault

### DIFF
--- a/dev-infrastructure/modules/keyvault/keyvault-private-endpoint.bicep
+++ b/dev-infrastructure/modules/keyvault/keyvault-private-endpoint.bicep
@@ -1,0 +1,72 @@
+param location string
+
+param keyVaultName string
+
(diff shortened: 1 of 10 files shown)


tony-schndr (approved):
lgtm

dev-infrastructure/modules/keyvault/keyvault-private-endpoint.bicep
```diff
@@ -0,0 +1,72 @@
+param location string
+
+param keyVaultName string
+
+param subnetId string = ''
+
+param vnetId string = ''
```
geoberle, line 7: can we get rid of the defaults?

geoberle (approved):

PR #663, opened 2024-09-29
frontend: Add azcore event logging, disabled by default

mbarnes:
### What this PR does

Debugging feature for [logging azcore events](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/log).

Enable by setting `AZURE_SDK_GO_LOGGING` to "all" in the `frontend-config` ConfigMap ("all" is the only valid value at present, which logs events for all [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go) clients).

Jira: Added this while working on [ARO-9674 - Frontend needs to run with at least two replicas](https://issues.redhat.com/browse/ARO-9674)
Link to demo recording:

### DIFF
--- a/frontend/cmd/cmd.go
+++ b/frontend/cmd/cmd.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	azlog "github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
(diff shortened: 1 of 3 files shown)


mbarnes:
Closed because the `AZURE_SDK_GO_LOGGING` environment variable doesn't behave the way the documentation implies.  Setting it to "all" enables the default logger, but calling [SetListener](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.14.0/log#SetListener) also enables logging regardless of the environment variable, which I wasn't expecting.

PR #676, opened 2024-09-30
ARO-10665 | Add docs to support local testing using first party appli…

cgiradkar:
…cation certificate bundle

### What this PR does

Jira: 
Link to demo recording:


dev-infrastructure/docs/development-setup.md
```diff
@@ -122,6 +122,12 @@ This will create:
 1. A custom role definition with required access as defined in `dev-infrastructure/scripts/mock-dev-role-definition.json`
 1. A service principal/application using the created cert as its authentication, and given access based on the custom role definition
 
+In the newly created Resource Group (test-RG):
```
miguelsorianod, line 125: For the other resources we don't specify test values in the documentation like test-RG, ... . We should be consistent with the guide and remove references. to example values. The same applies to test-Vnet, test-NSG, ...

dev-infrastructure/docs/development-setup.md
```diff
@@ -259,6 +265,13 @@ Option 2: You can follow the below manual steps from the root of the CS repo on
 # Setup the development.yml
 cp ./configs/development.yml .
 
+# Setup the azure-runtime-config.json
+cp ./configs/azure/example-config.json ./azure-runtime-config.json
+
+#  Get azure-first-party-application-client-id
+CLIENT_ID=$(az ad app list --display-name aro-hcp-dev-first-party --query '[*]'.appId -o tsv)
```
miguelsorianod, line 272: I am not sure that this display name is correct anymore. @janboll @JameelB could you check?

Aside from that, should we filter by display name or just specify the client id?

dev-infrastructure/docs/development-setup.md
```diff
@@ -259,6 +265,13 @@ Option 2: You can follow the below manual steps from the root of the CS repo on
 # Setup the development.yml
 cp ./configs/development.yml .
 
+# Setup the azure-runtime-config.json
+cp ./configs/azure/example-config.json ./azure-runtime-config.json
```
miguelsorianod, line 269: This works for now I guess but in the near future we will need to indicate exactly what content is the desired one

dev-infrastructure/docs/development-setup.md
```diff
@@ -344,7 +357,13 @@ instance_types:
 EOF
 ```
 
-3) Follow CS dev setup process:
+3) Get first-party application certificate bundle (Note that this will change when new FPA changes are introduced soon): </br>
```
miguelsorianod, line 360: I think with the current state the ("Note that this will change when new FPA are introduced soon)" can be removed.

There also seems to be some extra html markup (</br) not needed

dev-infrastructure/docs/development-setup.md
```diff
@@ -344,7 +357,13 @@ instance_types:
...
+Run the following command to generate a file containing the base64 decoded first-party application certificate bundle
+```shell
+$ az keyvault secret download --file fpa_cert --id=https://service-kv-aro-hcp-dev.vault.azure.net/secrets/firstPartyMock/<certi_id> | base64 -d > ~/fpa_cert_decoded
```
miguelsorianod, line 363: certi_id -> cert_id.

I think we might be able to make it easier to get the secret URL associated to the certficate than the user having to know the certificate id. cc @JameelB @janboll could you take a look?

Thanks

miguelsorianod, line 363: The command does not work.

The specified az keyvaut secret download command does not output to stdout so the base64 decoding through the pipe is not going to work.

Another way to specify this is in two commands. The first downloading the cert and the second base64 encoding.

We cleanup of the base64-encoded file should also be added as an instruction so there are no intermediate leftovers

dev-infrastructure/docs/development-setup.md
miguelsorianod, line 272: There are severa lclient IDs so I would be more specific and name the variable FPA_CLIENT_ID

dev-infrastructure/docs/development-setup.md
```diff
@@ -378,12 +397,12 @@ ocm login --url=http://localhost:8000 --use-auth-code
 ```bash
 NAME="<INSERT-NAME-HERE>"
 RESOURCENAME="<INSERT-NAME>"
-SUBSCRIPTION="<INSERT-NAME>"
-RESOURCEGROUPNAME="<INSERT-NAME>"
-TENANTID="<INSERT-NAME>"
+SUBSCRIPTION=$(echo $(az account subscription list | jq '.[] | select(.displayName == "ARO Hosted Control Planes (EA Subscription 1)")' | jq -r '.subscriptionId'))
+RESOURCEGROUPNAME="test-RG"
```
miguelsorianod, line 401: As mentioned in a comment above we can keep the <INSERT-NAME> content here

dev-infrastructure/docs/development-setup.md
```diff
@@ -378,12 +397,12 @@ ocm login --url=http://localhost:8000 --use-auth-code
...
+TENANTID=$(echo $(cat azure-creds.json | jq -r '.tenantId'))
```
miguelsorianod, line 402: I think we can keep <INSERT-NAME> here too. Even though this should work the usage of azure-creds.json is temporary so let's keep in unmodified

dev-infrastructure/docs/development-setup.md
```diff
@@ -378,12 +397,12 @@ ocm login --url=http://localhost:8000 --use-auth-code
...
 MANAGEDRGNAME="<INSERT-NAME>"
-SUBNETRESOURCEID="<INSERT-NAME>"
-$NSG="<INSERT-NAME>"
+SUBNETRESOURCEID="/subscriptions/$SUBSCRIPTION/resourceGroups/$RESOURCEGROUPNAME/providers/Microsoft.Network/virtualNetworks/test-Vnet/subnets/default-assoc"
```
miguelsorianod, line 404: Same comment as others about placeholders

dev-infrastructure/docs/development-setup.md
```diff
@@ -378,12 +397,12 @@ ocm login --url=http://localhost:8000 --use-auth-code
...
+$NSG="/subscriptions/$SUBSCRIPTION/resourceGroups/$RESOURCEGROUPNAME/providers/Microsoft.Network/networkSecurityGroups/test-NSG"
```
miguelsorianod, line 405: Same comment as others about placeholders

dev-infrastructure/docs/development-setup.md
```diff
@@ -122,6 +122,12 @@ This will create:
 1. A custom role definition with required access as defined in `dev-infrastructure/scripts/mock-dev-role-definition.json`
 1. A service principal/application using the created cert as its authentication, and given access based on the custom role definition
 
+In the newly created Resource Group (test-RG):
+1. Create a Virtual Network (test-Vnet) and a Network security group (test-NSG)
```
miguelsorianod, line 126: As an improvement these steps can be automated with commands. Not in scope of this PR necessarily

dev-infrastructure/docs/development-setup.md
JameelB, line 125: Not sure why this is in the `Step 2 - Create the Application and its dependencies` section? It's not related. This is a pre-requisite of creating an ARO HCP cluster, I would expect this to be closer to that documentation.

dev-infrastructure/docs/development-setup.md
JameelB: that's not the correct one any longer, I think the display name is `aro-dev-first-party`. The new client id is 57e54810-3138-4f38-bd3b-29cb33f4c358

If we specify the client id here and it changes, we would need to remember to update this doc too. Might be less likely for the name to change?

dev-infrastructure/docs/development-setup.md
JameelB: +1

alternatively, we could also just specify the vault name and name like `az keyvault secret download --vault-name "aro-hcp-dev-svc-kv" --name "firstPartyCert" -f`

Since these values are already used in the cs makefile, we could add a target for doing this. Doesn't have to be in this PR but would be nice in the future.

JameelB: Alternatively, you could use `az keyvault show .... --query "value"...` to output the b64 encoded cert then pipe it.

dev-infrastructure/docs/development-setup.md
```diff
@@ -357,10 +376,10 @@ make db/setup
 ./clusters-service init --config-file ./development.yml
 ```
 
-4) Start CS:
+5) Start CS:
 
 ```bash
-./clusters-service serve --config-file development.yml --runtime-mode aro-hcp --azure-auth-config-path azure-creds.json
+./clusters-service serve --config-file development.yml --runtime-mode aro-hcp --azure-auth-config-path azure-creds.json --azure-runtime-config-path ./azure-runtime-config.json --azure-first-party-application-certificate-bundle-path ~/fpa_cert_decoded --azure-first-party-application-client-id $CLIENT_ID
```
JameelB, line 382: All of these flags are in the development.yml file so you don't have to specify it as separate flags: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/configs/development.yml?ref_type=heads#L246-259

JameelB:
Not part of this MR but I think the whole section of running CS locally should really be in Cluster Service. Is there a reason for why that section is here? It would be nice to only have one place for this documentation rather than us updating two places.

cgiradkar:
> Not part of this MR but I think the whole section of running CS locally should really be in Cluster Service. Is there a reason for why that section is here? It would be nice to only have one place for this documentation rather than us updating two places.

If the same information is to be added in CS, then much of the steps will be overlapping with ARO-HCP docs. Was re-using the steps here.

JameelB:
> > Not part of this MR but I think the whole section of running CS locally should really be in Cluster Service. Is there a reason for why that section is here? It would be nice to only have one place for this documentation rather than us updating two places.
> 
> If the same information is to be added in CS, then much of the steps will be overlapping with ARO-HCP docs. Was re-using the steps here.

I'm not talking about the specific things you've added in btw, that was fine. But I do question if the documentation to run CS locally is in the right place. We already have overlapping documentation in CS atm: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/docs/aro_hcp_mode.md?ref_type=heads

cgiradkar:
> > > Not part of this MR but I think the whole section of running CS locally should really be in Cluster Service. Is there a reason for why that section is here? It would be nice to only have one place for this documentation rather than us updating two places.
> > 
> > 
> > If the same information is to be added in CS, then much of the steps will be overlapping with ARO-HCP docs. Was re-using the steps here.
> 
> I'm not talking about the specific things you've added in btw, that was fine. But I do question if the documentation to run CS locally is in the right place. We already have overlapping documentation in CS atm: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/docs/aro_hcp_mode.md?ref_type=heads
> 
> This is outside of your changes here, but I do want to start a discussion to consolidate these. If we update for example certain properties in the cluster payload, we'd have to update it here and the documentation in CS. Would be nice to just have one source of truth.

Agreed

dev-infrastructure/docs/development-setup.md
```diff
@@ -259,6 +259,13 @@ Option 2: You can follow the below manual steps from the root of the CS repo on
 # Setup the development.yml
 cp ./configs/development.yml .
 
+# Setup the azure-runtime-config.json
+cp ./configs/azure/example-config.json ./azure-runtime-config.json
```
JameelB, line 263: not clear here but the azure-runtime-config, the cert and the client id all needs to be set in the `development.yml` file. I would add a note to review the required properties there, it should be documented in the development.yml file itself. The section for getting the values for those required properties can be moved alongside that section. 

For example, FPA_CLIENT_ID env var is set below but it's not really used anywhere and it's not clear what/where it should be used for.

JameelB (approved):

mjlshen (approved):
/lgtm

PR #680, opened 2024-10-01
Add concurrency controls around Cosmos DB

mbarnes:
### What this PR does

In anticipation of running [multiple RP replicas](https://issues.redhat.com/browse/ARO-9674) as well as introducing a [polling backend for asynchronous operations](https://issues.redhat.com/browse/ARO-8598), we need to add concurrency controls around our Cosmos DB containers to ensure data integrity in the presence of multiple writers.

This PR takes a two-fold approach:

1. The first and simplest approach is to utilize [Optimistic Concurrency Control](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#optimistic-concurrency-control) when updating container items by setting the [IfMatchEtag option](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos#ItemOptions) and allowing for retries in the event of a **412 Precondition Failed** error.
   (covered in the 1st commit)

3. The second approach is to adapt Brian Dunnington[^1]'s [Cloud Distributed Lock pattern for Cosmos DB](https://github.com/briandunnington/CloudDistributedLock/) (see also [this YouTube video](https://www.youtube.com/live/Hreew-l5rCQ?si=F1DkRXHu3Tq4zeEk) of Brian demoing the pattern) to protect critical sections in request handlers. A critical section that requires protection through locks would involve the sequence of checking Cosmos for ongoing asynchronous operations that would block the current request, initiating a create, update, or delete operation through Cluster Service, and writing operation data, resource tags and system metadata to Cosmos.

   The [GitHub link](https://github.com/briandunnington/CloudDistributedLock/) explains the pattern best but I'll try to summarize:

   The pattern is somewhat similar to the leasing pattern ARO Classic uses to protect cluster documents, where all the data that needs protection is in one document and the lock (or lease) is built into the same document as the data.  The difference here is the lock is its own document, and leverages two built-in features of Cosmos:

   1. The lock container sets a short default time-to-live (TTL) as a fail safe in case the RP crashes or otherwise fails to release a lock.  Locks are only held briefly -- like a mutex -- so the TTL should only be a few seconds.
   1. Once a lock is acquired by creating a new container item, the RP uses [Optimistic Concurrency Control](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#optimistic-concurrency-control) to renew or release (delete) the lock.  Because of the short TTL, the RP must periodically renew the lock if it needs to hold it for longer periods.  This is done from a goroutine.

   For our purposes, we introduce a new Cosmos container named "Locks" and the IDs for items in this container are subscription IDs. So effectively the entire subscription is locked for the duration of a PUT, PATCH, or DELETE request (GET requests will not use locking).  If a PUT, PATCH, or DELETE request arrives while the subscription is locked, the caller gets a **409 Conflict** response with a **Retry-After** header set to the container's TTL value.

[^1]: Brian is a principal software engineering manager at Microsoft.

Jira: [ARO-10822 - Add concurrency controls around Cosmos DB](https://issues.redhat.com/browse/ARO-10822)
Link to demo recording: 

I think this is working well enough to merge, but it's not foolproof.

Because this is a cloud environment, weird things can happen and the RP can potentially lose the subscription lock it acquired.  This should be rare but is still possible.  When the lock is lost, the `context.Context` passed to the request handler is cancelled.  Depending on when the context is cancelled, the RP can potentially end up in an inconsistent state.

For example, suppose Cluster Service is taking a really long time to respond to a create cluster request.  During that time, the RP somehow loses its subscription lock and the context is cancelled.  Cluster Service eventually returns a response, but the context cancellation prevents the RP from recording the new cluster in Cosmos DB.

### DIFF
--- a/dev-infrastructure/modules/rp-cosmos.bicep
+++ b/dev-infrastructure/modules/rp-cosmos.bicep
@@ -27,6 +27,11 @@ var containers = [
   {
     name: 'Billing'
   }
+  {
+    name: 'Locks'
+    defaultTtl: 10
+    partitionKeyPaths: ['/id']
+  }
 ]
 
 param roleDefinitionId string = '00000000-0000-0000-0000-000000000002'
--- a/frontend/cmd/cmd.go
+++ b/frontend/cmd/cmd.go
@@ -118,7 +118,7 @@ func (opts *FrontendOpts) Run() error {
 			return err
 		}
 
-		dbClient, err = database.NewCosmosDBClient(cosmosDatabaseClient)
+		dbClient, err = database.NewCosmosDBClient(context.Background(), cosmosDatabaseClient)
 		if err != nil {
 			return fmt.Errorf("creating the database client failed: %v", err)
 		}
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -352,7 +352,6 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
 	}
 
 	var updating = (doc != nil)
-	var docUpdated bool // whether to upsert the doc
 	var operationRequest database.OperationRequest
 
 	var versionedCurrentCluster api.VersionedHCPOpenShiftCluster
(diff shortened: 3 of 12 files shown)


mjlshen (approved):
We discussed a bit offline about using the `CreateItem` vs `UpsertItem` API for creating resources. It's technically correct, but we will work on error handling/a background process/a Geneva Action to cleanup stale database records if needed in the future.

PR #701, opened 2024-10-10
add more templatize input parameters

geoberle:
### What this PR does

* support cloud, deploy-env and config-file as input parameters
* basic input parameter validation

Jira: 
Link to demo recording:

### DIFF
--- a/tooling/templatize/main.go
+++ b/tooling/templatize/main.go
@@ -1,10 +1,15 @@
 package main
 
 import (
+	"fmt"
 	"log"
+	"os"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/Azure/ARO-HCP/tooling/templatize/config"
 )
 
 func DefaultGenerationOptions() *GenerationOptions {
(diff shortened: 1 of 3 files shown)


tooling/templatize/main.go
```diff
@@ -1,24 +1,81 @@
...
-	User   string
+	ConfigFile string
+	Input      string
+	Cloud      string
+	DeployEnv  string
+	Region     string
+	User       string
+}
+
+func (opts *GenerationOptions) Validate() error {
```
janboll, line 28: Can be made more simple using cobra

janboll (changes requested):

tooling/templatize/main.go
```diff
@@ -1,24 +1,81 @@
...
+		}
+	}
+	return errors.NewAggregate(errs)
+}
+
+func (opts *GenerationOptions) validateFileExists(param, path string) error {
+	if len(path) == 0 {
+		return fmt.Errorf("parameter %s is missing", param)
+	}
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
geoberle: permission errors are now reported

tooling/templatize/main.go
geoberle: i covered the basic validations with cobra

whober0521 (approved):

tooling/templatize/main.go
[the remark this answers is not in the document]
stevekuznetsov: I know this is past, but wanted to make some comments here. There's no real point in trying to deduce all possible errors that can happen when opening or writing files. What if it's a network-attached filesystem and there's a desync? It's much better to just open and write to the files, and report the error, if there is one. In this case, even, doing _anything_ upfront might not be worth it. In the abstract, someone can race with the program and change filesystem permissions, so it's not like you can get away with not checking for errors when you actually go to use the files. Also, the tool takes O(microseconds) to run, not like doing it up-front saves much.

tooling/templatize/main.go
stevekuznetsov: I also know this is past, but a general comment from my experience: any CLI worth something is two things - a tiny wrapper of code to get input parameters from a command line, and a useful utility library on the inside. The utility library is where the value lies, and the moment it gets re-used, whether that's someone writing new automations that want to call the library as opposed to shelling out, or it's some CLI that embeds part of this one, that's the moment you can't use Cobra to validate your options. Your `func (opts *Options) Run()` needs to be certain about the content of `*Options` and if someone's creating that in Go code with a struct literal, your Cobra code won't run :)

PR #702, opened 2024-10-10
[Templatize Tooling] Local Development

whober0521:
### What this PR does
Local Development
- Get config values
- Create output folder if doesn't exist
- Generate output file using input template file and config values

Jira: 
Link to demo recording:


tooling/templatize/main.go
```diff
@@ -1,9 +1,13 @@
 package main
 
 import (
+	"context"
 	"fmt"
+	"html/template"
```
janboll, line 6: I think we want text templates? https://pkg.go.dev/text/template@go1.23.2

tooling/templatize/main.go
```diff
@@ -126,3 +130,34 @@ func main() {
...
+
+func (opts *GenerationOptions) ExecuteTemplate(ctx context.Context) error {
+	cfg := config.NewConfigProvider(opts.ConfigFile, opts.Region, opts.User)
+	vars, err := cfg.GetVariables(ctx, opts.Cloud, opts.DeployEnv)
+	if err != nil {
+		return err
+	}
+	// print the vars
+	for k, v := range vars {
+		println(k, v)
```
janboll, line 142: It's encouraged to use fmt.Printline. Please take a look https://go.dev/ref/spec#Bootstrapping

tooling/templatize/main.go
```diff
@@ -126,3 +130,34 @@ func main() {
 		log.Fatal(err)
 	}
 }
+
+func (opts *GenerationOptions) ExecuteTemplate(ctx context.Context) error {
```
janboll, line 134: I'd think we need to add some test from the start for this. Can you please add at least some basic test case?

janboll (approved):

PR #717, opened 2024-10-14
move maestro out of regional,svc,mgmt as independent conponent

jonathan34c:
### What this PR does
We want to move maestro out of svc, mgmt and regional as an independent component. The reason is that we want to deploy it individually when something is wrong. 

Jira: 
Link to demo recording:


dev-infrastructure/templates/maestro.bicep
```diff
@@ -0,0 +1,149 @@
...
+    privateEndpointVnetId: vnet.id
+    postgresServerPrivate: maestroPostgresPrivate
+    maestroServerManagedIdentityPrincipalId: svcUami[0].properties.principalId
+    maestroServerManagedIdentityName: 'maestro-server'
+    location: location
+  }
+}
+
+//
+//   M A E S T R O  C O N S U M E R
```
stevekuznetsov, line 117: Infra to server to consumer is a 1:1:N relationship - I don't know that we ever need to be deploying the server and the consumer(s) in one pass, and even if we do, there would be potentially many consumers for each server.

geoberle:
right now, the maestro infrastructure is part of three templates
* region.bicep - for the eventgrid infrastructure
* svc-cluster.bicep - manage eventgrid access from svc-cluster side
* mgmt-cluster.bicep - manage eventgrid access from the mgmt-cluster side

if we want to move out the maestro parts from these three templates, it should be again three templates with the mentioned scopes.

dev-infrastructure/templates/maestro.bicep
jonathan34c: i split the maestro template to 3 different components and added the MVP bicepparam files as well. thanks for the advice !

dev-infrastructure/templates/maestro-consumer.bicep
```diff
@@ -0,0 +1,61 @@
...
+
+resource mgmtUami 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' existing= [
+  for wi in mgmtWorkloadIdentities: {
+    name: wi.value.uamiName
+    scope: resourceGroup(mgmtResourceGroup)
+  }
+]
+func isValidMaestroConsumerName(input string) bool => length(input) <= 90 && contains(input, '[^a-zA-Z0-9_-]') == false
+
+module maestroConsumer '../modules/maestro/maestro-consumer.bicep' = if (deployMaestroConsumer) {
```
geoberle, line 49: i don't think we need this conditional anymore. this is not a "top-level" template and by running it we clearly indicate that we want the consumer stuff to happen

PR #720, opened 2024-10-15
Mociarain/refactor cluster service client

mociarain:
Small improvement based on feedback in [here](https://github.com/Azure/ARO-HCP/pull/606)

### DIFF
--- a/internal/ocm/ocm.go
+++ b/internal/ocm/ocm.go
@@ -13,9 +13,7 @@ import (
 
 type ClusterServiceClientSpec interface {
 	GetConn() *sdk.Connection
-	GetProvisionShardID() *string
-	GetProvisionerNoOpProvision() bool
-	GetProvisionerNoOpDeprovision() bool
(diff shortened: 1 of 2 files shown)


internal/ocm/ocm.go
```diff
@@ -26,6 +24,27 @@ type ClusterServiceClientSpec interface {
 	DeleteCSNodePool(ctx context.Context, internalID InternalID) error
 }
 
+// Get the default set of properties for the Cluster Service
+func getDefaultAdditionalProperities() map[string]string {
+	// additionalProperties should be empty in production, it is configurable for development to pin to specific
+	// provision shards or instruct CS to skip the full provisioning/deprovisioning flow.
+	additionalProperties := map[string]string{
```
mbarnes, line 31: nit: You could just return the map directly here.

mbarnes (approved):
Looks good.  One tiny nit but not a blocker for merging.

PR #725, opened 2024-10-15
tooling/templatize: use typed options, functional core

stevekuznetsov:
tooling/templatize: use typed options, functional core

The typed options pattern hides the validated and completed options in
opaque types that make it impossible for a consumer of the
`ExecuteTemplate()` function to pass some set of options that has NOT
been validated and completed. This makes it easy to trust the input and
make assumptions about it.

Furthermore, the `Complete()` implementation shows how to implement the
"imerative shell, functional core" pattern - imperative functionality
with side-effects like reading the filesystem can be done in a (small)
shell of imperative core, so that the real functionality of the library
is done in functional logic that is trivial to test.

An example of how to use this to our advantage is the new test for
template executioin - we pass raw data in, get raw data out. No need to
worry about filesytems, permissions, etc. We still *can* worry about
that in end-to-end tests if we need to.

---

*: use bingo, add formatter target

---


.bingo/README.md
```diff
@@ -0,0 +1,14 @@
+# Project Development Dependencies.
```
janboll, line 1: Bingo changes the go tooling for the entire repo. Can you please move this to a separate PR?

janboll (changes requested):

.bingo/README.md
SudoBrendan: I agree. +1 on Bingo and +1 on a separate PR to make that decision as a team.

geoberle (approved):

PR #727, opened 2024-10-16
install MCE via helm

geoberle:
### What this PR does

This PR replaces current OLM based ACM installation with a helm based one. In addition to that, this PR also reduced the ACM footprint and only installs the required components `multicluster-engine` (MCE) and `policy` (for RVMO).

Installation and configuration are split into two individual helm charts

### multicluster-engine

This helm chart installs the MCE operator. The chart is autogenerated from an OLM release-bundle of MCE by running `make helm-chart`. The resulting chart is then commited into the repository.

### multicluster-engine-config

This helm chart configures MCE and installs the `policy` component on top of it. The subchart for `policy` is imported into this repo by running `make helm-chart`. It can be found [in the ACM upstream Github org](https://github.com/stolostron/mce-install-kube) and is maintained by the ACM team.

⚠️ after merging this we need to reinstall the integrated DEV MC and CS RP MC as we can't switch from an OLM managed ACM to a helm one

Jira: https://issues.redhat.com/browse/ARO-10315


tony-schndr (approved):
lgtm

PR #746, opened 2024-10-21
tooling templateize - naming convention utils

geoberle:
### What this PR does

* add utils for azure resource name generation (e.g. max length, suffix handling for uniqueness)
* add region stamp and CX stamp as parameters for unique name generation
* add inputparameters under `.ctx` dict in templating paramters

especially the region and CX stamp flags introduce usage parity between personal DEV and higher environments. while in DEV, the region stamp would be `user`, in higher envs it would be usually a number indicating the stamp number of a setup within a region (usually `1` since we intend to have only one setup per region).

````
go run . --region-stamp $USER ...`
vs
go run . --region-stamp 1 ...`

Jira: <!-- optional: link to Jira issue -->
Link to demo recording: <!-- optional: link to a demo recording -->

<!-- optional -->

### CODE REVIEW

File: tooling/templatize/config/config.go [hunk 1]
```diff
@@ -50,19 +54,43 @@ func (cp *configProviderImpl) GetVariables(cloud, deployEnv string) (Variables,
...
+			"cloud":       cloud,
+			"deployEnv":   deployEnv,
+			"regionStamp": cp.regionStamp,
+			"cxStamp":     cp.cxStamp,
+		},
+	}
 
-	if err == nil {
-		var tmplBytes bytes.Buffer
+	functions := template.FuncMap{
```
whober0521, line 68: FuncMap should be in main.go instead of config.go. Since these functions are only used in local development, not in pipeline

tooling/templatize/config/config.go
```diff
@@ -50,19 +54,43 @@ func (cp *configProviderImpl) GetVariables(cloud, deployEnv string) (Variables,
...
+		"azureEventGridName": naming.AzureEventGridName,
+		"azurePostgresName":  naming.AzurePostgresName,
+		"azureKeyVaultName":  naming.AzureKeyVaultName,
+	}
```
whober0521, line 72: FuncMap should be in main.go instead of config.go. Since these functions are only used in local development, not in pipeline

whober0521 (changes requested):

janboll (approved):

tooling/templatize/config/config.go
janboll: I'm confused, these methods need to be defined here, since this is where the config is parsed. So why would you want to configure them in main.go? :smile:

geoberle: can you elaborate why these would be only used in local dev?

geoberle: i'm going forward with this PR for now to unblock some work (the stamp part is the major change here) but let's discuss and we can move this as you suggested once we clarified

PR #749, opened 2024-10-21
Implement operations status polling in backend pod

mbarnes:
### What this PR does

This adds functionality to the backend pod which updates Cosmos DB items with the status of active (non-terminal) asynchronous operations for clusters.  Status polling for node pools is not currently possible; this part will be completed when the new **/api/aro_hcp/v1** OCM endpoint becomes available.

This runs on two independent polling intervals, each of which can be overridden through environment variables.

The first polling, which by default runs every 30s, scans the Cosmos DB "Operations" container for items with a non-terminal status and stores an internal list.

The second polling, which by default runs every 10s, iterates over that list and queries Cluster Service for the status of each resource.  It then translates the Cluster Service status to a ProvisioningState value, takes an error message for failed operations, and -- with the subscription locked -- updates the appropriate "Operations" item and "Resources" item in Cosmos DB.

Additionally, if a deletion operation has completed successfully, it deletes the corresponding "Resources" item from Cosmos DB.

Jira: [ARO-8598 - Prototype async op status notification mechanism for RP](https://issues.redhat.com/browse/ARO-8598)
Link to demo recording: 

For the record, I'm not a fan of this design and consider it a "Mark I" iteration.  Polling introduces a lot of latency in the statuses reported by the `Microsoft.RedHatOpenShift/hcpOpenShiftClusters` API.  But we're stuck with polling for now for two reasons:

1. Microsoft's own [Go SDK for Cosmos](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos) is minimally functional and lacks _any_ support for change feeds.  So notification of changes to Cosmos containers through the SDK is currently not possible, meaning the backend cannot respond immediately to new operations. (There's other ways to address this but the team agreed to initially just poll for simplicity's sake.)
2. Cluster Service today lacks any kind of publish-subscribe mechanism and is, by design, unaware of the ARO-HCP RP.  This is something that could more easily be addressed if the [recently posted ADR](https://docs.google.com/document/d/1vb-e52rTQIULz4HUkCKJAaMO3eEXkkvp3keKIj59_nE/edit#heading=h.fbts9k7ez097) is approved.  For example, a dedicated Cluster Service backend for ARO-HCP could update Cosmos DB directly and eliminate the need for polling altogether.


internal/database/cache.go
```diff
@@ -22,6 +24,34 @@ type Cache struct {
...
+			if !yield(item) {
+				return
+			}
+		}
+	}
+}
+
+func (iter operationCacheIterator) GetError() error {
+	return iter.err
+}
```
SudoBrendan, line 53: this doesn't look concurrency safe - not sure if that's a requirement. Should we at least log `err` as we set it? How would we know what call to Items failed?

SudoBrendan (commented):
At a glance, this appears to perform what we need. This feels like code that should get some tests - can we add them to solidify our requirements as code? e.g. "a cluster eventually deletes", "a cluster eventually creates", "no more than 1 call is made to `Items` since it is not concurrency safe", etc

internal/database/cache.go
mbarnes: It's not meant to be concurrency safe, it's meant to be used as part of a [range-over-function](https://go.dev/blog/range-functions) pattern, which is new in Go 1.23.  (Think generator functions in Python, if you're familiar with that.)  The pattern as described in the blog post doesn't address the possibility of iteration failing (such as when paging over incoming data from a remote source) so I came up with this iterator interface as a way to stash an iteration error.

The idiom looks like:
```
for item := range iterator.Items(ctx) {
    ...
}

// Find out if iteration completed or aborted early.
err := iterator.GetError()
```

An iteration error would immediately break the "for" loop.

mbarnes:
> This feels like code that should get some tests because it is foundational - can we add them to solidify our requirements as code?

Are you referring to the new database functions or to the backend code paths... or both?  The backend might be difficult to write unit tests for at the moment since parts of it rely on database locking (see [#680](https://github.com/Azure/ARO-HCP/pull/680)), which I don't have a way to mock at the moment.  (Mocking should be doable, it's just not done.)

internal/database/cache.go
mbarnes: Also maybe worth mentioning, since I didn't realize you were highlighting the in-memory cache...

I don't believe this particular function is currently used.  I wrote it for the in-memory cache in order to fulfill the `DBClient` interface, but the in-memory cache is nowadays only used to mock database operations in unit tests.

backend/operations_scanner.go
SudoBrendan, line 1: I think some tests for this file would be pretty helpful over time

SudoBrendan (commented):
sorry I still haven't had time for a complete review here - a few more comments... I will take time early next week for something comprehensive - thank you!

backend/operations_scanner.go
mbarnes: Agree.  Working on such tests this week.

SudoBrendan (approved):

internal/database/document.go
```diff
@@ -132,6 +132,20 @@ func (doc *OperationDocument) ToStatus() *arm.Operation {
 	return operation
 }
 
+// UpdateStatus conditionally updates the document if the status given differs
+// from the status already present. If so, it sets the Status and Error fields
+// to the values given, updates the LastTransitionTime, and returns true. This
+// is intended to be used with DBClient.UpdateOperationDoc.
+func (doc *OperationDocument) UpdateStatus(status arm.ProvisioningState, err *arm.CloudErrorBody) bool {
+	if doc.Status != status {
```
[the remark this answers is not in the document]
mbarnes: I can't say for certain one way or another, but currently the only time we include error information in the operation status is when the provisioning state is "Failed" and I wouldn't expect the error code/message returned from Cluster Service to continue to change once a failure state is reached.

PR #751, opened 2024-10-22
Bump istioVersion param to 1-22

shitaljante:
### What this PR does

Bump istioVersion version to 1-22 to resolve the error while deploying svc-cluster -
`Message: Requested change in revisions is not allowed. Reason: Revision asm-1-21 is not supported by the service mesh add-on..`

Verified the changes in local setup.

### DIFF
--- a/cluster-service/Makefile
+++ b/cluster-service/Makefile
@@ -12,8 +12,9 @@ AZURE_FIRST_PARTY_APPLICATION_CLIENT_ID ?= "57e54810-3138-4f38-bd3b-29cb33f4c358
 deploy:
 	ZONE_RESOURCE_ID=$(shell az network dns zone list -g ${REGIONAL_RESOURCEGROUP} --query "[?zoneType=='Public'].id" -o tsv) && \
 	sed -e "s#ZONE_RESOURCE_ID#$${ZONE_RESOURCE_ID}#g" -e "s/REGION/${REGION}/g" -e "s/CONSUMER_NAME/${CONSUMER_NAME}/g" deploy/mvp-provisioning-shards.yml > deploy/tmp-provisioning-shard.yml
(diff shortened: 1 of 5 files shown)


cluster-service/Makefile
```diff
@@ -13,7 +13,7 @@ deploy:
 	ZONE_RESOURCE_ID=$(shell az network dns zone list -g ${REGIONAL_RESOURCEGROUP} --query "[?zoneType=='Public'].id" -o tsv) && \
 	sed -e "s#ZONE_RESOURCE_ID#$${ZONE_RESOURCE_ID}#g" -e "s/REGION/${REGION}/g" -e "s/CONSUMER_NAME/${CONSUMER_NAME}/g" deploy/mvp-provisioning-shards.yml > deploy/tmp-provisioning-shard.yml
 	oc process --local -f deploy/openshift-templates/arohcp-namespace-template.yml \
-	  -p ISTIO_VERSION=asm-1-21 | oc apply -f -
+	  -p ISTIO_VERSION=asm-1-22 | oc apply -f -
```
ahitacat, line 16: I will consider to use az to get this version, like
https://github.com/Azure/ARO-HCP/blob/5117c63b7ee7abb67c050b73ed51592d355d2630/maestro/Makefile#L23

shitaljante: output of below is - 
% az aks list --query "[?tags.clusterType == 'svc-cluster' && starts_with(resourceGroup, 'cs-integ-westus3-sc')].serviceMeshProfile.istio.revisions[-1]" -o tsv
asm-1-21

should I keep value as it is for "cs-integ-westus3-sc" RG? because this will create discrepancy in param values.

ahitacat: Not really, so when called this, it will use your RESOURCE_GROUP which if I check this I get the asm-1.22. 
I think that is the discrepancy. 
Can you check to only change this makefile?

I think the bicep scripts get modified in makefile call.

shitaljante: I have updated makefile , PTAL

ahitacat (approved):
LGTM, @geoberle wdyt?

geoberle (approved):

geoberle:
LGTM

PR #753, opened 2024-10-22
chore: remove install-config-file flag from the aro hcp service template

JameelB:
### What this PR does
The install-config-file is an optional flag that is only used within the Hive provisioner. This is not required for running CS on ARO HCP mode, therefore we should remove it from the template.

This removes the volume mount path '/configs/service' which only contains this install config yaml file and is not used by anything else. It can safely be removed.

Jira: [ARO-11242](https://issues.redhat.com/browse/ARO-11242)
Link to demo recording: N/A

N/A


ahitacat (approved):
LGTM :seal:

bennerv (approved):

PR #759, opened 2024-10-23
Block requests based on provisioning state

mbarnes:
### What this PR does

This pull request builds upon [#749 Implement operations status polling in backend pod](https://github.com/Azure/ARO-HCP/pull/749), so its commits are duplicated here until that pull request is merged. (The last two commits are unique to this PR.)

**Update:** Instead of reproducing all the commits from #749 here I just cherry-picked the one this PR actually needs.

This should complete the ARM contract for asynchronous operations by blocking requests on a resource based on its provisioning state and – for child resources – its parent resource's provisioning state.  The semantics for this are not covered in the [Resource Provider Contract on GitHub](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/) ([though they should be](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/issues/192)), so it has been derived from a number of other sources: [RPaaS behavior](https://eng.ms/docs/products/arm/rpaas/async), [ARO Classic behavior](https://github.com/Azure/ARO-RP/tree/master/pkg/frontend), and the [Resource Provider Contract Guidelines on eng.ms](https://eng.ms/docs/products/arm/api_contracts/guidelines/rpc).

Note: Until the new **/api/aro_hcp/v1** OCM endpoint becomes available, asynchronous operation statuses for node pools will not be accurate.  But we can still fulfill the high-level ARM requirements.

Jira: I don't think we have a JIRA Story specifically for this, but it falls under [XCMSTRAT-627 - ARO HCP - Async Operations](https://issues.redhat.com/browse/XCMSTRAT-627)
Link to demo recording:

### DIFF
--- a/frontend/pkg/frontend/helpers.go
+++ b/frontend/pkg/frontend/helpers.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	ocmerrors "github.com/openshift-online/ocm-sdk-go/errors"
@@ -16,6 +17,57 @@ import (
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/database"
 )
+
+// CheckForProvisioningStateConflict returns a "409 Conflict" error response if the
+// provisioning state of the resource is non-terminal, or any of its parent resources
+// within the same provider namespace are in a "Deleting" state.
+func (f *Frontend) CheckForProvisioningStateConflict(ctx context.Context, operationRequest database.OperationRequest, doc *database.ResourceDocument) *arm.CloudError {
+	switch operationRequest {
+	case database.OperationRequestCreate:
+		// Resource must already exist for there to be a conflict.
+	case database.OperationRequestDelete:
+		if doc.ProvisioningState == arm.ProvisioningStateDeleting {
+			return arm.NewCloudError(
+				http.StatusConflict,
+				arm.CloudErrorCodeConflict,
+				doc.Key.String(),
+				"Resource is already deleting")
+		}
+	case database.OperationRequestUpdate:
+		if !doc.ProvisioningState.IsTerminal() {
+			return arm.NewCloudError(
+				http.StatusConflict,
+				arm.CloudErrorCodeConflict,
+				doc.Key.String(),
+				"Cannot update resource while resource is %s",
+				strings.ToLower(string(doc.ProvisioningState)))
+		}
+	}
+
+	parent := doc.Key.GetParent()
+
+	// ResourceType casing is preserved for parents in the same namespace.
+	for parent.ResourceType.Namespace == doc.Key.ResourceType.Namespace {
+		parentDoc, err := f.dbClient.GetResourceDoc(ctx, parent)
+		if err != nil {
+			f.logger.Error(err.Error())
+			return arm.NewInternalServerError()
+		}
+
+		if parentDoc.ProvisioningState == arm.ProvisioningStateDeleting {
+			return arm.NewCloudError(
+				http.StatusConflict,
+				arm.CloudErrorCodeConflict,
+				doc.Key.String(),
+				"Cannot %s resource while parent resource is deleting",
+				strings.ToLower(string(operationRequest)))
+		}
+
+		parent = parent.GetParent()
+	}
+
+	return nil
+}
 
 func (f *Frontend) MarshalResource(ctx context.Context, resourceID *arm.ResourceID, versionedInterface api.Version) ([]byte, *arm.CloudError) {
 	var responseBody []byte
--- a/frontend/pkg/frontend/node_pool.go
+++ b/frontend/pkg/frontend/node_pool.go
@@ -42,7 +42,7 @@ func (f *Frontend) CreateOrUpdateNodePool(writer http.ResponseWriter, request *h
 		return
 	}
 
-	nodePoolResourceID, err := ResourceIDFromContext(ctx)
+	resourceID, err := ResourceIDFromContext(ctx)
 	if err != nil {
 		f.logger.Error(err.Error())
 		arm.WriteInternalServerError(writer)
@@ -58,41 +58,14 @@ func (f *Frontend) CreateOrUpdateNodePool(writer http.ResponseWriter, request *h
 
 	f.logger.Info(fmt.Sprintf("%s: CreateNodePool", versionedInterface))
 
-	clusterResourceID := nodePoolResourceID.GetParent()
-	if clusterResourceID == nil {
-		f.logger.Error(fmt.Sprintf("failed to obtain Azure parent resourceID for node pool %s", nodePoolResourceID))
-		arm.WriteInternalServerError(writer)
-		return
-	}
-
-	clusterDoc, err := f.dbClient.GetResourceDoc(ctx, clusterResourceID)
-	if err != nil {
-		f.logger.Error(err.Error())
-		arm.WriteInternalServerError(writer)
-		return
-	}
-
-	csCluster, err := f.clusterServiceClient.GetCSCluster(ctx, clusterDoc.InternalID)
-	if err != nil {
-		f.logger.Error(fmt.Sprintf("failed to fetch CS cluster for %s: %v", clusterResourceID, err))
-		arm.WriteInternalServerError(writer)
-		return
-	}
-
-	if csCluster.State() == cmv1.ClusterStateUninstalling {
-		f.logger.Error(fmt.Sprintf("failed to create node pool for cluster %s as it is in %v state", clusterResourceID, cmv1.ClusterStateUninstalling))
-		arm.WriteInternalServerError(writer)
-		return
-	}
-
-	nodePoolDoc, err := f.dbClient.GetResourceDoc(ctx, nodePoolResourceID)
+	doc, err := f.dbClient.GetResourceDoc(ctx, resourceID)
 	if err != nil && !errors.Is(err, database.ErrNotFound) {
 		f.logger.Error(err.Error())
 		arm.WriteInternalServerError(writer)
 		return
 	}
 
-	var updating = (nodePoolDoc != nil)
+	var updating = (doc != nil)
 	var operationRequest database.OperationRequest
 
 	var versionedCurrentNodePool api.VersionedHCPOpenShiftClusterNodePool
@@ -106,14 +79,14 @@ func (f *Frontend) CreateOrUpdateNodePool(writer http.ResponseWriter, request *h
 		// No special treatment here for "not found" errors. A "not found"
 		// error indicates the database has gotten out of sync and so it's
 		// appropriate to fail.
-		csNodePool, err := f.clusterServiceClient.GetCSNodePool(ctx, nodePoolDoc.InternalID)
+		csNodePool, err := f.clusterServiceClient.GetCSNodePool(ctx, doc.InternalID)
 		if err != nil {
-			f.logger.Error(fmt.Sprintf("failed to fetch CS node pool for %s: %v", nodePoolResourceID, err))
+			f.logger.Error(fmt.Sprintf("failed to fetch CS node pool for %s: %v", resourceID, err))
 			arm.WriteInternalServerError(writer)
 			return
 		}
 
-		hcpNodePool := ConvertCStoNodePool(nodePoolResourceID, csNodePool)
+		hcpNodePool := ConvertCStoNodePool(resourceID, csNodePool)
 
 		// Do not set the TrackedResource.Tags field here. We need
 		// the Tags map to remain nil so we can see if the request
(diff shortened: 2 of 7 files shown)


backend/main.go
```diff
@@ -64,20 +72,73 @@ func init() {
...
+	databaseClient, err := client.NewDatabase(argCosmosName)
+	if err != nil {
+		return nil, err
+	}
+
+	return database.NewCosmosDBClient(ctx, databaseClient)
+}
+
 func Run(cmd *cobra.Command, args []string) error {
+	ctx := context.Background()
```
stevekuznetsov, line 106: Might as well add `signal.NotifyContext` on `os.Interrupt` while you're here.

backend/main.go
```diff
@@ -64,20 +72,73 @@ func init() {
...
+		return fmt.Errorf("Failed to create OCM connection: %w", err)
+	}
+
 	logger.Info(fmt.Sprintf("%s (%s) started", cmd.Short, cmd.Version))
 
+	operationsScanner := NewOperationsScanner(dbClient, ocmConnection)
+
 	stop := make(chan struct{})
 	signalChannel := make(chan os.Signal, 1)
 	signal.Notify(signalChannel, syscall.SIGINT, syscall.SIGTERM)
```
stevekuznetsov, line 132: Actually - best to delete these lines and use the `ctx` you created before, decorate it with `signal.NotifyContext`

backend/operations_scanner.go
```diff
@@ -0,0 +1,346 @@
...
+		if err == nil {
+			return interval
+		} else {
+			logger.Warn(fmt.Sprintf("Cannot use %s: %s", envName, err.Error()))
+		}
+	}
+	return defaultVal
+}
+
+func (s *OperationsScanner) Run(ctx context.Context, logger *slog.Logger, stop <-chan struct{}) {
```
stevekuznetsov, line 60: What does it mean to take a context _and_ a stop channel? It looks like you could use `ctx.Done()` instead of needing `stop`.

backend/main.go
```diff
@@ -64,20 +72,73 @@ func init() {
...
 
+	go operationsScanner.Run(ctx, logger, stop)
+
 	sig := <-signalChannel
 	logger.Info(fmt.Sprintf("caught %s signal", sig))
 	close(stop)
 
+	operationsScanner.Join()
```
stevekuznetsov, line 140: What's the benefit of making `Run()` asynchronous and having a `Join()` as opposed to just making `Run()` synchronous and waiting for it to return?

backend/operations_scanner.go
```diff
@@ -0,0 +1,346 @@
...
+		}
+	}
+}
+
+func (s *OperationsScanner) Join() {
+	<-s.done
+}
+
+func (s *OperationsScanner) pollDBOperations(ctx context.Context, logger *slog.Logger) {
+	activeOperations := make([]*database.OperationDocument, 0)
```
stevekuznetsov, line 93: Not sure if there is a specific reason for this declaration, but the idiomatic way is `var activeOperations []*database.OperationDocument` - it more clearly indicates that a zero-length slice is being allocated. The `make` path is correct but adds review load to the PR as reviewers need to read carefully to see what the intent is.

backend/operations_scanner.go
```diff
@@ -0,0 +1,346 @@
...
+		s.activeOperations = activeOperations
+		if len(s.activeOperations) > 0 {
+			logger.Info(fmt.Sprintf("Tracking %d active operations", len(s.activeOperations)))
+		}
+	} else {
+		logger.Error(fmt.Sprintf("Error while paging through Cosmos query results: %s", err.Error()))
+	}
+}
+
+func (s *OperationsScanner) pollCSOperations(ctx context.Context, logger *slog.Logger, stop <-chan struct{}) {
```
stevekuznetsov, line 122: Similar comments here, I don't think you need a context _and_ a channel.

backend/operations_scanner.go
```diff
@@ -0,0 +1,346 @@
...
+	case cmv1.ClusterStateUninstalling:
+		opStatus = arm.ProvisioningStateDeleting
+	case cmv1.ClusterStatePending, cmv1.ClusterStateValidating:
+		// These are valid cluster states for ARO-HCP but there are
+		// no unique ProvisioningState values for them. They should
+		// only occur when ProvisioningState is Accepted.
+		if opStatus != arm.ProvisioningStateAccepted {
+			logger.Warn(fmt.Sprintf("Got ClusterState '%s' while ProvisioningState was '%s' instead of '%s'", clusterStatus.State(), opStatus, arm.ProvisioningStateAccepted))
+		}
+		return
```
stevekuznetsov, line 205: Naked `return`s with implicit variable names are extremely difficult to reason about and in my opinion should be avoided as a rule.

backend/operations_scanner.go
```diff
@@ -0,0 +1,346 @@
...
+	default:
+		logger.Warn(fmt.Sprintf("Unhandled ClusterState '%s' for operation '%s'", clusterStatus.State(), doc.ID))
```
stevekuznetsov, line 207: You're using `slog` for logging, so please add structured data to the logger appropriately:

```go
logger.With("clusterState", clusterStatus.State(), "operation", doc.ID").Warn("unhandled cluster state")
```

stevekuznetsov: (global comment for the whole PR)

backend/operations_scanner.go
stevekuznetsov: (global comment for the whole PR)

backend/operations_scanner.go
stevekuznetsov: (global comment for the whole PR)

backend/operations_scanner.go
stevekuznetsov: Also, note that while the default should certainly be to add structured data to a log line, if there's something you _need_ string-formatting for, `log/slog` already will do string formatting without needing to use `fmt.Sprintf`

backend/operations_scanner.go
```diff
@@ -0,0 +1,346 @@
...
+	err := s.dbClient.DeleteResourceDoc(ctx, doc.ExternalID)
+	if err != nil {
+		return err
+	}
+
+	// Save a final "succeeded" operation status until TTL expires.
+	const opStatus arm.ProvisioningState = arm.ProvisioningStateSucceeded
+	updated, err := s.dbClient.UpdateOperationDoc(ctx, doc.ID, func(updateDoc *database.OperationDocument) bool {
+		if opStatus != updateDoc.Status {
+			updateDoc.LastTransitionTime = time.Now()
```
stevekuznetsov, line 257: You likely want to store `now func() time.Time` as a member on `OperationsScanner` to allow any of this to be unit-tested while injecting a fake clock.

stevekuznetsov: (global comment for the whole PR)

backend/main.go
mbarnes: There will eventually be at least one other "scanner" routine running in parallel here so they need to remain asynchronous.

backend/operations_scanner.go
mbarnes: The intent here was to clarify what the boolean return value was for, but I take your point about named return variables; I can rewrite it to be more clear.

backend/operations_scanner.go
mbarnes: Structured data is added in the calling function using `logger.With`.  That's why I'm passing the `slog.Logger` instance as a function argument instead of making it a struct member.

> `log/slog` already will do string formatting without needing to use `fmt.Sprintf`

No, it won't.  Arguments past the first string argument are key/value pairs.

backend/operations_scanner.go
mbarnes: I wasn't aware of `signal.NotifyContext` and I can see the use cases for it, but I think I'm going to decline using it here.  The intent of the `stop` channel is to stop the timers upon receiving a terminate signal but let the last scan run to completion.  Cancelling the context on a terminate signal would abort the scanner immediately and potentially leave Cosmos data across multiple containers in an inconsistent state (though I realize this is still possible if the pod is forcefully killed).

backend/operations_scanner.go
stevekuznetsov: @mbarnes I think my comment was that `doc.ID` seems like exactly the sort of data you'd want to pull out into a structured logger attribute and filter on post-hoc when debugging.

backend/operations_scanner.go
stevekuznetsov: If you truly have two individual and distinct types of termination (why? is that really necessary? as you note, the application needs to be robust to this in any case, and if possible transactions etc need to be used), you should still use `context.Context` - that's why it exists. Passing through a `stop` channel is a much older and less powerful mechanism to achieve it. Mixing both approaches is pretty complicated.

backend/main.go
stevekuznetsov: If that's true, then this model as written in this PR will not work - one will block you from waiting on the other and you could deadlock. Prefer a `sync.WaitGroup` or `sync.ErrorGroup`

backend/operations_scanner.go
stevekuznetsov: (If it's already present in the logger, omit it from string-formatting in the log line, as it's redundant, right?)

backend/operations_scanner.go
mbarnes: As I understand, Cosmos transactions are limited to a single partition in a single container.  In our case we're juggling two containers with item state that needs to stay consistent.  (This may be a sign we need to rethink our database design, but that won't be addressed here.)

frontend/pkg/frontend/node_pool.go
```diff
@@ -42,7 +42,7 @@ func (f *Frontend) CreateOrUpdateNodePool(writer http.ResponseWriter, request *h
 		return
 	}
 
-	nodePoolResourceID, err := ResourceIDFromContext(ctx)
+	resourceID, err := ResourceIDFromContext(ctx)
```
SudoBrendan, line 46: nit: feel free to dismiss - but I think I personally prefer the old names (`nodePoolResourceID` and `nodePoolDoc`) because there are so many resources at play here (clusters, CSnodePools, HCPNodePools, etc)

frontend/pkg/frontend/helpers.go
```diff
@@ -17,6 +18,57 @@ import (
 	"github.com/Azure/ARO-HCP/internal/database"
 )
 
+// CheckForConflict returns a "409 Conflict" error response if the provisioning
+// state of the resource is non-terminal, or any of its parent resources within
+// the same provider namespace are in a "Deleting" state.
+func (f *Frontend) CheckForConflict(ctx context.Context, operationRequest database.OperationRequest, doc *database.ResourceDocument) *arm.CloudError {
+	switch operationRequest {
+	case database.OperationRequestCreate:
```
SudoBrendan, line 26: note to future reviewers: this is a [golang switch statement no-op case](https://go.dev/wiki/Switch#noop-case) not something like other languages where `Create` and `Delete` have the same path because there is no body defined; Figured I'd save some searches

frontend/pkg/frontend/helpers.go
```diff
@@ -17,6 +18,57 @@ import (
 	"github.com/Azure/ARO-HCP/internal/database"
 )
 
+// CheckForConflict returns a "409 Conflict" error response if the provisioning
+// state of the resource is non-terminal, or any of its parent resources within
+// the same provider namespace are in a "Deleting" state.
+func (f *Frontend) CheckForConflict(ctx context.Context, operationRequest database.OperationRequest, doc *database.ResourceDocument) *arm.CloudError {
```
SudoBrendan, line 24: idea: would this be clearer named as `CheckForProvisioningStateOperationConflict`? It is a longer name, but it might save us from having to write a clarifying comment on each call?

SudoBrendan (approved):
I think that ultimately this looks good to merge as long as the other prereq stuff does too. Thanks for always organizing commits so neatly!

Adding some unit tests for the helpers would really make this a standout and save some future maintenance I'm sure.

frontend/pkg/frontend/helpers.go
mbarnes: I'm not opposed to a longer name, but the clarifying comment on each call is there to address error handling since there's a conspicuous (but intentional) absence of a log entry call in the error block.

```
	// CheckForConflict does not log conflict errors but
	// does log unexpected errors like database failures.
	cloudError := f.CheckForConflict(ctx, operationRequest, doc)
	if cloudError != nil {
		arm.WriteCloudError(writer, cloudError)  <-- usually preceded by f.logger.Error()
		return
	}
```

frontend/pkg/frontend/node_pool.go
mbarnes: Yeah I take your point.  For months I've been trying to inch toward unifying the cluster and node pool request handlers but so far I've only managed to do so for GET requests.  I tend to make the logic (and variable names) as uniform as possible between the two but maybe that's at the expense of clarity.

Generally in this code base -- and this is just a vocabulary I've come up with -- "resource IDs" refer to Azure resources and "internal IDs" refer to Cluster Service resources.  The [InternalID](https://github.com/Azure/ARO-HCP/blob/main/internal/ocm/internalid.go) type is just the OCM API path (`/api/clusters_mgmt/v1/cluster/...`) with some convenience methods around it.

frontend/pkg/frontend/helpers.go
mbarnes: Yes.  In Go, case fallthroughs are explicit.  Implicit fallthroughs are a source of many bugs in C/C++ code.

frontend/pkg/frontend/helpers.go
mbarnes: I renamed it `CheckForProvisioningStateConflict`.

SudoBrendan (approved):
thanks for the tests!

PR #779, opened 2024-10-28
Add top-level github action to install specific azure-cli version

bennerv:
### What this PR does
Creates a top-level GitHub action in ARO-HCP for the following reasons:
1. azure/cli github action runs in a container which doesn't have a `make` binary present.  
2. allows us to pin the `azure-cli` version to a specific version, in the case where it's broken (https://github.com/Azure/azure-cli/issues/29828)

### Usage
We can now put this at the top of every github action to leverage this version (or define your own) of the azure-cli
```
- name: "install azure-cli'
  uses: "Azure/ARO-HCP@main"
```

### Tests: 
https://github.com/bennerv/ARO-HCP/actions/runs/11184797337/job/31096334705

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,0 +1,42 @@
+---
+# Custom install of the azure-cli
+# Note that the azure-cli ships out of the box with ubuntu-latest for
+# github actions runs.  This allows us to pin a version and also allows us
+# to use commands that aren't present in the azure/cli github action that's
+# published to the GitHub actions marketplace.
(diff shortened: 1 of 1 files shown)


bennerv:
I didn't add any usage of this because it's a chicken and egg problem.  We need the github action available in order to leverage it.  So this needs to merge first, then we can add the `uses` clause wherever needed in place of `azure/cli`.

SudoBrendan (approved):

PR #782, opened 2024-10-29
Enable usage of workload identity and azure DB in CS (including P1 environment)

miguelsorianod:
### What this PR does

This PR enables the usage of Azure Workload Identity for the CS deployment and also enables the usage of Azure Database for PostgreSQL using Microsoft Entra authentication.

It also changes the P1 environment to use both functionalities.

The changes specifically:
Update the CS image which enables the use
of Azure Workload Identity for the CS deployment itself.
    
We also create an additional Makefile target for CS that
allows deploying CS pointing to an Azure Database for PostgreSQL
Flexible Server. This approach is a workaround/temporary solution
to allow us using this in the P1 environment until we move to
Helm charts where it will be designed in a good way.
    
The changes also refactor the Makefile to avoid duplicating part of
the code.

Finally, the P1 configuration and pipelines are changed to use the new functionalities in the P1 environment.

Jira: 
Link to demo recording:

### DIFF
--- a/cluster-service/Makefile
+++ b/cluster-service/Makefile
@@ -41,12 +41,7 @@ deploy: configure-tmp-provision-shard deploy-namespace-template deploy-istio-con
 # be redesigned when we move to use Helm templates.
 # *Warning* if you change any parameter remember to change it in both the deploy and the deploy-using-azure-db targets
 # if they apply to both
-deploy-using-azure-db: configure-tmp-provision-shard deploy-namespace-template deploy-istio-configurations-template deploy-secrets-template
-	oc process --local -f deploy/openshift-templates/arohcp-db-template.yml \
-	-p DATABASE_USER=clusters-service \
-	-p DATABASE_NAME=clusters-service \
-	-p DATABASE_PASSWORD="" \
-	-p DATABASE_SERVICE=$(shell az postgres flexible-server list --resource-group ${RESOURCEGROUP} --query "[?starts_with(name, 'cs-pg-')].fullyQualifiedDomainName" -o tsv) | oc apply -f -
+deploy-using-azure-db: configure-tmp-provision-shard deploy-namespace-template deploy-istio-configurations-template deploy-secrets-template-using-azure-db
 	AZURE_CS_MI_CLIENT_ID=$(shell az identity show \
 			-g ${RESOURCEGROUP} \
 			-n clusters-service \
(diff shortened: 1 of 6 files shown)


miguelsorianod:
Do not merge this yet as I am doing some manual testing to verify the changes.

miguelsorianod:
> As an important note, the changes in here even if we use Azure Database for PostgreSQL in CS the postgreSQL K8s deployment and its associated K8s service will still be deployed, although unused. We are aware of this and we intend to stop doing that as soon as possible. There's work being done to move to Helm charts.

I actually went and I've updated the Makefile in the MR to only deploy the postgreSQL pod when the deploy target is called. when The deploy-using-azure-db target is called no postgreSQL pod is deployed anymore

cluster-service/Makefile
```diff
@@ -29,28 +24,82 @@ deploy:
...
+	ISTO_VERSION=$(shell az aks list --query "[?tags.clusterType == 'svc-cluster' && starts_with(resourceGroup, '${RESOURCEGROUP}')].serviceMeshProfile.istio.revisions[-1]" -o tsv)  && \
+	oc process --local -f deploy/openshift-templates/arohcp-namespace-template.yml \
+		-p ISTIO_VERSION=$${ISTO_VERSION} | oc apply -f -
+
+deploy-istio-configurations-template:
+	kubectl apply -f deploy/istio.yml
+
+deploy-secrets-template: configure-tmp-provision-shard
+	oc process --local -f deploy/openshift-templates/arohcp-secrets-template.yml \
+		-p PROVISION_SHARDS_CONFIG="$$( base64 -i deploy/tmp-provisioning-shard.yml)" | oc apply -f -
```
JameelB, line 86: Does this need to configure a db password? Otherwise the password will be `Password for the PostgreSQL connection user` 🤔  Alternatively, rather than having this as a default value, it can just be an empty string? And this make target then can set the db password?

dev-infrastructure/configurations/mvp-svc-cluster.bicepparam
```diff
@@ -20,7 +20,7 @@ param maestroPostgresServerStorageSizeGB = 32
 param deployMaestroPostgres = false
 param maestroPostgresPrivate = false
 
-param deployCsInfra = false
+param deployCsInfra = true
```
JameelB, line 23: what's the reason for changing this?

miguelsorianod:
I tested the makefile using the two targets:

deploy
deploy-using-azure-db

And I've been able to create ARO-HCP clusters in both scenarios. I also verified that the appropriate DB is used in each case.

dev-infrastructure/configurations/mvp-svc-cluster.bicepparam
miguelsorianod: This enables the deployment of an Azure Database for PostgreSQL instance in the P1 environment, which is the motivation of all these changes. We want CS to use that service instead of a postgresql pod. Also when using that service we want to authenticate against the DB using Microsoft Entra tokens

miguelsorianod:
This should be ready for review and merge if it looks good

cluster-service/Makefile
JameelB: thanks, saw the update, lgtm

JameelB:
lgtm!

geoberle (approved):
LGTM.
i'm fine with the make target duplication. we will dedup again in https://github.com/Azure/ARO-HCP/pull/762

PR #791, opened 2024-10-31
add cs platform identities config map

tony-schndr:
### What this PR does

Adds a config map containing platform identity role mappings required for CS.  Mapping is derived from production [RP-Config](https://dev.azure.com/msazure/AzureRedHatOpenShift/_git/RP-Config?path=/deploy/prod-config.yaml&version=GBmaster&line=2946&lineEnd=2947&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents). 

Jira: https://issues.redhat.com/browse/ARO-9598
Link to demo recording: 

The config map currently contains role identities from public prod environment.  This will be parameterized per environment once templatize refactor is complete and cluster service is deployable via helm to reduce rework.

### DIFF
--- a/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
+++ b/cluster-service/deploy/openshift-templates/arohcp-service-template.yml
@@ -368,6 +368,107 @@ objects:
         }
       }
 
+# TODO: parameterize data.platform-identities.yaml after templatize refactor and helm chart conversion
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: azure-operators-managed-identities-config
+    namespace: ${NAMESPACE}
+  data:
+    azure-operators-managed-identities-config.yaml: |
+      controlPlaneOperatorsIdentities:
+
+        cloud-controller-manager:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/ebe170ec-1247-536a-86d9-74c829dd9844'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Cloud Controller Manager - Dev'
+          optional: false
+
+        ingress:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/589ca160-4fac-501e-ad6c-006a19583727'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Cluster Ingress Operator - Dev'
+          optional: false
+
+        disk-csi-driver:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/4367fe74-0b43-5033-b629-15d9f28415ac'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Disk Storage Operator - Dev'
+          optional: false
+
+        file-csi-driver:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/fdc0aaaa-1c3e-548e-ad27-0321e5fab18b'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift File Storage Operator - Dev'
+          optional: false
+
+        image-registry:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/357b9263-656f-5d45-9d7a-ccb825f0683f'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Image Registry Operator - Dev'
+          optional: false
+
+        cloud-network-config:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/4e4f23fe-3fab-568b-a001-10b233b0f840'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Network Operator - Dev'
+          optional: false
+
+      dataPlaneOperatorsIdentities:
+
+        disk-csi-driver:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/4367fe74-0b43-5033-b629-15d9f28415ac'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Disk Storage Operator - Dev'
+          k8sServiceAccounts:
+            - name:  'azure-disk-csi-driver-operator'
+              namespace: 'openshift-cluster-csi-drivers'
+            - name: 'azure-disk-csi-driver-controller-sa'
+              namespace: 'openshift-cluster-csi-drivers'
+          optional: false
+
+        image-registry:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/357b9263-656f-5d45-9d7a-ccb825f0683f'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Image Registry Operator - Dev'
+          k8sServiceAccounts:
+            - name: 'cluster-image-registry-operator'
+              namespace: 'openshift-image-registry'
+            - name: 'registry'
+              namespace:  'openshift-image-registry'
+          optional: false
+
+        file-csi-driver:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/fdc0aaaa-1c3e-548e-ad27-0321e5fab18b'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift File Storage Operator - Dev'
+          k8sServiceAccounts:
+            - name: ':azure-file-csi-driver-operator'
+              namespace:  'openshift-cluster-csi-drivers'
+            - name: 'azure-file-csi-driver-controller-sa'
+              namespace:  'openshift-cluster-csi-drivers'
+            - name: 'azure-file-csi-driver-node-sa'
+              namespace:  'openshift-cluster-csi-drivers'
+          optional: false
+
+        ingress:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/589ca160-4fac-501e-ad6c-006a19583727'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Cluster Ingress Operator - Dev'
+          k8sServiceAccounts:
+            - name: 'ingress-operator'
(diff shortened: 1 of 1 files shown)


bennerv:
Can use [this](https://docs.google.com/spreadsheets/d/1D2o6Wi54yq75SLfIkP5wIMRamwre7pVDO9eJpCXRhXI/edit?gid=0#gid=0) to say which ones are HCP control plane vs. public cloud vs. not used at all.

Note that we don't have role definitions yet for some of them (CAPZ, CPO, KMS, etc.).  I'd just get it up to date with the info we know and we can come back and fix it.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -354,6 +354,91 @@ objects:
         "cloudEnvironment": "AzurePublicCloud"
       }
 
+# TODO: parameterize data.platform-identities.yaml after templatize refactor and helm chart conversion
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: clusters-service-platform-identity-roleset
```
machi1990, line 361: Can we align the naming to the one used by CS;
We are currently using `azure-operators-managed-identities-config`

machi1990: The reasoning for this suggestion is that it'll be easier for devs (notably CS ones) to work with this configMap if it contains the same naming.

machi1990:
> Can use [this](https://docs.google.com/spreadsheets/d/1D2o6Wi54yq75SLfIkP5wIMRamwre7pVDO9eJpCXRhXI/edit?gid=0#gid=0) to say which ones are HCP control plane vs. public cloud vs. not used at all.
> 
> Note that we don't have role definitions yet for some of them (CAPZ, CPO, KMS, etc.). I'd just get it up to date with the info we know and we can come back and fix it.

@bennerv @tony-schndr +1

machi1990:
@tony-schndr once https://github.com/Azure/ARO-HCP/pull/807 is in, perhaps we could mount the configMap and configure  CS using the CLI flag below
```
 - --azure-operators-managed-identities-config-path=<path to the mounted file.yaml>
```
what do you think?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -517,6 +643,9 @@ objects:
             readOnly: true
           - name: azure-runtime-config
             mountPath: /configs/azure-runtime-config
+          - name: azure-operators-managed-identities-config
+            mountPath: /configs/azure-operators-managed-identities-config.yaml
+            subPath: azure-operators-managed-identities-config.yaml
```
machi1990, line 648: From https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath, I see this NOTE
```
A container using a ConfigMap as a [subPath](https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath) volume mount will not receive ConfigMap updates.
```

This means that whenever this config map is changed, cs will need to be redeployed. This is fine for now  as CS doesn't autoreload the file when it changes, but mentioning the note for consideration in case that implementation changes in CS.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -368,6 +368,129 @@ objects:
...
+  metadata:
+    name: azure-operators-managed-identities-config
+    namespace: ${NAMESPACE}
+  data:
+    azure-operators-managed-identities-config.yaml: |
+      controlPlaneOperatorsIdentities:
+
+        cloud-controller-manager:
+          minOpenShiftVersion: 4.17
+          maxOpenShiftVersion: 4.17
```
machi1990, line 383: Why for all the operators min and max version is set to 4.17? Why not leave max as not set ? The implication of setting max to 4.17 is that when 4.18 is out for example, we'll need to update the version here otherwise we can't really use the config as it is with that version as it'll fail

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -368,6 +368,129 @@ objects:
...
+          azureRoleDefinitionResourceId: '/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/providers/Microsoft.Authorization/roleDefinitions/ebe170ec-1247-536a-86d9-74c829dd9844'
```
machi1990, line 384: I noticed that the roleDefinition is scoped to our subscription, I am wondering if this config as is can be  re-usable in say INT, Stage or Prod env? @tony-schndr @bennerv do you know or we would need to update the value across the different envs?

machi1990 (commented):
Thanks @tony-schndr 

I've left some comment

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
tony-schndr: INT and PROD will use different roles.  I am creating Custom roles in dev that are hardcoded in the repo for now.  I will be converting cluster-service deployment to use helm in the near future, apart of that change the config map will be able to use a different data.
I took this route for 2 reasons:
1. To unblock cluster-service work.
2. To prevent rework trying to make the config map data configurable in oc templates, since deployment of cluster-service is moving to helm charts.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
tony-schndr: I didn't realize I could leave that field unset.  I am concerned with leaving this unset though, future versions of OpenShift may add required permissions to the CredentialsRequests that were not in prior versions.

machi1990: Yes, the min/max openshift versions fields are optional. 
CS won't complain if you don't set them. 

> I am concerned with leaving this unset though, future versions of OpenShift may add required permissions to the CredentialsRequests that were not in prior versions.

That's a fair point. I think we'll notice it with a failing upgrade / installation (when trying out the new version before they reach out a prod env) to which we'll only need to update the permission. While with the max version being set, we'll need to first update the max version and then only to see that it was a missing permission issue and bump it.

I saw the max version useful for when we know that there is an operator that's only available up to that version. For future version, it has been removed, so for that operator we'll add the max version to guard it. 
The min version follows the same reasoning in my head.

tony-schndr: I removed the max openshift version fields.

machi1990 (approved):

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -368,6 +368,107 @@ objects:
...
+          optional: false
+
+      dataPlaneOperatorsIdentities:
+
+        disk-csi-driver:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/providers/Microsoft.Authorization/roleDefinitions/4367fe74-0b43-5033-b629-15d9f28415ac'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Disk Storage Operator - Dev'
+          k8sServiceAccounts:
+            - name:  'system:serviceaccount:openshift-cluster-csi-drivers:azure-disk-csi-driver-operator'
```
machi1990, line 424: I think this should just be the name of the service account. 
```suggestion
            - name:  'azure-disk-csi-driver-operator'
```

When CS is performing the federation, it'll construct the subject as expected, using the name and the namespace
```go
fmt.Sprintf("system:serviceaccount:%s:%s",namespace,name)
```

@bennerv can you confirm?

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -368,6 +368,107 @@ objects:
...
+
+        cloud-controller-manager:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/providers/Microsoft.Authorization/roleDefinitions/ebe170ec-1247-536a-86d9-74c829dd9844'
+          azureRoleDefinitionName: 'Azure Red Hat OpenShift Cloud Controller Manager - Dev'
+          optional: false
+
+        ingress:
+          minOpenShiftVersion: 4.17
+          azureRoleDefinitionResourceId: '/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/providers/Microsoft.Authorization/roleDefinitions/589ca160-4fac-501e-ad6c-006a19583727'
```
bennerv, line 389: We shouldn't include the `/subscription` information in here in my opinion, CS should append the information here because it's always relevant to the "remote" subscription.  

@machi1990 -> can we capture this work somewhere?  Whenever we create a resource group client we'll need to append the `/subscriptions/<customer-sub-id>` to the front of the roleDefinition.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
bennerv: These should always be hardcoded.  I'm actually wondering if CS needs namespace as a field in this config and instead we can just list the full `system:serviceaccount:<namespace>:<service-account-name>` as the config and call it good.  

CS configures these on the hosted cluster CR, so technically it doesn't need to know what namespaces correspond to what.  Dealers choice @machi1990 , but IMO i'd either remove namespace from this config and keep the "full" path, or just use the name of the service account.

machi1990: @bennerv @tony-schndr  We could've given the full `system:serviceaccount:<namespace>:<service-account-name>` and call it good but the way the config was designed is that `name` and `namespace` are to be given (ARO-7364) and CS would use use that info to construct the federation subject name among other. And config loading mechanism in CS has been implemented that way so I would suggest to keep it at it and not change it. 
Meaning that here we should give
```yaml
name: <service-account-name>
namespace: <service-account-namespace>
```
and not the full form for the name.

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
machi1990: This is a good point you are bringing because i was wondering what is this azure role definition id.
- is it a built-in role definition (akin to "Contributor" or "Owner" ) or a custom one?
- Who creates them? 
- Where are they created? On the customer's account

Is it that the customer will create them on the customer's subscription during cluster creation or these are pre-created roles definition with a bigger scope that are going to be created by us during rollout in a region?

tony-schndr: The roles are custom roles deployed to the dev sub.

bennerv: Update: when creating role assignments using the azure-sdk-for-go, you don't need to specify `/subscriptions/<sub-id>` as part of the role assignment resourceId.  The Authorization RP will infer it based on the scope you attempt to create it.  

This works for both custom roles and built-in roles.  We can remove `/subscriptions/<sub-id` from all of these.

machi1990 (approved):

bennerv (approved):

PR #819, opened 2024-11-08
provision shard generation with cx keyvaults

geoberle:
### What this PR does

adapts the provision shard configuration to the new format with the dedicated azure section, referencing the the DNS zone, and MGMT cluster KVs to use (CX secrets + CX MSI).

⚠️ only merge together with a CS promotion that implements the new provision shard structure

Jira: ARO-10856
Link to demo recording:

### DIFF
--- a/cluster-service/deploy/provisioning-shards.tmpl.yml
+++ b/cluster-service/deploy/provisioning-shards.tmpl.yml
@@ -11,8 +11,11 @@ provision_shards:
       "consumer_name": "{{ .maestroConsumerName }}"
     }
   status: active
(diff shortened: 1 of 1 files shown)


machi1990:
Hi @geoberle we can undraft the PR and move it ready to review. While doing so, the cs image tag needs to be bumped to `5528e96` which contains the change. Thanks

machi1990:
> Hi @geoberle we can undraft the PR and move it ready to review. While doing so, the cs image tag needs to be bumped to `5528e96` which contains the change. Thanks

Maybe he change in `5528e96` isn't the one we need as it'll require the env to be recreated 🤔

PR #825, opened 2024-11-11
templatize ev2 gap

geoberle:
### What this PR does

This PR tries to close the gap between `templatize` current haviour and EV2s mode of operations.

#### Name generation

This PR removes the name generator functions e.g. `azureKeyVaultName` since for EV2 the real names need to be determined at runtime not at buildtime. so we restrict ourselves to only use `ctx.region` `ctx.regionShort` and `ctx.stamp` 
this sadly has the impact that certain names for resources will change and that means it is disruptive for DEVs and our permantent DEV envs. after merging this, ENVs need to be recreated.

#### Generate with EV2 placeholders

generate an output file and replaces all placeholders with EV2 double-underscore ones when `--ev2-placeholders` is specified. this supports delayed evaluation of variables to distinguish between the EV2 build and run stage. this also means that the naming functions like `azureKeyVaultName` have been removed. Instead `region-short` has been introduced to have a short region name (as defined by EV2) available for resource naming that will not exceed lenght limitations.

```bash
templatize generate ... --ev2-placeholders --input xx --output xx`

SERVICE_KV ?= __SERVICEKEYVAULTNAME__
OIDC_STORAGE_ACCOUNT ?= __OIDCSTORAGEACCOUNTNAME__
IMAGE_REPO ?= __CLUSTERSERVICEIMAGEREPO__
```
#### EV2 ServiceConfig helper

in order to support the generation of a ServiceConfig.json, the `pkg.ev2.GetNonRegionalServiceConfigVariables` function can be used to fetch the variables from config.yaml that should go into a service configs global section. all variable values are formatted in a way to support EV2 system variables, e.g. `$location(), $stamp(), $(someOtherConfigValue)`

the `pkg.ev2.GetNonRegionalServiceConfigVariables` function can be used to fetch the regional overrides for a deploy env, again formatted for EV2 system variable usage.

#### EV2 ScopeBinding helper

the `pkg.ev2.ScopeBindingVariables` function can be used to extract a find/replace suitable variable list for an EV2 ScopeBinding. the result is a dictionary with entries like `__VAR__ = $config(var)` or `__PARENT_CHILD__ = $config(parent.child)` for nested configs.

#### EV2 file preprocess helper

the `pkg.ev2.PreprocessFileForEV2` function can be used to process a file with {{}} placeholders into one that uses EV2 system variables like $location() etc.

#### Pipeline.yaml runner

the EV2 generator from the ADO repository will use a pipeline.yaml file to generate EV2 artifacts (see example in this PR).
it is important that we use this pipeline.yaml declaration also in DEV to drive deployments. this way such declarations remain well maintained.

```sh
templatize run-pipeline --config-file config/config.yaml --cloud public --deploy-env personal-dev -p maestro/server/pipeline.yaml --region westus3 --region-short usw3gobe --stamp 1
```

Right now, only `Shell` runner is supported. the idea is to revamp the Makefiles so they don't require the `config.tmpl.mk` files anymore but rely exclusively on ENV variables to be injected as defined in the pipeline.yaml. in DEV, the run-pipeline command will ensure these env vars are set. in higher envs, EV2 will take care of it.

depending on the subscription and RG, the run-pipeline command will provide a suitable kubeconfig. in higher envs EV2 will take care of this.

Jira: https://issues.redhat.com/browse/ARO-12087
Link to demo recording:


tooling/templatize/internal/ev2/mapping_test.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+	expectedFlattened := map[string]string{
+		"__KEY1__": "key1",
+		"__KEY2__": "key2",
+		"__KEY3__": "key3",
+		"__PARENT_NESTED__": "parent.nested",
+		"__PARENT_DEEPER_DEEPEST__": "parent.deeper.deepest",
+	}
+	flattened, replace := EV2Mapping(testData, []string{})
+	assert.DeepEqual(t, expectedFlattened, flattened)
```
stevekuznetsov, line 30: I know this is just a PoC - in my experience the output of `cmp.Diff()` is 1000x better than `assert.DeepEqual`

geoberle: thank you. adapted

tooling/templatize/cmd/generate/options.go
```diff
@@ -38,38 +40,60 @@ func BindGenerationOptions(opts *RawGenerationOptions, cmd *cobra.Command) error
...
 }
 
 type ValidatedGenerationOptions struct {
 	// Embed a private pointer that cannot be instantiated outside of this package.
 	*validatedGenerationOptions
 }
 
+// completedGenerationOptions is a private wrapper that enforces a call of Complete() before config generation can be invoked.
+type completedGenerationOptions struct {
+	*ValidatedGenerationOptions
```
stevekuznetsov, line 62: Why not completed?

tooling/templatize/cmd/rolloutoptions.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func BindRolloutOptions(opts *RawRolloutOptions, cmd *cobra.Command) error {
+	err := BindOptions(opts.BaseOptions, cmd)
+	if err != nil {
+		return fmt.Errorf("failed to bind options: %w", err)
+	}
+	cmd.Flags().StringVarP(&opts.Region, "region", "r", opts.Region, "resources location")
+	cmd.Flags().StringVar(&opts.RegionShort, "region-short", opts.RegionShort, "short region string")
```
stevekuznetsov, line 39: Missing a short-flag.

tooling/templatize/cmd/run/options.go
```diff
@@ -0,0 +1,112 @@
...
+	// Embed a private pointer that cannot be instantiated outside of this package.
+	*completedRunOptions
+}
+
+func (o *RawRunOptions) Validate() (*ValidatedRunOptions, error) {
+	validatedRolloutOptions, err := o.RolloutOptions.Validate()
+	if err != nil {
+		return nil, err
+	}
+	// todo validate here
```
stevekuznetsov, line 68: Outstanding or for another PR?

tooling/templatize/cmd/run/options.go
```diff
@@ -0,0 +1,112 @@
...
+			Pipeline:            pipeline,
+		},
+	}, nil
+}
+
+func (o *RunOptions) RunPipeline() error {
+	variables, err := o.RolloutOptions.Options.ConfigProvider.GetVariables(
+		o.Cloud,
+		o.DeployEnv,
+		o.Region,
```
stevekuznetsov, line 101: I haven't read the code for this yet but at first blush it's surprising that you need to pass `o.Region` to `GetVariables()` as well as to `NewEvaluationContext()` - is it valid to pass two different regions?

tooling/templatize/pkg/config/config.go
```diff
@@ -0,0 +1,191 @@
...
+	config, err := cp.loadConfig(DefaultConfigEvaluationContext())
+	if err != nil {
+		return err
+	}
+	if _, ok := config.Overrides[cloud]; !ok {
+		return fmt.Errorf("the cloud %s is not found in %s", cloud, cp.config)
+	}
+
+	if _, ok := config.Overrides[cloud].Overrides[deployEnv]; !ok {
+		return fmt.Errorf("the deployment env %s is not found under cloud %s in %s", deployEnv, cloud, cp.config)
```
stevekuznetsov, line 117: What does the last `%s` do with `cp.config`?

tooling/templatize/pkg/config/config.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return nil, err
+	}
+	err = cp.Validate(cloud, deployEnv)
+	if err != nil {
+		return nil, err
+	}
+
+	variables := Variables{}
+	mergeVariables(variables, config.Defaults)
```
stevekuznetsov, line 133: Nice, this is quite concise & readable.

tooling/templatize/pkg/config/config_test.go
```diff
@@ -28,7 +28,7 @@ func TestConfigProvider(t *testing.T) {
 	assert.Equal(t, "aro-hcp-int.azurecr.io/maestro-server:the-stable-one", variables["maestro_image"])
 
 	// key is in the config file, default, varaible value
-	assert.Equal(t, fmt.Sprintf("hcp-underlay-%s-%s", region, regionStamp), variables["region_resourcegroup"])
+	assert.Equal(t, fmt.Sprintf("hcp-underlay-%s-%s", region, stamp), variables["region_resourcegroup"])
 }
 
 func TestInterfaceToVariable(t *testing.T) {
```
stevekuznetsov, line 34: Please add a test here with nested maps in `i`.

stevekuznetsov: Please add test cases to `TestMergeVariable` for nested sub-maps overwriting parent maps (both at one level of overwriting and more than one)

tooling/templatize/pkg/config/types.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (v Variables) AddNested(key string, other map[string]interface{}) error {
+	if _, exists := v[key]; exists {
+		return fmt.Errorf("%s exists already in Variables", key)
+	}
+	other[key] = other
+	return nil
+}
+
+type VariableOverrides struct {
```
stevekuznetsov, line 19: Structurally, what is the difference between `VariableOverrides`, `CloudVariableOverride` and `DeployEnvVariableOverride`? Why do we need three types?

stevekuznetsov: Why do the `*Override` structs have exported members if it seems like you need to use the getters for correctly traversing these? Perhaps an unexported struct you could unmarshal into + an interface would be better if you want to guarantee callers use the getters.

tooling/templatize/pkg/ev2/mapping_test.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	expectedFlattened := map[string]string{
+		"__KEY1__":                  "key1",
+		"__KEY2__":                  "key2",
+		"__KEY3__":                  "key3",
+		"__PARENT_NESTED__":         "parent.nested",
+		"__PARENT_DEEPER_DEEPEST__": "parent.deeper.deepest",
+	}
+	flattened, replace := EV2Mapping(testData, []string{})
+	assert.DeepEqual(t, expectedFlattened, flattened)
```
stevekuznetsov, line 32: I'd suggest `assert.Empty(t, cmp.Diff(expectedFlattened, flattened))` - the outputs on failure are much nicer.

tooling/templatize/pkg/pipeline/types.go
```diff
@@ -0,0 +1,24 @@
...
+type Step struct {
+	Name          string `yaml:"name"`
+	Subscription  string `yaml:"subscription"`
+	ResourceGroup string `yaml:"resourceGroup"`
+	Action        Action `yaml:"action"`
+}
+
+type Action struct {
+	Type      string   `yaml:"type"`
+	Command   string   `yaml:"command"`
```
stevekuznetsov, line 17: This should be `[]string` for the entire `cmdline` with args.

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -0,0 +1,190 @@
...
+	// build ENV vars
+	envVars := os.Environ()
+	envVars = append(envVars, fmt.Sprintf("KUBECONFIG=%s", kubeconfigFile))
+	for _, e := range s.Action.Variables {
+		value := vars[e.Value] // todo nested lookups
+		envVars = append(envVars, fmt.Sprintf("%s=%s", e.Name, value))
+	}
+
+	// execute the shell command with the environment variables
+	cmd := exec.Command("/bin/sh", "-c", s.Action.Command)
```
stevekuznetsov, line 101: I don't think it benefits us to assume this - the Ev2 configuration can support whatever. We should just run `[]string` into `exec` and call it a day.

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// Create a new subscriptions client
+	client, err := armsubscriptions.NewClient(cred, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to create subscriptions client: %v", err)
+	}
+
+	// List subscriptions and find the one with the matching name
+	ctx := context.Background()
```
stevekuznetsov, line 129: I would suggest creating a `ctx` child from `Background()` using `signal.NotifyContext()` and threading it through to these calls as well as your `exec.Cmd`. You'll be happy you did when you ctrl+C the CLI :)

https://pkg.go.dev/os/signal#NotifyContext

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -0,0 +1,190 @@
...
+	pager := client.NewListPager(nil)
+	for pager.More() {
+		page, err := pager.NextPage(ctx)
+		if err != nil {
+			return "", fmt.Errorf("failed to get next page of subscriptions: %v", err)
+		}
+		for _, sub := range page.Value {
+			if *sub.DisplayName == subscriptionName {
```
stevekuznetsov, line 137: Check for `nil` before dereferencing

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	defer tmpfile.Close()
+
+	// store the kubeconfig content into the temporary file
+	if _, err := tmpfile.Write([]byte(kubeconfigContent)); err != nil {
+		return "", fmt.Errorf("failed to write to temporary kubeconfigfile %s: %v", tmpfile.Name(), err)
+	}
+
+	// Run kubelogin to transform the kubeconfig
+	cmd := exec.Command("kubelogin", "convert-kubeconfig", "-l", "azurecli", "--kubeconfig", tmpfile.Name())
```
stevekuznetsov, line 183: `exec.CommandContext`

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -0,0 +1,190 @@
...
+	subscriptionID, err := lookupSubscriptionID(s.Subscription)
+	if err != nil {
+		return fmt.Errorf("failed to lookup subscription ID: %w", err)
+	}
+
+	// prepare kubeconfig
+	kubeconfigFile, err := buildAKSKubeConfig(
+		subscriptionID,
+		s.ResourceGroup,
+		"aro-hcp-aks",
```
stevekuznetsov, line 78: Looks like this might be an input in the `pipeline.yaml`?

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -0,0 +1,190 @@
...
+		envVars = append(envVars, fmt.Sprintf("%s=%s", e.Name, value))
+	}
+
+	// execute the shell command with the environment variables
+	cmd := exec.Command("/bin/sh", "-c", s.Action.Command)
+	cmd.Dir = "/Users/goberlec/dev/aro-hcp/maestro/server"
+	cmd.Env = append(cmd.Env, envVars...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to execute shell command: %s %w", output, err)
```
stevekuznetsov, line 106: I think you want `string(output)` or you get the byte array

tooling/templatize/pkg/pipeline/types.go
```diff
@@ -0,0 +1,24 @@
...
+	Variables []EnvVar `yaml:"variables"`
```
stevekuznetsov, line 18: My hunch is it would be much more ergonomic to have these in the `Makefile`

maestro/server/pipeline.yaml
```diff
@@ -0,0 +1,29 @@
...
+- name: deploy
+  subscription: {{ .serviceClusterSubscription }}
+  resourceGroup: {{ .serviceClusterRG }}
+  action:
+    type: Shell
+    command: "make deploy"
+    # we could infer the pwd from the location of this file and could avoid make -C
+    variables:
+    - name: EVENTGRID_NAME
+      value: maestroEventgridName
```
stevekuznetsov, line 13: What does this mean? I think you need the `{{ .maestro.eventGridName }}` syntax here?

tooling/templatize/cmd/run/options.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+type ValidatedRunOptions struct {
+	// Embed a private pointer that cannot be instantiated outside of this package.
+	*validatedRunOptions
+}
+
+// completedRunOptions is a private wrapper that enforces a call of Complete() before config generation can be invoked.
+type completedRunOptions struct {
+	*ValidatedRunOptions
```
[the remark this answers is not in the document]
geoberle: removed it

tooling/templatize/cmd/generate/options.go
geoberle: removed it

tooling/templatize/cmd/options.go
```diff
@@ -6,33 +6,25 @@ import (
...
+	"github.com/Azure/ARO-HCP/tooling/templatize/pkg/config"
 )
 
 func DefaultOptions() *RawOptions {
 	return &RawOptions{}
 }
 
 func BindOptions(opts *RawOptions, cmd *cobra.Command) error {
-	cmd.Flags().StringVar(&opts.ConfigFile, "config-file", opts.ConfigFile, "config file path")
+	cmd.Flags().StringVarP(&opts.ConfigFile, "config-file", "c", opts.ConfigFile, "config file path")
```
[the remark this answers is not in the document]
geoberle: i removed the short flags as i was not able to find meaningful conflict-free flags for every option

tooling/templatize/cmd/rolloutoptions.go
geoberle: decided to go without short flags

tooling/templatize/cmd/run/options.go
geoberle: added base validation

tooling/templatize/pkg/config/config.go
geoberle: cp.config is the configfile path.
i removed it since it is not really necessary that the config file name is part of the error message for this particular situation

tooling/templatize/pkg/config/config_test.go
geoberle: * test for nested maps for TestInterfaceToVariable done (+ fixed a bug)
* not sure about the additional `TestMergeVariable` test cases you asked about. have a look at the two onces i added

tooling/templatize/pkg/config/types.go
geoberle: > Why do we need three types?

their structure is the same but the struct tags for yaml parsing are different. i made the cloud and deployenv ones directly nested types within the outer one. let me know what you think.

> exported members

i made everything non-exported and defined accessors on an interface

tooling/templatize/pkg/ev2/mapping_test.go
geoberle: nice. adopted

tooling/templatize/pkg/pipeline/run.go
geoberle: done

tooling/templatize/pkg/pipeline/run.go
geoberle: done

tooling/templatize/pkg/pipeline/run.go
geoberle: done

maestro/server/pipeline.yaml
geoberle: as discussed, let's keep this a name based ref and refactor later to event not require this section at all.
for the time being i renamed `value` to `configRef` to make it clearer this is a ref to a config and not a value

tooling/templatize/pkg/pipeline/types.go
geoberle: agreed. let's keep it like this for the time being so we can move on to something that works now.
there is a way forward to not require this section at all.

tooling/templatize/pkg/pipeline/run.go
geoberle: done

tooling/templatize/pkg/pipeline/run.go
geoberle: done

tooling/templatize/cmd/run/options.go
geoberle:
```
GetVariables(cloud, deployEnv, region string, configEvalContext *ConfigEvaluationContext) (Variables, error)
```
* the region that is directly passed into `GetVariables` determines what regional overrides we want to be considered when returning config variables
* the region that is part of the configEvalContext is the one use as replacement value for `{{ .ctx.region }}` in config preprocessing

while in DEV both are the same, for EV2 they are different:
* configEvalContext.region is `$location()`
* the other is an actual region name

 i renamed `ConfigEvaluationContext` to `ConfigReplacements` and also renamed its fields to make it more apparent what they are about.

tony-schndr (approved):
lgtm

PR #829, opened 2024-11-12
Centralize writing HTTP response bodies

mbarnes:
### What this PR does

Adds functions to the `/internal/api/arm` package to correctly write HTTP response bodies with consistent JSON formatting:

```
// Marshal returns the JSON encoding of v.
//
// Call this function instead of the marshal functions in "encoding/json" for
// HTTP responses to ensure the formatting is consistent.
func Marshal(v any) ([]byte, error)

// WriteJSONResponse writes a JSON response body to the http.ResponseWriter in
// the proper sequence: first setting Content-Type to "application/json", then
// setting the HTTP status code, and finally writing a JSON encoding of body.
//
// The function accepts anything for the body argument that can be marshalled
// to JSON. One special case, however, is a byte slice. A byte slice will be
// written verbatim with the expectation that it was produced by Marshal.
func WriteJSONResponse(writer http.ResponseWriter, statusCode int, body any) (int, error)
```

Having these in the `arm` package allows them to be called from anywhere.

In the past I tried to centralize adding `Content-Type: application/json` headers to response bodies ([#677](https://github.com/Azure/ARO-HCP/pull/677)) but it didn't work because I misunderstood `http.ResponseWriter` semantics (and apparently didn't test it very well).  This PR reverts that attempted solution.

Jira: No Jira, just refactoring.
Link to demo recording:

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -249,14 +249,7 @@ func (f *Frontend) ArmResourceList(writer http.ResponseWriter, request *http.Req
 		}
 	}
 
-	resp, err := json.Marshal(pagedResponse)
-	if err != nil {
-		f.logger.Error(err.Error())
-		arm.WriteInternalServerError(writer)
-		return
-	}
-
-	_, err = writer.Write(resp)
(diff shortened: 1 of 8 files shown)


SudoBrendan (approved):
nice refactor, thanks!

PR #831, opened 2024-11-12
handle ResourceGroup ScopeLocked error

tony-schndr:
### What this PR does
Fixes cleanup automation so the script does not fail when a resource group is ScopeLocked and missing persist=true.

Also ignores some python files and bumps requests as it was yanked.

Jira: 
Link to demo recording:

### DIFF
--- a/tooling/azure-automation/resources-cleanup/src/resources_cleanup.py
+++ b/tooling/azure-automation/resources-cleanup/src/resources_cleanup.py
@@ -120,10 +120,10 @@ def process_resource_group(resource_group: ResourceGroup, resource_client: Resou
         result_poller = resource_client.resource_groups.begin_delete(resource_group_name)
         print(f"result_poller of resource group deletion: {result_poller}")
     except HttpResponseError as err:
-        target_error_code = "DenyAssignmentAuthorizationFailed"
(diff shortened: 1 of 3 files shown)


geoberle (dismissed):

janboll (dismissed):

tooling/azure-automation/resources-cleanup/src/resources_cleanup.py
```diff
@@ -120,10 +120,11 @@ def process_resource_group(resource_group: ResourceGroup, resource_client: Resou
         result_poller = resource_client.resource_groups.begin_delete(resource_group_name)
         print(f"result_poller of resource group deletion: {result_poller}")
     except HttpResponseError as err:
-        target_error_code = "DenyAssignmentAuthorizationFailed"
-        if err.error.code == target_error_code:
+        if err.error.code == "DenyAssignmentAuthorizationFailed":
```
bennerv, line 123: something like (my python probably isn't 100% valid, but you get the point)
```suggestion
ignored_error_codes = ("DenyAssignmentAuthorizationFailed", "ScopeLocked")
if err.error.code in ignored_error_codes: 
    print("skipping deletion of resource group due to error code %s", err.error.code)

PR #847, opened 2024-11-15
Add leader election to the backend

mbarnes:
### What this PR does

Adds leader election to the backend deployment using the [k8s.io/client-go/tools/leaderelection](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection) module so it can be horizontally scaled but still restrict active polling to one pod at a time.

Jira: No Jira, added by request
Link to demo recording: 

- The leader election module emits unstructured log messages through [klog](https://pkg.go.dev/k8s.io/klog/v2), so I had to catch these messages and try to adapt them to the standard library [structured logger](https://pkg.go.dev/log/slog), which required some futzing.  Please enlighten me if there's a better way to do this.

- I went ahead and bumped the backend replica count to 2 just to prove it works in dev environments.

- We should add a `/healthz` endpoint to the backend at some point.  The leader election module offers healthz integration for when it fails to renew the leader lease, but that's a pull request for another day.

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -11,15 +11,32 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime/debug"
+	"sync/atomic"
 	"syscall"
+	"time"
 
+	"github.com/go-logr/logr"
 	ocmsdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"k8s.io/klog/v2"
 
 	"github.com/Azure/ARO-HCP/internal/database"
 )
 
+const (
+	leaderElectionLockName      = "backend-leader"
+	leaderElectionLeaseDuration = 15 * time.Second
+	leaderElectionRenewDeadline = 10 * time.Second
+	leaderElectionRetryPeriod   = 2 * time.Second
+)
+
 var (
+	argKubeconfig         string
+	argNamespace          string
 	argLocation           string
 	argCosmosName         string
 	argCosmosURL          string
@@ -50,6 +67,8 @@ var (
 func init() {
 	rootCmd.SetErrPrefix(rootCmd.Short + " error:")
 
+	rootCmd.Flags().StringVar(&argKubeconfig, "kubeconfig", "", "Absolute path to the kubeconfig file")
+	rootCmd.Flags().StringVar(&argNamespace, "namespace", os.Getenv("NAMESPACE"), "Kubernetes namespace")
 	rootCmd.Flags().StringVar(&argLocation, "location", os.Getenv("LOCATION"), "Azure location")
 	rootCmd.Flags().StringVar(&argCosmosName, "cosmos-name", os.Getenv("DB_NAME"), "Cosmos database name")
 	rootCmd.Flags().StringVar(&argCosmosURL, "cosmos-url", os.Getenv("DB_URL"), "Cosmos database URL")
@@ -68,9 +87,42 @@ func init() {
 	}
 }
 
+func newKubeconfig(kubeconfig string) (*rest.Config, error) {
+	loader := clientcmd.NewDefaultClientConfigLoadingRules()
+	if kubeconfig != "" {
+		loader.ExplicitPath = kubeconfig
+	}
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loader, nil).ClientConfig()
+}
+
 func Run(cmd *cobra.Command, args []string) error {
 	handler := slog.NewJSONHandler(os.Stdout, nil)
 	logger := slog.New(handler)
+	klog.SetLogger(logr.FromSlogHandler(handler))
+
+	// Use pod name as the lock identity.
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+
+	kubeconfig, err := newKubeconfig(argKubeconfig)
+	if err != nil {
+		return fmt.Errorf("Failed to create Kubernetes configuration: %w", err)
+	}
+
+	leaderElectionLock, err := resourcelock.NewFromKubeconfig(
+		resourcelock.LeasesResourceLock,
+		argNamespace,
+		leaderElectionLockName,
+		resourcelock.ResourceLockConfig{
+			Identity: hostname,
+		},
+		kubeconfig,
+		leaderElectionRenewDeadline)
+	if err != nil {
+		return fmt.Errorf("Failed to create leader election lock: %w", err)
+	}
 
 	// Create the database client.
 	cosmosDatabaseClient, err := database.NewCosmosDatabaseClient(argCosmosURL, argCosmosName)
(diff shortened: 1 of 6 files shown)


backend/main.go
```diff
@@ -72,6 +92,14 @@ func init() {
 	}
 }
 
+func newKubeconfig(kubeconfig string) (*rest.Config, error) {
```
stevekuznetsov, line 95: Use the `clientcmd` package loaders (e.g. [this one](https://pkg.go.dev/k8s.io/client-go/tools/clientcmd#NewNonInteractiveDeferredLoadingClientConfig)) to model the chain of loaders you want to use in the order you want them. Easy to set up `$KUBECONFIG > --kubeconfig > in-cluster config`

backend/main.go
```diff
@@ -106,6 +134,52 @@ func Run(cmd *cobra.Command, args []string) error {
 	handler := slog.NewJSONHandler(os.Stdout, nil)
 	logger := slog.New(handler)
 
+	// Try to massage unstructured Kubernetes logs into
+	// something that fits better with structured logs.
+	klog.SetLoggerWithOptions(
```
stevekuznetsov, line 139: `klog/v2` allows you to use some other logging backend, why not just put `slog` in there as a `logr.Logger`?

backend/main.go
```diff
@@ -125,17 +199,39 @@ func Run(cmd *cobra.Command, args []string) error {
...
-
-	go operationsScanner.Run(logger, stop)
-
-	sig := <-signalChannel
-	logger.Info(fmt.Sprintf("caught %s signal", sig))
-	close(stop)
-
-	operationsScanner.Join()
+	go func() {
+		sig := <-signalChannel
```
stevekuznetsov, line 208: Please just use `signal.NotifyContext` on this thing? Didn't we already talk about this?

backend/operations_scanner.go
```diff
@@ -72,19 +72,19 @@ func (s *OperationsScanner) Run(logger *slog.Logger, stop <-chan struct{}) {
...
-
 	// Poll database immediately on startup.
-	s.pollDBOperations(ctx, logger)
+	s.pollDBOperations(logger)
 
-	for {
+	// The Context is cancelled when the process receives an interrupt signal.
+	// We let the periodic functions use their own Context so they can finish
+	// gracefully before we return.
+	for ctx.Err() == nil {
```
stevekuznetsov, line 81: IMO this is tricky code - just use `if ctx.Err() != nil { break }` in the `select` statement. Not obvious to me exactly the order of operations here - as written right now there are multiple steps:

1. `<-ctx.Done()`
2. `for ctx.Err() == nil`

If anyone ever changes this loop and execution is allowed to happen between the two, it will be harder to reason about the state of `ctx.Err()`.

backend/main.go
mbarnes: I want to log the signal received along with a timestamp so I can observe how much longer it takes the process to shut down.  I don't see any big advantage to `signal.NotifyContext`; it's a convenience wrapper that does the same thing I'm doing.

backend/main.go
mbarnes: Looks like [logr.FromSlogHandler](https://pkg.go.dev/github.com/go-logr/logr#FromSlogHandler) might work.

backend/main.go
mbarnes: Update 1st commit.

backend/main.go
mbarnes: Yep, it worked.  Updated 1st commit.

backend/operations_scanner.go
mbarnes: Label worked.  Updated 1st commit.

stevekuznetsov:
@mbarnes  I think you meant to push an update but may have forgotten :)

backend/main.go
stevekuznetsov: Is `signalChannel` just for your logger? Then the following is simpler:

```go
ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
defer cancel()

go func() {
    <-ctx.Done()
    logger.Info("Caught interrupt signal")
}()
```

And the only difference is you can't tell if you got SIGINT or SIGTERM (why do you care? in production, the kubelet only ever sends SIGTERM or SIGKILL ([ref](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination)), so this seems like a distinction without much meaning). Code is much less Rube-Goldberg. IDK. I respect you think this logging of the specific signal that is sent is somehow important but in a decade of writing k8s-native applications it has never been useful or required in the past.

stevekuznetsov: As far as big advantage, I did want to mention - using a more complex cascade of handlers and goroutines when it's not necessary is a downside for sure. Making the simple choice every time you can helps make this maintainable ten years down the line.

backend/operations_scanner.go
```diff
@@ -93,9 +97,11 @@ func (s *OperationsScanner) Join() {
 	<-s.done
 }
 
-func (s *OperationsScanner) pollDBOperations(ctx context.Context, logger *slog.Logger) {
+func (s *OperationsScanner) pollDBOperations(logger *slog.Logger) {
 	var activeOperations []*database.OperationDocument
 
+	ctx := context.Background()
```
SudoBrendan, line 103: I think we can preserve the root context here, because we aren't issuing a write or other non-interruptable operation with the context (it looks like it's just a List + Iterate - safe to cancel?)

backend/operations_scanner.go
```diff
@@ -123,35 +129,32 @@ func (s *OperationsScanner) pollDBOperations(ctx context.Context, logger *slog.L
 	}
 }
 
-func (s *OperationsScanner) pollCSOperations(ctx context.Context, logger *slog.Logger, stop <-chan struct{}) {
+func (s *OperationsScanner) pollCSOperations(logger *slog.Logger) {
 	var activeOperations []*database.OperationDocument
 
+	ctx := context.Background()
```
SudoBrendan, line 135: nit: let's add a note to why this context is needed (we are performing write operations that should not be canceled)

SudoBrendan (dismissed):

backend/operations_scanner.go
mbarnes: Agree, this function is safe to insta-terminate.

backend/operations_scanner.go
mbarnes: Will do.  Worth mentioning, too, is once we start using Cosmos DB transactions (soon) this concern goes away.

backend/operations_scanner.go
```diff
@@ -123,35 +124,42 @@ func (s *OperationsScanner) pollDBOperations(ctx context.Context, logger *slog.L
 	}
 }
 
-func (s *OperationsScanner) pollCSOperations(ctx context.Context, logger *slog.Logger, stop <-chan struct{}) {
+func (s *OperationsScanner) pollCSOperations(logger *slog.Logger) {
```
venkateshsredhat, line 127: Are we planning to get the ctx back ,Once we implement the batch transactions work in cosmo DB .

backend/main.go
```diff
@@ -96,17 +148,36 @@ func Run(cmd *cobra.Command, args []string) error {
...
+	// FIXME Integrate leaderelection.HealthzAdaptor into a /healthz endpoint.
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
+		Lock:          leaderElectionLock,
+		LeaseDuration: leaderElectionLeaseDuration,
+		RenewDeadline: leaderElectionRenewDeadline,
+		RetryPeriod:   leaderElectionRetryPeriod,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				startedLeading.Store(true)
+				go operationsScanner.Run(ctx, logger)
```
venkateshsredhat, line 170: Also a general Question around this feature , so with the replica being 2 . So the leader pod will be the one executing this method which executes pollDBOperations and pollCSOperations , what is the other pod doing meanwhile ? --- is it just for HA and will try to get the lease and be a leader or it will be doing more than that .

backend/operations_scanner.go
mbarnes: Yes.

backend/main.go
mbarnes: > is it just for HA and will try to get the lease and be a leader
Exactly that.  The other replicas are just on standby in case the leader pod dies.

venkateshsredhat (approved):

PR #851, opened 2024-11-18
RP pipelines

geoberle:
### What this PR does

pipeline definition for infra + RP

Jira: https://issues.redhat.com/browse/ARO-12212
Link to demo recording:


tooling/templatize/pkg/aks/admin.go
```diff
@@ -0,0 +1,181 @@
...
+			if err == nil {
+				fmt.Println("Cluster admin permissions are now effective")
+				return nil
+			}
+			fmt.Println("Waiting for role assignment to be effective...")
+		}
+	}
+}
+
+func TestClusterAdminPermissions(ctx context.Context, kubeconfigPath string) error {
```
janboll, line 79: Please use another prefix, not `test`.
I think this can be a private method

tooling/templatize/pkg/aks/admin.go
```diff
@@ -0,0 +1,181 @@
...
+	clientset, err := createKubeClient(kubeconfigPath)
```
janboll, line 80: Think it's okay to pass this in? Perhaps worth creating a client interface

tooling/templatize/pkg/config/types.go
```diff
@@ -10,6 +11,24 @@ type configProviderImpl struct {
 
 type Variables map[string]interface{}
 
+func (v Variables) GetByPath(path string) (interface{}, bool) {
+    keys := strings.Split(path, ".")
+    var current interface{} = v
```
janboll, line 16: please use `var current any = v`

tooling/templatize/pkg/pipeline/arm.go
```diff
@@ -0,0 +1,135 @@
...
+	// Ensure resourcegroup exists
+	err = s.ensureResourceGroupExists(ctx, executionTarget)
+	if err != nil {
+		return fmt.Errorf("failed to ensure resource group exists: %w", err)
+	}
+
+	// TODO handle dry-run
+
+	// Run deployment
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
```
janboll, line 34: An idea to make this testable, what do you off an interface for azure client, that you can mock in testing. See for instance https://github.com/app-sre/aws-resource-exporter/blob/master/pkg/awsclient/awsclient.go#L26 for an example on how we did in aws

backend/pipeline.yaml
```diff
@@ -0,0 +1,21 @@
+serviceGroup: Microsoft.Azure.ARO.Test
+rolloutName: RP - Backend
+resourceGroups:
+- name: {{ .svc.rg }}
+  subscription: {{ .svc.subscription }}
+  aksCluster: {{ .aksName }}
+  steps:
+  - name: deploy
+    action: Shell
+    command: ["/bin/bash", "-c", "make deploy"]
```
stevekuznetsov, line 10: Why do you need to run `/bin/bash`? I think this would be OK to just have `["make", "deploy"]`

dev-infrastructure/region-pipeline.yaml
```diff
@@ -0,0 +1,16 @@
...
+rolloutName: Region Rollout
+resourceGroups:
+- name: {{ .regionRG }}
+  subscription: {{ .serviceClusterSubscription }}
+  steps:
+  - name: region
+    action: ARM
+    template: templates/region.bicep
+    parameters: configurations/region.tmpl.bicepparam
+  - name: region
```
stevekuznetsov, line 11: @whober0521 good to validate duplicate names and that `dependsOn` actually refers to a named step before this one.

Another question - if these steps are in a list, and they are executed one after the other, is the `dependsOn` actually necessary?

tooling/templatize/cmd/pipeline/inspect/cmd.go
```diff
@@ -0,0 +1,42 @@
...
+		Long:  "inspect aspects of a pipeline.yaml file",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx, stop := signal.NotifyContext(cmd.Context(), os.Interrupt, syscall.SIGTERM)
+			defer stop()
+
+			return runInspect(ctx, opts)
+		},
+	}
+	if err := BindOptions(opts, cmd); err != nil {
+		log.Fatal(err)
```
stevekuznetsov, line 27: nit: almost always better to have `os.Exit()` type stuff only in a `func main()` and return errors elsewhere so you can reason about control flow.

tooling/templatize/pkg/aks/admin.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	// Validate assignment
+	err = CheckClusterAdminPermissions(ctx, kubeconfigPath)
+	if err == nil {
+		return nil
+	}
+
+	// Wait for role assignment to be effective
+	fmt.Println("Wait for role assignment to be effective")
```
stevekuznetsov, line 57: You use a logger elsewhere - why not here?

tooling/templatize/pkg/aks/admin.go
stevekuznetsov: What are you envisioning with an interface? `kubernetes.Interface` already exists to close over `kubernetes.ClientSet`

tooling/templatize/pkg/aks/admin.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func CheckClusterAdminPermissions(ctx context.Context, kubeconfigPath string) error {
+	clientset, err := createKubeClient(kubeconfigPath)
+	if err != nil {
+		return fmt.Errorf("failed to create Kubernetes client: %w", err)
+	}
+
+	// Implement the logic to test cluster admin permissions
```
stevekuznetsov, line 85: Depending on how the kubeconfig is created, the API server may not necessarily be running at that time - not sure if you will hit this using AKS, but in the past I've seen a lot of flakes around this kind of logic that did not handle network flakes or server being temporarily down.

tooling/templatize/pkg/aks/kubeconfig.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
+)
+
+func GetKubeConfig(ctx context.Context, subscriptionID, resourceGroupName, aksClusterName string) (string, error) {
```
stevekuznetsov, line 13: If you're shelling out to `kubelogin`, any specific reason not to shell out to `az aks get-credential`?

tooling/templatize/pkg/config/types.go
```diff
@@ -11,6 +12,24 @@ type configProviderImpl struct {
 
 type Variables map[string]any
 
+func (v Variables) GetByPath(path string) (any, bool) {
```
stevekuznetsov, line 15: This was precisely the thing I wished we would not need to write. `{{ .parent["child.foo"].grandchild }}` will break something like this. The quicker we stop supporting an `env` list, the better.

tooling/templatize/pkg/pipeline/arm.go
stevekuznetsov: I would suggest end-to-end rather than an integration test. This is just glue code and if you mock out all of the interactions your integration test is low-value.

tooling/templatize/pkg/aks/kubeconfig.go
stevekuznetsov: I see other parts of the code shelling out to `az` CLI as well - IMO save yourself the time and maintenance for this code if it's simple to just write `az` calls.

tooling/templatize/pkg/aks/admin.go
geoberle: for our scenario the kubeapi was always accessible once our bicep templates completed

tooling/templatize/pkg/aks/kubeconfig.go
geoberle: i wanted to avoid shelling out as much as possible.
i wanted to revisite if we can use `kubelogin` as a library to avoid shelling out there

tooling/templatize/pkg/config/types.go
geoberle: +1

tooling/templatize/pkg/aks/admin.go
geoberle: revamped for using logr

dev-infrastructure/region-pipeline.yaml
geoberle: i've added some basic validation that the referenced step exists, but needs more love (circular deps etc)

backend/pipeline.yaml
geoberle: you are right, it is fine without it. i removed it

tooling/templatize/pkg/aks/admin.go
```diff
@@ -0,0 +1,181 @@
...
+	_, err = clientset.CoreV1().Pods("default").List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list pods in the default namespace: %w", err)
+	}
+	return nil
+}
+
+func getCurrentUserObjectID(ctx context.Context) (string, error) {
+	// Create a Graph client using Azure Credentials
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
```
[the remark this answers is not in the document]
geoberle: we need to merge this for MSFT to be able to proceed with the config.msft.yaml file, the schema and the Makefile changes.
i noted an action item

tooling/templatize/pkg/aks/admin.go
geoberle: i noted down an action item as we need to merge this soon without this change being considered

tooling/templatize/pkg/pipeline/arm.go
geoberle: i wrote down an action item

tooling/templatize/pkg/aks/kubeconfig.go
geoberle: i wrote down an action item

janboll (approved):

PR #857, opened 2024-11-20
Update hypershift and CS image

ahitacat:
### What this PR does
Updates the hypershift-operator and CS image to have the latest changes of hypershift.

Jira: 
Link to demo recording:

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -19,7 +19,7 @@ defaults:
   # Hypershift
   hypershift:
     namespace: hypershift
-    additionalInstallArg: ''
+    additionalInstallArg: '--tech-preview-no-upgrade'
     externalDNSManagedIdentityName: external-dns
     externalDNSServiceAccountName: external-dns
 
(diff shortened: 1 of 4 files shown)


config/config.yaml
```diff
@@ -220,4 +220,4 @@ clouds:
           # Hypershift
           # uncomment the following line if you want to install the hypershift operator
           # with CRD support for managedIdentities
-          #hypershiftAdditionalInstallArg: '--tech-preview-no-upgrade'
+          hypershiftAdditionalInstallArg: '--tech-preview-no-upgrade'
```
ahitacat, line 223: This is required to create clusters with the latest hypershift-operator. 
I'm not sure if that will be needed in the rest of env (dev and cs-pr).

janboll: Please change to:
```yaml
hypershift:
   additionalInstallArg: "--tech-preview-no-upgrade"
```

janboll:
You also need to run `make materialize` in the config directory as well

janboll (approved):

machi1990:
Since we are bumping CS,
I'd like to point out that this change requires a change of the provision shard config.
Similar to what is done here https://github.com/Azure/ARO-HCP/pull/819 
Additionally, as that change is going to impact how we interact with the provision shard, it'll also require the dev env to be recreated as it is a breaking change - the current CS commit doesn't contain a migrations script

geoberle:
Do we need to recreate the dev env also when there are no active clusters in the DB?

geoberle:
Ah just read the no-mig-script comment

PR #874, opened 2024-11-22
Minor fixes

janboll:
Remove aspect parameter to algin with model
Add some error format strings

### What this PR does

Jira: 
Link to demo recording:

### DIFF
--- a/tooling/templatize/testdata/pipeline.yaml
+++ b/tooling/templatize/testdata/pipeline.yaml
@@ -2,7 +2,7 @@ serviceGroup: Microsoft.Azure.ARO.Test
 rolloutName: Test Rollout
 resourceGroups:
 - name: {{ .regionRG }}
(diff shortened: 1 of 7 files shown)


tooling/templatize/testdata/pipeline.yaml
```diff
@@ -2,15 +2,15 @@ serviceGroup: Microsoft.Azure.ARO.Test
 rolloutName: Test Rollout
 resourceGroups:
 - name: {{ .regionRG }}
-  subscription: {{ .serviceClusterSubscription }}
```
geoberle, line 5: if you change this, you also need to change testdata/config.yaml

janboll: AH, my bad I used the config from the root folder. Let me revert that

geoberle (approved):

PR #878, opened 2024-11-22
Dry Run and tests

janboll:
### What this PR does

- Add the capability to run dry runs on shell script steps.
- Did some slight refactorings to add tests.
- Also moved AKS Kubeconfig creation into Resource group to only call it once, as it is not required to run it for every stip since it can be reused

Jira: 
Link to demo recording:


tooling/templatize/testdata/pipeline.yaml
```diff
@@ -8,10 +8,11 @@ resourceGroups:
...
+        - name: DRY_RUN
+          value: "true"
+      command: ["make", "-n", "deploy"]
     env:
     - name: MAESTRO_IMAGE
-      configRef: maestro_image
-  - name: svc
-    action: ARM
-    template: templates/svc-cluster.bicep
-    parameters: test.bicepparam
```
geoberle, line 17: please don't remove this. this is required by another test

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -93,6 +93,22 @@ func (rg *resourceGroup) run(ctx context.Context, options *PipelineRunOptions) e
 	}
 
 	logger := logr.FromContextOrDiscard(ctx)
+
+	var kubeconfigFile string
+	if !options.DryRun {
```
geoberle, line 98: we will need the real kubeconfig file for dry run with `helm --dry-run`

tooling/templatize/testdata/zz_fixture_TestRawOptions.yaml
```diff
@@ -11,6 +11,13 @@ resourceGroups:
     env:
     - name: MAESTRO_IMAGE
       configRef: maestro_image
+  - name: dry-run
+    action: Shell
+    command: ["make", "deploy"]
+    dryRun:
+      envVars:
+      - name: DRY_RUN
+        value: "A very dry one"
```
geoberle, line 20: :)

geoberle (approved):

PR #882, opened 2024-11-22
.github: do a functional check for az login

stevekuznetsov:
Instead of checking for the "cause" of login failure, check for the symptom, so we don't get false positives.

### What this PR does

Jira: 
Link to demo recording:

### DIFF
--- a/.github/workflows/aro-hcp-login-check.yml
+++ b/.github/workflows/aro-hcp-login-check.yml
@@ -1,0 +1,52 @@
+---
+  name: Azure Login Check
+  on:
+    workflow_dispatch:
+    pull_request:
+      branches:
+        - main
+      paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/aro-hcp-login-check.yml
```diff
@@ -0,0 +1,52 @@
+---
+  name: ARO HCP Continuous Deployment
```
bennerv, line 2: just need to update the name to something else so we don't conflict, otherwise lgtm .

bennerv: @stevekuznetsov did you update it?  I'm still seeing 
```
  name: ARO HCP Continuous Deployment

bennerv (approved):

PR #883, opened 2024-11-22
Implement collection GETs for node pools

mbarnes:
### What this PR does

Per the [Resource Provider Contract rules for Get Resource](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/get-resource.md#rules) (specifically rule **RPC-Get-V1-05**), the only collection GET endpoint required for nested tracked resources is on the parent resource:
```
GET /subscriptions/.../resourceGroups/.../providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/.../nodePools
```

I managed to generalize the `Frontend.ArmResourceList` handler function for both clusters and node pools.  The logic there was already fairly complicated and I wanted to avoid having two copies of it.

There's also a bunch of enhancements here to my "[failable push iterator](https://github.com/Azure/ARO-HCP/pull/749#discussion_r1812998066)" pattern, such as extending it to encapsulate pagination of Cluster Service list responses.  Unfortunately I can't really add unit tests for these since they depend on actual Cluster Service responses.

Jira: I couldn't find a specific user story for this, but
- it's a followup to [ARO-11853 - Nodepools as Tracked or Proxy Resources?](https://issues.redhat.com/browse/ARO-11853) being decided (Tracked)
- it's tangentially related to [ARO-6009 - HCP Cluster Resource Deletion Cascading Subscription Delete](https://issues.redhat.com/browse/ARO-6009)

Link to demo recording:


frontend/pkg/frontend/frontend.go
```diff
@@ -185,24 +188,43 @@ func (f *Frontend) ArmResourceList(writer http.ResponseWriter, request *http.Req
...
+		var doc database.ResourceDocument
+
+		err = json.Unmarshal(item, &doc)
+		if err != nil {
+			f.logger.Error(err.Error())
+			arm.WriteInternalServerError(writer)
+			return
+		}
+
+		if strings.HasSuffix(strings.ToLower(doc.Key.ResourceType.Type), resourceTypeName) {
```
stevekuznetsov, line 219: Can this be done with a `WHERE` clause at the DB layer?

internal/database/cache.go
```diff
@@ -24,14 +22,14 @@ type Cache struct {
 	subscription map[string]*SubscriptionDocument
 }
 
-type operationCacheIterator struct {
-	operation map[string]*OperationDocument
-	err       error
+type cacheIterator struct {
```
stevekuznetsov, line 25: For my understanding, why is there an in-memory cache for CosmosDB? Was there a particular REST p95 that couldn't be reached without caching?

mbarnes: It's there for historical reasons, from early in development before we had access to Cosmos DB.  It was decided to keep it around as a unit test aid, and for a time it was helpful when our database use cases were simple.  But it's technical debt now; it's become a obstacle to leveraging more advanced features of Cosmos DB.  I'm honestly ready to ditch it and just use [gomock](https://github.com/uber-go/mock) or some such for unit tests, but I'll need to find time to pay that debt.

frontend/pkg/frontend/frontend.go
mbarnes: Technically yes, but see my in-memory cache answer below.

I'm going to leave this as is for this PR but I added a "FIXME" comment here to circle back to it once I've gotten rid of the generic database interface I described below.

SudoBrendan (approved):
Per usual - your commits are so well organized that a PR this large was easily digested! Thanks Matt, excellent work! LGTM

PR #887, opened 2024-11-25
refactor templatize ev2 package

geoberle:
### What this PR does

* `PrecompilePipelineForEV2` renamed to `PrecompilePipelineFileForEV2`
* `PrecompilePipelineForEV2` reintroduced but returns a parsed `pipeline.Pipeline` struct
* export all `config.Pipeline` sub-structs
* export `NewEv2ConfigReplacements`
* refactor tests to use fixtures

Jira: 
Link to demo recording:

### DIFF
--- a/tooling/templatize/pkg/config/config.go
+++ b/tooling/templatize/pkg/config/config.go
@@ -258,7 +258,7 @@ func PreprocessFile(templateFilePath string, vars map[string]any) ([]byte, error
 	return processedContent, nil
 }
 
-// PreprocessFile processes a gotemplate from memory
+// PreprocessContent processes a gotemplate from memory
 func PreprocessContent(content []byte, vars map[string]any) ([]byte, error) {
 	tmpl := template.New("file")
 	tmpl, err := tmpl.Parse(string(content))
--- a/tooling/templatize/pkg/config/config.go
+++ b/tooling/templatize/pkg/config/config.go
@@ -247,20 +247,28 @@ func (cp *configProviderImpl) loadConfig(configReplacements *ConfigReplacements)
 // PreprocessFile reads and processes a gotemplate
 // The path will be read as is. It parses the file as a template, and executes it with the provided variables.
 func PreprocessFile(templateFilePath string, vars map[string]any) ([]byte, error) {
-	tmpl := template.New("file")
 	content, err := os.ReadFile(templateFilePath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read file %s: %w", templateFilePath, err)
 	}
-
-	tmpl, err = tmpl.Parse(string(content))
+	processedContent, err := PreprocessContent(content, vars)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse template %s: %w", templateFilePath, err)
+		return nil, fmt.Errorf("failed to preprocess file %s: %w", templateFilePath, err)
+	}
+	return processedContent, nil
+}
+
+// PreprocessFile processes a gotemplate from memory
+func PreprocessContent(content []byte, vars map[string]any) ([]byte, error) {
+	tmpl := template.New("file")
+	tmpl, err := tmpl.Parse(string(content))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse template: %w", err)
 	}
 
 	var tmplBytes bytes.Buffer
 	if err := tmpl.Option("missingkey=error").Execute(&tmplBytes, vars); err != nil {
-		return nil, fmt.Errorf("failed to execute template %s: %w", templateFilePath, err)
+		return nil, fmt.Errorf("failed to execute template: %w", err)
 	}
 	return tmplBytes.Bytes(), nil
 }
--- a/tooling/templatize/pkg/config/config_test.go
+++ b/tooling/templatize/pkg/config/config_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/Azure/ARO-HCP/tooling/templatize/internal/testutil"
 )
 
 func TestConfigProvider(t *testing.T) {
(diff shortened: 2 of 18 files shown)


tooling/templatize/pkg/config/config.go
```diff
@@ -247,20 +247,28 @@ func (cp *configProviderImpl) loadConfig(configReplacements *ConfigReplacements)
 // PreprocessFile reads and processes a gotemplate
 // The path will be read as is. It parses the file as a template, and executes it with the provided variables.
 func PreprocessFile(templateFilePath string, vars map[string]any) ([]byte, error) {
-	tmpl := template.New("file")
```
janboll, line 250: these should be easy to test ;-)

tooling/templatize/pkg/pipeline/common.go
```diff
@@ -0,0 +1,47 @@
+package pipeline
```
janboll, line 1: Please add tests for these

tooling/templatize/pkg/ev2/pipeline.go
```diff
@@ -10,73 +11,109 @@ import (
...
-	if err != nil {
-		return "", err
+	processedFiles := make(map[string][]byte)
+	_, scopeBoundVars := EV2Mapping(vars, []string{})
+	for _, rg := range processingPipeline.ResourceGroups {
+		for _, step := range rg.Steps {
+			// preprocess the parameters file with scopebinding variables
+			if step.Parameters != "" {
+				paramFileContent, ok := referencedFiles[step.Parameters]
+				if !ok {
```
janboll, line 103: Is this check necessary?  isn't line 79 covering this?

tooling/templatize/pkg/ev2/pipeline.go
```diff
@@ -10,73 +11,109 @@ import (
...
 			}
 		}
 	}
-	return nil
+	return referencedFiles, nil
 }
 
-func precompileFileAndStore(filePath string, vars map[string]interface{}) (string, error) {
-	preprocessedBytes, err := config.PreprocessFile(filePath, vars)
+func processPipelineForEV2(p *pipeline.Pipeline, referencedFiles map[string][]byte, vars config.Variables) (*pipeline.Pipeline, map[string][]byte, error) {
```
janboll, line 91: this should be simple to test, and very helpful since it has quite some conditions and side effects

janboll (changes requested):

tooling/templatize/pkg/config/config.go
```diff
@@ -247,20 +247,28 @@ func (cp *configProviderImpl) loadConfig(configReplacements *ConfigReplacements)
...
 	}
+	processedContent, err := PreprocessContent(content, vars)
+	if err != nil {
+		return nil, fmt.Errorf("failed to preprocess file %s: %w", templateFilePath, err)
+	}
+	return processedContent, nil
+}
 
-	tmpl, err = tmpl.Parse(string(content))
+// PreprocessFile processes a gotemplate from memory
```
stevekuznetsov, line 261:
```suggestion
// PreprocessContent processes a gotemplate from memory
```

tooling/templatize/pkg/config/config_test.go
```diff
@@ -231,3 +233,78 @@ func TestConvertToInterface(t *testing.T) {
 	assert.IsType(t, expected, map[string]any{})
 	assert.IsType(t, expected["key2"], map[string]any{})
 }
+
+func TestPreprocessContent(t *testing.T) {
+	fileContent, err := os.ReadFile("../../testdata/test.bicepparam")
```
stevekuznetsov, line 238: Consider `//go:embed` for static test data, much faster.

tooling/templatize/pkg/ev2/pipeline.go
```diff
@@ -10,73 +11,109 @@ import (
...
+	}
+
+	// precompile the pipeline and referenced files
+	processedPipeline, processedFiles, err := processPipelineForEV2(originalPipeline, referencedFiles, vars)
+	if err != nil {
+		return nil, err
+	}
+
+	// store the processed files to disk relative to the pipeline directory
+	_, restoreDir, err := processedPipeline.EnterPipelineDir()
```
stevekuznetsov, line 53: If you ever start to use concurrency for speed, altering the global state like this using syscalls is going to be a nightmare. Can you pass the path prefix into the functions that need it and have them compute relative offsets instead?

stevekuznetsov: `$GOMAXPROCS=<num-cpu>` and using goroutines to read & write files, passing data using channels is a great way to make bulk filesystem processing 100x-1000x faster depending on the speed of your disk.

tooling/templatize/pkg/ev2/pipeline.go
stevekuznetsov: 79 is in a different function altogether right?

tooling/templatize/pkg/ev2/pipeline.go
stevekuznetsov: Should be easy to test with static data in, static data out, cmp.Diff

stevekuznetsov: This is tested in `TestProcessPipelineForEV2` right? At bottom with `for filePath, content := range newFiles {`

tooling/templatize/pkg/pipeline/common.go
stevekuznetsov: Although @geoberle you could do a round-trip test e.g. `assert.Empty(cmp.diff(foo, foo.DeepCopy(foo.PipelinePath))`

stevekuznetsov (commented):
Looking good!

tooling/templatize/pkg/ev2/pipeline.go
geoberle: > global state

good point. let me rework this

tooling/templatize/pkg/config/config_test.go
geoberle: i'll do some readup on go:embed. ty

whober0521 (approved):

PR #899, opened 2024-11-27
Receive mock managed identities rp identity in CS

machi1990:
### What this PR does

Receive mock managed identities rp identity in CS

Jira: 
Link to demo recording:


cluster-service/deploy/openshift-templates/arohcp-service-template.yml
```diff
@@ -272,6 +280,10 @@ objects:
             objectName: ${FPA_CERT_NAME}
             objectType: secret
             objectAlias: firstPartyApplicationCertificateBundle
+          - |
```
miguelsorianod, line 283: Do we want to approach it having a single SecretProviderClass with N secrets or N SecretProviderClass ? I see different approaches depending on the MR.

miguelsorianod:
Having a quick look the approach seems good to me.

I'll defer to someone from the lifecycle team to give the ok to the approach as it involves templating and such

cluster-service/deploy/openshift-templates/arohcp-service-template.yml
machi1990: Both approaches are valid but I opted for this because I see no difference in terms of configuration between the `SecretProviderClass` I needed for this purpose and the firstPartyApp one.  I'll defer this question to @janboll @geoberle on how they see it.

cgiradkar:
lgtm <sup>[*](https://github.com/Azure/ARO-HCP/pull/899/#discussion_r1860471756)</sup>

PR #905, opened 2024-11-27
adapt ev2 bicepparam preprocessor to generate any()

geoberle:
### What this PR does

when a bicepparam file is preprocessed for EV2 with scope binding placeholders, any non-string placeholders needs to be wrapped in the `any()` bicep type function since the placeholder string would not match the expected type for the respective parameter.

this PR also cleans up several unused functions in the `ev2` package that have become obsolete because of pipeline preprocessing. specifically the `PreprocessFileForEV2SystemVars`, `PreprocessFileForEV2ScopeBinding` helper functions are of little use now that we have the `PrecompilePipelineForEV2` functions. they are also hard to use considering the various contexts file preprocessing involves (e.g. EV2 system variables allowed or not, scope binding vs service config, bicepparam any() or not)

this PR also removed the `templatize` flag for EV2 preprocessing. since we integrate templatize as a library, there is no need to expose the preprocessing on the CLI as RH will not use it

Jira: 
Link to demo recording:


tooling/templatize/pkg/ev2/mapping.go
```diff
@@ -7,23 +7,65 @@ import (
...
+// and returns a flattenedKey and replaceVar, both of which are the generated placeholder string.
+//
+// Example:
+//
+//	key := []string{"foo", "bar"}
+//	flattenedKey, replaceVar := NewDunderPlaceholders()(key, nil)
+//	// flattenedKey and replaceVar will both be "__foo_bar__"
+func NewDunderPlaceholders() PlaceholderGenerator {
+	return func(key []string, _ interface{}) (flattenedKey string, replaceVar string) {
+		flattenedKey = fmt.Sprintf("__%s__", strings.Join(key, "_"))
```
whober0521, line 26: can we change separator from "_" to "."?

whober0521 (changes requested):

tooling/templatize/pkg/ev2/mapping.go
geoberle: done. this has the benefit of making the generated placeholders more unique. while we have the occasional `_` in a var name, we don't have dots there.

tooling/templatize/testdata/zz_fixture_TestPreprocessFilebicepparam
```diff
@@ -1,10 +0,0 @@
-// copy from dev-infrastructure/configurations/region.bicepparam
```
janboll, line 1: is this superseded by some fixture?

janboll: Ah you deleted it cause we won't use it.

janboll (approved):

PR #909, opened 2024-11-28
prepare global resources pipeline

geoberle:
### What this PR does

* introduce a `global` section in the config schema
* refine config for image sync
* define a pipeline yaml for the global ACRs and image sync
* introduce a KV for imagesync secrets living in the global context
* rework custom token management role code
  * make it optional (quota issues in other tenants)
  * ... and fallback to a wider role in that case

Jira: 
Link to demo recording:


dev-infrastructure/global-acr-pipeline.yaml
```diff
@@ -0,0 +1,25 @@
+$schema: "pipeline.schema.v1"
+serviceGroup: Microsoft.Azure.ARO.HCP.Service.Infra
```
whober0521, line 2: Microsoft.Azure.ARO.HCP.Global.ACR

whober0521 (changes requested):

dev-infrastructure/global-acr-pipeline.yaml
geoberle: fixed

dev-infrastructure/configurations/image-sync.tmpl.bicepparam
whober0521, line 1: `{{ .svcAcrName }}.azurecr.io` should be returned value in bicep

whober0521 (changes requested):

PR #915, opened 2024-12-03
fix command to fit tooling in aro-pipeline

jonathan34c:
### What this PR does
Pair with this PR in msft ADO https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO-Pipelines/pullrequest/11349241?path=/pkg/ev2ra/rolloutparameters.go 
to change the command to string instead of array

Jira: 
Link to demo recording:

### DIFF
--- a/dev-infrastructure/mgmt-pipeline.yaml
+++ b/dev-infrastructure/mgmt-pipeline.yaml
@@ -22,7 +22,7 @@ resourceGroups:
       output: cs.msi.resourceID
   - name: enable-metrics
     action: Shell
-    command: ["scripts/enable-aks-metrics.sh"]
+    command: "scripts/enable-aks-metrics.sh"
     env:
     - name: RESOURCEGROUP
       configRef: mgmt.rg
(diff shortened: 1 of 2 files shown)


mociarain (commented):
This looks good but I'm worried about some of the other Shell actions in other parts of the repo?  https://github.com/Azure/ARO-HCP/blob/a40992b120504ea9c511335c76bd16f7b9947a36/frontend/pipeline.yaml#L10

How are these affected?

geoberle:
> This looks good but I'm worried about some of the other Shell actions in other parts of the repo?
> 
> https://github.com/Azure/ARO-HCP/blob/a40992b120504ea9c511335c76bd16f7b9947a36/frontend/pipeline.yaml#L10
> 
> How are these affected?

the schema for `command` will change in MSFT tooling from array to string, so all pipeline files need to be adapted and also our tooling needs change. @janboll do you have time to work on this?

geoberle:
complete adaption to string commands can be found here https://github.com/Azure/ARO-HCP/pull/923

PR #918, opened 2024-12-03
post-infra creation task for local CS development permissions

geoberle:
### What this PR does

the `local-cs-permissions` task in `dev-infrastructure/Makefile` will set up permissions into OIDC storage accounts, SVC KVs and MC KVs

Jira: 
Link to demo recording:


dev-infrastructure/scripts/kv-permissions.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+az role assignment create \
+    --role "Key Vault Secrets Officer" \
+    --assignee ${PRINCIPAL_ID} \
+    --scope ${KV_RESOURCE_ID} \
+    --only-show-errors
+
+az role assignment create \
+    --role "Key Vault Certificates Officer" \
```
machi1990, line 21: "Key Vault Certificates Officer: perform any action on the certificates of a key vault, excluding reading the secret and key portions, and managing permissions"

I wonder if we need "Key Vault Certificate User" which has this: Read entire certificate contents including secret and key portion. 

The part that confuses me is; are the certificate officer permission enough to be able to read certificates?

geoberle: certificate officer together with secret officer should be sufficient. but let me check

geoberle: you are right. i added the KV Certificage User as well

machi1990: Thank you.

janboll (approved):

PR #920, opened 2024-12-03
HCP Cluster Resource Deletion Cascading Subscription Delete

mbarnes:
### What this PR does

When a subscription state changes to "**Deleted**", the RP now triggers a deletion of all HCP clusters under the subscription as per the [Resource Provider Contract](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/subscription-lifecycle-api-reference.md#subscription-states).

Behind the scenes, this introduces the concept of "implicit" and "explicit" async operations:

-  An "implicit" async operation has an "Operation" item in Cosmos DB, but no status endpoint for ARM to poll.
-  An "explicit" async operation starts as an "implicit" operation.  The `Frontend.ExposeOperation` method enriches the "Operation" item with information necessary to make the status endpoint accessible to ARM, and adds appropriate async headers to an `http.ResponseWriter`.

Importantly, the backend pod does not distinguish between implicit and explicit async operations.  The sole purpose of an "implicit" async operation at the moment, which is only used for deletions, is for the backend to delete the "Resource" item in Cosmos DB after the actual resource is deleted.

Jira: [ARO-13321 - Implement Cascading Subscription Deletion](https://issues.redhat.com/browse/ARO-13321)

- This duplicates a few database iterator commits from [#883](https://github.com/Azure/ARO-HCP/pull/883), which is still outstanding.
- I'll add unit tests for this in a follow-up PR _after_ I convert our existing unit tests to use [gomock](https://github.com/uber-go/mock) for Cosmos DB operations.  To add unit tests now would just create extra work for myself in the conversion effort.
- I still need to document asynchronous operation mechanics in general and this "implicit" vs "explicit" concept will be part of it.  I've been holding off on writing documentation until I can make some (imo) necessary changes to our database design.  This is just to say I haven't forgotten about it.

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -507,9 +507,18 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
 		}
 	}
 
-	operationDoc, err := f.StartOperation(writer, request, doc, operationRequest)
+	operationDoc := database.NewOperationDocument(operationRequest, doc.Key, doc.InternalID)
+
+	err = f.dbClient.CreateOperationDoc(ctx, operationDoc)
 	if err != nil {
-		f.logger.Error(fmt.Sprintf("failed to write operation document: %v", err))
+		f.logger.Error(err.Error())
+		arm.WriteInternalServerError(writer)
+		return
+	}
+
+	err = f.ExposeOperation(writer, request, operationDoc.ID)
+	if err != nil {
+		f.logger.Error(err.Error())
 		arm.WriteInternalServerError(writer)
 		return
 	}
@@ -638,9 +647,18 @@ func (f *Frontend) ArmResourceDelete(writer http.ResponseWriter, request *http.R
 		}
 	}
 
-	operationDoc, err := f.StartOperation(writer, request, resourceDoc, operationRequest)
+	operationDoc := database.NewOperationDocument(operationRequest, resourceDoc.Key, resourceDoc.InternalID)
+
+	err = f.dbClient.CreateOperationDoc(ctx, operationDoc)
 	if err != nil {
-		f.logger.Error(fmt.Sprintf("failed to write operation document: %v", err))
+		f.logger.Error(err.Error())
+		arm.WriteInternalServerError(writer)
+		return
+	}
+
+	err = f.ExposeOperation(writer, request, operationDoc.ID)
+	if err != nil {
+		f.logger.Error(err.Error())
 		arm.WriteInternalServerError(writer)
 		return
 	}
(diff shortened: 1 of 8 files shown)


internal/database/document.go
```diff
@@ -94,13 +95,15 @@ type OperationDocument struct {
 	Error *arm.CloudErrorBody `json:"error,omitempty"`
 }
 
-func NewOperationDocument(request OperationRequest) *OperationDocument {
+func NewOperationDocument(request OperationRequest, externalID *arm.ResourceID, internalID ocm.InternalID) *OperationDocument {
```
mociarain, line 98: Would we consider splitting this functionality into `NewImplicitOperationDocument` and `NewExplicitOperationDocument`? This would help cement the concept and make it more visible. We could even extend the OperationDocument Type with Implicit/Explicit but that might be overkill and I don't have a good feeling if that's worth it. 

What do people think?

frontend/pkg/frontend/frontend.go
```diff
@@ -507,9 +507,18 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
 		}
 	}
 
-	operationDoc, err := f.StartOperation(writer, request, doc, operationRequest)
+	operationDoc := database.NewOperationDocument(operationRequest, doc.Key, doc.InternalID)
```
mociarain, line 510: Side question: Any objection or issue if I make a PR to change `Key` -> `ResourceId` in the `ResourceDocument` Type just for readability?

mociarain (approved):
LGTM just a few small question and a thought

internal/database/document.go
mbarnes: In terms of the operation document, what distinguishes explicit from implicit operations is certain fields being populated: namely `TenantID`, `ClientID` and most importantly `OperationID`.  I'm all for making the code clearer, though I'm not sure if a separate document type is the way I'd go.  The backend pod, for example, doesn't care about this explicit vs implicit distinction and treats all operations the same.  So I wouldn't want the backend to have to handle two document types.

Maybe an `OperationDocument.IsExplicit` method, that just returns whether the `OperationID` field is non-empty, would be sufficent?  (Or `OperationDocument.IsExposed` to align with [Frontend.ExposeOperation](https://github.com/Azure/ARO-HCP/blob/ebfaa9e6e79d3803db27605fb6e0b9ceb85368c5/frontend/pkg/frontend/operations.go#L82-L129)... I don't know, I'm playing fast and loose with terminology here.)

In any case, I don't know that we have a use case for such a method at the moment but I'll keep this in the back of my mind.

mociarain: That's fair. I don't have my head around it yet but hopefully the "right" answer appears at some stage :D

PR #922, opened 2024-12-03
convert cs deployment to helm chart

tony-schndr:
### What this PR does
Fixes https://issues.redhat.com/browse/ARO-11949, https://issues.redhat.com/browse/ARO-12801, & https://issues.redhat.com/browse/ARO-13258
Jira: 
Link to demo recording:


cluster-service/deploy/helm/templates/azure-runtime-config.configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: azure-runtime-config
+  namespace: '{{ .Release.Namespace }}'
+data:
+  config.json: |
+    {
+      "cloudEnvironment": "AzurePublicCloud",
+      "managedIdentitiesDataPlaneAudienceResource": "https://dummy.org",
```
machi1990, line 11: this should also be parameterised - see ARO-12801

config/config.yaml
```diff
@@ -140,6 +140,87 @@ clouds:
       clusterService:
         imageTag: 6157c57
         imageRepo: app-sre/uhc-clusters-service
+        azureOperatorsManagedIdentitiesConfig: |
+          controlPlaneOperatorsIdentities:
+            cloud-controller-manager:
+              minOpenShiftVersion: 4.17
```
janboll, line 146: Will this differ from Env to Env? Or will this see progressive updates? Just asking, cause this large key might make the config harder to read then (especially when thinking about later rollouts with more envs).

cluster-service/deploy/openshift-templates/arohcp.devenv.md
```diff
@@ -1,184 +0,0 @@
-# How to run CS in DemoMode on AKS
```
janboll, line 1: Is this obsolete?

config/config.yaml
tony-schndr: It will differ from env to env and it may see updates if CS needs to add an identity..  I don't like it but I was having a hard time templatizing this object and making it work nice with config.mk.  The problem I hit was that templatize was not creating a valid multi line string in the config.mk.  To work around this I treat it as a string and base64 encode it so that it becomes a single line in config.mk and decode it in the helm template.

cluster-service/deploy/openshift-templates/arohcp.devenv.md
tony-schndr: Nah, just deleted by mistake.  I added it back.

cluster-service/deploy/helm/values.yaml
```diff
@@ -0,0 +1,265 @@
...
+databasePassword: "TheBlurstOfTimes"
+
+# Name of the PostgreSQL database accessed.
+databaseName: "ocm-cs-db"
+
+# Host port
+databasePort: "5432"
+
+# The name of the managed identities data plane audience resource.
+managedIdentitiesDataPlaneAudienceResource: "https://dummy.org"
```
machi1990, line 262: One constraint we've to take onto account regarding this field's value is that we cannot check it in code for higher envs (stage/prod). So when we get to those envs, please make sure that this is not checked in. @tony-schndr

tony-schndr: What should the values for those environments be?

machi1990: We don't know, but whatever the value of them, they shouldn't be checked in any repo.

machi1990: To be specific, 
The dev and int envs value can be any dummy https url; this can be checked in the repo.
For stage and prod envs, these will be real value of the msi dataplane. They can't be checked in any repo.

cluster-service/deploy/helm/values.yaml
```diff
@@ -0,0 +1,265 @@
...
+maximumExpiration: "0"
+
+# Duration since cluster creation after which the first notification for stale cluster should be sent.
+firstStaleClusterNotification: "24h"  # 1 day
+
+# Duration since cluster creation after which the second notification for stale cluster should be sent.
+secondStaleClusterNotification: "600h"  # 25 days
+
+# Duration after which a stale cluster can be cleaned up.
+staleClusterAutocleanupWindow: "720h"  # 30 days
```
machi1990, line 25: These don't apply to aro-hcp. 
They can be deleted.

cluster-service/deploy/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,215 @@
...
+          mountPath: /configs/azure-runtime-config
+        - name: azure-operators-managed-identities-config
+          mountPath: /configs/azure-operators-managed-identities-config.yaml
+          subPath: azure-operators-managed-identities-config.yaml
+        command:
+        - /usr/local/bin/clusters-service
+        - serve
+        - --log-level={{ .Values.logLevel }}
+        - --namespace={{ .Release.Namespace }}
+        - --runtime-mode={{ .Values.runtimeMode }}
```
machi1990, line 143: We could hardcode the mode here as we are in the context of aro-hcp.
```suggestion
        - --runtime-mode=aro-hcp
```

cluster-service/deploy/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,215 @@
...
+        - --api-listener-address=:8000
+        - --metrics-listener-network=tcp
+        - --metrics-listener-address=:8080
+        - --healthcheck-listener-network=tcp
+        - --healthcheck-listener-address=:8083
+        - --environment={{ .Values.environment }}
+        - --backplane-url={{ .Values.backplaneURL }}
+        - --provision-shards-config=/secrets/shards/config
+        - --proxy-config-file=/configs/proxy/config.yaml
+        - --aws-sts-policy-directory=/configs/policies
```
machi1990, line 173: This flag isn't applicable for aro-hcp.
We can clean up later;
At the moment if we don't define it, CS starts normally but we'll get a 
```
2024-12-12T15:00:50Z ERROR loader/aws_sts_policy_loader.go:50 [opid='dc2b7851-64e2-4869-b48a-d4a21c5a407a'] Got error while trying to load policies: open : no such file or directory
gitlab.cee.redhat.com/service/uhc-clusters-service/pkg/aws/loader.(*AWSPolicyLoader).run
	/cs/pkg/aws/loader/aws_sts_policy_loader.go:50
gitlab.cee.redhat.com/service/uhc-clusters-service/pkg/aws/loader.(*AWSPolicyLoader).Start
	cs/pkg/aws/loader/aws_sts_policy_loader.go:43
gitlab.cee.redhat.com/service/uhc-clusters-service/cmd/clusters-service/servecmd.runServ
```
error message at start time, which can be confusing.

machi1990 (approved):
left cleanup comments that can be addressed later on.
Many of the flags may not directly apply to aro-hcp.

cluster-service/deploy/helm/values.yaml
```diff
@@ -0,0 +1,265 @@
...
+insecure: "false"
+
+# The URL of the gateway.
+gatewayUrl: "http://127.0.0.1:9090"
+
+# Level of access that an app can request to a resource.
+clientScopes: "openid"
+
+# Environment associated with this instance.
+environment: "aro-hcp-dev"
```
machi1990, line 58: @geoberle @tony-schndr A note on the environment name.
It shouldn't be longer than 10 chars.
Otherwise, it breaks the cluster's base domain prefix limit. 

For example "aro-hcp-dev" is 11 chars, this means that the cluster's base domain prefix can't be longer than 14 chars while CS allows for it to be <= 15 chars.

tony-schndr: I will address this in a followup.  We can enforce the length in the config json schema.

PR #923, opened 2024-12-04
pipeline.yaml shell command changes from array to string

geoberle:
### What this PR does

this change follows MSFTs pipeline.yaml format change. the `command` is executed as a shell script in the bash context.

Jira: 
Link to demo recording:

### DIFF
--- a/tooling/templatize/pkg/pipeline/shell.go
+++ b/tooling/templatize/pkg/pipeline/shell.go
@@ -31,10 +31,7 @@ func (s *Step) createCommand(ctx context.Context, dryRun bool, envVars map[strin
 }
 
 func buildBashScript(command string) string {
-	return "set -o errexit\n" +
-		"set -o nounset\n" +
-		"set -o pipefail\n" +
-		command
+	return fmt.Sprintf("set -o errexit -o nounset  -o pipefail\n%s", command)
 }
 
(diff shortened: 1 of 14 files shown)


tooling/templatize/pkg/pipeline/shell.go
```diff
@@ -13,26 +13,30 @@ import (
...
-		// if dry-run is not enabled, use the actual command or also if no dry-run command is defined
-		cmd = exec.CommandContext(ctx, s.Command[0], s.Command[1:]...)
 	}
+	cmd := exec.CommandContext(ctx, "/bin/bash", "-c", buildBashScript(scriptCommand))
 	cmd.Env = append(cmd.Env, utils.MapToEnvVarArray(envVars)...)
 	return cmd, false
 }
 
+func buildBashScript(command string) string {
+	return "set -o errexit\n" +
```
janboll, line 34: Did you consider strings.Builder? https://stackoverflow.com/questions/1760757/how-to-efficiently-concatenate-strings-in-go/47798475#47798475

janboll (dismissed):

tooling/templatize/pkg/pipeline/shell.go
geoberle: i've opted for simply fmt.Sprintf

PR #927, opened 2024-12-04
feat: remove explicit properties enabling specific cs provisioner steps

miguelsorianod:
Now the steps are enabled by default, and the properties that enable/disable some of them have been renamed.

### What this PR does

Jira: 
Link to demo recording:

### DIFF
--- a/internal/ocm/mock.go
+++ b/internal/ocm/mock.go
@@ -42,7 +42,7 @@ func NewMockClusterServiceClient() MockClusterServiceClient {
 func (mcsc *MockClusterServiceClient) GetConn() *sdk.Connection { panic("GetConn not implemented") }
 
 func (csc *MockClusterServiceClient) AddProperties(builder *cmv1.ClusterBuilder) *cmv1.ClusterBuilder {
-	additionalProperties := getDefaultAdditionalProperities()
+	additionalProperties := map[string]string{}
 	return builder.Properties(additionalProperties)
 }
 
--- a/internal/ocm/ocm.go
+++ b/internal/ocm/ocm.go
@@ -24,27 +24,6 @@ type ClusterServiceClientSpec interface {
 	DeleteCSNodePool(ctx context.Context, internalID InternalID) error
 }
 
-// Get the default set of properties for the Cluster Service
-func getDefaultAdditionalProperities() map[string]string {
-	// additionalProperties should be empty in production, it is configurable for development to pin to specific
-	// provision shards or instruct CS to skip the full provisioning/deprovisioning flow.
-	additionalProperties := map[string]string{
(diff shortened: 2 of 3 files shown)


miguelsorianod:
This MR can only be merged after the corresponding MR in CS has been merged and rolled out

internal/ocm/ocm.go
```diff
@@ -28,20 +28,8 @@ type ClusterServiceClientSpec interface {
...
-		// installing state in CS.
-		"provisioner_managedcluster_step_enabled": "true",
-
-		// Enable the provisioning and deprovisioning of ARO-HCP Node Pools. For now, if not set the provisioning
-		// and deprovisioning of day 2 ARO-HCP Node Pools will not be performed on the Management Cluster.
-		"np_provisioner_provision_enabled":   "true",
-		"np_provisioner_deprovision_enabled": "true",
-	}
+	additionalProperties := map[string]string{}
+
```
mbarnes, line 32: I suggest just deleting this whole function and declaring a `map[string]string{}` where `getDefaultAdditionalProperities` is called.

mbarnes (changes requested):
Requested a change just to leave the code in a slightly cleaner state.

internal/ocm/ocm.go
miguelsorianod: That's fine for me if you want. The reason I kept the function is to have it there in case in the future we want to add default properties, to have it contained in that function.

Let me know if you still prefer what you suggested.

mbarnes: Understood.  If there's a need in the future for always-present additional properties we can reinstate that function, but I don't foresee it.  So I'd still prefer it removed, please.

miguelsorianod: 👍 I will change it then

miguelsorianod: updated

internal/ocm/mock.go
```diff
@@ -42,7 +42,7 @@ func NewMockClusterServiceClient() MockClusterServiceClient {
 func (mcsc *MockClusterServiceClient) GetConn() *sdk.Connection { panic("GetConn not implemented") }
 
 func (csc *MockClusterServiceClient) AddProperties(builder *cmv1.ClusterBuilder) *cmv1.ClusterBuilder {
-	additionalProperties := getDefaultAdditionalProperities()
+	additionalProperties := map[string]string{}
```
miguelsorianod, line 45: @mbarnes I also changed it here

mbarnes (approved):
LGTM, thanks!

miguelsorianod:
We are going to rollout the changes here together with CS Image update in https://github.com/Azure/ARO-HCP/pull/961. I am closing this issue and we moved the commits in here to the other MR.

PR #936, opened 2024-12-05
api: Make PlatformProfile.networkSecurityGroupId optional

mbarnes:
### What this PR does

Subject says it all.  Updated `ClusterBuilder` logic and unit tests.

Jira: [ARO-13341 - Updates in Frontend and API to make NSG Optional](https://issues.redhat.com/browse/ARO-13341)
Link to demo recording: 

- You can disregard the 1st commit.  That's just updating generated code with the latest `autorest` version. There's several API-related pull requests with this commit now ([#950](https://github.com/Azure/ARO-HCP/pull/950), [#951](https://github.com/Azure/ARO-HCP/pull/951), and this) because I don't know which one will merge first.
- I still need to add static validation to this and other resource ID fields.  That will be a separate effort.


bennerv (approved):
lgtm, but this won't work until CS relaxes the restriction so we're a bit early.

PR #949, opened 2024-12-09
Use the correct APIVersion

mociarain:
I found this when doing something else but I'm a bit confused by it. 

- I found the error `no kind "Template" is registered for version "v1" in scheme` when deploying `cluster-service` to the service cluster.
- The docs do indeed point to it being `template.openshift.io/v1` not `v1` [ref](https://docs.openshift.com/container-platform/4.17/rest_api/template_apis/template-apis-index.html#template-template-openshift-iov1)

I'm assuming this was working for others and there's something wrong but my setup but maybe not?

### DIFF
--- a/cluster-service/deploy/integration/cluster-service-namespace.yaml
+++ b/cluster-service/deploy/integration/cluster-service-namespace.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: v1
+apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
   name: cluster-service-admin


mbarnes (approved):
Seems correct to me based on [documentation](https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/template_apis/template-template-openshift-io-v1#template-template-openshift-io-v1).  Might be that it happens to work currently based on some fallback behavior in Helm or OpenShift, which if true we shouldn't be relying on.

mociarain:
> can you elaborate?

This error occurs when running `make deploy-namespace-template` and it looks like the error is coming from `oc` when it's trying to parse the template. Updating the API version lets me render the template and deploy the resources.

geoberle:
> deploy-namespace-template

ah i see. i think i've seen a discussion about this on slack. `oc` version related.
thanks for fixing.
LGTM

geoberle (approved):

SudoBrendan (approved):

PR #950, opened 2024-12-09
api: Define OperatorsAuthentication model

mbarnes:
### What this PR does

Adds `PlatformProfile.operatorsAuthentication` field to the Azure API, structured to match the [azure.operators_authentication](https://api.openshift.com/#/default/post_api_clusters_mgmt_v1_clusters) OCM schema.  (With minor tweaks to eliminate redundancy in field names, and the limitation that we can't use the built-in [Record\<Element\>](https://typespec.io/docs/language-basics/built-in-types/) TypeSpec type in Azure APIs for the operator maps, so instead they're arrays.)

Jira: [ARO-10911 - Frontend API Changes to support managed identities](https://issues.redhat.com/browse/ARO-10911)
Link to demo recording: 

* You can disregard the 1st commit. That's just updating generated code with the latest `autorest` version.
* There is no plumbing to Cluster Service yet.  I want to get initial sign-off on the API changes first.
* Visibility flags may need to be tweaked.  Obviously the operator arrays should be replaceable for OpenShift versions beyond the initial install, but I'm not sure what other fields we should allow to be patched.


api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -346,6 +350,77 @@ union OutboundType {
...
+  /** The set of Azure User-Assigned Managed Identities leveraged for the Data Plane
+   * operators of the cluster. The set of required managed identities is dependent on the
+   * Cluster's OpenShift version. */
+  @OpenAPI.extension("x-ms-identifiers", ["operatorName"])
+  dataPlaneOperators: DataPlaneOperatorsElement[];
+
+  /** The Managed Identities Data Plane Identity URL associated with the cluster. It is the URL
+   * that will be used to communicate with the Managed Identities Resource Provider (MI RP). */
+  @visibility("create", "read")
+  dataPlaneIdentityUrl: url;
```
bennerv, line 378: Don't need this.  It's sent in headers

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -346,6 +350,77 @@ union OutboundType {
...
+}
+
+/** Represents the information related to Azure User-Assigned managed identities needed
+ * to perform Operators authentication based on Azure User-Assigned Managed Identities */
+model ManagedIdentitiesProfile {
+  /** The set of Azure User-Assigned Managed Identities leveraged for the Control Plane
+   * operators of the cluster. The set of required managed identities is dependent on the
+   * Cluster's OpenShift version. */
+  @OpenAPI.extension("x-ms-identifiers", ["operatorName"])
+  controlPlaneOperators: ControlPlaneOperatorsElement[];
```
bennerv, line 367: We want to define these as a map type because otherwise replacement is difficult.  it'll go through additional scrutiny but we already have the use case built for classic.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -346,6 +350,77 @@ union OutboundType {
...
+  @visibility("create", "read")
+  serviceManagedIdentity: UserAssignedManagedIdentity;
+}
+
+/** Pairs a control plane operator name with its user assigned identity */
+model ControlPlaneOperatorsElement {
+  /** The control plane operator name */
+  operatorName: string;
+
+  ...UserAssignedManagedIdentity;
```
bennerv, line 391: the client-id and principal-id will be duplicated in both the `identity{}` field and this field if we choose to specify it here. 

```
properties: {
    "managedIdentityProfile": {
        "controlPlaneOperators": [
             {
             "operatorName": "csi-disk-driver", 
             "resourceId": "/subscriptions/.../Microsoft.Identity/managedIdentity/csi-disk-driver"
             "principalId": "csi-disk-driver",
             "clientId": "csi-disk-driver"
             }, 
             {...}
         ]
    }   
},
identity: {
    "/subscriptions/.../Microsoft.Identity/managedIdentity/csi-disk-driver": {
        "principalId": "csi-disk-driver",
        "clientId": "csi-disk-driver"
    }, 
    "/subscriptions/.../csi-file-driver": {
        "principalId": "csi-file-driver",
        "clientId": "csi-file-driver"
    }
}

File: api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp (hunk 1 above)
Comment (miguelsorianod) [reply api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp:378]: +1, unless we want to expose that to the end-users

File: api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp (hunk 2 above)
Comment (mbarnes) [reply api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp:367]: The way to do that in TypeSpec, as I understand, is with `Record<Type>` which defines a map with string keys and `Type` values.  But the TypeSpec compiler rejects it with
```
error @azure-tools/typespec-azure-resource-manager/arm-no-record: Model properties or
operation parameters should not be of type Record. ARM requires Resource provider teams
to define types explicitly.
```
so I'll have to look for a way to override it.

File: api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp (hunk 3 above)
Comment (mbarnes) [reply api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp:391]: Ah, I had no idea what the top-level `identity` section would contain.  In that case we can have both data and control plane operators be a map of `operatorName -> resourceId` where `resourceId` is a lookup key for the `identity` section.

**Comment by mbarnes** [comment]
> I described and reasoned why in the CS API we intentionally kept parts of the names that initially might seem "redundant"

I'm neutral on naming and will change it based on consensus, but FTR I felt the purpose of the operator maps were clear enough from the enclosing scopes.
```
properties: {
  platform: {
    operatorsAuthentication: {
      managedIdentities: {
        controlPlaneOperators: {...}
        dataPlaneOperators: {...}
      }
    }
  }
}
```

File: api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp (hunk 3 above)
Comment (mbarnes) [reply api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp:391]: Also, maybe the `serviceManagedIdentity` field just needs to be a resource ID lookup key as well?

File: api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp (hunk 2 above)
Comment (mbarnes) [reply api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp:367]: Figured it out...
```
#suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "operator name to resource ID pairings"
controlPlaneOperators: Record<UserAssignedIdentityResourceId>;
```

mbarnes: Fixed in a `fixup!` commit that I'll eventually squash.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mbarnes: Looks like it's `x-ms-identity-url` from step 6 of [How to onboard Managed Identities for your resource type](https://eng.ms/docs/products/arm/rbac/managed_identities/msionboardingsteps).
Please correct me if I'm wrong, the MS documentation for this is a maze.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mbarnes: Just to call out here, the RP does not currently parse the top-level `identity` section at all so we'll need to get that implemented immediately in addition to the plumbing needed for this new API.

SudoBrendan (approved):
I think this is clear on the various kinds of auth required and what fields should be provided for each. We should probably get BU feedback on names just to be sure, unless that's tracked somewhere I'm unaware of?

PR #951, opened 2024-12-09
api: Prefer built-in type "url" over "string" for URL fields

mbarnes:
### What this PR does

While working on [#950](https://github.com/Azure/ARO-HCP/pull/950) I realized TypeSpec has a [built-in type for URL strings](https://typespec.io/docs/language-basics/built-in-types/#string-types), which seemed appropriate to use for certain string fields in the API.

This does not change the generated Go code at all.  All it does is add a "format: uri" qualifier to these string fields in `openapi.json`, which I don't even know if ARM cares about but why not.

Jira: None, just tweaking.
Link to demo recording: 

* You can disregard the 1st commit. That's just updating generated code with the latest autorest version.  There's several API-related pull requests with this commit now ([#936](https://github.com/Azure/ARO-HCP/pull/936), [#950](https://github.com/Azure/ARO-HCP/pull/950), and this) because I don't know which one will merge first.


mociarain (approved):
LGTM

bennerv (approved):

PR #953, opened 2024-12-10
Use Job ID

janboll:
### What this PR does

Jira: 
Link to demo recording:


tooling/templatize/internal/end2end/e2e.go
```diff
@@ -91,15 +91,17 @@ func newE2E(tmpdir string) e2eImpl {
 }
 
 func (e *e2eImpl) UseRandomRG() func() error {
-
+	rgSuffx := ""
+	if jobID := os.Getenv("GITHUB_JOB_ID"); jobID != "" {
```
geoberle, line 95: I suggest to use `JOB_ID` as we don't really care that it is a GH one.

geoberle (approved):

PR #958, opened 2024-12-10
Fix config schema for istio and region

jonathan34c:
### What this PR does
Add istio and region schema to the config file.
Pipeline: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=110131012&view=results

Jira: 
Link to demo recording:

### DIFF
--- a/config/config.msft.yaml
+++ b/config/config.msft.yaml
@@ -34,7 +34,6 @@ defaults:
       istioctlUrl: "https://istio.io/downloadIstio"
       istioctlVersion: "1.23.1"
       tag: "prod-stable"
-      currentVersion: "asm-1-22"
       newVersion: ""
 
   # MGMT cluster specifics
--- a/config/config.msft.yaml
+++ b/config/config.msft.yaml
@@ -229,4 +229,3 @@ clouds:
                 - name: "uksouth"
                   settings:
                   mgmt: 3
-              
--- a/config/config.msft.yaml
+++ b/config/config.msft.yaml
@@ -221,7 +221,7 @@ clouds:
           # lets create this MSI manually for the time being and automate soon
           # but we should use the MSI name as an input and not the resource ID of the MSI
           aroDevopsMsiId: '/subscriptions/5299e6b7-b23b-46c8-8277-dc1147807117/resourcegroups/global-shared-resources/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aroint-int-public-oidc'
-          
+
           # Geographies
           geographies:
             - name: "United Kingdom"
@@ -229,4 +229,4 @@ clouds:
                 - name: "uksouth"
                   settings:
                   mgmt: 3
-                  
+              
--- a/config/config.msft.yaml
+++ b/config/config.msft.yaml
@@ -30,6 +30,12 @@ defaults:
     etcd:
       kvName: arohcp-etcd-{{ .ctx.regionShort }}
       kvSoftDelete: true
+    istio:
+      istioctlUrl: "https://istio.io/downloadIstio"
+      istioctlVersion: "1.23.1"
+      tag: "prod-stable"
+      currentVersion: "asm-1-22"
+      newVersion: ""
 
   # MGMT cluster specifics
   mgmt:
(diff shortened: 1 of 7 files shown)


config/config.msft.yaml
```diff
@@ -30,6 +30,12 @@ defaults:
     etcd:
       kvName: arohcp-etcd-{{ .ctx.regionShort }}
       kvSoftDelete: true
+    istio:
+      istioctlUrl: "https://istio.io/downloadIstio"
+      istioctlVersion: "1.23.1"
+      tag: "prod-stable"
+      currentVersion: "asm-1-22"
```
whober0521, line 37: duplicate with line 16

chiac831 (changes requested):

config/config.msft.yaml
```diff
@@ -30,6 +30,11 @@ defaults:
     etcd:
       kvName: arohcp-etcd-{{ .ctx.regionShort }}
       kvSoftDelete: true
+    istio:
+      istioctlUrl: "https://istio.io/downloadIstio"
```
stevekuznetsov, line 34: `istio` is already part of the parent key - this can simply be `url`

config/config.schema.json
```diff
@@ -620,7 +620,27 @@
           },
           "subscription": {
             "type": "string"
-          }
+          },
+          "istio": {
```
janboll, line 624: With the three different version attributes, does it makes sense to add some description field?

stevekuznetsov: Honestly, unless proven otherwise, description field should be expected for all IMO

config/config.msft.yaml
```diff
@@ -215,3 +221,11 @@ clouds:
           # lets create this MSI manually for the time being and automate soon
           # but we should use the MSI name as an input and not the resource ID of the MSI
           aroDevopsMsiId: '/subscriptions/5299e6b7-b23b-46c8-8277-dc1147807117/resourcegroups/global-shared-resources/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aroint-int-public-oidc'
+
+          # Geographies
+          geographies:
```
stevekuznetsov, line 226: @geoberle  likely want to make sure this structure is how we want to hold data for each region

config/config.msft.yaml
```diff
@@ -215,3 +221,11 @@ clouds:
...
+            - name: "United Kingdom"
+              regions:
+                - name: "uksouth"
+                  settings:
+                  mgmt: 3
```
stevekuznetsov, line 231: Is this mis-indented?

config/config.msft.yaml
```diff
@@ -30,6 +30,12 @@ defaults:
     etcd:
       kvName: arohcp-etcd-{{ .ctx.regionShort }}
       kvSoftDelete: true
+    istio:
+      url: "https://istio.io/downloadIstio"
+      istioctlVersion: "1.23.1"
+      tag: "prod-stable"
+      version: "asm-1-22"
+      newVersion: ""
```
stevekuznetsov, line 38: A quick overview of why we need this kind of value would be useful. As I remember it there's some namespace selection going on, but it would be good to confirm.

I don't know how this value gets corrupted when we have concurrent upgrades going on. If we had an upgrade from 1 to 2 running but failed, then we tried to upgrade to 3, we'd need to have config that specified 1->3, since 2->3 wouldn't pick up regions that still were on 1? I don't know if we want that.

stevekuznetsov: Actually, I don't see the old version used for anything in the script - https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/hcp?path=/ev2/svc/istio/archive/istio.sh&version=GBmain&line=13&lineEnd=14&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

Are we sure we need both parameters?

config/config.msft.yaml
geoberle: no the config.schema.json describes the result after interpreting a config.yaml in a cloud/env/region. if we want to support geographies as well, we need to make that part of the config.yaml structure itself

config/config.msft.yaml
jonathan34c: so like i wrote it in the [Readme](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/hcp?path=/ev2/svc/istio/README.md&_a=preview&anchor=istio-upgrade-procedue.)
So basically 
1. we need to add istio new version and old version in to SVC infra pipeline. and run it( so svc cluster would have both version in the cluster)
2. Run the istio pipeline to replace the version.
3. Update the svc config to have the new version, and run the svc infra again. 

Sorry for the confusion on having the old version in the script. it was a legacy code I forgot to delete from previous version of the shell script. I will delete it.

config/config.msft.yaml
jonathan34c: Gerd said he will use this version in the future.

stevekuznetsov: Can't this PR just do that? Just need to change the templates to use this one.

config/config.schema.json
jonathan34c: description added.

PR #979, opened 2024-12-12
Move istio to svc cluster rollout

jonathan34c:
### What this PR does
Move istio from hcp to ARO-HCP. Will run istio sh in every svc deployment. But will only upgrade if there's new version 
https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=110334014&view=results

https://dev.azure.com/msazure/AzureRedHatOpenShift/_build/results?buildId=110608387&view=results

Jira: 
Link to demo recording:


mociarain (approved):
LGTM

config/Makefile
```diff
@@ -15,7 +15,7 @@ detect-change: materialize
 		echo "Please review the diffs below:\n\n"; \
 		echo "$$diff_output"; \
 		echo "\n\n===================================================="; \
-		echo "\n\nOnce you reviewed the changes and consider them meaningful, commit them."; \
+		echo "\n\nOnce you reviewed the changes and consider them meaningful, commit them by running "make -C config/ materialize"."; \
```
stevekuznetsov, line 18:
```suggestion
		echo "\n\nOnce you reviewed the changes and consider them meaningful, update them by running "make -C config/ materialize" and commit your changes."; \
```

config/config.msft.yaml
```diff
@@ -30,6 +29,12 @@ defaults:
     etcd:
       kvName: arohcp-etcd-{{ .ctx.regionShort }}
       kvSoftDelete: true
+    istio:
+      url: "https://istio.io/downloadIstio"
+      istioctlVersion: "1.23.1"
+      tag: "prod-stable"
+      newVersion: ""
```
stevekuznetsov, line 36: IIRC @geoberle had suggested `targetVersion` and a versions array - are you making that change here?

config/config.schema.json
```diff
@@ -664,7 +661,32 @@
...
+              "istioctlVersion": {
+                "description": "Latest version of istioctl. This is used to download the istioctl",
+                "type": "string"
+              },
+              "tag": {
+                "description": "The revision tag used in istioctl",
+                "type": "string"
+              },
+              "version": {
+                "description": "The current istio version in the aks cluster",
```
stevekuznetsov, line 681:
```suggestion
                "description": "The current istio version in the AKS cluster",
```

config/config.schema.json
```diff
@@ -664,7 +661,32 @@
           },
           "subscription": {
             "type": "string"
-          }
+          },
+          "istio": {
+            "type": "object",
+            "properties": {
+              "url": {
+                "description": "The download Url for istioctl",
```
stevekuznetsov, line 669: Are we moving to an explicit link to a release artifact and checking SHASUM like @weinong mentioned?

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+echo "********** ISTIO Upgrade Started **************"
+# Get the current istio and check if it match target version
+export CURRENTVERSION=$(kubectl get pods -n aks-istio-system | tail -n +2 | tail -n 1 | cut -d '-' -f 2,3,4)
```
stevekuznetsov, line 7: General statement - never use string manipulation on structured data. e.g. here, please `-o json` and filter with `jq` or use `-o jsonpath`. Select what objects you want with `--selector`.

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,36 @@
...
+if [ "$CURRENTVERSION" == "$TARGET_VERSION" ] || [ -z "$TARGET_VERSION" ]; then
```
stevekuznetsov, line 8: nit: please install `shellcheck`. It will explain why `[` is better replaced with `[[`.

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,36 @@
...
+
+export NEWVERSION="$TARGET_VERSION"
+echo "********** Istio UpGrade Started with version ${NEWVERSION} **************"
+
+# Overwrite the tag with new istio version
+istioctl tag set "$TAG" --revision "${NEWVERSION}" -i aks-istio-system --overwrite
+
+# Get the namespaces with the label istio.io/rev=$TAG
+export namespaces=$(kubectl get namespaces --selector=istio.io/rev="$TAG") 
+for ns in $namespaces; do
```
stevekuznetsov, line 27: What we described in the last call was that we need to operate over all Pods in the cluster - regardless of their namespace - and restart them if their Istio sidecars are running the previous version of Istio. For pods owned by a ReplicaSet (owned by a Deployment, or StatefulSet), restart the owning resource using the appropriate API like rollouts.

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+echo "********** ISTIO Upgrade Started **************"
+# Get the current istio and check if it match target version
+export CURRENTVERSION=$(kubectl get pods -n aks-istio-system | tail -n +2 | tail -n 1 | cut -d '-' -f 2,3,4)
+if [ "$CURRENTVERSION" == "$TARGET_VERSION" ] || [ -z "$TARGET_VERSION" ]; then
+    echo "Istio is using Target Version. Exiting script."
```
stevekuznetsov, line 9: This check is superfluous - if there are no pods running the previous version of Istio to restart, we can exit implicitly.

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,36 @@
...
+istioctl tag set "$TAG" --revision "${NEWVERSION}" -i aks-istio-system --overwrite
+
+# Get the namespaces with the label istio.io/rev=$TAG
+export namespaces=$(kubectl get namespaces --selector=istio.io/rev="$TAG") 
+for ns in $namespaces; do
+    # Get all deployments in the namespace
+    deployments=$(kubectl get deployments -n "$ns" -o jsonpath='{.items[*].metadata.name}')
+    # Iterate over each deployment and restart it
+    for deployment in $deployments; do
+        kubectl rollout restart deployment "$deployment" -n "$ns"
```
stevekuznetsov, line 32: Also, after each `rollout restart` or Pod restart, we should ensure that the new Pods start successfully (become ready and healthy and do not start CrashLoopBackOff). Either handle this here or ensure it's captured as a high priority card for the next set of changes.

config/config.schema.json
jonathan34c: Im not hearing anything yet. i think we can keep the link here for now and update to prefer method in the future.

stevekuznetsov: This was in the Teams chat - it should be as easy as using `wget` on the literal release file you need (which the script also does, you can just inspect that) and `md5sum` to check with the published value.

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+echo "********** ISTIO Upgrade Started **************"
+# Followed this guide for istio upgrade https://learn.microsoft.com/en-us/azure/aks/istio-upgrade
+# Get the current istio and check if it match target version
+export CURRENTVERSION=$(kubectl get pods -n aks-istio-system -o jsonpath='{.items[*].metadata.name}' | tr ' ' '\n' | tail -n +2 | head -n 1 | cut -d '-' -f 2,3,4)
```
stevekuznetsov, line 8: The name of the pod is not a durable way to check for the version. Please look into `istioctl tag set` command - if it is idempotent, we do not need to have this check for versions whatsoever.

config/Makefile
```diff
@@ -15,7 +15,7 @@ detect-change: materialize
 		echo "Please review the diffs below:\n\n"; \
 		echo "$$diff_output"; \
 		echo "\n\n===================================================="; \
-		echo "\n\nOnce you reviewed the changes and consider them meaningful, commit them."; \
+		echo "\n\nOnce you reviewed the changes and consider them meaningful, commit them by running "make -C config/ materialize" and commit your changes."; \
```
stevekuznetsov, line 18:
```suggestion
		echo "\n\nOnce you reviewed the changes and consider them meaningful, update them by running "make -C config/ materialize" and commit your changes."; \
```

config/config.msft.yaml
```diff
@@ -30,6 +29,12 @@ defaults:
     etcd:
       kvName: arohcp-etcd-{{ .ctx.regionShort }}
       kvSoftDelete: true
+    istio:
+      url: "https://github.com/istio/istio/releases/download"
+      istioctlVersion: "1.23.1"
+      tag: "prod-stable"
+      targetVersion: ""
+      version: "asm-1-22"
```
stevekuznetsov, line 37: Doesn't this need to be an array?

stevekuznetsov: More to the point - we had talked about having an array here of the versions of the istio control plane installed on the cluster. So the upgrade steps would look like

```
# step 0, current state
istio:
  versions: ["asm-1-22"]
  targetVersion: "asm-1-22"

# step 1, begin upgrade by installing new control plane and using it
istio:
  versions: ["asm-1-22", "asm-1-23"]
  targetVersion: "asm-1-23"

# step 2, clean up previous
istio:
  versions: ["asm-1-23"]
  targetVersion: "asm-1-23"
```

config/config.yaml
```diff
@@ -28,6 +27,12 @@ defaults:
     etcd:
       kvName: arohcp-etcd-{{ .ctx.regionShort }}
       kvSoftDelete: true
+    istio:
+      url: "https://github.com/istio/istio/releases/download"
```
stevekuznetsov, line 31: Is there ever a case where we will download from some other GitHub release page? I don't think so - you are likely good to remove this as a parameter and simply hard-code it into the script.

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,77 @@
...
+cd istio-"${ISTIOCTL_VERSION}"
+export PATH=$PWD/bin:$PATH
+echo "=========================================================================="
+
+NEWVERSION="$TARGET_VERSION"
+echo "********** Istio UpGrade Started with version ${NEWVERSION} **************"
+
+istioctl tag set "$TAG" --revision "${NEWVERSION}" --istioNamespace aks-istio-system --overwrite
+# Get the namespaces with the label istio.io/rev=$TAG
+namespaces=$(kubectl get namespaces --selector=istio.io/rev="$TAG" -o jsonpath='{.items[*].metadata.name}' | xargs -n1 echo)
```
stevekuznetsov, line 43: Why do you need `xargs -n1 echo`?

stevekuznetsov: It looks like you were trying to get newlines between names? I don't think you need to. Shell constructs will happily do a loop over space-delimited data, so you can simply write:

```sh
for namespace in $( kubectl get namespaces --selector=istio.io/rev="$TAG" -o jsonpath='{.items[*].metadata.name}' ); do
    
done
```

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,77 @@
...
+
+for ns in $namespaces; do
+    pods=$(kubectl get pods -n "$ns" -o jsonpath='{.items[*].metadata.name}'| xargs -n1 echo)
```
stevekuznetsov, line 46: Instead of making many `kubectl` client calls to the server, you can simply record the data and operate over it:

```sh
for namespace in $( kubectl get namespaces --selector=istio.io/rev="$TAG" -o jsonpath='{.items[*].metadata.name}' ); do
    pods="$( kubectl get pods --namespace "${namespace}" -o json )"
    for pod in "$( jq <<<"${pods}" --raw-output --arg NEWVERSION "${NEWVERSION}" '.items[] | select(.metadata.annotations.sidecar\.istio\.io/status | fromjson.revision != $NEWVERSION) | .metadata.name' )"; do
        owner_kind="$( jq <<"${pods}" --raw-output --arg NAME "${pod}" '.items[] | select(.metadata.name == $NAME) | .metadata.ownerReferences[0].kind' )"
        # etc
    done
done
```

stevekuznetsov: I may have messed up the shell script, they are very hard to write correctly without a REPL. Let's pair on this one. Saving the output of `get pods -o json` in a file also lets your script run super duper fast when you are testing this locally, as listing all pods might take some time on the server.

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,106 @@
...
+    for pod in $(jq <<<"${pods}" --raw-output --arg NEWVERSION "${NEWVERSION}" '.items[] | select(.metadata.annotations["sidecar.istio.io/status"] | fromjson.revision != $NEWVERSION) | .metadata.name'); do
+        owner_kind=$(jq <<<"${pods}" --raw-output --arg NAME "${pod}" '.items[] | select(.metadata.name == $NAME) | .metadata.ownerReferences[0].kind')
+        owner_name=$(jq <<<"${pods}" --raw-output --arg NAME "${pod}" '.items[] | select(.metadata.name == $NAME) | .metadata.ownerReferences[0].name')        
+            case "$owner_kind" in
+                "ReplicaSet")
+                    deployment=$(kubectl get replicaset "$owner_name" -n "$namespace" -o jsonpath='{.metadata.ownerReferences[0].name}')
+                    if [[ -n "$deployment" ]]; then
+                        echo "in ReplicaSet restart deployment $deployment"
+                        kubectl rollout restart deployment "$deployment" -n "$namespace"
+                        continue 2
```
stevekuznetsov, line 86: TIL `continue 2`

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,106 @@
...
+
+NEWVERSION="$TARGET_VERSION"
+echo "********** Istio Upgrade Started with version ${NEWVERSION} **************"
+
+istioctl tag set "$TAG" --revision "${NEWVERSION}" --istioNamespace ${ISTIO_NAMESPACE} --overwrite
+for namespace in $(kubectl get namespaces --selector=istio.io/rev="$TAG" -o jsonpath='{.items[*].metadata.name}'); do
+    echo "in namespace $namespace"
+    pods="$(kubectl get pods --namespace "${namespace}" -o json)"
+    for pod in $(jq <<<"${pods}" --raw-output --arg NEWVERSION "${NEWVERSION}" '.items[] | select(.metadata.annotations["sidecar.istio.io/status"] | fromjson.revision != $NEWVERSION) | .metadata.name'); do
+        owner_kind=$(jq <<<"${pods}" --raw-output --arg NAME "${pod}" '.items[] | select(.metadata.name == $NAME) | .metadata.ownerReferences[0].kind')
```
stevekuznetsov, line 78: Will this error if there are no `ownerReferences`?

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,106 @@
...
+        owner_name=$(jq <<<"${pods}" --raw-output --arg NAME "${pod}" '.items[] | select(.metadata.name == $NAME) | .metadata.ownerReferences[0].name')        
+            case "$owner_kind" in
+                "ReplicaSet")
+                    deployment=$(kubectl get replicaset "$owner_name" -n "$namespace" -o jsonpath='{.metadata.ownerReferences[0].name}')
```
stevekuznetsov, line 82: This is a RS, not a Deployment?

stevekuznetsov: Usually Deployment owns RS

geoberle: fixed in the recent commit

dev-infrastructure/scripts/istio.sh
geoberle: adressed. we have now handling for owned and bare pods

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,111 @@
...
+    exit 1
+fi
+
+NEWVERSION="$TARGET_VERSION"
+echo "********** Istio Upgrade Started with version ${NEWVERSION} **************"
+
+istioctl tag set "$TAG" --revision "${NEWVERSION}" --istioNamespace ${ISTIO_NAMESPACE} --overwrite
+for namespace in $(kubectl get namespaces --selector=istio.io/rev="$TAG" -o jsonpath='{.items[*].metadata.name}'); do
+    echo "in namespace $namespace"
+    # bare pods
```
stevekuznetsov, line 76: if nothing owns them, they won't get recreated, right? yikes! this just nukes them out of prod

stevekuznetsov: probably good to fix in our follow-ups

geoberle: i'm just wondering what would would create such pods in our namespaces without having clear ownership. maybe we should just fail the script and the pipeline if there are any

dev-infrastructure/scripts/istio.sh
```diff
@@ -0,0 +1,111 @@
...
+    for pod in $(kubectl get pods --namespace "${namespace}" -o json | jq -r --arg NEWVERSION "${NEWVERSION}" '.items[] | select(.metadata.annotations["sidecar.istio.io/status"] | fromjson.revision != $NEWVERSION) | select(.metadata.ownerReferences | length == 0) | .metadata.name'); do
+        echo "recycle pod $pod"
+        kubectl delete pod "$pod" -n "$namespace"
+    done
+    # pods with owners
+    currentDeloyment=""
```
stevekuznetsov, line 82: probably want to be more rigorous with this in the future

stevekuznetsov (commented):
Seems reasonable, let's merge it! @jonathan34c  could you please squash your commits into something sensible? Up to you and @geoberle if you want to split authorship, etc.

PR #985, opened 2024-12-13
frontend: Refetch resource document after updating

mbarnes:
### What this PR does

Noticed while testing Tags field patching that the PATCH response did not contain the updated Tags. I had to GET the resource again to see the correct set of Tags.

Jira: During testing of [ARO-10911 - API: HCP supports managed identities](https://issues.redhat.com/browse/ARO-10911), though not really related
Link to demo recording:

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -513,6 +513,13 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
 		if updated {
 			f.logger.Info(fmt.Sprintf("document updated for %s", resourceID))
 		}
+		// Get the updated resource document for the response.
+		doc, err = f.dbClient.GetResourceDoc(ctx, resourceID)
+		if err != nil {
+			f.logger.Error(err.Error())
+			arm.WriteInternalServerError(writer)
+			return
+		}
 	}
 
 	responseBody, err := marshalCSCluster(csCluster, doc, versionedInterface)
--- a/frontend/pkg/frontend/node_pool.go
+++ b/frontend/pkg/frontend/node_pool.go
@@ -242,6 +242,13 @@ func (f *Frontend) CreateOrUpdateNodePool(writer http.ResponseWriter, request *h
 		if updated {
 			f.logger.Info(fmt.Sprintf("document updated for %s", resourceID))
 		}
+		// Get the updated resource document for the response.
+		doc, err = f.dbClient.GetResourceDoc(ctx, resourceID)
+		if err != nil {
+			f.logger.Error(err.Error())
+			arm.WriteInternalServerError(writer)
+			return
+		}
 	}
 
 	responseBody, err := marshalCSNodePool(csNodePool, doc, versionedInterface)


mociarain:
That seems broken and is very very annoying. 

- I assume that conditionally re-fetching just on the tags update is too much work but did you consider adding some note to highlight that we could stop re-fetching once this is fixed? Or have you just lost faith and want to re-fetch every time for certainty? 
- Do we have some way to raise this issue back with MSFT?

mbarnes:
It's not broken, it's just the way we use [optimistic concurrency control](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#optimistic-concurrency-control) in Cosmos DB.

When we want to update an existing container item, we:
- Take a [callback function](https://github.com/Azure/ARO-HCP/blob/f26a94acc37e731b2c972795c5ff6f71751c8384/internal/database/database.go#L210) and [start a retry loop](https://github.com/Azure/ARO-HCP/blob/f26a94acc37e731b2c972795c5ff6f71751c8384/internal/database/database.go#L218).
- [Fetch the item to be updated to obtain its current etag](https://github.com/Azure/ARO-HCP/blob/f26a94acc37e731b2c972795c5ff6f71751c8384/internal/database/database.go#L222).
- [Pass the fetched item to the callback function](https://github.com/Azure/ARO-HCP/blob/f26a94acc37e731b2c972795c5ff6f71751c8384/internal/database/database.go#L227). This is where modifications are made.
- [Submit the modified item to Cosmos DB with the etag as a precondition](https://github.com/Azure/ARO-HCP/blob/f26a94acc37e731b2c972795c5ff6f71751c8384/internal/database/database.go#L236-L237).
- [If we get a PreconditionFailed error](https://github.com/Azure/ARO-HCP/blob/f26a94acc37e731b2c972795c5ff6f71751c8384/internal/database/database.go#L244), meaning our etag is out-of-date because the item was updated since we fetched it, we let the retry loop run again up to a fixed limit.

Now here's where the bug comes in.  When Cosmos DB creates or updates an item, by default it does not return the updated item content in the response.  You have to explicitly request that per-operation through an option named [EnabledContentResponseOnWrite](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos#ItemOptions).

In almost all cases we don't need the response content, so we leave the option false.  Here, however, is the one and only case so far where we _do_ need the response content, because we're expected to return the updated tags in our own response content.  Problem is we don't expose the [EnabledContentResponseOnWrite](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos#ItemOptions) option (or indeed _any_ options) through our current `DBClient` abstraction interface.  So easiest solution here is just refetch the item content after the update.

Is is the best solution?  Probably not.  But since I'm planning to dismantle the `DBClient` abstraction interface early next year (because it's getting in the way of using Cosmos DB more efficiently; this being one more example), I felt it was a good enough interim solution.

mociarain:
Ah yes, the ol'DBClient chestnut. Big +1 to doing something better with that. It's been a pain in Classic too

mociarain (approved):

SudoBrendan (approved):

PR #986, opened 2024-12-13
*: add acrpull controller, binding

stevekuznetsov:
### What this PR does

Jira: 
Link to demo recording:

### DIFF
--- a/dev-infrastructure/modules/aks-cluster-base.bicep
+++ b/dev-infrastructure/modules/aks-cluster-base.bicep
@@ -469,6 +469,36 @@ resource uami_fedcred 'Microsoft.ManagedIdentity/userAssignedIdentities/federate
   }
 ]
 
+resource pullerIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' = {
+    location: location
+    name: 'image-puller'
+}
+
+module acrPullRole 'acr/acr-permissions.bicep' = [
+  for (_, i) in acrPullResourceGroups: {
+    name: guid(acrRg[i].id, aksCluster.id, acrPullRoleDefinitionId)
+    scope: acrRg[i]
+    params: {
+      principalId: pullerIdentity.properties.principalId
+      acrResourceGroupid: acrRg[i].id
+    }
+  }
+]
+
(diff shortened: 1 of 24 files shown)


frontend/deploy/helm/frontend/templates/acrpullbinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: pull-binding
+spec:
+  acr:
+    environment: PublicCloud
+    scope: {{ .Values.pullBinding.registry }}
+    server: {{ .Values.pullBinding.scope }}
+  auth:
+    workloadIdentity:
+      serviceAccountRef: frontend
```
stevekuznetsov, line 12: The SA whose credentials are projected is not necessarily the same SA as where they are projected, on purpose. So if we wanted a different identity with pull rights federated to some other SA in the namespace, we could.

dev-infrastructure/modules/aks-cluster-base.bicep
```diff
@@ -460,15 +460,26 @@ resource uami_fedcred 'Microsoft.ManagedIdentity/userAssignedIdentities/federate
     parent: uami[i]
     name: '${workloadIdentities[i].value.uamiName}-${location}-fedcred'
     properties: {
-      audiences: [
```
stevekuznetsov, line 463: @geoberle  please provide feedback on all this

frontend/deploy/helm/frontend/templates/acrpullbinding.yaml
weinong: it's easy to use a separate one now. lets do that.

frontend/deploy/helm/frontend/templates/acrpullbinding.yaml
```diff
@@ -0,0 +1,15 @@
...
+      clientID: {{ .Values.pullBinding.workloadIdentityClientId }}
+      tenantID: {{ .Values.pullBinding.workloadIdentityTenantId }}
```
stevekuznetsov, line 14: Newest version of the controller allows us to reference the puller's MI for this token, letting multiple MI to be federated to the SA

PR #995, opened 2024-12-17
Add geo-replications to svc and ocp ACRs

jfchevrette:
### What this PR does

* Add an `acr-replication` bicep module used to set up a replication
* Use the `acr-replication` module in `region.bicep` to deploy ACR replications for both the svc and ocp ACRs
* An ACR replication cannot exist in the same region as the parent ACR and the module usage definitions checks for that

Jira: https://issues.redhat.com/browse/ARO-13411
Link to demo recording: 

### DIFF
--- a/dev-infrastructure/configurations/region.tmpl.bicepparam
+++ b/dev-infrastructure/configurations/region.tmpl.bicepparam
@@ -1,5 +1,14 @@
 using '../templates/region.bicep'
 
+// general
+param globalRegion = '{{ .global.region }}'
+param globalResourceGroup = '{{ .global.rg }}'
+param regionalRegion = '{{ .region }}'
+
+// acr
+param ocpAcrName = '{{ .ocpAcrName }}'
+param svcAcrName = '{{ .svcAcrName }}'
+
 // dns
 param baseDNSZoneName = '{{ .baseDnsZoneName }}'
 param baseDNSZoneResourceGroup = '{{ .baseDnsZoneRG }}'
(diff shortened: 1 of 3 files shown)


jfchevrette:

SudoBrendan (commented):
I think this work is well done, thanks! I don't have questions on the code, only on Acceptance Criteria. The jira is a bit vague so everything here fits well within that, but I imagine we will eventually want replication to every Azure region we support, meaning we'd need an array of locations, which I think would come from our global config file Gerd has stood up, if I'm understanding all of this correctly :sweat_smile:

At minimum I imagine we'd want that for the OCP ACR so that cluster installs aren't pulling images over an ocean and affecting install/repair times of clusters, right? SVC might be more lenient... but I could also see replication playing a role in our services' resiliency (e.g. we don't want a regional outage in data center A affecting data center B's resiliency - there might actually be MSFT requirements around this (?))

So I guess all that to say - this PR looks great, but what are we trying to build? What do we need to be successful as a service with our replication? Has a DDR been made (do we even need one, or just make that choice here)?

Thanks again for the hard work :+1:

tony-schndr:
@SudoBrendan the region template contains the replication resource that references the ACRs (ocp/svc) in the global resource group.  Anytime we deploy a region we will get replication without having to maintain a separate list.  No reason to maintain a separate list with it implemented this way.  There is a DDR btw, SD-DDR-0044, not much is said about global replication besides the ACRs should be replicated in every region the service is provided in.

tony-schndr (approved):
Nice work @jfchevrette, lgtm

SudoBrendan (approved):

jfchevrette:
Thank you both for the reviews!

geoberle:
well done. thank you

PR #1005, opened 2024-12-18
introduce step action based structs

geoberle:
### What this PR does

* added `Step` minimal interface for step metadata
* introduce `ShellStep` and `ARMStep` structs for existing steps to hold action specific parameters
* introduce new step types for DNS child zone delegation and KV issuer configuration

Jira: 
Link to demo recording:

### DIFF
--- a/tooling/templatize/pkg/pipeline/run_test.go
+++ b/tooling/templatize/pkg/pipeline/run_test.go
@@ -8,83 +8,46 @@ import (
 )
 
 func TestStepRun(t *testing.T) {
-	fooundOutput := ""
-	s := &Step{
-		Name:    "test",
-		Action:  "Shell",
-		Command: "echo hello",
-		outputFunc: func(output string) {
-			fooundOutput = output
+	foundOutput := ""
+	s := NewShellStep("step", "echo hello").WithOutputFunc(
+		func(output string) {
+			foundOutput = output
 		},
-	}
-	_, err := s.run(context.Background(), "", &executionTargetImpl{}, &PipelineRunOptions{}, nil)
+	)
+	_, err := RunStep(s, context.Background(), "", &executionTargetImpl{}, &PipelineRunOptions{}, nil)
 	assert.NilError(t, err)
-	assert.Equal(t, fooundOutput, "hello\n")
-}
-
-func TestStepRunSkip(t *testing.T) {
-	s := &Step{
-		Name: "step",
-	}
-	// this should skip
-	_, err := s.run(context.Background(), "", &executionTargetImpl{}, &PipelineRunOptions{Step: "skip"}, nil)
-	assert.NilError(t, err)
-
-	// this should fail
-	_, err = s.run(context.Background(), "", &executionTargetImpl{}, &PipelineRunOptions{Step: "step"}, nil)
-	assert.Error(t, err, "unsupported action type \"\"")
+	assert.Equal(t, foundOutput, "hello\n")
 }
 
 func TestResourceGroupRun(t *testing.T) {
 	foundOutput := ""
 	rg := &ResourceGroup{
-		Steps: []*Step{
-			{
-				Name:    "step",
-				Action:  "Shell",
-				Command: "echo hello",
-				outputFunc: func(output string) {
+		Steps: []Step{
+			NewShellStep("step", "echo hello").WithOutputFunc(
+				func(output string) {
 					foundOutput = output
 				},
-			},
+			),
 		},
 	}
-	err := rg.run(context.Background(), &PipelineRunOptions{}, &executionTargetImpl{}, make(map[string]output))
+	err := RunResourceGroup(rg, context.Background(), &PipelineRunOptions{}, &executionTargetImpl{}, make(map[string]output))
 	assert.NilError(t, err)
 	assert.Equal(t, foundOutput, "hello\n")
 }
 
 func TestResourceGroupError(t *testing.T) {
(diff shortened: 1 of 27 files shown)


janboll (approved):
:rocket:

tooling/templatize/pkg/pipeline/types.go
```diff
@@ -14,25 +16,170 @@ type Pipeline struct {
...
+	for i, rawStep := range rawRg.Steps {
+		switch rawStep.meta.Action {
+		case "Shell":
+			rg.Steps[i] = &ShellStep{}
+		case "ARM":
+			rg.Steps[i] = &ARMStep{}
+		default:
+			return fmt.Errorf("unknown action type %s", rawStep.meta.Action)
+		}
+		err := rawStep.unmarshal(rg.Steps[i])
```
janboll, line 48: don't you have to call UnmarshalYAML?

geoberle: `UnmarshalYAML` this is a "magic" function the golang yaml library calls

tooling/templatize/pkg/pipeline/run_test.go
```diff
@@ -9,82 +9,54 @@ import (
 
 func TestStepRun(t *testing.T) {
 	fooundOutput := ""
```
stevekuznetsov, line 11:
```suggestion
	foundOutput := ""
```

tooling/templatize/pkg/pipeline/run_test.go
```diff
@@ -9,82 +9,54 @@ import (
...
+	)
+	_, err := RunStep(s, context.Background(), "", &executionTargetImpl{}, &PipelineRunOptions{}, nil)
 	assert.NilError(t, err)
 	assert.Equal(t, fooundOutput, "hello\n")
 }
 
-func TestStepRunSkip(t *testing.T) {
-	s := &Step{
-		Name: "step",
+/*func TestStepRunSkip(t *testing.T) {
```
stevekuznetsov, line 22: Prefer to delete, comments get stale.

tooling/templatize/pkg/pipeline/types.go
```diff
@@ -14,25 +16,170 @@ type Pipeline struct {
...
+	}
+	msg.unmarshal = unmarshal
+	return nil
+}
+
+func (msg *rawStep) Unmarshal(v interface{}) error {
+	return msg.unmarshal(v)
+}
+
+type Step interface {
```
stevekuznetsov, line 94: No `Run()`?

tooling/templatize/pkg/pipeline/types.go
```diff
@@ -14,25 +16,170 @@ type Pipeline struct {
...
+		Command: command,
+	}
+}
+
+type ShellStep struct {
+	StepMeta   `yaml:",inline"`
+	Command    string     `yaml:"command,omitempty"`
+	Variables  []Variable `yaml:"variables,omitempty"`
+	DryRun     DryRun     `yaml:"dryRun,omitempty"`
+	outputFunc outPutHandler
```
stevekuznetsov, line 116: This smells a tiny bit - and makes static validation of your code harder by storing function pointers with data, which renders things like `cmp.Diff()` useless. Who creates the output handler, is it a runtime concern? Why does it need to be attached to the step as opposed to somehow passed in to a `Run()`?

tooling/templatize/pkg/pipeline/types.go
geoberle: i had a `Run` in there to begin with but then reconsidered. i have high hopes that these structures will be the ones we will harmonize onto when we merge the code from ARO-HCP + ADO-Pipelines into the new repo. and for the EV2 context, a run function does not make any sense. so  i tried to keep them as data-only as possible.

tooling/templatize/pkg/pipeline/run_test.go
geoberle: gone

tooling/templatize/pkg/pipeline/run_test.go
geoberle: ty

tooling/templatize/pkg/pipeline/types.go
geoberle: i agree. it has been here already so no change. let's tackle this in a followup PR. it should be part of the Run func somehow

tooling/templatize/pkg/pipeline/types.go
```diff
@@ -14,25 +16,170 @@ type Pipeline struct {
...
+func (m *StepMeta) Dependencies() []string {
+	return m.DependsOn
+}
+
+type rawStep struct {
+	meta      *StepMeta
+	unmarshal func(interface{}) error
+}
+
+func (msg *rawStep) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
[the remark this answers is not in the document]
geoberle: this is a way to remember the actual full content of a step, stored as the unmarshal func. this way we can turn a `rawStep` into `XXXStep` later. 

do you propose another option?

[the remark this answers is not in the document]
stevekuznetsov: The uninteresting approach would be e.g. https://goplay.tools/snippet/yBhVozOXrvB - up to you, IMO a little cleaner as you are in control of the codepath explicitly as opposed to implicitly having UnmarshalYAML called and storing the func for later. Little easier to wrap my brain around.

PR #1012, opened 2024-12-19
Add Pipeline Run Script

janboll:
### What this PR does
This PR switches frontline service deployments to use `templatize` pipeline runs. It adds Makefile targets to run deployment and additionally, it adds dry run capabilities. Dry run is implemented using the helm diff plugin. In order to prevent credential leaking `--suppress-secrets` is set. Nevertheless, extra care should be taken when adopting this for other helm deployments to prevent secret leaking (looking at you cluster-service :smiley: )

Jira:  https://issues.redhat.com/browse/ARO-13774
Link to demo recording:


run_pipeline.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
+
+PROJECT_ROOT_DIR=$(dirname "$(realpath "${BASH_SOURCE[0]}")")
```
geoberle, line 3: this is basically a copy of tempatlize.sh
can we try to pull the common parts into a script we include in both templatlize.sh and run_pipeline.sh?

janboll: Sure can do. I somehow feared having a single script can be quite complicated, but I'll give it a shot.

SudoBrendan (approved):

PR #1016, opened 2024-12-20
introduce pipeline.yaml file for CS

geoberle:
### What this PR does

* remove config.mk
* and introduce pipeline.yaml instead

adopting pipeline.yaml required different handling for certain vars since functions like `ternary` are not available there.
for the DB settings, i followed the same approach as for maestro, namely having dedicated preset value override files for the two scenarios.

additonally the operator configmap handling needed touching as well. since b64 encoding is not available the 6 operator role names are now provided via dedicated config.yaml vars. the rest of the information in the configmap is either the same in all environments (and can be defined directly in the configmap itself) or can be determined at runtime (e.g. the actual role resource IDs).

Jira: 
Link to demo recording:


cluster-service/deploy/helm/templates/azure-operators-managed-identities-config.configmap.yaml
```diff
@@ -6,4 +6,83 @@ metadata:
   namespace: {{ .Release.Namespace }}
 data:
   azure-operators-managed-identities-config.yaml: |
-{{ .Values.azureOperatorsManagedIdentitiesConfig | b64dec | indent 4 }}
+    controlPlaneOperatorsIdentities:
```
tony-schndr, line 9: With this approach we should be able to remove the linting exception I added https://github.com/Azure/ARO-HCP/blob/9c74aae617d4dd2007f1aea0a3a2144177e97f18/.yamllint.yml#L7

cluster-service/deploy/helm/templates/deployment.yaml
```diff
@@ -16,6 +16,14 @@ spec:
       labels:
         app: clusters-service
         azure.workload.identity/use: "true"
+      annotations:
```
tony-schndr, line 19: Nice addition of the checksums :+1:

tony-schndr (commented):
LGTM, nice work!

PR #1017, opened 2024-12-20
.github: fix az login check workflow and re-introduce fork check

jfchevrette:
### What this PR does

The jobs/steps names for workflow `Azure Login Check` which were added in #882 are misleading. One example can be seen in https://github.com/Azure/ARO-HCP/actions/runs/12396846632/job/34606970167?pr=992

This also re-introduces the job that checks if PRs are opened from a different repo, which I believe may have been erroneously removed in #882 

Jira: 
Link to demo recording:

### DIFF
--- a/.github/workflows/aro-hcp-cd.yml
+++ b/.github/workflows/aro-hcp-cd.yml
@@ -41,6 +41,15 @@
     cancel-in-progress: false
 
   jobs:
+    is_running_on_fork:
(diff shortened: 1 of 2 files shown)


jfchevrette:
The existing code is correct. The azure/login is in fact checking if the PR is opened from a fork due to the fact that on forks a write token won't be there and the azure login will fail because it does not have credentials

PR #1018, opened 2024-12-20
Add extra step to check the health of istio version before update

jonathan34c:
edit: dev-infrastructure/scripts/istio.sh to check if the target vers…

Per discuss with AKS team, we need to check if the istio is up running before do the istio upgrade to makesure.

### What this PR does

Jira: 
Link to demo recording:

### DIFF
--- a/dev-infrastructure/scripts/istio.sh
+++ b/dev-infrastructure/scripts/istio.sh
@@ -1,6 +1,13 @@
 #!/bin/bash
 
 set -euo pipefail
+ISTIO_NAMESPACE="aks-istio-system"
+echo "********** Check istio is up and running **************"
+ISTIO_PODS_COUNT=$(kubectl get pods -n ${ISTIO_NAMESPACE} -o jsonpath='{range .items[?(@.status.phase=="Running")]}{.metadata.name}{"\n"}{end}' | grep -c "${TARGET_VERSION}")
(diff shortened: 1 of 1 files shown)


dev-infrastructure/scripts/istio.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 
 set -euo pipefail
+ISTIO_NAMESPACE="aks-istio-system"
```
stevekuznetsov, line 4: Can we use `az aks mesh versions`?

weinong: use that for what?

stevekuznetsov: For checking what versions of the `istio` control plane are available, without having to assume things like where the AKS plugin happens to run them, how they are annotated, whatever. Not sure we care about any of those implementation details, and we certainly do not care about making sure that we know how many replicas the plugin decides to run. What if the plugin thinks 5 replicas are necessary? Or 3? Or just 1?

jonathan34c: the az aks mesh ```az aks mesh get-revisions --location``` only check the avilable versions to upgrade to in the certain location. it does not indicate the running istio in the current cluster

weinong (approved):
LGTM

mociarain (approved):

PR #1019, opened 2024-12-20
Refactor frontend/backend make deploy to use most recent image tags if COMMIT is not provided

jfchevrette:
### What this PR does

Small refactor of frontend & backend Makefiles with the following functional changes:
* If COMMIT is provided, no change (same behavior as before)
* If COMMIT is not provided, lookup the most recent tag in the registry and use that
* Adjust documentation to that effect

Jira: https://issues.redhat.com/browse/ARO-13263
Link to demo recording:

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -1,10 +1,15 @@
 -include ../setup-env.mk
 
-ifndef COMMIT
-COMMIT := $(shell git rev-parse --short=7 HEAD)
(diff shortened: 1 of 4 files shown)


jfchevrette:
I just realized the CD jobs will push tags to the repos as part of the PR checks which make using the most recent image tag in dev environments not viable

PR #1028, opened 2025-01-02
ARO-13402 | Adding Identity field to the HCPOpenShiftCluster

venkateshsredhat:
### What this PR does

The frontend needs to parse and validate the top-level ARM-defined "identity" resource field to fully support managed identities.

Jira:  https://issues.redhat.com/browse/ARO-13402 
Link to demo recording: 

The ticket says  Visibility tags are TBD so have left them blank .

### DIFF
--- a/internal/api/arm/identity.go
+++ b/internal/api/arm/identity.go
@@ -1,0 +1,27 @@
+package arm
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+// Represents to support the ManagedServiceIdentity ARM resource.
+type Identity struct {
+	PrincipalID            string                           `json:"principalId,omitempty"`
+	TenantID               string                           `json:"tenantId,omitempty"`
+	Type                   ManagedServiceIdentityType       `json:"type"`
+	UserAssignedIdentities map[string]*UserAssignedIdentity `json:"userAssignedIdentities,omitempty"`
+}
+
+// UserAssignedIdentity - User assigned identity properties https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/reference/data-types/#Azure.ResourceManager.CommonTypes.UserAssignedIdentity
+type UserAssignedIdentity struct {
+	ClientID    *string
+	PrincipalID *string
+}
+
+type ManagedServiceIdentityType string
+
+const (
+	ManagedServiceIdentityTypeNone                       ManagedServiceIdentityType = "None"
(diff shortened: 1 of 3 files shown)


venkateshsredhat:
@microsoft-github-policy-service agree company="RedHat"

miguelsorianod:
I see that the MR are related to changes in the cluster's frontend payload side.

Is the logic to send that information to the CS payload already performed and we are only missing the parsing from the frontend side? or are we missing there something?

I am asking this because we intended to rollout CS side changes that will make the managed identities required at API level, and we need to have the frontend side changes related to sending that information to CS implemented. If I understand correctly without the changes in this MR that wouldn't be possible, so we need to wait until this MR is merged. Let me know if this understanding is not accurate.

miguelsorianod:
>  If I understand correctly without the changes in this MR that wouldn't be possible, so we need to wait until this MR is merged. Let me know if this understanding is not accurate.

I think this part might not be accurate. Although the changes on this MR seem to be missing, they might not be strictly required for the CS side payload requirements.

mbarnes:
> Is the logic to send that information to the CS payload already performed and we are only missing the parsing from the frontend side? or are we missing there something?

Yes, that was added in #977 but with no validation on the frontend side.  Just a blind pass-through.

This PR is fulfilling Azure requirements for how ARM passes managed identities to resource providers.  After this merges we'll need validation logic that pairs each MI here with the operator-specific MI fields in #977.

mbarnes (commented):
Looking great so far.

FYI, the GET logic for this structure is going to be messy.  On GET, each user-assigned identity has to include ClientId and PrincipalId properties.  But that information comes from Cluster Service by way of its operator-specific managed identity properties.  So we'll have to construct the `identity.userAssignedIdentities` section from data returned by Cluster Service (in `/frontend/pkg/frontend/ocm.go`).  I'm cool with deferring that messiness to another PR if you'd rather just get the structs defined here.

SudoBrendan (commented):
Looking good so far - LMK when you're ready to take this out of draft :)

venkateshsredhat:
> Looking great so far.
> 
> FYI, the GET logic for this structure is going to be messy. On GET, each user-assigned identity has to include ClientId and PrincipalId properties. But that information comes from Cluster Service by way of its operator-specific managed identity properties. So we'll have to construct the `identity.userAssignedIdentities` section from data returned by Cluster Service (in `/frontend/pkg/frontend/ocm.go`). I'm cool with deferring that messiness to another PR if you'd rather just get the structs defined here.

okay sure , since this PR was raised from my forked branch due to permission issue . I will raise a new PR for this and for ClientId and principalId properties will raise a new one .

PR #1029, opened 2025-01-02
doc: document ARO-HCP Clusters creation in CS with Managed Identities

miguelsorianod:
### What this PR does

We document how to create the N needed Managed Identities when creating ARO-HCP Clusters directly against CS.

As an important remark, this assumes 4.17.x clusters creation. The set of required managed identities can differ between OCP versions. For now assuming 4.17.x is good enough but in the future we might want to improve on this.

As an important note, similar documentation / scripts should be created for the frontend side.


miguelsorianod:
This MR is a copy of https://github.com/Azure/ARO-HCP/pull/858 but with the branch created in the ARO-HCP repository itself. #858 was already reviewed so I am setting automerge on this one.

oriAdler (approved):

PR #1036, opened 2025-01-07
fix: register frontend metrics to global registry

simonpasquier:
### What this PR does

This change registers the frontend metrics to the global registry instead of creating a new registry which isn't exposed to the /metrics endpoint.


simonpasquier:

PR #1067, opened 2025-01-09
Buildout

jonathan34c:
### What this PR does

Jira: 
Link to demo recording:


geoberle:
i think the topology.yaml file from https://github.com/Azure/ARO-HCP/pull/1626 covers this

PR #1069, opened 2025-01-09
use MCR image for istio cert refresher pod

geoberle:
### What this PR does

* use MCR image for istio cert refresher pod
* use `sleep infinity` to prevent noise from a continously failing pod

Jira: 
Link to demo recording:


mociarain (approved):
Oh nice. Was there some issue with the standard busybox and Azure that meant we needed a special MSFT busybox?

geoberle:
> Oh nice. Was there some issue with the standard busybox and Azure that meant we needed a special MSFT busybox?

there was nothing technically wrong with it. but we should use only images from our private ACRs (and sync stuff into it if required) or from MCR.

PR #1093, opened 2025-01-13
add onecert signer to SVC KV

geoberle:
### What this PR does

* provide cert officer permissions to devops MSI (so EV2 can act on the KV)
* expose SVC KV url on svc-infra.bicep ...
* ... so it can be consumed by the issuer setup set

Jira: 
Link to demo recording:


mociarain (approved):
LGTM. Once this is merged is there anything else to enable?

tony-schndr (approved):

PR #1095, opened 2025-01-14
feat: automate creation and deletion of clusters against ARO-HCP frontend using UAMIs

miguelsorianod:
### What this PR does

Recently introduced changes in the Frontend and in Clusters Service have introduced the requirement of precreating a set of Azure Managed Identities and passing references to them during cluster creation API requests in both the Frontend API and the Clusters Service API side.

This commit updates the scripts in the `demo` directory to automate the precreation of Azure Managed Identities for the cluster being created and also the automated generation of the ARO-HCP Frontend ARO-HCP Cluster creation payload to include the information about the precreated Azure Managed Identities. It also automates the cleanup of the precreated Managed Identities of the cluster when the cluster is deleted.

Although precreation the identities and passing them as part of the part of the ARO-HCP Cluster creation payload is now a requirement, the Managed Identities will not be leveraged. They are API side changes only. In the future we will enable the actual usage of Managed Identities once the corresponding functionality is finalized.

Jira: 
Link to demo recording:

### DIFF
--- a/demo/03-create-cluster.sh
+++ b/demo/03-create-cluster.sh
@@ -1,21 +1,218 @@
 #!/bin/bash
 
+set -e
+
 source env_vars
 source "$(dirname "$0")"/common.sh
 
-CLUSTER_TMPL_FILE="cluster.tmpl.json"
-CLUSTER_FILE="cluster.json"
+initialize_control_plane_identities_uamis_names() {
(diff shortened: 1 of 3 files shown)


geoberle (approved):

miguelsorianod:
Thanks for the fixes in the deletion script 👍

PR #1096, opened 2025-01-14
setup DEV KV permissions during infra provisioning

geoberle:
### What this PR does

up until now, CS developers needed to manually run the `local-cs-permissions` make target if they wanted to have permissions to the shared OIDC storage, the SVC KV and the MC KVs.

with this patch, such permissions are set up automatically during personal dev environment creation

integrated DEV, CSPR and higher envs are unaffected by this change

Jira: 
Link to demo recording:


miguelsorianod:
lgtm

PR #1097, opened 2025-01-14
Ensure map exists

mociarain:
Context: https://redhat-external.slack.com/archives/C075PHEFZKQ/p1736847149670269

This PR ensures a map is present which should solve the immediate issue.

Note for the future: 
For other maps we initialize on the fly (https://github.com/Azure/ARO-HCP/blob/9396f053c1c2319c05e63339f706156e94038690/internal/api/utils.go#L72). I have a weak preference, since we have a default initialization, to initialize all the maps there but I mainly want one or the other. What do people think?

### DIFF
--- a/internal/api/hcpopenshiftcluster.go
+++ b/internal/api/hcpopenshiftcluster.go
@@ -124,8 +124,5 @@ func NewDefaultHCPOpenShiftCluster() *HCPOpenShiftCluster {
 				},
 			},
 		},
-		Identity: arm.Identity{
-			UserAssignedIdentities: make(map[string]*arm.UserAssignedIdentity),
-		},
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/api/hcpopenshiftcluster.go
```diff
@@ -124,5 +124,8 @@ func NewDefaultHCPOpenShiftCluster() *HCPOpenShiftCluster {
 				},
 			},
 		},
+		Identity: arm.Identity{
+			UserAssignedIdentities: make(map[string]*arm.UserAssignedIdentity),
+		},
```
mbarnes, line 129: I think a better place to create the map might be in [normalizeIdentityUserAssignedIdentities](https://github.com/Azure/ARO-HCP/blob/026ce1cfdb9eb0b437f75b149b3a9e69f117e988/internal/api/v20240610preview/hcpopenshiftclusters_methods.go#L547-L556).

mociarain: I'm fine with that

mbarnes:
> I have a weak preference, since we have a default initialization, to initialize all the maps there but I mainly want one or the other. What do people think?

This is a little tricky because there's places where the difference between a nil map and an initialized-but-empty map is significant to the frontend.  It may only apply to tags at the moment.  A nil map indicates no "tags" field was present in the payload.  A non-nil map, even if empty, signifies a "tags" field in the payload.  It may not apply in this case, but that's why we initialize maps on-demand.

mbarnes (approved):
LGTM.  We discussed on a call whether the `nil` check is necessary.  It probably will be necessary to eventually support patching this field, though we don't properly support patching maps yet.

PR #1098, opened 2025-01-14
Deploy kubelet-parameters ds to mgmt clusters

mmazur:
### What this PR does

Jira: https://issues.redhat.com/browse/ARO-13357
Link to demo recording: 

- `mgmt-fixes` is the best name I could come up with.

### DIFF
--- a/mgmt-fixes/Makefile
+++ b/mgmt-fixes/Makefile
@@ -5,8 +5,6 @@ HELM_CMD ?= helm upgrade --install
 deploy:
 	${HELM_CMD} mgmt-fixes \
 		deploy/helm/mgmt-fixes/
-#	kubectl create namespace mgmt-fixes --dry-run=client -o json | kubectl apply -f - && \
-#		--namespace mgmt-fixes
 .PHONY: deploy
 
 undeploy:
--- a/mgmt-fixes/Makefile
+++ b/mgmt-fixes/Makefile
@@ -1,0 +1,14 @@
+-include ../setup-env.mk
+-include ../helm-cmd.mk
+HELM_CMD ?= helm upgrade --install
+
+deploy:
+	${HELM_CMD} mgmt-fixes \
+		deploy/helm/mgmt-fixes/
+#	kubectl create namespace mgmt-fixes --dry-run=client -o json | kubectl apply -f - && \
+#		--namespace mgmt-fixes
+.PHONY: deploy
+
+undeploy:
+	helm uninstall mgmt-fixes
+.PHONY: undeploy
--- a/mgmt-fixes/deploy/helm/mgmt-fixes/Chart.yaml
+++ b/mgmt-fixes/deploy/helm/mgmt-fixes/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v2
+name: Fixes
+description: Any fixes that need to be uplied but have not (yet) been upstreamed
+version: 0.1.0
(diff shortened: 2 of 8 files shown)


mgmt-fixes/deploy/helm/mgmt-fixes/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v2
+name: Fixes
+description: Any fixes that need to be uplied but have not (yet) been upstreamed
```
mociarain, line 3:
```suggestion
description: Any fixes that need to be applied but have not (yet) been upstreamed
```
I think this was the typo

mgmt-fixes/Makefile
```diff
@@ -0,0 +1,14 @@
+-include ../setup-env.mk
+-include ../helm-cmd.mk
+HELM_CMD ?= helm upgrade --install
+
+deploy:
+	${HELM_CMD} mgmt-fixes \
+		deploy/helm/mgmt-fixes/
+#	kubectl create namespace mgmt-fixes --dry-run=client -o json | kubectl apply -f - && \
+#		--namespace mgmt-fixes
```
mociarain, line 9: Is this a leftover from a possible dedicated namespace or do we need it?

mgmt-fixes/deploy/helm/mgmt-fixes/templates/ds-kubelet-parameters.yaml
```diff
@@ -0,0 +1,75 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    component: kubelet-parameters
+  name: kubelet-parameters
```
mociarain, line 6:
```suggestion
  name: set-kubelet-parameters-for-scale
```
Or something else more descriptive maybe?

mociarain (changes requested):
Looks good but I think we need to add some references and context so we can understand where this comes from and help determine when we can remove it in the future. I would add something at the top of "ds-kubelet-parameters.yaml" and possibly add a README.md but I think you have better context to make that call than me.

Some questions:
- Do we expect to have more "management-fix" type things in future?

mgmt-fixes/Makefile
mmazur: Yup, obsolete bit I forgot to remove, gone now.

mgmt-fixes/deploy/helm/mgmt-fixes/templates/ds-kubelet-parameters.yaml
mmazur: More descriptive more better, so +1 from me.

mmazur:
Agreed on more context being useful, so I've added both a short readme and context to the ds yaml itself.

As for expectation for more things like this: I can't predict that.

Reasons why I went with a generic `mgmt-fixes` mechanism:
1. My experience tells me a need for additional stuff like this is likely to go up over time.
2. It felt weird to put something as specific as that kubelet config change as a top level target in the repo. To me it looks prettier if it's a generic mechanism at the top level, that I just use for this one thing in particular underneath.

geoberle (approved):

geoberle:
> My experience tells me a need for additional stuff like this is likely to go up over time.

having a dedicated chart like this for all kinds of MC pathes sounds reasonable to me. t

PR #1099, opened 2025-01-14
Mock DBClient with GoMock

mbarnes:
### What this PR does

- Introduces [GoMock](https://github.com/uber-go/mock) as a development tool, installed by Bingo.
- Generates a mocked `DBClient` implementation in a new "mocks" package under `/internal/mocks`.
- Converts existing unit tests to use the mocked `DBClient` instead of the in-memory cache `DBClient`.
- Removes the `--use-cache` command-line option.
- Finally, discards the in-memory cache `DBClient`.

The justification for all this is explained in the epic [ARO-14167 - Use Cosmos DB's Transactional Batch Operations in RP](https://issues.redhat.com/browse/ARO-14167).

Jiras:
- [ARO-14189 - Track new mock CLI utilities in Bingo](https://issues.redhat.com/browse/ARO-14189)
- [ARO-14168 - Use GoMock to mimic database operations in unit tests](https://issues.redhat.com/browse/ARO-14168)

Link to demo recording: 

The last commit was arrived at after much trial and error.  I don't know if there's a better way to adapt the GitHub workflow -- I couldn't find any predefined actions for installing Bingo, nor could I suss out how ARO-RP does it -- so someone with more experience can feel free to correct me.


.github/workflows/ci-go.yml
```diff
@@ -40,6 +40,11 @@ jobs:
           go-version-file: 'go.work'
           check-latest: true
 
+      - name: 'Install Bingo'
+        run: |
+          go install github.com/bwplotka/bingo@latest
+          bingo get --link github.com/bwplotka/bingo
```
SudoBrendan, line 46: We shouldn't need this change, here's why:

Bingo is basically just fancy `go vendor` but for tools. This means that we shouldn't need to use `bingo get` directly in CI, because the `.bingo` directory should already be built and committed as code in the repo. We'll be able to use any installed tool via the included `.mk` in any subsequent `make` calls, or, we can  `. .bingo/variables.env` file bingo already built for us for `/bin/sh`. After those files get sourced for their respective `make` or `sh` environments, we can use the variables they define (in our case, we'll have `${MOCKGEN}` that is a path to the "vendored" binary.

frontend/pkg/frontend/frontend_test.go
```diff
@@ -219,12 +238,39 @@ func TestSubscriptionsPUT(t *testing.T) {
...
+			// MiddlewareLockSubscription
+			// (except when MiddlewareValidateStatic fails)
+			mockDBClient.EXPECT().
+				GetLockClient().
+				MaxTimes(1)
+			if test.expectedStatusCode != http.StatusBadRequest {
+				// ArmSubscriptionPut
+				mockDBClient.EXPECT().
+					GetSubscriptionDoc(gomock.Any(), gomock.Any()).
+					DoAndReturn(func(context.Context, string) (*database.SubscriptionDocument, error) {
```
SudoBrendan, line 250: nit: This is the second time I see this in-line function - should we consider making it a named function defined in this test suite?

frontend/pkg/frontend/frontend_test.go
```diff
@@ -219,12 +238,39 @@ func TestSubscriptionsPUT(t *testing.T) {
...
+						UpdateSubscriptionDoc(gomock.Any(), gomock.Any(), gomock.Any())
 				}
 			}
+			// MiddlewareMetrics
+			// (except when MiddlewareValidateStatic fails)
+			mockDBClient.EXPECT().
+				GetSubscriptionDoc(gomock.Any(), gomock.Any()).
+				DoAndReturn(func(ctx context.Context, subscriptionID string) (*database.SubscriptionDocument, error) {
+					return database.NewSubscriptionDocument(subscriptionID, test.subscription), nil
+				}).
```
SudoBrendan, line 272: We can save putting a `Do` function call on the stack by just using [Return](https://pkg.go.dev/github.com/golang/mock/gomock#Call.Return) here.

.github/workflows/ci-go.yml
mbarnes: How does the Bingo executable otherwise get installed in a CI pipeline?

The first line just installs `$(GOBIN)/bingo`.
The second line establishes a versioned softlink `$(GOBIN)/bingo-v0.9.0` that our [.bingo/Variables.mk wants](https://github.com/Azure/ARO-HCP/blob/35408ec1b539bbfd69eb29c56017fb81801e4079/.bingo/Variables.mk#L20).

So this is just to bootstrap Bingo itself, not install other tools under Bingo's control.

frontend/pkg/frontend/helpers_test.go
```diff
@@ -149,21 +152,25 @@ func TestCheckForProvisioningStateConflict(t *testing.T) {
...
-					_ = frontend.dbClient.CreateResourceDoc(ctx, parentDoc)
-				}
+				mockDBClient.EXPECT().
+					GetResourceDoc(gomock.Any(), gomock.Any()).
+					DoAndReturn(func(ctx context.Context, resourceID *arm.ResourceID) (*database.ResourceDocument, error) {
+						resourceDoc := database.NewResourceDocument(resourceID)
+						// Hold the provisioning state to something benign.
+						resourceDoc.ProvisioningState = arm.ProvisioningStateSucceeded
+						return resourceDoc, nil
+					}).
```
SudoBrendan, line 172: I think this is another case where we should use `Return` since we don't have any logic; we can define our `resourceDoc` in the test, then make a call to Return with it since it's the same every time.

frontend/pkg/frontend/helpers_test.go
```diff
@@ -194,9 +203,13 @@ func TestCheckForProvisioningStateConflict(t *testing.T) {
 
 				parentResourceID := resourceID.GetParent()
 				if parentResourceID.ResourceType.Namespace == resourceID.ResourceType.Namespace {
-					parentDoc := database.NewResourceDocument(parentResourceID)
-					parentDoc.ProvisioningState = parentState
-					_ = frontend.dbClient.CreateResourceDoc(ctx, parentDoc)
+					mockDBClient.EXPECT().
+						GetResourceDoc(gomock.Any(), gomock.Any()).
+						DoAndReturn(func(ctx context.Context, resourceID *arm.ResourceID) (*database.ResourceDocument, error) {
```
SudoBrendan, line 208: same - it's a static value; use `Return`. There's more examples like these in other tests as well below but I won't comment on them to keep this review from getting redundant ;)

internal/mocks/generate.go
```diff
@@ -0,0 +1,6 @@
+package mocks
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+//go:generate $MOCKGEN -source=../database/database.go -destination=dbclient.go -package mocks github.com/Azure/ARO-HCP/internal/database DBClient
```
SudoBrendan, line 6: You might reach out to Amber about this... I realized while looking at ARO-RP that we specifically do track mockgen in .bingo - but we do _not_ use $MOCKGEN in `go generate` calls, and I'm sure there's a reason why, which she will know...

SudoBrendan (changes requested):
Here's a prelim review, a few things to fix related to installing Bingo in `make install-tools` and also some migrations from `DoAndReturn` into just `Return` for static values. Thanks Matt, great work!

Makefile
```diff
@@ -14,10 +14,18 @@ all: test lint
...
-test:
+test: mocks
 	go list -f '{{.Dir}}/...' -m |RUN_TEMPLATIZE_E2E=true xargs go test -timeout 1200s -tags=$(GOTAGS) -cover
 .PHONY: test
 
+mocks: install-tools
+	MOCKGEN=${MOCKGEN} go generate ./internal/mocks
+.PHONY: mocks
+
+install-tools:
```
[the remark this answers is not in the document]
mbarnes: Ah ha!  Thanks, that explains why it wasn't bootstrapping itself.

internal/mocks/generate.go
mbarnes: Found it.  Need to call `$(BINGO) get` with the `--link` option from the Makefile to create unversioned symlinks.
e.g.
```
$(GOBIN)/mockgen -> $(GOBIN)/mockgen-v0.5.0
```

frontend/pkg/frontend/frontend_test.go
mbarnes: It's a little tricky due to the variable scoping of `test.subDoc`.  But I suppose a named function could take a document pointer argument and return the inline function.  Bonus points if I can make it a generic function for any document type.

.github/workflows/ci-go.yml
mbarnes: Force-pushed an update that removes this hunk.

Makefile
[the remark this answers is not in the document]
mbarnes: Force-pushed an update that adds `$(BINGO)` as a prereq target.

internal/mocks/generate.go
mbarnes: Force-pushed an update to fix this.

frontend/pkg/frontend/node_pool_test.go
```diff
@@ -129,6 +130,39 @@ func TestCreateNodePool(t *testing.T) {
...
+			// CreateOrUpdateNodePool
+			mockDBClient.EXPECT().
+				CreateOperationDoc(gomock.Any(), gomock.Any())
+			// ExposeOperation
+			mockDBClient.EXPECT().
+				UpdateOperationDoc(gomock.Any(), gomock.Any(), gomock.Any())
+			// CreateOrUpdateNodePool
+			mockDBClient.EXPECT().
+				CreateResourceDoc(gomock.Any(), gomock.Any())
+
```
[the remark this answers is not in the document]
mbarnes: They're not all happening in one method.  This is tracking one HTTP request's journey through the frontend, including middleware and helper functions.  The comments I added are the function names where the DB call is being made from, for clarity.

mbarnes:
> Since we're now starting do we want to consider just going with Ginkgo from the start? I'm happy to apply it to this PR so we have a comparison.

I'm open to whatever testing framework the team prefers, but I'd rather defer any alternative to a future PR.  For this PR, my only interest is getting rid of the in-memory cache and getting _something_ in it's place for unit tests.  At least on ARO Classic I've seen strong opinions for and against Behavior Driven Development.  I personally am fine with Ginkgo for E2E tests, but not yet sure how I feel about it for unit tests.

frontend/pkg/frontend/node_pool_test.go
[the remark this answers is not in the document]
venkateshsredhat: We should rather think about splitting this , in case we adapt Ginkgo kind of frameworks ?

frontend/pkg/frontend/frontend_test.go
mbarnes: Turned out simpler than I thought, and I did manage to make it a generic function.

frontend/pkg/frontend/node_pool_test.go
```diff
@@ -40,6 +43,12 @@ var dummyVMSize = "Big"
...
 var dummyVersionID = "dummy"
 
+func equalResourceID(expectResourceID *arm.ResourceID) gomock.Matcher {
+	return gomock.Cond(func(actualResourceID *arm.ResourceID) bool {
+		return strings.EqualFold(actualResourceID.String(), expectResourceID.String())
+	})
+}
+
 func TestCreateNodePool(t *testing.T) {
 	clusterResouceID, _ := arm.ParseResourceID(dummyClusterID)
```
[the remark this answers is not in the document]
mbarnes: LOL, this exact typo is repeated like 8 times in the file.  Don't think it's mine but fixing.

internal/mocks/generate.go
mbarnes: I actually ended up reverting this back to what I had for a few reasons:

- I don't have `$GOBIN` in my `$PATH` and I don't want that to be a silent requirement.
- `.bingo/Variable.mk` defines `MOCKGEN` as `$(GOBIN)/mockgen-v0.5.0` so it's an absolute path.
- The generated file ends up recording which version of `mockgen` it was generated with, which seems potentially valuable since the generated files are committed to source control.
```
// Generated by this command:                                                                                                                                                                    
//                                                                                                                                                                                               
//    mockgen-v0.5.0 -source=../database/database.go -destination=dbclient.go -package mocks github.com/Azure/ARO-HCP/internal/database DBClient                                               
//                                                                                                                                                                                            
```

mbarnes:
All review comments are addressed.  I was able to simplify all the `DoAndReturn` calls to just `Return`.

backend/operations_scanner_test.go
```diff
@@ -74,20 +77,26 @@ func TestDeleteOperationCompleted(t *testing.T) {
...
-				_ = scanner.dbClient.CreateResourceDoc(ctx, resourceDoc)
-			}
+			mockDBClient.EXPECT().
+				DeleteResourceDoc(gomock.Any(), resourceID).
+				Do(func(ctx context.Context, resourceID *arm.ResourceID) {
+					resourceDocDeleted = tt.resourceDocPresent
+				})
+			mockDBClient.EXPECT().
+				UpdateOperationDoc(gomock.Any(), operationDoc.ID, gomock.Any()).
+				DoAndReturn(func(ctx context.Context, operationID string, callback func(*database.OperationDocument) bool) (bool, error) {
```
SudoBrendan, line 97: `Return` instead?

internal/mocks/generate.go
SudoBrendan: Ok - thanks for the double-check!

Makefile
```diff
@@ -14,10 +14,18 @@ all: test lint
 
 # There is currently no convenient way to run tests against a whole Go workspace
 # https://github.com/golang/go/issues/50745
-test:
+test: mocks
```
SudoBrendan, line 17: I hope this isn't too much scope creep - but we've got new code that's auto-generated to consider in this PR, so a few things I want to note:

CONTEXT:

While I too was originally on-board with your pattern, where `go generate` commands ultimately get called as a direct dependency of their requirements, in Classic, we eventually made the decision _not_ to do this for a few key reasons:

1. The golang maintainers say that [`go generate` should _not_ be part of your build process](https://go.googlesource.com/proposal/+/refs/heads/master/design/go-generate.md) - it should be independent and committed as code (rather than generated on the fly).
2. In the case here - tests in CI might pass even though you _didn't_ run `go generate` locally and committed the diff. This means that we haven't guaranteed the code in git is up-to-date with what is expected in a given commit on PRs. Instead, we should probably do the same in classic: if there's a diff after running `go generate`, that should produce a CI error.
3. This extends the duration of our test suite long-term, especially when we have quite a few mocks to generate in the long-game of this repo.

More of me being ripped tearing and screaming away from this pattern can be read here: https://redhat-internal.slack.com/archives/C02ULBRS68M/p1724353443729429?thread_ts=1724304200.163579&cid=C02ULBRS68M

WHAT I THINK WE SHOULD DO:

1. Remove `mocks` as a direct dependency of `test`, instead, the dev is expected to run `make mocks` if interfaces have changed.
2. Add CI logic to run an equivalent of `make mocks && exit 1 on git diff` to ensure that `make mocks` was run correctly before merge: https://github.com/Azure/ARO-RP/blob/b6a5f9ca0d2862ba3672271204659d835d91470b/.github/workflows/ci-go.yml#L29-L42

SudoBrendan (changes requested):
Thanks Matt! A few more comments here.

backend/operations_scanner_test.go
mbarnes: I don't think so here, because we have to invoke the callback that's passed in as a function argument.

internal/mocks/dbclient.go
SudoBrendan, line 1: nit: thinking long-game - do we potentially want to reflect the package directory structure inside of `mocks` to prevent the `mocks` directory from being too much of a dumping ground? I would think we'd at least want something like:

```
./internal/
├── database
└── mocks
    ├── database
    │   └── dbclient.go
    └── generate.go
```

mbarnes: I don't foresee us mocking much else beside _maybe_ the `ClusterServiceClient` interface.  I'm not opposed to the idea _if_ it does start to become a dumping ground, but I think it's a little premature now.

SudoBrendan: and it looks like that is already mocked out by hand... ack

Makefile
mbarnes: I removed `mocks` as a prerequisite for `test` in an existing commit and appended a new commit for a new "Regenerate mocks" CI step.

mbarnes: Also ran a test with a now-reverted commit where I modified `/internal/mocks/dbclient.go` and it caught it.

SudoBrendan (approved):

PR #1101, opened 2025-01-15
fix MI deletion on cluster deletion

geoberle:
### What this PR does

the while read logic skipped the last element in the MI list so it was not deleted.

Jira: 
Link to demo recording:


miguelsorianod:
I think the reason the last element was not read is because there's a newline missing at the end of the input being processed because of the usage of `-n` in the echo commands there. Just removing the `-n` from them I think would also solve the issues

miguelsorianod:
lgtm

PR #1103, opened 2025-01-15
Use of tagged IP Addresses (IPTags) for Cluster Egress and Istio Ingress

jfchevrette:
### What this PR does

- Add a Public IP Address bicep module
  - Allows creating a Public IP Address with or without a RoleAssignment
- Updates AKS cluster (managedCluster) bicep template to:
  - Create a Public IP Address to be used as cluster egress (optionally with IPTags)
  - Create a Public IP Address to be used for Istio Ingress (optionally with IPTags)
  - Change cluster egress loadbalancer parameters to use the IP address
  - Annotate the AKS-managed ASM IngressGateway service to make it use the desired IP address
- Update configs to pass IPTags for msft environments (defaults to none)

Jira:
- https://issues.redhat.com/browse/ARO-13083
- https://issues.redhat.com/browse/ARO-12529


dev-infrastructure/modules/aks-cluster-base.bicep
```diff
@@ -37,6 +37,31 @@ param podSubnetPrefix string
...
+param istioIngressGatewayIpAddressIPTags string = ''
+var istioIngressGatewayIpAddressIPTagsArray = [
+  for tag in (istioIngressGatewayIpAddressIPTags == '') ? [] : split(istioIngressGatewayIpAddressIPTags, ','): {
+    ipTagType: split(tag, ':')[0]
+    tag: split(tag, ':')[1]
+  }
+]
+
+@description('List of Availability Zones for zone-redundant resources')
+param zoneRedundancyZones array = [
```
tony-schndr, line 59: I think we will need to check the region to see if it supports zones 
https://learn.microsoft.com/en-us/azure/reliability/availability-zones-region-support 

Does it make sense to move that logic to the publicipaddress module?

jfchevrette: We currently do not do that check for the AKS cluster itself, so I figure this could be left out of this task for now to avoid scope creep

I did parametrize the zones array which is passed to the IP Address module and now used in the AKS cluster definition

tony-schndr: Ah, you're right. It makes sense to get that in a follow up.

dev-infrastructure/modules/aks-cluster-base.bicep
```diff
@@ -218,6 +243,40 @@ resource aksClusterAdminRoleAssignment 'Microsoft.Authorization/roleAssignments@
   }
 }
 
+var istioIngressGatewayIPAddressName = 'aro-hcp-istio-ingress'
```
geoberle, line 246: nit: this value is also used in the `istio.sh` script.
do you think it would make sense to have it in config.yaml and feed it into both places?

jfchevrette: I originally had it and then removed it before pushing my PR. The main (only?) benefit is that there is no hardcoding, but the flipside is this requires quite a bit of plumbing for little benefits IMO. I would not expect us to ever have to change this name but maybe there is a use case I am not thinking about?

Added in d3562a3 - PTAL

geoberle: i understand your concern. thanks for adding it anyways for the time being.
we can discuss in a team meeting how we want to do things like these in the future and revamp.

PR #1104, opened 2025-01-15
s/Key/ResourceId for ResourceDocument Type

mociarain:
Small refactor PR as a result of a discussion here: https://github.com/Azure/ARO-HCP/pull/920#discussion_r1916309106

### DIFF
--- a/internal/database/document.go
+++ b/internal/database/document.go
@@ -37,7 +37,7 @@ func newBaseDocument() BaseDocument {
 type ResourceDocument struct {
 	BaseDocument
 
-        // FIXME: Change the JSON field name when we're ready to break backward-compat.
+	// FIXME: Change the JSON field name when we're ready to break backward-compat.
 	ResourceId        *arm.ResourceID       `json:"key,omitempty"`
 	PartitionKey      string                `json:"partitionKey,omitempty"`
 	InternalID        ocm.InternalID        `json:"internalId,omitempty"`


internal/database/document.go
```diff
@@ -37,7 +37,7 @@ func newBaseDocument() BaseDocument {
 type ResourceDocument struct {
 	BaseDocument
 
-	Key               *arm.ResourceID       `json:"key,omitempty"`
+	ResourceId        *arm.ResourceID       `json:"key,omitempty"`
```
mbarnes, line 40:
```suggestion
        // FIXME: Change the JSON field name when we're ready to break backward-compat.
	ResourceId        *arm.ResourceID       `json:"key,omitempty"`
```

mbarnes (changes requested):
I suggested a comment in the struct definition as a reminder to myself when I go to rework these types.  Otherwise looks great.

venkateshsredhat:
LGTM , just a quick question what does this comment mean " // FIXME: Change the JSON field name when we're ready to break backward-compat."

mbarnes:
> what does this comment mean " // FIXME: Change the JSON field name when we're ready to break backward-compat."

That's a reminder for myself.  Maitiu is changing the Go field name but subtly leaving the JSON field name alone for backward-compatibility.  Meanwhile I'm working through a series of breaking changes to our database design, and as long as I'm breaking things I may as well also resolve this field name discrepancy at the same time.

venkateshsredhat:
LGTM

mbarnes (approved):

PR #1107, opened 2025-01-16
restructure AKS config schema in config.yaml

geoberle:
### What this PR does

introduce a reusable schema for AKS clusters and their nodepools and use it under `svc.aks` and `mgmt.aks`

Jira: 
Link to demo recording:

### DIFF
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -14,6 +14,81 @@
         "required": [
           "roleName"
         ]
+      },
+      "aksAgentPool": {
+        "type": "object",
+        "properties": {
+          "maxCount": {
+            "type": "number"
+          },
+          "minCount": {
+            "type": "number"
+          },
+          "osDiskSizeGB": {
+            "type": "number"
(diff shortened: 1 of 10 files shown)


config/config.schema.json
```diff
@@ -741,55 +743,11 @@
...
-            "additionalProperties": false,
-            "required": [
-              "azCount",
-              "maxCount",
-              "minCount",
-              "osDiskSizeGB",
-              "vmSize"
-            ]
+          "aks": {
+            "$ref": "#/definitions/aksConfig"
```
janboll, line 750: Do you have future use of aksConfig in mind? I really like this approach for node pools, but I think it could be overused in this particular case.

geoberle: It's in use twice in the config, once for SC and once for the MC. Right now, both of them are the same so I thought reuse makes sense. I'm open to suggestions though.

janboll (approved):

PR #1109, opened 2025-01-16
support global and regional infra to be in different subscriptions

geoberle:
### What this PR does

- add a global output bicep template to feed values about ACR and DNS parent zones into regional infra templates
- region template accepts resource IDs for ACR and DNS

Jira: 
Link to demo recording:


config/config.yaml
```diff
@@ -339,6 +337,12 @@ clouds:
       personal-dev:
         # this is the personal DEV environment
         defaults:
+          #global:
```
jfchevrette, line 340: What are these commented sections for?

Also is a bit confusing to me to have `defaults.*` at the top level and `defaults.*` here

dev-infrastructure/svc-pipeline.yaml
```diff
@@ -1,34 +1,86 @@
...
-    dependsOn:
-      - svc-infra
-    vaultBaseUrl:
-      input:
-        name: svcKeyVaultUrl
-        step: svc-infra
-    provider:
-      name: provider
-      value: OneCertV2-PrivateCA
+  # Configure certificate issuers for the SVC KV
```
jfchevrette, line 42: Did you mean to leave this one commented or should this be removed?

jfchevrette (commented):
I quickly did a pass over the changes and did not spot any glaring issues.

This PR is pretty difficult to review in details. Is there a chance this could be split into a smaller change that pertains specificly to the two bullet points from the PR description? I see changes that I cannot connect to the main intent, like changes in image tags and code changes - though that last one could be due to my lack of deeper knowledge of the tool stack

config/config.yaml
geoberle: each level has their own defaults section, so top level, cloud, env
this format was defined by MSFT

geoberle:
i can try to split it once i verified everything works in EV2

dev-infrastructure/configurations/mgmt-cluster.tmpl.bicepparam
```diff
@@ -43,4 +41,4 @@ param mgmtKeyVaultName = '{{ .mgmtKeyVault.name }}'
 param aroDevopsMsiId = '{{ .aroDevopsMsiId }}'
 
 // Azure Monitor Workspace
-param azureMonitorWorkspaceName = '{{ .monitoring.workspaceName }}'
+param azureMonitoringWorkspaceId = '__azureMonitoringWorkspaceId__'
```
jharrington22, line 44: What is `__azureMonitoringWorkspaceId__` a variable?

geoberle: it is an output chaining variable. so it is filled with an output value from another bicep run

geoberle:
close in favor of https://github.com/Azure/ARO-HCP/pull/1179 (which is smaller since a lot of different aspects have been handled in dedicated PRs

PR #1110, opened 2025-01-16
ARO-12097 | feat: Use snake case for azure runtime config

philipwu08:
### What this PR does
Update the keys in the azure runtime config from camel case to snake case.

Jira: 
Link to demo recording:

### DIFF
--- a/cluster-service/deploy/azure-runtime-config.tmpl.yaml
+++ b/cluster-service/deploy/azure-runtime-config.tmpl.yaml
@@ -1,11 +1,11 @@
 {
-  "cloudEnvironment": "AzurePublicCloud",
-  "tenantId": "{{ .extraVars.tenantId }}",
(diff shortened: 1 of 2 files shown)


JameelB:
I saw that there's another template for this config in the `deploy/helm` directory: https://github.com/Azure/ARO-HCP/blob/main/cluster-service/deploy/helm/templates/azure-runtime-config.configmap.yaml which I believe is used for the CS rollout pipeline? @tony-schndr @geoberle can you confirm? I think we need to change this in that config as well. 

It might be best to add both the camelCase and snake_case fields in these configs to ensure we've a successful rollout? We may need to merge this first before the CS changes, ensure that both versions of the fields are there. Rollout CS that changes from consuming from the camelCase -> snake_case fields. Once that's rolled out, at that point we can remove the old camelCase fields. wdyt?

philipwu08:
@microsoft-github-policy-service agree

PR #1111, opened 2025-01-16
feat: update frontend aro-hcp cluster deletion script to perform active waiting

miguelsorianod:
### What this PR does

This commit updates the demo/05-delete-cluster.sh script to wait until the cluster being deleted is fully deleted before continuing with its associated managed identities deletion. This ensures an ordered cleanup.
An environment variable WAIT_FOR_CLUSTER_DELETION can be set to enable/disable the active waiting during cluster deletion, but should be used with caution as not waiting means that the cluster and the managed identities are deleted in parallel, which could cause issues as in that case the identities deleted are still used by the cluster being deleted at that point in time.

Jira: 
Link to demo recording:


machi1990 (approved):
lgtm

I tested this locally

PR #1123, opened 2025-01-20
Refactor Cosmos database client creation

simonpasquier:
### What this PR does

This change moves the creation of the CosmosDB client to the internal/ directory because the code was identical between the fronend and backend components.

Jira: 
Link to demo recording:

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -13,10 +13,6 @@ import (
 	"runtime/debug"
 	"syscall"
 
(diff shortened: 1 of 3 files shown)


mbarnes (approved):
Makes sense to centralize this.  I'll probably add a `azcore.ClientOptions` parameter to `NewCosmosDatabaseClient` at some point but we're not quite ready for it yet.

PR #1128, opened 2025-01-20
arm: Drop azcorearm.ResourceID wrapper

mbarnes:
### What this PR does

The `azcorearm.ResourceID` text marshalling feature I [submitted upstream](https://github.com/Azure/azure-sdk-for-go/pull/23381), and was the reason for the wrapper, got released in [azcore 1.17.0](https://github.com/Azure/azure-sdk-for-go/releases/tag/sdk/azcore/v1.17.0).

Jira: no Jira; code cleanup

### DIFF
--- a/backend/operations_scanner_test.go
+++ b/backend/operations_scanner_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"go.uber.org/mock/gomock"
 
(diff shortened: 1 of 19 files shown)


venkateshsredhat:
@mbarnes As i understand we added the wrapper to implement MarshalText and Unmarshall methods but since you have done the contribution upstream we dont need it anymore . 
That makes sense , but i see we never used MarshalText or UnmarshallText , is that correct or we removed it sometime back?

venkateshsredhat (commented):
@mbarnes As i understand we added the wrapper to implement MarshalText and Unmarshall methods but since you have done the contribution upstream we dont need it anymore .
That makes sense , but i see we never used MarshalText or UnmarshallText , is that correct or we removed it sometime back?

venkateshsredhat (approved):

PR #1129, opened 2025-01-20
frontend: suppress warnings when running tests

simonpasquier:
### What this PR does

This change adds a default logger to
`TestCheckForProvisioningStateConflict()` otherwise the test output gets cluttered with warnings with the `-v` option.

Jira: 
Link to demo recording:


mbarnes (approved):
Oversight on my part.  Thanks for this!

PR #1132, opened 2025-01-21
frontend: add otel-sdk for inital trace support

frzifus:
### What this PR does

- install the opentelemetry-sdk and **disable trace export by default**
- the only impact will be outgoing requests to cluster-service will contain a `traceparent` in the header.
- configure trace middleware for incoming requests
- configure round-tripper for outgoing requests

It can be used to export traces to texting instances like all-in-one jaeger by patching the enviroment variables of the frontend deployment.

```diff
+        env:
+        - name: OTEL_EXPORTER_OTLP_ENDPOINT
+          value: https://example.<namespace>:4318
+        - name: OTEL_TRACES_EXPORTER
+          value: otlp
```

Jira: https://issues.redhat.com/browse/ARO-14430
Link to demo recording: 

How to start from scratch:
```bash
SKIP_CONFIRM=1 make infra.all
... install acrpull
... install frontend
... install cluster-service
```

For testing I deployed Jaeger:
```bash
kubectl create ns observability
kubectl apply -n observability -f https://gist.githubusercontent.com/frzifus/a5d8644a9103d3ce49e60d0f84f1e1ae/raw/7a70892125d98db01a468657d74bf4aaf89e5e6c/jaeger-simple.yaml
```

```
kubectl patch -n aro-hcp deployment aro-hcp-frontend \
  --type=json \
  -p='[
    {
      "op": "replace",
      "path": "/spec/template/spec/containers/0/image",
      "value": "ghcr.io/frzifus/aro-hcp-frontend:7bec29a"
    },
    {
      "op": "add",
      "path": "/spec/template/spec/containers/0/env",
      "value": [
        {
          "name": "OTEL_EXPORTER_OTLP_ENDPOINT",
          "value": "http://jaeger.observability:4318"
        },
        {
          "name": "OTEL_TRACES_EXPORTER",
          "value": "otlp"
        }
      ]
    }
  ]'
```
---
The follow-up PR for cluster service can be found here:
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/9060


frontend/cmd/cmd.go
```diff
@@ -103,7 +105,11 @@ func (opts *FrontendOpts) Run() error {
 	}
 
 	// Initialize Clusters Service Client
+
 	conn, err := sdk.NewUnauthenticatedConnectionBuilder().
+		TransportWrapper(func(r http.RoundTripper) http.RoundTripper {
+			return otelhttp.NewTransport(r)
```
frzifus, line 111: Consider setting correlation and requestID as attributes and put it into the baggage.

frontend/pkg/frontend/otel_sdk.go
```diff
@@ -0,0 +1,77 @@
...
+
+func InstallOpenTelemetryTracer(ctx context.Context, logger *slog.Logger, resourceAttrs ...attribute.KeyValue) (
+	func(context.Context) error,
+	error,
+) {
+	// NOTE: Auto span exporter sends traces to https://localhost:4318/v1/traces by default.
+	// We overwrite the default value with "none".
+	// See:
+	// https://github.com/open-telemetry/opentelemetry-go-contrib/blob/f6667f6f9eab2370f46d0903cf323cda3b7ca2bd/exporters/autoexport/spans.go#L32-L60
+	if v, _ := os.LookupEnv("OTEL_TRACES_EXPORTER"); v == "" {
```
simonpasquier, line 31: (nit) https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport#WithFallbackSpanExporter seems to imply that we can override the default exporter without messing up the env vars.

frontend/cmd/cmd.go
frzifus: Maybe in the future if needed.

frontend/pkg/frontend/routes.go
```diff
@@ -44,6 +44,7 @@ func (f *Frontend) routes() *MiddlewareMux {
 	metricsMiddleware := MetricsMiddleware{dbClient: f.dbClient, Emitter: f.metrics}
 
 	mux := NewMiddlewareMux(
+		MiddlewareTracing,
```
simonpasquier, line 47: Should we move it after the panic middleware? I assume that the panic middleware should wrap all other middlewares.

frontend/cmd/cmd.go
```diff
@@ -103,7 +105,11 @@ func (opts *FrontendOpts) Run() error {
 	}
 
 	// Initialize Clusters Service Client
+
```
simonpasquier, line 108: (nit)
```suggestion
```

frontend/cmd/cmd.go
```diff
@@ -125,7 +131,15 @@ func (opts *FrontendOpts) Run() error {
...
-
+	shutdown, err := frontend.InstallOpenTelemetryTracer(ctx, logger)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := shutdown(ctx); err != nil {
+			logger.Error("otel-sdk shutdown failed", "error", err)
+		}
+	}()
```
simonpasquier, line 142: (nit)
```suggestion

	shutdown, err := frontend.InstallOpenTelemetryTracer(ctx, logger)
	if err != nil {
		return err
	}
	defer func() {
		if err := shutdown(ctx); err != nil {
			logger.Error("otel-sdk shutdown failed", "error", err)
		}
	}()

```

frontend/pkg/frontend/middleware_tracing.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+	"net/http"
+	"strings"
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func MiddlewareTracing(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
```
simonpasquier, line 16: (suggestion) can we add a comment explaining the purpose of the middleware for future readers?

frontend/pkg/frontend/middleware_tracing.go
```diff
@@ -0,0 +1,28 @@
...
+	attrs := []attribute.KeyValue{}
+	for k, v := range r.Header {
+		attrs = append(attrs, attribute.String(fmt.Sprintf("http.request.header.%s", k), strings.Join(v, ",")))
```
simonpasquier, line 19: Are we sure that it's ok to report all of the incoming HTTP headers? E.g. any personal/private/sensitive information which shouldn't be reported?
Maybe it's better to start with an empty set of custom attributes?

frontend/pkg/frontend/otel_sdk.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/sdk/resource"
+)
+
+func InstallOpenTelemetryTracer(ctx context.Context, logger *slog.Logger, resourceAttrs ...attribute.KeyValue) (
+	func(context.Context) error,
+	error,
+) {
+	logger.InfoContext(ctx, "initialising OpenTelemetry tracer")
```
simonpasquier, line 27: (suggestion) would it be possible to log if the span exporter is a no-op exporter? E.g. either it's a noopSpanExporter instance or autoexport.IsNoneSpanExporter(...) returns true.

frontend/cmd/cmd.go
```diff
@@ -125,7 +131,15 @@ func (opts *FrontendOpts) Run() error {
 		return errors.New("location is required")
 	}
 	logger.Info(fmt.Sprintf("Application running in %s", opts.location))
-
+	shutdown, err := frontend.InstallOpenTelemetryTracer(ctx, logger)
```
simonpasquier, line 134: Should we add a region attribute (see lines above)?

frontend/pkg/frontend/routes.go
frzifus: Thats a good  question. I wasnt sure. How its setup we should complete the trace and report the `panic` and stack trace. I assume it we place it in front the trace will be gone.

frzifus: ftr: Discussed this with @simonpasquier and @dinhxuanvu.
Maybe it makes sense to register the panicMiddleware twice?
e.g.
```
MiddlewarePanic
MiddlewareTracing
metricsMiddleware.Metrics()
MiddlewareLogging
MiddlewarePanic
...
others
```

frontend/cmd/cmd.go
```diff
@@ -161,6 +165,16 @@ func (opts *FrontendOpts) Run() error {
...
+	// NOTE: In the future we may want to include "opts.location" as resource attribute.
+	shutdown, err := frontend.InstallOpenTelemetryTracer(ctx, logger)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := shutdown(ctx); err != nil {
+			logger.Error("otel-sdk shutdown failed", "error", err)
+		}
+	}()
```
mbarnes, line 177: Not a requirement, but I wonder if this could be moved into `Frontend.Run`?
I'd like to keep `cmd.go` as simple as possible.

frzifus: Sure, it just needs to be initialized before the service starts.  I placed it on top of the `Frontend.Run` method.

frontend/pkg/frontend/frontend.go
```diff
@@ -86,11 +86,21 @@ func (f *Frontend) Run(ctx context.Context, stop <-chan struct{}) {
...
 	logger := LoggerFromContext(f.server.BaseContext(f.listener))
 
+	errs, ctx := errgroup.WithContext(ctx)
+	errs.Go(func() error {
+		shutdown, err := InstallOpenTelemetryTracer(ctx, logger)
+		defer func() {
+			if err := shutdown(ctx); err != nil {
+				logger.Error("otel-sdk shutdown failed", "error", err)
+			}
+		}()
```
mbarnes, line 96: Looks like this defer function should be conditional.  If `InstallOpenTelemetryTracer` fails then the `shutdown` function will be `nil`, and attempting to call it will panic.  Am I reading that right?

mbarnes (changes requested):
One nit, but otherwise looks good.

frontend/pkg/frontend/frontend.go
```diff
@@ -86,11 +86,24 @@ func (f *Frontend) Run(ctx context.Context, stop <-chan struct{}) {
...
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if err := shutdown(ctx); err != nil {
+				logger.Error("otel-sdk shutdown failed", "error", err)
+			}
+		}()
+		return nil
+	})
```
mbarnes, line 101: Forgive me if this is a dumb question but when does `shutdown` actually run here?

Since this is now in a goroutine and `InstallOpenTelemetryTracer` doesn't block, I'm concerned that `shutdown` will just get called immediately since it looks like the goroutine exits immediately.

Either I'm misunderstanding the semantics of `errgroup` or maybe `InstallOpenTelemetryTracer` should just be called directly in the `Run` method.

Thoughts?

frzifus: y, it got called immediately and skilled any trace export. I moved the shutdown call to the stop routine.

mbarnes (approved):
I _think_ this looks correct now.  Sorry for the trouble.

Since I'm not a SME on tracing I'll defer to @simonpasquier for final approval.

frontend/pkg/frontend/middleware_tracing.go
```diff
@@ -0,0 +1,17 @@
...
+	"net/http"
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+// MiddlewareTracing starts a span wrapping all incoming HTTP requests.
+// Other middlwares or actual request handlers can extend its metadata or create
+// their own associated spans.
+func MiddlewareTracing(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	otelhttp.NewHandler(http.Handler(next), "middleware").ServeHTTP(w, r)
```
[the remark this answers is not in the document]
frzifus: Should we change it to `HTTP  {method}` then? Since dont have access to the URL patterns/template.  wdyt?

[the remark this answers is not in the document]
simonpasquier: It sounds good to me for now. IIUC we could change the name of the span after it was processed by the ServeMux but it would need testing.

frontend/pkg/frontend/frontend.go
```diff
@@ -74,23 +74,33 @@ func NewFrontend(logger *slog.Logger, listener net.Listener, metricsListener net
 }
 
 func (f *Frontend) Run(ctx context.Context, stop <-chan struct{}) {
+	// This just digs up the logger passed to NewFrontend.
+	logger := LoggerFromContext(f.server.BaseContext(f.listener))
+
+	errs, ctx := errgroup.WithContext(ctx)
+	var otelShutdown func(context.Context) error
+
+	errs.Go(func() (err error) {
```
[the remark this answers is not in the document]
mbarnes: Good point about the race.  Does this really need to be done in a goroutine at all?  It wasn't when this call was in `cmd.go`.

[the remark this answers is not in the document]
frzifus: Nope, placed it there that way to match the way other errors are handled.  :see_no_evil:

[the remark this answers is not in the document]
frzifus: @simonpasquier @mbarnes I changed the initialization. Let me know what you think. :smile:

simonpasquier (approved):
lgtm

dinhxuanvu (commented):
/lgtm

mbarnes (approved):
LGTM as well.

PR #1137, opened 2025-01-21
specify output format when expecting json from az

sanchezl:
### What this PR does

There are some instances where we expect JSON output from `az`, but we do not explicity specify the output format.
A developer changing the default output format of `az` would cause the affected scripts to fail.


sanchezl:

geoberle:
thank you. looking good so far. i'll do another review once the PR is out of Draft

geoberle (approved):

PR #1141, opened 2025-01-22
create public IP before AKS cluster

geoberle:
### What this PR does

create implicit dependency between public egress IP address and AKS cluster. this way bicep will only start the AKS creation/modification once the public IP is created and ready.

additionally persist the registration of the `Microsoft.Network/AllowBringYourOwnPublicIpAddress` AFEC flag, which is required to bring our own egress IP for the AKS cluter.

Jira: 
Link to demo recording:

### DIFF
--- a/dev-infrastructure/modules/aks-cluster-base.bicep
+++ b/dev-infrastructure/modules/aks-cluster-base.bicep
@@ -459,6 +459,7 @@ resource aksCluster 'Microsoft.ContainerService/managedClusters@2024-04-02-previ
   dependsOn: [
     aksNetworkContributorRoleAssignment
     aks_keyvault_crypto_user
+    aksClusterOutboundIPAddress
   ]
 }
 
--- a/dev-infrastructure/templates/feature-registration.bicep
+++ b/dev-infrastructure/templates/feature-registration.bicep
@@ -9,6 +9,7 @@ param features array = [
   'Microsoft.ContainerService/DisableSSHPreview'
   'Microsoft.ContainerService/IstioNativeSidecarModePreview'
   'Microsoft.Compute/EncryptionAtHost'
+  'Microsoft.Network/AllowBringYourOwnPublicIpAddress'
 ]
 
 resource featureReg 'Microsoft.Features/featureProviders/subscriptionFeatureRegistrations@2021-07-01' = [


dev-infrastructure/modules/aks-cluster-base.bicep
```diff
@@ -389,7 +389,7 @@ resource aksCluster 'Microsoft.ContainerService/managedClusters@2024-04-02-previ
         outboundIPs: {
           publicIPs: [
             {
-              id: resourceId('Microsoft.Network/publicIPAddresses', aksClusterOutboundIPAddressName)
+              id: aksClusterOutboundIPAddress.outputs.publicIPAddressId
```
jfchevrette, line 392: That did not work in my previous testing and I just tried it again:

> Preflight validation check for resource(s) for container service aro-hcp-aks in resource group hcp-underlay-usw3jche-svc failed. Message: Error parsing provided IP of [reference(resourceId('Microsoft.Resources/deployments', variables('aksClusterOutboundIPAddressName')), '2022-09-01').outputs.publicIPAddressId.value]. Error parsing failed for [reference(resourceId('Microsoft.Resources/deployments', variables('aksClusterOutboundIPAddressName')), '2022-09-01').outputs.publicIPAddressId.value]. Invalid resource Id format.

My understanding is this is apparently due to how bicep modules are processed in ARM and how it only produces outputs after the module deployment is completed. As such, it cannot provide the ID to the AKS cluster resource and ARM wants an ID, not a reference to one

jfchevrette:
Is this meant to solve an issue you encountered? I must have provisioned at least 20 clusters throughout my testing of #1103 and never encountered a dependency or missing resource error

dev-infrastructure/modules/aks-cluster-base.bicep
geoberle: i just realized you are right. it works well when the IP exists already but fails on a preflight check in EV2. i'll try something else. see the updated PR

geoberle:
> Is this meant to solve an issue you encountered? I must have provisioned at least 20 clusters throughout my testing of #1103 and never encountered a dependency or missing resource error

there can be timing issues, e.g. when IP address creation is delayed and the AKS cluster creation is making faster progress. i reverted the `id` reference change and instead created an explicit dependency from the AKS cluster to the public IP address module.

dev-infrastructure/modules/aks-cluster-base.bicep
jfchevrette: I had also refrained to use dependsOn because Azure [discourages it](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/resource-dependencies#explicit-dependency) and having never encountered timing issues in my testing I assumed somehow Azure was figuring out the dependency with `resourceId()`

That said nothing complains if we add the resource to dependsOn so lets do that for now to avoid the timing issues

jfchevrette (approved):

PR #1143, opened 2025-01-22
frontend: inject the correlation ID into CosmosDB requests

simonpasquier:
### What this PR does

This change adds a new policy to the frontend CosmosDB client which reads the correlation request ID from the request's context and injects the header into the request being sent to the CosmosDB API. It enables better traceability of incoming requests.

Jira: https://issues.redhat.com/browse/ARO-14434
Link to demo recording: 

I couldn't find a practical way to test the functionality without a "real" CosmosDB client.

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -16,6 +16,9 @@ import (
 	ocmsdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/spf13/cobra"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+
 	"github.com/Azure/ARO-HCP/internal/database"
 )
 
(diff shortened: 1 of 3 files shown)


internal/database/database.go
```diff
@@ -515,15 +515,19 @@ func (d *CosmosDBClient) UpdateSubscriptionDoc(ctx context.Context, subscription
...
-	azcoreClientOptions := azcore.ClientOptions{
+func NewCosmosDatabaseClient(url string, dbName string, opts ...func(*azcore.ClientOptions)) (*azcosmos.DatabaseClient, error) {
+	clientOptions := azcore.ClientOptions{
 		// FIXME Cloud should be determined by other means.
 		Cloud: cloud.AzurePublic,
 	}
 
+	for _, o := range opts {
+		o(&clientOptions)
+	}
```
mbarnes, line 526: I think I'd prefer if `*azcore.ClientOptions` were just passed to the function directly, unless I'm not seeing the advantage to this callback pattern.  Like the "FIXME" comment just above says, at some point we're going to have to stop hard-coding `cloud.AzurePublic` and determine that by other means (e.g. environment variable or IMDS).  I expect to eventually have a centralized function that returns a configured `azcore.ClientOptions`.

mbarnes (changes requested):
Suggested a simplification, but feel free to push back if you disagree.

mbarnes (approved):
Yeah, I like that better.  At least the database client code is cleaner.  Thanks!

backend/main.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
```
mbarnes, line 21: Here's the source of the lint error.  These need to be moved to the preceding group of imports, alongside `ocmsdk`.

simonpasquier: thanks! I was too lazy to run golangci-lint locally :stuck_out_tongue:

mbarnes (approved):

PR #1146, opened 2025-01-22
Add Wait for Deployment

janboll:
### What this PR does
This PR changes the behavior of ARM deployments. The current behavior is to run the deployment, this will cause errors. With this PR the runner will block if a previous deployment is already running. 

Jira: 
Link to demo recording:

### DIFF
--- a/tooling/templatize/pkg/pipeline/arm.go
+++ b/tooling/templatize/pkg/pipeline/arm.go
@@ -16,9 +16,12 @@ import (
 )
 
 type armClient struct {
-	creds          *azidentity.DefaultAzureCredential
-	SubscriptionID string
-	Region         string
+	deploymentClient        *armresources.DeploymentsClient
+	resourceGroupClient     *armresources.ResourceGroupsClient
+	deploymentRetryWaitTime int
+
+	Region        string
+	GetDeployment func(ctx context.Context, rgName, deploymentName string) (armresources.DeploymentsClientGetResponse, error)
 }
 
 func newArmClient(subscriptionID, region string) *armClient {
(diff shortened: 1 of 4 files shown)


geoberle (approved):

tooling/templatize/pkg/pipeline/arm.go
```diff
@@ -32,6 +33,45 @@ func newArmClient(subscriptionID, region string) *armClient {
 	}
 }
 
+func getDeployment(ctx context.Context, rgName, deploymentName string, client *armresources.DeploymentsClient) (*armresources.DeploymentsClientGetResponse, error) {
+	resp, err := client.Get(ctx, rgName, "asd", nil)
```
jfchevrette, line 37:
```suggestion
	resp, err := client.Get(ctx, rgName, deploymentName, nil)
```
unless I'm missing something?

jfchevrette (changes requested):
Other than the issue I spotted, this looks good from a code perspective

As a reviewer I would appreciate at least a short description of what this change is meant to do. The PR title alone does not tell me much....

tooling/templatize/pkg/pipeline/arm.go
```diff
@@ -15,23 +16,76 @@ import (
...
+
+	resp, err := a.getExistingDeployment(ctx, rgName, deploymentName)
+	if err != nil {
+		return fmt.Errorf("Error getting deployment %w", err)
+	}
+	if resp.Properties == nil {
+		return nil
+	}
+
+	for !canContinue {
```
jfchevrette, line 71: As far as I can tell this is meant to loop infinitely? Can you use `for { }` and avoid the `canContinue` variable?

tooling/templatize/pkg/pipeline/arm.go
```diff
@@ -15,23 +16,76 @@ import (
...
+		if *resp.Properties.ProvisioningState == armresources.ProvisioningStateRunning {
+			time.Sleep(time.Duration(a.deploymentRetryWaitTime) * time.Second)
+			currentTimeout -= a.deploymentRetryWaitTime
+			if currentTimeout < 0 {
+				return fmt.Errorf("Timeout exeeded waiting for deployment %s in rg %s", deploymentName, rgName)
+			}
+		} else {
+			return nil
+		}
+		resp, err = a.getExistingDeployment(ctx, rgName, deploymentName)
```
jfchevrette, line 81: Since you're already checking for the existence of the deployment at the top of the method, is this necessary here?

tooling/templatize/pkg/pipeline/arm.go
```diff
@@ -15,23 +16,76 @@ import (
...
+
+	for !canContinue {
+		if *resp.Properties.ProvisioningState == armresources.ProvisioningStateRunning {
+			time.Sleep(time.Duration(a.deploymentRetryWaitTime) * time.Second)
+			currentTimeout -= a.deploymentRetryWaitTime
+			if currentTimeout < 0 {
+				return fmt.Errorf("Timeout exeeded waiting for deployment %s in rg %s", deploymentName, rgName)
+			}
+		} else {
+			return nil
```
jfchevrette, line 79: nil: in go it is generally prefered to avoid else statements when possible to reduce nesting and improve readability... you could do something like this
```go
if !deploymentExist {
    // return error deployment does not exist
}
if deploymentRunning {
    // ...
}
if !deploymentRunning { // Or some other more specific conditions like Success or Failed
    return nil 
}
```

tooling/templatize/pkg/pipeline/arm.go
janboll: Thanks for spotting that

tooling/templatize/pkg/pipeline/arm.go
janboll: I refactored the method

geoberle (approved):

PR #1149, opened 2025-01-23
Fix RP frontend and backend make run command

ahitacat:
### What this PR does

Add CosmosDB related parameter for running both backend and frontend.
Remove extra parameters in the frontend that are required by CS.

Jira: 
Link to demo recording:

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -15,8 +15,11 @@ backend:
 .PHONY: backend
 
 run:
+	DB_URL=$$(az cosmosdb show -n ${DB_NAME} -g ${RESOURCEGROUP} --query documentEndpoint -o tsv) && \
 	./aro-hcp-backend --location ${LOCATION} \
-		--clusters-service-url http://localhost:8000
+		--clusters-service-url http://localhost:8000 \
+		--cosmos-name ${DB_NAME} \
+		--cosmos-url $${DB_URL}
 .PHONY: run
 
 clean:
--- a/frontend/Makefile
+++ b/frontend/Makefile
@@ -19,11 +19,12 @@ info:
 	@echo "COMMIT: ${COMMIT}"
 
 run:
+	DB_URL=$$(az cosmosdb show -n ${DB_NAME} -g ${RESOURCEGROUP} --query documentEndpoint -o tsv) && \
 	./aro-hcp-frontend --location ${LOCATION} \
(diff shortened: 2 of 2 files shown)


mociarain (approved):

backend/Makefile
```diff
@@ -17,6 +17,8 @@ backend:
 run:
 	./aro-hcp-backend --location ${LOCATION} \
 		--clusters-service-url http://localhost:8000
+		--cosmos-name ${DB_NAME} \
+		--cosmos-url ${DB_URL}
```
geoberle, line 21: where are these two env vars coming from? is the user setting them when running the cmd?

ahitacat: From envars, as the deploy make target. I just notice I need to reuse how the DB_URL is get.

geoberle: ack. we tried to not have heavy lookup calls like this one outside of targets. even if it is duplication, i suggest to have the DB_URL lookup in the deploy and run separately

backend/Makefile
```diff
@@ -39,7 +42,7 @@ deploy:
 			-g ${RESOURCEGROUP} \
 			-n backend \
 			--query clientId -o tsv) && \
-	DB_URL=$$(az cosmosdb show -n ${DB_NAME} -g ${RESOURCEGROUP} --query documentEndpoint -o tsv) && \
+	DB_URL=$(az cosmosdb show -n ${DB_NAME} -g ${RESOURCEGROUP} --query documentEndpoint -o tsv) && \
```
geoberle, line 45: I'm always confused by single and double dollar. Was this broken before?

backend/Makefile
```diff
@@ -15,8 +15,11 @@ backend:
 .PHONY: backend
 
 run:
+	DB_URL?= $(az cosmosdb show -n ${DB_NAME} -g ${RESOURCEGROUP} --query documentEndpoint -o tsv) && \
```
geoberle, line 18: Is the ? A valid symbol here?

backend/Makefile
ahitacat: It is a mistake, should be double dollar

backend/Makefile
ahitacat: No, it should be the same as the deployment.

geoberle (approved):

PR #1152, opened 2025-01-24
frontend: fix edge cases in MiddlewareValidateSubscriptionState

simonpasquier:
### What this PR does

This change fixes 2 edge cases:
* When there's no db client in the request's context.
* When the subscription state value isn't valid.

Jira: 
Link to demo recording: 

Extended the unit tests to cover the cases.


mbarnes (approved):
Good catch, thanks!

PR #1156, opened 2025-01-27
Miscellaneous fluff: typos, type names, etc.

mbarnes:
### What this PR does

Just a pile of simple "fluff" commits that have floated up from my ongoing database work.

Submitting these separately so they don't distract from the "meatier" changes to follow.

Jira: Loosely related to [ARO-14170 - Merge Cosmos DB containers](https://issues.redhat.com/browse/ARO-14170)

### DIFF
--- a/backend/operations_scanner.go
+++ b/backend/operations_scanner.go
@@ -96,7 +96,7 @@ func (s *OperationsScanner) Join() {
 func (s *OperationsScanner) pollDBOperations(ctx context.Context, logger *slog.Logger) {
 	var activeOperations []*database.OperationDocument
 
-	iterator := s.dbClient.ListAllOperationDocs(ctx)
+	iterator := s.dbClient.ListAllOperationDocs()
 
 	for item := range iterator.Items(ctx) {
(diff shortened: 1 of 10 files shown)


backend/operations_scanner.go
```diff
@@ -96,7 +96,7 @@ func (s *OperationsScanner) Join() {
 func (s *OperationsScanner) pollDBOperations(ctx context.Context, logger *slog.Logger) {
 	var activeOperations []*database.OperationDocument
 
-	iterator := s.dbClient.ListAllOperationDocs(ctx)
```
venkateshsredhat, line 99: Is there a particular reason why we are getting rid of the context ?

venkateshsredhat: Ok, i think they were unused .

venkateshsredhat (approved):
LGTM

backend/operations_scanner.go
mbarnes: It's not used.  The "List" methods prepare an iterator but don't do any I/O that would require a context.

PR #1159, opened 2025-01-27
fix azureOperatorsManagedIdentities

tony-schndr:
### What this PR does
Removes network/ingress identities from data plane identities
Adds a new identity for the control plane operator with Contributor Role.
Adds a new identity for cluster-azure-api with Contributor Role.

Jira: 
Link to demo recording:

### DIFF
--- a/cluster-service/deploy/helm/templates/azure-operators-managed-identities-config.configmap.yaml
+++ b/cluster-service/deploy/helm/templates/azure-operators-managed-identities-config.configmap.yaml
@@ -7,10 +7,15 @@ metadata:
 data:
   azure-operators-managed-identities-config.yaml: |
     controlPlaneOperatorsIdentities:
-      control-plane-operator:
+      cluster-api-azure:
         minOpenShiftVersion: 4.17
-        azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/{{ .Values.azureOperatorsMI.controlPlaneOperator.roleId }}'
-        azureRoleDefinitionName: '{{ .Values.azureOperatorsMI.controlPlaneOperator.roleName }}'
+        azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/{{ .Values.azureOperatorsMI.clusterApiAzure.roleId }}'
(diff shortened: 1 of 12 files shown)


geoberle (dismissed):

cluster-service/deploy/helm/templates/azure-operators-managed-identities-config.configmap.yaml
```diff
@@ -7,6 +7,11 @@ metadata:
 data:
   azure-operators-managed-identities-config.yaml: |
     controlPlaneOperatorsIdentities:
```
miguelsorianod, line 9: We are still missing another control plane operator identity:

The CAPZ one: cluster-api-azure

config/config.msft.yaml
```diff
@@ -166,6 +166,8 @@ clouds:
         imageTag: ecd15ad
         imageRepo: app-sre/uhc-clusters-service
         azureOperatorsManagedIdentities:
+          controlPlaneOperator:
+            roleName: Contributor
```
miguelsorianod, line 170: For dev at least we should have a more narrow set of permissions as with the other operators. For now this is ok but we should fix it. The work is tracked in https://issues.redhat.com/browse/ARO-14868.
The same applies to the capz operator.

cluster-service/deploy/helm/templates/azure-operators-managed-identities-config.configmap.yaml
```diff
@@ -7,6 +7,11 @@ metadata:
...
+      control-plane-operator:
```
miguelsorianod, line 10: This is not correct. It should be control-plane .

This is important because the names we define here *MUST* match the names specified in the spreadsheet where we define them (Proposed Key Name Changes for ARO Workload Identity) as they need to be agreed and consistent between offerings (ARO-HCP, ARO Classic, ...)

config/config.msft.yaml
tony-schndr: Noted, when the work is complete we need to update the custom roles in dev and then we can remove contributor.

miguelsorianod:
lgtm

geoberle (approved):

geoberle (approved):

PR #1161, opened 2025-01-28
Create custom role for MSI Mock

janboll:
### What this PR does

Jira: 
Link to demo recording:

### DIFF
--- a/dev-infrastructure/templates/mock-identities.bicep
+++ b/dev-infrastructure/templates/mock-identities.bicep
@@ -85,3 +85,26 @@ module msiRPMockIdentity '../modules/keyvault/key-vault-cert.bicep' = {
     validityInMonths: 1000
   }
 }
+
+resource msiCustomRole 'Microsoft.Authorization/roleDefinitions@2022-04-01' = {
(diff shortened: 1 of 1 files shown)


miguelsorianod:
The changes look good to me.

As an important note, I think the set of permission here will not be enough.

If the any of the control plane operators needs to interact with resources outside of the managed resource groups we will need to add them to the MSI Mock too. We still do not have an accurate set of permissions needed by the control plane operators themselves so I guess we will find if that's the case when actually trying the MIs functionality and/or we are given the accurate set of permissions.
Additionally, for the service managed identity I suspect we will need to have permissions to interact with other resources outside of the managed resource groups, in which case we will need to add them to the MSI Mock too. For now that's not the case but I suspect in the future that will occur.

Finally, once those changes are applied, do we also need to remove the currently set permissions on the MSI mock manually?

miguelsorianod:
Taking into consideration what I mentioned in my previous comment, lgtm

janboll:
We will update the role once additional required permissions are required.

geoberle (approved):

PR #1164, opened 2025-01-28
CODEOWNERS: Give RP maintainers ownership over go.work[.sum]

mbarnes:
Every time there's a `go.work.sum` change in a pull request I'm locked out of merging it, even if the rest of the changes are strictly RP focused.


mociarain (approved):
Same. +1 for this

PR #1165, opened 2025-01-28
database: Make DBClientIterator generic

mbarnes:
### What this PR does

Converts `DBClientIterator` to a generic type.

The type parameter is any Cosmos DB document type. This allows `DBClientIterator.Items` to return documents directly instead of raw JSON-encoded data that must be unmarshalled by the caller.

Truth be told, I wanted to do this when I first introduced `DBClientIterator` but didn't have my head around Go generics well enough at the time.

Jira: Loosely related to [ARO-14170 - Merge Cosmos DB containers](https://issues.redhat.com/browse/ARO-14170)

### DIFF
--- a/backend/operations_scanner.go
+++ b/backend/operations_scanner.go
@@ -5,7 +5,6 @@ package main
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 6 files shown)


mociarain (approved):
Nice. I really like generics

PR #1170, opened 2025-01-28
Aks name refactor

tony-schndr:
### What this PR does
Refactors clusters names so they are unique.  This change was mostly driven from the requirement of needing unique metrics as Azure Managed Prometheus does not support adding additional labels to default metrics (api server, node, etc) and there is no "resource_group" label to uniquely identity a metric.

### How was it tested?
I was able to provision all the infrastructure for a personal dev cluster and create a hosted cluster.

Jira: 
Link to demo recording:


tony-schndr:
This change is a break change, the errors in the "what-if" look to be expected.  We will need to fully redeploy environments

```
ERROR: InvalidTemplateDeployment - The template deployment 'svc-dev-13060613027' is not valid according to the validation procedure. The tracking id is 'c11f03b1-db2e-45b5-80c1-abfa9fc5f446'. See inner errors for details.
NodeResourceGroupAlreadyExists - Preflight validation check for resource(s) for container service dev-svc in resource group hcp-underlay-dev-svc failed. Message: Node resource group 'hcp-underlay-dev-svc-aks1' under subscription '***' already exists, please provide a new one.. Details: 
```

geoberle (dismissed):

geoberle:
i've tested this locally again after the rebase with a new personal dev env because the PR checks don't work on the current DEV and CS PR checks due to the AKS renaming

PR #1180, opened 2025-01-29
frontend: carry correlation and request ID in trace baggage

frzifus:
### What this PR does

Jira: https://issues.redhat.com/browse/ARO-14934
Link to demo recording:

### DIFF
--- a/frontend/pkg/frontend/middleware_logging.go
+++ b/frontend/pkg/frontend/middleware_logging.go
@@ -78,6 +78,7 @@ func MiddlewareLoggingPostMux(w http.ResponseWriter, r *http.Request, next http.
 
 	correlationData := arm.NewCorrelationData(r)
 
+	// ContextWithTraceCorrelationData is defined in middleware_tracing.go
 	ctx = ContextWithTraceCorrelationData(ctx, correlationData)
 
 	ctx = ContextWithCorrelationData(ctx, correlationData)
--- a/frontend/pkg/frontend/middleware_logging.go
+++ b/frontend/pkg/frontend/middleware_logging.go
@@ -11,10 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/baggage"
-	"go.opentelemetry.io/otel/trace"
-
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 )
@@ -82,30 +78,7 @@ func MiddlewareLoggingPostMux(w http.ResponseWriter, r *http.Request, next http.
 
 	correlationData := arm.NewCorrelationData(r)
 
-	// NOTE: Here the middleware span is extended by further attributes.
-	// If the tracingMiddleware is not registered, this lines will have no effect.
-	span := trace.SpanFromContext(ctx)
-	span.SetAttributes(attribute.String("correlation.id", correlationData.CorrelationRequestID))
-	span.SetAttributes(attribute.String("client.request.id", correlationData.ClientRequestID))
-
-	cID, err := baggage.NewMemberRaw("correlation.id", correlationData.CorrelationRequestID)
-	if err != nil {
-		span.RecordError(fmt.Errorf(`unable to create baggage member "correlation.id": %w`, err))
-	}
-	rID, err := baggage.NewMemberRaw("request.id", correlationData.RequestID.String())
-	if err != nil {
-		span.RecordError(fmt.Errorf(`unable to create baggage member "request.id": %w`, err))
-	}
-	crID, err := baggage.NewMemberRaw("client.request.id", correlationData.ClientRequestID)
-	if err != nil {
-		span.RecordError(fmt.Errorf(`unable to create baggage member "client.request.id": %w`, err))
-	}
-
-	bag, err := baggage.New(cID, rID, crID)
-	if err != nil {
-		span.RecordError(fmt.Errorf("unable to generate new baggage: %w", err))
-	}
-	ctx = baggage.ContextWithBaggage(ctx, bag)
+	ctx = ContextWithTraceCorrelationData(ctx, correlationData)
 
 	ctx = ContextWithCorrelationData(ctx, correlationData)
 
--- a/frontend/pkg/frontend/middleware_logging.go
+++ b/frontend/pkg/frontend/middleware_logging.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/baggage"
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/Azure/ARO-HCP/internal/api"
@@ -86,6 +87,25 @@ func MiddlewareLoggingPostMux(w http.ResponseWriter, r *http.Request, next http.
 	span := trace.SpanFromContext(ctx)
 	span.SetAttributes(attribute.String("correlation.id", correlationData.CorrelationRequestID))
 	span.SetAttributes(attribute.String("client.request.id", correlationData.ClientRequestID))
+
+	cID, err := baggage.NewMemberRaw("correlation.id", correlationData.CorrelationRequestID)
+	if err != nil {
+		span.RecordError(fmt.Errorf(`unable to create baggage member "correlation.id": %w`, err))
+	}
+	rID, err := baggage.NewMemberRaw("request.id", correlationData.RequestID.String())
+	if err != nil {
+		span.RecordError(fmt.Errorf(`unable to create baggage member "request.id": %w`, err))
+	}
+	crID, err := baggage.NewMemberRaw("client.request.id", correlationData.ClientRequestID)
+	if err != nil {
+		span.RecordError(fmt.Errorf(`unable to create baggage member "client.request.id": %w`, err))
+	}
+
+	bag, err := baggage.New(cID, rID, crID)
+	if err != nil {
+		span.RecordError(fmt.Errorf("unable to generate new baggage: %w", err))
+	}
+	ctx = baggage.ContextWithBaggage(ctx, bag)
 
 	ctx = ContextWithCorrelationData(ctx, correlationData)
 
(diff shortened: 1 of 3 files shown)


frontend/pkg/frontend/middleware_logging.go
```diff
@@ -86,6 +87,19 @@ func MiddlewareLoggingPostMux(w http.ResponseWriter, r *http.Request, next http.
 	span := trace.SpanFromContext(ctx)
 	span.SetAttributes(attribute.String("correlation.id", correlationData.CorrelationRequestID))
 	span.SetAttributes(attribute.String("client.request.id", correlationData.ClientRequestID))
+	cID, err := baggage.NewMember("correlationID", correlationData.CorrelationRequestID)
```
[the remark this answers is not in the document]
frzifus: We create two members and store them in the baggage. Each member has its own key. What would you propose instead?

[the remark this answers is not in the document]
frzifus: The propagated header should look like this:
```
Baggage: correlationID=7f64af13-0208-486f-a162-108d406e675d,requestID=f471f0af-f11f-4a86-bbe0-42727bf61623
```

frontend/pkg/frontend/middleware_logging.go
```diff
@@ -86,6 +87,19 @@ func MiddlewareLoggingPostMux(w http.ResponseWriter, r *http.Request, next http.
...
+	if err != nil {
+		span.AddEvent("unable to create correlationID baggage member")
```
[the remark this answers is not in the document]
dinhxuanvu: Should we record the error just like how you do in the `baggage` step? May as well keep it consistent across the board

frontend/pkg/frontend/middleware_logging.go
```diff
@@ -86,6 +87,19 @@ func MiddlewareLoggingPostMux(w http.ResponseWriter, r *http.Request, next http.
 	span := trace.SpanFromContext(ctx)
 	span.SetAttributes(attribute.String("correlation.id", correlationData.CorrelationRequestID))
 	span.SetAttributes(attribute.String("client.request.id", correlationData.ClientRequestID))
+	cID, err := baggage.NewMember("correlation.id", correlationData.CorrelationRequestID)
+	if err != nil {
+		span.AddEvent("unable to create correlationID baggage member")
+	}
+	rID, err := baggage.NewMember("request.id", correlationData.RequestID.String())
```
simonpasquier, line 94: in relation to  https://github.com/Azure/ARO-HCP/pull/1185, I wonder if we need to store the request ID in the baggage since CS can read it from the X-Request-ID header. But adding `ClientRequestID` would be useful.

frontend/pkg/frontend/middleware_logging.go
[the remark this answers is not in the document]
frzifus: Isn it beneficial when the invalid behaviour is marked at the trace itself? That way, you dont need to jump to the logs to understand why data is missing?

Or should we maybe do both?
e.g.:
image

frontend/pkg/frontend/middleware_logging.go
frzifus: That probably works too. The idea here was to transfer all request-specific data points for the correlation in the same way.

frontend/pkg/frontend/middleware_logging.go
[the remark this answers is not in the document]
simonpasquier: > Or should we maybe do both?

:+1: for doing both

frontend/pkg/frontend/middleware_logging.go
simonpasquier: sounds good to me.

frontend/pkg/frontend/middleware_logging.go
```diff
@@ -87,6 +88,25 @@ func MiddlewareLoggingPostMux(w http.ResponseWriter, r *http.Request, next http.
 	span.SetAttributes(attribute.String("correlation.id", correlationData.CorrelationRequestID))
 	span.SetAttributes(attribute.String("client.request.id", correlationData.ClientRequestID))
 
+	cID, err := baggage.NewMemberRaw("correlation.id", correlationData.CorrelationRequestID)
```
simonpasquier, line 91: (suggestion) move to a separate function so it can be unit-tested?

frontend/pkg/frontend/middleware_tracing.go
```diff
@@ -16,3 +22,44 @@ import (
...
+
+// ContextWithTraceCorrelationData adds correlationData as attributes to a propagated span.
+// It also registers a baggage with correlation data to the context.
+// If the context does not maintain a span, it has no effect.
+func ContextWithTraceCorrelationData(ctx context.Context, data *arm.CorrelationData) context.Context {
+	logger := LoggerFromContext(ctx)
+	// NOTE: Here the middleware span is extended by further attributes.
+	// If the tracingMiddleware is not registered, this lines will have no effect.
+	span := trace.SpanFromContext(ctx)
+	span.SetAttributes(attribute.String("correlation.id", data.CorrelationRequestID))
```
simonpasquier, line 34: suggestion: we could reduce duplication a bit.

````golang
    b, _ := bagagge.New() // will never err
    for _, := range []struct{
      name string
      value string
    }{
    } {
        {
            name: "correlation.id",
            value: data.CorrelationRequestID,
        },
        {
            ...
        },
        {
            ...
        },
    } {
        span.SetAttributes(attribute.String(name, value), value)
        m, err := baggage.NewMemberRaw(name, value)
        ...
        b.SetMember(m)
    }
````

Comment (frzifus) [reply frontend/pkg/frontend/middleware_tracing.go]: mh.. changed it :+1:

File: frontend/pkg/frontend/middleware_tracing_test.go [hunk 6]
```diff
@@ -0,0 +1,61 @@
...
+		"request.id":        requestID.String(),
+	}
+
+	data := &arm.CorrelationData{
+		CorrelationRequestID: expected["correlation.id"],
+		ClientRequestID:      expected["client.request.id"],
+		RequestID:            requestID,
+	}
+
+	// NOTE: no span, no effect
```
dinhxuanvu, line 34: Just a nitpick. I think this unit test is fine but technically you test a few scenarios here. So I wonder if it would be more readable if you construct this into test cases:

``` suggestion
tests := []struct {
	name    string
	data	something
	wantErr string
}{
	{
		// test spec
	},
}

for _, tt := range tests {
	// do your test
}
```
Feel free to ignore this suggestion if it ends up overcomplicating the test.

dinhxuanvu (commented):
Overall, looking good. Just a tiny suggestion.

frontend/pkg/frontend/middleware_tracing_test.go
frzifus: mh.. from point of view that would make it more compliated then its supposed to be.

Since 3 out of 4 tests, do just a simple function call with emtpy data or no active span to ensure, no active tracer will simply have no effect.

frontend/pkg/frontend/middleware_tracing.go
```diff
@@ -16,3 +22,46 @@ import (
...
+		{
+			name:  "request.id",
+			value: data.RequestID.String(),
+		},
+	} {
+		span.SetAttributes(attribute.String(e.name, e.value))
+
+		m, err := baggage.NewMemberRaw(e.name, e.value)
+		if err != nil {
+			fmtStr := `unable to create baggage member "%s"`
```
simonpasquier, line 58: (nit)
```suggestion
			fmtStr := "unable to create baggage member %q"
```

frontend/pkg/frontend/middleware_tracing.go
```diff
@@ -16,3 +22,46 @@ import (
...
+// ContextWithTraceCorrelationData adds correlationData as attributes to a propagated span.
+// It also registers a baggage with correlation data to the context.
+// If the context does not maintain a span, it has no effect.
+func ContextWithTraceCorrelationData(ctx context.Context, data *arm.CorrelationData) context.Context {
+	logger := LoggerFromContext(ctx)
+	// NOTE: Here the middleware span is extended by further attributes.
+	// If the tracingMiddleware is not registered, this lines will have no effect.
+	span := trace.SpanFromContext(ctx)
+
+	bag, _ := baggage.New()
```
simonpasquier, line 35:
```suggestion
	// Calling New() without any member never returns an error.
	bag, _ := baggage.New()
```

frontend/pkg/frontend/middleware_tracing.go
```diff
@@ -16,3 +22,46 @@ import (
...
+			value: data.RequestID.String(),
+		},
+	} {
+		span.SetAttributes(attribute.String(e.name, e.value))
+
+		m, err := baggage.NewMemberRaw(e.name, e.value)
+		if err != nil {
+			fmtStr := `unable to create baggage member "%s"`
+			span.RecordError(fmt.Errorf(fmtStr+": %w", e.name, err))
+			logger.ErrorContext(ctx, fmt.Sprintf(fmtStr, e.name), "error", err)
```
simonpasquier, line 60: (nit) 
```suggestion
			logger.ErrorContext(ctx, fmt.Sprintf(fmtStr, e.name), "error", err)
			continue
```

simonpasquier (dismissed):
lgtm

frzifus:
cc @mbarnes :smiley:

frontend/pkg/frontend/middleware_logging.go
```diff
@@ -81,11 +78,7 @@ func MiddlewareLoggingPostMux(w http.ResponseWriter, r *http.Request, next http.
 
 	correlationData := arm.NewCorrelationData(r)
 
-	// NOTE: Here the middleware span is extended by further attributes.
-	// If the tracingMiddleware is not registered, this lines will have no effect.
-	span := trace.SpanFromContext(ctx)
-	span.SetAttributes(attribute.String("correlation.id", correlationData.CorrelationRequestID))
-	span.SetAttributes(attribute.String("client.request.id", correlationData.ClientRequestID))
+	ctx = ContextWithTraceCorrelationData(ctx, correlationData)
```
mbarnes, line 81: I debated whether to suggest moving the function definition to this file, because each middleware file was intended to be self-contained.  I assume it's elsewhere to keep all the OpenTelemetry logic in one place?

Can you at least add a comment here like
```
// ContextWithTraceCorrelationData is defined in middleware_tracing.go
```

mbarnes (changes requested):
Just a small nit.

frontend/pkg/frontend/middleware_logging.go
frzifus: I see, y the thought was to keep the tracing peaces on one place. I will add a comment for now :)

mbarnes (approved):

PR #1189, opened 2025-01-29
database: Add PartitionKeys container

mbarnes:
### What this PR does

In anticipation of merging the "Resources", "Operations" and "Subscriptions" Cosmos DB containers, this PR adds a new container named "PartitionKeys" and two new `DBClient` methods: `ListOperationDocs` and `ListAllSubscriptionDocs`.

The PartitionKeys container is essentially the same hack [we're currently using for the Operations container](https://github.com/Azure/ARO-HCP/blob/760f2f748ca7c94cbc4bf74e7f4a04752b91d71b/internal/database/database.go#L28-L41), but the Operations container will soon be merged into the Resources container.  After the merge, Operation documents will be partitioned by Azure subscription ID.

The PartitionKeys container is an index of partition keys (Azure subscription IDs) for the Resources container.  This is necessary because the [Azure SDK for Go does not support cross-partition Cosmos DB queries](https://github.com/Azure/azure-sdk-for-go/issues/18578), and so the partition keys of a container are otherwise undiscoverable.

We overcome this same problem in the PartitionKeys container itself by keeping all items in a single partition (and a known partition key) so they can be found with a single-partition query.

If Azure's Go SDK ever catches up to their SDK for other languages, we can simply drop this container and start using cross-partition queries on the Resources container. **No data migration needed.**

As for the new `DBClient` methods:

- `ListOperationDocs` is a variation of `ListAllOperationDocs` that limits the query to operations within a single Azure subscription.  After the Cosmos DB containers are merged, this method will be reimplemented as a single-partition query and `DBClient.ListAllOperationDocs` will be removed.

- `ListAllSubscriptionDocs` should be self-explanatory.

Jira: [ARO-14170 - Merge Cosmos DB containers](https://issues.redhat.com/browse/ARO-14170)

The new `DBClient` methods are not used in this pull request.  They will be used in the [_next_ pull request for this Jira](https://github.com/Azure/ARO-HCP/pull/1200).  I thought it made more sense to introduce them here since they both leverage the PartitionKeys container.

### DIFF
--- a/internal/database/partitionkeys.go
+++ b/internal/database/partitionkeys.go
@@ -1,0 +1,114 @@
+package database
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
+)
+
+// XXX The Azure SDK for Go does not support cross-partition Cosmos DB
+//     queries which means the partition keys of a Cosmos DB container
+//     cannot be discovered through the Go SDK.
+//
+//     The Azure SDK for other languages like .NET, Java, and Python
+//     already support cross-partition queries, so there is reason to
+//     hope the Go SDK may some day catch up.
+//
+//     In the meantime, this container serves as a workaround. This is
+//     a single-partition container that simply collects the partition
+//     keys of the Resources container, like an index.
+//
+//     Once [1] is fixed we could simply drop this container and begin
+//     using cross-partition queries on the Resources container.
+//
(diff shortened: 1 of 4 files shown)


internal/database/partitionkeys.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// partitionKeyIterator implements DBClientIterator for subscriptions.
+type partitionKeyIterator struct {
+	client DBClient
+	pager  *runtime.Pager[azcosmos.QueryItemsResponse]
+	err    error
+}
+
+func upsertPartitionKey(ctx context.Context, containerClient *azcosmos.ContainerClient, id string) error {
```
mociarain, line 51: Why upsert and not create? I guess my real question is: How do we compose `PartitionKeys`?

mociarain: I get you now. Thanks for explaining

bennerv (approved):

PR #1196, opened 2025-01-30
fix(local-cs): ensure that recently added control-plane and cluster-api-azure are properly generated in the operators config for local CS

machi1990:
### What this PR does

Ensure that recently added control-plane and cluster-api-azure are properly generated in the operators config for local CS

Follows up https://github.com/Azure/ARO-HCP/pull/1159

Jira: 
Link to demo recording:


JameelB (approved):
Tested locally, these values were generated correctly.

PR #1204, opened 2025-01-30
fix(demo): attach service managed identities to the "identities.userAssignedIdentities" field

machi1990:
### What this PR does

Attach service managed identities to the "identities.userAssignedIdentities" field

This fixes the newly added validation in the RP which returned an error

```
{
    "error": {
        "message": "must be same as count of OperatorsAuthentication Managed Identities",
        "target": "identity.userAssignedIdentities"
    }
}
```

The fix ensures that all control plane identities + the service identity are copied to the identities.userAssignedIdentities map (as map keys)

Jira: 
Link to demo recording:


machi1990:
@mbarnes can you take a look as well?

PR #1211, opened 2025-01-31
disable grafana monitoring workspace registration temporarily

geoberle:
### What this PR does

mitigates bug introduced in https://github.com/Azure/ARO-HCP/pull/1179
i'll work on a fix while the GH actions are unblocked

Jira: 
Link to demo recording:

### DIFF
--- a/.github/workflows/environment-infra-cd.yml
+++ b/.github/workflows/environment-infra-cd.yml
@@ -59,7 +59,7 @@
         - name: 'Deploy'
           run: |
             cd dev-infrastructure/
-            make region metrics-infra
+            make region
 
     deploy_service_cluster_rg:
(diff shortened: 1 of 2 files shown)


.github/workflows/environment-infra-cd.yml
```diff
@@ -59,7 +59,7 @@
         - name: 'Deploy'
           run: |
             cd dev-infrastructure/
-            make region metrics-infra
```
geoberle, line 62: the metrics-infra target is now part of the region pipeline so calling it explicitely is not required anymore. i missed that in my previous PR. GH actions are hard to test :(

tony-schndr (approved):

PR #1222, opened 2025-02-03
arm: Add NewPagedResponse function

mbarnes:
### What this PR does

An empty response from a "list items" endpoint used to look like:

```
{
    "value": null
}
```

The [Resource Provider Contract](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/tree/master/v1.0) does not _explicitly_ say this is wrong, but since `value` is normally a list of items I think it would be better to return an empty list:

```
{
    "value": []
}
```

Jira: no Jira, just a bug I noticed


venkateshsredhat (approved):

venkateshsredhat (commented):
Looks like we are just initializing the variable here so LGTM.

venkateshsredhat (approved):

PR #1223, opened 2025-02-03
ARO-14927: add custom roles for cpo, kms and capz

tjungblu:
This adds custom roles for control plane operator, KMS and capz. KMS is unused by Hypershift for now, but pre-created to enable the hypershift changes later.

Those custom roles will be replaced by builtin roles for INT after we're done with the onboarding and approval process.

### What this PR does

Jira: ARO-14927

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -187,9 +187,9 @@ clouds:
         imageRepo: app-sre/uhc-clusters-service
         azureOperatorsManagedIdentities:
           clusterApiAzure:
-            roleName: Contributor
+            roleName: Azure Red Hat OpenShift Control Plane Operator Role - Dev
           controlPlane:
-            roleName: Contributor
+            roleName: Azure Red Hat OpenShift Cluster API Role - Dev
           cloudControllerManager:
             roleName: Azure Red Hat OpenShift Cloud Controller Manager - Dev
           ingress:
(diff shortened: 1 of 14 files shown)


config/config.yaml
```diff
@@ -183,9 +183,9 @@ clouds:
         imageRepo: app-sre/uhc-clusters-service
         azureOperatorsManagedIdentities:
           clusterApiAzure:
-            roleName: Contributor
+            roleName: Azure Red Hat OpenShift Control Plane Operator Role - Dev
           controlPlane:
-            roleName: Contributor
+            roleName: Azure Red Hat OpenShift Cluster API Role - Dev
```
tjungblu, line 188: I assume we need to leave this at `Contributor` until we make the KMS changes in Hypershift

p0lyn0mial: perhaps https://github.com/openshift/hypershift/blob/main/cmd/infra/azure/create.go#L831

how does `Azure Red Hat OpenShift Image Registry Operator - Dev` maps to `Azure Red Hat OpenShift Image Registry Operator Role` ?

tjungblu: thanks. Should be a 1:1 copy for the most part, unless anyone wants to change permissions going forward.
In DEV they want to keep the custom roles for this reason.

tjungblu: actually, this might get mapped from cluster service to the hypershift API via:
https://github.com/openshift/hypershift/pull/4811/files#diff-11b7c56a5faf0c5e2154fc35474456e2aac1b6ea8e22e5bbda6c3a3996d91afaR423

there is also a KMS field already

janboll (dismissed):

janboll (approved):

PR #1229, opened 2025-02-03
add maestro agent metrics

tony-schndr:
### What this PR does
Adds ability to scrape maestro agent metrics using nginx container source from MCR.
The init pod reads the service account token and copies it into nginx.conf that is then picked up by the metrics-proxy sidecar container.

Why is this needed?  Because Azure Managed Prometheus agent cannot scrape metrics using a token since it incorrectly masks the value AND maestro agent is built on ACM and removing authentication is not an option at this time.

Jira:: https://issues.redhat.com/browse/ARO-10274
Link to demo recording:

### DIFF
--- a/config/config.msft.yaml
+++ b/config/config.msft.yaml
@@ -88,6 +88,9 @@ defaults:
     restrictIstioIngress: true
     consumerName: hcp-underlay-{{ .ctx.regionShort }}-mgmt-{{ .ctx.stamp }}
     imageBase: quay.io/redhat-user-workloads/maestro-rhtap-tenant/maestro/maestro
+    agentSideCar:
+      imageBase: mcr.microsoft.com/azurelinux/base/nginx
+      imageTag: '1.25'
 
   # Cluster Service
   clusterService:
--- a/maestro/agent/helm/templates/maestro-agent.deployment.yaml
+++ b/maestro/agent/helm/templates/maestro-agent.deployment.yaml
@@ -17,7 +17,37 @@ spec:
         checksum/credsstore: {{ include (print $.Template.BasePath "/maestro.secretproviderclass.yaml") . | sha256sum }}
         checksum/config: {{ include (print $.Template.BasePath "/maestro.secret.yaml") . | sha256sum }}
     spec:
+      initContainers:
+      - name: init
+        image: "{{ .Values.sideCar.imageBase }}:{{ .Values.sideCar.imageTag }}"
+        env:
+          - name: TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: metrics-access-token
+                key: token
+        volumeMounts:
+        - name: nginx-config
+          mountPath: /etc/nginx/
+        - name: nginx-config-tmp
+          mountPath: /tmp/nginx/nginx.conf
+          subPath: nginx.conf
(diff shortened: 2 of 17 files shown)


config/config.msft.yaml
```diff
@@ -109,7 +112,7 @@ defaults:
     componentSync:
       enabled: true
       imageRepo: image-sync/component-sync
-      repositories: arohcpsvcdev.azurecr.io/image-sync/component-sync,arohcpsvcdev.azurecr.io/arohcpfrontend,arohcpsvcdev.azurecr.io/arohcpbackend,quay.io/acm-d/rhtap-hypershift-operator,arohcpsvcdev.azurecr.io/image-sync/oc-mirror,quay.io/app-sre/uhc-clusters-service,quay.io/package-operator/package-operator-package
+      repositories: arohcpsvcdev.azurecr.io/image-sync/component-sync,arohcpsvcdev.azurecr.io/arohcpfrontend,arohcpsvcdev.azurecr.io/arohcpbackend,quay.io/acm-d/rhtap-hypershift-operator,arohcpsvcdev.azurecr.io/image-sync/oc-mirror,quay.io/app-sre/uhc-clusters-service,quay.io/package-operator/package-operator-package,mcr.microsoft.com/azurelinux/base/nginx
```
geoberle, line 115: We don't need to sync MCR images tonight ACR. These are fine to use

config/config.msft.yaml
```diff
@@ -88,6 +88,9 @@ defaults:
     restrictIstioIngress: true
     consumerName: hcp-underlay-{{ .ctx.regionShort }}-mgmt-{{ .ctx.stamp }}
     imageBase: quay.io/redhat-user-workloads/maestro-rhtap-tenant/maestro/maestro
+    agentSideCar:
+      imageBase: arohcpsvcint.azurecr.io/azurelinux/base/nginx
```
geoberle, line 92: We can directly reference the MCR image

maestro/agent/helm/templates/metrics-proxy.configmap.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nginx-config
+data:
+  nginx.conf: |
+    worker_processes auto;
+    pid /run/nginx.pid;
+
```
jfchevrette, line 9: Suggestion: send errors to `stderr` to be picked up by the cluster so we have some visibility over any errors. Error log volume should be very light
```suggestion
error_log /dev/stderr;

```

maestro/agent/helm/templates/maestro-agent.deployment.yaml
```diff
@@ -36,6 +68,13 @@ spec:
           readOnly: true
       serviceAccountName: maestro
       volumes:
+      - name: nginx-config-tmp
+        configMap:
+          name: nginx-config
+      - name: nginx-config
+        emptyDir: {}
+      - name: logs
```
jfchevrette, line 76: We won't need this volume if both access and error logs are sent to null/stderr (not files)

jfchevrette (approved):
Beside two suggestions, LGTM

PR #1236, opened 2025-02-04
Fix missing image bug.

janboll:
Add script to resolve the latest available tag.

### What this PR does

[ARO-13263](https://issues.redhat.com/browse/ARO-13263)

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -2,12 +2,10 @@
 -include ../helm-cmd.mk
 HELM_CMD ?= helm upgrade --install
 
-ifndef COMMIT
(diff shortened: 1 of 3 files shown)


backend/Makefile
```diff
@@ -2,14 +2,10 @@
 -include ../helm-cmd.mk
 HELM_CMD ?= helm upgrade --install
 
-ifndef COMMIT
```
geoberle, line 5: if the `COMMIT` env var has a value, it comes from config.yaml and should be used. only if it is empty, the resolution process for the latest image should be used

janboll: Right, that got lost. I'll add it to the script

janboll: fixed

geoberle (approved):

PR #1241, opened 2025-02-04
frontend: refactor tracing and logging middlewares

simonpasquier:
### What this PR does

This change adds a new middleware (MiddlewareCorrelation) which is responsible for populating the correlation data into the incoming request's context. The rationale is that both logging and tracing middlewares rely on the correlation data and it's easier to have a single middleware responsible for it.

The commit also refactors the existing unit tests to exercise the middlewares rather than the inner functions.

Jira: 
Link to demo recording:

### DIFF
--- a/frontend/pkg/frontend/middleware_correlation.go
+++ b/frontend/pkg/frontend/middleware_correlation.go
@@ -1,0 +1,44 @@
+package frontend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+)
+
+// MiddlewareCorrelationData reads the correlation data from the incoming
+// request, extends the contextual logger with correlation attributes and adds
+// the necessary X-ms-* headers to the HTTP response.
+func MiddlewareCorrelationData(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	var (
+		ctx    = r.Context()
+		logger = LoggerFromContext(ctx)
+	)
+
+	correlationData := arm.NewCorrelationData(r)
+	ctx = ContextWithCorrelationData(ctx, correlationData)
+
+	logger = logger.With("request_id", correlationData.RequestID.String())
+	if correlationData.ClientRequestID != "" {
+		logger = logger.With("client_request_id", correlationData.ClientRequestID)
+	}
+
+	if correlationData.CorrelationRequestID != "" {
+		logger = logger.With("correlation_request_id", correlationData.CorrelationRequestID)
+	}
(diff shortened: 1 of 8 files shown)


simonpasquier:
cc @frzifus

frontend/pkg/frontend/middleware_tracing.go
```diff
@@ -16,24 +16,45 @@ import (
...
-	otelhttp.NewHandler(http.Handler(next), fmt.Sprintf("HTTP %s", r.Method)).ServeHTTP(w, r)
+	otelhttp.NewHandler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var (
+			ctx    = r.Context()
+			logger = LoggerFromContext(ctx)
+		)
+
+		data, err := CorrelationDataFromContext(ctx)
+		if err != nil {
+			logger.ErrorContext(ctx, "failed to find correlation data in context", "error", err)
```
frzifus, line 33: Should we also add an event to the current span using [recordError](https://github.com/open-telemetry/opentelemetry-go/blob/eb988d7a850f4e09763f77ed9f2ff3f02bb7a05a/sdk/trace/span.go#L518-L522)?

frzifus (commented):
Just a small thing, since we skip empty values and the entire call of `addCorrelationDataToSpanContext` when no correlation data is attached to the context, we should indicate that on the span itself.

frzifus (dismissed):

frontend/pkg/frontend/middleware_correlation.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
```
mbarnes, line 11: nit: For consistency, can you move the copyright above the `import`?

frontend/pkg/frontend/middleware_correlation_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
```
mbarnes, line 17: Same nit: move the copyright above `import`.

mbarnes (changes requested):
I like this a lot.  I originally based `MiddlewareLogging` on similar middleware code from ARO Classic but always felt it was a bit overloaded.  Splitting out the correlation data handling makes a lot more sense.

Just had a couple stupid nits.

mbarnes (approved):

PR #1242, opened 2025-02-04
ARO-15000 | Add no-CNI support to frontend

cgiradkar:
Made changes in the demo script files to form cluster creation payload accordingly if $NETWORK_TYPE is redefined.

### What this PR does

Jira:  [ARO-15000](https://issues.redhat.com/browse/ARO-15000)
Link to demo recording: 

Since the no-CNI feature depends solely on cluster creation payload, the only change needed here was to update demo scripts.

### DIFF
--- a/demo/03-create-cluster.sh
+++ b/demo/03-create-cluster.sh
@@ -207,6 +207,7 @@ main() {
 
   jq \
     --arg managed_rg "$MANAGED_RESOURCE_GROUP" \
+    --arg network_type "$NETWORK_TYPE" \
     --arg subnet_id "$SUBNET_ID" \
     --arg nsg_id "$NSG_ID" \
     --argjson uamis_json_map "$UAMIS_JSON_MAP" \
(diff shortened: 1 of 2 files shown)


frontend/pkg/frontend/const.go
```diff
@@ -17,5 +17,6 @@ const (
 	PathSegmentOperationID       = "operationid"
 	PathSegmentResourceGroupName = "resourcegroupname"
 	PathSegmentResourceName      = "resourcename"
+	PathSegmentNoCni             = "nocni"
```
bennerv, line 20: Why is `nocni` part of the path segment?  IIRC this is only part of the url, and `nocni` should be representative of an API change.

cgiradkar: The info is only passed through payload now

demo/env_vars
```diff
@@ -4,6 +4,9 @@
 if [ -z "${CUSTOMER_RG_NAME:-}" ]; then
     export CUSTOMER_RG_NAME="$USER-net-rg"
 fi
+if [ -z "${NETWORK_TYPE:-}" ]; then
+    export NETWORK_TYPE="OVNKubernetes"
+fi
```
bennerv, line 9: @mbarnes we default this to OVN right, if so @cgiradkar do we want to set defaults here?

cgiradkar: I have introduced NETWORK_TYPE as a new variable to be fed here  [#210](https://github.com/Azure/ARO-HCP/pull/1242/files#diff-43aeae6c6b9f7909dac7a4534c6132c2e1907ac5d4991e27d1fc60b69c6c9777R210). Hence, I have set default here. It can be removed if I move the check to makefile. WDYT?

SudoBrendan: I think Ben is saying that we already default this value in our API Spec: https://github.com/Azure/ARO-HCP/blob/2a122ebb5926989733235310f129f5557a05f902/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp#L215

cgiradkar: ack

demo/03-create-cluster.sh
```diff
@@ -216,7 +217,8 @@ main() {
       .properties.platform.subnetId = $subnet_id |
       .properties.platform.networkSecurityGroupId = $nsg_id |
       .properties.platform.operatorsAuthentication.userAssignedIdentities = $uamis_json_map |
-      .identity.userAssignedIdentities = $identity_uamis_json_map
+      .identity.userAssignedIdentities = $identity_uamis_json_map |
+      .properties.spec.network.networkType = $network_type
```
bennerv, line 221: This property doesn't exist anymore.  For the API review for MSFT we removed any field where the feature wasn't complete (or close to complete) so I don't know if this PR is valid anymore from frontend perspective.

cgiradkar:
closing th PR for now, feature will be added later

PR #1245, opened 2025-02-05
Replace yamllint with yamlfmt

bennerv:
### What this PR does

Replaces the megalinter yamllint with yamlfmt so that instead of us doing manual work to format yaml files correctly, we can just let the computers do it for us.


.github/workflows/yamlfmt.yml
```diff
@@ -0,0 +1,21 @@
...
+# MegaLinter GitHub Action configuration file
+# More info at https://oxsecurity.github.io/megalinter
+name: mega-linter
+on:
+  pull_request:
+permissions:
+  contents: read
+jobs:
+  build:
+    name: mega-linter
```
jfchevrette, line 11:
```suggestion
    name: yamlfmt
```

.github/workflows/yamlfmt.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        fetch-depth: 1
+    - name: 'run yamlfmt'
+      run: |-
+        make yamlfmt
+        git diff --exit-code\
```
jfchevrette, line 21: Is the trailing slash expected?

backend/pipeline.yaml
```diff
@@ -1,18 +1,19 @@
+---
 $schema: "pipeline.schema.v1"
 serviceGroup: Microsoft.Azure.ARO.HCP.RP.Backend
 rolloutName: RP Backend Rollout
 resourceGroups:
-- name: {{ .svc.rg }}
-  subscription: {{ .svc.subscription }}
-  aksCluster: {{ .aksName }}
+- name: {? {.svc.rg: ''} : ''}
```
jfchevrette, line 6: Are these changes with go templates conditional/ternary operators expected as part of this PR?

jfchevrette: This also changes the behavior where instead of the template failing if the variable is not set, it'll set defaults to empty strings which may not produce desired outcome. Seems like its better to let it fail than to use empty strings and hope for the best?

jfchevrette:
+1 from me

.github/workflows/yamlfmt.yml
jfchevrette, line 21: small UX suggestion :)
```suggestion
        git diff --exit-code
        [ $? -ne 0 ] && echo "YAML formatting needed. Please run `make yamlfmt` to resolve automatically"
```

backend/pipeline.yaml
geoberle: the templates fail and should fail if a variable can't be resolved

bennerv: The yaml formatter changed this for us.  I think we have to convert things to strings as a result.

jfchevrette: Oh wow TIL https://yaml.org/spec/1.2.2/#rule-e-node
I didn't think this was valid YAML... To essentially this not only add some yaml ternary blocks but breaks our templating?

Is there a rule we can exclude from the formatter to avoid it messing with our templating?

bennerv:
Implemented in #1498

PR #1251, opened 2025-02-05
ACR: Enable zone redundancy

janboll:
Adding just a parameter to the configuration for now, can be added to the configuration later if it is required to make it toggable

### What this PR does

Jira: 
Link to demo recording:

### DIFF
--- a/config/config.msft.yaml
+++ b/config/config.msft.yaml
@@ -246,7 +246,9 @@ clouds:
 
           # ACR
           svcAcrName: arohcpsvcint
+          svcAcrZoneRedundancy: Disabled
           ocpAcrName: arohcpocpint
+          ocpAcrZoneRedundancy: Disabled
 
           # Frontend
           frontend:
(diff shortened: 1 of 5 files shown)


jfchevrette:
I was planning to use the function introduced in #1240 (just merged) to dynamically configure resources based on whether the region provides zone redundancy. 

Something like this don't need the added parameters:
```bicep

import {locationIsZoneRedundant} from 'common.bicep'

module svcAcr '../modules/acr/acr.bicep' = {
  name: svcAcrSku
  params: {
    acrName: svcAcrName
    acrSku: svcAcrSku
    location: location
    zoneRedundancy: locationIsZoneRedundant(location) ? 'Enabled' : 'Disabled'
  }
}
```

Alternatively we can set the parameter default based on the function call
```bicep
param zoneRedundancy string = locationIsZoneRedundant(location) ? 'Enabled' : 'Disabled'
```

I prefer the first option as it avoid having parameters/configurations we may never need

WDYT?

geoberle (approved):

PR #1259, opened 2025-02-06
Add -o json to fix KeyVault Secret Fetch Issue

YiqinZhang:
### What this PR does
Add `-o json` to return the output in json format. Without explicitly specifying a format, no output is returned. This caused keyvault secret fetch to fail and led to service cluster deployment crashing.

Jira: 
[ARO-13654](https://issues.redhat.com/browse/ARO-13654)

Link to demo recording:


geoberle (approved):
thank you

PR #1269, opened 2025-02-07
Add templates to configure prometheus scraping

janboll:
### What this PR does
This adds a service and a service monitor. 
I manually applied this for testing before to ensure it works.  Afterwards I deleted the resources again.

[ARO-13247](https://issues.redhat.com/browse/ARO-13247)

### DIFF
--- a/pko/helm/templates/service.yaml
+++ b/pko/helm/templates/service.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: package-operator-metrics
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: package-operator
+    port: metrics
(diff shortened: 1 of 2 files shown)


pko/helm/templates/service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: package-operator-metrics
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: package-operator
+    port: metrics
```
jfchevrette, line 8: nit: Is this additional label and label selector in the ServiceMonitor needed since the ServiceMonitor already looks for an endpoint port named metrics?

jfchevrette (approved):

PR #1272, opened 2025-02-07
frontend: refactor subscription metrics

simonpasquier:
### What this PR does

This change implements a custom collector for the subscription lifecycle metrics. In practice, the collector maintains a local cache of subscriptions which is refreshed periodically from the database.

The benefits are:
* The metrics are persisted upon restart of the process (previously a subscription would only exist as a metric if it was accessed by some HTTP operation).
* All frontend replicas report the same metrics.
* Only the latest state of a subscription is reported (in the previous implementation, the old state would still be reported after an update).

The cache is refreshed every 30 seconds which should be good enough assuming that the metrics are scraped at a similar (or greater) interval.

Jira: 
Link to demo recording:

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -20,8 +20,10 @@ import (
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
+	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/sync/errgroup"
 
+	"github.com/Azure/ARO-HCP/frontend/pkg/metrics"
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/database"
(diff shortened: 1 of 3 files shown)


frontend/cmd/cmd.go
```diff
@@ -168,18 +170,33 @@ func (opts *FrontendOpts) Run() error {
...
+		f.Run(ctx, stop)
+	}()
+
+	collector := metrics.NewSubscriptionCollector(prometheus.DefaultRegisterer, dbClient, opts.location)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		collector.Run(logger, stop)
+	}()
```
mbarnes, line 193: Could this be moved inside `Frontend.Run`?  We're already using a wait group (actually an [error group](https://pkg.go.dev/golang.org/x/sync/errgroup)) there.  Just modify `SubscriptionCollector.Run` to return an `error`.

mbarnes (changes requested):
Suggested a simplification that would avoid another wait group.

frontend/cmd/cmd.go
simonpasquier: much simper indeed!

mbarnes (approved):
Very nice!

This will introduce a small conflict with changes I have staged in [#1215 - database: Merge Cosmos DB containers](https://github.com/Azure/ARO-HCP/pull/1215) but I'll adapt it rather than hold this PR up any longer.

PR #1283, opened 2025-02-10
use OCP 4.18rc6 in demo scripts

geoberle:
### What this PR does

Jira: 
Link to demo recording:


miguelsorianod:
We should update the nodepool template too.

Also in some of the scripts there are code comments mentioning 4.17. We should update them mentioning 4.18

PR #1290, opened 2025-02-10
Improve Availability Zone selection when provisioning AKS and zone-redundant resources

jfchevrette:
### What this PR does

* Refactor helper script to get regions and AZs in a better format to put in the `common.bicep` file
* Updated common.bicep with AZ availability for each region, add func to get list of AZs for a region
* Update aks-cluster, svc-cluster, mgmt-cluster templates to use the AZs provided by the updated data
  * Regions are not provisioned appropriately based on AZ availability 

Jira: https://issues.redhat.com/browse/XCMSTRAT-1067

Link to demo recording:

### DIFF
--- a/dev-infrastructure/modules/aks-cluster-base.bicep
+++ b/dev-infrastructure/modules/aks-cluster-base.bicep
@@ -25,6 +25,10 @@ param dnsServiceIP string = '10.130.0.10'
 // Passed Params and Overrides
 param location string
 
+@description('List of Availability Zones to use for the AKS cluster')
+param locationAvailabilityZones array
+var locationHasAvailabilityZones = length(locationAvailabilityZones) > 0
+
 @description('Set to true to prevent resources from being pruned after 48 hours')
 param persist bool = false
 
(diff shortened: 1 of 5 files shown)


dev-infrastructure/modules/aks-cluster-base.bicep
```diff
@@ -255,7 +252,7 @@ module istioIngressGatewayIPAddress '../modules/network/publicipaddress.bicep' =
     name: istioIngressGatewayIPAddressName
     ipTags: istioIngressGatewayIPAddressIPTagsArray
     location: location
-    zones: zoneRedundancyZones
+    zones: locationHasAvailabilityZones ? locationAvailabilityZones : null
```
tony-schndr, line 255: nit: could set this as a variable at the top, right?

tony-schndr (approved):
lgtm

dev-infrastructure/modules/aks-cluster-base.bicep
jfchevrette: I think it reads better without another variable which may not convey the two possible outcomes

geoberle (approved):

PR #1302, opened 2025-02-11
Use json output from azure cli

janboll:
### What this PR does
Use Json output, since it's more resilient

(not available with bicep though :( )


geoberle (approved):

janboll:
https://github.com/Azure/ARO-HCP/pull/1295 looks way better!

PR #1305, opened 2025-02-11
frontend: fix correlation headers in the HTTP response

simonpasquier:
### What this PR does

50dd65dc introduces a regression: the RP frontend stopped returning correlation headers in the HTTP response. The reason was that changing the Header map after calling `ResponseWriter.WriteHeader()` has no effect. The existing tests didn't catch the issue because they were accessing `httptest.ResponseRecorder` directly instead of testing the `*http.Response` value returned by `writer.Result()`.

Jira: 
Link to demo recording: 

I notice that other middleware tests have the same flaw (e.g. working directly on the ResponseRecorder instead of calling `*ResponseRecorder.Result()`). I'll post a follow-up PR to clean them up too.


mbarnes (approved):
Nice, that's bit me too in the past.  The `ResponseWriter` API is a bit unintuitive imo.

PR #1309, opened 2025-02-11
extract cs environment to config

tony-schndr:
### What this PR does
extracts cs environment to config and sets min/max char length in schema.

Jira: https://issues.redhat.com/browse/ARO-14975
Link to demo recording: 

It seems that in order to implement this change we must remove hosted clusters, recreate database (or drop tables) and redeploy cluster service.  There are a handful of other changes that require environment recreation so lets hold off on merging until we are ready to recreate the cs-pr and dev mvp environments.

### DIFF
--- a/cluster-service/deploy/helm/values.yaml
+++ b/cluster-service/deploy/helm/values.yaml
@@ -59,7 +59,7 @@ gatewayUrl: "http://127.0.0.1:9090"
 clientScopes: "openid"
 
 # Environment associated with this instance.
-environment: "aro-hcp-dev"
+environment: ""
 
 # The URL of the Backplane API which is exposed via an endpoint and is dynamically consumed by the Backplane CLI.
 # Backplane CLI expects a URL with the following format https://api.<OCM environment>.backplane.<domain>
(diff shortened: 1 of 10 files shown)


config/config.schema.json
```diff
@@ -172,6 +172,11 @@
       "clusterService": {
         "type": "object",
         "properties": {
+          "environment" :{
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 10
```
machi1990, line 178: 👍

cluster-service/deploy/helm/values.yaml
```diff
@@ -59,7 +59,7 @@ gatewayUrl: "http://127.0.0.1:9090"
 clientScopes: "openid"
 
 # Environment associated with this instance.
-environment: "aro-hcp-dev"
+environment: ""
```
miguelsorianod, line 62: Does this mean that by default it is the "" empty string? Is this intentional?

CS will fail to start if that's not set. I understand that we are setting this variable externally from the Makefile. If someone were to run the helm chart directly I understand it would fail.

This might be intentional and we might just want to be it that way. I am raising it just to be sure we are intentional/aware of it.

geoberle: this is a pattern. good that CS will fail. it means that we will spot an error if we fail to pass something valid to helm

PR #1311, opened 2025-02-12
various smaller fixes for MSFT INT deployment

geoberle:
### What this PR does

* missing AKS metadata read permissions for the devops MSI (this is now required by EV2 shell steps)
* update image references for various tools to bring in MIWI work to INT
* fake CS SP creds
* some pipeline step dependencies for svc and mgmt pipelines

Jira: 
Link to demo recording:


cluster-service/Makefile
```diff
@@ -41,7 +40,7 @@ deploy:
 	  --set tenantId=$${TENANT_ID} \
 	  --set region=${REGION} \
 	  --set serviceKeyvaultName=${SERVICE_KV} \
-	  --set csServicePrincipalCredsBase64=$${CS_SERVICE_PRINCIPAL_CREDS_BASE64} \
+	  --set csServicePrincipalCredsBase64="eyAidGVuYW50SWQiOiAiIiwgImNsaWVudElkIjogIiIsICJjbGllbnRTZWNyZXQiOiAiIiB9Cg==" \
```
tony-schndr, line 43: Just unset json?  Could we move this to the helm template?

geoberle: this will be removed soon alltogether but needs a patch in CS.
for now CS reads the content of the secret, verifies that the expected fields are there and then ignores it

@miguelsorianod PTAL at this

tony-schndr (dismissed):

PR #1320, opened 2025-02-13
frontend: add aro prefix to custom span attributes

frzifus:
### What this PR does

Since e.g. aws uses similar IDs for tracking and this could lead to conflicts in cs, we have decided to introduce a prefix. As of today, correlation id, request id and client request id are not listed in the [azure semantic-conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/azure.md). 

This PR proposes `aro` as custom prefix.

Jira: 
Link to demo recording: 

cc @simonpasquier @dinhxuanvu

### DIFF
--- a/frontend/pkg/frontend/middleware_tracing.go
+++ b/frontend/pkg/frontend/middleware_tracing.go
@@ -62,15 +62,15 @@ func addCorrelationDataToSpanContext(ctx context.Context, data *arm.CorrelationD
 		value string
 	}{
 		{
-			name:  "correlation.id",
+			name:  "aro.correlation.id",
 			value: data.CorrelationRequestID,
 		},
 		{
-			name:  "client.request.id",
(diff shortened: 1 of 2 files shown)


simonpasquier (dismissed):
lgtm

dinhxuanvu (commented):
/lgtm

mbarnes (commented):
Changes look fine.

Are conflicts with ARO Classic not a concern?  Just wondering if `arohcp` might be a less ambiguous prefix.

frzifus:
> Are conflicts with ARO Classic not a concern? Just wondering if arohcp might be a less ambiguous prefix.

I think aro should be fine. Since aro classic receives the same identifiers from ARM as the HCP frontend.

mbarnes (approved):

PR #1328, opened 2025-02-14
CS: remove ARO-HCP Clusters static credentials configuration

miguelsorianod:
### What this PR does

Now that the Managed Identities mechanism is used we can remove the `azure-auth-config` CLI flag and the usage of the credentials it references.

Jira: 
Link to demo recording:


dev-infrastructure/docs/development-setup.md
```diff
@@ -426,9 +407,9 @@ First install the agent
     ```bash
     SUBSCRIPTION_NAME="ARO Hosted Control Planes (EA Subscription 1)"
     RESOURCENAME="<INSERT-NAME>"
-    SUBSCRIPTION=$(echo $(az account subscription list | jq '.[] | select(.displayName == $SUBSCRIPTION_NAME)' | jq -r '.subscriptionId'))
```
miguelsorianod, line 429: This was actually broken before this MR so I updated it

miguelsorianod:
There's some PR checks failing but they are unrelated to this PR. They occur because other existing issues, as discussed with @geoberle .
Once those are fixed we'll rerun the PR and when the checks pass merge it, assuming the changes look good to the reviewers.

machi1990 (approved):

geoberle:
the only failing step is a linter one. i'll fix that in an upcoming PR so we can merge this one here

Issue #1163, opened 2025-01-28
frontend: ci test seems to be flaky


frzifus:
https://github.com/Azure/ARO-HCP/actions/runs/13010842188/job/36287996416?pr=1132

```
	github.com/Azure/ARO-HCP/frontend		coverage: 0.0% of statements
	github.com/Azure/ARO-HCP/frontend/pkg/config		coverage: 0.0% of statements
	github.com/Azure/ARO-HCP/frontend/cmd		coverage: 0.0% of statements
	github.com/Azure/ARO-HCP/frontend/utils		coverage: 0.0% of statements
{"time":"2025-01-28T12:54:15.228403686Z","level":"ERROR","msg":"error retrieving value from context, value obtained was '<nil>' and type obtained was '*slog.Logger'"}
{"time":"2025-01-28T12:54:15.228567801Z","level":"INFO","msg":"read request","request_method":"PUT","request_path":"/subscriptions/00000000-0000-0000-0000-000000000000","request_proto":"HTTP/1.1","request_query":"api-version=2.0","request_remote_addr":"127.0.0.1:33948","request_user_agent":"Go-http-client/1.1"}
{"time":"2025-01-28T12:54:15.228606082Z","level":"ERROR","msg":"error retrieving value from context, value obtained was '<nil>' and type obtained was '*slog.Logger'"}
{"time":"2025-01-28T12:54:15.228665262Z","level":"ERROR","msg":"error retrieving value from context, value obtained was '<nil>' and type obtained was '*slog.Logger'"}
{"time":"2025-01-28T12:54:15.228814279Z","level":"ERROR","msg":"error retrieving value from context, value obtained was '<nil>' and type obtained was '*slog.Logger'"}
{"time":"2025-01-28T12:54:15.228823837Z","level":"ERROR","msg":"error retrieving value from context, value obtained was '<nil>' and type obtained was '*slog.Logger'"}
{"time":"2025-01-28T12:54:15.228837633Z","level":"ERROR","msg":"error retrieving value from context, value obtained was '<nil>' and type obtained was '<nil>'","request_id":"f38eaa54-6092-497e-80cb-1ba334486a75","client_request_id":"","correlation_request_id":"","subscription_id":"00000000-0000-0000-0000-000000000000"}
[...]

PR #1333, opened 2025-02-15
frontend: Do not handle panics when running "go test"

mbarnes:
### What this PR does

This effectively disables `MiddlewarePanic` when running unit tests.

Catching a panic can sometimes obscure the root cause of a test failure, especially when the panic is intermittent and happens during a CI pipeline run where there's no opportunity to debug.

This has been observed at least once before in [PR #833](https://github.com/Azure/ARO-HCP/pull/833) and may be happening again according to [issue #1163](https://github.com/Azure/ARO-HCP/issues/1163).

### DIFF
--- a/frontend/pkg/frontend/middleware_panic.go
+++ b/frontend/pkg/frontend/middleware_panic.go
@@ -7,18 +7,22 @@ import (
 	"fmt"
 	"net/http"
 	"runtime/debug"
+	"testing"
 
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 )
 
 func MiddlewarePanic(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
-	defer func() {
-		if e := recover(); e != nil {
-			logger := LoggerFromContext(r.Context())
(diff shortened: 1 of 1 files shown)


frzifus (approved):

frontend/pkg/frontend/middleware_panic.go
```diff
@@ -7,18 +7,22 @@ import (
...
 func MiddlewarePanic(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
-	defer func() {
-		if e := recover(); e != nil {
-			logger := LoggerFromContext(r.Context())
-			logger.Error(fmt.Sprintf("panic: %#v\n%s\n", e, string(debug.Stack())))
-			arm.WriteInternalServerError(w)
-		}
-	}()
+	// Do not catch panics when running "go test".
+	if !testing.Testing() {
```
venkateshsredhat, line 17: This will ensure only for non testing envs .

venkateshsredhat (approved):

PR #1335, opened 2025-02-15
Ensure all JSON timestamps use UTC time

mbarnes:
### What this PR does

Did a quick audit of the RP code to make sure all timestamps that appear in JSON data - whether for logging or response bodies - use UTC time.  Found one case to fix and one case to remove.

### DIFF
--- a/internal/api/arm/correlation.go
+++ b/internal/api/arm/correlation.go
@@ -5,7 +5,6 @@ package arm
 
 import (
 	"net/http"
-	"time"
 
 	"github.com/google/uuid"
 )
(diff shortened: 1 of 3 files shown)


venkateshsredhat:
@mbarnes There's a CI failure .

mbarnes:
Appears to be a service lifecycle issue, unrelated to the changes here.  But I'll rerun the jobs.

hbhushan3:
Based on a quick [search](https://github.com/search?q=repo%3AAzure%2FARO-HCP%20time.Now()&type=code) in the repo, I see `time.Now()` being used in 10 files. Can we switch to` time.Now().UTC()` in all these files or probably create a ticket to look into this? Thanks.

mbarnes:
> Based on a quick [search](https://github.com/search?q=repo%3AAzure%2FARO-HCP%20time.Now()&type=code) in the repo, I see time.Now() being used in 10 files. Can we switch to time.Now().UTC() in all these files or probably create a ticket to look into this? Thanks.

All those other cases are used as timers.  You'll see they're each followed shortly by a `time.Since` call to get a duration.  The time value itself does not appear in any output: HTTP responses, logs, metrics, etc.  So the `time.UTC` call isn't needed for these cases.

hbhushan3 (approved):
/approve /lgtm

PR #1375, opened 2025-02-20
Support & enable audit logging for Key Vault and AKS for dev & cspr environments

jfchevrette:
### What this PR does

* Add a Log Analytics workspace in global
  * New config flag `log.enableLogAnalytics` - defaults to false, true for dev and cspr)  
* Add support for Key Vault audit logging via `Diagnostic Dettings`
* Enable Key Vault audit logging for the following Key Vaults in (dev & cspr only)
  * svc: svc
  * mgmt: cx, mgmt, msi
* Enable AKS audit logging for svc and mgmt clusters (dev & cspr only)
  * kube-audit
  * Container logs

Jira:
* https://issues.redhat.com/browse/ARO-15221
* https://issues.redhat.com/browse/ARO-15223
* https://issues.redhat.com/browse/ARO-15224

Suggest reviewing commit by commit

### DIFF
--- a/dev-infrastructure/modules/aks-cluster-base.bicep
+++ b/dev-infrastructure/modules/aks-cluster-base.bicep
@@ -62,6 +62,8 @@ var istioIngressGatewayIPAddressIPTagsArray = [
 @maxLength(24)
 param aksKeyVaultName string
 
+param logAnalyticsWorkspaceId string = ''
+
 // Local Params
 @description('Optional DNS prefix to use with hosted Kubernetes API server FQDN.')
 param dnsPrefix string = aksClusterName
@@ -459,6 +461,25 @@ resource aksCluster 'Microsoft.ContainerService/managedClusters@2024-04-02-previ
     aksClusterOutboundIPAddress
   ]
 }
+
+resource aksDiagnosticSettings 'Microsoft.Insights/diagnosticSettings@2017-05-01-preview' = if (logAnalyticsWorkspaceId != '') {
+  scope: aksCluster
+  name: aksClusterName
+  properties: {
+    logs: [
+      {
+        category: 'kube-audit'
+        enabled: true
+      }
+      {
+        category: 'kube-audit-admin'
+        enabled: true
+      }
+    ]
+    workspaceId: logAnalyticsWorkspaceId
+  }
+}
+
 
 resource userAgentPools 'Microsoft.ContainerService/managedClusters/agentPools@2024-04-02-preview' = [
   for i in range(0, userAgentPoolAZCount): {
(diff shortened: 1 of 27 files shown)


dev-infrastructure/modules/aks-cluster-base.bicep
```diff
@@ -460,6 +462,82 @@ resource aksCluster 'Microsoft.ContainerService/managedClusters@2024-04-02-previ
...
+      {
+        category: 'kube-audit-admin'
+        enabled: true
+      }
+    ]
+    workspaceId: logAnalyticsWorkspaceId
+  }
+}
+
+resource aksClusterDcr 'Microsoft.Insights/dataCollectionRules@2023-03-11' = if (logAnalyticsWorkspaceId != '') {
```
tony-schndr, line 483: There is already a DCR in place [here](https://github.com/Azure/ARO-HCP/blob/main/dev-infrastructure/modules/metrics/datacollection.bicep) which I _think_ we could add this configuration to.  Are any downsides to having multiple DCRs configured on cluster?

dev-infrastructure/modules/aks-cluster-base.bicep
```diff
@@ -460,6 +462,82 @@ resource aksCluster 'Microsoft.ContainerService/managedClusters@2024-04-02-previ
...
+            }
+          }
+          extensionName: 'ContainerInsights'
+        }
+      ]
+    }
+    destinations: {
+      logAnalytics: [
+        {
+          name: 'ciworkspace'
```
geoberle, line 512: what does `ci` stand for?

geoberle (commented):
nice work! 
can you enable log analytics for eventgrid as well?

dev-infrastructure/modules/aks-cluster-base.bicep
jfchevrette: I thought different dcr for different purpose. I don't think it matters from a technical standpoint whether we have one or two. I also could not tell if the name of the existing one (MSProm-...) is intentional or not and I opted to avoid renaming/re-creating existing resources.

dev-infrastructure/modules/aks-cluster-base.bicep
jfchevrette: it's the default name when creating these resources from the Azure Portal. I believe `ci` stand for `Container Insights`. It's an internal name used to link dataflows and log analytics workspaces in the DCR. I suppose there could be a use case to have multiple so the name would be used to distinguish.

jfchevrette:
> nice work! can you enable log analytics for eventgrid as well?

Added in 75ca2cf5197f0625fc3c105e35bf75aea687c6ed

dev-infrastructure/modules/aks-cluster-base.bicep
geoberle: ah understood. thank you

geoberle (dismissed):

PR #1376, opened 2025-02-20
grafana monitoring workspace cross subscription registration

geoberle:
### What this PR does

* grafana and monitoring workspaces can now exist in different subscriptions
* grafana setup extracted to dedicated modules
* cleanup of unused resources (monitoring msi)
* explicit zone redundancy setup
* ensure linked azure monitor workspaces don't get wiped during grafana reconcile
* use raw az commands to register a regional azure monitor workspace with grafana instead of the az plugin, because the plugin does not support cross subscription registration :(

Jira: https://issues.redhat.com/browse/ARO-14980
Link to demo recording:


janboll:
Are we going to have a single Grafana instance for all regions? So we'd have to configure datasources for each region?

geoberle:
> Are we going to have a single Grafana instance for all regions? So we'd have to configure datasources for each region?

yes the Grafana instance is global and the references all Azure Montor Workspaces from all regions

dev-infrastructure/templates/global-grafana.bicep
```diff
@@ -0,0 +1,58 @@
...
+
+resource grafana 'Microsoft.Dashboard/grafana@2023-09-01' = {
+  name: grafanaName
+  location: resourceGroup().location
+  sku: {
+    name: 'Standard'
+  }
+  identity: {
+    type: 'SystemAssigned'
+  }
```
jfchevrette, line 35: Could I request that you add `zoneRedundancy` here and get the desired value based on zone redundancy availability like we've done for other resources? See https://issues.redhat.com/browse/ARO-14959

geoberle: done

janboll (approved):

PR #1395, opened 2025-02-24
rm istio from personal-dev

katherinelc321:
### What this PR does

Removing Istio from personal dev clusters since we won't have arm in dev
Jira: ARO-15108
Link to demo recording:


geoberle:
i think this might be a misunderstanding.
istio serves also in-cluster communciation purposes, not just ingress. hence we can't remove it from any environment.

i think the original idea for this PR was to disable ingress in such envs as we don't have a way to implement authn with the lack of MISE outside of MSFT envs. but @tony-schndr found a way to use MISE in our dev tenant. let's put this is PR on hold until we have confirmation we are allowed to actually use MISE in this context

PR #1405, opened 2025-02-26
github: Make dependabot more efficient

mbarnes:
### What this PR does

This uses the [relatively new "directories" key](https://github.com/dependabot/dependabot-core/issues/1595) to better handle dependencies in a multi-module repository.

For example instead of dependabot submitting three separate pull requests to update the same Go module in `/frontend`, `/backend`, and `/internal`, it should now submit ONE pull request to update the Go module in all three directories.

Remains to be seen if dependabot pull requests will now update `go.work.sum`.  I will investigate further if not.


mbarnes:
Looks like workspace support for monorepos is still an open issue.
https://github.com/dependabot/dependabot-core/issues/6012

DennisJWheeler (dismissed):
Seems a reasonable consolidation.

mociarain (approved):
LGTM

SudoBrendan (approved):

PR #1410, opened 2025-02-26
added the orphaned namespace cleaner namespace definition

anshulvermapatel:
This is an addition to a previous PR where the namespace definitions was missed in the template.

### DIFF
--- a/cluster-service/deploy/integration/cluster-service-namespace.yaml
+++ b/cluster-service/deploy/integration/cluster-service-namespace.yaml
@@ -4,13 +4,19 @@ kind: Template
 metadata:
   name: cluster-service-admin
 parameters:
-- name: NAMESPACE
-  description: The namespace to create
-  required: true
-  value: cluster-service-admin
-- name: CLIENT_ID
-  description: The Azure Client ID used for federation
-  required: true
+  - name: NAMESPACE
+    description: The namespace to create
+    required: true
+    value: cluster-service-admin
+  - name: CLIENT_ID
+    description: The Azure Client ID used for federation
+    required: true
+  - name: ORPHANED_NAMESPACE_CLEANER
+    description: The namespace to create to have a cronjob which will delete the orphaned namespace which are not deleted due to any issues with the jenkins job.
+    value: orphaned-namespace-cleaner
+  - name: KUBECTL_IMAGE
+    description: An image which have the `kubectl` binary in it.
+    value: quay.io/rhn_support_ansverma/ubi8-minimal-kubectl:latest
 
 objects:
   - apiVersion: v1
@@ -27,12 +33,12 @@ objects:
     metadata:
       name: namespace-admin
     rules:
-    - apiGroups:
-      - ""
-      resources:
-      - namespaces
-      verbs:
-      - "*"
+      - apiGroups:
+          - ""
+        resources:
+          - namespaces
+        verbs:
+          - "*"
   - apiVersion: rbac.authorization.k8s.io/v1
     kind: ClusterRoleBinding
     metadata:
@@ -42,9 +48,9 @@ objects:
       kind: ClusterRole
       name: admin
     subjects:
-    - kind: ServiceAccount
-      name: cluster-service-mgmt
-      namespace: cluster-service-admin
+      - kind: ServiceAccount
+        name: cluster-service-mgmt
+        namespace: cluster-service-admin
   - apiVersion: rbac.authorization.k8s.io/v1
     kind: ClusterRoleBinding
     metadata:
@@ -54,9 +60,9 @@ objects:
       kind: ClusterRole
       name: namespace-admin
     subjects:
-    - kind: ServiceAccount
-      name: cluster-service-mgmt
-      namespace: cluster-service-admin
+      - kind: ServiceAccount
+        name: cluster-service-mgmt
+        namespace: cluster-service-admin
   - apiVersion: v1
     kind: Secret
     metadata:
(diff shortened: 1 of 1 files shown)


cluster-service/deploy/integration/cluster-service-namespace.yaml
```diff
@@ -4,13 +4,19 @@ kind: Template
...
-  description: The Azure Client ID used for federation
-  required: true
+  - name: NAMESPACE
+    description: The namespace to create
+    required: true
+    value: cluster-service-admin
+  - name: CLIENT_ID
+    description: The Azure Client ID used for federation
+    required: true
+  - name: ORPHANED_NAMESPACE_CLEANER
```
miguelsorianod, line 14: The component name is ORPHANED_NAMESPACE_CLEANER. Therefore, to designate the namespace of that component I would expect the parameter to be something like NAMESPACE_ORPHANED_NAMESPACE_CLEANER or ORPHANED_NAMESPACE_CLEANER_NAMESPACE

cluster-service/deploy/integration/cluster-service-namespace.yaml
```diff
@@ -4,13 +4,19 @@ kind: Template
 metadata:
```
miguelsorianod, line 4: What is the meaning of "integration" in this context? We are in the ARO-HCP repository, so I might expect integration to mean ARO-HCP Integration environment

miguelsorianod, line 4: What is the meaning of "cluster-service-namespace" in this context?

cluster-service/deploy/integration/cluster-service-namespace.yaml
```diff
@@ -4,13 +4,19 @@ kind: Template
...
+    value: cluster-service-admin
+  - name: CLIENT_ID
+    description: The Azure Client ID used for federation
+    required: true
+  - name: ORPHANED_NAMESPACE_CLEANER
+    description: The namespace to create to have a cronjob which will delete the orphaned namespace which are not deleted due to any issues with the jenkins job.
+    value: orphaned-namespace-cleaner
+  - name: KUBECTL_IMAGE
+    description: An image which have the `kubectl` binary in it.
+    value: quay.io/rhn_support_ansverma/ubi8-minimal-kubectl:latest
```
miguelsorianod, line 19: Is there some reason we are referencing a repository of a specific developer for the container image?

cluster-service/deploy/integration/cluster-service-namespace.yaml
```diff
@@ -27,12 +33,12 @@ objects:
...
-    - apiGroups:
-      - ""
-      resources:
-      - namespaces
-      verbs:
-      - "*"
+      - apiGroups:
+          - ""
+        resources:
+          - namespaces
```
miguelsorianod, line 39: If I understand correctly, this grants performing any actions against K8s namespace resources. Is that what we intend?

cluster-service/deploy/integration/cluster-service-namespace.yaml
```diff
@@ -72,3 +78,49 @@ objects:
...
+              serviceAccountName: orphaned-namespace-cleaner-cronjob
+              containers:
+                - name: kubectl-container
+                  image: ${KUBECTL_IMAGE}
+                  command: ["/bin/sh", "-c"]
+                  args:
+                    - |
+                      echo "Starting to clear orphaned namespaces"
+                      echo "deleting the orphaned namespaces"
+                      kubectl get namespaces -o json | jq -r '.items[] | select(.metadata.labels."sandbox-jenkins-type"=="aro-hcp") | select((now - (.metadata.creationTimestamp | fromdate)) / 60 > 60) | .metadata.name' | xargs kubectl delete namespace
```
miguelsorianod, line 124: What happens to this command if there are no items?

cluster-service/deploy/integration/cluster-service-namespace.yaml
```diff
@@ -72,3 +78,49 @@ objects:
...
+    roleRef:
+      kind: ClusterRole
+      name: namespace-admin
+      apiGroup: rbac.authorization.k8s.io
+  - apiVersion: batch/v1
+    kind: CronJob
+    metadata:
+      name: orphaned-namespace-cleaner
+      namespace: ${ORPHANED_NAMESPACE_CLEANER}
+    spec:
```
miguelsorianod, line 107: What happens if for example there's another job already running? are we limiting the concurrency of the jobs?

cluster-service/deploy/integration/cluster-service-namespace.yaml
miguelsorianod, line 124: Could you explain what's the meaning of the / 60 > 0 part and date manipulation? maybe we could add it as a comment to help readers?

cluster-service/deploy/integration/cluster-service-namespace.yaml
anshulvermapatel: I did not add this, maybe @geoberle can comment on this.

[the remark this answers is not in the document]
anshulvermapatel: No reason, the file was not that big, so I included in the same file itself.

cluster-service/deploy/integration/cluster-service-namespace.yaml
anshulvermapatel: I have addressed in the new PR: 
https://github.com/Azure/ARO-HCP/pull/1412/files
renamed it to `ORPHANED_NAMESPACE_CLEANER_NAMESPACE`

cluster-service/deploy/integration/cluster-service-namespace.yaml
anshulvermapatel: We can push this image to ACR as well once we are ready for that, till then we can use this one.

cluster-service/deploy/integration/cluster-service-namespace.yaml
anshulvermapatel: Yes, we want to get, list and delete permissions. So I used the ClusterRole which was already there in the template.

cluster-service/deploy/integration/cluster-service-namespace.yaml
anshulvermapatel: The command will return error and the job pod will get completed.

cluster-service/deploy/integration/cluster-service-namespace.yaml
anshulvermapatel: Already this job will run once a day. It is going to be very rare that a job pod will still be running.

cluster-service/deploy/integration/cluster-service-namespace.yaml
anshulvermapatel: This mean that the namespaces which are more than 1 hour old, only those namespaces will be deleted as the time out for the pr_check is 30 mins so we took double of that time.

cluster-service/deploy/integration/cluster-service-namespace.yaml
miguelsorianod: It is not for the integration env. It is for the CS PR Check environment. We should fix this

cluster-service/deploy/integration/cluster-service-namespace.yaml
anshulvermapatel: I think the error which kubectl returns is self explanatory and it gives the information that there are no namespaces which comes under the conditions on which they should be deleted. So I think it will be okay to keep it as is.

cluster-service/deploy/integration/cluster-service-namespace.yaml
anshulvermapatel: Fix as in we create another directory in `deploy` and move this template there?

cluster-service/deploy/integration/cluster-service-namespace.yaml
miguelsorianod: Although the error could be self explanatory my understanding is that it would cause the job ending in failure, which does not seem correct. Why would we mark the job as failed when there are no namespaces to clean. That's not an error situation. There's simply no namespaces to clean.

PR #1411, opened 2025-02-26
feat: add Standard_D8ps_v6 as supported instance type in CS

miguelsorianod:
### What this PR does

To have an allowed instance type that is based on arm64.

This is needed so we can deploy and test ARM64 based Node Pools.

Jira: 
Link to demo recording:


tony-schndr (approved):
LGTM, though I wonder if this is something worth extracting to `config/config.yaml` where we could have very specific node pool instance types per environment/region.

miguelsorianod:
> LGTM, though I wonder if this is something worth extracting to `config/config.yaml` where we could have very specific node pool instance types per environment/region.

Yes I was thinking that before too. I think we should probably make it more flexible in the medium term.

JameelB (approved):

PR #1412, opened 2025-02-26
added the orphaned namespace cleaner namespace definition

anshulvermapatel:
This is an addition to a previous PR where the namespace definitions was missed in the template.


shitaljante:
LGTM

miguelsorianod:
The MR looks good to me in this context but I'd like to see the namespace cleaner configuration in a different template than the other elements. Let's work on that in follow-up MRs

PR #1415, opened 2025-02-26
chore: change default demo nodepool size from 30 to 64 GiB

miguelsorianod:
### What this PR does

The reason is that we found out that for some instance types 30 GiB is not enough.

Jira: 
Link to demo recording:


tony-schndr (approved):
lgtm

PR #1419, opened 2025-02-27
backend: add /metrics endpoint

simonpasquier:
### What this PR does

Jira: 
Link to demo recording:

### DIFF
--- a/backend/deploy/helm/backend/templates/peerauthentication.yaml
+++ b/backend/deploy/helm/backend/templates/peerauthentication.yaml
@@ -1,0 +1,11 @@
+apiVersion: security.istio.io/v1beta1
+kind: PeerAuthentication
+metadata:
+  name: aro-hcp-backend-metrics
+spec:
+  selector:
+    matchLabels:
+      app: aro-hcp-backend
+  portLevelMtls:
+    8081:
+      mode: PERMISSIVE
--- a/backend/operations_scanner.go
+++ b/backend/operations_scanner.go
@@ -20,6 +20,8 @@ import (
 	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	ocmerrors "github.com/openshift-online/ocm-sdk-go/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/database"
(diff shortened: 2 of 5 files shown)


backend/deploy/helm/backend/templates/peerauthentication.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: security.istio.io/v1beta1
```
simonpasquier, line 1: this is similar to #1420

mbarnes (changes requested):
Overall looks good to me.  This sets the stage nicely for also adding a `/healthz` endpoint with leader-election integration.

I had one suggestion, but not a requirement if you don't like it or it wouldn't work.

backend/operations_scanner.go
```diff
@@ -147,11 +226,19 @@ loop:
...
 
 // collectSubscriptions builds an internal list of Azure subscription IDs by
 // querying Cosmos DB.
 func (s *OperationsScanner) collectSubscriptions(ctx context.Context, logger *slog.Logger) {
+	startTime := time.Now()
+	s.operationsCount.WithLabelValues(collectSubscriptionsLabel).Inc()
+	defer func() {
+		s.operationsDuration.WithLabelValues(collectSubscriptionsLabel).Observe(time.Since(startTime).Seconds())
+		s.lastOperationTimestamp.WithLabelValues(collectSubscriptionsLabel).SetToCurrentTime()
+	}()
```
mbarnes, line 240: Would it work to make this anonymous function a proper method so it doesn't have to keep being repeated?  Something like...
```
func (s *OperationsScanner) updateOperationMetrics(label string, startTime time.Time)
```
and then here you could call
```
defer s.updateOperationMetrics(collectSubscriptionsLabel, startTime)
```

mbarnes: Honestly what I really wish is golang had the concept of decorators like in Python, because this would be a perfect use case for it.

simonpasquier: good idea! I've added a slight modification to your suggestion.

backend/operations_scanner.go
```diff
@@ -147,11 +226,24 @@ loop:
...
+}
+
+func (s *OperationsScanner) updateOperationMetrics(label string) func() {
+	startTime := time.Now()
+	s.operationsCount.WithLabelValues(label).Inc()
+	return func() {
+		s.operationsDuration.WithLabelValues(label).Observe(time.Since(startTime).Seconds())
+		s.lastOperationTimestamp.WithLabelValues(label).SetToCurrentTime()
+	}
 }
```
mbarnes, line 239: I like!

backend/operations_scanner.go
```diff
@@ -147,11 +226,24 @@ loop:
...
 
 // collectSubscriptions builds an internal list of Azure subscription IDs by
 // querying Cosmos DB.
 func (s *OperationsScanner) collectSubscriptions(ctx context.Context, logger *slog.Logger) {
+	done := s.updateOperationMetrics(collectSubscriptionsLabel)
+	defer done()
```
mbarnes, line 245: nit: Could we compact this even further to
```
defer s.updateOperationMetrics(collectSubscriptionsLabel)()
```

mbarnes (changes requested):
Love it.  Had a further suggestion but, again, not required.

backend/operations_scanner.go
simonpasquier: Done. The API is a bit error-prone because `defer s.updateOperationMetrics(collectSubscriptionsLabel)` compiles though it's not doing what you expect but it's probably fine if the service doesn't run new types of operation in the future.

mbarnes: I'm adding a new operation now, actually, but I'll be careful to abide by this.

mbarnes (approved):

PR #1420, opened 2025-02-27
frontend: add PeerAuthentication resource for metrics

simonpasquier:
### What this PR does

This change disables mTLS for the metrics port otherwise the collection fails because Envoy rejects the scrape request. It is similar to what exists for the Maestro server.

After applying the manifest, metrics are ingested in managed Prometheus.

image

Jira: 
Link to demo recording:


simonpasquier:
cc @mmazur @tony-schndr

tony-schndr (approved):
Thanks for fixing this!

PR #1422, opened 2025-02-27
test: Start of the E2E test suite and basis of E2E docs

patriksuba:
### What this PR does
Add test directory with README and initialize E2E framework with client which connects to development environment. Create basic tests (cluster list and get).

Jira: [ARO-14974](https://issues.redhat.com/browse/ARO-14974), [ARO-15274](https://issues.redhat.com/browse/ARO-15274) 
Link to demo recording: 
Scope: Deliver minimal framework code to demonstrate connection of client to development environment.


test/e2e/Makefile
```diff
@@ -0,0 +1,5 @@
+SHELL = /bin/bash
+
+run:
+	go run github.com/onsi/ginkgo/v2/ginkgo run --tags E2Etests --junit-report ./report-$(shell date --iso-8601).xml ./
```
mbukatov, line 4: Here I would rather use a timestamp with at least minute precision, so something like `date -Iminute` or sheer unix epoch timestamp like `date +'%s'`, so that someone rerunning tests locally don't accidently overwrite previous test results.

```
$ date -Iminute
2025-03-20T17:07+01:00
```

```
$ date +'%s'
1742486970
```

mbukatov (commented):
Thank you for the initial code.

test/go.mod
```diff
@@ -0,0 +1,39 @@
...
+	github.com/google/go-cmp v0.7.0 // indirect
+	github.com/google/pprof v0.0.0-20241210010833-40e02aabc2ad // indirect
+	golang.org/x/net v0.35.0 // indirect
+	golang.org/x/sys v0.30.0 // indirect
+	golang.org/x/text v0.22.0 // indirect
+	golang.org/x/tools v0.28.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace github.com/Azure/ARO-HCP/internal => ../internal
```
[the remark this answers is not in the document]
patriksuba: Without replace line, go mod tidy command is trying to get ARO-HCP/internal module from public repository which will result in error. Code of internal module is importable only by code in the directory tree rooted at the parent of "internal", so replace line has to be present in order to use package outside of internal module and inside of its parent directory.

mbukatov (approved):

mbarnes (approved):

test/e2e/README.md
```diff
@@ -0,0 +1,68 @@
+# E2E Testing
+
+The E2E test suite will work in every environment of the ARO-HCP project. Its main purpose is to ensure specific functionality based on the environment and its usage. You can find more details about the test suite design in [ARO-12804](https://issues.redhat.com/browse/ARO-12804).
+
+- The test suite client connects to the RP frontend in the development environment using port forwarding to localhost.
+- In the integration environment, the client will connect to the RP frontend using a public IP (when available) but only within the MSFT corporate network.
```
SudoBrendan, line 6: We should be able to use ARM at this point as well.

test/e2e/get_test.go
```diff
@@ -0,0 +1,32 @@
...
+	var (
+		clustersClient *api.HcpOpenShiftClustersClient
+	)
+
+	BeforeEach(func() {
+		By("Prepare HCP clusters client")
+		clustersClient = clients.NewHcpOpenShiftClustersClient()
+	})
+
+	It("Get non existing cluster", labels.Medium, labels.Negative, func(ctx context.Context) {
```
SudoBrendan, line 24: nit/opinion: It's a norm in Ginkgo and other behavioral test suites to make test cases read like sentences. This makes it so even lay people can understand them and what expectations of the code are. Some development teams even put the ownership of these sentences onto their BU ("give me a test case sentence, I'll write up the code for it").

As written, our complete test case sentence reads like: "Get operation: it get non existing cluster by prepare HCP clusters client (and) by Send get request for cluster", which does get the point across, but is a bit garbled. We could reframe this to a comprehensive and correct sentence given our test case: "Get HCPOpenShiftCluster: it should fail to get a nonexistent cluster with a Not Found error by preparing HCP clusters client (and) by sending a GET request for the nonexistent cluster"

Without reading any code, just the test case, I can tell what we intend to do with this code (and it becomes very obvious when the code _doesn't_ do that). This is especially helpful when tests _fail later,_ and I'm writing some feature way in the future that breaks the test and I see this sentence in the logs.

Given that we're greenfielding this entire suite - now feels like a good time to set that norm for our team: all E2E test cases should read like sentences.

SudoBrendan (approved):
I only have nits and opinions here, I think we would be well-served to follow Behavioral Test norms for our test descriptions. Generally, LGTM as an MVP, if we want to add this stuff later that's fine too. I am happy to merge if you don't want to address what I've written or it feels unnecessary/unhelpful - just give me a ping :)

SudoBrendan (approved):

PR #1427, opened 2025-02-28
[Logs] Add logs values in config

whober0521:
[ARO-5569](https://issues.redhat.com/browse/ARO-5569)

### What

- config fields for MDSD and log account setup
- log account pipeline step schema adaptions to changes in https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/hcp/pullrequest/11859510?_a=files

### Why

To run logs account & release pipelines


tony-schndr (approved):
LGTM

PR #1404, opened 2025-02-25
[API] Remove External Auth

jonathan34c:
### What this PR does
Remove External Auth per API review doc : https://docs.google.com/document/d/1k6sqn7uFe46pnweQU_BPu07ux6U_gAiZPVYdGKKCKww/edit?tab=t.0
JIRA ticket: 
https://issues.redhat.com/browse/ARO-15336

Fields that I removed
```
ExternalAuthConfigProfile
ExternalAuthProfile
TokenIssuerProfile
ExternalAuthClientProfile
ExternalAuthClaimProfile
ExternalAuthClientComponentProfile
TokenClaimMappingsProfile
TokenClaimValidationRuleProfile
ClaimProfile
```

Jira: 
Link to demo recording:

PR #1380, opened 2025-02-20
[API]Remove proxy and child resources

jonathan34c:
### What this PR does
Remove ClusterSpec.Proxy, ClusterPatchSpec.Proxy and child resources, may be re-added in the future.
Jira Ticket: https://issues.redhat.com/browse/ARO-15334

Jira: 
Link to demo recording:

PR #1434, opened 2025-03-03
adapt demo script cluster template to ARM cluster payload changes

geoberle:
### What

the cluster payload has been changed during the ARM API review and the demo payload we use for cluster cretion does not match that schema

* https://github.com/Azure/ARO-HCP/pull/1404
* https://github.com/Azure/ARO-HCP/pull/1380


miguelsorianod:
lgtm

PR #1437, opened 2025-03-03
backend: Process node pool deletions

mbarnes:
### What

Node pool status is not available yet, but we can at least check for "404 Not Found" when fetching node pools from Cluster Service.

### Why

I've had several complaints about not being able to reuse node pool names through the RP following a cluster or node pool deletion.  This is because a Cosmos DB document for the deleted node pool was being left behind.

Jira: This is somewhat by-request, but also related to [ARO-5737 - Implement Async API ARM Requirements for Long-Running Operations](https://issues.redhat.com/browse/ARO-5737)


DennisJWheeler (approved):
Looks good to me.

PR #1438, opened 2025-03-04
Init deployment of automation account

janboll:
[ARO-15556](https://issues.redhat.com/browse/ARO-15556)

### What

Create bicep template for already deployed automation accounts. Makes management easier and allows reuse of other purposes. 

### Why

Want to use this for nightly rebuild of infra.


tony-schndr (approved):
Nice work!  LGTM.

PR #1440, opened 2025-03-04
*: use -typed for mockgen

stevekuznetsov:
There's little utility to generating loosey-goosey mocks that don't even assert types at compile time. Getting feedback on types there instead of runtime is strictly better.


mbarnes (approved):
Nice.  I wasn't aware of that option and I'm surprised it's not the default.

Looks like it already caught an issue in the backend tests.
(Anonymous function for `Do` needs an `error` return.)

geoberle (approved):

mbarnes (approved):
Makefile change breaks my ability to merge, but still LGTM!

SudoBrendan (approved):

PR #1441, opened 2025-03-04
[API] Add hcpOpenshiftCluster.Properties.Capabilities.Disabled 

sanchezl:
### What

Add `hcpOpenshiftCluster.Properties.Capabilities.Disabled` to resource provider API and map the values to OCM API's `Cluster.Azure.Capabilties.Disabled` attribute.

### Why

To allow the creation of hosted clusters with certain OpenShift capabilities disabled. Only `ImageRegistry` is supported at first.

OCM API changes:
* https://github.com/openshift-online/ocm-api-model/pull/1020
* https://github.com/openshift-online/ocm-sdk-go/pull/1029


api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -90,6 +90,25 @@ model HcpOpenShiftClusterProperties {
...
+
+/** Cluster capabilities configuration. */
+model ClusterCapabilitiesProfile {
+  /** Disabled cluter capabilities. */
+  @visibility("create", "read")
+  disabled?: OptionalClusterCapability[];
+}
+
+/** Cluster capabilities that can be disabled. */
+union OptionalClusterCapability {
```
vkareh, line 107: Is there a reason to call this "Optional"? Is that convention or just preference?

go.work
```diff
@@ -11,3 +11,5 @@ use (
 	./tooling/templatize
 	tooling/prometheus-rules
 )
+
+replace github.com/openshift-online/ocm-sdk-go => github.com/sanchezl/ocm-sdk-go v0.0.0-20250228013841-940a99155963
```
vkareh, line 15: Is this https://github.com/openshift-online/ocm-sdk-go/pull/1029 ?

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
sanchezl: This is not an exhaustive list of capabilities. Only the capabilities defined as optional by HCP.

internal/api/hcpopenshiftcluster_test.go
```diff
@@ -35,6 +36,7 @@ func newTestValidator() *validator.Validate {
 		arm.ManagedServiceIdentityTypeSystemAssigned,
 		arm.ManagedServiceIdentityTypeSystemAssignedUserAssigned,
 		arm.ManagedServiceIdentityTypeUserAssigned))
+	validate.RegisterAlias("enum_optionalclustercapability", EnumValidateTag(generated.PossibleOptionalClusterCapabilityValues()...))
```
mbarnes, line 39: nit: There's nothing technically wrong with this, but for consistency's sake I'd prefer the unversioned enum here (`OptionalClusterCapabilityImageRegistry`).

internal/api/v20240610preview/hcpopenshiftclusters_methods.go
```diff
@@ -125,6 +126,18 @@ func (v version) NewHCPOpenShiftCluster(from *api.HCPOpenShiftCluster) api.Versi
...
+	out := &generated.ClusterCapabilitiesProfile{
+		Disabled: make([]*generated.OptionalClusterCapability, len(from.Disabled)),
+	}
+
+	for index, item := range from.Disabled {
+		out.Disabled[index] = api.Ptr(generated.OptionalClusterCapability(item))
+	}
+
+	return out
+}
```
mbarnes, line 139: note to self: `api.StringSliceToStringPtrSlice` would not work here as is because of the string subtypes, but look into making it generic with `~string` constraints for the input and output types.

mbarnes (changes requested):
One nit (below) and one missing piece:

You still need to update the `Normalize` method in `internal/api/v20240610preview/hcpopenshiftclusters_methods.go`, which converts from _versioned_ to _unversioned_ clusters.  Basically the reverse of `NewHCPOpenShiftCluster`.

internal/api/utils.go
```diff
@@ -17,6 +18,20 @@ func DeleteNilsFromPtrSlice[S ~[]*E, E any](s S) S {
...
+	return func(yield func(int, *E) bool) {
+		for i := range a {
+			if a[i] != nil {
+				if !yield(i, a[i]) {
+					return
+				}
+			}
+		}
+	}
+}
```
mbarnes, line 33: This is great!  Would you mind renaming it to something like `NonNilSlice` because I'd like to write a similar function for maps.

mbarnes (changes requested):
Requested a rename but otherwise looks great!

internal/api/utils.go
sanchezl: Updated to `NonNilSliceValues`.

mbarnes (approved):
Approved.  Thanks for humoring me on the generic test functions!

PR #1443, opened 2025-03-04
[API] Correct the Visibility of all non-secret objects

jonathan34c:
correct the Visibility of all non-secret objects so that they are “create”, “read”, “update” rather than “create”, “update”.

https://issues.redhat.com/browse/ARO-15339

Per https://docs.google.com/document/d/1k6sqn7uFe46pnweQU_BPu07ux6U_gAiZPVYdGKKCKww/edit?tab=t.0, review and correct the Visibility of all non-secret objects so that they are “create”, “read”, “update” rather than “create”, “update”.

### DIFF
--- a/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
+++ b/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
@@ -78,7 +78,7 @@ model HcpOpenShiftClusterProperties {
   api: ApiProfile;
 
   /** Disable user workload monitoring */
-  @visibility("create", "update")
+  @visibility("create", "update", "read")
   disableUserWorkloadMonitoring?: boolean = false;
 
   /** Azure platform configuration */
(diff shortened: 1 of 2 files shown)


mbarnes (changes requested):
I think you misread the Jira.  Sorry if it wasn't stated clearly enough.

We just need to ensure all `@visibility` attributes include `"read"`.  The absence of `"read"` implies the field is a secret, which nothing in the current API should be.  All fields should be readable.

Do not add `"update"` where it's not already present, as this alters which fields can be patched on day-2.

jonathan34c:
> I think you misread the Jira. Sorry if it wasn't stated clearly enough.
> 
> We just need to ensure all `@visibility` attributes include `"read"`. The absence of `"read"` implies the field is a secret, which nothing in the current API should be. All fields should be readable.
> 
> Do not add `"update"` where it's not already present, as this alters which fields can be patched on day-2.

PR updated

PR #1461, opened 2025-03-05
harmonize paths for helm charts within service directories

geoberle:
### What

most services followed already some sort of `$service/deploy/<chart content>` structure, but not every service exactly the same. this PR establishes consistency on this pattern and documents it


janboll (dismissed):
thank you for cleaning this up :bouquet:

stevekuznetsov:
Will there be anything other than `helm` in the `deploy` directories? What about anything other than `service-name` in the `helm` directories? I know it's "idiomatic" but having two levels of directory nesting that do nothing could also be seen as a waste of time ;)

geoberle:
> Will there be anything other than `helm` in the `deploy` directories? What about anything other than `service-name` in the `helm` directories? I know it's "idiomatic" but having two levels of directory nesting that do nothing could also be seen as a waste of time ;)

i agree on the `helm` intermediary not being of big value. there could be other stuff but we don't have anything else right now. dropping `helm` would leave us with `$service/deploy/$service` in most cases which for sure looks awkward. we have some situations where we have more than one chart per service, but we could solve that differently, e.g.

* `acm/multicluster-engine/deploy/<chart goes here>`
* `acm/multicluster-engine-config/deploy/<chart goes here>`
* `acm/clc-state-metrics/deploy/<chart goes here>`
* `maestro/server/deploy/<chart goes here>`
* `maestro/agent/deploy/<chart goes here>`
* `maestro/agent-registration/deploy/<chart goes here>`

this way we have less no-value dirs but still a common structure. wdyt?

stevekuznetsov:
I like the suggestion!

geoberle:
Okidoki. I'll adapt

PR #1472, opened 2025-03-09
pipeline and ev2 deployment docs

geoberle:
https://issues.redhat.com/browse/ARO-15617

### What

* document existing pipelines (review by looking at the rendered doc because of the table)
* describe ev2 deployment + ADO pipelines (review by looking at the rendered doc becaue of the mermaid chart)


dev-infrastructure/global-pipeline.yaml
```diff
@@ -1,3 +1,12 @@
+#
+# Purpose: Manage global infrastructure for ARO HCP
```
janboll, line 2: Could add a required `description` field and compile it into an additional document instead of comments. WDYT?

docs/pipelines.md
```diff
@@ -0,0 +1,24 @@
+# ARO HCP Pipelines
```
janboll, line 1: This is destined to be outdated soon. I think we should generate this info from the Pipeline files instead. ADO Pipelines information could be part of some mapping file. (or we can get a token to retrieve this info?)

geoberle: good point. i've created a ticket to track this
https://issues.redhat.com/browse/ARO-15637

janboll (approved):

docs/ev2-deployment.md
```diff
@@ -0,0 +1,107 @@
+# Deployment via EV2
+
+This document describes the deployment process of ARO HCP instances into Microsoft tenants using EV2 tighing in all artifact types and configuration.
```
mmazur, line 3:
```suggestion
This document describes the deployment process of ARO HCP instances into Microsoft tenants using EV2 tying in all artifact types and configuration.
```

docs/ev2-deployment.md
```diff
@@ -0,0 +1,107 @@
...
+
+> [!NOTE]
+> Please note that this document does not cover the following details yet
+>
+> * using the EV2 portal
+> * AME deployments yet
```
mmazur, line 9:
```suggestion
> * AME deployments
```

PR #1482, opened 2025-03-11
Revist secret sync configuration

janboll:
### What

create file hierarchy of secrets and with this use the file name information to infer secret names from it.


config/config.yaml
```diff
@@ -162,8 +162,7 @@ defaults:
     rg: hcp-underlay-{{ .ctx.regionShort }}-imagesync
     acrRG: global
     environmentName: aro-hcp-image-sync
-    # secretsToSync: "dev_global_imagesync_bearer-secret.enc;dev_global_imagesync_component-sync-pull-secret.enc;"
-    secretsToSync: ""
+    secretsToSync: ewogICAgImJlYXJlci1zZWNyZXQiOiAiZGV2X2dsb2JhbF9pbWFnZXN5bmNfYmVhcmVyLXNlY3JldC5lbmMiLAogICAgImNvbXBvbmVudC1zeW5jLXB1bGwtc2VjcmV0IjogImRldl9nbG9iYWxfaW1hZ2VzeW5jX2NvbXBvbmVudC1zeW5jLXB1bGwtc2VjcmV0LmVuYyIKfQ==
```
geoberle, line 165: i would prefer to not have unreadable config values.

i wonder if we can reduce the need for configuration a lot. e.g. group secrets to sync to an env/KV in directory structures and name the files exactly as the secrets will be named in the KV. the config could be something as simple as "what is the dir and kv to process"

config/config.yaml
```diff
@@ -162,8 +162,7 @@ defaults:
     rg: hcp-underlay-{{ .ctx.regionShort }}-imagesync
     acrRG: global
     environmentName: aro-hcp-image-sync
-    # secretsToSync: "dev_global_imagesync_bearer-secret.enc;dev_global_imagesync_component-sync-pull-secret.enc;"
-    secretsToSync: ""
+    secretsToSync: "dev/global/arohcp-imagesync-dev/bearer-secret.enc,dev/global/arohcp-imagesync-dev/component-sync-pull-secret.enc"
```
geoberle, line 165: how about one more simplification
`secretsToSync: "dev/global/arohcp-imagesync-dev"`

the script can find the files in the directory itself

janboll: Hm, I'd prefer explicit configuration, i.e. to control secret rotation

tooling/secret-sync/decrypt-all.sh
```diff
@@ -29,7 +29,8 @@ if [[ ${DRY_RUN} == "true" ]]; then
     command="echo"
 fi
 
-while  read -d';' line
+echo ${SECRETS} | tr ',' '\n' | while  read line
```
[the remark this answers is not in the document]
janboll: Changed it to simpler approach. I updated the description

config/config.yaml
geoberle: hmm. if you don't want a secret to be updated on the target, don't update it in the source.

please elaborate on how a secrets get updated in the ARO-HCP repo

geoberle: let's discuss the update process for secrets first

janboll: Two ways to update:

- Create a new secret with a suffix/prefix (i.e. v2), run the pipeline; update the secret reference to use new version
or
- Update the secret and run the pipeline; perhaps recycle pods if needed

geoberle (approved):

PR #1485, opened 2025-03-11
Add ID of MSI Mock to cluster-service-config

janboll:
### Why

Cluster service PR scripts need this ID during runtime. However it can not look it up, cause this would require elevated AD permissions. If read was possible we'd still need to pass information on the name, cause the name might change, i.e. if we have to recreate the SP. So simply passing the ID here to proceed.


jfchevrette (approved):

machi1990:
I think there are some overlaps with https://github.com/Azure/ARO-HCP/pull/1544?

PR #1490, opened 2025-03-12
Add Mise msft config

tony-schndr:
### What

Updates mise deployment to use proper configuration in Microsoft INT environment.


config/config.msft.yaml
```diff
@@ -144,10 +144,9 @@ defaults:
   mise:
     deploy: true
     azureAdInstance: https://login.microsoftonline.com/
-    azureAdClientId: ""
+    azureAdClientId: 7f4a113a-c61d-412a-bea1-85dee5baf4a8 # ARO-HCP 1p App ID
```
geoberle, line 147: This is the FPA app ID but the property is named client ID. Is that intended?

tony-schndr: updated

geoberle (dismissed):

geoberle (dismissed):

geoberle (dismissed):

geoberle (approved):

PR #1494, opened 2025-03-13
Add a make target to run mega lint

janboll:
### What

Add mega lint to makefile

### Why

Makes it easier to change yaml/bash and test locally


stevekuznetsov (approved):
Works for me

PR #1496, opened 2025-03-13
Adding healthz endpoint to backend 

venkateshsredhat:
### What
Adding healthz endpoint to backend and exposing the healthz metric as well  . This will be used in the Leader Election logic of Backend .

### Why
Currently the Backend Leader election doesn't make use of the watchdog feature due to the lack of healthz endpoint . This PR should fix the problem .

Currently Testing it . I will update this PR as i test .

### DIFF
--- a/backend/deploy/templates/backend.deployment.yaml
+++ b/backend/deploy/templates/backend.deployment.yaml
@@ -51,6 +51,8 @@ spec:
           ports:
             - containerPort: 8081
               protocol: TCP
+            - containerPort: 8080
+              protocol: TCP              
           resources:
             limits:
               memory: 1Gi
@@ -65,5 +67,18 @@ spec:
             runAsNonRoot: true
             seccompProfile:
               type: RuntimeDefault
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 8080
+            initialDelaySeconds: 15
+            periodSeconds: 20
+            failureThreshold: 3
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: 8080
+            initialDelaySeconds: 5
+            periodSeconds: 10              
       restartPolicy: Always
       terminationGracePeriodSeconds: 30
--- a/backend/main.go
+++ b/backend/main.go
@@ -32,6 +32,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 
+	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/database"
 )
 
@@ -51,6 +52,7 @@ var (
 	argClustersServiceURL   string
 	argInsecure             bool
 	argMetricsListenAddress string
+	argPortAddress          string
 
 	processName = filepath.Base(os.Args[0])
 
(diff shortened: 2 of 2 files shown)


backend/main.go
```diff
@@ -251,3 +282,16 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func backendCheckReady(ctx context.Context, dbClient database.DBClient, logger *slog.Logger) bool {
+	var ready atomic.Bool
+	// Verify the DB is available and accessible
+	if err := dbClient.DBConnectionTest(ctx); err != nil {
```
mbarnes, line 289: We had some debate awhile back [in an old pull request](https://github.com/Azure/ARO-HCP/pull/529) about what the `/healthz` endpoint in the frontend should actually do.  I'm not sure if full agreement was ever reached but one consensus was the health of the pod should _**not**_ rely on dependencies like Cosmos DB.  I think this still needs revisited in the frontend.

I think until integration with leader election is completed the `/healthz` endpoint should just return "200 OK" unconditionally.

venkateshsredhat: Some Interesting Conversation there , so I will get rid of the DB connection setup stuff but the reason I thought was backend is pretty much dependent on the DB . 
So Healthz endpoint should simply return 200 as soon as the server is up ? not be bothered about leader election ?

mbarnes: Well your PR is missing any actual [leader election integration](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#HealthzAdaptor).  I had assumed that was intentional, with the integration to be added later as implied in the description.
> This will be eventually used in the Leader Election logic of Backend .

The `/healthz` endpoint should ultimately just call [HealthzAdaptor.Check](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#HealthzAdaptor.Check).

mbarnes:
Would like @simonpasquier's input as well since this adds a Prometheus metric.

backend/main.go
```diff
@@ -165,6 +170,32 @@ func Run(cmd *cobra.Command, args []string) error {
 
 	group, ctx := errgroup.WithContext(context.Background())
 
+	// Handle requests directly for /healthz endpoint
+	if argPortListenAddress != "" {
+		backendHealthGauge := promauto.With(prometheus.DefaultRegisterer).NewGauge(prometheus.GaugeOpts{Name: "backend_healthz_state", Help: "backend_healthz_state is 1 when healthy"})
```
simonpasquier, line 175: we have `frontend_health` in the RP frontend. Maybe we should be consistent (one way or the other)?

mbarnes: Agree.

venkateshsredhat: sure , thanks !. Will change it .

backend/main.go
venkateshsredhat: yup that was intentional  , Idea was to have this until I figure out how to integrate this with the Leaderelection HealthzAdaptor check .

venkateshsredhat: @mbarnes I have added [HealthzAdaptor.Check](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#HealthzAdaptor.Check) .

backend/deploy/templates/backend.deployment.yaml
```diff
@@ -64,6 +66,19 @@ spec:
...
+              port: 8083
+            initialDelaySeconds: 15
+            periodSeconds: 20
+            failureThreshold: 3
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: 8083
+            initialDelaySeconds: 5
+            periodSeconds: 10
```
mbarnes, line 82: One last nit is we need both probes since it's the same endpoint and therefore same criteria.  I'm not sure which one to recommend.  I think I would slightly favor `readinessProbe`.

mbarnes (changes requested):
One last requested change.  The rest looks good.

backend/deploy/templates/backend.deployment.yaml
venkateshsredhat: Thanks , @mbarnes done .

backend/deploy/templates/backend.deployment.yaml
```diff
@@ -65,5 +67,11 @@ spec:
           runAsNonRoot: true
           seccompProfile:
             type: RuntimeDefault
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 8083
+          initialDelaySeconds: 5
+          periodSeconds: 10
```
mbarnes, line 75: It looks like there's trailing spaces on this line.  That's probably what's tripping the CI failures.

venkateshsredhat:
https://github.com/Azure/ARO-HCP/pull/1534 Opened this PR to cover these changes .

PR #1499, opened 2025-03-13
Added a separate template for namespace cleaner

anshulvermapatel:
Fixes: [ARO-15555](https://issues.redhat.com/browse/ARO-15555)

### What

This separates the template for the namespace cleaner and modified the cronjob to handle the error gracefully when 0 namespaces are there for deletion.

### DIFF
--- a/cluster-service/cspr/namespace-cleaner.yaml
+++ b/cluster-service/cspr/namespace-cleaner.yaml
@@ -1,0 +1,77 @@
+---
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: namespace-cleaner
+parameters:
+  - name: ORPHANED_NAMESPACE_CLEANER_NAMESPACE
+    description: The namespace to create to have a cronjob which will delete the orphaned namespace which are not deleted due to any issues with the jenkins job.
+    value: orphaned-namespace-cleaner
+  - name: NAMESPACE_CLEANER_CLUSTERROLE_NAME
+    value: namespace-cleaner
+  - name: KUBECTL_IMAGE
+    description: An image which have the `kubectl` binary in it.
+    value: quay.io/rhn_support_ansverma/ubi8-minimal-kubectl:latest
+objects:
+  - apiVersion: rbac.authorization.k8s.io/v1
+    kind: ClusterRole
+    metadata:
+      name: ${NAMESPACE_CLEANER_CLUSTERROLE_NAME}
+    rules:
+      - apiGroups:
+          - ""
+        resources:
+          - namespaces
+        verbs: ["get", "list", "delete", "watch"]
+  - apiVersion: v1
+    kind: Namespace
+    metadata:
+      name: ${ORPHANED_NAMESPACE_CLEANER_NAMESPACE}
+  - apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: orphaned-namespace-cleaner-cronjob
+      namespace: ${ORPHANED_NAMESPACE_CLEANER_NAMESPACE}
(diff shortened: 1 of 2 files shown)


cluster-service/cspr/namespace-cleaner.yaml
```diff
@@ -0,0 +1,77 @@
...
+            spec:
+              serviceAccountName: orphaned-namespace-cleaner-cronjob
+              containers:
+                - name: kubectl-container
+                  image: ${KUBECTL_IMAGE}
+                  command: ["/bin/sh", "-c"]
+                  args:
+                    - |
+                      echo "Starting to clear orphaned namespaces"
+                      # `select((now - (.metadata.creationTimestamp | fromdate)) / 60 > 60)` selects the namespaces which are older than 60 minutes.
```
miguelsorianod, line 68: The comment doesn't seem to match the executed command

cluster-service/cspr/namespace-cleaner.yaml
```diff
@@ -0,0 +1,77 @@
+---
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: namespace-cleaner
```
miguelsorianod, line 5: Should it be named orphaned-namespace-cleaner?

cluster-service/cspr/namespace-cleaner.yaml
```diff
@@ -0,0 +1,77 @@
...
+parameters:
+  - name: ORPHANED_NAMESPACE_CLEANER_NAMESPACE
+    description: The namespace to create to have a cronjob which will delete the orphaned namespace which are not deleted due to any issues with the jenkins job.
+    value: orphaned-namespace-cleaner
+  - name: NAMESPACE_CLEANER_CLUSTERROLE_NAME
```
miguelsorianod, line 10: Same comment as https://github.com/Azure/ARO-HCP/pull/1499/files#r1995376661

cluster-service/cspr/namespace-cleaner.yaml
```diff
@@ -0,0 +1,77 @@
+---
```
miguelsorianod, line 1: Now we have a new OCP Template, but is it being deployed? I only see the addition of the new template file

cluster-service/cspr/namespace-cleaner.yaml
anshulvermapatel: Yes, I have only mentioned the part which selects the namespaces based on how old they are so that anyone looking at the definition, understand that it is for the namespaces older than 1 hour(60 mins).

cluster-service/cspr/namespace-cleaner.yaml
anshulvermapatel: Yeah, okay, I'll change it to that.

cluster-service/cspr/namespace-cleaner.yaml
anshulvermapatel: yes, for now, I have added the template only for you to take a look, I will make the changes to have it deployed also.

cluster-service/cspr/namespace-cleaner.yaml
miguelsorianod: Yes but what I meant is that that part doesn't match the command either. One shows /60 > 60 and the other shows /1 > 1

anshulvermapatel: Oh, I am sorry, Let me correct it.

anshulvermapatel:
New PR created since this one was from a forked repo. I have addressed the change requests in the new PR, please continue on new one.
https://github.com/Azure/ARO-HCP/pull/1536

PR #1468, opened 2025-03-06
[API] Remove NodePool TuningConfigs

jonathan34c:
Per https://docs.google.com/document/d/1k6sqn7uFe46pnweQU_BPu07ux6U_gAiZPVYdGKKCKww/edit?tab=t.0, remove the NodePoolSpec.tuningConfigs array since we haven't implemented the feature yet.

Jira: https://issues.redhat.com/browse/ARO-15668

PR #1510, opened 2025-03-14
More Azure API adjustments

mbarnes:


api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -377,22 +379,6 @@ model NodePoolProperties {
...
-   * - https://github.com/openshift/cluster-node-tuning-operator/tree/b41042d42d4ba5bb2e99960248cf1d6ae4935018/pkg/apis/performanceprofile/v2
-   */
-  /** Tuning configs, TODO provide meaningful explanation
-   * TuningConfig is a list of references to ConfigMaps containing serialized
-   * Tuned resources to define the tuning configuration to be applied to
-   * nodes in the NodePool.
-   * Each ConfigMap must have a single key named "tuned" whose value is the
-   * JSON or YAML of a serialized Tuned or PerformanceProfile.
-   */
-  tuningConfigs?: string[];
```
jharrington22, line 395: Did we decided if this was a requirement?

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.RedHatOpenShift;
 model HcpOpenShiftClusterResource
   is TrackedResource<HcpOpenShiftClusterProperties> {
   /** Name of HCP cluster */
-  @pattern("^[a-zA-Z][a-zA-Z0-9-]$")
+  @pattern("^[a-zA-Z][-a-zA-Z0-9]+[a-zA-Z0-9]$")
```
jharrington22, line 23: I believe this will fail but `^[a-zA-Z0-9]([-0-9A-Za-z])+$` or `^[a-zA-Z0-9]([-0-9A-Za-z]+){4}$` would be more precise.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
SudoBrendan: Async API Review showcased it as removed.

SudoBrendan: https://github.com/Azure/ARO-HCP/pull/1468 was tracking this too.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mbarnes: I wrote it this way to eliminate any trailing hypen(s), which Cluster Service does not accept for a cluster name.  (Has to be a valid [RFC 1035 \<label\>](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1).)

Wasn't sure whether to expose this restriction in the ARM API.  But if we don't we'll potentially have to truncate trailing hyphens from the ARM resource name for Cluster Service (e.g. `myCluster-----` in ARM becomes `myCluster` in CS).  Perhaps we don't care?

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -463,9 +380,24 @@ model Label {
...
 
+/** Versions represents an OpenShift version. */
+model NodePoolVersionProfile {
+  /** ID is the unique identifier of the version. */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  id?: string;
+
+  /** ChannelGroup is the name of the set to which this version belongs. Each version belongs to only a single set. */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  channelGroup?: string = "stable";
```
vkareh, line 391: This should also include `Lifecycle.Update`

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -146,28 +96,28 @@ union ProvisioningState {
...
-  @visibility("create", "read")
-  id: string;
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  id?: string;
 
   /** ChannelGroup is the name of the set to which this version belongs. Each version belongs to only a single set. */
-  @visibility("create", "read")
-  channelGroup: string;
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  channelGroup?: string = "stable";
```
vkareh, line 104: Same here, this should include `Lifecycle.Update`

internal/api/hcpopenshiftcluster.go
```diff
@@ -93,10 +93,19 @@ func NewDefaultHCPOpenShiftCluster() *HCPOpenShiftCluster {
 			Type: arm.ManagedServiceIdentityTypeNone,
 		},
 		Properties: HCPOpenShiftClusterProperties{
+			Version: VersionProfile{
+				ChannelGroup: "stable",
+			},
```
bennerv, line 98: Do we want to set these if CS sets them already?

internal/api/hcpopenshiftclusternodepool.go
```diff
@@ -49,12 +53,19 @@ type NodePoolAutoScaling struct {
...
+	Key    string `json:"key,omitempty"    validate:"required_for_put,min=1,max=316"`
+	Value  string `json:"value,omitempty"  validate:"omitempty,min=1,max=63"`
 }
 
 func NewDefaultHCPOpenShiftClusterNodePool() *HCPOpenShiftClusterNodePool {
 	return &HCPOpenShiftClusterNodePool{
-		Properties: HCPOpenShiftClusterNodePoolProperties{},
+		Properties: HCPOpenShiftClusterNodePoolProperties{
+			Version: NodePoolVersionProfile{
+				ChannelGroup: "stable",
```
bennerv, line 64: Same for this.  Do we want to set defaults on this if CS already does?

internal/api/v20240610preview/nodepools_methods.go
```diff
@@ -8,11 +8,11 @@ import (
 	"github.com/Azure/ARO-HCP/internal/api/v20240610preview/generated"
 )
 
-type HcpOpenShiftClusterNodePoolResource struct {
-	generated.HcpOpenShiftClusterNodePoolResource
+type NodePool struct {
```
bennerv, line 11: if we're changing the struct type to just node pools, do we want to change the internal API files to just nodepools?

internal/api/hcpopenshiftcluster.go
bennerv: I guess it would prevent breaking changes if CS changed the default for some reason.  I'm guessing that's what the response will be too.  If we define a default in our swagger, I'm cool with having a default in our code.

mbarnes: I think we should.  These defaults are all explicitly named in the API spec.  They all currently agree with CS, but if CS were to change a default value we'd want the RP to favor the API spec.

internal/api/v20240610preview/nodepools_methods.go
mbarnes: Maybe, though I don't think it really matters.  I only changed the TypeSpec model name so that [ResourceNameParameter](https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/reference/data-types/#Azure.ResourceManager.ResourceNameParameter) would do the right thing without overriding a bunch of its parameters.

internal/api/hcpopenshiftcluster.go
mbarnes: It did get me thinking, though, that if CS ever _did_ change a default and we wanted to match it in a new API version, I guess we'd have to make `NewDefaultHCPOpenShiftCluster` part of the versioned API interface since different API versions would now have different defaults.  We'll cross that bridge when we get there.

internal/api/v20240610preview/nodepools_methods.go
mbarnes: Gonna postpone this decision.  Right now I kinda prefer the longer name so I'll leave it alone in this PR.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
SudoBrendan: we have to care - because if someone creates `mycluster` ResourceID and someone else creates `mycluster--` resourceID, they will conflict.

SudoBrendan (commented):
Thanks Matt - as always a pleasure to review these commits. The only thing that sticks in my mind is how operator identities will change as OCP versions do - I imagine we may need to have some connection between versions and what identities exist, right?

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mbarnes: I'm not sure it will.  Cluster Service identifies clusters by an assigned ID, not by name, which suggests duplicate names are allowed.  And we would still pass the full Azure resource ID verbatim.

We do have to make sure we submit a valid name to Cluster Service but I _think_ the name is just cosmetic for human identification.  In ARO-HCP I'm not even sure if the Cluster Service name would be exposed to customers, so we could potentially just use a constant name.  (Someone correct me if I'm wrong.)

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -305,12 +305,14 @@ model UserAssignedIdentitiesProfile {
    * operators of the cluster. The set of required managed identities is dependent on the
    * Cluster's OpenShift version. */
   #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "operator name to user assigned identity pairings"
+  @visibility("create", "read")
```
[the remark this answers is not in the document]
mbarnes: I think so.  Cluster Service isn't ready for updates just yet.  @miguelsorianod is working on it.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -459,6 +459,21 @@ model Label {
...
 }
 
+/** Versions represents an OpenShift version. */
+model NodePoolVersionProfile {
+  /** ID is the unique identifier of the version. */
+  @visibility("create", "read", "update")
+  id?: string;
+
+  /** ChannelGroup is the name of the set to which this version belongs. Each version belongs to only a single set. */
+  @visibility("create", "read")
```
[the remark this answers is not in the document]
mbarnes: @vkareh pointed that out too.  Fixed in a later commit.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -498,6 +498,22 @@ model NodePoolPlatformProfile {
   availabilityZone?: string;
 }
 
+/** Supported Azure storage account types
+ * - https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types
+ */
```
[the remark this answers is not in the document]
mbarnes: Yeah, I meant to see if TypeSpec has a suitable doc tag for this.

mbarnes:
>The only thing that sticks in my mind is how operator identities will change as OCP versions do - I imagine we may need to have some connection between versions and what identities exist, right?

It's in the works: [ARO-13376](https://issues.redhat.com/browse/ARO-13376)

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
miguelsorianod: > I'm not sure it will. Cluster Service identifies clusters by an assigned ID, not by name, which suggests duplicate names are allowed. And we would still pass the full Azure resource ID verbatim.
> We do have to make sure we submit a valid name to Cluster Service but I think the name is just cosmetic for human identification. In ARO-HCP I'm not even sure the Cluster Service name would be exposed to customers, so we could potentially just use a constant name if we wanted (not that I'm advocating for that).

This is not accurate. Although CS identifies clusters by an assigned ID, there are other restrictions around the name. For example, there cannot be two Clusters with the same CS Cluster Name within the same Resource Group belonging to the same Azure Subscription.

Aside from that, the CS Cluster Name ends up being part of K8s resource names, where some for some of the K8s resource types must be a RFC 1035 label, which doesn't allow ending its value with a hyphen character.

Truncating should not be applied because of the reasons mentioned above.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -17,17 +17,11 @@ namespace Microsoft.RedHatOpenShift;
...
-  @maxLength(54)
-  @key("hcpOpenShiftClusterName") // sets the alternative name for the name propery
-  @path
-  @segment("hcpOpenShiftClusters")
-  name: string;
-
+model HcpOpenShiftCluster is TrackedResource<HcpOpenShiftClusterProperties> {
+  ...ResourceNameParameter<
+    HcpOpenShiftCluster,
+    NamePattern = "^[a-zA-Z][-a-zA-Z0-9]{1,52}[a-zA-Z0-9]$"
```
miguelsorianod, line 23: This is not correct because of the reasons mentioned in https://github.com/Azure/ARO-HCP/pull/1510/files#diff-4dec67f51651c81d6090e5ee3cf8021338af5ffd93e456d8c88eb585743aec9cR23.

Aside from that, if the intention is to simulate a RFC 1035 dns label minimum length the length check seems to also be wrong. The RFC 1035 allows something like "a", which we don't allow here. Here we are introducing a minimum of 3 characters. If we want to add this additional restriction of minimum 3 characters that's fine, but it is because of an additional restriction we decide, not of RFC 1035 dns label or CS.

The easiest would be separating the regex from the minimum and maximum lengths if possible

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -37,93 +31,49 @@ model HcpOpenShiftClusterResource
...
-  @key("nodePoolName")
-  @path
-  @segment("nodePools")
-  name: string;
-
+@parentResource(HcpOpenShiftCluster)
+model NodePool is TrackedResource<NodePoolProperties> {
+  ...ResourceNameParameter<
+    NodePool,
+    NamePattern = "^[a-zA-Z][-a-zA-Z0-9]{1,13}[a-zA-Z0-9]$"
```
miguelsorianod, line 38: This is not correct because of the same reasons explained in https://github.com/Azure/ARO-HCP/pull/1510#discussion_r2008032037 but applied to node pool names.

The only difference in this case is the max length which for nodepools name should 15 characters

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
miguelsorianod: > But the RP lowercases the name before passing it to CS.

As long as the RP is able to return the originally provided casing (as that is an ARM requirement) that should be fine.
The other implication of that route is that if in the future there's any other component other than the RP directly interacting with the CS API it might need to do the same.

miguelsorianod: Independently on the previous point around case sensitivity, there are the open questions around min (https://github.com/Azure/ARO-HCP/pull/1510#discussion_r2008099894) and max (https://github.com/Azure/ARO-HCP/pull/1510#discussion_r2008032037) length for clusters, and min (https://github.com/Azure/ARO-HCP/pull/1510#discussion_r2008103022) length for nodepools

mbarnes: >Finally, the BaseDomainPrefix should also have the same regex because of the same reasons.

This is covered.  It's a plain string in the API spec but the RP catches it during static validation:
```
BaseDomainPrefix string `validate:"omitempty,dns_rfc1035_label,max=15" ...`
```

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mbarnes: > Here we are introducing a minimum of 3 characters. If we want to add this additional restriction of minimum 3 characters that's fine, but it is because of an additional restriction we decide, not of RFC 1035 dns label or CS.

FTR, 3 character minimum has been in place since the API spec was first drafted.  I'm not introducing it here, just coding it differently.  I don't know where it came from but I assume the BU has final say.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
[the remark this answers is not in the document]
mbarnes: [Apparently not](https://typespec.io/docs/language-basics/documentation/#doc-comment-tags).

Introducing a blank line ala [Python docstrings](https://peps.python.org/pep-0257/#multi-line-docstrings) also doesn't work (but it _should_, imo).

I like having the URL there though, if only for my own reference, so I'm inclined to leave it alone for now.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
miguelsorianod: 👍 Yes I was just going to update the max cluster length insight.

miguelsorianod: Do we know why 3 is the minimum length?

mbarnes: Jerome didn't give a reason, so I still don't know if we arrived at 3 intentionally or just because it's Azure's default minimum for resource names.  He just said "3 seems reasonable to me".  I did point out that Cluster Service accepts single-character names.  It seems like the BU is more concerned with the maximum than the minimum.

SudoBrendan: Trying to read this thread - there's at least one open discussion here I want marked as OK before we merge to my eye (may be others):

> Are we sure that information implies that you have to accept any casing?

That is how I read `OAPI012: Resource IDs must not be case sensitive`, we have to accept any casing, and we are required to use the same (given) casing in responses.

If a customer says `az aro hcp create --name FOO` we will accept that, and then return a response that says they created `FOO`. Afterwards, when a customer runs `az aro hcp get --name foo` then they should get the same resource they just created, but in the response it will be called `foo`. In addition, if the customer attemps to `az aro hcp create --name fOo` that will fail with a conflict error, since the resource already exists.

What we decide to do with the casing provided information on the backend is our choice, but we must adhere to that contract as laid out in MSFT spec.

Given this, and that the frontend implementation adheres to this, and passes resourceIDs in entirely lower-case to CS, is there anything more to discuss in this thread, or do we foresee issues with it?

SudoBrendan (approved):
I believe that all comments/feedback has been addressed except for Miguel's comments on naming of resources and casing. If we plan to go to prod, I don't think we have a choice but to follow the MSFT spec here, so I think this is good to merge. @miguelsorianod if you could have a look and ensure what we have proposed (always lower-case resourceIDs before interacting with CS) is OK, and give this a :heavy_check_mark: we can merge.

mbarnes:
N.b. the most recent commit "[api: Make networkSecurityGroupId required again](https://github.com/Azure/ARO-HCP/pull/1510/commits/2384b7abc841fa9740b55a98c8cf4cbe69edcba5)" based on recent discussions.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
miguelsorianod: > That is how I read OAPI012: Resource IDs must not be case sensitive, we have to accept any casing, and we are required to use the same (given) casing in responses.

In the own MS documentation, you can find resources where casing restrictions are applied during creation for some resources. For example, the azure database for postgresql resource that we use: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftdbforpostgresql where you can observe only lowercase is accepted. I tried creating that resource and I verified that that's the case.

Based on that, that does not necessarily imply that. One could interpret it as the resource has been created and accepted, it must be treated in a case insensitive way. It might not necessarily imply you need to *accept* any casing during creation. In the own MS documentation you can find resources whose casing restrictions are limited. It is either some MS resources are non compliant or the other interpretation could apply.

In any case, given that the RP lowercases everything to CS and assuming the understand the RP will comply with treating resource names in a case insensitive way I am fine with it.

PR #1498, opened 2025-03-13
automate yaml formatting

stevekuznetsov:
.bingo,Makefile: add yamlfmt

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

*: wrap templated strings in quotes to make valid YAML

Generated with ymlfmt.wrap.sh && ymlfmt.unwrap.sh

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

*: run make yamlfmt

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

PR #1512, opened 2025-03-16
Tweaks related to Typespec and code generation

mbarnes:
### What

Couple small fixes:

- The VSCode devcontainer has been complaining about not being able to find the TypeSpec language server (`tsp-server`) on startup.  This is because it was installed in a location that VSCode does not check by default.  I figured out how to tell VSCode where to find `tsp-server`. 

- Since [`fmt` was recently added as a `make` dependency to `all-tidy`](https://github.com/Azure/ARO-HCP/pull/1498), the "Validate API Spec" CI check has been failing because `autorest`-generated Golang code often has improper formatting.  Add a step to "Validate API Spec" to run `make fmt` to clean up the formatting after code generation.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,7 +15,8 @@
                 "gopls": {
                     "ui.semanticTokens": true
                 },
-                "go.lintTool": "golangci-lint"
+                "go.lintTool": "golangci-lint",
+                "typespec.tsp-server.path": "${workspaceFolder}/ARO-HCP/api/node-modules/.bin"
             }
         }
     },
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -16,6 +16,7 @@
                     "ui.semanticTokens": true
                 },
                 "go.lintTool": "golangci-lint"
```
sanchezl, line 18:
```suggestion
                "go.lintTool": "golangci-lint",
```
?

sanchezl (changes requested):

.devcontainer/devcontainer.json
mbarnes: Good catch, thanks.

bennerv (approved):

PR #1513, opened 2025-03-16
observability: provide frontend and cs helm values

frzifus:

### DIFF
--- a/cluster-service/deploy/templates/deployment.yaml
+++ b/cluster-service/deploy/templates/deployment.yaml
@@ -136,6 +136,13 @@ spec:
         - name: azure-operators-managed-identities-config
           mountPath: /configs/azure-operators-managed-identities-config.yaml
           subPath: azure-operators-managed-identities-config.yaml
+        {{- if .Values.tracing.enabled }}
+        env:
+        - name: OTEL_EXPORTER_OTLP_ENDPOINT
+          value: "{{ .Values.tracing.address }}"
+        - name: OTEL_TRACES_EXPORTER
+          value: "{{ .Values.tracing.exporter }}"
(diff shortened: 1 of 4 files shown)


cluster-service/deploy/templates/deployment.yaml
```diff
@@ -136,6 +136,13 @@ spec:
         - name: azure-operators-managed-identities-config
           mountPath: /configs/azure-operators-managed-identities-config.yaml
           subPath: azure-operators-managed-identities-config.yaml
+        {{- if .Values.tracing.enabled }}
+        env:
+        - name: OTEL_EXPORTER_OTLP_ENDPOINT
+          value: "{{ .Values.tracing.address }}"
+        - name: OTEL_TRACES_EXPORTER
+          value: "{{ .Values.tracing.exporter }}"
```
frzifus, line 144:
```suggestion
          value: "otlp"
```

Perhaps we want to put the value `otlp` here constantly?

cluster-service/deploy/templates/deployment.yaml
```diff
@@ -136,6 +136,13 @@ spec:
         - name: azure-operators-managed-identities-config
           mountPath: /configs/azure-operators-managed-identities-config.yaml
           subPath: azure-operators-managed-identities-config.yaml
+        {{- if .Values.tracing.enabled }}
```
frzifus, line 139:
```suggestion
        {{- if ne .Values.Tracing.Address "" }}
```

Maybe we want to replace the `enabled` field with a check for an empty address?

frzifus:
Continue here:
- https://github.com/Azure/ARO-HCP/pull/1514

PR #1560, opened 2025-03-26
Grafana dashboard sync

janboll:
### What

Add Grafana dashboard sync

### Why

Want to sync dashboards to managed grafana instances

Dry run output:

```
WARNING: Preview version of extension is disabled by default for extension installation, enabled for modules without stable versions. 
WARNING: Please run 'az config set extension.dynamic_install_allow_preview=true or false' to config it specifically. 
WARNING: The command requires the extension amg. It will be installed first.
would create folder 'testing' on managed grafana arohcp-dev in rg global
would create/update dashboard '"FxooBar"' on managed grafana arohcp-dev in rg global
```

### DIFF
--- a/grafana/deploy.sh
+++ b/grafana/deploy.sh
@@ -41,6 +41,7 @@ do
         then
             echo "would create/update dashboard '${dashboard_name}' on managed grafana ${GRAFANA_NAME} in rg ${RESOURCEGROUP}"
         else
+            grep -q XYZTOBESETBYPIPELINEZYX ${dashboard} || echo "Magic string XYZTOBESETBYPIPELINEZYX not found in dashboard file ${dashboard} >&2
             sed -i "s/XYZTOBESETBYPIPELINEZYX/${folderUid}/" ${dashboard}
             az grafana dashboard update --overwrite true -g ${RESOURCEGROUP}  -n ${GRAFANA_NAME} --definition ${dashboard}
         fi
--- a/grafana/deploy.sh
+++ b/grafana/deploy.sh
@@ -6,7 +6,7 @@
 #  - check if stale folder needs to be deleted
 #  - check if stale dashboard needs to be deleted 
 
-set -eu
+set -euo pipefail
 
 export RESOURCEGROUP=${GLOBAL_RESOURCEGROUP:-global}
 
--- a/grafana/deploy.sh
+++ b/grafana/deploy.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# This is a super simple approach to syncing dashboards
+# TODO:
+#  - check if dashboard really needs update (will cause many versions in grafana db)
+#  - check if stale folder needs to be deleted
+#  - check if stale dashboard needs to be deleted 
+
+set -eu
+
+export RESOURCEGROUP=${GLOBAL_RESOURCEGROUP:-global}
+
+if [[ -z ${GRAFANA_NAME} ]];
+then
+    echo "GRAFANA_NAME needs to be set"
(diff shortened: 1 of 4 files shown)


grafana/deploy.sh
```diff
@@ -0,0 +1,49 @@
...
+then
+    echo "GRAFANA_NAME needs to be set"
+    exit 1
+fi
+
+cd dashboards
+
+existing_folders=$(az grafana folder list -g ${RESOURCEGROUP}  -n ${GRAFANA_NAME})
+
+IFS=$'\n'; for d in $(ls -1)
```
jfchevrette, line 23: This will trip if there is ever a file in the the directory

Maybe a more robust option:
```suggestion
for d in $(find . -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)
```

grafana/deploy.sh
```diff
@@ -0,0 +1,49 @@
...
+    fi
+    cd ${d}
+    IFS=$'\n'; for dashboard in $(ls -1)
+    do
+        dashboard_name=$(cat ${dashboard} | jq '.dashboard.title' )
+        if [[ ${DRY_RUN} == "true" ]];
+        then
+            echo "would create/update dashboard '${dashboard_name}' on managed grafana ${GRAFANA_NAME} in rg ${RESOURCEGROUP}"
+        else
+            sed -i "s/XYZTOBESETBYPIPELINEZYX/${folderUid}/" ${dashboard}
```
jfchevrette, line 44: sed will not error if this magic string is not found (and replaced). Maybe we should first check for this magic string and error out if it is not found? In case someone forgets to set it in the dashboard json...

```suggestion
            grep -q XYZTOBESETBYPIPELINEZYX ${dashboard} || echo "Magic string XYZTOBESETBYPIPELINEZYX not found in dashboard file ${dashboard} >&2
            sed -i "s/XYZTOBESETBYPIPELINEZYX/${folderUid}/" ${dashboard}
```

jfchevrette (commented):
Overall LGTM, nothing major aside from a couple ideas to improve the deploy script and make it more robust

grafana/deploy.sh
```diff
@@ -0,0 +1,55 @@
...
+    else
+        if [[ ${DRY_RUN} == "true" ]];
+        then
+            echo "would create folder '${d}' on managed grafana ${GRAFANA_NAME} in rg ${RESOURCEGROUP}"
+        else
+            folderUid=$(az grafana folder create --only-show-errors  -g ${RESOURCEGROUP}  -n ${GRAFANA_NAME} --title ${d} |jq -r '.uid')
+        fi
+    fi
+    pushd ${d}
+    IFS=$'\n'; for dashboard in $(ls -1)
```
jfchevrette, line 38: same as the other suggestion but for files
```suggestion
   for file in $(find . -mindepth 1 -maxdepth 1 -type f)
```

I had not seen this inner loop earlier, but now that I have I think you could simplify the two loops into one

```sh
cd dashboards
find . -mindepth 1 -maxdepth 2 -type f | while read -r file; do
  dashboard_dir=$(basename "$(dirname "$file")")
  dashboard_file=$(basename "$file")
  # handle creating/updating dashboards as before...
done
```

not deal-breaking by any means, but switching to a `find ... | while read` will also make it properly handle when/if there are spaces in the directory or file names which I had not thought about in my earlier suggestions

jfchevrette: As discussed in chat lets leave it as is for now

jfchevrette (approved):

PR #1567, opened 2025-03-28
ARO-15197 | Add KeyVault name and certificate names to CS configmap.

cgiradkar:
Since hardcoding the above things was to be avoided and that information is already maintained in the config file, the same was added in the CS configmap.


machi1990 (approved):

geoberle:
what is the context for this?

PR #1569, opened 2025-03-28
tracing: log Clusters Service attributes

simonpasquier:
This change injects the following attributes which also exist in Clusters Service:
* `cs.cluster.id`
* `cs.cluster.name`
* `cs.cluster.state`

The goal is to facilitate correlation and querying across the 2 services.

https://issues.redhat.com/browse/ARO-15044

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -23,6 +23,7 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/Azure/ARO-HCP/frontend/pkg/metrics"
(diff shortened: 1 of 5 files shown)


frzifus (approved):

simonpasquier:
@mbarnes can you please review? thanks!

internal/tracing/attributes.go
```diff
@@ -1,6 +1,10 @@
 package tracing
 
-import "go.opentelemetry.io/otel/attribute"
+import (
+	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
```
mbarnes, line 7: nit: For consistency with other source code files please move the imports below the copyright.

frontend/pkg/frontend/frontend.go
```diff
@@ -512,6 +515,7 @@ func (f *Frontend) ArmResourceCreateOrUpdate(writer http.ResponseWriter, request
 			arm.WriteInternalServerError(writer)
 			return
 		}
+		tracing.SetClusterAttributes(trace.SpanFromContext(ctx), csCluster)
```
mbarnes, line 518: Why is `tracing.SetClusterAttributes` being called twice in this function?  Seems like after this "if" block might be a better place, then you would catch it for both creating and updating operations.

mbarnes (changes requested):
Question and a small nit.  I'm not sure I understand the intent here.

simonpasquier:
> I'm not sure I understand the intent here.

I was experimenting with the Jaeger UI which forces you to select a service first when looking for traces. So if you have a CS cluster ID and you select "ARO HCP Frontend" as the service, you get no result.
I'm not sure if Jarvis has the same limitation or not though.

mbarnes (approved):
Looks good now, thanks.

PR #1578, opened 2025-04-01
ARO-15197 | Add KeyVault name and certificate names to CS configmap.

cgiradkar:
Since hardcoding the above things was to be avoided and that information is already maintained in the config file, the same was added in the CS configmap.


machi1990 (approved):

cluster-service/Makefile
```diff
@@ -87,13 +87,13 @@ deploy-pr-env-deps:
 	AZURE_CS_MI_CLIENT_ID=$(shell az identity show -g ${RESOURCEGROUP} -n clusters-service --query clientId -o tsv) && \
 	oc process --local -f cspr/cluster-service-namespace.yaml \
 		-p CLIENT_ID=$${AZURE_CS_MI_CLIENT_ID} \
-		-p KEY_VAULT_NAME=${SERVICE_KV}
-		-p FIRST_PARTY_APP_CERTIFICATE_NAME=${FPA_CERT_NAME}
+		-p KEY_VAULT_NAME=${SERVICE_KV} \
+		-p FIRST_PARTY_APP_CERTIFICATE_NAME=${FPA_CERT_NAME} \
 		-p FPA_CLIENT_ID=${AZURE_FIRST_PARTY_APPLICATION_CLIENT_ID} \
-		-p ARM_HELPER_CERTIFICATE_NAME=${FPA_CERT_NAME}
+		-p ARM_HELPER_CERTIFICATE_NAME=${FPA_CERT_NAME} \
```
machi1990, line 93: Hmm this is not correct actually. It should be the arm helper cert name

machi1990:
```suggestion
		-p ARM_HELPER_CERTIFICATE_NAME=${ARM_HELPER_CERT_NAME} \
```

machi1990: Fixed in https://github.com/Azure/ARO-HCP/pull/1585

PR #1579, opened 2025-04-01
Use Azure CLI to replace Bicep for linking the job schedule to the Automation Runbook.

YiqinZhang:
### What

Since Bicep has limitations in handling job schedule linkage, particularly due to guid() conflicts and the immutability of job schedules, it's better to switch to CLI for this part of the deployment. However, Azure CLI does not provide a direct command for managing jobSchedules in Azure Automation.

To work around this, add a PowerShell CLI step in the GitHub Actions workflow to run the `Register-AzureAutomationScheduledRunbook` command. This allows for manually linking a runbook to a schedule, which is currently not supported directly via Azure CLI.

[ARO-15539](https://issues.redhat.com/browse/ARO-15539)

### DIFF
--- a/.github/workflows/runbook-cicd.yml
+++ b/.github/workflows/runbook-cicd.yml
@@ -24,6 +24,11 @@ on:
 permissions:
   id-token: write
   contents: read
+env:
+  AUTOMATION_ACCOUNT_RG: hcp-dev-automation-account
+  AUTOMATION_ACCOUNT_NAME: hcp-dev-automation
+  RUNBOOK_NAME: hcp-dev-automation_roleAssignmentsCleanup
+  SCHEDULE_NAME: hcp-dev-automation_nightly-schedule
 jobs:
   ci:
     name: Lint & Test Runbook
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
> particularly due to guid() conflicts

This usually means you simply did not include all of the independent inputs to `guid()`. Are you certain ARM templates can't work for this?

> immutability of job schedules

This will require something imperative. Perhaps a post-hoc cleanup?

YiqinZhang:
I've tried various types of input for guid(), but they all failed during redeployment due to a resource conflict—since there's no explicit Azure CLI command to delete jobSchedules, the existing one remains and causes the conflict.

Adding a timestamp to generate a unique ID solves the issue, but it introduces nondeterministic warnings during deployment. PR is here: https://github.com/Azure/ARO-HCP/pull/1575


So now, I’m exploring another workaround to avoid both the conflict and the warnings.

> > particularly due to guid() conflicts
> 
> This usually means you simply did not include all of the independent inputs to `guid()`. Are you certain ARM templates can't work for this?

stevekuznetsov:
>I've tried various types of input for guid(), but they all failed during redeployment due to a resource conflict

Sure, seems like you need the script anyway but FWIW what you're describing should just be a lack of _enough_ inputs to `guid()`.

PR #1581, opened 2025-04-02
Promtool-rules change configuration and add rule unit testing

janboll:
[ARO-15912](https://issues.redhat.com/browse/ARO-15912)
## What
Adding call to promtool

Adding support for service owned alerts, where alerts folder and files are stored in service directory:

- cluster-service/alerts
- backend/alerts
 .... 
By putting alerts into service folders we can easily grant code ownership to these files.

These folders will be referenced by a configuration file in observability/alerts:
```
rulesFolders:
 - cluster-service/alerts
 - backend/alerts
untestedRules:
additionalRules:
 - kubernetesControlPlane-prometheusRule.yaml
outputFile:
 - dev-infrastructure/obsevability/generated-prometheus-rules.bicep
```
All files in rulesFolders must have tests
Files referenced in additionalRulesuntestedRules are exemptions and don't require a test

## Why
Want to have strict testing for prometheus alerts
Need to support more than one alert file


tooling/prometheus-rules/main.go
```diff
@@ -72,11 +68,62 @@ func main() {
 	if err := o.complete(flag.Args()); err != nil {
 		logrus.WithError(err).Fatal("could not complete options")
 	}
-	if err := generate(o.rules, o.output); err != nil {
+	b, _ := os.ReadFile("testing-prometheusRule_test.yaml")
```
stevekuznetsov, line 71: a) why is this hard-coded?
b) why is the error ignored?

tooling/prometheus-rules/main.go
```diff
@@ -45,10 +45,6 @@ func (o *options) complete(args []string) error {
 	if err := yaml.Unmarshal(rawRules, &o.rules); err != nil {
 		return fmt.Errorf("failed to parse input rules: %v", err)
 	}
-	o.output, err = os.Create(o.outputBicep)
```
stevekuznetsov, line 48: Why did this move out of `complete`? The point of this pattern is to have all necessary data read in and created here so testing past this step is easy.

tooling/prometheus-rules/main.go
```diff
@@ -72,11 +68,62 @@ func main() {
...
+		return fmt.Errorf("missing filename extension or using '.' in filename")
+	}
+
+	testFile := fmt.Sprintf("%s%s%s_test.%s", dir, string(os.PathSeparator), fileNameParts[0], fileNameParts[1])
+	err = os.WriteFile(testFile, testFileContent, 0644)
+	if err != nil {
+		return fmt.Errorf("error writing rule groups test file %v", err)
+	}
+
+	cmd := exec.Command("promtool", "test", "rules", testFile)
```
stevekuznetsov, line 118: It's unfortunate that they put everything into `package main`. The drive behind having options as input and non-`main` packages is that you can just import stuff like https://github.com/prometheus/prometheus/blob/main/cmd/promtool/unittest.go#L52 and use it inline.

tooling/prometheus-rules/main.go
janboll: Sorry, this PR should be a draft. going to change that. It's simply not ready

tooling/prometheus-rules/main.go
janboll: Oh, my point was, why create the output if the test could fail?

stevekuznetsov: Fair, but losing out on having options as useful input to execution is a shame.

tony-schndr (dismissed):
Overall this is great!  Some improvements/nits for a follow up if you wish:

When testing I don't see any output on what tests were ran, it would be nice to see a format looking something like:
:green_circle: cluster-service/alert/InstancesDownV1 passed
:x:  cluster-service/alert/InstancesDownV1 failed
    <failure details here>
I saw output only when it failed.

I also tried to generate the alerts in bicep, I updated the name `InstancesDownV1` to `InstancesDownV2` and I didn't see the update in the generated bicep module.  Not sure if I was doing something wrong.

Maybe you're already working on it, but could we also get some documentation on how this is used?

.github/workflows/ci-observability.yml
```diff
@@ -0,0 +1,68 @@
+name: 'ci-observability'
+on:
+  pull_request:
+    branches:
+    - main
+    paths:
+    - '.github/workflows/ci-python.yml'
```
mmazur, line 7:
```suggestion
    - '.github/workflows/ci-observability.yml'
```

janboll:
@tony-schndr regarding the missing update, did you run `make alerts` in the observability folder?

tony-schndr (approved):

PR #1586, opened 2025-04-02
Report status in asynchronous node pool operations

mbarnes:
[ARO-15560 - Update the RP to check against node pool status](https://issues.redhat.com/browse/ARO-15560)

### What

Cluster Service recently added a status endpoint for node pools, so we can finally complete* this missing piece of the RP backend.

\* Major caveat: Cluster Service does not yet provide error messages for failed node pool operations, so at the moment all we can report is an opaque "Internal Server Error".  This is being tracked in [ARO-14969 - Node Pool error handling design and implementation](https://issues.redhat.com/browse/ARO-14969).

### DIFF
--- a/internal/ocm/internalid.go
+++ b/internal/ocm/internalid.go
@@ -19,8 +19,9 @@ const (
 	v1NodePoolPattern             = v1ClusterPattern + "/node_pools/*"
 	v1BreakGlassCredentialPattern = v1ClusterPattern + "/break_glass_credentials/*"
 
-	aroHcpV1Alpha1Pattern        = "/api/aro_hcp/v1alpha1"
-	aroHcpV1Alpha1ClusterPattern = aroHcpV1Alpha1Pattern + "/clusters/*"
+	aroHcpV1Alpha1Pattern         = "/api/aro_hcp/v1alpha1"
+	aroHcpV1Alpha1ClusterPattern  = aroHcpV1Alpha1Pattern + "/clusters/*"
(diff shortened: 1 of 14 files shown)


internal/ocm/internalid.go
```diff
@@ -67,6 +68,12 @@ func (id *InternalID) validate() error {
 		return nil
 	}
 
+	if match, _ = path.Match(aroHcpV1Alpha1NodePoolPattern, id.path); match {
+		// Temporarily use cmv1 constant for backward-compatibility.
```
SudoBrendan, line 72: can you explain this comment a bit more - what would be our path forward for this, and how long is "temporary"?

SudoBrendan (commented):
Everything's intuitive other than the last use of the old API - how do we migrate out of that, do we need to groom additional work?

internal/ocm/internalid.go
mbarnes: It's a bit of a mess still from the partial transition to the `aro_hcp` API.  **Cluster credentials** is the last thing still using the `clusters_mgmt` API.  As soon as CS makes that available through `aro_hcp` (maybe they have already... need to check) I'll come back and remove all trace of `clusters_mgmt`, including these "kind" constants.

SudoBrendan (approved):

internal/ocm/internalid.go
nimrodshn: @mbarnes not yet afaik.

PR #1590, opened 2025-04-03
bump mise image for MSFT envs

geoberle:
### What

bump MISE image digest


frontend/pipeline.yaml
```diff
@@ -14,7 +14,7 @@ resourceGroups:
         value: "true"
     variables:
     - name: TARGET_ACR
-      configRef: 'acr.svc.name'
+      configRef: acr.svc.name
```
geoberle, line 17: unrelated cosmitic fix

PR #1591, opened 2025-04-03
helm deployment diagnostics

geoberle:
### What

observe the in-cluster state after a helm rollout, showing deployments, pods, service accounts that belong to respective namespace.
the diagnostics are only collected when running in EV2. GH action logs are unprotected and the diagnostic information might  not be suitable for being shared publicly.

### Why

having post deploy diagnostics available in the EV2 portal allows quick first-pass problem analysis in case a deployment fails without the need to aquire cluster admin.

### DIFF
--- a/hack/deployment-diagnostics.sh
+++ b/hack/deployment-diagnostics.sh
@@ -62,10 +62,11 @@ for DEPLOY in $DEPLOYMENTS; do
     kubectl describe deployment "$DEPLOY" -n "$NAMESPACE"
 done
 
+# TODO - replace this with ready-to-click kusto links once kusto is ready
 PODS=$(kubectl get pods -n "$NAMESPACE" -o jsonpath='{.items[*].metadata.name}' 2>/dev/null || echo "")
 for POD in $PODS; do
     echo -e "\n>>> Logs from pod: $POD"
-    kubectl logs "$POD" -n "$NAMESPACE" --tail=50
+    kubectl logs "$POD" -n "$NAMESPACE" --previous --all-containers --tail=50
 done
 
 echo -e "\n--- ServiceAccounts in $NAMESPACE ---"
--- a/hack/deployment-diagnostics.sh
+++ b/hack/deployment-diagnostics.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [[ -z "${EV2:-}" ]]; then
+   # this script only executes in EV2
+   # executing this in other environments with less restricted access to CD logs
+   # can lead to leaking sensitive information
+   exit 0
+fi
+
+HELM_RELEASE_NAME="$1"
+NAMESPACE="$2"
+
+cat <<'EOF'
+
+
+  _          _                      _                       _        __
+ | |__   ___| |_ __ ___    _ __ ___| | ___  __ _ ___  ___  (_)_ __  / _| ___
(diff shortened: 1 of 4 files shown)


hack/deployment-diagnostics.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [[ -z "${EV2:-}" ]]; then
+   # this script only executes in EV2
+   # executing this in other environments with less restricted access to CD logs
+   # can lead to leaking sensitive information
```
stevekuznetsov, line 8: n.b. anyone who can send a PR that edits files under `.github` has an attack vector for this

hack/deployment-diagnostics.sh
```diff
@@ -0,0 +1,86 @@
...
+DEPLOYMENTS=$(kubectl get deployments -n "$NAMESPACE" -o jsonpath='{.items[*].metadata.name}' 2>/dev/null || echo "")
+for DEPLOY in $DEPLOYMENTS; do
+    echo -e "\n--- Describe Deployment: $DEPLOY ---"
+    kubectl describe deployment "$DEPLOY" -n "$NAMESPACE"
+done
+
+PODS=$(kubectl get pods -n "$NAMESPACE" -o jsonpath='{.items[*].metadata.name}' 2>/dev/null || echo "")
+for POD in $PODS; do
+    echo -e "\n>>> Logs from pod: $POD"
+    kubectl logs "$POD" -n "$NAMESPACE" --tail=50
```
stevekuznetsov, line 68: Will this always default to some container in a multi-container Pod?

stevekuznetsov: Also, in the case that the pod is in CrashLoopBackOff you might want `--previous`

hack/deployment-diagnostics.sh
```diff
@@ -0,0 +1,86 @@
...
+
+DEPLOYMENTS=$(kubectl get deployments -n "$NAMESPACE" -o jsonpath='{.items[*].metadata.name}' 2>/dev/null || echo "")
+for DEPLOY in $DEPLOYMENTS; do
+    echo -e "\n--- Describe Deployment: $DEPLOY ---"
+    kubectl describe deployment "$DEPLOY" -n "$NAMESPACE"
+done
+
+PODS=$(kubectl get pods -n "$NAMESPACE" -o jsonpath='{.items[*].metadata.name}' 2>/dev/null || echo "")
+for POD in $PODS; do
+    echo -e "\n>>> Logs from pod: $POD"
```
stevekuznetsov, line 67: Is it possible to have output artifacts from Ev2? Might be nicer to have this as a downloadable file - all logs for all pods in the namespace could quickly balloon to megabytes of text and be annoying to operate over in a web view.

hack/deployment-diagnostics.sh
geoberle: ty. adressed

hack/deployment-diagnostics.sh
geoberle: as discussed in chat, it makes a lot of sense to craft links to Kusto instead of showing logs here.

hack/deployment-diagnostics.sh
geoberle: reconsidred: i removed the section and will work on kusto links asap

hack/deployment-diagnostics.sh
geoberle: i checked that only project member PRs actually run PR checks

janboll (approved):

PR #1592, opened 2025-04-04
Update go to 1.24

janboll:


go.work
```diff
@@ -1,6 +1,6 @@
-go 1.23.3
+go 1.24.0
 
-toolchain go1.23.3
+toolchain go1.24.1
```
geoberle, line 3: should these two be in sync?

geoberle:
LGTM for me from SL perspective/needs
cc @mbarnes this affects backend/frontend so your input is required

mbarnes (approved):
Looks like there's quite a few exclusions for RP code that should get addressed ASAP but it doesn't need to be a blocker for merging.  Approving in hopes of never having to read "File is not \`goimports\`-ed" again.

PR #1596, opened 2025-04-07
Msft stg cs rollout

geoberle:
### What

* ARMHelper and MSIMock CLI args and SecretProviderClass references are now optional
* moved ACR detail lookup into global-output step to adress cross subscription deployment

https://issues.redhat.com/browse/ARO-16274


machi1990 (approved):
The cs deployment looks good to me after the modification.

janboll (approved):

machi1990 (approved):

PR #1603, opened 2025-04-08
Grafana dashboard script, fixes

janboll:
### What

Adding additional fixes and some more testing

### Why

Ensure keys are always set and making code easier to read

### DIFF
--- a/observability/grafana/grafana.py
+++ b/observability/grafana/grafana.py
@@ -6,6 +6,8 @@ import os
 import subprocess
 import tempfile
 
+AZURE_MANAGED_FOLDERS = ["Azure Monitor", "Microsoft Defender for Cloud"]
+
 
 def run_command(command):
     if os.getenv("PRINT_COMMANDS"):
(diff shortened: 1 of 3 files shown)


observability/grafana/grafana.py
```diff
@@ -90,6 +91,84 @@ def fs_get_dashboard_folders(search_root: str) -> list[str]:
...
+    existing_folders: list[dict[str, any]],
+    g: GrafanaRunner,
+) -> None:
+    k = f"{d['folderUid']}_{d['title']}"
+    if k not in dashboards_visited:
+        for amf in AZURE_MANAGED_FOLDERS:
+            uid = get_folder_uid(amf, existing_folders)
+            if uid and uid in k:
+                return
+        g.delete_dashboard(k.split("_")[1])
```
geoberle, line 169: * should this function be named `delete_stale_dashboard` (singular)?
* from what i understand you only need `k` to check if this dashboard still exists in the repo
* instead of `uid in k` you could directly check `uid == d['folderUid']`
* instead `k.split("_")[1]` you can directly use `d['title']`

janboll: makes sense, fixed!

geoberle (approved):
sweet! thanks

PR #1609, opened 2025-04-09
Replace bingo with go tool

janboll:
### What

Replace bingo with go tool
### Why

go 1.24 brings tooling support. No more need to maintain extra infra for this.


janboll:
@stevekuznetsov, since you drove bingo adoption, can you please take a look?

jfchevrette:
Are the version bumps strictly needed to do this?

If at all possible I would leave those out so we know the functional change does not affect the codebases

janboll:
I assume new versions came with tooling install and has recursive effects, since modules are used in our code as well.

mbarnes (commented):
Conflicts need resolved but :+1: from me on the intention.

stevekuznetsov:
@janboll so sorry I missed this - please ping me on Slack in the future, my GitHub review queue gets slammed with OSS stuff from the last ten years.

re: these changes - I say go for it, 1.24 `go tool` is great! The one thing to worry about a tiny bit - the MSFT component governance security scanning today uses `go mod graph` if I remember correctly to find the list of modules your project depends on, and is fairly naive in the sense that it does not worry about minimum version selection or module pruning, etc - in classic it has already alerted on many false positives. With this 1.24 feature, you _may_ be opening yourself up to have all of your build-time dependencies included in CVE-related scans, which would be 100% noise. If this happens, you could always revert to `bingo`.

PR #1514, opened 2025-03-16
observability: provide frontend and cs helm values

frzifus:

PR #1612, opened 2025-04-10
Calling bicep fails in latest ubuntu runner

janboll:
### Why

see: https://github.com/actions/runner-images/issues/11987


frzifus (approved):
Link: https://github.com/Azure/ARO-HCP/pull/1514#issuecomment-2789683183

PR #1614, opened 2025-04-10
Refactor grafana.py

janboll:
### What

Refactor grafana.py

### Why

Refactor grafana.py to leverage observability.yaml


tony-schndr (approved):
lgtm

PR #1619, opened 2025-04-13
Ai review

jianzzha:
### What

Use custom build so that the openshift version check is bypassed; and use quay.io image for assisted service to avoid pull secret use.


jianzzha:
This is for sourcery ai testing, closing

PR #1631, opened 2025-04-15
Remove backplane deploy for now

janboll:
### What

remove backplane api from svc make target

### Why

Should have been disabled anyhow, since it's still work in progress and blocks others from running since there are still open issues.


Nanyte25:
approved

PR #1635, opened 2025-04-16
Fixes for backplane api deployment

janboll:
### What

Fixes for backplane api deployment


Nanyte25:
looks like the service_pr_check is failing because it deploying to incorrect namespace @janboll

backplane-api/Makefile
```diff
@@ -4,12 +4,11 @@
 deploy:
 	@kubectl create namespace aro-hcp --dry-run=client -o json | kubectl apply -f - && \
```
Nanyte25, line 5: This should be ``kubectl create namespace backplane-api --dry-run=client -o json | kubectl apply -f

PR #1637, opened 2025-04-16
frontend: Add MiddlewareReferer

mbarnes:
### What

`MiddlewareReferer` ensures a `Referer` header is present in the request. This header is always added by ARM but is often forgotten in testing environments. If missing, derive a `Referer` from the `http.Request`.

`Referer` header is required for RP responses to include operation status or result endpoints in `Azure-AsyncOperation` or `Location` headers.  Currently if there's no `Referer` header the RP just omits these response headers but that's not ideal for testing.

No JIRA, this is just a test aid.


venkateshsredhat (approved):

venkateshsredhat:
LGTM , Thanks !

PR #1642, opened 2025-04-17
ARO-15755 | Display inflight check errors

oriAdler:
If the cluster got into an error state with status code OCM4001, fetch the inflight check errors and populate the `CloudErrorBody` with the details.

### DIFF
--- a/backend/operations_scanner.go
+++ b/backend/operations_scanner.go
@@ -358,7 +358,7 @@ func (s *OperationsScanner) pollClusterOperation(ctx context.Context, op operati
 		return
 	}
 
-	opStatus, opError, err := convertClusterStatus(clusterStatus, op.doc.Status)
+	opStatus, opError, err := s.convertClusterStatus(ctx, clusterStatus, op.doc.Status, op.doc.InternalID)
 	if err != nil {
 		s.operationsFailedCount.WithLabelValues(pollClusterOperationLabel).Inc()
 		op.logger.Warn(err.Error())
@@ -524,7 +524,8 @@ func (s *OperationsScanner) postAsyncNotification(ctx context.Context, op operat
 // convertClusterStatus attempts to translate a ClusterStatus object from
 // Cluster Service into an ARM provisioning state and, if necessary, a
 // structured OData error.
-func convertClusterStatus(clusterStatus *arohcpv1alpha1.ClusterStatus, current arm.ProvisioningState) (arm.ProvisioningState, *arm.CloudErrorBody, error) {
+func (s *OperationsScanner) convertClusterStatus(ctx context.Context, clusterStatus *arohcpv1alpha1.ClusterStatus,
+	current arm.ProvisioningState, internalId ocm.InternalID) (arm.ProvisioningState, *arm.CloudErrorBody, error) {
 	var opStatus arm.ProvisioningState = current
 	var opError *arm.CloudErrorBody
 	var err error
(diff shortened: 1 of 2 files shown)


backend/operations_scanner.go
```diff
@@ -567,3 +575,49 @@ func convertClusterStatus(clusterStatus *arohcpv1alpha1.ClusterStatus, current a
...
+
+func (s *OperationsScanner) convertInflightChecks(ctx context.Context, internalId ocm.InternalID) (*arm.CloudErrorBody, error) {
+	inflightChecks, err := s.clusterService.GetClusterInflightChecks(ctx, internalId)
+	if err != nil {
+		return &arm.CloudErrorBody{}, err
+	}
+
+	var cloudErrors []arm.CloudErrorBody
+	for _, inflightCheck := range inflightChecks.Items() {
+		if inflightCheck.State() == "failed" {
```
mbarnes, line 587: Probably should use the `InflightCheckStateFailed` constant here.

backend/operations_scanner.go
```diff
@@ -567,3 +575,49 @@ func convertClusterStatus(clusterStatus *arohcpv1alpha1.ClusterStatus, current a
...
+			cloudErrors = append(cloudErrors, arm.CloudErrorBody{
+				// TODO: set a custom inflight error?
+				Code:    arm.CloudErrorCodeInternalServerError,
```
mbarnes, line 590: I'm still tempted to use `inflightCheck.Name()` for the code, at least initially.

backend/operations_scanner.go
```diff
@@ -567,3 +576,49 @@ func convertClusterStatus(clusterStatus *arohcpv1alpha1.ClusterStatus, current a
...
+func convertInflightCheckDetails(inflightCheck *arohcpv1alpha1.InflightCheck) string {
+	details, ok := inflightCheck.GetDetails()
+	if !ok {
+		return ""
+	}
+
+	detailsMap, ok := details.(map[string]interface{})
+	if !ok {
+		return ""
+	}
```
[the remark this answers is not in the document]
oriAdler: I agree, updated to return a boolean value, and log the error accordingly https://github.com/Azure/ARO-HCP/pull/1646

backend/operations_scanner.go
oriAdler: I couldn't find an existing constant for that, so I created one here https://github.com/Azure/ARO-HCP/pull/1646/files#diff-8e0a776b7f44845193da0fd0a483d869eceaa15415306a81b85b647292d7a306R62, please let me know if that's the right place.

backend/operations_scanner.go
oriAdler: I'm on the fence here, I think there is no harm in specifying an error code per inflight. 
I need to go through the inflight service code to understand if there is a distinction in CS between a failure due to the cluster spec versus an internal error in the inflight service.

@miguelsorianod any second thoughts on the error code per inflight?

oriAdler:
I closed the PR in favor of https://github.com/Azure/ARO-HCP/pull/1646

backend/operations_scanner.go
oriAdler: Checking the inflight service in CS code, there isn't a distinction between a bad request and an internal server error. In both cases, the inflight will fail with the error details.
Maybe that's an improvement we want to consider before making this distinction in the RP?

PR #1644, opened 2025-04-17
Backport API changes from ARM review

mbarnes:
[ARO-4681 - New API to be reviewed and approved by ARM Team at Microsoft](https://issues.redhat.com/browse/ARO-4681)

### What

These are the remaining API backports from the [ARM API review](https://github.com/Azure/azure-rest-api-specs-pr/pull/22123).

Most significant change was removing `disableUserWorkloadMonitoring`, which it turns out is ROSA-specific.


bennerv (approved):

nimrodshn:
I was under the impression `disableUserWorkloadMonitoring` is also an ARO-HCP property.

@ahitacat  @miguelsorianod  @machi1990  Please keep me honest.

machi1990:
`disableUserWorkloadMonitoring` we've it in the cluster spec but the desired effect of the property doesn't take place as the mechanism for aro-hcp isn't built. Only rosa-hcp have it wired up end to end; we need a similar mechanism in aro-hcp (perhaps via ACM policy or similar) to make it work. There is a dedicated card for that

PR #1645, opened 2025-04-18
observability: respect jaeger deployment config

frzifus:
### What

Currently Jaeger is not automatically deployed when calling `deployall`. This PR extends the config schema to enable the jaeger deployment and enables it by default on the personal-dev environment.

### Why

Tracing is on the personal-dev environment enabled by default. Deploying Jaeger by default two makes our lives easier :smiley:


observability/pipeline.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - name: deploy
+    action: Shell
+    command: make deploy
+    dryRun:
+      variables:
+      - name: DRY_RUN
+        value: "true"
+    variables:
+    - name: DEPLOY
+      configRef: svc.jaeger.deploy
```
frzifus, line 36: Will we need another service group to do this individually on the svc and mgmt cluster?

frzifus: But perhaps a setting for svc and mgmt is also sufficient? wdyt  @simonpasquier @dinhxuanvu ?

frzifus: It is :) 
I just moved the file into `observability/tracing`.

simonpasquier (dismissed):
lgtm

frzifus:
cc @janboll

frzifus:
cc @mbarnes

dinhxuanvu (approved):
/lgtm

janboll (approved):

PR #1646, opened 2025-04-20
ARO-15755 | Display inflight check errors

oriAdler:
https://issues.redhat.com/browse/ARO-15755

### What

If the cluster got into an error state with status code OCM4001, fetch the inflight check errors and populate the `CloudErrorBody` with the details.

### Why

Improve the UX.

e.g.
```
curl http://localhost:8443/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationsStatus/53b18a92-ddb6-43fc-8b8b-c354d3d24baf?api-version=2024-06-10-preview
{
    "id": "/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationsStatus/53b18a92-ddb6-43fc-8b8b-c354d3d24baf",
    "name": "53b18a92-ddb6-43fc-8b8b-c354d3d24baf",
    "status": "Failed",
    "startTime": "2025-04-20T13:42:19.101376032Z",
    "endTime": "2025-04-20T13:42:48.096984902Z",
    "error": {
        "code": "InternalServerError",
        "message": "Managed resource group 'oadler-rg' should not exist prior to cluster creation"
    }
}
```

### DIFF
--- a/backend/operations_scanner.go
+++ b/backend/operations_scanner.go
@@ -69,6 +69,10 @@ const (
 	NodePoolStateError            NodePoolStateValue = "error"
 )
 
+const (
+	InflightChecksFailedProvisionErrorCode = "OCM4001"
+)
+
 type operation struct {
 	id     string
 	pk     azcosmos.PartitionKey
@@ -387,7 +391,7 @@ func (s *OperationsScanner) pollClusterOperation(ctx context.Context, op operati
 		return
 	}
 
-	opStatus, opError, err := convertClusterStatus(clusterStatus, op.doc.Status)
+	opStatus, opError, err := s.convertClusterStatus(ctx, op.logger, clusterStatus, op.doc.Status, op.doc.InternalID)
 	if err != nil {
 		s.operationsFailedCount.WithLabelValues(pollClusterOperationLabel).Inc()
 		op.logger.Warn(err.Error())
@@ -671,7 +675,9 @@ func (s *OperationsScanner) postAsyncNotification(ctx context.Context, op operat
 // convertClusterStatus attempts to translate a ClusterStatus object from
 // Cluster Service into an ARM provisioning state and, if necessary, a
 // structured OData error.
-func convertClusterStatus(clusterStatus *arohcpv1alpha1.ClusterStatus, current arm.ProvisioningState) (arm.ProvisioningState, *arm.CloudErrorBody, error) {
+func (s *OperationsScanner) convertClusterStatus(ctx context.Context, logger *slog.Logger,
+	clusterStatus *arohcpv1alpha1.ClusterStatus, current arm.ProvisioningState,
+	internalId ocm.InternalID) (arm.ProvisioningState, *arm.CloudErrorBody, error) {
 	var opStatus = current
 	var opError *arm.CloudErrorBody
 	var err error
@@ -679,9 +685,8 @@ func convertClusterStatus(clusterStatus *arohcpv1alpha1.ClusterStatus, current a
 	switch state := clusterStatus.State(); state {
 	case arohcpv1alpha1.ClusterStateError:
 		opStatus = arm.ProvisioningStateFailed
-		// FIXME This is guesswork. Need clarity from Cluster Service
-		//       on what provision error codes are possible so we can
-		//       translate to an appropriate cloud error code.
+		// Provision error codes are defined in the CS repo:
+		// https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/pkg/api/cluster_errors.go
 		code := clusterStatus.ProvisionErrorCode()
 		if code == "" {
 			code = arm.CloudErrorCodeInternalServerError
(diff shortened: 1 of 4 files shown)


backend/operations_scanner.go
```diff
@@ -58,6 +58,11 @@ const (
 	NodePoolStateError            NodePoolStateValue = "error"
 )
 
+const (
+	InflightCheckStateFailed               = "failed"
+	InflightChecksFailedProvisionErrorCode = "OCM4001"
```
oriAdler, line 63: @mbarnes, please let me know if this is the right place for the constants.

backend/operations_scanner.go
```diff
@@ -660,17 +665,18 @@ func (s *OperationsScanner) postAsyncNotification(ctx context.Context, op operat
...
+	clusterStatus *arohcpv1alpha1.ClusterStatus, current arm.ProvisioningState,
+	internalId ocm.InternalID) (arm.ProvisioningState, *arm.CloudErrorBody, error) {
 	var opStatus = current
 	var opError *arm.CloudErrorBody
 	var err error
 
 	switch state := clusterStatus.State(); state {
 	case arohcpv1alpha1.ClusterStateError:
 		opStatus = arm.ProvisioningStateFailed
-		// FIXME This is guesswork. Need clarity from Cluster Service
```
oriAdler, line 671: @mbarnes, I added a reference to the cluster service repo. Does this meet your expectations?

backend/operations_scanner_test.go
```diff
@@ -468,6 +469,7 @@ func TestConvertClusterStatus(t *testing.T) {
 	}
 
 	for _, tt := range tests {
+		var operationsScanner *OperationsScanner
```
oriAdler, line 472: @mbarnes, I noticed the `OperationsScanner` has a `clusterService` field of type `ocm.ClusterServiceClient`, is that intentional?
To use the mock and add more test coverage, we'll have to change it to the `ClusterServiceClientSpec` interface. I could handle that in a separate PR to keep this one concise and add tests later.

backend/operations_scanner.go
mbarnes: That helps, thanks.  I'm still not sure if it's worth calling `clusterStatus.Description()` as a fallback.  Would that ever return a message appropriate for error details?  `ProvisionErrorCode` and `ProvisionErrorMessage` seem like the right things to check for a failure during provisioning, but what about a failed cluster update?

backend/operations_scanner_test.go
mbarnes: Good catch, that is in fact _not_ intentional.  I guess the existing tests weren't mocking any CS calls.  Please fix however you think is best.

internal/ocm/ocm.go
```diff
@@ -22,6 +22,9 @@ type ClusterServiceClientSpec interface {
 	// GetClusterStatus sends a GET request to fetch a cluster's status from Cluster Service.
 	GetClusterStatus(ctx context.Context, internalID InternalID) (*arohcpv1alpha1.ClusterStatus, error)
 
+	// GetClusterInflightChecks sends a GET request to fetch a cluster's inflight checks from Cluster Service.
+	GetClusterInflightChecks(ctx context.Context, internalID InternalID) (*arohcpv1alpha1.InflightCheckList, error)
```
mbarnes, line 26: Might Cluster Service ever paginate the inflight check list?  Since we moved to Go 1.23 I've been building [iterators](https://pkg.go.dev/iter) around CS list types to encapsulate pagination.  See [/internal/ocm/iterators.go](https://github.com/Azure/ARO-HCP/blob/main/internal/ocm/iterators.go).  I don't know if it would be worth it in this case, but just wanted to mention it for consideration.

mbarnes (changes requested):
Looks good overall.  Withholding approval only for a nit (unnecessary constant definition) and a question to consider.

backend/operations_scanner.go
```diff
@@ -736,3 +751,68 @@ func convertNodePoolStatus(nodePoolStatus *arohcpv1alpha1.NodePoolStatus, curren
...
+	if err != nil {
+		return &arm.CloudErrorBody{}, err
+	}
+
+	var cloudErrors []arm.CloudErrorBody
+	for _, inflightCheck := range inflightChecks.Items() {
+		if inflightCheck.State() == InflightCheckStateFailed {
+			cloudErrors = append(cloudErrors, convertInflightCheck(inflightCheck, logger))
+		}
+	}
```
mbarnes, line 770: This _should_ never happen, but maybe we should return an error here if `len(cloudErrors) == 0` after filtering on failed inflight checks... just to ensure we never have a `MultipleErrorsOccurred` with no nested errors.

backend/operations_scanner.go
oriAdler: Going through CS code mappers, it seems reasonable to use description as a fallback https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/pkg/mappers/outbound/clusters.go?ref_type=heads#L787 (if both are empty, a generic error message is set).

Can you please elaborate on a failed cluster update in this context? Is the flow for cluster update also trigger `convertClusterStatus`?

backend/operations_scanner_test.go
oriAdler: I created this ticket to track the changes https://issues.redhat.com/browse/ARO-17009, will do in a follow-up PR.

internal/ocm/ocm.go
oriAdler: Let me take a closer look at the CS backend.
The number of inflights won't exceed 10-20 inflights soon, and with relatively small struct per inflight, I doubt we'll use paging for this endpoint.

backend/operations_scanner.go
oriAdler: Added.

internal/ocm/ocm.go
oriAdler: After double-checking, the default page size in CS is 100, so in the case of inflights, it will be one page.
I don't see a scenario where the number of inflights reaches that number.

mbarnes (approved):
Everything looks good to me.  Thanks for this!

PR #1651, opened 2025-04-21
Detect Template Language Expressions (TLEs) in preflight payloads

mbarnes:
### What

Detect and skip preflight resources with Template Language Expressions (TLEs).

Template Language Expressions are easy to identify and [Microsoft is very precise about how to do so](https://armwiki.azurewebsites.net/api_contracts/PreflightImplementation.html#identifying-tles): they simply begin with an opening square bracket (`[`) and end with a closing square bracket (`]`).  So `"[ ... ]"` is a TLE.

### Why

Getting ahead of an [upcoming Preflight KPI](https://armwiki.azurewebsites.net/api_contracts/PreflightKPI.html#kpi-3-preflight-resource-failing-with-tles):

> **KPI 3: Preflight Resource Failing with TLEs**
> Your implementation MUST accept resource bodies containing Template Language Expressions (TLEs), and be able to skip validation on the TLEs that are detected.

This implementation simply skips any preflight resources with a TLE. Strictly speaking this satisfies the KPI, but it's not really what Microsoft intended. Microsoft wants RPs to skip validating _fields_ with a TLE, not entire resources.  But I settled on this for a couple reasons:

1. Cluster Service has no idea what a TLE is, so the RP can't pass it any resource with a TLE because it would likely fail Cluster Service's validation. And I'm not about to try to figure out what fields Cluster Service does and does not inspect when validating.

2. It's possible for the RP to accept TLEs in its string fields and still do _some_ static validation itself. I even started down this road initially and had a partially-working implementation.  What killed it was a) it was way more complex and error-prone and not at all future-proof, and b) the inter-dependencies between some fields introduced all kinds of subtle corner cases with TLEs that would likely result in false validation errors (hello, managed identities).

All the refactoring around making validation functions take an `http.Request` is left over from my original attempt to handle _some_ TLEs and not strictly needed. But I like how it turned out so I kept it.


internal/api/arm/preflight.go
```diff
@@ -123,3 +124,45 @@ func WriteDeploymentPreflightResponse(w http.ResponseWriter, preflightErrors []C
 
 	_, _ = WriteJSONResponse(w, http.StatusOK, response)
 }
+
+func recurseDetectTLE(v any) bool {
+	switch v := v.(type) {
```
venkateshsredhat, line 129: @mbarnes I am sure you would have checked it but are these the only types?

mbarnes: Yeah, the only data structures in JSON are lists (Go slices) and objects (Go maps).  We know that a preflight resource is a JSON object, so we let [json.Unmarshal](https://pkg.go.dev/encoding/json#Unmarshal) convert it to a `map[string]any` and recurse from there.

PR #1653, opened 2025-04-21
remove aks cluster admin from aks UAMSI

tony-schndr:
### What

Removes cluster admin role from aks UAMSI

### Why

Doesn't look like this role is needed.


bennerv (approved):

bennerv (changes requested):
The `aksClusterAdminRoleId` section at the top oof this file can probably be removed since it's no longer referenced anywhere.

tony-schndr:
>The aksClusterAdminRoleId section at the top oof this file can probably be removed since it's no longer referenced anywhere.
Oops... I would have thought that should have been caught in bicep lint.

bennerv (approved):

PR #1659, opened 2025-04-22
Unit test refactoring

mbarnes:
### What

Refactoring unit tests to be easier to maintain.  Changes here fall into four categories:

1. Use available constants in tests instead of writing out the constant value.

2. Use the package "[github.com/stretchr/testify/require](https://pkg.go.dev/github.com/stretchr/testify/require)" during test setup where an unexpected condition should abort the whole test.  Just makes the test code more compact and consistent.

3. Use the package "[github.com/stretchr/testify/assert](https://pkg.go.dev/github.com/stretchr/testify/assert)" with functions being tested where an unexpected condition should fail the test case but continue the test.  Just makes the test code more compact and consistent.

4. Establish `/internal/api/testhelpers.go` as a place to share common constants and helper functions for unit tests.

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/openshift-online/ocm-sdk-go v0.1.464
 	github.com/prometheus/client_golang v1.21.0
 	github.com/spf13/cobra v1.9.1
+	github.com/stretchr/testify v1.10.0
 	go.uber.org/mock v0.5.0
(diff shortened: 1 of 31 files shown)


venkateshsredhat:
/internal/api/testhelpers.go -- This looks super helpful . While writing the unit tests for MI felt the pain . This makes it look so neat . Thanks !.

venkateshsredhat (approved):
LGTM . Thanks Matt .

PR #1661, opened 2025-04-23
backend: add subscription metrics

simonpasquier:
### What

This commit adds a new metric to the backend RP
(`backend_subscriptions`) reporting the number of subscriptions per state.

Example

```
backend_subscriptions{state="Deleted"} 0
backend_subscriptions{state="Registered"} 1
backend_subscriptions{state="Suspended"} 0
backend_subscriptions{state="Unregistered"} 0
backend_subscriptions{state="Warned"} 0
```

### Why

Provide more metrics on subscriptions.

### DIFF
--- a/backend/operations_scanner.go
+++ b/backend/operations_scanner.go
@@ -85,8 +85,8 @@ type OperationsScanner struct {
 	lockClient          *database.LockClient
 	clusterService      ocm.ClusterServiceClient
 	notificationClient  *http.Client
-	subscriptions       []string
 	subscriptionsLock   sync.Mutex
+	subscriptions       []string
 	subscriptionChannel chan string
 	subscriptionWorkers sync.WaitGroup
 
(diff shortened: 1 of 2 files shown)


simonpasquier:
cc @mbarnes (when you have time, no urgency)

backend/operations_scanner.go
```diff
@@ -151,6 +159,16 @@ func NewOperationsScanner(dbClient database.DBClient, ocmConnection *ocmsdk.Conn
 		s.lastOperationTimestamp.WithLabelValues(v)
 	}
 
+	for _, v := range []arm.SubscriptionState{
+		arm.SubscriptionStateRegistered,
+		arm.SubscriptionStateUnregistered,
+		arm.SubscriptionStateWarned,
+		arm.SubscriptionStateDeleted,
+		arm.SubscriptionStateSuspended,
```
mbarnes, line 167: I found it handy to have a "generator" function (in Python lingo) for ARM resource provisioning states - [ListProvisioningStates](https://github.com/Azure/ARO-HCP/blob/a16e7926fb42e87d22bcffc595f04156794dd0c0/internal/api/arm/resource.go#L130-L142) - to avoid having to write out all the constants.

I could see a similar "ListSubscriptionStates" function being handy too.  It would save you from having to write out all the constants here and in `collectSubscriptions`.

mbarnes (approved):
I like it.  Had a suggestion below for extra credit but feel free to pass on it.  Approved either way.  Let me know what you decide.

mbarnes (approved):

PR #1662, opened 2025-04-23
Ensure all .go files have copyright headers

mbarnes:
SSIA... well, almost.  By "all" I mean .go files under First-Party Team's purview, and excluding generated files.


frzifus (approved):

stevekuznetsov:
Best to add a CI check for this so you don't regress in the future. `find -type f -name '*.go' | xargs head -n 4 {} | grep` or something might get you there for a pretty low cost.

mbarnes:
> Best to add a CI check for this so you don't regress in the future. find -type f -name '*.go' | xargs head -n 4 {} | grep or something might get you there for a pretty low cost.

:+1: [ARO-RP](https://github.com/Azure/ARO-RP) has something for this.  I'll add it separately.

bennerv (approved):

PR #1664, opened 2025-04-23
Makefile: add a target to ensure licenses in Go files

stevekuznetsov:
Makefile: add a target to ensure licenses in Go files

---

*: add licenses to Go files

---


bennerv (approved):

stevekuznetsov:
CODEOWNERS wants the QE folks to review, since we touched their files - but we can be certain they are OK with this change, so I'll force it in to save us review cycles and time.

PR #1665, opened 2025-04-23
chore: use consts from OCM v1alpha1 package

simonpasquier:
### What

This PR updates the RP frontend and backend to use constants from the `github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1` package instead of `github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1` whenever applicable.

### Why

While the constant values are the same, it's more consistent to import them from the "right" package.

### DIFF
--- a/internal/ocm/internalid_test.go
+++ b/internal/ocm/internalid_test.go
@@ -16,10 +16,10 @@ package ocm
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"testing"
 
+	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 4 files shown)


simonpasquier:
cc @mbarnes

internal/ocm/internalid_test.go
```diff
@@ -85,14 +92,13 @@ func TestInternalID(t *testing.T) {
 			str := internalID.String()
 			assert.Equal(t, tt.path, str)
 
-			if kind == cmv1.NodePoolKind {
+			if kind == cmv1.NodePoolKind || kind == arohcpv1alpha1.NodePoolKind {
```
mbarnes, line 95: Don't need to check for both, it's the same string `"NodePool"`.
[cmv1.NodePoolKind](https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1#NodePoolKind)
[arohcpv1alpha1.NodePoolKind](https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1#NodePoolKind)

mbarnes (changes requested):
Thanks for this.  Just one stupid nit.

mbarnes (approved):

PR #1669, opened 2025-04-24
maintain topology.yaml

geoberle:
### What

* add stage ADO pipeline IDs
* remove pipelines that are part of other pipelines already
  * ACRPull runs as a step in each respective cluster pipeline
  * istio runs as part of the svc-cluster pipeline


janboll (approved):
:medal_military:

PR #1673, opened 2025-04-24
Fix commit SHA version in backend/frontend images

simonpasquier:
### What

This change ensures that the version reported by the backend and the frontend is effectively the commit SHA.

The tooling also adds a `-dirty` suffix if the Git repository has modified/untracked files.

### Why

The version reported from the container images was always "unknown".

It can supersede https://github.com/Azure/ARO-HCP/pull/1666


simonpasquier:
cc @mbarnes

stevekuznetsov:
Any reason not to cargo-cult the `tag.#commits-commit.{dirty|clean}` from OCP elsewhere? I've found the additional data quite useful.

stevekuznetsov:
@simonpasquier [this](https://github.com/openshift/origin/blob/6dc0a25d36c9491352d8a4c82edb902afbfe3e16/hack/lib/build/version.sh#L25) is the source

simonpasquier:
@stevekuznetsov do you feel that we need to handle tags (\*)? Or should we just append "-dirty" in case the repository isn't clean?

(\*)  IIUC there are no tags for now on ARO-HCP.

stevekuznetsov:
If there are no tags and we never expect to have any, no worries on omitting it. I thought we already had them! Oops.

simonpasquier:
Updated, the version should now be `<7 digits SHA>` or `<7 digits SHA>-dirty`  in case of uncommited changes.

stevekuznetsov (approved):

simonpasquier:
cc @SudoBrendan @mbarnes  for the frontend and backend changes.

mbarnes (approved):
Looks good to me.  Not sure if we'll end up using tags or not but this commit SHA logic has been revised several times already.  Probably won't be the last.

PR #1674, opened 2025-04-24
Disable dry run to actually clean up resources

janboll:
### What

Enable usage of managed identity by passing in MI id as parameter.

In order to deploy `register-scheduledrunbook.ps1` needs to be enhanced to support parameters


geoberle (approved):

YiqinZhang (approved):
/lgtm

PR #1677, opened 2025-04-24
Make ARM resources plural

mociarain:
This PR makes the custom ARM resource type plural. This is required


internal/api/registry.go
```diff
@@ -23,13 +23,13 @@ import (
...
-	NodePoolResourceTypeName        = "nodePools"
-	OperationResultResourceTypeName = "hcpOperationResults"
-	OperationStatusResourceTypeName = "hcpOperationsStatus"
-	ResourceTypeDisplay             = "Hosted Control Plane (HCP) OpenShift Clusters"
+	ProviderNamespace                 = "Microsoft.RedHatOpenShift"
+	ProviderNamespaceDisplay          = "Azure Red Hat OpenShift"
+	ClusterResourceTypeName           = "hcpOpenShiftClusters"
+	NodePoolResourceTypeName          = "nodePools"
+	OperationResultResourceTypeName   = "hcpOperationResults"
+	OperationStatusesResourceTypeName = "hcpOperationsStatuses"
```
mbarnes, line 31: I think we agreed over Slack to make it `hcpOperationStatuses` (double plurals is weird).

Also I'd prefer if the constant name stays singular like the others.  It reads better in code (imo).

mbarnes (changes requested):

mbarnes:
The CI failure is due to the pull request being from a forked repo.  But for this simple change I think we can ignore it.

mbarnes:
We went back and forth a bit more on the resource type name and finally settled on `hcpOperationStatuses`.
I ended up squashing the requested changes into your commit since you had left for the day.

SudoBrendan (approved):

PR #1693, opened 2025-04-28
clusters service pluralize

geoberle:
### What

Pluarlize `clusterService` to `cluster(s)Service` while remaining an no-op for all environments.
The actual rename of the namespace will be done in a followup PR that will then be a lot easier to review.

https://issues.redhat.com/browse/ARO-15469


miguelsorianod:
lgtm

PR #1697, opened 2025-04-28
add negative cluster and nodepool test examples

mgahagan73:
[ARO-15277](https://issues.redhat.com/browse/ARO-15277)
[ARO-15993](https://issues.redhat.com/browse/ARO-15993)

### What

This PR adds 2 sample negative tests for cluster and node pool creation

### Why

These are reference examples for cluster/node pool creation. Plus they also serve as tests to confirm we are handling failure to create a cluster or node pool properly.


test/e2e/create_cluster_test.go
```diff
@@ -0,0 +1,39 @@
+package e2e
```
patriksuba, line 1: Use name of file from minimal structure document i.e. cluster_put_test.go.

test/e2e/create_cluster_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	api "github.com/Azure/ARO-HCP/internal/api/v20240610preview/generated"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+)
+
+var _ = Describe("Create Cluster operation", func() {
```
patriksuba, line 14: Instead of "Create" use "Put" so it is consistent with HTTP method used by client. It is better to use "HCPOpenShiftCluster" instead of just "cluster" to keep full RP resource naming. Do no use word operation so that description can be read as a complete sentence. Final sentence "Put HCPOpenShiftCluster: it Attempts to create cluster with non-existent Resource Group by: preparing HCP clusters client (and) by: sending request to create cluster".

test/e2e/create_cluster_test.go
```diff
@@ -0,0 +1,39 @@
...
+	It("Attempt to create cluster with non-existant Resource Group", labels.Medium, labels.Negative, func(ctx context.Context) {
+		clusterName := "non-existing-cluster"
+		customerRGName := "non-existing-group"
+		var (
+			clusterResource api.HcpOpenShiftCluster
+			clusterOptions  *api.HcpOpenShiftClustersClientBeginCreateOrUpdateOptions
+		)
+		By("Send request to create cluster")
+		_, err := clustersClient.BeginCreateOrUpdate(ctx, customerRGName, clusterName, clusterResource, clusterOptions)
+		// _, err := clustersClient.Get(ctx, customerRGName, clusterName, nil)
```
patriksuba, line 33: Please do not leave commented code which is not used.

test/e2e/create_nodepool_test.go
```diff
@@ -0,0 +1,39 @@
+package e2e
```
patriksuba, line 1: The content of the file is the same as create_cluster_test.go. Probably a mistake.

patriksuba (commented):
Thank you for the PR. Most comments are description/structure related and opinions which should be addressed to keep code consistent in all files. Test for negative cluster creation is OK. Take a look at nodepool test and address lint error.

test/e2e/create_cluster_test.go
mbukatov: > Instead of "Create" use "Put" so it is consistent with HTTP method used by client. 

Here I disagree a bit as I would prefer sematic descriptions ... (as I noted in the document draft).

> Do no use word operation so that description can be read as a complete sentence.

+1

test/e2e/create_cluster_test.go
mbukatov: Here I disagree a bit, as I would prefer a semantic naming (let's discuss it in the labels doc).

patriksuba (dismissed):

test/e2e/create_cluster_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Expect(err).ToNot(BeNil())
+		//errMessage := fmt.Sprintf("The resource 'hcpOpenShiftClusters/%s' under resource group '%s' was not found.", clusterName, customerRGName)
+		errMessage := fmt.Sprintf("RESPONSE 500: 500 Internal Server Error")
```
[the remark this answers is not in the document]
mbukatov: Ah, ok.

mbukatov (approved):

test/e2e/nodepool_create_test.go
```diff
@@ -0,0 +1,51 @@
...
+		By("Preparing HCP nodepools client")
+		NodePoolsClient = clients.NewNodePoolsClient()
+	})
+
+	It("Attempts to create a nodepool for a non-existant HCPOpenshiftCluster", labels.Medium, labels.Negative, func(ctx context.Context) {
+		var (
+			nodePoolName     = "mynodepool"
+			clusterName      = "non-existing_cluster"
+			nodePoolResource api.NodePool
+			nodePoolOptions  *api.HcpOpenShiftClustersClientBeginCreateOrUpdateOptions
```
stevekuznetsov, line 42: Why not type this as `*api.NodePoolsClientBeginCreateOrUpdateOptions` and not need to cast? You only ever use it once, where does this type come from?

mbukatov (approved):

test/e2e/nodepool_create_test.go
mgahagan73: I think when I first started writing this there wasn't a NodePoolsClientBeginCreateOrUpdateOptions yet. Thanks for catching this, I've changed nodePoolOptions to use NodePoolsClientBeginCreateOrUpdateOptions.

PR #1698, opened 2025-04-29
Create NSP on MGMT and global, refactor SVC

janboll:
### What

Create NSP on MGMT and global, refactor SVC

### Why

Need to configure more than one NSP profile, having overlaps between templates, hence create a dedicated one for mgmt.
Add nsp for global and refactor svc to align with new modules.

### DIFF
--- a/dev-infrastructure/modules/network/nsp-profile.bicep
+++ b/dev-infrastructure/modules/network/nsp-profile.bicep
@@ -1,0 +1,64 @@
+param location string
+
+@description('Access mode for this NSP')
+@allowed(['Audit', 'Enforced', 'Learning'])
+param accessMode string
+
+@description('Resource IDs to associate with this NSP')
+param associatedResources array
+
+@description('The Name the NSP should have')
+param nspName string
+
+@description('The Name the NSP profile should have')
+param profileName string
+
+@description('Array of IPs that will be allowd to access NSP')
+param addressPrefixes array = []
+
+@description('Array of Service Tags that will be allowd to access NSP')
+param serviceTags array = []
+
+@description('Array of Subscription ids that will be allowd to access NSP')
+param subscriptions array = []
+
+var subscriptionObjects = [for s in subscriptions: { 'id': s }]
+
+resource nsp 'Microsoft.Network/networkSecurityPerimeters@2024-06-01-preview' existing = {
+  name: nspName
+}
+
+resource nspProfile 'Microsoft.Network/networkSecurityPerimeters/profiles@2024-06-01-preview' = {
+  parent: nsp
+  location: location
+  name: profileName
+}
+
+resource accessRule 'Microsoft.Network/networkSecurityPerimeters/profiles/accessRules@2024-06-01-preview' = if (length(addressPrefixes) > 0 || length(serviceTags) > 0 || length(subscriptionObjects) > 0) {
+  parent: nspProfile
+  location: location
+  name: '${nspName}-inbound-accessRule'
+  properties: {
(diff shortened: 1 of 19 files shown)


dev-infrastructure/templates/svc-cluster.bicep
```diff
@@ -614,13 +614,25 @@ module fpaCertificate '../modules/keyvault/key-vault-cert.bicep' = if (manageFpa
...
+}
+
+module svcNSPProfile '../modules/network/nsp-profile.bicep' = {
+  name: 'profile-${uniqueString(resourceGroup().name)}'
   params: {
     accessMode: svcNSPAccessMode
     nspName: svcNSPName
+    profileName: svcNSPName
     location: location
     associatedResources: [
```
geoberle, line 630: please add a TODO for service KV, postgres DBs and EV2 service tags

janboll: Postgres DBs are not supported :(

dev-infrastructure/modules/network/nsp-profile.bicep
```diff
@@ -0,0 +1,64 @@
+param location string
+
+@description('Access mode for this NSP')
+@allowed(['Audit', 'Enforced', 'Learning'])
```
geoberle, line 4: please remove this. EV2 precompiles the bicepparam files but leaves the placeholders in them during validation, which then fails this validation

dev-infrastructure/templates/svc-infra.bicep
```diff
@@ -116,5 +123,34 @@ module serviceKeyVaultDevopsSecretsOfficer '../modules/keyvault/keyvault-secret-
...
+
+module svcNSPProfile '../modules/network/nsp-profile.bicep' = {
+  name: 'profile-svc-kv-${uniqueString(resourceGroup().name)}'
+  params: {
+    accessMode: svcKVNSPAccessMode
+    nspName: svcKVNSPName
+    profileName: svcKVNSPName
+    location: serviceKeyVaultLocation
+    associatedResources: [
+      serviceKeyVault.outputs.kvId
```
geoberle, line 146: we should not manage the same resource (the profile) in two places. i suggest we move the svc kv association to service-cluster.bicep

janboll: It's a different NSP in the service key vault rg (global)

geoberle: ah i see. it is just in the global RG in DEV for reuse, in MSFT envs it is in the SVC RG. is there a way to get it into to NSP in the SVC RG or is there something limiting us from doing it?

janboll: Hm, it's a trade off, cause we have it in the global in integrated dev and in the svc rg in int/stage. I tend to let it live near the KV, cause otherwise the template resources would be split over two RGs in integrated dev otherwise.

geoberle: i'm not sure i understand. isn't this "just" adding the SVC KV resourceID to the SVC NSP in svc-cluster.bicep?

janboll: Please check lines above: https://github.com/Azure/ARO-HCP/pull/1698/files/64fbe064c7ce82a4e0edefb2545853e5f87e887e#diff-3af1c68f034209b2f567963719477f3483c0a6cf97a24497fe79efa81683c439R130-R136

geoberle (approved):

PR #1702, opened 2025-04-29
bugfix: support to marshal additionalProperties unformed types

bennerv:
no jira

### Why
When attempting to re-register the subscription with the resource provider, we failed to marshal the sent PUT request `additionalProperties` section.

`cannot unmarshal object into Go struct field SubscriptionProperties.properties.additionalProperties of type string`

The resource provider contract says the below, hence why we just make it an `any` type.  
> Resource Providers should develop code that can handle modifications, and should not strongly parse this dictionary.

### DIFF
--- a/internal/api/arm/subscription.go
+++ b/internal/api/arm/subscription.go
@@ -48,7 +48,7 @@ type SubscriptionProperties struct {
 	SpendingLimit        *string              `json:"spendingLimit,omitempty"`
 	AccountOwner         *AccountOwner        `json:"accountOwner,omitempty"`
 	ManagedByTenants     *[]map[string]string `json:"managedByTenants,omitempty"`
-	AdditionalProperties *map[string]string   `json:"additionalProperties,omitempty"`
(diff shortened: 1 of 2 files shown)


internal/api/arm/subscription.go
```diff
@@ -48,7 +48,7 @@ type SubscriptionProperties struct {
 	SpendingLimit        *string              `json:"spendingLimit,omitempty"`
 	AccountOwner         *AccountOwner        `json:"accountOwner,omitempty"`
 	ManagedByTenants     *[]map[string]string `json:"managedByTenants,omitempty"`
-	AdditionalProperties *map[string]string   `json:"additionalProperties,omitempty"`
+	AdditionalProperties any                  `json:"additionalProperties,omitempty"`
```
mbarnes, line 51: If we ever want to read anything from `AdditionalProperties`, defining it as `map[string]any` would be better. But for now it doesn't matter.

mbarnes (approved):
Approved with a non-blocking note.

PR #1704, opened 2025-04-30
Remove quotes from OP_[A-Z_]+_NAME variables in CS Makefile

kostola:
### What

`OP_[A-Z_]+_NAME` variables in `cluster-service/Makefile` contains already double quoted values, so double quoting them again causes commands to look like:

```
helm template deploy -s templates/azure-operators-managed-identities-config.configmap.yaml \
	  --set azureOperatorsMI.cloudControllerManager.roleName=""Azure Red Hat OpenShift Cloud Controller Manager - Dev"" \
	  ...
```

which returns an error.

This PR aims to fix this.


kostola:
#1706 fixes the issue in a cleaner way. Closing this.

PR #1708, opened 2025-04-30
Improve handling of Cluster Service errors in RP frontend

mbarnes:
Related to [ARO-15570 - NodePool Static Validation](https://issues.redhat.com/browse/ARO-15570)

### What

This is an attempt to improve handling of Cluster Service error responses, primarily for the purpose of propagating validation errors that the RP does not catch.

### Why

But for a few cases where we check for a "404 Not Found" response, any error from Cluster Service currently triggers an opaque "500 Internal Server Error" response from the RP.  This needs improved upon.

This PR introduces a `CSErrorToCloudError` helper function that attempts to preserve select status codes from Cluster Service and turn them into compliant ARM error responses:

- **400 Bad Request**
  This status code has a broad range of applications but in lieu of a distinct secondary code from Cluster Service we assume this means a validation error and propagate the "[reason](https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go/errors#Error.Reason)" message from Cluster Service with an "[InvalidRequestContent](https://github.com/Azure/ARO-HCP/blob/b1e287d13121c677a48a10942471ad3ac18ce0b1/internal/api/arm/error.go#L29)" Azure error code.

- **404 Not Found**
  If this error occurs in the context of an ARM resource request, we use our own error message that references the resource ID.  Otherwise we propagate the "[reason](https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go/errors#Error.Reason)" message from Cluster Service.

- **409 Conflict**
  Cluster Service will soon return this status code in the context of [break-glass credentials](https://docs.google.com/document/d/1Qttlxpnv_d5zsoj4LAgLpr8ZUMCne92e8-prn8G5ILY/edit?tab=t.0).  There may be other cases I'm not aware of where Cluster Service returns this status code.  Here too we propagate the "[reason](https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go/errors#Error.Reason)" message from Cluster Service.

Any other status code becomes an opaque "500 Internal Server Error" response, as before.

The list of recognized status codes and the particulars of how they're handled may change over time.  The main motivation here was to establish a centralized place to do these kinds of error conversions.

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -317,7 +317,7 @@ func (f *Frontend) ArmResourceList(writer http.ResponseWriter, request *http.Req
 	// Check for iteration error.
 	if err != nil {
 		logger.Error(err.Error())
-		arm.WriteInternalServerError(writer)
+		arm.WriteCloudError(writer, CSErrorToCloudError(err, nil))
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


venkateshsredhat (approved):
Lgtm

PR #1709, opened 2025-04-30
Align node pool defaults with Cluster Service

mbarnes:
Related to [ARO-15570 - NodePool Static Validation](https://issues.redhat.com/browse/ARO-15570)

### What

Corrects a couple default value inconsistencies with Cluster Service.

- `NodePoolProperties.autoRepair` : [Cluster Service defaults to true](https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/service/azure/node_pool_defaults.go?ref_type=heads#L43-47)
  ARM API has had the default set to **false** for some reason since its initial draft.  I'm assuming the default CS chose had more intention behind it than our initial API draft.

- `NodePoolPlatformProfile.diskSizeGiB` : [Cluster Service defaults to 64](https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/service/azure/node_pool_defaults.go?ref_type=heads#L60-69)
  ARM API has no default, nor does it mark the field as required.  So 64 GiB is currently an _implicit_ default that CS assigns.  But changing a default, whether it's explicit or implicit, has the same downstream impacts which is why Microsoft considers it a breaking change.  There's no real benefit to leaving it implicit so we may as well declare it in the API, imo.

### DIFF
--- a/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
+++ b/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
@@ -411,7 +411,7 @@ model NodePoolPlatformProfile {
   vmSize: string;
 
   /** The OS disk size in GiB */
-  diskSizeGiB?: int32;
+  diskSizeGiB?: int32 = 64;
 
   /** The type of the disk storage account
    * - https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types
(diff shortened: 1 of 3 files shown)


SudoBrendan (approved):

mbarnes:
I'll roll this into our azure-rest-api-specs-pr branch.

PR #1711, opened 2025-05-01
Finish node pool static validation

mbarnes:
[ARO-15570 - NodePool Static Validation](https://issues.redhat.com/browse/ARO-15570)

### What

More static validation for clusters and node pools.  A few gaps remain yet for clusters but the focus here was on node pools so we can close the [node pool day-2 operations epic](https://issues.redhat.com/browse/ARO-10276).

This introduces a couple new dependencies to assist with syntax parsing:

- `github.com/hashicorp/go-version` for semantic version parsing (`properties.version.id`)
- `k8s.io/apimachinery/pkg/util/validation` for node pool label and taint parsing

This also adds a little bit more semantic validation, where certain node pool fields are compared to parent cluster fields.

~The commits from #1708 are duplicated here because there was one commit that calls `CSErrorToCloudError`.~


SudoBrendan (approved):

internal/api/validate.go
```diff
@@ -93,6 +94,43 @@ func NewValidator() *validator.Validate {
...
+	err = validate.RegisterValidation("openshift_version", func(fl validator.FieldLevel) bool {
+		field := fl.Field()
+		if field.Kind() != reflect.String {
+			panic("String type required for openshift_version")
+		}
+		_, err := NewOpenShiftVersion(field.String())
+		return err == nil
+	})
+	if err != nil {
+		panic(err)
```
bennerv, line 131: Should we be panic'ing on error?  

If so, are we handling goroutines panicking not taking down he whole app?

mbarnes: > Should we be panic'ing on error?

We're panic'ing on a registration error, which will happen on startup either 100% of the time or never, kinda like calling `regex.MustCompile`.  In this particular case, the only way the registration fails is if you pass nonsense arguments like a `nil` function.

> If so, are we handling goroutines panicking not taking down he whole app?

More generally, yes we do have middleware in the frontend to catch a panic.

PR #1712, opened 2025-05-02
setup new int identities

tony-schndr:
### What
Adds a make target and updates bicep module for creating new identities for the INT environment.

### Why

Previously the INT environment was using the current development EntraId Apps (fp, msi mock, arm helper).

Creating / setting up the identities is complete, need to deploy CS (from this branch) to INT to make them active/test.  Before that hosted clusters should be deleted as the RP will lose access to them after these identities become active.  The new Entra Id Apps have permissions over `ARO SRE Team - INT (EA Subscription 3)`.


jfchevrette:
Merged via https://github.com/Azure/ARO-HCP/pull/1780

PR #1716, opened 2025-05-05
Conditionally create NSP profile for SVC KeyVault

janboll:
### What
Conditionally create NSP profile for SVC KeyVault
### Why

Key vault is shared. A resource can only be assigned to one NSP in learning mode. Hence ensure it is only associated to an NSP once. Since in integrated dev all resources share the same subscription access should still work

### DIFF
--- a/dev-infrastructure/templates/svc-cluster.bicep
+++ b/dev-infrastructure/templates/svc-cluster.bicep
@@ -239,6 +239,9 @@ param svcNSPName string
 @description('Access mode for this NSP')
 param svcNSPAccessMode string
 
+@description('Access mode for this NSP')
+param rhDevFixSVCKVAsignNSP bool = true
+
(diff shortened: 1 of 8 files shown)


mmazur (approved):

dev-infrastructure/templates/svc-cluster.bicep
```diff
@@ -629,6 +632,22 @@ module svcNSPProfile '../modules/network/nsp-profile.bicep' = {
...
       rpCosmosDb.outputs.cosmosDbAccountId
+    ]
+    // TODO Add EV2 access here
+    subscriptions: [
+      subscription().id
+    ]
+  }
+}
+
+module svcKVNSPProfile '../modules/network/nsp-profile.bicep' = if (rhDevFixSVCKVAsignNSP) {
```
geoberle, line 643: late review: would the condition `resourceGroup().name == serviceKeyVaultResourceGroup` work as well?

PR #1718, opened 2025-05-05
manual SRE breakglass script

geoberle:
### What

this PR contains a quick and dirty POC for a manual SRE breakglass script, executed directly towards the MC of an HCP

TLDR:
* use the `sre-break-glass` signer to sign a CSR for an HCP
  * use the organization `O` of the CSR subject to carry the information about groupmembership
* use the resulting certificate to craft a kubeconfig
* use an ACM policy to grant the group mentioned in `O` certain permissions on the HCP
* setup kubectl portforward to the HCP KAS

(!) don't merge! this is only to show that the procedure basically works and grants access to an HCP in conjunction with an ACM policy.

https://issues.redhat.com/browse/ARO-17406

### DIFF
--- a/hack/hcp-sre-breakglass.sh
+++ b/hack/hcp-sre-breakglass.sh
@@ -1,10 +1,7 @@
 #!/bin/sh
 #
 # E X P E R I M E N T A L   -   N O T   I N T E N D E D   F O R   P R O D U C T I O N
-#
-# Use at your own risk. This script does not do sanity checks and does not have error handling.
-# In case of errors, resources may be left behind, etc.
-# TLDR: this script is the worst but good enough for a quick and dirty POC.
+# Use at your own risk. This script does not do sanity checks and only partial error handling.
 #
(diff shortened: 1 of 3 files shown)


hack/hcp-sre-breakglass.sh
```diff
@@ -0,0 +1,128 @@
...
+    api.openshift.com/name: $USER
+    api.openshift.com/type: break-glass-credential
+  name: ${CSR_NAME}
+  namespace: ${HCP_NAMESPACE}
+EOF
+
+echo "Hypershift CSR Approval ${CSR_NAME} applied to the cluster."
+
+# wait for the CSR to be approved
+while true; do
```
stevekuznetsov, line 83: `kubectl wait --for` on conditions is what you want here :)

geoberle: good point. done

venkateshsredhat:
thanks for adding the SOP , Backplane will be doing the exact same way but we would rely on Maestro for applying Manifest. This SOP would be a go to during some kind of Maestro issue . :+1:

PR #1724, opened 2025-05-06
Enable advanced container networking

janboll:
### What

Enable advanced container networking on clusters that run cilium

[ARO-16981](https://issues.redhat.com/browse/ARO-16981)


janboll:
Failur in `[CS PR Check Environment Bicep What-If / what-if (pull_request)](https://github.com/Azure/ARO-HCP/actions/runs/14854313032/job/41703872609?pr=1724)` is caused by nodepool deployment, hence ignoring error since dev passed.

PR #1729, opened 2025-05-06
add infra nodes for prometheus workloads

tony-schndr:
https://issues.redhat.com/browse/ARO-16948

### What
This PR adds a new infra agent pool to svc/mgmt clusters and configures the scheduling of Prometheus pods to those nodes.  
HA is handled with topology spread constraints using the failure domain zone label which ensures that each Prometheus shard and its replica are not deployed to the same az.

### Why

Ensure prometheus runs on dedicate nodes with HA.


jfchevrette (approved):
LGTM

PR #1730, opened 2025-05-06
move PIPs from svc RG to regional RG

jfchevrette:
[ARO-16089](https://issues.redhat.com/browse/ARO-16089)

### What

Move the PIP resources used by the AKS cluster from the SVC resource group to the regional resource group

### Why

MSFT prefers the SVC cluster RG to be about the AKS cluster. All other resources like DBs, storage accounts, managed identities for service, etc should be in the regional RG.

See [ARO-16089](https://issues.redhat.com/browse/ARO-16089)

As this will require a rebuild of the dev & cspr environments we should only merge this when we are ready to do so


jfchevrette:
Merged via https://github.com/Azure/ARO-HCP/pull/1780

PR #1732, opened 2025-05-06
add cluster label to oss Prometheus scraped metrics

tony-schndr:
https://issues.redhat.com/browse/ARO-16995

### What
Adds a relabel config to remote write configuration to add a "cluster" label
### Why

This label was provided by default through Azure Managed Prometheus but was lost when moving to OSS Prometheus.


simonpasquier:
an alternative is to setup the label as an external label which is automatically added not only to remote write samples but also alerts and federation (in case these are used later).

simonpasquier (approved):
lgtm

PR #1745, opened 2025-05-09
Define billing doc object

gouthamMN:
### What
Add a struct for billing documents.

### Why
We need this object for hcp cluster billing

### DIFF
--- a/internal/database/document.go
+++ b/internal/database/document.go
@@ -52,6 +52,24 @@ type DocumentProperties interface {
 	GetValidTypes() []string
 }
 
+type BillingDocument struct {
+	// The cluster creation time represents the time when the cluster was provisioned successfully
+	CreationTime time.Time `json:"creationTime,omitempty"`
+	// The cluster deletion time
+	DeletionTime *time.Time `json:"deletionTime,omitempty"`
+	// The cluster last billing time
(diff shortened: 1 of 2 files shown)


internal/database/document.go
```diff
@@ -52,6 +52,24 @@ type DocumentProperties interface {
...
+	// The cluster last billing time
+	LastBillingTime *time.Time `json:"lastBillingTime,omitempty"`
+
+	// The location of the HCP cluster
+	Location string `json:"location,omitempty"`
+	// The tenant ID of the HCP cluster
+	TenantID string `json:"tenantId,omitempty"`
+	// The hcp cluster ARM resource ID
+	ResourceID string `json:"resourceId,omitempty"`
+	// The managed resource group of the hcp cluster
```
mbarnes, line 69: I can't remember if we decided this is just the managed resource group _name_ or the full resource ID (/subscriptions/...).  Can you clarify the comment for my bad memory?  Either
```
// The name of the managed resource group for the HCP cluster
```
or
```
// The resource ID of the managed resource group for the HCP cluster
```

gouthamMN: That's a full resource ID (/subscriptions/...) from which we can get the sub ID and other info.

mbarnes (approved):
Perfect.  Thanks for fixing the lint error!

PR #1747, opened 2025-05-11
Use Cosmos DB transactions and patching

mbarnes:
This finishes the "[Use Cosmos DB's Transactional Batch Operations in RP](https://issues.redhat.com/browse/ARO-14167)" epic.

### What

Adds methods to the `DBClient` interface to utilize [transactional batch operations](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/transactional-batch) as well as [partial document updates](https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update) (a.k.a. JSON patches) instead of full document replacement.

The backend, it turns out, does not need transactions because it always patches documents with a precondition predicate that we expect may fail.  A failed precondition in a transaction would abort the whole transaction and that's not what we want in the backend.  A failed precondition in the backend does not jeopardize the integrity of data in Cosmos DB.

The frontend, on the other hand, uses transactions extensively to ensure resource and operation document writes succeed or fail together.  A failed transaction in the frontend would still introduce discrepancies between the state of Cluster Service and Cosmos DB, but that is [easier to reconcile](https://issues.redhat.com/browse/ARO-10838).

### Why

See the [epic description](https://issues.redhat.com/browse/ARO-14167) for a long-winded explanation of why.

As usual, I strongly suggest reviewing commits one at a time.


simonpasquier (approved):
LGTM

venkateshsredhat (approved):
LGTM

PR #1755, opened 2025-05-14
Added new attributes to the managed identities config according to the new schema

anshulvermapatel:
https://issues.redhat.com/browse/ARO-17371

### What

To support multiple azure role definition for both the defined control plane and data plane operators managed identities, a new schema has been finalized and this is to add attributes for the clusters-service managed identities config according to the new schema. This is to make sure the successful deployed after the clusters-service code change to support multiple managed identities for the azure operators.

### DIFF
--- a/cluster-service/Makefile
+++ b/cluster-service/Makefile
@@ -72,13 +72,31 @@ deploy:
 	  --set azureOperatorsMI.imageRegistry.roleId="$${OP_IMAGE_REGISTRY_DRIVER_ROLE_ID}" \
 	  --set azureOperatorsMI.cloudNetworkConfig.roleName="${OP_CLOUD_NETWORK_CONFIG_ROLE_NAME}" \
 	  --set azureOperatorsMI.cloudNetworkConfig.roleId="$${OP_CLOUD_NETWORK_CONFIG_ROLE_ID}" \
+	  --set azureOperatorsMI.kms.roleName="${OP_KMS_ROLE_NAME}" \
+	  --set azureOperatorsMI.kms.roleId="$${OP_KMS_ROLE_ID}" \
+	  --set azureOperatorsMI.clusterApiAzure.roleDefinitions[0].name="${OP_CLUSTER_API_AZURE_ROLE_NAME}" \
+	  --set azureOperatorsMI.clusterApiAzure.roleDefinitions[0].id="$${OP_CLUSTER_API_AZURE_ROLE_ID}" \
+	  --set azureOperatorsMI.controlPlane.roleDefinitions[0].name="${OP_CONTROL_PLANE_ROLE_NAME}" \
+	  --set azureOperatorsMI.controlPlane.roleDefinitions[0].id="$${OP_CONTROL_PLANE_ROLE_ID}" \
+	  --set azureOperatorsMI.cloudControllerManager.roleDefinitions[0].name="${OP_CLOUD_CONTROLLER_MANAGER_ROLE_NAME}" \
+	  --set azureOperatorsMI.cloudControllerManager.roleDefinitions[0].id="$${OP_CLOUD_CONTROLLER_MANAGER_ROLE_ID}" \
+	  --set azureOperatorsMI.ingress.roleDefinitions[0].name="${OP_INGRESS_ROLE_NAME}" \
+	  --set azureOperatorsMI.ingress.roleDefinitions[0].id="$${OP_INGRESS_ROLE_ID}" \
+	  --set azureOperatorsMI.diskCsiDriver.roleDefinitions[0].name="${OP_DISK_CSI_DRIVER_ROLE_NAME}" \
+	  --set azureOperatorsMI.diskCsiDriver.roleDefinitions[0].id="$${OP_DISK_CSI_DRIVER_ROLE_ID}" \
+	  --set azureOperatorsMI.fileCsiDriver.roleDefinitions[0].name="${OP_FILE_CSI_DRIVER_ROLE_NAME}" \
+	  --set azureOperatorsMI.fileCsiDriver.roleDefinitions[0].id="$${OP_FILE_CSI_DRIVER_ROLE_ID}" \
+	  --set azureOperatorsMI.imageRegistry.roleDefinitions[0].name="${OP_IMAGE_REGISTRY_DRIVER_ROLE_NAME}" \
+	  --set azureOperatorsMI.imageRegistry.roleDefinitions[0].id="$${OP_IMAGE_REGISTRY_DRIVER_ROLE_ID}" \
+	  --set azureOperatorsMI.cloudNetworkConfig.roleDefinitions[0].name="${OP_CLOUD_NETWORK_CONFIG_ROLE_NAME}" \
+	  --set azureOperatorsMI.cloudNetworkConfig.roleDefinitions[0].id="$${OP_CLOUD_NETWORK_CONFIG_ROLE_ID}" \
+	  --set azureOperatorsMI.kms.roleDefinitions[0].name="${OP_KMS_ROLE_NAME}" \
+	  --set azureOperatorsMI.kms.roleDefinitions[0].id="$${OP_KMS_ROLE_ID}" \
 	  --set pullBinding.workloadIdentityClientId="$${IMAGE_PULLER_MI_CLIENT_ID}" \
 	  --set pullBinding.workloadIdentityTenantId="$${IMAGE_PULLER_MI_TENANT_ID}" \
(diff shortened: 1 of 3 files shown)


cluster-service/deploy/templates/azure-operators-managed-identities-config.configmap.yaml
```diff
@@ -8,52 +8,102 @@ data:
     controlPlaneOperatorsIdentities:
       cluster-api-azure:
         minOpenShiftVersion: 4.18
+        roleDefinitions:
```
miguelsorianod, line 11: Assuming that this defines a roleDefinitions where the value is an array of structs where the structs have the resourceId and name keys it seems good to me

cluster-service/deploy/templates/azure-operators-managed-identities-config.configmap.yaml
```diff
@@ -8,52 +8,102 @@ data:
...
+            name: '{{ .name }}'
+        {{- end }}
         azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/{{ .Values.azureOperatorsMI.fileCsiDriver.roleId }}'
         azureRoleDefinitionName: '{{ .Values.azureOperatorsMI.fileCsiDriver.roleName }}'
         optional: false
       image-registry:
         minOpenShiftVersion: 4.18
+        roleDefinitions:
+        {{- range .Values.azureOperatorsMI.imageRegistry.roles }}
+          - resourceId: '/providers/Microsoft.Authorization/roleDefinitions/{{ .id }}'
```
miguelsorianod, line 73: Would the variable be .id or .resourceId?

miguelsorianod:
I'll defer the review of changes around helm to the lifecycle team

cluster-service/deploy/templates/azure-operators-managed-identities-config.configmap.yaml
anshulvermapatel: Yes indeed it is an array of roledefinitions.. I had to deploy a dev env .. and I deployed with this configuration .. have a look at the CS configmap:
```
file-csi-driver:
        minOpenShiftVersion: 4.18
        roleDefinitions:
          - resourceId: '/providers/Microsoft.Authorization/roleDefinitions/xxxxxxxxxxxxxx'
            name: 'File Storage Operator - Dev'
        azureRoleDefinitionResourceId: '/providers/Microsoft.Authorization/roleDefinitions/xxxxxxxxxxxxxxxxxx'
        azureRoleDefinitionName: 'File Storage Operator - Dev'
```

cluster-service/deploy/templates/azure-operators-managed-identities-config.configmap.yaml
anshulvermapatel: I just put `id` thinking of the context .. however, `.resourceId` should be good.

anshulvermapatel: I am sorry .. actually, this is roleId .. and since this is the array of roleIds and role names .. I chose just `id` and `name`. I am for now reverting back to `id` from `resourceId`

cluster-service/deploy/templates/azure-operators-managed-identities-config.configmap.yaml
```diff
@@ -8,52 +8,102 @@ data:
     controlPlaneOperatorsIdentities:
       cluster-api-azure:
         minOpenShiftVersion: 4.18
+        roleDefinitions:
+        {{- range .Values.azureOperatorsMI.clusterApiAzure.roles }}
+          - resourceId: '/providers/Microsoft.Authorization/roleDefinitions/{{ .id }}'
```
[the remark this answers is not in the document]
anshulvermapatel: So it is not the full resource id, if you look at it, the role id is getting embedded in the resource id. So it is actually roleId not resourceId, I am reverting my previous changes also to `id` from `resourceId`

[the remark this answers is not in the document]
machi1990: ack, thanks for confirming

machi1990:
lgtm

cluster-service/Makefile
```diff
@@ -72,13 +72,31 @@ deploy:
 	  --set azureOperatorsMI.imageRegistry.roleId="$${OP_IMAGE_REGISTRY_DRIVER_ROLE_ID}" \
 	  --set azureOperatorsMI.cloudNetworkConfig.roleName="${OP_CLOUD_NETWORK_CONFIG_ROLE_NAME}" \
 	  --set azureOperatorsMI.cloudNetworkConfig.roleId="$${OP_CLOUD_NETWORK_CONFIG_ROLE_ID}" \
+	  --set azureOperatorsMI.kms.roleName="${OP_KMS_ROLE_NAME}" \
+	  --set azureOperatorsMI.kms.roleId="$${OP_KMS_ROLE_ID}" \
+	  --set azureOperatorsMI.clusterApiAzure.roles[0].name="${OP_CLUSTER_API_AZURE_ROLE_NAME}" \
```
miguelsorianod, line 77: I see it references roles, name and id.  The attribute names don't seem to match what's specified in the values.yaml file.
Is that correct and intended?

anshulvermapatel: no they are incorrect. 

Correted them now, please have a look.

miguelsorianod: Although I see now we reference roleDefinitions I still see references to .name and .id instead of .roleName and .resourceId. Is that correct and intended? same applies to the references in the configmap yaml

anshulvermapatel: Yes that is intended. I kept it like that as this is an array of roleDefinitions.

miguelsorianod: If it is intended, why is that some are referenced as .id, name ( azureOperatorsMI.kms.roleName)  and others as .roleId, .roleName ( azureOperatorsMI.clusterApiAzure.roleDefinitions[0].name) ?

miguelsorianod: and how does that relate  to the attribute names in the values.yaml file ?

anshulvermapatel: `azureOperatorsMI.kms.roleName` this refer to the old schema of the configuration file which we need to keep until we deploy the new CS code. THe KMS one, I just shifted it with other similar flags just to keep them together.

miguelsorianod: Understood 👍 . That makes sense to me

anshulvermapatel:
> is there some way to generate the changes without needing to deploy an environment?

Yes there is.
https://github.com/Azure/ARO-HCP/blob/main/cluster-service/Makefile#L127-L156
@miguelsorianod

machi1990:
> > is there some way to generate the changes without needing to deploy an environment?
> 
> Yes there is. https://github.com/Azure/ARO-HCP/blob/main/cluster-service/Makefile#L127-L156 @miguelsorianod

+1. Can you run the command and share the output as part of the PR's description? This could be part of the manual verifications that the reviewers can see that they've already been done. It sets the tone for the expected output.

miguelsorianod:
> Here is how its is getting generated. I even deployed it in my dev env and it is working as expected.
> 
> ```
> % cat azure-operators-managed-identities-config.yaml
> controlPlaneOperatorsIdentities:
>   cluster-api-azure:
>     minOpenShiftVersion: 4.18
>     roleDefinitions:
>       - resourceId: '<resource_id>'
>         name: '<name>'
>     azureRoleDefinitionResourceId: '<resource_id>'
>     azureRoleDefinitionName: '<name>'
>     optional: false
>   control-plane:
>     minOpenShiftVersion: 4.18
>     roleDefinitions:
>       - resourceId: '<resource_id>'
>         name: '<name>'
>     azureRoleDefinitionResourceId: '<resource_id>'
>     azureRoleDefinitionName: '<name>'
>     optional: false
> .
> .
> .
> <enacted>
> .
> .
> .
> dataPlaneOperatorsIdentities:
>   disk-csi-driver:
>     minOpenShiftVersion: 4.18
>     roleDefinitions:
>       - resourceId: '<resource_id>'
>         name: '<name>'
>     azureRoleDefinitionResourceId: '<resource_id>'
>     azureRoleDefinitionName: '<name>'
>     k8sServiceAccounts:
>       - name: '<name>'
>         namespace: '<ns>'
>       - name: '<name>'
>         namespace: '<ns>'
>     optional: false
> ```

Thanks, this kind of information is very useful for the reviewers

miguelsorianod (approved):

janboll (approved):

PR #1765, opened 2025-05-15
Document asynchronous operations in the ARO-HCP RP

mbarnes:
Jira: [DDR to define async op status notification mechanism for RP](https://issues.redhat.com/browse/ARO-7951) (sort of)

### What

This covers Cosmos DB document structures and how the ARO-HCP resource provider uses them to implement the ARM long-running operation protocol in its frontend and backend pods.

### Why

This is my penance for never finishing the DDR I was assigned last year.  :slightly_smiling_face: 

Let's call this a first edition.  There's still more material to cover like admin credential operations, the Cosmos DB billing container and the RP's interaction with it, etc.  But I need to get a baseline merged or I'll just continue tweaking it forever.


frontend/asynchronous-operations.md
```diff
@@ -0,0 +1,761 @@
...
+The second thing is documents in a Cosmos DB container – aside from a few [system-generated fields](https://learn.microsoft.com/en-us/rest/api/cosmos-db/documents) – are free-form JSON blobs. Unlike with tables in a schema-based database, [documents within a Cosmos DB container do not have to be homogeneous](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/modeling-data#distinguish-between-different-document-types). This is worth taking advantage of considering the aforementioned scoping constraints.
+
+With these points in mind, ARO-HCP has settled on the following data model.
+
+### Containers
+
+The ARO-HCP resource provider interacts with the following containers:
+
+#### Resources
+"Resources" is the primary container where all the registered Azure subscriptions, hosted control plane (HCP) cluster and node pool metadata, and asynchronous operation tracking that the ARO-HCP resource provider is reponsible for live.
```
mociarain:
```suggestion
"Resources" is the primary container where all the registered Azure subscriptions, hosted control plane (HCP) cluster and node pool metadata, and asynchronous operation tracking that the ARO-HCP resource provider is responsible for live.
```

frontend/asynchronous-operations.md
```diff
@@ -0,0 +1,761 @@
...
+   ```
+   Within the ARO-HCP resource provider we call this the "internal ID" since it is hidden from the end-user.
+
+<a name="resource-document-activeoperationid-field"></a>
+
+3. The `activeOperationId` field refers to the `id` field of the Cosmos DB document representing the current asynchronous operation acting on the resource. In Azure there can only be one active operation on a resource at a time. This field is removed when the active asynchronous operation reaches a terminal state.
+
+<a name="resource-document-provisioningstate-field"></a>
+
+4. The `provisioningState` field stays syncrhonized with the [`status` field](#operation-document-status-field) of the Cosmos DB document representing the latest asynchronous PUT, PATCH, or DELETE operation initiated on the resource. The reason the value is duplicated here is to preserve the last operation's terminal status, potentially beyond the [limited lifespan](#operation-document-time-to-live) of asynchronous operation documents.
```
mociarain:
```suggestion
4. The `provisioningState` field stays synchronized with the [`status` field](#operation-document-status-field) of the Cosmos DB document representing the latest asynchronous PUT, PATCH, or DELETE operation initiated on the resource. The reason the value is duplicated here is to preserve the last operation's terminal status, potentially beyond the [limited lifespan](#operation-document-time-to-live) of asynchronous operation documents.
```

mociarain (approved):
This is great! I learned a lot more about our system reading this and I think with some more time and this I would be confident making changes to this system

PR #1769, opened 2025-05-16
Fix roleassignment cleanup bug

YiqinZhang:
### What

Add `-SubscriptionId `and `-AccountId` parameters to `Connect-AzAccount -Identity` so it login with the managed identity and directly sets the subscription context.

Fix the error:

```
Error
Please provide a valid tenant or a valid subscription.
Error
No subscription was found in the default profile and no scope was specified. Either specify a scope or use a tenant with a subscription to run the command.
Error
Wrong value for Objecttype, perhaps missing Directory Reader permissions on identity or IDs changed

```

### DIFF
--- a/tooling/azure-automation/resources-cleanup/src/clean-orphaned-role-assignments.ps1
+++ b/tooling/azure-automation/resources-cleanup/src/clean-orphaned-role-assignments.ps1
@@ -1,19 +1,23 @@
 Param
 (
   [Parameter (Mandatory= $false)]
-  [System.Boolean] $dryRun = $false
+  [System.Boolean] $dryRun = $false,
+  [Parameter (Mandatory= $false)]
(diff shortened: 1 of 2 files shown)


tooling/azure-automation/resources-cleanup/src/clean-orphaned-role-assignments.ps1
```diff
@@ -1,19 +1,23 @@
 Param
 (
   [Parameter (Mandatory= $false)]
-  [System.Boolean] $dryRun = $false
+  [System.Boolean] $dryRun = $false,
+  [Parameter (Mandatory= $false)]
+  [string] $SubscriptionId = "1d3378d3-5a3f-4712-85a1-2485495dfc4b",
+  [Parameter (Mandatory= $false)]
+  [string] $ManagedIdentityId
 )
```
tony-schndr, line 9: Is there a calling script/module that needs to be updated with these parameters?

PR #1781, opened 2025-05-21
POC - ACM policy to mirror an ingress TLS cert into a HCP KAS

geoberle:
this PR showcases how an ACM policy can be used to copy an HCP specific secret into the HCP KAS

this relies on the ACM `copySecretData` and `policy.spec.hubTemplate.serviceAccountName` functionality, as well as the `ManagedCluster.metadata.labels.hosted-cluster-namespace` label to identify the source namespace (`ocm-xxx-$clusterID`) for the secret that is supposed to be mirrored.

(!) this PR uses a quick and dirty new service account in the `open-cluster-management-policies` namespace with cluster admin permissions. we need to narrow these permissions down to list and watch on Secrets in the relevant `ocm-xxx-$clusterID` namespaces.


geoberle:
closed in favor of #1899

PR #1785, opened 2025-05-21
Use AKSCluster from ShellStep instead of ResourceGroup

janboll:
### Why

Implements refactoring in ARO-Tools: https://github.com/Azure/ARO-Tools/pull/9


janboll:
https://github.com/Azure/ARO-Tools/pull/9 needs to be merged first, after that this PR needs a `go get -u`

geoberle:
looks good. this allows us to cleanup a couple of more things in pipelines like `mgmt-pipeline.yaml` where the same sub/rg was mentioned twice, just once with and once without the aks reference. with your change, that is not necessary anymore

geoberle (approved):

PR #1787, opened 2025-05-21
fix admin api route

katherinelc321:
### What

Removes placeholder route with admin/helloworld route

### Why

Geneva Actions will be able to call the helloworld endpoint

### DIFF
--- a/admin/pkg/admin/routes.go
+++ b/admin/pkg/admin/routes.go
@@ -15,6 +15,7 @@
 package admin
 
 import (
+	"fmt"
 	"net/http"
 )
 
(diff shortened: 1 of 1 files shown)


admin/pkg/admin/routes.go
```diff
@@ -15,15 +15,18 @@
...
 	"net/http"
 )
 
 func (a *Admin) adminRoutes() *http.ServeMux {
 
 	adminMux := http.NewServeMux()
 
-	adminMux.HandleFunc("/v1/<something>/", func(writer http.ResponseWriter, request *http.Request) {
-		// Queries something
+	adminMux.HandleFunc("/admin/hellowworld", func(writer http.ResponseWriter, request *http.Request) {
```
mbarnes, line 26: There appears to be a typo here.
```suggestion
	adminMux.HandleFunc("/admin/helloworld", func(writer http.ResponseWriter, request *http.Request) {
```

mbarnes (changes requested):
I think I caught a typo.  Otherwise LGTM.

mbarnes (approved):

SudoBrendan (approved):

PR #1788, opened 2025-05-21
add second azure monitor workspace for hosted control plane metrics

tony-schndr:
https://issues.redhat.com/browse/ARO-17156
https://issues.redhat.com/browse/ARO-17158

### What
Deploys a second Azure Monitor Workspace for storing hosted control plane metrics.  A second data collection rule is created to route hcp metrics to the hcp amw.  The remote write config for services amw drops all metrics coming from `ocm-<environment>*` namespaces, whereas the remote write configuration for hosted control planes amw only allows metrics from `ocm-<environment>*` namespaces.

### Why

For storing customer related metrics in a separate from aks/service metrics.


simonpasquier (approved):
lgtm

PR #1789, opened 2025-05-22
config: make use of environment value

stevekuznetsov:
This commit de-duplicates configuration values for staging environment and integration by using the new `{{ .ctx.environment }}` value. We are already getting into some really unfortunate outcomes like the name of the subscription in which the service and management clusters live - we have a environment-agonstic value at the top-level default, but then hard-code overwrite it with envrionment-specific values for staging. I imagine we will also have to overwrite it for production. I would love for us to simply have environment-sensitive names for *all* environments. This would promote simplicity and predictability.

This commit does *not* de-duplicate other values that are identical between these environments but should likely be. I would recommend setting global or per-cloud defaults for all of these values and only overriding (perhaps in int, specifically) where necessary. By forcing new additions to the config to be added at the top-level it will promote consistency from the outset and will make our configuration smaller over time.

### DIFF
--- a/config/config.msft.yaml
+++ b/config/config.msft.yaml
@@ -28,20 +28,25 @@ defaults:
   # The long Azure region name
   region: "{{ .ctx.region  }}"
   regionRG: "{{ .ctx.region }}-shared-resources"
+  # OIDC
+  oidcStorageAccountName: "arohcp{{ .ctx.environment }}oidc{{ .ctx.regionShort }}" # [globally-unique]
   global:
     rg: global-shared-resources
     subscription: hcp-global
     globalMSIName: "global-ev2-identity"
     safeDnsIntAppObjectId: "" # intentionally blank - only required in INT
     keyVault:
+      name: 'arohcp{{ .ctx.environment }}-global' # [globally-unique]
       private: false
       softDelete: true
     secretsToSyncDir: "none"
   # ACR
   acr:
     svc:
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
Config change failures ...

```
     "postgres": {
-      "name": "arohcpstg-cs-usw3",
+      "name": "arohcpstg-csdb-usw3",
```
How far along are we in staging environment? Can we just undo this mistake? We should really try not to have such things differ between environments.

```
     "logs": {
-      "san": "MGMT.GENEVA.KEYVAULT.ARO-HCP-STG.AZURE.COM"
+      "san": "MGMT.GENEVA.KEYVAULT.ARO-HCP-stg.AZURE.COM"
     },
```
Did we ever decide to put in that package of useful funcs for the template engine? We can use `| toUpper` if so. Is the SAN case-sensitive?

geoberle:
> How far along are we in staging environment? Can we just undo this mistake? We should really try not to have such things differ between environments.

we are about to rebuild it for a different reason so it is a good point in time for adjusting names

> Did we ever decide to put in that package of useful funcs for the template engine? We can use `| toUpper` if so. Is the SAN case-sensitive?

we could but would need to be careful not to use them on variables like the stamp, region or regionShort that will remain a variable when translated into EV2. for every variable that gets replaced with an actual value we could use sprig funcs etc. but that would leave is with `environment` for the time being

geoberle:
can you also adapt `mgmt.subscription` and `svc.subscription` ?

PR #1793, opened 2025-05-22
Grant NSP access from the Service Cluster subscription to the managem…

janboll:
### What

Grant NSP access from the Service Cluster subscription to the management NSP

### Why

Need to access Keyvaults from service clusters


dev-infrastructure/configurations/mgmt-nsp.tmpl.bicepparam
```diff
@@ -15,3 +15,5 @@ param aksKeyVaultName = '{{ .mgmt.aks.etcd.kvName }}'
 
 param mgmtNSPName = '{{ .mgmt.nsp.name }}'
 param mgmtNSPAccessMode = '{{ .mgmt.nsp.accessMode }}'
+
+param serviceClusterSubscriptionId = ''
```
geoberle, line 19: you need to use the __xxx__ syntax instead of an empty string.

https://github.com/Azure/ARO-HCP/blob/main/docs/bicep.md#output-templates-and-output-chaining

geoberle (dismissed):

geoberle (approved):

PR #1802, opened 2025-05-26
Add Logs Public Production Values

chiac831:
[ARO-15676](https://issues.redhat.com/browse/ARO-15676)

### What

Add Logs Staging Value

### Why

Deploy Staging Kusto

### DIFF
--- a/config/config.msft.yaml
+++ b/config/config.msft.yaml
@@ -372,11 +372,13 @@ clouds:
       # SVC cluster settings
       svc:
         logs:
-          san: "svc.geneva.keyvault.aro-hcp-{{ .ctx.environment }}.azure.com"
+          san: "svc.geneva.keyvault.aro-hcp.azure.com"
+          configVersion: "1.1"
       # MC cluster settings
       mgmt:
         logs:
-          san: "mgmt.geneva.keyvault.aro-hcp-{{ .ctx.environment }}.azure.com"
+          san: "mgmt.geneva.keyvault.aro-hcp.azure.com"
+          configVersion: "1.1"
       imageSync:
         componentSync:
           image:
@@ -387,6 +389,12 @@ clouds:
       mise:
         azureAdInstance: https://login.microsoftonline.com/
         armInstance: https://management.azure.com
+
+      # Logs
+      logs:
+        mdsd:
+          subscriptions:
+            - 9a53d80e-dae0-4c8a-af90-30575d253127  # Azure Red Hat OpenShift v4.x - HCP Global
     environments:
       int:
         # this is the MSFT INT environment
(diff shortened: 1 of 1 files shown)


config/config.msft.yaml
```diff
@@ -372,11 +372,13 @@ clouds:
       # SVC cluster settings
       svc:
         logs:
-          san: "svc.geneva.keyvault.aro-hcp-{{ .ctx.environment }}.azure.com"
+          san: "svc.geneva.keyvault.aro-hcp.azure.com"
```
weinong, line 375: this can be RA?

stevekuznetsov: Do we need a different SAN per environment, per cloud, or per region? If so, could we use the `{{ .ctx.<thing> }}` approach?

stevekuznetsov: Hm yeah, looks like you have an `-int` suffix. I know we do this a lot in classic where int (and sometimes stage) have special names, and prod has a non-special name ... but that is hard(er) to grok. Instead of sometimes having a suffix and sometimes not having the suffix, _always_ having the suffix just makes this config simpler, smaller, and more predictable.

geoberle: let's have a look at this.
currently we achieve RA configuration by leveraging placeholders from `.ctx`. all these values are passed in on configuration instantiation

`ctx.region`
`ctx.regionShort`
`ctx.stamp`
`ctx.cloud`
`ctx.environment`

especially the `ctx.region`, `ctx.regionShort` and `ctx.environment` are very useful to achieve region (and even environment) agnosticity. if you have a look at the environment level overrides for int and stage, you will see that it is revolving mostly about things that are truly environment specific (ip service tags, base domain, digests and versions, appIDs).

i'm open to add more `ctx.xxx` variables as we see fit. i'm not sure though it is needed to the log SANS, as we have a RA configuration for it already, namely `svc.geneva.keyvault.aro-hcp-{{ .ctx.environment }}.azure.com` 

coming back to the log SAN. what are the requirements for the value? is a respective domain registration in onecert required? could we pull off `svc.geneva.keyvault.aro-hcp-prod.azure.com` or `svc.geneva.keyvault.prod.aro-hcp.azure.com` (or any flavor of it that uses ctx.environment) for prod?

stevekuznetsov: Chatted with Gerd async, realizing that even if we keep `san: "svc.geneva.keyvault.aro-hcp-{{ .ctx.environment }}.azure.com"` we still end up in a pickle for int with `azure-test.net` and FairFax with `azure.us`. So ideally some small environment-specific context could be injected to allow for `san: "svc.geneva.keyvault.aro-hcp-{{ .ctx.environment }}.{{ .cfg.domainSuffix }}"` to be the DRY value at the top-level. Of course if we have that, we don't necessarily need the `{{ .ctx.environment }}` value in the SAN (but are free to use it elsewhere).

PR #1805, opened 2025-05-27
Devops msi output chaining

stevekuznetsov:
What
the ID of the global MSI used for deployment scripts and helm deployments is statically defined in the configuration right now. since this identity is also created by our bicep templates (global-infra.bicep) based on the configuration und global, this is an annoying two-pass setup in an environment:

run global pipeline
update config again with the ID from the freshly created global MSI
to prevent this, we will not store the ID in the config anymore but use output chaining for the lookup instead.

@whober0521 this is a draft and i'm not entirely sure if this is even feasible to be used for shell extension in its current form.

Why
Special notes for your reviewer

### DIFF
--- a/dev-infrastructure/scripts/service-image-backfill.sh.bak
+++ b/dev-infrastructure/scripts/service-image-backfill.sh.bak
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+SCRIPT_DIR=$(dirname "$(realpath "${BASH_SOURCE[0]}")")
+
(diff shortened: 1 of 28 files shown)


dev-infrastructure/scripts/service-image-backfill.sh.bak
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
stevekuznetsov, line 1: @geoberle did you mean to check this in?

PR #1807, opened 2025-05-28
add additional config schema validations

geoberle:
### What

* name lengths for azure resources
* character restrictions

### Why

it's better to check certain restrictions as early as possible in the process

PR #1809, opened 2025-05-28
env agnostic configuration

geoberle:
the config.msft.yaml configuration started using `ctx.environment` a lot for environment agnostic configuration. instead of having env overrides to add `int` or `stage` pre/in/suffixes to names, the names are constructed based on `ctx.environment` and are therefore environment agnostic. this enables us have a lot of configuration in the top level `defaults` section instead of leveraging the `cloud > env` defaults.

`ctx.environment` was just recently added to the ARO HCP tooling and reflects the name of an environment under `clouds.public.environments.xxx`.

example: `arohcp{{ .ctx.environment }}-global` becomes `arohcpint-global` in the MSFT INT environment or `arohcpdev-global` in the integrated DEV environment

to be able to use the environment agnostic (EA) configuration in more places in DEV, we need to adapt some of our environment names because these names will be used in places where
* the characterset is limited (e.g. no hyphens)
* lenght limitations apply

to make the name change safe with upfront validation, #1807 introduced more validation to the config schema in regards to naming patterns and lengths.

this PR  prepares the env names to be used in EA configurations for DEV. while env names are changes, this will not have any effect on existing envs. we should strive for env names that have no hypens and numbers in them, as well as being limited to 5 chars.

proposal:

* `personal-dev` -> `pers`
* `personal-perf` -> `perf`
* `cs-pr` -> `cspr`
* `nightly` -> `nitey` - length reasons, at least sounds like the original one. i'm open for suggestions

this PR also starts to use `southwestus` as region for config materialization, because it gives us a short region name that is 5 chars long and ends in a number. this enables better testing the new limits and restrictions early on.


config/Makefile
```diff
@@ -1,16 +1,24 @@
...
 	@echo "Materializing public cloud int env of config/config.yaml"
-	@../templatize.sh dev > public-cloud-dev.json
-	@echo "Materializing public cloud cs-pr env of config/config.yaml"
-	@../templatize.sh cs-pr > public-cloud-cs-pr.json
-	@echo "Materializing nightly env of config/config.yaml"
-	@../templatize.sh nightly > public-cloud-nightly.json
+	@REGION=$(REGION) ../templatize.sh dev > public-cloud-dev.json
+	@echo "Materializing public cloud cspr env of config/config.yaml"
+	@REGION=$(REGION) ../templatize.sh cspr > public-cloud-cspr.json
+	@echo "Materializing nitey env of config/config.yaml"
```
stevekuznetsov, line 14:
```suggestion
	@echo "Materializing nightly env of config/config.yaml"
```

stevekuznetsov: (and elsewhere)

config/Makefile
```diff
@@ -1,16 +1,24 @@
+# this is a region used for materializing configuration
```
stevekuznetsov, line 1: n.b. if we want to add another CI job that iterates through every possible value of (cloud,env,region) and validates the output for them, I think we could do that - the Go tool I wrote to do this for classic runs in milliseconds. We don't have to check in the output anywhere.

config/config.yaml
```diff
@@ -204,7 +204,7 @@ defaults:
     image:
```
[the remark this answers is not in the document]
geoberle: cleaned up

config/Makefile
stevekuznetsov: Seemed like a typo - if we are going for shorter, let's try `ntly` or something

geoberle: thanks for the name suggestion

janboll (approved):

PR #1814, opened 2025-05-29
ARO HCP E2E Setup

mbukatov:
### What

This PR introduces E2E Test cluster setup and helper scripts for development and testing purposes.

Implements ARO-15239

### Why

The setup will help:

- ARO HCP sub teams to create ARO HCP clusters for development or testing purposes
- ARO HCP QE sub team as a test setup for ARO HCP E2E Test suite runs
- ARO HCP QE, Doc and other sub teams to define or learn about existing supported cluster configurations

- Setup step code is based on the demo scripts.
- Eventually the idea is to keep a set of predefined cluster configuration we test with, for testing as well as demonstration purpose (so it will eventually replaces the demo scripts).
- See the README file for design decisions and reasoning.
- Updates to make this work in INT are not part of the PR, but will be included shortly (in this PR or as a follow up PR)


e2e-setup/setup.cluster-demo.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Run of Setup Steps scripts
+
+./10-infra-setup.sh
+./11-infra-mi-setup.sh
+export CLUSTER_TMPL_FILE=templates/cluster.default.json
+./20-cluster-create.arocurl.sh
+export NODEPOOL_TMPL_FILE=templates/nodepool.default.json
+export NODEPOOL_NAME=pool-one
+./30-nodepool-create.arocurl.sh
```
mgahagan73, line 13: We need some sort of wait/poll loop here where we need to ensure the cluster is in the "ready" state before making the nodepool... hitting this error:

`HTTP/1.1 400 Bad Request
X-Ms-Client-Request-Id: C1475200-B927-4DD7-AEE5-DA763F5326C7
X-Ms-Error-Code: InvalidRequestContent
X-Ms-Request-Id: 987159ea-c34c-4dcf-b21a-f69ffd5eb3ce
Date: Fri, 30 May 2025 15:25:04 GMT
Content-Length: 191

{
    "error": {
        "message": "Node pools can only be created on clusters in 'ready' state, cluster requested is in 'validating' state."
    }
`

mgahagan73 (commented):
I'm currently testing this out, so far the  main issue I'm seeing is in node pool create before cluster is ready.

e2e-setup/aro-curl-wait.sh
```diff
@@ -0,0 +1,75 @@
...
+while true; do
+  ./arocurl.sh GET "${AZURE_PATH}" > "${TMP_FILE}"
+  TS=$(date +%s)
+  TS_STR=$(date -u +"%Y-%m-%dT%H:%M:%S+00:00" --date="@${TS}")
+  STATE=$(jq --raw-output "${STATUS_QUERY}" "${TMP_FILE}")
+  if [[ "${STATE}" = "${TARGET_STATE}" ]]; then
+    echo "${TS_STR} $AZURE_PATH reached ${TARGET_STATE}"
+    break
+  fi
+  if [[ "${TS}" -gt $(("${TS_START}" + "${MAX_TIME}")) ]]; then
```
mgahagan73, line 66: I wouldn't hold up review for this but we might want to mention in the README for other Mac users that MacOS ships an ancient version of Bash from 2007 (3.2.57) that blows up on line 66 (Fedora 42 works fine)

e2e-setup/aro-curl-wait.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# Create a temporary file for cluster status
+TMP_FILE=$(mktemp)
+
+TS_START=$(date +%s)
+
+while true; do
+  ./arocurl.sh GET "${AZURE_PATH}" > "${TMP_FILE}"
+  TS=$(date +%s)
+  TS_STR=$(date -u +"%Y-%m-%dT%H:%M:%S+00:00" --date="@${TS}")
```
mgahagan73, line 60: MacOS date doesn't support this, might just want to mention in the README for other users who might be trying this on MacOS.

e2e-setup/teardown.sh
```diff
@@ -0,0 +1,17 @@
...
+#  * We assume that no resources are created outside of customer resource
+#    group during E2E setup.
+#  * In CI runs, we assume that the E2E tests run tests for cluster and node
+#    pool removal, so that when this script is executed, the cluster is no
+#    longer present.
+#  * If that is not the case, the teardown script will still remove the cluster
+#    as any other resource in the resource group (unles this is done in dev
+#    enviroment).
+
+az group delete --name "${CUSTOMER_RG_NAME}" --location "${LOCATION}"
```
patriksuba, line 17: Location option for command `az group delete` does not exists.

patriksuba (commented):
I assumed that we wanted to keep test framework related work in the test directory. If the wider team thinks the new directory is needed, let's keep it.
Thanks, let me know your thoughts.

e2e-setup/aro-curl-wait.sh
mbukatov: Good point.

e2e-setup/setup.infra-only-demo.sh
```diff
@@ -0,0 +1,22 @@
...
+SETUP_FILENAME="${CLUSTER_NAME}.e2e-setup.json"
+
+# Describe the setup configuration in a json structure
+./aro-setup-metadata.sh - "${SETUP_FILENAME}" << EOF
+{
+  "e2e_setup": {
+    "name": "infra-only-demo",
+    "tags": ["customer-infra-only"]
+  },
+  "cluster": {},
```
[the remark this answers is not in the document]
mbukatov: This is possible, we can include it there and some test can reuse it. But I would not make it a hard requirement tests will rely on. Test should be able to come up with a better name(s) if it needs to create clusters on top of resources created during setup. From setup perspective, you still need to specify `CLUSTER_NAME` variable even if you don't create cluster, as this value is used in few naming convetions (eg. managed identities stuff or name of the setup file itself).

e2e-setup/.gitignore
```diff
@@ -0,0 +1,7 @@
+.*.swp
+*~
+env.*
```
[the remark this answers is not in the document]
mbukatov: when I add the defaults and example files (fixing the problem with these files missing), this will work as intended, as there won't be a reason to commit another env file

mbukatov:
Based on further feedback, I'm going to add the following changes to the PR:

- persist tag for customer resource group
- improve failure handling on multiple places (the setup should stop when a first error is encountered)

tony-schndr:
Just my personal opinion, I think bash is the wrong language to use for this.  These scripts are already fairly complicated and would be better off written in Golang where we could have better test coverage.  These scripts will be hard to maintain as we add more features and even harder debug when they break.

Bash complexity aside, I recently merged some bicep templates to create hosted control planes and node pools.  This script can only be used in environments with ARM, https://github.com/Azure/ARO-HCP/pull/1871

There was also a recent merge that cleans up the existing demo/ scripts so that az rest or curl is used depending on environment, might be helpful. https://github.com/Azure/ARO-HCP/pull/1868

stevekuznetsov:
The end-to-end tests are already a separate Go module. There is no dependence between modules unless you import something and make it happen. I don't see any reason your constraints imply something other than `.bicep` templates for infra and a test runner written in a maintainable language.

e2e-setup/setup.infra-only-demo.sh
mgahagan73, line 1: From my cluster lifecycle testing I've done so far, we need to write out the subnet info to the e2e-setup.json file otherwise this file does not provide enough information about the environment to be usable for creating a cluster. I'm working around this by naming the subnet "worker-subnet" which happens to be the default value apperantly.

PR #1816, opened 2025-05-30
Validation refactoring

mbarnes:
No Jira.

### What

Refactors some of the validation code to capture common patterns and move "complex" validation out of the versioned API directory since it's not version-specific.

### Why

I have a few more more cluster validations to add but wanted to get this out of the way first.

The last commit looks big, but it's basically just moving validation code and tests from `/internal/api/v20240610preview` to `/internal/api`.


venkateshsredhat (approved):
lgtm

PR #1817, opened 2025-06-02
support for multiple role definitions for operator managed identities…

anshulvermapatel:
[ARO-17371](https://issues.redhat.com/browse/ARO-17371)

### What

This is to add support for multiple role definitions for operator managed identities for ev2 configs. There were few changes which were added previously, however, having ev2 configs populating envs containing the role definition names were left to be fixed.

This PR allows to add another role definition for any operator by adding it separating by a comma(without giving any space after the comma) in the ev2 main [config.yaml](https://github.com/Azure/ARO-HCP/blob/main/config/config.yaml#L299-L317) file. 

This can be verified by making the local azure operators managed identities config. The same logic is used to actual deployment as well.
~~~
make local-azure-operators-managed-identities-config 
~~~


geoberle:
you could use the shell script you introduced to build an additional values.yaml file that you pass in to helm via `-f roles-values.yaml` 

such a file would be easier to validate

cluster-service/Makefile
```diff
@@ -1,10 +1,13 @@
 -include ../setup-env.mk
 
+export
```
miguelsorianod, line 3: This seems unnecessary

cluster-service/generate_helm_set_flags.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash 
+
+# The 'roleName' field in the ../config/config.yaml file is a comma separated list of role names.
+# This script separates them and generates `--set` flags for the `helm deploy` command.
+# Please note that since here, the role names are being separated by commas, the role name itself
+# cannot have a comma in it's name.
+
+genereate_set_flags() {
```
miguelsorianod, line 8: Typo in the name of the function. genereate_set_flags -> generate_set_flags

cluster-service/generate_helm_set_flags.sh
```diff
@@ -0,0 +1,46 @@
...
+    for ((i=0; i<${#ROLE_NAMES[@]}; i++)); do
+        ROLE_NAME="${ROLE_NAMES[$i]}"
+        HELM_SETS+=(--set "azureOperatorsMI.${1}.roleDefinitions[$i].name=${ROLE_NAME}") 
+        ID=$(az role definition list --name "${ROLE_NAME}" --query "[].name" -o tsv)
+        HELM_SETS+=(--set "azureOperatorsMI.${1}.roleDefinitions[$i].id=${ID}")
+    done
+    printf "%q " "${HELM_SETS[@]}"
+}
+
+OP_HELM_SET_FLAGS=()
```
miguelsorianod, line 20: I would do all of this only when explicitly called

cluster-service/generate_helm_set_flags.sh
anshulvermapatel: > I would also be explicit about it being helm related: generate_helm_set_flags

I have named the whole script to be this name, so I think the helm context can be gotten from the script name.

miguelsorianod: > I have named the whole script to be this name, so I think the helm context can be gotten from the script name.

That is true for the context *within* the file but not outside of it. My comment here is coupled to the https://github.com/Azure/ARO-HCP/pull/1817/files#r2125972658 where you would call the function by name (or another one) from another file (the Makefile)

cluster-service/Makefile
miguelsorianod: That should not be done that way as that would be very broad. If you want to export a particular set of variables then you should export them and only them. Exporting everything can cause unintended side effects and it is unnecessary.

cluster-service/generate_helm_set_flags.sh
anshulvermapatel: Okay, I can agree with the function name to add `helm` in it. But I will still ask why not do all these in the script itself and keep the make file clean. https://github.com/Azure/ARO-HCP/pull/1817/files#r2126794479

cluster-service/generate_helm_set_flags.sh
anshulvermapatel: Sorry, can you please elaborate?

We are already doing this only when 2 of the make targets(`deploy` and `local-azure-operators-managed-identities-config`) are executed and where this is required.

cluster-service/generate_helm_set_flags.sh
```diff
@@ -0,0 +1,46 @@
...
+# The 'roleName' field in the ../config/config.yaml file is a comma separated list of role names.
+# This script separates them and generates `--set` flags for the `helm deploy` command.
+# Please note that since here, the role names are being separated by commas, the role name itself
+# cannot have a comma in it's name.
+
+genereate_set_flags() {
+    HELM_SETS=()
+    IFS=',' read -a ROLE_NAMES <<< "${2}" 
+    for ((i=0; i<${#ROLE_NAMES[@]}; i++)); do
+        ROLE_NAME="${ROLE_NAMES[$i]}"
```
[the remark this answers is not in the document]
anshulvermapatel: Okay.

cluster-service/Makefile
miguelsorianod: > It does exports the variables, i.e make them available for any script running for any target.

I don't think it does. As I mentioned above, according to the documentation when you run export without variable names it prints their names. It does not mention anything about all of them being exported.

To make sure about that, I've created the following scripts:
```
msoriano@localhost:~/go/src/github.com/Azure/ARO-HCP/cluster-service (main)(ocm:?)$ cat myscript1.sh 
#!/bin/bash

echo script1
MYVAR=3
export
./myscript2.sh
echo endscript1

msoriano@localhost:~/go/src/github.com/Azure/ARO-HCP/cluster-service (main)(ocm:?)$ cat myscript2.sh 
#!/bin/bash

echo "script2"

echo "'$MYVAR'"
echo "end script 2"
```

If you run `script1.sh` you will observehow the output is `''`. This indicates that MYVAR has not been exported. You will also see how it will have printed your environment variables, including MYVAR.
If you change the `export` line in myscript1.sh to `export MYVAR` and run again `script1.sh` you will observe how now `'3'` is printed. This further confirms what I mentioned.

This means that if those variables are exported, it is because of something else or some other side effect. But `export`  without arguments does not export the environment.

Independently on that, as I mentioned it is not the best idea to export the whole environment. If that is needed I would redesign that part so you can include / export whatever we need. One option as you said would be explicitly exporting what we need, having the Makefile a bit longer could be acceptable. There could also be other approaches that can be explored.

cluster-service/generate_helm_set_flags.sh
miguelsorianod: The idea is for the file to only define functions. Those functions would be available to the consumers of this file. Including/sourcing this file would not generate any side / implicit effects, which can be harder to follow and reason about when issues arise.

cluster-service/generate_helm_set_flags.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash 
+
+# The 'roleName' field in the ../config/config.yaml file is a comma separated list of role names.
```
miguelsorianod, line 3: It is now called roleNames

cluster-service/generate_helm_set_flags.sh
anshulvermapatel: As discussed in our meet. We are going to keep the function as is.

stevekuznetsov:
Why are we storing comma-delimited lists in a string field, as opposed to using a real array in config?

geoberle:
We faced issues whenever we wanted to use arrays in conjunction with EV2. We would love to use arrays

PR #1830, opened 2025-06-03
change schedule start time

YiqinZhang:
### What

- Modified runbook schedules, adjusting their start times to avoid a race condition between them.
- Use ordered dictionary [ordered]@{} in the PowerShell script to maintain the parameters registration order


tony-schndr (approved):
Nice work!  Thanks @YiqinZhang

PR #1832, opened 2025-06-04
Final pass at cluster validation

mbarnes:
If there was a Jira for this, it's been closed already.  This is just filling some gaps.

### What

This fills the remaining gaps in cluster validation after examining Cluster Service's validation code in detail.  It doesn't replicate _everything_ Cluster Service checks, but cherry-picks things that are unlikely to ever change: no CIDR ranges can overlap, customer-provided resources cannot be in the managed resource group, etc.

I also reworked the managed identity checks after I realized we were comparing them case-sensitively, and resource IDs (at least the kinds we deal with) are always case-_insensitive_.  For example, if we got a cluster payload that included:

```
"properties": {
    "platform": {
        "operatorsAuthentication": {
            "userAssignedIdentities": {
                "controlPlaneOperators": {
                    "ingress": "/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/resourcegroups/dev-rg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/ingress"
                }
            }
        }
    }
},
"identity":
    "userAssignedIdentities": {
        "/SUBSCRIPTIONS/1D3378D3-5A3F-4712-85A1-2485495DFC4B/RESOURCEGROUPS/DEV-RG/PROVIDERS/MICROSOFT.MANAGEDIDENTITY/USERASSIGNEDIDENTITIES/INGRESS": {}
    }
}
```

We need to recognize that these two user-assigned identity references are the same resource and allow it.

Finally, being uncomfortable with hard-coding JSON paths in error messages for some validation checks, I wrote a test function to verify the JSON path against the`"json:"` struct tags of the cluster or node pool type definition.  So if a hard-coded path has an error, it will be caught in unit tests.

### Why

The RP's validation error messages are specific to the ARM API, as opposed to Cluster Service's error messages which are specific to the OCM API and may not make sense to an ARO-HCP user.  Also the RP can return multiple validation errors at once, making for a better user experience with fewer retries.

### DIFF
--- a/internal/api/hcpopenshiftcluster.go
+++ b/internal/api/hcpopenshiftcluster.go
@@ -16,9 +16,11 @@ package api
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"strings"
 
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	validator "github.com/go-playground/validator/v10"
 
 	"github.com/Azure/ARO-HCP/internal/api/arm"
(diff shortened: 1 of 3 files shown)


internal/api/hcpopenshiftcluster.go
```diff
@@ -100,8 +102,8 @@ type OperatorsAuthenticationProfile struct {
 // OpenShift operators using user-assigned managed identities.
 // Visibility for the entire struct is "read create".
 type UserAssignedIdentitiesProfile struct {
-	ControlPlaneOperators  map[string]string `json:"controlPlaneOperators,omitempty"  validate:"dive,resource_id=Microsoft.ManagedIdentity/userAssignedIdentities"`
-	DataPlaneOperators     map[string]string `json:"dataPlaneOperators,omitempty"     validate:"dive,resource_id=Microsoft.ManagedIdentity/userAssignedIdentities"`
+	ControlPlaneOperators  map[string]string `json:"controlPlaneOperators,omitempty"  validate:"dive,keys,required,endkeys,resource_id=Microsoft.ManagedIdentity/userAssignedIdentities"`
```
venkateshsredhat, line 105: Nice Addition ..

PR #1833, opened 2025-06-04
harmonize DEV and MSFT config

geoberle:
### What

harmonize configuration between config.yaml and config.msft.yaml
approach:
* pull as much as possible into the defaults section of config.yaml
* config.msft.yaml inherts the defaults section of config.yaml (see the changes in the Makefile how merging is done)
* use ctx.environment as much as possible to build up env agnostic naming
* restrict name lengths so that 3-digit stamps are still within valid naming lengths

this implies that several names will change in dev and msft envs, but that should be fine as we are rebuilding anyways

this is a preparational step that still keeps a config.msft.yaml copy in this repo. in a followup step we will move the public-cloud-overlay for msft into an ADO repo

review suggestions:
* focus on the materialized configs to get a feeling for the actual change
* don't review config.msft.yaml as it is mostly generated now
* ... review config.yaml and config.msft.public-cloud-overlay.yaml instead

https://issues.redhat.com/browse/ARO-17629


config/config.yaml
```diff
@@ -318,105 +410,101 @@ clouds:
...
       # Backplane API
       backplaneAPI:
         image:
-          registry: arohcpsvcdev.azurecr.io
-          repository: app-sre/backplane-api
           digest: sha256:822477832a73c7eab7fe27200994f10030f708f4a752f33ded3f8f8eaa0470f6
       # Frontend
       frontend:
+        cert:
+          name: frontend-cert-{{ .ctx.regionShort }}
```
stevekuznetsov, line 421: `issuer: Self` surely has to remain in `dev` cloud but stuff like this name seems like a good candidate for promotion to top-level default

config/config.yaml
```diff
@@ -318,105 +410,101 @@ clouds:
...
-        softDelete: false
-        private: false
-      mgmtKeyVault:
-        softDelete: false
-        private: false
+        softDelete: true
       # disable soft delete on etcd KVs in DEV
       svc:
         subscription: ARO Hosted Control Planes (EA Subscription 1)
+        rg: hcp-underlay-{{ .ctx.regionShort }}-svc
```
stevekuznetsov, line 449: Good candidate for promotion up to top-level default?

config/config.msft.public-cloud-overlay.yaml
```diff
@@ -0,0 +1,326 @@
...
+      # Logs
+      logs:
+        mdsd:
+          cert:
+            name: logs-mdsd
+            type: x-pem-file # GCS certificate file in PEM format
+            issuer: OneCertV2-PrivateCA
+      # SVC cluster settings
+      svc:
+        subscription: "hcp-{{ .ctx.environment }}-svc-{{ .ctx.region }}"
```
stevekuznetsov, line 64: Good candidate for promotion up to top-level default? Then let dev cloud override with the hard-coded name?

config/public-cloud-dev.json
```diff
@@ -92,7 +92,7 @@
     }
   },
   "cxKeyVault": {
-    "name": "arohcp-cx-dev-1",
+    "name": "ah-dev-cx-dev-999",
```
stevekuznetsov, line 95: Ah... is `ah-` `AroHcp-`? If so, almost seems like we can snip off that prefix since presumably everything in the subscription deploying this will be for ARO HCP - so every resource would have the prefix.

config/config.msft.public-cloud-overlay.yaml
stevekuznetsov: I'm thinking from the perspective of someone building out a sovereign cloud in MSFT env - ideally they don't need to copy this block (or any of the blocks in this overlay that are fairly generic but just-so-happen-to-be overridden by `dev` cloud.) So maybe the litmus test is - if this naming scheme would also be the one used in Fairfax or Mooncake, let's see if we can hoist it up to the top-level defaults and just let `dev` override.

config/config.yaml
geoberle: done

config/config.yaml
geoberle: done

config/config.yaml
```diff
@@ -318,105 +410,101 @@ clouds:
...
+          prometheusOperator:
+            image:
+              digest: a5bf4407cb83dc93d4e29ef680e0a4d621256e0f004822f53b2ff1c592bf2a82
+          prometheusSpec:
+            image:
+              digest: 2dcc22f4a8ea5c198e1c9eb6e7f04d127c55924da72e0f4334e659633185283c
+            shards: 1
       mgmt:
+        subscription: ARO Hosted Control Planes (EA Subscription 1)
+        rg: hcp-underlay-{{ .ctx.regionShort }}-mgmt-{{ .ctx.stamp }}
```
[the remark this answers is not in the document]
geoberle: done

config/config.yaml
```diff
@@ -318,105 +410,101 @@ clouds:
...
         aks:
+          name: "{{ .ctx.regionShort }}-mgmt-{{ .ctx.stamp }}"
```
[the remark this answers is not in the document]
geoberle: done

config/config.yaml
```diff
@@ -318,105 +410,101 @@ clouds:
...
       svc:
         subscription: ARO Hosted Control Planes (EA Subscription 1)
+        rg: hcp-underlay-{{ .ctx.regionShort }}-svc
+        nsp:
+          accessMode: 'Learning'
+        istio:
+          targetVersion: "asm-1-25"
+          versions: "asm-1-25"
         aks:
+          name: "{{ .ctx.regionShort }}-svc"
```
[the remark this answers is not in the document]
geoberle: done

config/config.schema.json
```diff
@@ -86,6 +86,11 @@
       },
       "additionalProperties": false
     },
+    "resourceGroupName": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 90
```
geoberle, line 92: add regexp pattern

PR #1847, opened 2025-06-06
tracing: add node pool attributes

simonpasquier:
Closes [ARO-15048](https://issues.redhat.com/browse/ARO-15048)

### What

This commit instruments the Clusters Service calls with traces and adds
the following span attributes when available:
* `cs.nodepool.id`
* `cs.nodepool.state`

### Why

To be on par with the Clusters Service traces which include the same attributes.

Instead of instrumenting the frontend and backend separately, I've decided to implement tracing at the CS client level. It reduces the risk of missed attributes.


internal/ocm/tracing.go
```diff
@@ -0,0 +1,231 @@
...
+
+	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/Azure/ARO-HCP/internal/tracing"
+)
+
+type clusterServiceClientWithTracing struct {
+	csc        ClusterServiceClientSpec
```
mbarnes, line 28: Consider embedding `ClusterServiceClientSpec` here so you don't have to do `csc.csc` everywhere.
```
type clusterServiceClientWithTracing struct {
        ClusterServiceClientSpec
        tracerName string
}
```

mbarnes (approved):
Appreciate the `clusterServiceClient` cleanup.  We really need a better name for the interface than `ClusterServiceClientSpec`.  I had a suggestion below but it's optional.

PR #1848, opened 2025-06-09
api: Location is required for tracked resources

mbarnes:
### What

It's right there in the [RPC](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/put-resource.md#request) but I somehow missed it until the INT environment came online.


demo/03-create-cluster.sh
```diff
@@ -205,12 +205,14 @@ main() {
   create_azure_managed_identities_for_cluster
 
   jq \
+    --arg location "$LOCATION" \
```
venkateshsredhat, line 208: all this while was it using default location?

venkateshsredhat (approved):

SudoBrendan (approved):

PR #1849, opened 2025-06-09
fix templatize e2e test after environment rename

tony-schndr:
### What

fixes templatize e2e by updating resource group / cluster name to match new names

### Why

Values are hard coded and recent redeploy broke the e2e tests.  Will follow up to make this dynamic.

### DIFF
--- a/tooling/templatize/internal/end2end/e2e_test.go
+++ b/tooling/templatize/internal/end2end/e2e_test.go
@@ -82,9 +82,9 @@ func TestE2EKubernetes(t *testing.T) {
 	tmpDir := t.TempDir()
 
 	e2eImpl := newE2E(tmpDir)
-	e2eImpl.AddStep(types.NewShellStep("test", "kubectl get namespaces").WithAKSCluster("dev-svc"), 0)
+	e2eImpl.AddStep(types.NewShellStep("test", "kubectl get namespaces").WithAKSCluster("dev-westus3-svc-1"), 0)
 
-	e2eImpl.SetConfig(config.Configuration{"defaults": config.Configuration{"rg": "hcp-underlay-dev-svc"}})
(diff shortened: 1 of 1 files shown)


tooling/templatize/internal/end2end/e2e_test.go
```diff
@@ -82,9 +82,9 @@ func TestE2EKubernetes(t *testing.T) {
 	tmpDir := t.TempDir()
 
 	e2eImpl := newE2E(tmpDir)
-	e2eImpl.AddStep(types.NewShellStep("test", "kubectl get namespaces").WithAKSCluster("dev-svc"), 0)
+	e2eImpl.AddStep(types.NewShellStep("test", "kubectl get namespaces").WithAKSCluster("dev-westus3-svc-1"), 0)
 
-	e2eImpl.SetConfig(config.Configuration{"defaults": config.Configuration{"rg": "hcp-underlay-dev-svc"}})
+	e2eImpl.SetConfig(config.Configuration{"defaults": config.Configuration{"rg": "hcp-underlay-dev-westus3-svc"}})
```
mbarnes, line 87: Looks like it should work.  Maybe want to define a location const in `e2e.go` though?  I see `westus3` hard-coded in a few places there too.

tony-schndr: I'd like to see if we can make this dynamic instead, ie use whatever is in `config/`.  It's annoying to have to change this if we decide to update the resource group / aks cluster naming schema or if we want to run this in another environment.

mbarnes (commented):
LGTM as a quick fix.

Reminder to update the resource group names in the [docs](https://github.com/Azure/ARO-HCP/blob/main/docs/integrated-dev.md#access-integrated-dev-environment), but it can be done separately.

mmazur (approved):

PR #1851, opened 2025-06-10
api: Add PtrOrNil helper function

mbarnes:
### What

`PtrOrNil(p)` returns a pointer to `p` or `nil` if `p` is an empty value as would be determined by the `"omitempty"` option in [json.Marshal](https://pkg.go.dev/encoding/json#Marshal).

This function is used when converting a "normalized" API model to a "versioned" API model which consists entirely of pointer fields. If a value in a normalized model is empty, it is omitted from the versioned model as it would be when marshaling to JSON.

### Why

This turns out to be important for correct handling of read-only fields when updating a resource.

`api.ValidateVisibility` is supposed to disregard `nil` input values from a request payload, but we were previously never passing `nil` input values.

When processing a resource update via PUT, the frontend handler:

1. Creates a normalized API model containing only default values.
2. Converts the normalized API model to a versioned API model.
3. Unmarshals the request payload onto the versioned API model.

Step 2 is where the problem occurs. Fields in the normalized model with no default value will have an empty value. Previously we were using `api.Ptr` in the conversion which meant all those empty values would become _pointers_ to empty values in the versioned model. Then `api.ValidateVisibility` would handle those pointers to empty values as though they were explicitly provided in the request payload.

We now use `api.PtrOrNil` during conversion so empty values in the normalized model become `nil` pointers in the versioned model, and `api.ValidateVisiblity` handles them correctly.

### DIFF
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -16,6 +16,7 @@ package api
 
 import (
 	"iter"
+	"reflect"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


tony-schndr (approved):
LGTM!

PR #1856, opened 2025-06-10
PKO: Add sync steps

janboll:
### What

Add steps to sync PKO images

### Why

In int this is the way to sync images. in Order to deploy we need the images.


jfchevrette (dismissed):
LGTM

PR #1861, opened 2025-06-10
only run ci-go workflow if changes were made to go-related files

jfchevrette:
### What

ci-go workflow should only run when changes were made to files which relate or may relate to go code

### Why

ci-go workflow can take upward of 10-15 minutes to run so we should not run it when not necessary


janboll (changes requested):
this requires splitting out yaml format. Since this happens during make all-tidy

PR #1863, opened 2025-06-10
test: Add support for the INT environment to E2E Test Suite client

patriksuba:
[ARO-17452](https://issues.redhat.com/browse/ARO-17452)

### What

Add support for different environments configurations.
- create a global variable with current endpoint based on testing environment
- Update README with more accurate description of testing environments.
- Expanded test configuration details in README

### Why

Add support for the integration environment to HcpOpenShiftClient.


test/util/environment/environments.go
```diff
@@ -0,0 +1,39 @@
...
+	Staging     Environment = "staging"
+	Production  Environment = "production"
+)
+
+var environmentUrl = map[Environment]string{
+	Development: "http://localhost:8443",
+	Integration: "https://centraluseuap.management.azure.com",
+	Staging:     "https://...",
+	Production:  "https://...",
+}
```
mbukatov, line 31: Do we expect other differences than url and if we ignore TLS (as we do in dev)? I'm asking because I have two concerns here:

- maintaining this definition of environments in the test code
- ability to change it from the outside (eg. in a CI job executing the tests)

I recall we discussed this before, but if there are no other changes in the client wrt environment, I would suggest to reconsider:

- pass information that one uses "dev" env. (with localhost:8443 as a default url while disabling TLS) as we do now
- ability to pass url of the env from the outside

That said, if we are actually able to fill the gaps in right now, and be sure that we don't need to update it later, I would be ok with the current design.

mbukatov (commented):
Let's recheck that our assumptions make sense before merging.

test/util/environment/environments.go
patriksuba: Development environment requires additional setup. It would be better to set default to environment with least effort to setup like staging. We can keep some default for now.

I don't expect the URLs to change often, if at all. That's why I think we should hardcode them instead of letting them be changed from the outside. If a URL does change, it can be updated directly in the code.

We can use global variable *endpointUrl* and *Environment.CompareUrl* method to check what environment is being used. This lets us skip certain tests based on the result of check. See setup of TLS in configuration.

test/e2e/README.md
```diff
@@ -1,25 +1,39 @@
...
-- In the integration environment, the client will connect to the ARM using a public IP but only within the MSFT corporate network.
-- For stage and production environments, the client will connect through ARM once they are set up.
-
-The client expects a subscription to be already registered. To assign the client to a specific subscription, set its ID in the environment variable **CUSTOMER_SUBSCRIPTION**. If not set, the default all-zero subscription will be used.
-
-The E2E Test Suite requires a JSON setup file with a deterministic structure. The suite expects the path to this JSON file in the **SETUP_FILEPATH** environment variable. This JSON file is created by the ARO HCP E2E Setup code and contains all the necessary values for testing.
-
+- Development: The client connects directly to the RP frontend via port forwarding to localhost.
+- Integration: The client connects to Azure Resource Manager (ARM) using a public IP from a specific subscription.
+- Staging & Production: The client connects directly to the environment's public ARM endpoint.
```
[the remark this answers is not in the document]
patriksuba: This part is not directly trying to discuss or explain any environment. It is a simple sentence with little meaning behind it. I would rather not change it.

mgahagan73 (approved):

PR #1827, opened 2025-06-03
backend: instrument the CosmosDB client

simonpasquier:
https://issues.redhat.com/browse/ARO-15050

### What

This change is similar to 3a1188726 which applied to the RP frontend.

PR #1865, opened 2025-06-10
backend: create root spans before lock

simonpasquier:
### What

This change ensures that the spans created by the CosmosDB client are always attached to a root span. Before that the withSubscription() function which wraps the processOperations() call would create a new unattached span.

Example of orphan spans:

image

With this commit:

image

image

### Why

The issue happened after #1827 merged.


simonpasquier:
@mbarnes for review

PR #1868, opened 2025-06-10
Cleanups and enhancements to demo scripts

mbarnes:
### What

Code cleanups for the scripts in `/demo`.

If the ARO-HCP INT subscription is default according to `az account`, the scripts will issue RP requests with `az rest` instead of `curl`.


tony-schndr (approved):
Nice work!

PR #1869, opened 2025-06-10
bicep lint/fmt should run in parallel to reduce run time

jfchevrette:
### What

Improve bicep linting and formatting with the following changes:
- Use `AZURE_BICEP_CHECK_VERSION=False` to skip version check on every run of `az bicep` (saves a few seconds overall)
- Remove empty lines in output so it is easier to read
- Run lint/fmt make targets in parallel to speed up the process (using detected CPU cores minus 2 to avoid overloading the system)

The changes should be backward compatible with wherever we already use the make targets

Performance comparison before and after the changes on a 12-core system:

```
BEFORE
make fmt  106.76s user 32.07s system 98% cpu 2:21.19 total
make lint  146.68s user 32.37s system 105% cpu 2:50.30 total

AFTER
make fmt  169.38s user 67.47s system 771% cpu 30.700 total
make lint  244.80s user 70.59s system 871% cpu 36.171 total
```

The ` Bicep Lint / validate_bicep` workflow step went from **10 minutes** to **5 minutes** (GitHub runners have 4 cores)

### DIFF
--- a/dev-infrastructure/Makefile
+++ b/dev-infrastructure/Makefile
@@ -11,6 +11,7 @@ PROMPT_TO_CONFIRM = "--confirm-with-what-if"
 endif
 
 PERSIST ?= false
+NPROC ?= $(shell command -v nproc > /dev/null 2>&1 && nproc || echo 4)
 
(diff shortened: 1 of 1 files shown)


tony-schndr (approved):
Nice enhancement!

PR #1870, opened 2025-06-10
bump INT/Stage frontend version

tony-schndr:
### What

Deploy the fix from https://github.com/Azure/ARO-HCP/pull/1851 to INT and Stage

### Why

So that creating/updating hcp/nodepools via bicep/arm is idempotent.

This was deployed and tested in INT.


mbarnes (dismissed):
Approved, but I lack merge power over `/config`.  :frowning_face:

PR #1871, opened 2025-06-10
add a demo script to deploy HCP/NP via bicep

tony-schndr:
### What

Adds a demo script and bicep templates to deploy HCP + Node pools to INT


demo/deploy-hcp-bicep.sh
```diff
@@ -0,0 +1,46 @@
...
+# Create an ARO HCP Cluster + Node pool using bicep.
+set -o errexit
+set -o nounset
+set -o pipefail
+
+source env_vars
+LOCATION=uksouth # The ONLY supported region for ARO-HCP in INT
+
+# This is the only supported subscription for creating INT hcp/nodepools
+az account set -n "ARO SRE Team - INT (EA Subscription 3)"
```
mbarnes, line 11: This could be an unexpected side-effect of running the script.  Instead of changing the default can we just use `--subscription` in all the `az` commands?

mbarnes (changes requested):
Looks good, just one request.

PR #1872, opened 2025-06-10
frontend: Add a trivial /location endpoint

mbarnes:
### What

This is strictly for development environments to help discover the frontend's Azure region when port forwarding with kubectl.

e.g. `LOCATION=$(curl http://localhost:8443/location)`

### Why

@geoberle asked me about it today and I couldn't think of a better alternative than an endpoint.


geoberle (approved):
thank you

PR #1877, opened 2025-06-11
dev-infrastructure: fix bicep lint warnings

jfchevrette:
### What

This fixes a bunch of bicep linting warnings
- `dev-infrastructure/modules/network/nsp-profile.bicep`: Change property name from string to identifier
- `dev-infrastructure/modules/rp-cosmos.bicep`: Refactored the container array to use a user-defined type, enforcing the object keys and moving the default from within the code into the actual definitions. The result is values are more obvious and easier to find
- `dev-infrastructure/templates/global-image-sync.bicep`: Remove unused imports

### Why

Clean code is happy code

```
WARNING: dev-infrastructure/modules/network/nsp-profile.bicep(25,54)
Warning prefer-unquoted-property-names: Property names that are valid identifiers should be declared without quotation marks and accessed using dot notation. [https://aka.ms/bicep/linter/prefer-unquoted-property-names]

INFO: dev-infrastructure/modules/rp-cosmos.bicep(97,24)
Info BCP187: The property "defaultTtl" does not exist in the resource or type definition, although it might still be valid. If this is a resource type definition inaccuracy, report it using https://aka.ms/bicep-type-issues. [https://aka.ms/bicep/core-diagnostics#BCP187]

INFO: dev-infrastructure/modules/rp-cosmos.bicep(113,21)
Info BCP187: The property "partitionKeyPaths" does not exist in the resource or type definition, although it might still be valid. If this is a resource type definition inaccuracy, report it using https://aka.ms/bicep-type-issues. [https://aka.ms/bicep/core-diagnostics#BCP187]
```

### DIFF
--- a/dev-infrastructure/modules/network/nsp-profile.bicep
+++ b/dev-infrastructure/modules/network/nsp-profile.bicep
@@ -22,7 +22,7 @@ param serviceTags array = []
 @description('Array of Subscription ids that will be allowd to access NSP')
 param subscriptions array = []
 
-var subscriptionObjects = [for s in subscriptions: { 'id': s }]
+var subscriptionObjects = [for s in subscriptions: { id: s }]
 
 resource nsp 'Microsoft.Network/networkSecurityPerimeters@2024-06-01-preview' existing = {
   name: nspName
--- a/dev-infrastructure/modules/rp-cosmos.bicep
+++ b/dev-infrastructure/modules/rp-cosmos.bicep
@@ -10,13 +10,20 @@ param userAssignedMIs array
 param private bool
 
 // Local Params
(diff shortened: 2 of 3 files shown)


stevekuznetsov:
@jfchevrette why don't these lint warnings fail CI?

jfchevrette:
> @jfchevrette why don't these lint warnings fail CI?

Very good question that I was going to try to address next once all warnings are cleared. I'd like to propose that CI fails on warnings. It already fails on errors but `bicep lint` exits 0 on warnings and has no flag to do otherwise. I'm thinking we'll have to log linting output and grep for warnings or something like that.

PR #1892, opened 2025-06-13
CODEOWNERS: fix up .github, .devcontainers directories

stevekuznetsov:
The first-party team should not be the only ones that can touch these.


stevekuznetsov:
Going to self-merge here - the CODEOWNERS semantic is not intuitive. When an entry is added for a path, it _overrides_ the other entries, instead of adding to them. I don't think the first party team intended to say they were the only ones that could ever approve changes to GitHub Actions, and even if they were, that's wrong - so this change is good to go.

PR #1894, opened 2025-06-13
cleanup old pipeline files

geoberle:
### What

* istio pipeline is part of the svc-pipeline.yaml
* acrpull pipeline is part of the svc-pipeline.yaml and mgmt-pipeline.yaml


stevekuznetsov:
Some linter is complaining - but maybe can we also add a CI check with this PR that ensures no pipeline files exist that are not in topology?

PR #1897, opened 2025-06-13
add helm charts for admi-api deploy

katherinelc321:
### What

Created helm charts to deploy the admin-api
Created pipelines to deploy in dev
ARO-17908

### Why

We need to deploy admin-api to dev to test it


admin/deploy/templates/admin.deployment.yaml
```diff
@@ -0,0 +1,27 @@
...
+    matchLabels:
+      app: aro-hcp-admin
+  template:
+    metadata:
+      labels:
+        app: aro-hcp-admin
+    spec:
+      containers:
+        - name: aro-hcp-admin
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
geoberle, line 20: please use digest based image references instead of tag based ones.
frontend/backend decided on using a `deployment.imageName` helm variable used in the deployment

```
image: '{{ .Values.deployment.imageName }}'
```

.. which is constructured in the Makefile as a digest based image reference

acm/deploy/helm/multicluster-engine-config/charts/policy/charts/grc/templates/grc-policy-addon-controller.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
   namespace: '{{ .Values.global.namespace  }}'
 spec:
   progressDeadlineSeconds: 600
-  replicas: {{ .Values.hubconfig.replicaCount }}
+  replicas: {? {.Values.hubconfig.replicaCount: ''} : ''}
```
geoberle, line 11: something went wront here. this looks like the yamlfmt did not finish it's job

PR #1898, opened 2025-06-13
tooling/templatize: add a command to render and validate config

stevekuznetsov:
templatize: use logger from context in pipeline validation

---

tooling/templatize: add a command to render and validate config

This commit introduces a new set of tooling to validate that changes to
the service configuration files result in expected outcomes when they
are rendered into a region. This is a natural extension of the
`materialize` target, but this commit does not remove that mechanism
yet. If we are happy with this approach, we may as well remove that
mechanism in the future.

This commit consists of two related commands:

  templatize configuration render

The `render` subcommand does what it says - spits out a service
configuration for a given region. This is roughly the same thing that
the existing `generate` subcommand does, but it uses the ARO-Tools
libraries to be specific and explicit - instead of being a generic
template runner.

  templatize configuration validate

The `validate` subcommand runs in one of two modes - consuming developer
settings, or production configuration. The distinction is simply in how
the command determines which combinations of clouds, environments and
regions it will be operating over.

For a given set of clouds, environments, and regions (deployment
contexts) the command will render the service configuration and record
a SHA256 digest of the output to a records file. If the digest is
different than what is checked in, the command will do some `git`
workflows to determine what is likely the reference version of the
service configuration that generated that previous digest, save any
current work in the local branch, check out that reference version of
the service configuration and render the regional configuration on the
fly, to give the user a clear diff between what they've got and what was
there previously. This lets us record and manage configuration for every
possible region in every possible cloud, without checking in gigabytes
of data to this repository.

The output directory of the `validate` subcommand may be set to some
real path in the repository to keep the outputs, but it defaults to a
temporary directory that gets cleaned up after the run. Setting this
directory to `config/rendered` and pointing the tool at the developer
settings file lets us get almost the same behavior as the `materialize`
target today.

---

### DIFF
--- a/.github/workflows/config-change-detection.yml
+++ b/.github/workflows/config-change-detection.yml
@@ -32,7 +32,7 @@ jobs:
           echo "##################################################################"
           echo "##                                                              ##"
           echo "##   If you intended to change the service configuration, run   ##"
-          echo "##       make -C config/ render                                 ##"
+          echo "##       make -C config/ materialize                            ##"
           echo "##   and check in the result.                                   ##"
           echo "##                                                              ##"
           echo "##################################################################"
--- a/config/Makefile
+++ b/config/Makefile
@@ -13,7 +13,7 @@ prepare-msft-config:
 	yq '. | .clouds = (load("config.msft.public-cloud-overlay.yaml") | .clouds)' config.yaml >> config.msft.yaml
 
 
-materialize: prepare-msft-config
+materialize: render
 	@echo "Materializing dev cloud int env of config/config.yaml"
 	@REGION=$(REGION) ../templatize.sh dev -x $(STAMP) > public-cloud-dev.json
 	@echo "Materializing dev cloud cspr env of config/config.yaml"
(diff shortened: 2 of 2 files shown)


tooling/templatize/cmd/configuration/render/options.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	rawCfg, err := resolver.GetRegionConfiguration(opts.Region)
+	if err != nil {
+		return fmt.Errorf("failed to get region config: %w", err)
+	}
+
+	cfg, ok := config.InterfaceToConfiguration(rawCfg)
+	if !ok {
+		return fmt.Errorf(" invalid configuration")
```
weinong, line 188: extra whitespace, heh

weinong (dismissed):
LGTM

config/config.schema.json
```diff
@@ -642,6 +648,181 @@
         "tracing"
       ]
     },
+    "ev2": {
```
geoberle, line 651: is this supposed to be part of this PR or commited by accident?

geoberle:
so can introduce a GH action that will show the diff between old and new materialized without the need to have the actual diff in a PR. neat!

geoberle (dismissed):

stevekuznetsov:
Not sure why we're requiring top-level CODEOWNERS again, sigh.

PR #1912, opened 2025-06-16
tooling: add a command to only validate changed pipelines

stevekuznetsov:
Instead of running the (kind of long) validation on every single pipline in every single possible region, we can do some git magic to try to deliver the minimal "did I break something?" button.

### DIFF
--- a/tooling/templatize/cmd/pipeline/validate/options.go
+++ b/tooling/templatize/cmd/pipeline/validate/options.go
@@ -17,7 +17,9 @@ package validate
 import (
 	"context"
 	"fmt"
+	"os/exec"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -136,6 +143,19 @@ func (opts *ValidationOptions) ValidatePipelineConfigReferences(ctx context.Cont
 	if err != nil {
 		return fmt.Errorf("failed to create logger: %w", err)
 	}
+	shouldHandleService := func(string) bool {
+		return true
```
jonathan34c, line 147: why is this always return ```true```?

geoberle (approved):

PR #1913, opened 2025-06-16
config: use Ev2 value for regional az count

stevekuznetsov:
One unfortunate outcome of the "pre-templated config should parse as YAML" decision is that we cannot have raw template directives in the config, like:

  azCount: {{ .ev2.availabilityZoneCount }}

If we did that, the template block gets interpreted as some YAML and fails. Instead, we need everything to be a string if it's going to be templated, and use the `int()` function in `.bicepparam`.

It's possible we can run the config through some kind of templating with dummy values during load to relax this constraint...


geoberle:
i can see that you fixed the problem described in the PR description as the configuration shows plain {{}}

LGTM once the remaining PR checks are resolved

config/rendered/dev/cspr/westus3.yaml
```diff
@@ -342,7 +343,7 @@ svc:
       kvName: ah-cspr-se-usw3-1
       kvSoftDelete: false
     infraAgentPool:
-      azCount: 1
+      azCount: 3
```
stevekuznetsov, line 346: Wasn't quite sure if this should actually be the number of available zones or just `1` always?

config/config.yaml
```diff
@@ -529,9 +532,9 @@ clouds:
             maxCount: 6
             vmSize: 'Standard_D4s_v3'
             osDiskSizeGB: 100
-            azCount: 3
+            azCount: {{ .ev2.availabilityZoneCount }}
           infraAgentPool:
-            azCount: 1
+            azCount: {{ .ev2.availabilityZoneCount }}
```
geoberle, line 537: i think for this we should remain at hardcoded `1` in DEV to save some CPU quota in busy regions

geoberle: or at least for the `pers` env

geoberle:
can you remove the comment from the top of config/Makefile? https://github.com/Azure/ARO-HCP/pull/1913/files#diff-b30b43ebe195c6a3c05d58e2e592329e106a238d5f00cde0e1d658807b21bfefL2-L8

geoberle (approved):

PR #1915, opened 2025-06-17
Update VMSize of nodepool

patilsuraj767:
[ARO-17958](https://issues.redhat.com/browse/ARO-17958)

### What

Reducing VM Size for ARO-HCP NodePools this would save costs and allow us to provision more of those resources hitting quota limits later.

Both `Standard_D2s_v3` (amd64) and `Standard_D2ps_v6` (arm64) have been tested and verified to ensure that all cluster operators run successfully on node pools created with these VM sizes.

### DIFF
--- a/cluster-service/deploy/templates/cloud-resources-config.configmap.yaml
+++ b/cluster-service/deploy/templates/cloud-resources-config.configmap.yaml
@@ -9,6 +9,23 @@ data:
   instance-types.yaml: |
     instance_types:
     # DEMO INSTANCE TYPES
+    - id: Standard_D2s_v3
+      name: Standard_D2s_v3 - General purpose
+      cloud_provider_id: azure
+      cpu_cores: 2
+      memory: 8589934592
+      category: general_purpose
+      size: d2s_v3
+      generic_name: standard-d2s_v3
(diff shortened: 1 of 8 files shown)


cluster-service/deploy/templates/cloud-resources-config.configmap.yaml
```diff
@@ -9,6 +9,24 @@ data:
...
     instance_types:
     # DEMO INSTANCE TYPES
+    - id: Standard_D2s_v3
+      name: Standard_D2s_v3 - General purpose
+      cloud_provider_id: azure
+      cpu_cores: 2
+      memory: 8589934592
+      category: general_purpose
+      size: d2s_v3
+      ccs_only: true
```
oriAdler, line 19: Is that attribute needed, as it's not added to the other type?

IIRC it stands for **customer cloud subscription** for validation in CS. I'm unsure if that applies to ARO HCP, as all ARO HCP clusters will be provisioned into the customers' accounts.

oriAdler (commented):
I left one minor comment. Other than that LGTM.

oriAdler (approved):

PR #1920, opened 2025-06-17
tooling: handle changed schemas during diff generation

stevekuznetsov:
The following tooling has an issue when the config *and* the schema have changed - as we will check out the previous version of the config, but not the schema, so we will fail to render the previous region-scoped configuration as the previous RA service config does not pass the current schema. This commit makes sure to check out the JSON schema at the previous commit as well, to make sure we can render a diff.


weinong (approved):
LGTM

PR #1923, opened 2025-06-18
test: E2E Test Suite Containerization

patriksuba:
[ARO-17552](https://issues.redhat.com/browse/ARO-17552)

### What

Containerization of E2E Test Suite.
- Building binary of E2E Test Suite
- Building container image
- Running container

### Why

Prepare container for running E2E Testing in CI.

### DIFF
--- a/.github/workflows/aro-hcp-e2e-tests-image.yaml
+++ b/.github/workflows/aro-hcp-e2e-tests-image.yaml
@@ -30,12 +30,6 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 1
-    - name: Podman login to ACR
-      uses: redhat-actions/podman-login@v1
-      with:
-        registry: ${{ env.ARO_HCP_E2ETESTS_IMAGE_REGISTRY }}
-        username: ${{ secrets.AZURE_CLIENT_ID }}
-        password: ${{ secrets.AZURE_CLIENT_SECRET }}
     - name: Install Azure CLI
       uses: Azure/ARO-HCP@main
     - name: Az CLI login
@@ -44,6 +38,18 @@ jobs:
         client-id: ${{ secrets.AZURE_CLIENT_ID }}
         tenant-id: ${{ secrets.AZURE_TENANT_ID }}
         subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+    - name: Get ACR refreshToken # Workaround for podman login to ACR
+      id: acr_token
+      run: |
+        REFRESH_TOKEN=$(az acr login --name ${{ env.ARO_HCP_E2ETESTS_IMAGE_REGISTRY }} --expose-token --output tsv --query refreshToken)
+        echo "::add-mask::$REFRESH_TOKEN"
+        echo "refresh_token=$REFRESH_TOKEN" >> $GITHUB_OUTPUT
+    - name: Podman login to ACR with refreshToken
+      uses: redhat-actions/podman-login@v1
+      with:
+        registry: ${{ env.ARO_HCP_E2ETESTS_IMAGE_REGISTRY }}
+        username: 00000000-0000-0000-0000-000000000000
+        password: ${{ steps.acr_token.outputs.refresh_token }}
     - name: Build e2e test image
       uses: redhat-actions/buildah-build@v2
       with:
--- a/.github/workflows/aro-hcp-e2e-tests-image.yaml
+++ b/.github/workflows/aro-hcp-e2e-tests-image.yaml
@@ -1,4 +1,4 @@
-name: Build and Push E2E Test Image to ACR
+name: Build and Push E2E Tests Image to ACR
 on:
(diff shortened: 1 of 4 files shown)


test/Containerfile.e2e
```diff
@@ -0,0 +1,16 @@
...
+COPY --from=ARO-HCP test ./test
+COPY --from=ARO-HCP internal ./internal
+RUN touch junit-report.xml
+RUN cd test/e2e && make e2etest
+
+FROM --platform=linux/amd64 mcr.microsoft.com/cbl-mariner/distroless/base:2.0-nonroot@sha256:a5d5928601847a40c81fa397df33136c285866c974bf91ed94d688390fcc33f3
+
+WORKDIR /
+ENV SETUP_FILEPATH=e2e-setup.json
+COPY --from=builder /app/test/e2e/$SETUP_FILEPATH $SETUP_FILEPATH
```
mbukatov, line 13: I would not include the setup file into a container, but would rather pass it via a volume (see docker/podman --volume).

This means that the container would assume that a file is present on given path, but it will be passed only when the test is executed later, not during a build.

test/e2e/Makefile
```diff
@@ -1,5 +1,32 @@
...
+	-f ../Containerfile.e2e \
+	-t $(ARO_HCP_E2ETESTS_IMAGE):$(COMMIT_SHA)
+.PHONY: image
+
+e2etest-run:
+	podman run --rm \
+	-e CUSTOMER_SUBSCRIPTION=$(CUSTOMER_SUBSCRIPTION) \
+	-e AZURE_TENANT_ID=$(AZURE_TENANT_ID) \
+	-e AZURE_CLIENT_ID=$(AZURE_CLIENT_ID) \
+	-e AZURE_CLIENT_SECRET=$(AZURE_CLIENT_SECRET) \
```
mbukatov, line 30: Here is the place we could pass the setup via a volume.

mbukatov (changes requested):
I would rather use podman/docker volumes to pass data to and from the container.

test/Containerfile.e2e
mgahagan73: It sounds like we need a check somewhere (probably out of scope of this PR) to make sure we have the appropriate e2e-setup file already exists.

test/e2e/Makefile
mgahagan73: How would we get the junit report file out of the e2etest: target here?

mbukatov: > How would we get the junit report file out of the e2etest: target here?
I suggested to use a volume for that as well.

patriksuba: My initial idea was to use `podman cp`. Copying is not in a scope of the PR. I would leave that functionality for a actual pipeline.
I will explore more about volumes. Sounds like what we need.

test/Containerfile.e2e
```diff
@@ -0,0 +1,16 @@
+FROM --platform=linux/amd64 mcr.microsoft.com/oss/go/microsoft/golang:1.24-fips-cbl-mariner2.0@sha256:374281fe968134a44bcc7bd4b9e2d55dda61dcaa942fee5dda28cd38d54619cd as builder
+
+WORKDIR /app
+COPY --from=ARO-HCP test ./test
+COPY --from=ARO-HCP internal ./internal
+RUN touch junit-report.xml
```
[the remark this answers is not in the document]
patriksuba: Distroless Linux require junit file created this way. 

Yes we are dropping {unixtime} from junit report file. The intention is to collect it as an artifact after each run.

mbukatov (dismissed):

mgahagan73 (dismissed):
Curent changes look good to me.

patriksuba:
Dropped commit with temporary testing change to run image pipeline on pull request change.
Now the pipeline will run when changes are made to main branch.
PR is ready to get merged.

mbukatov (approved):

.github/workflows/aro-hcp-e2e-tests-image.yaml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        fetch-depth: 1
+    - name: Install Azure CLI
+      uses: Azure/ARO-HCP@main
+    - name: Az CLI login
+      uses: azure/login@v2
+      with:
+        client-id: ${{ secrets.AZURE_CLIENT_ID }}
+        tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+        subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
```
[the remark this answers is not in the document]
tony-schndr: Mirroring requires additional infra setup/changes.  I think it would be easier to just tag it for quay.io and push.

1. Add quay credentials to a place this pipeline can access.
2. Update step to tag image for quay.io.
3. Auth and push to quay.io.

tony-schndr (approved):
lgtm

PR #1931, opened 2025-06-19
api spec update to support NodePool OSDisk Encryption With Customer-…

Nanyte25:
To enable support for NodePool OSDisk Encryption With Customer-Managed Keys, the API spec has been updated.


Nanyte25:
@mbarnes Can you review this PR and see where I might be going wrong generating the rest api specs thank you

mociarain:
@Nanyte25 I don't think we support development from a fork for this repo because of Azure permissions. Can you move this to a branch on the upstream?

PR #1932, opened 2025-06-19
Add authorizedIpRanges

janboll:
https://issues.redhat.com/browse/ARO-15671

### What

Draft spec for adding authorizedIpRanges


api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -195,6 +195,10 @@ model ApiProfile {
   /** The internet visibility of the OpenShift API server */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   visibility?: Visibility = "public";
+
+  /** The IP ranges authorized to access the Kubernetes API server. */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  authorizedIpRanges?: string[];
```
miguelsorianod, line 201: Given that this is to the kubernetes api server, should the attribute name be more specific? For example, would it limit access in the ingress, or other endpoints?

miguelsorianod, line 201: Do we know if Hypershift supports this in some way?

enxebre: It does not atm. This is how we originally had capture this functional req https://issues.redhat.com/browse/CNTRLPLANE-203

Are we saying this would be a list of CIDRs? Is this mutable? Is there an upper limit?
Let's flesh out all those details and capture in CNTRLPLANE that supersedes 203

mbarnes: My read of the [XCMSTRAT](https://issues.redhat.com/browse/XCMSTRAT-517) is this is only for the restricting API server access.

IMO, its placement in `ApiProfile` along with the doc comment makes the context clear enough.
```
{
    "properties": {
        "api" {
            "url": "https://...",
            "visibility": "Public",
            "authorizedIpRanges": [
                "192.168.1.0/24"
            ]
        }
    }
}
```

If we wanted something similar for ingress I imagine we'd add an `IngressProfile` model.

mbarnes:
Sorry for the conflicts, we had a few other PRs merge today.

mbarnes:
>@janboll Sorry for joining the party late. Can we rephrase to authorizedCIDRBlocks or allowedCIDRBlocks? The reason is that IpRanges can also mean the following format 1.1.1.1-2.2.2.2 which I do not think we would like to support at the very least can we make sure that it is documented that we only support CIDR notations [e.g. 0.0.0.0/N]. Is that correct?

It's a good point.  Would `authorizedCIDRs` (or `allowedCIDRs`) be ok?  Only because our network settings have, for example, `PodCIDR` instead of `PodCIDRBlock`, and I see the term "CIDRs" being used throughout Hypershift.

nimrodshn:
@mbarnes Both sound good to me.

mbarnes:
I'm going to close this since it's been superseded by https://github.com/Azure/ARO-HCP/pull/1978.
Having conversations for the same changes in two PRs is confusing.

PR #1939, opened 2025-06-20
Add the nodepool architecture property to Hcpcluster Nodepool 

venkateshsredhat:
https://issues.redhat.com/browse/ARO-17716 

### What

Support for ARM64 in HCP worker nodes for Azure Red Hat Openshift

### Why

XCMSTRAT-523  :  Supporting Node Pools with x64 and arm64 architectures

Draft PR to discuss and finalize the models/design .


mociarain (commented):
I don't think we need this. Based on the Azure VM design the choice of VM Size explicitly chooses the OS architecture:
"If no feature letter for a CPU is listed, the series uses Intel x86-64 CPUs. If the CPU is AMD, it will be listed as a. If the CPU is ARM based (Microsoft Cobalt or Ampere Altra), it will be listed as p. "
(Ref: point 3 [here](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/overview?tabs=breakdownseries%2Cgeneralsizelist%2Ccomputesizelist%2Cmemorysizelist%2Cstoragesizelist%2Cgpusizelist%2Cfpgasizelist%2Chpcsizelist) )

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -410,6 +410,9 @@ model NodePoolPlatformProfile {
    * - https://learn.microsoft.com/en-us/azure/virtual-machines/sizes */
   vmSize: string;
 
+  /** The Architecture for the  Nodepool */
+  arch?: ArchitectureType;
```
mbarnes, line 414: Optional fields should have a default value.
```suggestion
  arch?: ArchitectureType = "amd64";
```

PR #1940, opened 2025-06-20
api: Make networkSecurityGroupId optional again

mbarnes:
https://issues.redhat.com/browse/ARO-17688

### What

Note, this is both an ARM API change and a frontend change.

It basically just reverts https://github.com/Azure/ARO-HCP/commit/2384b7abc841fa9740b55a98c8cf4cbe69edcba5.

### Why

We're charging ahead with ARM APIs before all the features are ready.  YOLO!


mbarnes:
Holding for review by CS and BU.

mociarain (commented):
LGTM but I think we need input from CS

miguelsorianod:
This is unclear.
As of now, it is not clear whether we are going to make NSG mandatory or not. There are some pending talks to BU next week to determine this.

PR #1942, opened 2025-06-20
tooling/templatize: load settings file everywhere

stevekuznetsov:
While we can continue exposing the raw flags for setting cloud, environment, region, stamp, etc, it's most likely that users will prefer to use the settings file and specify just the specific environment they are interested in.

Now that the service configuration validator knows how to read and evaluate every setting, we can remove the old materialize target and JSON files.

We are getting close to removing the templatize.sh script entirely, which is great - that means that the CLI itself is easy to use and it's now portable to other projects that want it.


config/config.msft.yaml
```diff
@@ -112,8 +112,11 @@ defaults:
   # Monitoring
   monitoring:
     grafanaName: "arohcp-{{ .ctx.environment }}"
-    # Format: Multiline string using '>-' YAML block scalar
+    # Format:
```
stevekuznetsov, line 115: @geoberle @tony-schndr given that I did not edit the config myself in this PR, I'm wondering how this diff snuck in and didn't trip the CI?

tooling/templatize/cmd/rolloutoptions.go
```diff
@@ -37,33 +37,25 @@ func NewRolloutOptions(config config.Configuration) *RolloutOptions {
 	}
 }
 
-func EV2RolloutOptions() *RawRolloutOptions {
```
stevekuznetsov, line 40: This was unused.

weinong (dismissed):
LGTM

PR #1944, opened 2025-06-20
Add HcpOperatorIdentityRoleSet resource to API spec

tsatam:
[ARO-18086](https://issues.redhat.com/browse/ARO-18086) 

### What

Adds the `HcpOperatorIdentityRoleSet` resource to our API spec. This definition is based on a combination of the frontend API proposed in [XCMSTRAT-1095](https://issues.redhat.com/browse/XCMSTRAT-1095), as well as the similar-purpose `PlatformWorkloadIdentityRoleSet` resource defined in ARO Classic: https://learn.microsoft.com/en-us/rest/api/openshift/platform-workload-identity-role-set/get?view=rest-openshift-2024-08-12-preview&tabs=HTTP#platformworkloadidentityroleset 

### Why

Consumers require the ability to retrieve what the required identities and their expected role assignments for a given OpenShift version are, as they will be required to pre-create these identities and RAs before creating a cluster.


api/redhatopenshift/HcpCluster.Management/hcpOperatorIdentityRoles.tsp
```diff
@@ -0,0 +1,82 @@
...
+/** HcpOperatorIdentityRoles represents a location based representation of
+ * the required platform workload identities and their required roles for a given
+ * OpenShift version
+ */
+@locationResource
+model HcpOperatorIdentityRoleSet
+  is ProxyResource<HcpOperatorIdentityRoleSetProperties> {
+  ...ResourceNameParameter<
+    HcpOperatorIdentityRoleSet,
+    NamePattern = "^\\d+\\.\\d+"
```
mbarnes, line 19: Since this is a read-only resource do we _need_ to restrict the name pattern?

For installable OCP versions ([XCMSTRAT-613](https://issues.redhat.com/browse/XCMSTRAT-613)) was thinking of having a special resource named "default" in addition to the "X.Y.Z" resource names.  I don't know if we want to have something similar for this resource type, but maybe we shouldn't preclude it in the API spec?

mbarnes (commented):
Looking good to me.  Just had a thought about resource names below.

api/redhatopenshift/HcpCluster.Management/hcpOperatorIdentityRoles.tsp
```diff
@@ -0,0 +1,82 @@
...
+   * by Control Plane operators on a cluster. */
+  controlPlaneOperators: OperatorIdentityRoles[];
+
+  /** The role definitions required for the User-Assigned managed identities used
+   * by Data Plane operators on a cluster. */
+  dataPlaneOperators: OperatorIdentityRoles[];
+}
+
+/** Role definitions for a specific operator */
+model OperatorIdentityRoles {
```
machi1990, line 39: Is this missing min/max supported openshift version?

Or the info is not needed since the user will do something like `Give me the required identity roles for the X openshift version I am going to install?`
i.e
```
GET /subscriptions/<subscriptio	n-id>/locations/<location>/providers/Microsoft.RedHatOpenShift/hcpOperatorIdentityRoleSets/4.14.
```

? 

if so, how does the response to the query below look like 

```
GET /subscriptions/<subscriptio	n-id>/locations/<location>/providers/Microsoft.RedHatOpenShift/hcpOperatorIdentityRoleSets/4.14.
```
? or we won't be supporting that endpoint

machi1990: Actually, ignore what I said. The info I was asking is on the `.name` field of the returned response.

api/redhatopenshift/HcpCluster.Management/hcpOperatorIdentityRoles.tsp
```diff
@@ -0,0 +1,82 @@
...
+/** HcpOperatorIdentityRoles represents a location based representation of
+ * the required platform workload identities and their required roles for a given
+ * OpenShift version
+ */
+@locationResource
+model HcpOperatorIdentityRoleSet
+  is ProxyResource<HcpOperatorIdentityRoleSetProperties> {
+  ...ResourceNameParameter<
+    HcpOperatorIdentityRoleSet,
+    NamePattern = "^[a-zA-Z0-9-\\.]{3,24}$"
```
miguelsorianod, line 19: What does this name pattern designate?

tsatam: Validation for the "name" of each roleset. Our expectation is that these will generally be named either an OCP version (`X.Y` or `X.Y.Z`), or a pointer name (e.g. `default`). 

I'm just overriding Azure's default resource name validation regex here (`^[a-zA-Z0-9-]{3,24}$`), which ordinarily does not permit the use of `.` in resource names. This validation doesn't matter too much in practice, the only time we accept a name as user input is on the path param when `GET`ting an individual resource.

tsatam: "Roleset" is just the name of this resource type, indicating "the set of roles required for a given OCP version". The name is a carryover from ARO Classic's `PlatformWorkloadIdentityRoleSet` resource that serves a similar purpose: https://learn.microsoft.com/en-us/rest/api/openshift/platform-workload-identity-role-set/get?view=rest-openshift-2024-08-12-preview&tabs=HTTP#platformworkloadidentityroleset.

My expectation is that the RP will need to assemble the responses here, either using the CS API's query parameters to filter down to a specific OCP version for single-version get requests, or getting the required identities for each supported minor OCP version on for the list request. 

I'm open to changing the name of the resource type with consensus, if we have any better ideas, I just leaned on Classic's naming conventions here since I was familiar with it. I'm also open to changing the structure of this endpoint entirely to match the CS API if that makes sense to do and is straightforward for downstream usecases, again the current structure is inherited from Classic which just maintains static definitions of the required operators+roleassignments for each OCP version (e.g. https://github.com/Azure/ARO-RP/blob/master/hack/devtools/platform_workload_identity_role_sets.json).

machi1990 (dismissed):

mbarnes (changes requested):

mbarnes (dismissed):
LGTM now.  I'm gonna copy your resource name pattern for versions.

jboutaud:
lgtm

mbarnes (approved):

PR #1945, opened 2025-06-20
adding support for nodepool OSDisk encryption Customer Managed Key BYOK.

Nanyte25:

### DIFF
--- a/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
+++ b/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
@@ -457,6 +457,34 @@ model NodePoolAutoScaling {
   max?: int32;
 }
 
+/** Encryption support disabled by default */
+model Encryption {
+  /** Encryption support disabled by default */
+  infrastructureEncryption?: "enabled" | "disabled" | string = "disabled";
+
+  /** Encryption support disabled by default */
+  customerManagedKeyEncryption?: CustomerManagedKeyEncryption;
+}
+/** Enable customer Managed encryption */
+model CustomerManagedKeyEncryption {
+  /** Enable customer Managed encryption */
+  keyEncryptionKeyIdentity: KeyEncryptionKeyIdentity;
+
+  /** Enable customer Managed encryption */
+  keyEncryptionKeyUrl: string;
+}
+/** Enable Encryption Key Identity assignment */
+model KeyEncryptionKeyIdentity {
+  /** Enable Encryption Key Identity assignment */
+  identityType: "systemAssignedIdentity" | "userAssignedIdentity" | string;
+
+  /** user  Identity assignment */
+  userAssignedIdentityResourceId?: UserAssignedIdentityResourceId;
+
+  /** Federated ClientId Identity assignment */
+  federatedClientId?: string;
+}
+
 /*
  * =======================================
  * End NodePool resources
--- a/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
+++ b/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
@@ -428,6 +428,9 @@ model NodePoolPlatformProfile {
    * - https://learn.microsoft.com/en-us/azure/availability-zones/az-overview
    */
   availabilityZone?: string;
+
+  /** Disk encryption settings for the OS and data disks */
+  encryption?: Encryption;
 }
 
 /** Supported Azure storage account types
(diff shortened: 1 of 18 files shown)


api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -457,6 +457,34 @@ model NodePoolAutoScaling {
   max?: int32;
 }
 
+/** Encryption support disabled by default */
+model Encryption {
```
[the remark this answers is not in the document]
Nanyte25: @miguelsorianod This is just the propose restspec to support bring you own for OSDisk encryption, this might need a DDR?

Nanyte25:
> What is the difference with #1931?
@miguelsorianod 
This one should be closed out, it was created on my fork incorrectly.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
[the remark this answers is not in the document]
miguelsorianod: > @miguelsorianod Do we have support for this via the nodepool endpoint in cluster service?
At the moment of writing this, we do not.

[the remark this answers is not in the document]
miguelsorianod: > @miguelsorianod This is just the propose restspec to support bring you own for OSDisk encryption,
I understand, but even with that, my point was that the MR does not seem to include anything related to OS Disk encryption in the restspec. It defines encryption related types, but they are not used anywhere in the restspec at least as of now

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -456,7 +459,35 @@ model NodePoolAutoScaling {
   @minValue(0)
   max?: int32;
 }
+/** Disk encryption settings for the OS and data disks */
+model Encryption {
+  /** Infrastructure-level encryption for the disks. Defaults to "disabled". */
+  infrastructureEncryption?: "enabled" | "disabled" | string = "disabled";
```
mbarnes, line 465: That's interesting, I've not seen this style of defining string enums before.

Nanyte25: I think it works just to have disabled by default, but I can change it @mbarnes

mbarnes: No I like it, it's more compact than the one-off string union types we've been doing.

I'll have to see what the generated JSON and Go code looks like, but if it's equivalent I'm tempted to apply this style to other existing enums like API visibility and outbound type.

mbarnes:
> @mbarnes This is ready for review

Aside from my distraction with enum declarations, the models you've added look good at a technical level.  But I have to withhold approving until we have more input from CS and BU folks.  I don't yet know if these are the _correct_ models for the encryption feature.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mbarnes: Just to close this thread, I decided _not_ to convert existing enums because it means we lose doc strings for the enum values.  (AFAIK there's no way to have doc strings for inline enum values.)  For super obvious cases like what you have here ("enabled" | "disabled") I think the inline syntax is fine.  Other enum values are less obvious and I think will benefit from having doc strings.

mbarnes, line 465: Unrelated to the syntax but rather to the enum values themselves, can you please capitalize them?
I just fixed this up across the API in another PR: https://github.com/Azure/ARO-HCP/pull/1965
```suggestion
  infrastructureEncryption?: "Enabled" | "Disabled" | string = "Disabled";
```

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -456,7 +459,35 @@ model NodePoolAutoScaling {
...
+
+  /** Configuration for using a customer-managed key (CMK) for disk encryption. */
+  customerManagedKeyEncryption?: CustomerManagedKeyEncryption;
+}
 
+/** Configuration for customer-managed disk encryption using Azure Key Vault. */
+model CustomerManagedKeyEncryption {
+  /** Identity used to authenticate with the Azure Key Vault for CMK. */
+  keyEncryptionKeyIdentity: KeyEncryptionKeyIdentity;
```
miguelsorianod, line 474: As we mentioned through other channels, this does not seem to be the interface that configuration of OS Disks SSE disk encryption with customer managed keys seems to expect.
Looking at https://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-customer-managed-keys-portal, this seems to be done through disk encryption sets.

You can also observe how hypershift seems to be exposing it: https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/azure.go#L315

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mociarain: Done

mociarain:
Closing in favor of https://github.com/Azure/ARO-HCP/pull/2069

PR #1949, opened 2025-06-23
Frontend: Support HCP-Level Autoscaling for All NodePools

venkateshsredhat:
https://issues.redhat.com/browse/ARO-17695

### What
1P STUB Frontend: Support HCP-Level Autoscaling for All NodePools

### Why
https://issues.redhat.com/browse/XCMSTRAT-1037

This is just the initial draft and will be subjected to changes with reviews on the Spec changes PR : https://github.com/Azure/ARO-HCP/pull/1936


miguelsorianod:
@machi1990 are you familiar with this on CS side? back then you explored cluster level functionalities in a spreadsheet

mbarnes:
Note, if updates are allowed then there's some _deep_ questions about how async operations will work.  This PR is wiring up the internal API model, which is fine, but is not actually supporting the feature yet.

venkateshsredhat:
> Note, if updates are allowed then there's some [_deep_ questions](https://github.com/Azure/ARO-HCP/pull/1936#discussion_r2171834818) about how async operations will work. This PR is wiring up the internal API model, which is fine, but is not actually supporting the feature yet.

I thought its these features are still not implemented in ARO HCP !? so probably this PR will mostly be around touching the internal structs .

machi1990:
Correct. This feature is not yet implemented in CS for aro-hcp. 
There is already a design serving commercial offering (e.g rosa-hcp) that isn't wired end to end and will error out.
There is a CS epic card under the same STRAT to design this for aro-hcp with async consideration in mind etc.

PR #1950, opened 2025-06-23
update CS image to include preparation changes related to ARO-HCP OCP versions configuration

miguelsorianod:
Update CS image to include newest changes.

Includes preparation work of ARO-HCP OCP Versions configuration

The CS image SHA 3c603f59797f7602aa59f888a8830d414932c7d6590c8721e3cdde85a9d3cb31 corresponds to CS image with tag 69ed2ee


miguelsorianod:
I put the materialize results commit in another PR to show what is that I changed manually and what's been generated.

One of the reasons is because the files that are generated have changed and there are other shas that changed other than the ones I manually modified.

Let me know if you want it in the same commit.

janboll (approved):

PR #1952, opened 2025-06-23
MSFT configuration cutover to ADO

geoberle:
### What

the MSFT specific configuration needs to be stored within ADO sdp-pipelines/hcp.
this PR updates the documentation and removes the respective msft configuration files from ARO HCP


stevekuznetsov:
Did we also want to remove it from the `settings.yaml`?

PR #1955, opened 2025-06-23
api: Restore the ProxyProfile model

mbarnes:
[ARO-17739 - REST Specs API: Support Cluster-Wide Proxy](https://issues.redhat.com/browse/ARO-17739)

### What

This was removed in https://github.com/Azure/ARO-HCP/pull/1380.

Restoring the ARM API as it was with one exception:

-  The OCM API defines the `noProxy` field as
   > a comma-separated list of domains and CIDRs for which the proxy should not be used

   I changed this field from a string to a string array.  [This matches what AKS does](https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/create-or-update?view=rest-aks-2025-04-01&tabs=HTTP#managedclusterhttpproxyconfig).
We'll just join the array elements to make the comma-separated list that CS wants.


tsatam (dismissed):

mbarnes:
I failed to notice this feature is for **Public Preview** rather than **Private Preview**.  Not sure if we should hold or go ahead and add it early.   I did have this all wired up to Cluster Service before it was removed.

SudoBrendan:
Adding a Hold to this since it was moved into Public Preview API version, and we shouldn't expect CS to have implementation anytime soon so this field would remain dead for a while.

raelga:
@mbarnes Does this still apply?

mbarnes:
Closing since it's not clear if we're ever gonna need this and the effort required to bring it up to speed with the current RP architecture is not worth it anymore.

PR #1959, opened 2025-06-24
Spike implementation of image mirror step

janboll:


tooling/templatize/pkg/pipeline/image_mirror.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating script temp file %w", err)
+	}
+
+	err = os.Chmod(tmpFile.Name(), 0755)
+	if err != nil {
+		return fmt.Errorf("error make script temp file executable %w", err)
+	}
+
+	defer os.Remove(tmpFile.Name())
```
stevekuznetsov, line 37: prefer to defer a closure so you can log errors

tooling/templatize/pkg/pipeline/image_mirror.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/Azure/ARO-Tools/pkg/types"
+)
+
+func runImageMirrorStep(ctx context.Context, step *types.ImageMirrorStep, options *PipelineRunOptions, inputs map[string]Output, outputWriter io.Writer) error {
+	tmpFile, err := os.CreateTemp("", "")
+	if err != nil {
+		return fmt.Errorf("error creating script temp file %w", err)
+	}
+
+	err = os.Chmod(tmpFile.Name(), 0755)
```
stevekuznetsov, line 32: perfer to inline these in the `if` block to not leak `err` to outer scope (comment applies a couple times)

```go
if err := thing(); err != nil { }
```

tooling/templatize/pkg/pipeline/image_mirror.go
janboll: Added

stevekuznetsov (dismissed):

geoberle:
hmm. the failing job is about backend/frontend digests not being set in the the RH DEV config. we interpret a missing DIGEST as an indicator to let the deploy targets pick the image, e.g. the image that belongs to the commit sha that is  currently deployed.

the old on-demand sync script had some handling for this. when the script was running within a GH-action, a missing digest just meand "skip"

https://github.com/Azure/ARO-HCP/blob/9f676ac322a52076be29fa8e371f46a1433823ce/image-sync/on-demand/sync.sh#L10-L13

geoberle:
we could just skip the step in templatize if the digest is not given

PR #1961, opened 2025-06-24
add script + pipelines to delete region,svc,mgmt resource groups

tony-schndr:
### What

Adds region/svc/mgmt resource group delete pipelines

### Why

So we can cleanup resource groups in touchless environments (AME)


config/rendered/public/int/uksouth.yaml
```diff
@@ -395,6 +395,6 @@ svc:
       replicas: 2
       shards: 2
       version: ""
-  rg: hcp-underlay-int-uksouth-svc
+  rg: hcp-underlay-uksouth-svc
```
geoberle, line 398: temporary change to delete the RGs that still follow the old names in STG

geoberle (dismissed):

PR #1964, opened 2025-06-25
make RP image digests explicit in dev config

geoberle:
### What

in dev environments, the rp image digests were undefined and we had a deployment mechanism in place to lookup digests at deploytime. this was supposed to bring some dev-ease-of-life. but it also makes reasoning about the config harder and is problematic for processes that require proper configuration.

this PR proposed the following changes:
* RH config needs to mention the image digests and the `make deploy` targets as well as the GH action based delivery will rely on that
* there will be an additional target `make deploy.current` that will provide the current behaviour and can be used in personal dev environments

this implies that 1P needs to bump the digests in config at relevant points in time to make sure the changes that should go to integrated dev are actually promoted.

### DIFF
--- a/get-digest.sh
+++ b/get-digest.sh
@@ -1,38 +1,25 @@
 #!/bin/bash
 
-if [ "$#" -ne 2 ]
+if [ "$#" -ne 3 ]
 then
-    echo "Need ARO_HCP_IMAGE_ACR and REPOSITORY parameters"
+    echo "Usage: $0 <ARO_HCP_IMAGE_ACR> <REPOSITORY> <TAG>"
     exit 1
 fi
 
 aro_hcp_image_acr=${1}
 repository=${2}
+tag=${3}
 
-if [ -n "${IMAGE_DIGEST_OVERRIDE}" ];
+tags=$(az acr repository show-tags --orderby time_desc --n "${aro_hcp_image_acr}" --repository "${repository}" --detail)
+
+# find the digest for the specified tag
+suggested_digest=$(jq -r --arg TAG "${tag}" \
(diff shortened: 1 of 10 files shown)


get-digest.sh
```diff
@@ -1,38 +1,25 @@
...
 then
-    echo ${suggested_digest}
+    echo "${suggested_digest}"
     exit 0
 fi
 
-jq -r 'first(.[] | .digest)' $tags
\ No newline at end of file
+# if there is no digest for the specified tag, return the first digest from the list of non-dirty tags
+jq -r 'first(.[] | select(.name | endswith("-dirty") | not) | .digest)' <<< "${tags}"
```
geoberle, line 25: i'm not sure we should actually keep this
* non-RP devs will never hit this situation as they rely on the digests from the config
* RP devs will build and push the image as part of `make deploy.current` anyways

get-digest.sh
```diff
@@ -1,38 +1,25 @@
...
-    echo ${IMAGE_DIGEST}
-    exit 0
-fi
-
-
-tags=$(mktemp)
-trap "rm ${tags}" EXIT
-
-az acr repository show-tags --orderby time_desc --n ${aro_hcp_image_acr} --repository ${repository} --detail > $tags
+tags=$(az acr repository show-tags --orderby time_desc --n "${aro_hcp_image_acr}" --repository "${repository}" --detail)
```
stevekuznetsov, line 13: I would expect a "current" deploy to be paired with the actual build of the image, so instead of looking it up from a central ACR and hoping you win the race for "latest push" could we not just record what we built and pushed, when we did it?

PR #1965, opened 2025-06-25
Azure API corrections

mbarnes:
### What

Fixes a few minor issues I noticed in the TypeSpec models:
- Node pool taint and label fields were marked optional when they should be required.
- Make all of our enum values use "PascalCase" instead of "camelCase"

### Why

Our API is coming due soon and I don't want to be stuck with these flaws.


mbarnes:
> Out of curiosity is the PascalCase a MSFT/Typespec thing or just our convention?

It appears to be a MSFT convention.  I mentioned this in the commit comment:
```
In lieu of an explicit requirement, follow ARM as an example.
All built-in enum values I've seen in ARM models use "PascalCase".
```

PR #1966, opened 2025-06-25
Make the SubscriptionLock resilient to handler panics

deads2k:
This ensures the stop method is called when the handler ends or panics.


mbarnes (dismissed):
Nice, I like the lock interface.
I'd like to better understand the raciness you commented on.

mbarnes (approved):
Thanks.  Builds now and tests all pass.
I see no reason why this wouldn't pass CI so I'm gonna merge and :crossed_fingers:.

PR #1979, opened 2025-06-26
Add etcd encryptionKey REST Spec

mociarain:
https://issues.redhat.com/browse/XCMSTRAT-981

### What

This implements a basic structure for the ETCD encryption but a user. It is designed to allow a user to specific version of a customer managed key to use for encryption.

Note:
The key is not rotatable or patchable. 

Some context is in here: https://docs.google.com/document/d/1MD6KPGDRAgiuok93ut9lSG8QS5RmM-F_buWomj-5MaU/edit?tab=t.0

### DIFF
--- a/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
+++ b/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
@@ -72,6 +72,14 @@ model HcpOpenShiftClusterProperties {
   /** Configure cluter capabilities. */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   capabilities?: ClusterCapabilitiesProfile;
+
+  /** Configure etcd encryption KMS key.
+   * Your Microsoft Entra application used to create the cluster must be authorized to access this keyvault,
+   * e.g using the AzureCLI: `az keyvault set-policy -n $KEYVAULT_NAME --key-permissions decrypt encrypt --spn <YOUR APPLICATION CLIENT ID>`
+   *
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  etcdEncryptionKey: KmsKey;
 }
 
 /** Cluster capabilities configuration. */
@@ -290,6 +298,27 @@ scalar UserAssignedIdentityResourceId
       type: "Microsoft.ManagedIdentity/userAssignedIdentities",
     }
   ]>;
+
+/** A represention of a KeyVault Secret. */
+model KmsKey {
+  /** vaultName is the name of the keyvault that contains the secret */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  @maxLength(255)
+  @minLength(1)
+  vaultName: string;
+
+  /** name is the name of the keyvault key used for encrypt/decrypt */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  @maxLength(255)
+  @minLength(1)
+  name: string;
+
+  /** version contains the version of the key to use */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  @maxLength(255)
+  @minLength(1)
+  version: string;
+}
 
 /*
  * =======================================
--- a/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
+++ b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
@@ -1171,10 +1171,19 @@
             "read",
             "create"
           ]
+        },
+        "etcdEncryptionKey": {
+          "$ref": "#/definitions/KmsKey",
+          "description": "Configure etcd encryption KMS key.\nYour Microsoft Entra application used to create the cluster must be authorized to access this keyvault,\ne.g using the AzureCLI: `az keyvault set-policy -n $KEYVAULT_NAME --key-permissions decrypt encrypt --spn <YOUR APPLICATION CLIENT ID>`",
+          "x-ms-mutability": [
+            "read",
+            "create"
+          ]
         }
       },
       "required": [
-        "platform"
+        "platform",
+        "etcdEncryptionKey"
       ]
     },
     "HcpOpenShiftClusterPropertiesUpdate": {
@@ -1204,6 +1213,47 @@
         {
           "$ref": "#/definitions/Azure.ResourceManager.CommonTypes.TrackedResourceUpdate"
         }
+      ]
+    },
+    "KmsKey": {
+      "type": "object",
+      "description": "A represention of a KeyVault Secret.",
+      "properties": {
+        "vaultName": {
+          "type": "string",
+          "description": "vaultName is the name of the keyvault that contains the secret",
+          "minLength": 1,
+          "maxLength": 255,
+          "x-ms-mutability": [
+            "read",
+            "create"
+          ]
+        },
+        "name": {
+          "type": "string",
+          "description": "name is the name of the keyvault key used for encrypt/decrypt",
+          "minLength": 1,
+          "maxLength": 255,
+          "x-ms-mutability": [
+            "read",
+            "create"
+          ]
+        },
+        "version": {
+          "type": "string",
+          "description": "version contains the version of the key to use",
+          "minLength": 1,
+          "maxLength": 255,
+          "x-ms-mutability": [
+            "read",
+            "create"
+          ]
+        }
+      },
+      "required": [
+        "vaultName",
+        "name",
+        "version"
       ]
     },
     "Label": {
--- a/internal/api/v20240610preview/generated/models.go
+++ b/internal/api/v20240610preview/generated/models.go
@@ -159,6 +159,11 @@ type HcpOpenShiftClusterListResult struct {
 
 // HcpOpenShiftClusterProperties - HCP cluster properties
 type HcpOpenShiftClusterProperties struct {
+	// REQUIRED; Configure etcd encryption KMS key. Your Microsoft Entra application used to create the cluster must be authorized
+	// to access this keyvault, e.g using the AzureCLI: az keyvault set-policy -n
+	// $KEYVAULT_NAME --key-permissions decrypt encrypt --spn <YOUR APPLICATION CLIENT ID>
+	EtcdEncryptionKey *KmsKey
+
 	// REQUIRED; Azure platform configuration
 	Platform *PlatformProfile
 
@@ -212,6 +217,18 @@ type HcpOpenShiftClusterUpdate struct {
 
 	// READ-ONLY; The type of the resource. E.g. "Microsoft.Compute/virtualMachines" or "Microsoft.Storage/storageAccounts"
 	Type *string
+}
+
+// KmsKey - A represention of a KeyVault Secret.
+type KmsKey struct {
+	// REQUIRED; name is the name of the keyvault key used for encrypt/decrypt
+	Name *string
+
+	// REQUIRED; vaultName is the name of the keyvault that contains the secret
+	VaultName *string
+
+	// REQUIRED; version contains the version of the key to use
+	Version *string
 }
 
 // Label represents the Kubernetes label
--- a/internal/api/v20240610preview/generated/models_serde.go
+++ b/internal/api/v20240610preview/generated/models_serde.go
@@ -487,6 +487,7 @@ func (h HcpOpenShiftClusterProperties) MarshalJSON() ([]byte, error) {
 	populate(objectMap, "capabilities", h.Capabilities)
 	populate(objectMap, "console", h.Console)
 	populate(objectMap, "dns", h.DNS)
+	populate(objectMap, "etcdEncryptionKey", h.EtcdEncryptionKey)
 	populate(objectMap, "network", h.Network)
 	populate(objectMap, "platform", h.Platform)
 	populate(objectMap, "provisioningState", h.ProvisioningState)
@@ -514,6 +515,9 @@ func (h *HcpOpenShiftClusterProperties) UnmarshalJSON(data []byte) error {
 			delete(rawMsg, key)
 		case "dns":
 			err = unpopulate(val, "DNS", &h.DNS)
+			delete(rawMsg, key)
+		case "etcdEncryptionKey":
+			err = unpopulate(val, "EtcdEncryptionKey", &h.EtcdEncryptionKey)
 			delete(rawMsg, key)
 		case "network":
 			err = unpopulate(val, "Network", &h.Network)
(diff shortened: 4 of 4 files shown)


api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -72,6 +72,14 @@ model HcpOpenShiftClusterProperties {
...
   @visibility(Lifecycle.Read, Lifecycle.Create)
   capabilities?: ClusterCapabilitiesProfile;
+
+  /** Configure etcd encryption KMS key.
+   * Your Microsoft Entra application used to create the cluster must be authorized to access this keyvault,
+   * e.g using the AzureCLI: `az keyvault set-policy -n $KEYVAULT_NAME --key-permissions decrypt encrypt --spn <YOUR APPLICATION CLIENT ID>`
+   *
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  etcdEncryptionKey: KmsKey;
```
miguelsorianod, line 82: The attribute is defined in the properties section of the cluster resource API. In that section different configurations related to different concepts coexist (network, console, dns ...).

When in the future we introduce additional configuration related to etcd encryption, with the way it is designed right now it means that other attributes would need to be added at the same level, so we would end up with configuration related to etcd encryption with multiple attributes at that level, coexisting with other concepts. Aside from that, at some point we want to support etcd encryption with platform managed keys which means this attribute won't be needed.

I think we can try to make this a bit more organized by placing the etcd encryption related configuration in its own section. We should also design the api in a way that is prepared for expansion for the needs we foresee. Particularly:
* For customer managed keys, we will want to support rotation
* We will want to make customer managed keys optional and use platform managed keys instead

A proposal I can think of is:

```
etcdDataEncryption (or etcdEncryption?):
    encryptionKey (or activeEncryptionKey, or activeKey, or key):
```

This contains the etcd data  encryption in its own section. In the future if we need to add a "backup key" or similar related to rotation would be added there. If the section is not provided I guess it would signal that platform managed encryption would be used. The behavior of selection between customer and platform managed keys is semi-implicit. If or some reason in the future it turns out that platform managed keys have some sort of configuration, adding it would be non natural because we would have platform managed related configuration and customer managed related configuration mixed in the same section. If for some reason for platform/customer  encryption we want to support other encryption algorithms adding that would be non natural because of the same reason.

Another proposal:

```
etcdDataEncryption:
  managementMode (or keyManagementMode): managed | unmanaged (or customerManaged | platformManaged ?)
  unmanaged:
    encryptionType: kms
    kms:
      encryptionKey  (or activeEncryptionKey, or activeKey, or key):
        ...
```

This last approach gives space to:
* Define if you want your etcd data encryption with platform managed keys or customer managed keys. It also is explicit about it. It also gives room to potentially add configuration related to platform managed encryption
* Add other encryption types for customer managed keys
* It gives space to add additional keys in kms, potentially needed for rotation

I would like to hear views on this or if you think there are other better approaches. I personally like the explicit approach and that allows you to grow on several dimensions (different management modes, different encryption algorithms within the management mode, space for specific configuration for each of those dimensions and so on...) but maybe I am missing some disadvantages or people think otherwise.

Independently on that, we will need to take a decision on how to define mandatory / optional attributes at API level. The decision has been that initially we will only support etcd data encryption with customer managed keys using kms, but in the future we will want to offer etcd data encryption with platform managed keys. I can think of the following potential decisions and implications around this:
* If we decide to make the customer managed encryption keys with kms mandatory by default it means that that will be required by default forever, even when we implement platform managed keys, because changing that would break the API.
* If we decide to make platform managed encryption keys by default, it means that right now if the user does not provide the fields related to customer managed keys, it will error at RP/app level because it is not supported, but once  the platform managed approach is implemented that will start to work. This can be a bit confusing initially because if someone creates a cluster it will fail once the request is submitted if no customer managed keys are provided
* If we use the explicit approach, make specifying the desired encryption type mandatory, no defaults. This means however that consumers of the API will need to always specify it, even when we support platform managed keys, because adding a default later would be an api breaking change.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -291,6 +299,27 @@ scalar UserAssignedIdentityResourceId
...
+  @maxLength(255)
+  @minLength(1)
+  name: string;
+
+  /** version contains the version of the key to use */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  @maxLength(255)
+  @minLength(1)
+  version: string;
+}
```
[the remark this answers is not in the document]
mociarain: This point on optional is interesting and as of now unclear. I'm hoping we can chat it out later. @miguelsorianod gives a great overview of the concerns here: https://github.com/Azure/ARO-HCP/pull/1979#discussion_r2171149221.

The point on visibility is interesting. I did a little more reading and I went with something more explicit based on [this](https://typespec.io/docs/language-basics/visibility/#visibility-filters).

LMKWYT

mociarain:
We need to discuss with PMs about what we want to do with defaults but that aside let me know what you think of the new models (@miguelsorianod 🙌 for help with re-working them)

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -291,6 +295,64 @@ scalar UserAssignedIdentityResourceId
...
+  dataEncryption: EtcdDataEncryptionProfile;
+}
+
+/** The ETCD data encryption settings. */
+model EtcdDataEncryptionProfile {
+  /** Specify whether the key is customer or platform managed. */
+  managementMode: "customerManaged" | "platformManaged" | string;
+
+  /** Specify customer managed encryption key details. */
+  customerManagedConfig?: CustomerManagedConfig;
```
miguelsorianod, line 334: Maybe just customerManaged as the name instead of customerManagedConfig

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -291,6 +295,64 @@ scalar UserAssignedIdentityResourceId
...
+}
+
+/** Customer managed encryption key details. */
+model CustomerManagedConfig {
+  /** The encryption type used. */
+  encryptionType: "kms" | string;
```
miguelsorianod, line 340: Should this be marked as extensible enum?

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -291,6 +295,64 @@ scalar UserAssignedIdentityResourceId
...
+/** The ETCD settings and configuration options. */
+model EtcdProfile {
+  /** ETCD Data Encryption settings. */
+  dataEncryption: EtcdDataEncryptionProfile;
+}
+
+/** The ETCD data encryption settings. */
+model EtcdDataEncryptionProfile {
+  /** Specify whether the key is customer or platform managed. */
+  managementMode: "customerManaged" | "platformManaged" | string;
```
miguelsorianod, line 331: Should this be marked as extensible enum?

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -291,6 +295,64 @@ scalar UserAssignedIdentityResourceId
...
+/** A KMS Key that is immutable. */
+@withVisibilityFilter(#{ all: #[Lifecycle.Create, Lifecycle.Read] })
+model NonUpdatableKmsKey {
+  ...KmsKey;
+}
+
+/** The ETCD settings and configuration options. */
+model EtcdProfile {
+  /** ETCD Data Encryption settings. */
+  dataEncryption: EtcdDataEncryptionProfile;
```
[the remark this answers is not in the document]
miguelsorianod: We should also document that if not provided, then it defaults to platform managed keys

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -291,6 +295,64 @@ scalar UserAssignedIdentityResourceId
...
+  customerManagedConfig?: CustomerManagedConfig;
+}
+
+/** Customer managed encryption key details. */
+model CustomerManagedConfig {
+  /** The encryption type used. */
+  encryptionType: "kms" | string;
+
+  /** The kms encryption key details. */
+  kmsConfig: KmsConfig;
```
miguelsorianod, line 343: I think here the attribute name can just be kms

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mociarain: We've decided to go with: a more expressive and well defined API

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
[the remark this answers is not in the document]
mociarain: FWIW I like "customerManaged" | "platformManaged". They're never "unmanaged" it's more a function of who is responsible. 

I also prefer "keyManagementMode" and will make this change now

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mociarain: Switching it only generates a small change in the 'openapi.json' so I was planning to leave it simple until we need it to be more. 

I don't feel strongly about it.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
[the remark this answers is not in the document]
mociarain: Yes to both. Done

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
miguelsorianod: I think we need it so it is prepared for extension without being considered a breaking change in the uture

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
miguelsorianod: See https://github.com/Azure/ARO-HCP/pull/1979#discussion_r2175138559

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -291,6 +295,68 @@ scalar UserAssignedIdentityResourceId
...
+model EtcdProfile {
+  /** ETCD Data Encryption settings.
+   * If not specified platform managed keys are used.
+   */
+  dataEncryption?: EtcdDataEncryptionProfile;
+}
+
+/** The ETCD data encryption settings. */
+model EtcdDataEncryptionProfile {
+  /** Specify whether the key is customer or platform managed.
```
miguelsorianod, line 332: We don't need to say "customer or platform managed". We can just say specify the key management mode, and then describe what is the effect of each.  In that way it is not limited to just two

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -291,6 +295,64 @@ scalar UserAssignedIdentityResourceId
...
+}
+
+/** Configure etcd encryption KMS key.
+ * Your Microsoft Entra application used to create the cluster must be authorized to access this keyvault,
+ * e.g using the AzureCLI: `az keyvault set-policy -n $KEYVAULT_NAME --key-permissions decrypt encrypt --spn <YOUR APPLICATION CLIENT ID>`
+ *
+ */
+model KmsConfig {
+  /** The details of the active key. */
+  activeKey: NonUpdatableKmsKey;
```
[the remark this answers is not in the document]
mociarain: Done

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mociarain: Done

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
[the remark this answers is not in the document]
mociarain: This is now solved, and I went back on what I had done with the specific limited visibility model to just defining it on the attribute directly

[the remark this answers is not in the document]
mbarnes: > Also I think the @visibility tags are unnecessary in this model since you have @visibility(Lifecycle.Read, Lifecycle.Create) on the etcdEncryptionKey field.

Actually you were right to make it explicit everywhere.  I was wrong about how visibility works; I thought it cascaded into child models but it does not.  Sorry to mislead.

mbarnes:
Sorry for the conflicts, we had a few other PRs merge today.

jharrington22:
This LGTM, it differs from ROSA but in my opinion much cleaner implementation.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -293,6 +297,86 @@ scalar UserAssignedIdentityResourceId
...
+}
+
+/** Customer managed encryption key details. */
+model CustomerManagedConfig {
+  /** The encryption type used.
+   * By default, "kms" is used.
+   */
+  encryptionType?: CustomerManagedEncryptionType = CustomerManagedEncryptionType.Kms;
+
+  /** The kms encryption key details. */
```
miguelsorianod, line 346: We should add:

Required when encryptionType is "kms".

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -70,9 +70,13 @@ model HcpOpenShiftClusterProperties {
   @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   platform: PlatformProfile;
 
-  /** Configure cluter capabilities. */
+  /** Configure cluster capabilities. */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   capabilities?: ClusterCapabilitiesProfile;
+
+  /** Configure ETCD. */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
```
miguelsorianod, line 78: Does this propagate to subsections? What happens if in the future we want to allow updating something inside the etcd section or subsections? will adding update be easy or will it be considered breaking somehow?

miguelsorianod:
I left a couple of comments but once those are addressed it will lgtm

tsatam (commented):
LGTM outside of the suggestions already made

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -322,6 +326,94 @@ scalar UserAssignedIdentityResourceId
...
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  encryptionType?: CustomerManagedEncryptionType = CustomerManagedEncryptionType.Kms;
+
+  /** The Key Management Service (KMS) encryption key details.
+   *
+   * Required when encryptionType is "KMS".
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  kms?: KmsConfig;
```
mbarnes, line 384:
```suggestion
  kms?: KmsEncryptionProfile;
```

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -322,6 +326,94 @@ scalar UserAssignedIdentityResourceId
...
+}
+
+/** Configure etcd encryption Key Management Service (KMS) key.
+ * Your Microsoft Entra application used to create the cluster must be authorized to access this keyvault,
+ * e.g using the AzureCLI: `az keyvault set-policy -n $KEYVAULT_NAME --key-permissions decrypt encrypt --spn <YOUR APPLICATION CLIENT ID>`
+ *
+ */
+model KmsConfig {
```
mbarnes, line 392:
```suggestion
model KmsEncryptionProfile {
```

mbarnes (changes requested):
One more naming nit, sorry.

PR #1983, opened 2025-06-26
enable size tagging to define ClusterSizingConfiguration

mukrishn:
### What

Hypershift `ClusterSizingConfiguration` is required to size control plane components for workloads.
This is disabled by default and can be enabled by adding `--enable-size-tagging` at install.  
Details of CSC [here](https://hypershift-docs.netlify.app/how-to/azure/scheduler/#clustersizingconfiguration)

```
--enable-size-tagging                            If true, HyperShift will tag the HostedCluster with a size label corresponding to the number of worker nodes
```

### Why

This is needed to configure HCP size for large scale clusters.

The default CSC config looks like this,
```
apiVersion: scheduling.hypershift.openshift.io/v1alpha1
kind: ClusterSizingConfiguration
metadata:
  name: cluster
spec:
  concurrency:
    limit: 5
    slidingWindow: 0s
  sizes:
  - criteria:
      from: 0
      to: 10
    name: small
  - criteria:
      from: 11
      to: 100
    name: medium
  - criteria:
      from: 101
    name: large
  transitionDelay:
    decrease: 0s
    increase: 0s
```

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -92,7 +92,7 @@ defaults:
       registry: quay.io
       repository: acm-d/rhtap-hypershift-operator
     namespace: hypershift
-    additionalInstallArg: ''
+    additionalInstallArg: '--enable-size-tagging=true'
   # Log settings
   logs:
     mdsd:
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -92,7 +92,7 @@ defaults:
       registry: quay.io
       repository: acm-d/rhtap-hypershift-operator
     namespace: hypershift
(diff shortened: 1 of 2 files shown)


geoberle:
@mukrishn is this something we need in all environments?
if yes please add it to https://github.com/Azure/ARO-HCP/blob/82e50995a413f25e7c396eff7213188d450e309d/hypershiftoperator/deploy/templates/installer.job.yaml

mukrishn:
The default config(which I've attached in the description) does nothing, this additional arg only enables a CRD to let us config the HCP sizing.

PR #1997, opened 2025-06-27
add target in CS Makefile to generate aro-hcp versions config file for dev

miguelsorianod:
With the source being the helm file. Useful for local development of CS.

### DIFF
--- a/cluster-service/Makefile
+++ b/cluster-service/Makefile
@@ -113,7 +113,7 @@ local-azure-operators-managed-identities-config:
 	@source ./generate_helm_set_flags.sh && \
 	helm template deploy -s templates/azure-operators-managed-identities-config.configmap.yaml \
 	  "$${OP_HELM_SET_FLAGS[@]}" \
-	  | yq '.data."azure-operators-managed-identities-config.yaml"' > ./azure-operators-managed-identities-config.yaml
+	  | yq '.data["azure-operators-managed-identities-config.yaml"]' > ./azure-operators-managed-identities-config.yaml
 .PHONY: local-azure-operators-managed-identities-config
 
 local-aro-hcp-ocp-versions-config:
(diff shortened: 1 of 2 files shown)


cluster-service/Makefile
```diff
@@ -113,9 +113,13 @@ local-azure-operators-managed-identities-config:
 	@source ./generate_helm_set_flags.sh && \
 	helm template deploy -s templates/azure-operators-managed-identities-config.configmap.yaml \
 	  "$${OP_HELM_SET_FLAGS[@]}" \
-	  | yq '.data."azure-operators-managed-identities-config.yaml"' > ./azure-operators-managed-identities-config.yaml
+	  | yq '.data["azure-operators-managed-identities-config.yaml"]' > ./azure-operators-managed-identities-config.yaml
 .PHONY: local-azure-operators-managed-identities-config
 
+local-aro-hcp-ocp-versions-config:
+	helm template deploy -s templates/aro-hcp-ocp-versions-config.configmap.yaml | yq '.data["aro-hcp-ocp-versions-config.yaml"]' > ./aro-hcp-ocp-versions-config.yaml
```
p0lyn0mial, line 120: matches https://github.com/Azure/ARO-HCP/blob/77f36d10dde7cf4c7870e06ae28d54aac394408f/cluster-service/deploy/templates/aro-hcp-ocp-versions-config.configmap.yaml#L7

miguelsorianod: Oh that's correct.

The reasons I added it using helm is because the same it is done for the MIs requirements config file target. And that target uses helm because the config file is a helm config file (same as this one), and in that case it has references to helm variables. For this specific config we currently do not have any helm variable so as you pointed out there's no need to use the helm command line. It wouldn't surprise me if in the future some is added so I am fine keeping using the helm command line although not strictly needed right now.

p0lyn0mial: sgtm

geoberle (approved):

PR #2008, opened 2025-06-28
Fix ARM deployment names conflicts for outputOnly steps

geoberle:
### What

output only steps are used to lookup complex values cross subscription. since they don't modify anything, we can run multiple of them at the same time without running into issues. to make sure we don't run into deployment name conflicts while doing so, output only steps will get a randomized name suffix added to the azure deployment name.

### DIFF
--- a/tooling/templatize/pkg/pipeline/arm.go
+++ b/tooling/templatize/pkg/pipeline/arm.go
@@ -16,6 +16,8 @@ package pipeline
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


mmazur (approved):

tooling/templatize/pkg/pipeline/arm.go
```diff
@@ -285,19 +304,19 @@ func doWaitForDeployment(ctx context.Context, client *armresources.DeploymentsCl
 	if step.DeploymentLevel == "Subscription" {
 		// Hardcode until schema is adapted
```
stevekuznetsov, line 305: @geoberle what needs to be fixed here? I can do it if you let me know the intent

PR #2042, opened 2025-06-30
secret-sync-controller helm chart

geoberle:
### What

helm chart and CI integration for secret-sync-controller

https://issues.redhat.com/browse/ARO-17118


.yamlfmt.yaml
```diff
@@ -2,6 +2,7 @@ continue_on_error: true
 match_type: doublestar
 exclude:
 - '**/zz_fixture_*.yaml'
+- 'secret-sync-controller/deploy/templates/*'
```
stevekuznetsov, line 5: @geoberle  I am surprised you had to do this - `yamlfmt` will just ignore any file that isn't valid YAML (e.g. has template braces in there)

yamlfmt.wrap.sh
```diff
@@ -3,15 +3,45 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-set -o xtrace
 
 source "hack/utils.sh"
 
-while IFS= read -r file; do
+# Function to check if a file matches any ignore pattern
+should_ignore_file() {
```
stevekuznetsov, line 10: We're crossing the "shouldn't be in shell" bar here IMO - what was the behavior that necessitated this? If the wrap/unwrap is correctly written it should fully undo itself, so it shouldn't really matter if it operates over files that the linters are not touching.

.yamlfmt.yaml
geoberle: forget about it. it was a stupid attempt. fixed by adapting the unwrap script. take a look there

yamlfmt.wrap.sh
geoberle: i reverted this change, so wrap script is untouched

yamlfmt.unwrap.sh
```diff
@@ -9,9 +9,9 @@ source hack/utils.sh
 
 while IFS= read -r file; do
   # Step 1: Unwrap quotes from specific fields
-  os::util::sed -E "s/([pP]ort|replicas|enabled|azCount): '\{\{\s*([^}]+)\s*\}\}'$/\1: {{ \2 }}/g" "$file"
+  os::util::sed -E "s/([pP]ort|replicas|enabled|azCount|expression): '\{\{\s*([^}]+)\s*\}\}'$/\1: {{ \2 }}/g" "$file"
```
stevekuznetsov, line 12: Makes sense, it's the smallest sin, of the ones we have available to us:

1. ignore the directory entirely
2. make some uber-complex loading & parsing thing for files that are not YAML but may become YAML after a template pass
3. add more sins to this list of sins

tony-schndr (dismissed):
lgtm

PR #2046, opened 2025-06-30
add change to trigger prow CI

deads2k:
/hold

this should give a point to try CI over and over again.


deads2k:
Well, the trigger worked.  Now we get to see whether the bot manages to make an ARO HCP cluster.

PR #2055, opened 2025-06-30
demo: correcting case issues in cluster bicep files

sclarkso:
### What

Updates case of 'Public' and 'LoadBalancer' in cluster generation bicep file for the demo

### Why

Uppercase 'Public' and 'LoadBalancer' cause an error on provisioning.

### DIFF
--- a/demo/bicep/cluster.bicep
+++ b/demo/bicep/cluster.bicep
@@ -93,12 +93,12 @@ resource hcp 'Microsoft.RedHatOpenShift/hcpOpenShiftClusters@2024-06-10-preview'
     }
     console: {}
     api: {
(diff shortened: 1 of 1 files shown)


mbarnes:
Yeah, this was a [small breaking API change](https://github.com/Azure/ARO-HCP/pull/1965) that hasn't been pushed to INT yet. `Public` and `LoadBalancer` are ultimately correct but, you can use your suggested change as a workaround until the frontend image in INT gets an update.  I'll see to getting it updated.

mbarnes (approved):
This is a temporary fix but I'll take it since the pipeline for pushing out newer INT images is broke at the moment.

PR #2069, opened 2025-07-01
Add support for OSDisk encryption

mociarain:

Outstanding Questions:
- Can we PATCH this? 
  - This is a question for Hypershift


api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -435,6 +435,19 @@ model NodePoolPlatformProfile {
...
+   * This needs to exist in the same subscription id listed in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.SubscriptionID.
+   * DiskEncryptionSetID should also exist in a resource group under the same subscription id and the same location
+   * listed in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.Location.
+   *
+   * Details on how to create a Disk Encryption Set can be found here:
+   * https://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-customer-managed-keys-portal#set-up-your-disk-encryption-set
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  @maxLength(285)
+  diskEncryptionSetId?: url;
```
miguelsorianod, line 449: I would make clear that this is for the *OS Disk* as part of the name. Unless we are intentionally designing this same one be used for potential future data disks

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -435,6 +435,19 @@ model NodePoolPlatformProfile {
   @visibility(Lifecycle.Read, Lifecycle.Create)
   diskStorageAccountType?: DiskStorageAccountType = "Premium_LRS";
 
+  /** The ID of the DiskEncryptionSet resource to use to encrypt the OS disks for the VMs.
+   * Can be used with either platform (Azure) managed, or customer managed encryption keys.
```
miguelsorianod, line 439: Is this information accurate? Can you have a disk encryption set only with platform managed keys?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
```diff
@@ -1462,6 +1462,16 @@
             "create"
           ]
         },
+        "diskEncryptionSetId": {
+          "type": "string",
+          "format": "uri",
+          "description": "The ID of the DiskEncryptionSet resource to use to encrypt the OS disks for the VMs.\nCan be used with either platform (Azure) managed, or customer managed encryption keys.\nThis needs to exist in the same subscription id listed in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.SubscriptionID.\nDiskEncryptionSetID should also exist in a resource group under the same subscription id and the same location\nlisted in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.Location.\n\nDetails on how to create a Disk Encryption Set can be found here:\nhttps://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-customer-managed-keys-portal#set-up-your-disk-encryption-set",
+          "maxLength": 285,
+          "x-ms-mutability": [
```
miguelsorianod, line 1470: Pending to define this

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mociarain: All the other attributes for attributes are just disk so I'd rather keep them all as 'disk***' or nest them all under some 'NodePoolOSDiskSpec' object so the context is clear (this is what [hypershift does](https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/azure.go#L273))... but maybe we've already gone too far and this would be a breaking change already?? @mbarnes-azure any thoughts on this?

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mociarain: Good spot. So, reading deeper you can have double encryption with platform and customer managed keys but not just platform managed. 

I'll update the string

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
```diff
@@ -1462,6 +1462,16 @@
             "create"
           ]
         },
+        "diskEncryptionSetId": {
+          "type": "string",
+          "format": "uri",
+          "description": "The ID of the DiskEncryptionSet resource to use to encrypt the OS disks for the VMs.\nCan be used with either platform (Azure) managed, or customer managed encryption keys.\nThis needs to exist in the same subscription id listed in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.SubscriptionID.\nDiskEncryptionSetID should also exist in a resource group under the same subscription id and the same location\nlisted in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.Location.\n\nDetails on how to create a Disk Encryption Set can be found here:\nhttps://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-customer-managed-keys-portal#set-up-your-disk-encryption-set",
+          "maxLength": 285,
```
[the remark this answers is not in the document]
mociarain: https://github.com/openshift/hypershift/blob/9720e90964cf527422033550889835ee450ba46a/api/hypershift/v1beta1/azure.go#L313

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
mbarnes: I think renaming or grouping all the `disk*` fields at this point would have to be a BU decision.  Maybe bring it up in the next API meeting?

My suggestion for now is just make sure all the doc comments for these fields say "OS disk".  The comment for `diskStorageAccountType` might need clarified.

miguelsorianod: I would prefer having them as os_disk_* or within an os disk section. If we leave them with generic name and in the future we add support for data disks we will end up with generic attribute names and then data_* names which is not ideal, because we will probably want to have differentiated configuration for them.

mociarain: We decided to nest things under OSDisk

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -491,6 +510,13 @@ model NodePoolAutoScaling {
   max?: int32;
 }
 
+scalar DiskEncryptionSetResourceId
+  extends Azure.Core.armResourceIdentifier<[
```
miguelsorianod, line 514: What are the implications of this?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
mociarain: Waiting on Hypershift to tell me if we can PATCH this. Thread https://redhat-external.slack.com/archives/C075PHEFZKQ/p1751380214598679

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
miguelsorianod: Does this mean we still need maxLen 285 then? What is the length that that defines?

mociarain: A quick google suggests that there is no official limit. I've seen people suggest a [theoretical max](https://learn.microsoft.com/en-us/answers/questions/1189747/max-length-of-azure-resource-id-and-whether-resour) of:
- ~1000: based on the sum of the max length of the various possible components
- 260:  based on the max path size allowed in Windows.... 

Either way OCP says 285 so I think this is useful

mbarnes (approved):

PR #2084, opened 2025-07-03
Add Authorized IP to Api Server

hbhushan3:
Replaces https://github.com/Azure/ARO-HCP/pull/1932

Ticket -
https://issues.redhat.com/browse/ARO-15671

What
Spec for adding authorizedIpRanges to api server

Support for up to 5000 authorized IP ranges (IPv4 CIDR)
Configurable at cluster creation time and updatable post-deployment
Feature can be disabled by providing an empty array (reverts to default access)
Added co-pilot generated examples

### DIFF
--- a/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
+++ b/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
@@ -200,6 +200,11 @@ model ApiProfile {
   /** The internet visibility of the OpenShift API server */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   visibility?: Visibility = Visibility.Public;
+
+  /** The list of authorized IPv4 CIDR blocks allowed to access the API server. Maximum 5000 entries. */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  @maxItems(5000)
+  authorizedCidrs?: string[];
 }
 
 /** The internet visibility of the OpenShift API server */
--- a/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
+++ b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
@@ -1108,6 +1108,19 @@
             "read",
             "create"
           ]
+        },
+        "authorizedCidrs": {
+          "type": "array",
+          "description": "The list of authorized IPv4 CIDR blocks allowed to access the API server. Maximum 5000 entries.",
+          "maxItems": 5000,
+          "items": {
+            "type": "string"
+          },
+          "x-ms-mutability": [
+            "read",
+            "update",
+            "create"
+          ]
         }
       },
       "required": [
--- a/api/redhatopenshift/HcpCluster.Management/examples/2024-06-10-preview/HcpOpenShiftClusters_CreateOrUpdate_MaximumSet_Gen.json
+++ b/api/redhatopenshift/HcpCluster.Management/examples/2024-06-10-preview/HcpOpenShiftClusters_CreateOrUpdate_MaximumSet_Gen.json
@@ -23,7 +23,11 @@
           "hostPrefix": 21
         },
         "api": {
-          "visibility": "Public"
+          "visibility": "Public",
+          "authorizedCidrs": [
+            "192.168.1.0/24",
+            "10.0.0.0/16"
+          ]
         },
         "platform": {
           "managedResourceGroup": "nhyhywrxupo",
--- a/api/redhatopenshift/HcpCluster.Management/examples/2024-06-10-preview/HcpOpenShiftClusters_Get_MaximumSet_Gen.json
+++ b/api/redhatopenshift/HcpCluster.Management/examples/2024-06-10-preview/HcpOpenShiftClusters_Get_MaximumSet_Gen.json
@@ -28,7 +28,11 @@
           },
           "api": {
             "visibility": "Public",
-            "url": "https://api.test.shrd.usw3test.hcp.osadev.cloud:443"
+            "url": "https://api.test.shrd.usw3test.hcp.osadev.cloud:443",
+            "authorizedCidrs": [
+              "192.168.1.0/24",
+              "203.0.113.0/24"
+            ]
           },
           "platform": {
             "managedResourceGroup": "nhyhywrxupo",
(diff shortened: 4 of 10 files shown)


mociarain:
Does this replace this [PR](https://github.com/Azure/ARO-HCP/pull/1932) and can we close it?

hbhushan3:

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -196,8 +196,17 @@ model ApiProfile {
   /** The internet visibility of the OpenShift API server */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   visibility?: Visibility = Visibility.Public;
+
+  /** The list of authorized IP ranges (IPv4 CIDRs) allowed to access the API server. Maximum 5000 entries. */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  @maxItems(5000)
+  authorizedIpRanges?: ipCidr[];
```
mbarnes, line 203: There was a suggestion in https://github.com/Azure/ARO-HCP/pull/1932 to rename this to `authorizedCIDRs`.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -196,8 +196,17 @@ model ApiProfile {
...
 }
 
+/** IP CIDR block (supports IPv4) */
+@pattern("^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\/([0-9]|[1-2][0-9]|3[0-2]))$")
```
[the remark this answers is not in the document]
hbhushan3: The pattern is correct. In TypeSpec, backslashes must be double-escaped (\\) because they're inside string literals. Otherwise, we will get a compile error by tsp. Also tested this out using prism mock.

We're escaping:

. → \. (literal dots in IP addresses)
/ → \/ (literal forward slash in CIDR notation)

For regex101 tester, use 
^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\/([0-9]|[1-2][0-9]|3[0-2]))$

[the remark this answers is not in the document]
mbarnes: We don't need this pattern in the ARM API; the field type can just be `string`.  The RP will validate that it's a valid CIDR.  (See [NetworkProfile](https://github.com/Azure/ARO-HCP/blob/68e4fac9371a007a5cbca543d547d17c94c2341e/internal/api/hcpopenshiftcluster.go#L64-L66) for examples.)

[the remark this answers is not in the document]
hbhushan3: Also updated existing MaximumSet_Gen files with AuthorizedCidrs.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
hbhushan3: We are using authorizedCidrs now. Thanks!

hbhushan3:

mbarnes (approved):
LGTM.  I'll defer merging to @mociarain in case he has any final comments.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -200,6 +200,11 @@ model ApiProfile {
   /** The internet visibility of the OpenShift API server */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   visibility?: Visibility = Visibility.Public;
+
+  /** The list of authorized IPv4 CIDR blocks allowed to access the API server. Maximum 5000 entries. */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  @maxItems(5000)
```
miguelsorianod, line 206: How does this match with the fact that if I understand correctly the corresponding attribute in Hypershift is https://github.com/openshift/hypershift/blob/2892457510cc7a86b5362ce10d0e8757a7e1fbec/api/hypershift/v1beta1/hostedcluster_types.go#L1082 which has a limit of 25 entries? is that going to be increased on that side too?

I don't think we are going to get an answer to this today so we can merge if needed but we should follow up on this as there seems to be a big disconnect

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -200,6 +200,11 @@ model ApiProfile {
   /** The internet visibility of the OpenShift API server */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   visibility?: Visibility = Visibility.Public;
+
+  /** The list of authorized IPv4 CIDR blocks allowed to access the API server. Maximum 5000 entries. */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
```
miguelsorianod, line 205: Did we verify that updates are supported on the "backend" side?

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -200,6 +200,11 @@ model ApiProfile {
   /** The internet visibility of the OpenShift API server */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   visibility?: Visibility = Visibility.Public;
+
+  /** The list of authorized IPv4 CIDR blocks allowed to access the API server. Maximum 5000 entries. */
```
miguelsorianod, line 204: Here I think we have a lack of definition of constraints around this:
* Are we going to allow overlapping of ranges?
* Are we going to somehow consider the NSG / Subnets related to the Cluster / Node Pools?
* Are we going to prevent from some specific blocks to be provided?

Similarly to other comments, we might not get the needed answers today but we should follow up on this

cc @nimrodshn

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
```diff
@@ -1108,6 +1108,19 @@
             "read",
             "create"
           ]
+        },
+        "authorizedCidrs": {
+          "type": "array",
+          "description": "The list of authorized IPv4 CIDR blocks allowed to access the API server. Maximum 5000 entries.",
+          "maxItems": 5000,
```
jharrington22, line 1115: We got clarification that this should be 500.

PR #2086, opened 2025-07-03
tooling: bump ARO-Tools

stevekuznetsov:


tooling/templatize/internal/end2end/e2e.go
```diff
@@ -54,7 +52,7 @@ type bicepTemplate struct {
 }
 
 type e2eImpl struct {
-	config   config.Configuration
```
stevekuznetsov, line 57: This is NOT what this type means. This code was (ab)using things.

PR #2090, opened 2025-07-04
Add key vault to support etcd encryption

patilsuraj767:
[ARO-18943](https://issues.redhat.com/browse/ARO-18943)

- Add Key Vault creation and RBAC configuration in customer infrastructure setup
- Configure Azure operators managed identities
- Update demo scripts and environment variables for KMS support

### DIFF
--- a/cluster-service/cluster-creation.md
+++ b/cluster-service/cluster-creation.md
@@ -129,7 +129,20 @@ This document outlines the process of creating an HCP via the Cluster Service ru
         export SERVICE_MANAGED_IDENTITY_UAMI=$(az identity show -n ${USER}-${CS_CLUSTER_NAME}-service-managed-identity-${OPERATORS_UAMIS_SUFFIX} -g <resource-group> | jq -r '.id')
         ```
 
+    * Create Azure key vault for storing etcd data encryption key
+        ```bash
+        RESOURCENAME=<resource-group>
+        SUBSCRIPTION_ID=<subscription-id>
+        CURRENT_USER_ID=$(az ad signed-in-user show -o json | jq -r '.id')
+        az keyvault create --name ${USER}-${CS_CLUSTER_NAME}-"kv" -g ${RESOURCENAME} --location "westus3" --enable-rbac-authorization true
+        az role assignment create --assignee "${CURRENT_USER_ID}" --role "Key Vault Crypto Officer" --scope "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${RESOURCENAME}/providers/Microsoft.KeyVault/vaults/${USER}-${CS_CLUSTER_NAME}-kv"
+        az keyvault key create --vault-name ${USER}-${CS_CLUSTER_NAME}-"kv" --name "etcd-data-kms-encryption-key" --kty RSA --size 2048
+        
+        export etcd_key_version=$(az keyvault key list-versions --vault-name ${USER}-${CS_CLUSTER_NAME}-"kv" --name etcd-data-kms-encryption-key  --query "[].kid" -o tsv | cut -d'/' -f6)
+        ```
+
 1. Create the cluster. This assumes OCP 4.19 clusters will be created.
+
     > [!NOTE] See the [Cluster Service API](https://api.openshift.com/#/default/post_api_clusters_mgmt_v1_clusters) documentation
     > for further information on the properties within the payload below
 
@@ -166,6 +179,21 @@ This document outlines the process of creating an HCP via the Cluster Service ru
         "managed_resource_group_name": "$MANAGEDRGNAME",
         "subnet_resource_id": "$SUBNETRESOURCEID",
         "network_security_group_resource_id":"$NSG",
+        "etcd_encryption": {
+          "data_encryption": {
+            "key_management_mode": "customer_managed",
+            "customer_managed": {
+              "encryption_type": "kms",
+              "kms": {
+                "active_key": {
+                  "key_vault_name": ${USER}-${CS_CLUSTER_NAME}-"kv",
+                  "key_name": "etcd-data-kms-encryption-key",
+                  "key_version": ${etcd_key_version}
+                }
+              }
+            }
+          }
+        },
         "operators_authentication": {
           "managed_identities": {
             "managed_identities_data_plane_identity_url": "https://dummyhost.identity.azure.net",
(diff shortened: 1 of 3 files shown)


cluster-service/deploy/templates/azure-operators-managed-identities-config.configmap.yaml
```diff
@@ -77,7 +77,7 @@ data:
           - resourceId: '/providers/Microsoft.Authorization/roleDefinitions/{{ .id }}'
             name: '{{ .name }}'
         {{- end }}
-        optional: true
+        optional: false
```
miguelsorianod, line 80: This is not correct. It is still an optional identity. It is only required when etcd data level encryption with kms is enabled. We refer to MIs that become mandatory when their corresponding functionality is enabled  as "optional: true" in this config file. This should remain unchanged

demo/02-customer-infra.sh
```diff
@@ -18,3 +18,29 @@ az network vnet create \
...
   --nsg "${NSG_ID}"  --location ${LOCATION}
+
+az keyvault create \
+  --name "${CUSTOMER_KV_NAME}" \
+  --resource-group "${CUSTOMER_RG_NAME}" \
+  --location "${LOCATION}" \
+  --enable-rbac-authorization true
+
+# Get current user's object ID
+CURRENT_USER_ID=$(az ad signed-in-user show --query id -o tsv)
```
miguelsorianod, line 29: We could set it to:

$(az ad signed-in-user show -o json | jq -r '.id')

as done in other places

demo/02-customer-infra.sh
```diff
@@ -18,3 +18,29 @@ az network vnet create \
...
+
+# Assign Key Vault Crypto Officer role to current user
+az role assignment create \
+  --assignee "${CURRENT_USER_ID}" \
+  --role "Key Vault Crypto Officer" \
+  --scope "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CUSTOMER_RG_NAME}/providers/Microsoft.KeyVault/vaults/${CUSTOMER_KV_NAME}"
+
+# Wait for role assignment to propagate
+echo "Waiting for RBAC role assignment to propagate..."
+sleep 30
```
miguelsorianod, line 39: This does not seem very reliable. Can we come with a better method?

demo/02-customer-infra.sh
```diff
@@ -18,3 +18,29 @@ az network vnet create \
...
+
+# Create RSA key in Key Vault
+az keyvault key create \
+  --vault-name "${CUSTOMER_KV_NAME}" \
+  --name "encryption-key" \
```
miguelsorianod, line 44: This name is too generic.

Let's name it something like:
etcd-data-kms-encryption-key

demo/02-customer-infra.sh
```diff
@@ -18,3 +18,29 @@ az network vnet create \
...
+  --location "${LOCATION}" \
+  --enable-rbac-authorization true
+
+# Get current user's object ID
+CURRENT_USER_ID=$(az ad signed-in-user show --query id -o tsv)
+
+# Assign Key Vault Crypto Officer role to current user
+az role assignment create \
+  --assignee "${CURRENT_USER_ID}" \
+  --role "Key Vault Crypto Officer" \
```
miguelsorianod, line 34: Can we ask the lifecycle team that there are no concerns about using this role?

I understand that there shouldn't as it is scoped to the KV that we are creating, but just in case they have a specific policy about this.

cluster-service/deploy/templates/azure-operators-managed-identities-config.configmap.yaml
miguelsorianod: Yes that's the idea. As you can see in that link you referenced, in a couple  of places the following is mentioned:
```
		// the identity is only required when the corresponding feature is enabled.
		// At the moment of writing, we don't have optional features that requires MI
		// once we start adding such features, an extra check needs to go here to
		// determine if the feature is disabled before adding the identity to the slices
		// of identities that are not needed
```

etcd data level encryption with kms is one of such features.

demo/02-customer-infra.sh
patilsuraj767: Modified it.

demo/02-customer-infra.sh
```diff
@@ -18,3 +18,39 @@ az network vnet create \
...
+max_retries=30
+counter=0
+while ! az keyvault key list --vault-name "${CUSTOMER_KV_NAME}" &>/dev/null; do
+    if [ $counter -eq $max_retries ]; then
+        echo "Timeout waiting for RBAC role assignment to propagate"
+        exit 1
+    fi
+    echo "Waiting for RBAC permissions... (attempt $((counter + 1))/$max_retries)"
+    counter=$((counter + 1))
+    sleep 10
```
miguelsorianod, line 48: This seems like a big sleep. Can we reduce it a bit? take into account that max retries should be increased in that case too

demo/02-customer-infra.sh
```diff
@@ -18,3 +18,39 @@ az network vnet create \
...
+az role assignment create \
+  --assignee "${CURRENT_USER_ID}" \
+  --role "Key Vault Crypto Officer" \
+  --scope "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CUSTOMER_RG_NAME}/providers/Microsoft.KeyVault/vaults/${CUSTOMER_KV_NAME}"
+
+# Wait for role assignment to propagate with timeout
+echo "Waiting for RBAC role assignment to propagate..."
+max_retries=30
+counter=0
+while ! az keyvault key list --vault-name "${CUSTOMER_KV_NAME}" &>/dev/null; do
```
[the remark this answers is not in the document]
patilsuraj767: Yes, Verified. 
User is not able to list the keys if role is not assigned. Once role assignment is propagated it works.

cluster-service/cluster-creation.md
```diff
@@ -31,7 +31,7 @@ This document outlines the process of creating an HCP via the Cluster Service ru
     ocm login --url=http://localhost:8000 --use-auth-code
     ```
 
-1) Create pre-requisite resources for cluster creation
+2) Create pre-requisite resources for cluster creation
```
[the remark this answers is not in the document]
patilsuraj767: In the markdown preview I can't see numbering in automatically working - https://github.com/Azure/ARO-HCP/blob/main/cluster-service/cluster-creation.md#creating-a-cluster

I tired 1. but not working in my local maybe some indentation or some other issue.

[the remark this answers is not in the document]
patilsuraj767: Created the PR to fix the indentation - https://github.com/Azure/ARO-HCP/pull/2120

demo/02-customer-infra.sh
miguelsorianod: Discussed through other channels and it should be fine

PR #2091, opened 2025-07-04
Frontend API struct Support: HCP-Level Autoscaling for All NodePools

venkateshsredhat:
https://issues.redhat.com/browse/ARO-17695

What
1P STUB Frontend: Support HCP-Level Autoscaling for All NodePools

Why
https://issues.redhat.com/browse/XCMSTRAT-1037

Special notes for your reviewer
Adding Frontend Struct level changes for Spec changes PR : https://github.com/Azure/ARO-HCP/pull/1936


internal/api/v20240610preview/hcpopenshiftclusters_methods.go
```diff
@@ -308,6 +320,21 @@ func normalizePlatform(p *generated.PlatformProfile, out *api.PlatformProfile) {
 	}
 }
 
+func normailzeAutoscaling(p *generated.ClusterAutoscalingProfile, out *api.ClusterAutoscalingProfile) {
```
mbarnes, line 323:
```suggestion
func normalizeAutoscaling(p *generated.ClusterAutoscalingProfile, out *api.ClusterAutoscalingProfile) {
```

internal/api/v20240610preview/hcpopenshiftclusters_methods.go
```diff
@@ -212,6 +221,9 @@ func (c *HcpOpenShiftCluster) Normalize(out *api.HCPOpenShiftCluster) {
 			if c.Properties.Platform != nil {
 				normalizePlatform(c.Properties.Platform, &out.Properties.Platform)
 			}
+			if c.Properties.Autoscaling != nil {
+				normailzeAutoscaling(c.Properties.Autoscaling, &out.Properties.Autoscaling)
```
mbarnes, line 225:
```suggestion
				normalizeAutoscaling(c.Properties.Autoscaling, &out.Properties.Autoscaling)
```

mbarnes (changes requested):
Looks good so far, aside from a misspelling.  We'll worry about validation later, but one thing I _would_ like to see here are any non-zero default values for these fields added to `api.NewDefaultHCPOpenShiftCluster`.

mbarnes:
> But my question is will it be even required ? Autoscaling is a optional field right , why do we set non zero values here ?

`Network` is also an optional field.  We set non-zero default values _because_ it's optional.  If the user does not provide their own autoscaling values then these default values are used.  Either way, responses to cluster GET/PUT/PATCH requests will show an `autoscaling` section with the current values.

> Edit : probably the MaxNodesTotal can be removed ? without setting it . It wont have a limit , the number 5 now is just random .

Yes, remove `MaxNodesTotal`.

mbarnes (approved):

PR #2105, opened 2025-07-07
Implement hcpopenshiftversion

katherinelc321:
Related to this Jira: https://issues.redhat.com/browse/ARO-17700

### What

Implement HCPOpenShiftVersion internal model 

### Why

API Specs were merged in pull [1953 ](https://github.com/Azure/ARO-HCP/pull/1953) and this is the implementation of the internal model


mbarnes (changes requested):
There's some bits that can be dropped on account of `hcpOpenShiftVersion` being a read-only resource type.  Makes the implementation a lot simpler.  The remaining code looks great.

mbarnes (approved):

PR #2110, opened 2025-07-08
tracing: fix OpenTelemetry collector config

simonpasquier:
### What

After deploying the new setup with the OTEL collector, I realized that the collector was sending gRPC OTLP to the HTTP ports of Jaeger and lgtm.

It also updates the README.md with the latest changes.


simonpasquier:
PTAL @janboll @geoberle @frzifus

PR #2111, opened 2025-07-08
Add wait/conditional scaledown

janboll:
### What

Scale down ACM after setup to reduce likelihood of CRD ownership conflict

### Why

https://redhat-external.slack.com/archives/C075PHEFZKQ/p1751015987160579

### DIFF
--- a/acm/deploy.sh
+++ b/acm/deploy.sh
@@ -14,7 +14,7 @@ set -e
 # Check if MCE resource exists
 if kubectl get mce multiclusterengine -n ${MCE_NS} >/dev/null 2>&1; then
     phase=$(kubectl -n ${MCE_NS} get mce multiclusterengine -o json | jq -r '.status.phase')
-    
+
     if [ "${phase}" = "Paused" ] && [ "${MCE_PAUSE_RECONCILIATION}" = "true" ]; then
         echo "MCE is already paused, skipping deploy"
         exit 0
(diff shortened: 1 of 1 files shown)


acm/pipeline.yaml
```diff
@@ -20,3 +20,18 @@ resourceGroups:
...
+    aksCluster: '{{ .mgmt.aks.name  }}'
+    action: Shell
+    command: make scale-down
+    dryRun:
+      variables:
+      - name: DRY_RUN
+        value: "true"
+    variables:
+    - name: ARO_HCP_IMAGE_ACR
+      configRef: acr.svc.name
```
geoberle, line 33: do we need this in the scaledown target?

acm/deploy.sh
```diff
@@ -0,0 +1,33 @@
...
+    clc-state-metrics \
+    --namespace ${MCE_NS}
+set -e
+
+phase=$(kubectl -n multicluster-engine get mce multiclusterengine -o json | jq -r '.status.phase')
+
+if [ "${phase}" = "Paused" ]; then
+    echo "MCE is already paused, skipping deploy"
+    exit 0
+fi
```
geoberle, line 19: how will unpause work?

janboll: Manually remove the annotation

geoberle: maybe we can enhance the condition for `exit 0` and check for MCE_PAUSE_RECONCILIATION == true as well.

* in case MCE does not exist yet, `phase` will be empty, so not matching `Pause` and will proceed with installing MCE (potentially pausing it at the end of the script and scaling it down in the next step
* in case MCE exists and is paused + MCE_PAUSE_RECONCILIATION indicates it should be paused, we can exit since there is nothing to do here
* in case MCE exists and is paused but MCE_PAUSE_RECONCILIATION indicates it should not be paused, we proceed with the MCE deployment (which scales it up again) - at the end of the script, the pause annotation is also set to false

PR #2114, opened 2025-07-08
Query hcp cluster versions in ocm

katherinelc321:
ARO-17734

### What

Implement methods in ocm to query hcp cluster versions GetVersions(), ListVersions()

### Why

RP Frontend needs to support looking up HCP Cluster versions

### DIFF
--- a/internal/ocm/internalid.go
+++ b/internal/ocm/internalid.go
@@ -29,10 +29,12 @@ const (
 	v1ClusterPattern              = v1Pattern + "/clusters/*"
 	v1NodePoolPattern             = v1ClusterPattern + "/node_pools/*"
 	v1BreakGlassCredentialPattern = v1ClusterPattern + "/break_glass_credentials/*"
+	v1VersionPattern              = v1Pattern + "/versions/*"
 
 	aroHcpV1Alpha1Pattern         = "/api/aro_hcp/v1alpha1"
 	aroHcpV1Alpha1ClusterPattern  = aroHcpV1Alpha1Pattern + "/clusters/*"
 	aroHcpV1Alpha1NodePoolPattern = aroHcpV1Alpha1ClusterPattern + "/node_pools/*"
+	aroHcpV1Alpha1VersionPattern  = aroHcpV1Alpha1Pattern + "/versions/*"
 )
 
 func GenerateClusterHREF(clusterName string) string {
@@ -77,6 +79,11 @@ func (id *InternalID) validate() error {
 		return nil
 	}
 
+	if match, _ = path.Match(v1VersionPattern, id.path); match {
+		id.kind = cmv1.VersionKind
+		return nil
+	}
+
 	if match, _ = path.Match(aroHcpV1Alpha1ClusterPattern, id.path); match {
 		id.kind = arohcpv1alpha1.ClusterKind
 		return nil
@@ -84,6 +91,11 @@ func (id *InternalID) validate() error {
 
 	if match, _ = path.Match(aroHcpV1Alpha1NodePoolPattern, id.path); match {
 		id.kind = arohcpv1alpha1.NodePoolKind
+		return nil
+	}
+
+	if match, _ = path.Match(aroHcpV1Alpha1VersionPattern, id.path); match {
+		id.kind = arohcpv1alpha1.VersionKind
 		return nil
 	}
 
@@ -198,4 +210,11 @@ func (id *InternalID) GetBreakGlassCredentialClient(transport http.RoundTripper)
 		return nil, false
 	}
 	return cmv1.NewBreakGlassCredentialClient(transport, id.path), true
+}
+
+func (id *InternalID) GetVersionClient(transport http.RoundTripper) (*cmv1.VersionClient, bool) {
+	if id.Kind() != cmv1.VersionKind {
+		return nil, false
+	}
+	return cmv1.NewVersionClient(transport, id.path), true
 }
--- a/internal/ocm/iterators.go
+++ b/internal/ocm/iterators.go
@@ -28,6 +28,11 @@ type ClusterListIterator struct {
 	err     error
 }
 
+type VersionsListIterator struct {
+	request *arohcpv1alpha1.VersionsListRequest
+	err     error
+}
+
 // Items returns a push iterator that can be used directly in for/range loops.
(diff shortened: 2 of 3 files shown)


internal/ocm/ocm.go
```diff
@@ -383,3 +391,27 @@ func (csc *clusterServiceClient) ListBreakGlassCredentials(clusterInternalID Int
 	}
 	return BreakGlassCredentialListIterator{request: breakGlassCredentialsListRequest}
 }
+
+func (csc *clusterServiceClient) GetVersion(ctx context.Context, internalID InternalID) (*arohcpv1alpha1.Version, error) {
+	client, ok := internalID.GetHCPClusterVersionClient(csc.conn)
+	if !ok {
+		return nil, fmt.Errorf("OCM path is not a version: %s", internalID)
+	}
```
mbarnes, line 399: I still could be proven wrong here, but I don't foresee a need to use `InternalID` with version resources.  Let's keep this simple for now and just take a version string argument, and get a version client similar to what you're doing in `ListVersions`.
```suggestion
func (csc *clusterServiceClient) GetVersion(ctx context.Context, id string) (*arohcpv1alpha1.Version, error) {
	client := csc.conn.AroHCP().V1alpha1().Versions().Version(id)
```

internal/ocm/internalid.go
```diff
@@ -199,3 +199,10 @@ func (id *InternalID) GetBreakGlassCredentialClient(transport http.RoundTripper)
 	}
 	return cmv1.NewBreakGlassCredentialClient(transport, id.path), true
 }
+
+func (id *InternalID) GetHCPClusterVersionClient(transport http.RoundTripper) (*arohcpv1alpha1.VersionClient, bool) {
+	if id.Kind() != arohcpv1alpha1.VersionKind {
+		return nil, false
+	}
+	return arohcpv1alpha1.NewVersionClient(transport, id.path), true
+}
```
mbarnes, line 208: I think we can drop this.  See my comment above.

mbarnes (changes requested):

frontend/pkg/frontend/helpers.go
```diff
@@ -219,6 +219,22 @@ func (f *Frontend) MarshalResource(ctx context.Context, resourceID *azcorearm.Re
 
 	logger := LoggerFromContext(ctx)
 
+	if resourceID.ResourceType.Type == (api.ClusterVersionTypeName) {
```
mbarnes, line 222: This comparison should be case-insensitive.
```suggestion
	if strings.EqualFold(resourceID.ResourceType.Type, api.ClusterVersionTypeName) {
```

internal/ocm/ocm.go
```diff
@@ -383,3 +391,25 @@ func (csc *clusterServiceClient) ListBreakGlassCredentials(clusterInternalID Int
 	}
 	return BreakGlassCredentialListIterator{request: breakGlassCredentialsListRequest}
 }
+
+func (csc *clusterServiceClient) GetVersion(ctx context.Context, versionName string) (*cmv1.Version, error) {
+	client := cmv1.NewVersionClient(csc.conn, versionName)
```
mbarnes, line 396: We should be using `arohcpv1alpha1` instead of `cmv1`.  Correct me if you find I'm wrong, but I believe version resources have been ported over to the `aro_hcp/v1alpha1` API.
```suggestion
func (csc *clusterServiceClient) GetVersion(ctx context.Context, versionName string) (*arohcpv1alpha1.Version, error) {
	client := arohcpv1alpha1.NewVersionClient(csc.conn, versionName)
```

internal/ocm/ocm.go
```diff
@@ -383,3 +391,25 @@ func (csc *clusterServiceClient) ListBreakGlassCredentials(clusterInternalID Int
...
+	}
+	version, ok := resp.GetBody()
+	if !ok {
+		return nil, fmt.Errorf("empty response body")
+	}
+	return version, nil
+}
+
+func (csc *clusterServiceClient) ListVersions(searchExpression string) VersionsListIterator {
+	versionsListRequest := csc.conn.ClustersMgmt().V1().Versions().List()
```
mbarnes, line 410: Same as above.  I guess I overlooked this on the first go round.
```suggestion
	versionsListRequest := csc.conn.AroHCP().V1alpha1().Versions().List()
```

internal/ocm/iterators.go
```diff
@@ -184,3 +189,52 @@ func (iter BreakGlassCredentialListIterator) Items(ctx context.Context) iter.Seq
 func (iter BreakGlassCredentialListIterator) GetError() error {
 	return iter.err
 }
+
+// Items returns a push iterator that can be used directly in for/range loops.
+// If an error occurs during paging, iteration stops and the error is recorded.
+func (iter VersionsListIterator) Items(ctx context.Context) iter.Seq[*cmv1.Version] {
+	return func(yield func(*cmv1.Version) bool) {
```
mbarnes, line 196:
```suggestion
func (iter VersionsListIterator) Items(ctx context.Context) iter.Seq[*arohcpv1alpha1.Version] {
	return func(yield func(*arohcpv1alpha1.Version) bool) {
```

internal/ocm/tracing.go
```diff
@@ -254,3 +254,11 @@ func (csc *clusterServiceClientWithTracing) DeleteBreakGlassCredentials(ctx cont
 func (csc *clusterServiceClientWithTracing) ListBreakGlassCredentials(clusterInternalID InternalID, searchExpression string) BreakGlassCredentialListIterator {
 	return csc.csc.ListBreakGlassCredentials(clusterInternalID, searchExpression)
 }
+
+func (csc *clusterServiceClientWithTracing) GetVersion(ctx context.Context, versionName string) (*cmv1.Version, error) {
```
mbarnes, line 258:
```suggestion
func (csc *clusterServiceClientWithTracing) GetVersion(ctx context.Context, versionName string) (*arohcpv1alpha1.Version, error) {
```

mbarnes (changes requested):
This is looking good.

Main correction this time is we need to be using the `/api/aro_hcp/v1alpha1` endpoint instead of `/api/clusters_mgmt/v1`.

mbarnes (changes requested):
One correction left.  Everything else looks good.

mbarnes (approved):

PR #2115, opened 2025-07-08
Add indication that errors originate from cluster-service.

deads2k:
### What

This allows us to determine which part of the system has lost track of a particular cluster or nodepool.

This is part one of two. Part two will require ripples for better information about internal errors.

### Why

Without this we are unable to determine why a call is failing and thus unable correct the field problem.

/assign @mbarnes

### DIFF
--- a/frontend/pkg/frontend/ocm.go
+++ b/frontend/pkg/frontend/ocm.go
@@ -437,7 +437,7 @@ func CSErrorToCloudError(err error, resourceID *azcorearm.ResourceID) *arm.Cloud
 			return arm.NewCloudError(
 				statusCode,
 				arm.CloudErrorCodeInvalidRequestContent,
-				"", "%s", ocmError.Reason())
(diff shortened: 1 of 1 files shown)


mociarain (commented):
AFAIK this message will get bubbled up to the user. Mentioning internals to them in a message seems off to me? Would logging this be better?

deads2k:
> AFAIK this message will get bubbled up to the user. Mentioning internals to them in a message seems off to me? Would logging this be better?

No, I don't think so.  I'm happy to call it anything we like, but in a bug report it is far more useful to get an identifier of what failed.  Right now we're fighting a case of, "it's literally impossible to write a decent bug report".  So we can debate how we describe it.  I'm happy to simply place a UID there.  But we have to have something.

deads2k:
moved PR and conversation to https://github.com/Azure/ARO-HCP/pull/2131

PR #2118, opened 2025-07-08
Add location details for all internal errors.

deads2k:
### What

This allows us to determine which particular area had an internal are without risking internal details of the service like IPs, service identities, etc being leaked via unvetted library error messages.

### Why

We need to be able to debug  our service when something fails.  Without this there are numerous different spots to have internal errors for individual actions and we have no way to track which is causing an issue.

/assign @mbarnes

### DIFF
--- a/internal/api/arm/error.go
+++ b/internal/api/arm/error.go
@@ -141,17 +141,23 @@ func WriteCloudError(w http.ResponseWriter, err *CloudError) {
 	_, _ = WriteJSONResponse(w, err.StatusCode, err)
 }
 
-// NewInternalServerError creates a CloudError for an internal server error
-func NewInternalServerError() *CloudError {
+// NewInternalServerError creates a CloudError for an internal server error.
+// locationHint is a means to find which internal error was encountered.  It must be a constant or known pattern string.
+// This is critical for us to avoid accidentally reporting things like internal IP addresses, internal user identities, or
+// other internal service knowledge via the API.
+func NewInternalServerError(locationHint string) *CloudError {
 	return NewCloudError(
 		http.StatusInternalServerError,
 		CloudErrorCodeInternalServerError, "",
-		"Internal server error.")
+		"Internal server error.  Origin: %q", locationHint)
 }
 
 // WriteInternalServerError writes an internal server error to the given ResponseWriter
-func WriteInternalServerError(w http.ResponseWriter) {
-	WriteCloudError(w, NewInternalServerError())
+// locationHint is a means to find which internal error was encountered.  It must be a constant or known pattern string.
+// This is critical for us to avoid accidentally reporting things like internal IP addresses, internal user identities, or
+// other internal service knowledge via the API.
+func WriteInternalServerError(w http.ResponseWriter, locationHint string) {
+	WriteCloudError(w, NewInternalServerError(locationHint))
 }
 
 // NewConflictError creates a CloudError for a conflict error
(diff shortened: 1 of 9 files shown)


internal/api/arm/error.go
```diff
@@ -141,17 +141,23 @@ func WriteCloudError(w http.ResponseWriter, err *CloudError) {
...
+// NewInternalServerError creates a CloudError for an internal server error.
+// locationHint is a means to find which internal error was encountered.  It must be a constant or known pattern string.
+// This is critical for us to avoid accidentally reporting things like internal IP addresses, internal user identities, or
+// other internal service knowledge via the API.
+func NewInternalServerError(locationHint string) *CloudError {
 	return NewCloudError(
 		http.StatusInternalServerError,
 		CloudErrorCodeInternalServerError, "",
-		"Internal server error.")
+		"Internal server error.  Origin: %q", locationHint)
```
mbarnes, line 152: This is useful information but there might be a more structured way to include it.  Not sure if you can access this yet, but the [Resource Provider Contract defines the structure for error responses](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/common-api-details.md#error-response-content).

It gives the example below.  (Nested errors are possible but wouldn't apply here.)  But the `additionalInfo` field could _maybe_ be utilized for this.  Neither ARO-HCP nor ARO Classic has done anything with this field to date so I might need to research what other Azure services do with it.

My concern is we don't want to force future client code to potentially have to parse error messages to extract this kind of information.
```
{
  "error": {
    "code": "",
    "message": "",
    "target": "",
    "additionalInfo": [],
    "details": [
      {
        "code": "",
        "target": "",
        "message": "",
        "additionalInfo": [
          {
            "type": "PolicyViolation",
            "info": {
              "policySetDefinitionDisplayName": "Secure the environment",
              "policySetDefinitionId": "/subscriptions/00000-00000-0000-000/providers/Microsoft.Authorization/policySetDefinitions/TestPolicySet",
              "policyDefinitionDisplayName": "Allowed locations",
              "policyDefinitionId": "/subscriptions/00000-00000-0000-000/providers/Microsoft.Authorization/policyDefinitions/TestPolicy1",
              "policyAssignmentDisplayName": "Allow Central US and WEU only",
              "policyAsssignmentId": "/subscriptions/00000-00000-0000-000/providers/Microsoft.Authorization/policyAssignments/TestAssignment1"
            }
          },
          {
            "type": "SomeOtherViolation",
            "info": {
              "innerException": "innerException Details"
            }
          }
        ]
      }
    ]
  }
}
```
| Element name | Type | Description |
|----|----|----|
| additionalInfo| array (optional)| An array of objects with "type" (string), and "info" (object) properties. The schema of "info" is service-specific and dependent on the "type" string. |

mbarnes: Perhaps it could be structured like:
```
"additionalInfo": [
  {
    "Type": "Hints",
    "Info": {
      "Location": locationHint,
    },
  }
]
```
("Component" might be a better term since "Location" implies Azure region.)
Not yet sure what other hints we might want, but at least the structure is extendable.

bennerv: If a client has a bug report, what's stopping them from just passing us a correlationID and we look into our internal systems to see what went wrong?  It's not a great experience from a customer perspective, but unknown or sensitive errors have to follow [MSFT's SDL requirements](https://liquid.microsoft.com/Web/Object/Read/MS.Security/Requirements/Microsoft.Security.SystemsADM.10202) (which I'll be the first to admit is unfortunate). 

If we're lacking logging internally for error messages, I think it would be good to ensure we're logging appropriately so we can debug any issues that come our way.

PR #2131, opened 2025-07-10
Add location details for all internal errors.

deads2k:
### What

This allows us to determine which particular area had an internal are without risking internal details of the service like IPs, service identities, etc being leaked via unvetted library error messages.

### Why

We need to be able to debug  our service when something fails.  Without this there are numerous different spots to have internal errors for individual actions and we have no way to track which is causing an issue.

/assign @mbarnes 

replaces https://github.com/Azure/ARO-HCP/pull/2118


mbarnes (commented):
Changes look good to me on a technical level.  There may still be an open discussion about whether we _should_ do this, so I'll withhold approval for the moment and let others chime in.

mociarain:
Bringing this one back up and it seems more important . I'm happy to make this change now and I brought it up to date. 

Note: In the last commit I switched to using the "CS" abbreviation for "ClusterService" in the constant names to help readability. This is only done for CS<Object> as this is an existing convention we have .

geoberle:
closing this. @deads2k if you still need this, please reopen

PR #1956, opened 2025-06-23
ARO-17962: Provision shard generation with cx_secrets_key_vault_managed_identity_client_id

jubittajohn:
This PR has the changes for adding a new shard config property for `cx_secrets_key_vault_managed_identity_client_id`:
The changes were tested the following way:
1. Ran the following make target to generate the provision_shards config `make --directory=cluster-service local-deploy-provision-shard`:
```
provision_shards:
- id: 1
  maestro_config: |
    {
      "rest_api_config": {
        "url": "http://localhost:8001"
      },
      "grpc_api_config": {
        "url": "localhost:8090"
      },
      "consumer_name": "hcp-underlay-usw3jujo-mgmt-1"
    }
  status: active
  management_cluster_id: local-cluster
  region: 'westus3'
  cloud_provider: azure
  topology: dedicated
  azure_shard:
    public_dns_zone_resource_id: "/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/resourceGroups/hcp-underlay-pers-usw3jujo/providers/Microsoft.Network/dnszones/usw3jujo.hcp.osadev.cloud"
    cx_secrets_key_vault_url: "https://ah-pers-cx-usw3jujo-1.vault.azure.net/"
    cx_managed_identities_key_vault_url: "https://ah-pers-mi-usw3jujo-1.vault.azure.net/"
    cx_secrets_key_vault_managed_identity_client_id: "f52765c3-c50a-42d1-85ca-acd9c1adf243"
```

2. Ran the following make command to deploy the CS `make cluster-service.deploy_pipeline`:

```
> oc get secrets -n clusters-service provision-shards -oyaml | yq '.data.config' | base64 -d
provision_shards:
- id: 1
  maestro_config: |
    {
      "rest_api_config": {
        "url": "http://maestro.maestro.svc.cluster.local:8000"
      },
      "grpc_api_config": {
        "url": "maestro-grpc.maestro.svc.cluster.local:8090"
      },
      "consumer_name": "hcp-underlay-usw3jujo-mgmt-1"
    }
  status: active
  management_cluster_id: local-cluster
  region: 'westus3'
  cloud_provider: azure
  topology: dedicated
  azure_shard:
    public_dns_zone_resource_id: "/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/resourceGroups/hcp-underlay-pers-usw3jujo/providers/Microsoft.Network/dnszones/usw3jujo.hcp.osadev.cloud"
[...]

PR #2132, opened 2025-07-10
add mgmt-output step to resolve azure keyvault secrets provider clientId

tony-schndr:
Follow up to https://github.com/Azure/ARO-HCP/pull/1956
### What
Fetch the azure keyvaults secrets provider client id from an output stage so that clusters service deployments work in multi subscription environments.

### Why

to fix clusters service deployment to environments with multiple subscriptions.  az commands in shell steps don't work nicely if the resource lives in a separate subscription.


jfchevrette (approved):
LGTM

PR #2133, opened 2025-07-10
use svcParentZoneName for oidc

tony-schndr:
### What

oic should be based on the svcParentZoneName


cluster-service/pipeline.yaml
```diff
@@ -78,8 +78,8 @@ resourceGroups:
       configRef: miMockCertName
     - name: ARM_HELPER_CERT_NAME
       configRef: armHelperCertName
-    - name: CX_PARENT_DNS_ZONE_NAME
-      configRef: dns.cxParentZoneName
```
geoberle, line 82: keep the CX one, just add the SVC

geoberle (approved):

PR #2134, opened 2025-07-10
EV2 signaling env var name change

geoberle:
### What

SDP-pipelines recently introduced env var conflict avoidance and because of that, the env var to signal an EV2 run changed from `EV2` to `__zz_injected_EV2__`.


jfchevrette (approved):
❤️

PR #2138, opened 2025-07-10
Support kube-API IP Access Control

hbhushan3:
Support for kube-API IP Access Control has been successfully implemented in the ARO-HCP project through the addition of an `authorizedCidrs` field. This field allows users to specify a list of authorized IPv4 CIDR blocks that can access the API server.

Waiting for merge of https://github.com/openshift-online/ocm-api-model/pull/1059/files#diff-364d063fef766917b654a76528b2154563eed08f9d346c352e6fd1eba8f5fd34

Ticket - https://issues.redhat.com/browse/ARO-15672

### DIFF
--- a/frontend/pkg/frontend/ocm.go
+++ b/frontend/pkg/frontend/ocm.go
@@ -81,6 +81,14 @@ func convertOutboundTypeRPToCS(outboundTypeRP api.OutboundType) (outboundTypeCS 
 	return
 }
 
+// convertAuthorizedCidrs extracts authorized CIDRs from CS cluster API object
+func convertAuthorizedCidrs(clusterAPI *arohcpv1alpha1.ClusterAPI) []string {
+	if cidrs := clusterAPI.AllowedCIDRBlocks(); cidrs != nil {
+		return cidrs
+	}
+	return nil
+}
+
 func convertEnableEncryptionAtHostToCSBuilder(in api.NodePoolPlatformProfile) *arohcpv1alpha1.AzureNodePoolEncryptionAtHostBuilder {
 	var state string
 
@@ -132,8 +140,9 @@ func ConvertCStoHCPOpenShiftCluster(resourceID *azcorearm.ResourceID, cluster *a
 				URL: cluster.Console().URL(),
 			},
 			API: api.APIProfile{
-				URL:        cluster.API().URL(),
-				Visibility: convertListeningToVisibility(cluster.API().Listening()),
+				URL:             cluster.API().URL(),
+				Visibility:      convertListeningToVisibility(cluster.API().Listening()),
+				AuthorizedCidrs: convertAuthorizedCidrs(cluster.API()),
 			},
 			Platform: api.PlatformProfile{
 				ManagedResourceGroup:   cluster.Azure().ManagedResourceGroupName(),
@@ -250,7 +259,8 @@ func withImmutableAttributes(clusterBuilder *arohcpv1alpha1.ClusterBuilder, hcpC
 			MachineCIDR(hcpCluster.Properties.Network.MachineCIDR).
 			HostPrefix(int(hcpCluster.Properties.Network.HostPrefix))).
 		API(arohcpv1alpha1.NewClusterAPI().
-			Listening(convertVisibilityToListening(hcpCluster.Properties.API.Visibility)))
+			Listening(convertVisibilityToListening(hcpCluster.Properties.API.Visibility)).
+			AllowedCIDRBlocks(hcpCluster.Properties.API.AuthorizedCidrs))
 
 	azureBuilder := arohcpv1alpha1.NewAzure().
 		TenantID(tenantID).
--- a/demo/README.md
+++ b/demo/README.md
@@ -43,6 +43,7 @@ CLUSTER_NAME=abc ./03-create-cluster.sh
 
 Observe the cluster creation with `./query-cluster-rp.sh` until `properties.provisioningState` is (hopefully) `Succeeded`.
 `properties.api.url` holds the URL to the API server of the HCP.
+`properties.api.authorizedCidrs` (optional) can be used to restrict access to the API server to specific IPv4 CIDR blocks.
 
 See [Get the kubeconfig for an HCP](#get-the-kubeconfig-for-an-hcp) on how to get the kubeconfig for the HCP.
 
--- a/demo/cluster.tmpl.json
+++ b/demo/cluster.tmpl.json
@@ -15,7 +15,11 @@
     },
     "console": {},
     "api": {
-      "visibility": "Public"
+      "visibility": "Public",
+      "authorizedCidrs": [
+        "192.168.1.0/24",
+        "10.0.0.0/16"
+      ]
     },
     "platform": {
       "managedResourceGroup": "$managed-resource-group",
--- a/internal/api/hcpopenshiftcluster.go
+++ b/internal/api/hcpopenshiftcluster.go
@@ -75,8 +75,9 @@ type ConsoleProfile struct {
 
 // APIProfile represents a cluster API server configuration.
 type APIProfile struct {
-	URL        string     `json:"url,omitempty"        visibility:"read"`
-	Visibility Visibility `json:"visibility,omitempty" visibility:"read create" validate:"omitempty,enum_visibility"`
+	URL             string     `json:"url,omitempty"             visibility:"read"`
+	Visibility      Visibility `json:"visibility,omitempty"    visibility:"read create" validate:"omitempty,enum_visibility"`
+	AuthorizedCidrs []string   `json:"authorizedCidrs,omitempty" visibility:"read create update" validate:"max=500,dive,cidrv4"`
 }
 
 // PlatformProfile represents the Azure platform configuration.
--- a/internal/api/v20240610preview/register_test.go
+++ b/internal/api/v20240610preview/register_test.go
@@ -95,6 +95,7 @@ func TestClusterStructTagMap(t *testing.T) {
 		"Properties.API":                                                     skip,
 		"Properties.API.URL":                                                 skip,
 		"Properties.API.Visibility":                                          api.VisibilityRead | api.VisibilityCreate,
+		"Properties.API.AuthorizedCidrs":                                     api.VisibilityRead | api.VisibilityCreate | api.VisibilityUpdate,
 		"Properties.Platform":                                                skip,
 		"Properties.Platform.ManagedResourceGroup":                           api.VisibilityRead | api.VisibilityCreate,
 		"Properties.Platform.SubnetID":                                       api.VisibilityRead | api.VisibilityCreate,


frontend/pkg/frontend/ocm.go
```diff
@@ -52,6 +52,8 @@ const (
 	csOutboundType                      string = "load_balancer"
 	csUsernameClaimPrefixPolicyNoPrefix string = "NoPrefix"
 	csUsernameClaimPrefixPolicyPrefix   string = "Prefix"
+	csCIDRBlockAllowAccessModeAllowAll  string = "allow_all"
```
mbarnes, line 55: I don't see `AllowAll` being used.  Is this implied when setting `API.CIDRBlockAccess(nil)`?

rajdeepc2792: Pushed the change to use `AllowAll` in case of nil/empty array.

mbarnes (dismissed):
I like the follow-up change, that's clearer.
I don't know if we're ready to merge yet but I'll approve so you're not blocked.

rajdeepc2792:
Blocking the PR until the dependent CS feature implementation is completed: https://redhat-external.slack.com/archives/C075PHEFZKQ/p1757528386532159?thread_ts=1756387629.670409&cid=C075PHEFZKQ

After CS changes are merged, this PR can be tested for the same.

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -0,0 +1,466 @@
...
+					currentCluster.HcpOpenShiftCluster,
+					nil,
+				)
+				Expect(err).NotTo(HaveOccurred())
+
+				_, err = poller.PollUntilDone(ctx, nil)
+				Expect(err).NotTo(HaveOccurred())
+
+				// Wait a bit for the change to propagate
+				time.Sleep(30 * time.Second)
```
SudoBrendan: Let's poll until a timeout instead of `sleep`, which introduces race conditions.

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -0,0 +1,466 @@
...
+
+				// Test connectivity using VM run command
+				connectivityTest := fmt.Sprintf("curl -k -s -o /dev/null -w '%%{http_code}' --connect-timeout 10 %s/healthz", apiURL)
+				output, err := runVMCommand(ctx, tc, *resourceGroup.Name, vmName, connectivityTest)
+				Expect(err).NotTo(HaveOccurred())
+
+				// Should get HTTP response (likely 401 or 200, but not connection refused)
+				httpCode := strings.TrimSpace(output)
+				By(fmt.Sprintf("VM received HTTP status code: %s", httpCode))
+				Expect(httpCode).To(MatchRegexp("^[2-5][0-9][0-9]$"), "Should receive valid HTTP status code from authorized IP")
```
SudoBrendan: wrap in Eventually for robustness?

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -0,0 +1,466 @@
...
+	privateKeyPEM := &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(privateKeyData),
+	}
+	privateKeyStr := string(pem.EncodeToMemory(privateKeyPEM))
+
+	// Generate public key in SSH format
+	pub, err := ssh.NewPublicKey(&privateKeyData.PublicKey)
+	if err != nil {
+		return "", "", err
```
SudoBrendan: nit: should we wrap this error with some context like "failed to generate ssh key" or something?

test/e2e/cluster_authorized_cidrs_connectivity.go
mbarnes: I think the intent is to put the below check in a retry loop that sleeps, say, 5 seconds between tries and retries up to 6 times.  There's your 30 seconds.  Or adjust these values to whatever you think is reasonable before giving up.

test/e2e-setup/bicep/modules/cluster.bicep
```diff
@@ -95,6 +98,7 @@ resource hcp 'Microsoft.RedHatOpenShift/hcpOpenShiftClusters@2024-06-10-preview'
     }
     api: {
       visibility: 'Public'
+      authorizedCidrs: authorizedCidrs
```
mbukatov: What would an empty list as a value of `authorizedCidrs` mean? Would it basically block all traffic to the API, disable the feature or report an error?

I'm asking because I'm little confused, as this almost looks as if an empty list here disables the feature (so that access is possible from all IP addresses), which is not what would I expect.

test/e2e-setup/bicep/modules/test-vm.bicep
```diff
@@ -0,0 +1,148 @@
...
+  parent: vm
+  location: location
+  properties: {
+    publisher: 'Microsoft.Azure.Extensions'
+    type: 'CustomScript'
+    typeHandlerVersion: '2.1'
+    autoUpgradeMinorVersion: true
+    settings: {}
+    protectedSettings: {
+      commandToExecute: 'sh -c "export DEBIAN_FRONTEND=noninteractive && rm -rf /var/lib/apt/lists/* && apt-get update && apt-get install -y curl wget && curl -LO https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl && install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl"'
```
mbukatov: Could we use RHEL image with openshift build of kubectl to avoid relying on binary artefacts from dl.k8s.io? If there is no image let's just prepare one and host it on quay and/or in our acr. This will reduce risks of flakynes caused by issues of external infra.

internal/ocm/convert_test.go
```diff
@@ -566,3 +567,147 @@ func TestBuildCSExternalAuth(t *testing.T) {
...
+			name:     "CREATE - sets CIDRBlockAccess with empty AuthorizedCIDRs",
+			updating: false,
+			hcpCluster: &api.HCPOpenShiftCluster{
+				CustomerProperties: api.HCPOpenShiftClusterCustomerProperties{
+					API: api.CustomerAPIProfile{
+						AuthorizedCIDRs: []string{},
+					},
+				},
+			},
+			expectedCSCluster: getBaseCSClusterBuilder(false),
```
mbukatov: Isn't an empty list of authorized CIDRS basically a request for blocking everyone (until this is updated with CIDR later)?

internal/ocm/convert_test.go
```diff
@@ -566,3 +567,147 @@ func TestBuildCSExternalAuth(t *testing.T) {
...
+	return builder.
+		NodeDrainGracePeriod(arohcpv1alpha1.NewValue().
+			Unit(csNodeDrainGracePeriodUnit).
+			Value(float64(0))).
+		API(clusterAPIBuilder.CIDRBlockAccess(arohcpv1alpha1.NewCIDRBlockAccess().
+			Allow(arohcpv1alpha1.NewCIDRBlockAllowAccess().
+				Mode(csCIDRBlockAllowAccessModeAllowAll))))
+}
+
+func TestBuildCSCluster(t *testing.T) {
```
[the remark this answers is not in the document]
mbukatov: Is there any IP address range we should not allow (such as the example/doc or local one)?

mbukatov (commented):
Providing some comments related to E2E testing.

internal/ocm/convert_test.go
nimrodshn: > Isn't an empty list of authorized CIDRS basically a request for blocking everyone

No. the opposite - an empty list is "allow-all" by default. (similar to current behavior).

nimrodshn:
> Is there any IP address range we should not allow (such as the example/doc or local one)?

@mbukatov See https://docs.google.com/document/d/1WHeFrgQ1wzsO5qqMQEM5cPfOjHKxlKDTBVsCrl_hM-U/edit?pli=1&tab=t.0

nonlocal-e2e-specs.txt
```diff
@@ -20,5 +23,8 @@
   "should be able to test admin credentials before cluster ready, then full admin credential lifecycle",
   "should confirm the HCP cluster is deleted (not found)",
   "should create an HCP cluster and validate TLS certificates",
-  "should not be able to create a 4.18 HCP cluster"
+  "should not be able to create a 4.18 HCP cluster",
+  "should reject cluster creation with IPv6 CIDR",
+  "should reject cluster creation with invalid CIDR format",
+  "should reject cluster creation with too many CIDRs"
```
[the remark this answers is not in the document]
nimrodshn: Oh, according to @venkateshsredhat I need to add these here. Is that wrong?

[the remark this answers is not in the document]
nimrodshn: OK thanks for the clarification.

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -0,0 +1,255 @@
...
+					},
+					TestArtifactsFS,
+				)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("generating SSH key pair for VM")
+				sshPublicKey, _, err := framework.GenerateSSHKeyPair()
+				Expect(err).NotTo(HaveOccurred())
+
+				By("deploying test VM")
```
stevekuznetsov: Do this in parallel with the cluster creation.

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -0,0 +1,255 @@
...
+					30*time.Minute,
+				)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("extracting VM public IP from deployment outputs")
+				vmPublicIP, err := framework.GetOutputValueString(vmDeployment, "publicIP")
+				Expect(err).NotTo(HaveOccurred())
+				Expect(vmPublicIP).NotTo(BeEmpty(), "VM public IP should be in deployment outputs")
+
+				By("setting authorized CIDRs with VM IP")
```
stevekuznetsov: Any way to not make this dependent on the VM already running? Does it need a public IP? Any network magic to use something internal to this resource group?

test/e2e/cluster_create_authorized_cidrs.go
```diff
@@ -0,0 +1,308 @@
...
+					clusterName,
+					cluster,
+					nil,
+				)
+				Expect(err).To(HaveOccurred())
+				Expect(strings.ToLower(err.Error())).To(ContainSubstring("invalid cidr"))
+			},
+		)
+
+		It("should reject cluster creation with IPv6 CIDR",
```
stevekuznetsov: ditto

nonlocal-e2e-specs.txt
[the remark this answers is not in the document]
stevekuznetsov: This PR cannot merge until the tests are updated to support local API.

test/e2e/cluster_authorized_cidrs_connectivity.go
nimrodshn: That wont work since we are pinging the api server the request passes through the internet. If the traffic were to be strictly inside the vnet or subnet than once can consider some network magic.

What is tested here is the likely scenario the user would do.

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -0,0 +1,255 @@
...
+				var httpCode string
+				Eventually(func(g Gomega) {
+					output, err := framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, connectivityTest)
+					g.Expect(err).NotTo(HaveOccurred())
+
+					// Should get HTTP response (likely 401 or 200, but not connection refused)
+					httpCode = strings.TrimSpace(output)
+					By(fmt.Sprintf("VM received HTTP status code: %s", httpCode))
+					g.Expect(httpCode).To(MatchRegexp("^[2-5][0-9][0-9]$"), "Should receive valid HTTP status code from authorized IP")
+				}, 2*time.Minute, 10*time.Second).Should(Succeed())
```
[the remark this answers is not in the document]
nimrodshn: From what I understand, this function runs every 10 seconds up to a timeout of 2min (See [here](https://onsi.github.io/gomega/#eventually), for more details).

test/e2e/cluster_create_authorized_cidrs.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+
+	hcpsdk "github.com/Azure/ARO-HCP/test/sdk/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+)
+
+var _ = Describe("Authorized CIDRs", func() {
+	Context("Cluster Creation", func() {
+		It("should be able to create a HCP cluster with valid authorized CIDRs",
```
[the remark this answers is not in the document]
nimrodshn: I was asked to add these here, I can remove these.

test/e2e/cluster_create_authorized_cidrs.go
```diff
@@ -0,0 +1,308 @@
...
+				for i, cidr := range cluster.Properties.API.AuthorizedCIDRs {
+					if cidr != nil {
+						cidrs[i] = *cidr
+					}
+				}
+				Expect(cidrs).To(ConsistOf("10.0.0.0/16", "192.168.1.0/24"))
+			},
+		)
+
+		It("should be able to create a HCP cluster with empty authorized CIDRs",
```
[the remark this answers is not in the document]
nimrodshn: Agreed. I was asked to add them. I can remove it.

internal/validation/validate_cluster.go
```diff
@@ -485,6 +485,7 @@ func validateCustomerAPIProfile(ctx context.Context, op operation.Operation, fld
 
 	// AuthorizedCIDRs []string   `json:"authorizedCidrs,omitempty"`
 	errs = append(errs, MaxItems(ctx, op, fldPath.Child("authorizedCidrs"), newObj.AuthorizedCIDRs, nil, 500)...)
+	errs = append(errs, MinItems(ctx, op, fldPath.Child("authorizedCidrs"), newObj.AuthorizedCIDRs, nil, 1)...)
```
mbarnes: This is preventing an empty `authorizedCidrs` list, correct?

internal/ocm/convert.go
```diff
@@ -335,6 +348,21 @@ func convertEtcdRPToCS(in api.EtcdProfile) (*arohcpv1alpha1.AzureEtcdEncryptionB
...
+func convertCIDRBlockAllowAccessRPToCS(in api.CustomerAPIProfile) (*arohcpv1alpha1.CIDRBlockAccessBuilder, error) {
+	cidrBlockAllowAccess := arohcpv1alpha1.NewCIDRBlockAllowAccess()
+
+	if in.AuthorizedCIDRs == nil {
+		cidrBlockAllowAccess.Mode(csCIDRBlockAllowAccessModeAllowAll)
+	} else if len(in.AuthorizedCIDRs) > 0 {
+		cidrBlockAllowAccess.Mode(csCIDRBlockAllowAccessModeAllowList)
+		cidrBlockAllowAccess.Values(in.AuthorizedCIDRs...)
+	} else {
+		return nil, fmt.Errorf("AuthorizedCIDRs cannot be an empty list")
```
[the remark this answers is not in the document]
rajdeepc2792: Comment added.

test/e2e/cluster_authorized_cidrs_connectivity.go
tsatam: Does that prevent the creation of the cluster itself and VM ARM resources being parallelized? As far as I understand the VM deployment, its only dependencies are on the vnet/subnet (which should be put down by CreateClusterCustomerResources). I could very well be off-base or missing something, though.

tsatam: correction - as of now the cluster creation depends on the VM (not the reverse), as the VM Bicep template deploys the publicIpAddresses resource that we need to reference in both VM and cluster creation. 

A follow-up fix could be to split this IP address resource into its own deployment, but I don't think we need to block on that, unless we already are observing this VM creation taking the full ~30 minutes we timeout on.

nimrodshn: > Does that prevent the creation of the cluster itself and VM ARM resources being parallelized?

Yes. The  underlying fw infrastructure will block the traffic otherwise.

tsatam (approved):
LGTM. There's an ongoing improvement opportunity in this (very expensive) E2E test to reduce its time cost, but if we're not seeing this being a severe-enough issue already, this can be a follow-up improvement.

PR #2157, opened 2025-07-14
fix: do not set CS MultiAZ attribute during cluster creation

miguelsorianod:
With recent changes, the Multi AZ attribute in CS is readonly for the ARO-HCP offering and it cannot be provided during cluster creation anymore. The Control Plane Multi AZness is determined depending on whether the region where it is deployed supports multiple AZs or not.

This decision was reflected in https://issues.redhat.com/browse/ARO-17289, which was approved by api domain owners in CS too.

### DIFF
--- a/frontend/pkg/frontend/ocm.go
+++ b/frontend/pkg/frontend/ocm.go
@@ -36,7 +36,6 @@ const (
 	csCloudProvider    string = "azure"
 	csProductId        string = "aro"
 	csHypershifEnabled bool   = true
-	csMultiAzEnabled   bool   = true
 	csCCSEnabled       bool   = true
 
 	// The OCM SDK does not provide these constants.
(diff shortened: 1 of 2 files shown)


mbarnes (approved):

bennerv (approved):
Makes sense to me.  The only question I think we need to clarify is our SLA will be lower for non-zonal HCPs.  

How do customers tell what their SLA is without an external field?  Will it always be tied to having zones within an Azure region, or will it depend on capacity within all zones, etc.?

miguelsorianod:
> Makes sense to me. The only question I think we need to clarify is our SLA will be lower for non-zonal HCPs.
> 
> How do customers tell what their SLA is without an external field? Will it always be tied to having zones within an Azure region, or will it depend on capacity within all zones, etc.?

My understanding is that it will be tied to whether an Azure region supports availability zones. If the region does not support availability zones then the control plane is not deployed in multiple AZs. If it does then the control plane is deployed in multiple AZs.

This means that when we deploy management clusters, we should make sure that the instance types chosen for the regions where multiple availability zones are supported, support multi AZ.

PR #2159, opened 2025-07-14
update demo/bicep scripts to specify 4.19.0 as the OCP version

miguelsorianod:
With recent changes we added support for OCP 4.19.0 and stopped supporting 4.18.x.


deads2k:
/lgtm
/hold

hold until the changes end up in our integration environment.  That should happen quickly. If for some reason it isn't happening quickly, please start a slack thread so we can drive out what is causing our deployment to be slow.

geoberle (approved):

PR #2168, opened 2025-07-15
chore: update global image sync related to 4.19 to be stable channel

miguelsorianod:
Now 4.19 GA images are released.

This MR:
* Renames candidate-4.19 channel to stable-4.19
* changes minVersion 4.19.0-rc.2 to 4.19.0


miguelsorianod:
This assumes that the existing GA images won't be removed. If that's not the case let me know because then it could have some impact

geoberle (approved):

PR #2175, opened 2025-07-15
*: add subscription provisioning config

stevekuznetsov:
*: add subscription provisioning config

---

config: add admin reader groups

---


stevekuznetsov:
Config-only change that I need for sdp-pipelines, going to send it in.

PR #2181, opened 2025-07-16
update acm policy helm chart

zhiweiyin318:
### What

Update the ACM Policy helm chart using the update-policy-chart.sh script, and update the image digest of policy from the 
acm-operator-bundle 2.13.1 image. 

### Why

1. The previous repo https://github.com/stolostron/mce-install-kube for ACM Policy helm chart cannot be updated timely, and the the manifests in the chart are synced from the https://github.com/stolostron/multiclusterhub-operator , so it will be more convenient and maintainable to update the policy helm chart using the script directly in this repo.
2. The RHACM version and MCE version have strict alignment, so we need to set the correct image digest for the MCE and Policy helm charts.


geoberle:
thank you. i will test once MCE 2.9 is avaiable

PR #2203, opened 2025-07-18
frontend: Cleanup routes

mbarnes:
### What

This cleans up some misleading code in `/frontend/pkg/frontend/routes.go`:

- Remove some misleading comments from early development.

- Split out `ArmResourceRead` endpoints into a separate group whose middleware list does NOT include `MiddlewareLockSubscription`.

- Similarly, give `ArmSubscriptionGet` its own middleware list that does NOT include `MiddlewareLockSubscription`.

- `MiddlewareLockSubscription` now _always_ applies where used.  Request method is no longer a factor.

### Why

These came to my attention when responding some recent questions from @deads2k.


frontend/pkg/frontend/middleware_locksubscription.go
```diff
@@ -40,12 +37,8 @@ func MiddlewareLockSubscription(w http.ResponseWriter, r *http.Request, next htt
...
 	subscriptionID := r.PathValue(PathSegmentSubscriptionID)
 
-	switch r.Method {
-	case http.MethodGet, http.MethodHead:
-		// These methods are read-only and don't require locking.
-	default:
-		lockClient = dbClient.GetLockClient()
-	}
+	// This may be nil when running "go test".
+	lockClient := dbClient.GetLockClient()
```
deads2k, line 41: future-us should make this panic on missing dbclient.

frontend/pkg/frontend/routes.go
```diff
@@ -102,17 +102,29 @@ func (f *Frontend) routes(r prometheus.Registerer) *MiddlewareMux {
...
+		MuxPattern(http.MethodGet, PatternSubscriptions, PatternProviders, PatternLocations, PatternVersions),
+		postMuxMiddleware.HandlerFunc(f.ArmResourceRead))
+
+	// Resource create/update/delete endpoints
+	postMuxMiddleware = NewMiddleware(
+		MiddlewareResourceID,
+		MiddlewareLoggingPostMux,
+		MiddlewareValidateAPIVersion,
+		MiddlewareLockSubscription,
+		MiddlewareValidateSubscriptionState)
```
deads2k, line 127: fwiw, I think future-us should find a factorization that makes it easier to reason about, "did I add the right stuff to this endpoint", but this is an improvement.

PR #2208, opened 2025-07-18
convert version from x.y to x.y.z

katherinelc321:
ARO-19982

### What

Validates that control plane version is x.y and converts it to x.y.x


internal/api/hcpopenshiftcluster.go
```diff
@@ -200,6 +201,20 @@ func (cluster *HCPOpenShiftCluster) validateVersion() []arm.CloudErrorBody {
 		})
 	}
 
+	versionID := cluster.Properties.Version.ID
+	if versionID != "" {
```
fahlmant, line 205: What happens if the version ID is blank? Is this a possibility?

internal/api/hcpopenshiftcluster.go
```diff
@@ -200,6 +201,20 @@ func (cluster *HCPOpenShiftCluster) validateVersion() []arm.CloudErrorBody {
...
+		// Check if version is in X.Y format and append ".0" to convert X.Y to X.Y.Z
+		if matched, _ := regexp.MatchString(`^\d+\.\d+$`, versionID); matched {
+			cluster.Properties.Version.ID = versionID + ".0"
+		} else {
```
fahlmant, line 209: Won't this trigger if the control plane version has been set to X.Y.Z previously?

katherinelc321: for the august release, we only want the customer to be able to specify x.y for the control planes version. We'll have CS implement x.y later but for now, cs can only take x.y.z. This is a temporary solution

frontend/pkg/frontend/ocm.go
```diff
@@ -470,6 +474,13 @@ func (f *Frontend) BuildCSNodePool(ctx context.Context, nodePool *api.HCPOpenShi
 			Value(float64(*nodePool.Properties.NodeDrainTimeoutMinutes)))
 	}
 
+	nodepoolVersion := nodePool.Properties.Version.ID
+	nodepoolCSversion, err := ocm.ConvertOpenshiftVersionAddPrefix(nodepoolVersion)
+	if err != nil {
+		return nil, err
+	}
+	nodePool.Properties.Version.ID = nodepoolCSversion
```
mbarnes, line 482: This is happening too late; the version is already set in `npBuilder` by this point.
This code needs to happen before the `arohcpv1alpha1.NewVersion()` call above.

mbarnes (changes requested):
Spotted a couple more bugs from testing.

mbarnes (approved):

PR #2212, opened 2025-07-18
update aro-hcp-e2e to accept LOCATION env var

deads2k:
This allows different test invocations to select where to create stuff. This does prevent us from installing to multiple different locations for cross location tests.  We'll cross that bridge later.

Preferably, we land https://github.com/openshift/release/pull/67234 first to pass the env var and have zero downtime.

/assign @mbukatov


mbukatov (dismissed):

deads2k:
I opened with the wrong branch :(

closing for https://github.com/Azure/ARO-HCP/pull/2227

PR #2215, opened 2025-07-18
config: move to new provider and feature registration step

stevekuznetsov:


stevekuznetsov:
Seems to work as expected, need to pull this in to unblock myself. If I break things, please forgive me.

PR #2224, opened 2025-07-21
Bump ocm-sdk-go to v0.1.472

mociarain:
This bumps the OCM-sdk since there was a bug in `v0.1.471`


tooling/prometheus-rules/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.24.3
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/alertsmanagement/armalertsmanagement v0.10.0
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.82.2
-	github.com/prometheus/common v0.64.0
+	github.com/prometheus/common v0.65.0
```
mociarain, line 8: This resulted from running `make all-tidy` and seems a little odd to me but not dangerous. Just calling it out

PR #2230, opened 2025-07-21
Cluster lifecycle test bicep - WIP

mgahagan73:
https://issues.redhat.com/browse/ARO-18147

### What

cluster lifecycle test using bicep template

### Why

Experimental lifecycle test using bicep template.

Work in progress.


test/e2e/cluster_lifecycle_test.go
```diff
@@ -0,0 +1,261 @@
...
+			"persistTagValue": "false",
+		}
+
+		// Use the framework's CreateBicepTemplateAndWait function for consistency
+		deploymentResult, err := framework.CreateBicepTemplateAndWait(
+			ctx,
+			armResourcesClientFactory.NewDeploymentsClient(),
+			resourceGroup,
+			deploymentName,
+			bicepContent,
```
deads2k, line 120: this part is a little tricky.  we can't use bicep directly, it  needs to be compiled first. Try converting in this file: https://github.com/Azure/ARO-HCP/blob/aa885890cfff7d2c717ddb1d964526659256e0d8/test/hack/update-bicep-json.sh#L55

You choose a source bicep file and a destination json file.  If you place your destination in `${OUTPUT_DIR}/<subdir>/foo.json`, then you can access it via the embedded filesystem like this: `framework.Must(TestArtifactsFS.ReadFile("test-artifacts/generated-test-artifacts/standard-cluster-create/customer-infra.json")),`

test/e2e/cluster_lifecycle_test.go
```diff
@@ -0,0 +1,261 @@
...
+		// Get deployment outputs to retrieve managed identity information
+		deploymentOutputs := deploymentResult.Properties.Outputs
+		Expect(deploymentOutputs).NotTo(BeNil(), "deployment outputs should not be nil")
+
+		// Get deployment outputs to retrieve managed identity information
+		outputs := deploymentOutputs.(map[string]interface{})
+
+		// Extract identityValue and userAssignedIdentitiesValue from outputs
+		identityValue := outputs["identityValue"].(map[string]interface{})
+		userAssignedIdentitiesValue := outputs["userAssignedIdentitiesValue"].(map[string]interface{})
```
deads2k, line 135: try using this function here: https://github.com/Azure/ARO-HCP/blob/f68fa8de08934df1e05b4e18dce16a90ea27216a/test/util/framework/deployment_helper.go#L46-L49

PR #2231, opened 2025-07-21
Add `platform.loadbalancer` to API spec

tsatam:
[ARO-17691](https://issues.redhat.com/browse/ARO-17691)

### What

Adds a new `platform.loadBalancer` property to the hcpOpenShiftClusters resource. This currently contains a `managedOutboundIps.count` property, modifable by the customer on creation or updates to set the number of outbound IP addresses on the cluster's public load balancer, as well as an `effectiveOutboundIps` array containing resourceID references to the actual `publicIpAddresses` resources backing that count. 

### Why

Needed in order to support user requests to scale the number of outbound IP addresses on a cluster's load balancer, which is in-turn required to scale the number of nodes on a cluster (60 nodes per outbound IP) 

As implemented in the spec, the entire `platform.loadBalancer` section is optional on cluster creation. Any missing values here across the chain of nested properties will assume the defined default of `platform.loadBalancer.managedOutboundIps.count = 1` (to be implemented within the frontend later).


mociarain (approved):
LGTM

miguelsorianod:
I think the design needs might be refined a bit more.

This adds a new .platform.loadbalancer section to the api spec. However, at the same time we have .platform.outboundType which was the first attribute in preparation for this, and related to this. It might be a bit strange to have them in parallel.

On the CS API side, we defined and implemented the API in the following way, having in mind this upcoming work some time ago:

```
nodes_outbound_connectivity: {
  outbound_type: load_balancer //currently the only supported type but others would be added
}
```

Then, when finally, the configuration associated to the LB is to be added, which is going to happen soon as we can see in the MR, we would add:

```
nodes_outbound_connectivity: {
  outbound_type: load_balancer //currently the only supported type but others would be added
  load_balancer: { // this one is mandatory when outbound_type is load_balancer
    //new config associated to lb
  }
}
```

With this design:
* We have space for other outbound type other than lbs
* We have a centralized configuration for the lb one, and not having mixed concerns in the parent section about different functionalities
* We have space for adding per outbound type specific configuration, and that can grow over time too
* We are explicit with what's being set
* We do not mix different functionalities concerns in the same layer/section

Can we consider if something similar might make sense on the ARM api side?

api/redhatopenshift/HcpCluster.Management/examples/2024-06-10-preview/HcpOpenShiftClusters_CreateOrUpdate_MaximumSet_Gen.json
```diff
@@ -33,6 +33,11 @@
           "managedResourceGroup": "nhyhywrxupo",
           "subnetId": "/subscriptions/FDEA43EA-0230-4A7D-BDEE-F3AFF2183B1D/resourceGroups/resourceGroupName/providers/Microsoft.Network/virtualNetworks/hcp-network-example/subnets/example-subnet",
           "outboundType": "LoadBalancer",
+          "loadBalancer": {
+            "managedOutboundIps": {
```
miguelsorianod, line 37: In the future, other than specifying the number of desired IPs, do we foresee other mechanisms of configuring them? I am asking because back then it wasn't clear what were the requirements around this and how do we see that evolving.

Is the need/possibility of the end-users providing the specific IPs they want to leverage at some point? If there's a chance for that to happen it would be a good idea to design around expansion to it.  See my other comment which talks about the api design in general, because in that case something similar here might apply.

internal/api/v20240610preview/generated/models.go
```diff
@@ -559,6 +559,33 @@ type Label struct {
 	Value *string
 }
 
+// LoadBalancerProfile - A representation of the cluster's public load balancer.
+type LoadBalancerProfile struct {
+	// REQUIRED; The desired managed outbound IPs for the cluster's public load balancer
+	ManagedOutboundIPs *ManagedOutboundIPs
+
+	// READ-ONLY; The list of effective outbound IP addresses of the cluster's public load balancer
+	EffectiveOutboundIPs []*string
```
miguelsorianod, line 568: Why this name "EffectiveOutboundIPs" the "effective" part seems a bit odd. What about ProvisionedOutboundIPs.

Why is this at this level? Maybe it would make sense in ManagedOutboundIPs. See my other comments about API design.

mbarnes (changes requested):
Small nit about a type name.  Otherwise the design seems reasonable to me; I trust we're following Classic here.

PR #2237, opened 2025-07-22
Support for adding Y stream version upgrades.

Nanyte25:
### What 
Rest Spec Update for Y stream version Upgrade https://issues.redhat.com/browse/ARO-17745


Nanyte25:
@mociarain Can you tell what might the issue be here with linting ?

kaushalm1:
@Nanyte25   Do add the openapi example too ! ty !!

Nanyte25:
@mociarain this is ready now to be merged.

mociarain (commented):
So, the changes in the models look good but all the autorest versions changes shouldn't be in here. I'm not very familiar with go package management but I don't understand why we don't pin to a specific version to avoid all the constant noise we get in these PRs.

Is there a good reason we don't pin the autorest version?

If not, I suggest we merge this as is, and then do a subsequent PR where we pin it and re-run the generation.

@mbarnes @tsatam @venkateshsredhat @SudoBrendan @bennerv

PR #2240, opened 2025-07-23
geneva config refactor

weinong:
### What

* consolidate geneva config per account and namespace
* use placeholder values

### Why

* the config belongs to account and namespace. it doesn't belong to the cluster placement as we may have svc logs from mgmt cluster, too!

### DIFF
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -51,6 +51,41 @@
         "name"
       ]
     },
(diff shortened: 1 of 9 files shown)


janboll:
I refactored this into: https://github.com/Azure/ARO-HCP/pull/2241

PR #2221, opened 2025-07-21
Add geneva authentication certificate

janboll:
### What

- Add certificate for geneva authentication
- Refactor maestro certificate to not copy and paste same code everywhere

This will result in the following SAN:
**RP Logs:**
- **mgmt**: mgmt.rplogs.hcpsvc.osadev.cloud
- **svc** svc.rplogs.hcpsvc.osadev.cloud

**ClusterLogs**
- **mgmt** mgmt.clusterlogs.hcpsvc.osadev.cloud

PR #2241, opened 2025-07-23
Geneva certificate management

janboll:
### What

Update schema and bicep files to create certificates for Geneva log authentication. 

Reuse work from these PRs:
- https://github.com/Azure/ARO-HCP/pull/2221
- https://github.com/Azure/ARO-HCP/pull/2240

### Why

Need to refactor keyvault certificate createion used in Maestro, cause it can be used here. Need to refactor schema, cause we need to make it more flexible since we will use RP account access from management cluster as well.


geoberle (approved):

tony-schndr (approved):

geoberle (dismissed):

weinong (dismissed):
LGTM

PR #2244, opened 2025-07-23
Redesign OS Disk profile API Spec

venkateshsredhat:
As there is some Redesign of CS Node Pools API to move os disk related configuration under a single section  . Updating the Frontend API spec to accommodate those changes .

Looks like we dont have the ephemeral_os_disk_enabled defined earlier hence adding the new attribute "persistence" to the API spec to allow end user to pick .

https://issues.redhat.com/browse/XCMSTRAT-1289 

CS API design work : https://issues.redhat.com/browse/ARO-19918 .

Frontend changes : https://issues.redhat.com/browse/ARO-19899 .


venkateshsredhat:
Moved to https://github.com/Azure/ARO-HCP/pull/2352 .

PR #2250, opened 2025-07-23
postgres breakglass

geoberle:
### What

provide breakglass for CS and maestro postgres

https://issues.redhat.com/browse/ARO-20082


maestro/server/deploy/templates/postgres-breakglass.deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+          echo "Getting PostgreSQL access token..."
+          ACCESS_TOKEN=$(az account get-access-token --resource-type oss-rdbms -o json | jq .accessToken -r)
+
+          if [ -z "$ACCESS_TOKEN" ] || [ "$ACCESS_TOKEN" = "null" ]; then
+            echo "Failed to get access token"
+            exit 1
+          fi
+
+          echo "Writing access token and connect script to shared volume..."
```
jfchevrette, line 41: Suggestion:  use a block like you did in the container command script

maestro/server/deploy/templates/postgres-breakglass.deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - name: init
+        image: 'mcr.microsoft.com/azure-cli:cbl-mariner2.0'
+        imagePullPolicy: IfNotPresent
+        command:
+        - /bin/sh
+        - -c
+        - |
+          echo "PostgreSQL breakglass init container started at $(date)"
+          echo "Logging into Azure using workload identity..."
+          az login --federated-token "$(cat $AZURE_FEDERATED_TOKEN_FILE)" --service-principal -u $AZURE_CLIENT_ID -t $AZURE_TENANT_ID
```
jfchevrette, line 31: Suggestion: read the token file into a variable prior to using it in the az command
Suggestion 2: ensure environment variables exist and are set prior to using them, ex
```bash
set -euo pipefail

: "${AZURE_CLIENT_ID:?Environment variable is not set}"
: "${AZURE_TENANT_ID:?Environment variable is not set}"
# ...
az login ...
```

maestro/server/deploy/templates/postgres-breakglass.deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+          echo "Getting PostgreSQL access token..."
+          ACCESS_TOKEN=$(az account get-access-token --resource-type oss-rdbms -o json | jq .accessToken -r)
+
+          if [ -z "$ACCESS_TOKEN" ] || [ "$ACCESS_TOKEN" = "null" ]; then
+            echo "Failed to get access token"
+            exit 1
+          fi
+
+          echo "Writing access token and connect script to shared volume..."
+          echo "$ACCESS_TOKEN" > /breakglass/postgres_token
```
jfchevrette, line 42: I don't see this file being used anywhere?

docs/ops/postgres-breakglass.md
```diff
@@ -0,0 +1,30 @@
...
+   Wait for the deployment to be ready
+1. Exec into the pod of the deployment
+
+   ```/bin/sh
+   kubectl exec -it postgres-breakglass-<pod-name> -n clusters-service -- /bin/bash
+   or
+   kubectl exec -it postgres-breakglass-<pod-name> -n maestro -- /bin/bash
+   ```
+
+2. Run `connect` within the container to start a `psql` session
```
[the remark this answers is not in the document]
geoberle: this token expires after an hour. existing psql sessions will still work but the token can't be used to establish new ones. such a pod can't be used for a long time, it needs to be recycled after a while when new sessions need to be established. i added a hint to the docs

maestro/server/deploy/templates/postgres-breakglass.deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+          mountPath: /tmp
+        - name: var-tmp
+          mountPath: /var/tmp
+        - name: breakglass
+          mountPath: /breakglass
+      volumes:
+      - name: tmp
+        emptyDir: {}
+      - name: var-tmp
+        emptyDir: {}
```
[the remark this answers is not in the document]
geoberle: the psql client can write tmp data to these paths so i thought i back it with an empty dir, but since these are one-shot short lived containers, it most likely does not make any sense. i removed it

maestro/server/deploy/templates/postgres-breakglass.deployment.yaml
geoberle: good spot. this was my first solution and i was unsure if i should keep it around for something convenient that might come up. not worth the effort. i removed it

maestro/server/deploy/templates/postgres-breakglass.deployment.yaml
geoberle: both done

janboll (approved):

PR #2251, opened 2025-07-23
feat: update cs image to 5a131d6 in int env

JameelB:
### What
Updates cs image to 5a131d6 in the int environment

### Why
Promote CS release 5a131d6 from integrated dev to int environment.

N/A


JameelB:
cc @deads2k, here's the PR to promote to int.

PR #2274, opened 2025-07-28
udpate demo cluster creation bicep module with MI usage

geoberle:
### What

in STAGE the MI permissions have an actual impact, hence we need to carefully craft these permissions.

this PR is the current minimal-"best-to-our-knowledge" permissions implementation for the MIs of an HCP

### DIFF
--- a/demo/bicep/cluster.bicep
+++ b/demo/bicep/cluster.bicep
@@ -1,62 +1,364 @@
+// TODO
+// * KMS identity + role assignments need to be created
+// * pass the KMS identity to the cluster
+
 @description('Name of the hypershift cluster')
 param clusterName string
 
 @description('The Hypershift cluster managed resource group name')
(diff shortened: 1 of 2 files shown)


demo/bicep/cluster.bicep
```diff
@@ -1,80 +1,464 @@
+// TODO
+// * KMS identity + role assignments need to be created
+// * pass the KMS identity to the cluster
+
```
deads2k, line 4: can we make each of these separate bicep templates so we can easily test good and bad scenarios?

tony-schndr (approved):
lgtm

demo/bicep/cluster.bicep
geoberle: we can address that in future changes. makes sense for sure

PR #2275, opened 2025-07-28
MSI Refresher - add msiKeyvaultUri to config schema

cadenmarchese:
### What

Adds a new parameter (msiKeyvaultUri) to config schema, and then run make materialize

### Why

The msi credential refresher needs to know what MSI keyvault to watch and subsequently refresh certs from. Long term, there will be an inventory service that it will read, which knows where each MSI keyvault lives. For now, the refresher just takes a --keyvault-uri flag on startup that contains the URI to the keyvault, so we need to pass that to helm via a config.

### DIFF
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -1039,6 +1039,10 @@
           "type": "string",
           "description": "The client ID of the first party application that will be used by the refresher"
         },
+        "msiKeyvaultUri": {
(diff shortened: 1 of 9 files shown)


cadenmarchese:
Closing this as we will construct the full URI using `msiKeyVault.name`.

PR #2278, opened 2025-07-29
New pipeline for RMO

roivaz:
https://issues.redhat.com/browse/ARO-19978
https://issues.redhat.com/browse/ARO-20117

### What

Adds a new pipeline to deploy the RMO (route-monitor-operator) in the management clusters. RMO provides monitoring of endpoints via the blackbox-exporter.
The PR also includes alerts to monitor Kubernetes API Server availability of customer Hosted Control Planes.

#### Implementation details

Currently, RMO does not offer native support for the ARO-HCP environment. Until native integration is implemented, several workarounds using ACM policies are deployed alongside RMO to ensure it creates the correct resources in the management cluster, thereby enabling monitoring for all Hosted Control Planes managed there. The ACM policy works like this:
* Looks for HostedControlPlane custom resources and creates a RouteMonitor resource for each
* RMO will create the blackbox-exporter resources and a ServiceMonitor (monitoring.rhobs) per HCP to gather enpoint availability metrics
* Prometheus in management clusters don't reconcile servicemonitors.monitoring.rhobs resources, so the policy watches for these and creates a corresponding servicemonitors.monitoring.coreos.com custom resource. This will be prooperly reconciled by the in-cluster prometheus-operator instance.

RMO Helm chart generation:
* RMO is distributed with OLM. There is a new make target in `route-monitor-operator/Makefile` to generate the Helm chart using `tooling/olm-bundle-repkg` (see https://github.com/Azure/ARO-HCP/pull/2225). This make target picks up the bundle version from the configuration in `config`. The chart generation must be manually triggered and the resulting chart commited to the ARO-HCP repo.

### Why

https://issues.redhat.com/browse/ARO-19708?page=com.docminer.jira.issue-links%3Acom.kintosoft.jira.links.tab-panel

"As a/an SRE I want  the availability metrics for each hosted API server endpoint to be available so that  SREs can view them and set up alerts for them."


route-monitor-operator/scaffold/kas-monitoring-hack.managedclustersetbinding.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: cluster.open-cluster-management.io/v1beta2
+kind: ManagedClusterSetBinding
+metadata:
+  name: default
+  namespace: openshift-route-monitor-operator
+spec:
+  clusterSet: default
```
geoberle, line 7: i already completely forgot how managedclustersets and their bindings work.
i just know we have another clusterset hypershift-management-clusters

you seem to know your way around this. can you elaborate this a bit?

roivaz: ACM uses this mechanism to impose restrictions on which users can manage which clusters using policies. In practice, when you bind a clusterset to a namespace, any user with permission to deploy policies in that namespace can apply policies to any cluster that belongs to that clusterset.

In this particular case, we are binding the "default" clusterset, which only has one member (the "local-cluster"), to the "openshift-route-monitor-operator" namespace. Effectively, any user with permissions to create/modify/delete policies in this namespace can affect the management cluster.

geoberle: ah i see. `default`== `local-cluster` `hypershift-management-clusters` == `HCPs`

geoberle (approved):

geoberle (approved):

PR #2301, opened 2025-07-30
manage OCP image cache also for non-dev environments

geoberle:
### What

in order to be able to use nightly OCP images in higher environments, we need to move from mirrored images to cached ones. this PR enables an image cache for the quay.io/openshift-release-dev repositories

the oc-mirrored images can still be used and components can move over to cached images one at a time, e.g. CS and HO

https://issues.redhat.com/browse/ARO-20177


dev-infrastructure/templates/global-acr.bicep
```diff
@@ -23,16 +27,55 @@ resource globalMSI 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31'
...
+        targetRepo: 'openshift-release-dev/*'
+        userIdentifier: 'quay-username'
+        passwordIdentifier: 'quay-password'
+        loginServer: 'quay.io'
+      }
+    ]
+    purgeJobs: [
+      {
+        name: 'openshift-release-dev-purge'
+        purgeFilter: 'quay.io/openshift-release-dev/.*:.*'
```
bennerv, line 63: Why quay.io? Shouldn't this just be `openshift-release-dev/.*:.*`?

dev-infrastructure/templates/global-acr.bicep
```diff
@@ -23,16 +27,55 @@ resource globalMSI 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31'
...
+      {
+        ruleName: 'openshiftReleaseDev'
+        sourceRepo: 'quay.io/openshift-release-dev/*'
+        targetRepo: 'openshift-release-dev/*'
+        userIdentifier: 'quay-username'
+        passwordIdentifier: 'quay-password'
+        loginServer: 'quay.io'
+      }
+    ]
+    purgeJobs: [
```
bennerv, line 60: We probably want to swap `purgeJobs` -> retention policy since these are untagged manifests now.  

https://learn.microsoft.com/en-us/azure/container-registry/container-registry-retention-policy

dev-infrastructure/templates/global-acr.bicep
```diff
@@ -23,16 +27,55 @@ resource globalMSI 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31'
...
+      {
+        name: 'openshift-release-dev-purge'
+        purgeFilter: 'quay.io/openshift-release-dev/.*:.*'
+        purgeAfter: '${ocpImagesPurgeAfterDays}d'
+        imagesToKeep: 100
```
bennerv, line 65: when you have `purgeAfter` and `imagesToKeep`, does that mean:
1. we will only purge images after the `purgeAfter` date?
2. we will purge images after the `purgeAfter` date but also only keep 100 images max
3. something else?

dev-infrastructure/templates/global-acr.bicep
geoberle: good question. i we had this code in use already for DEV but we never checked if it is actually doing something. i will check the docs

dev-infrastructure/templates/global-acr.bicep
geoberle: ack. i will change

dev-infrastructure/templates/global-acr.bicep
geoberle: 3) we will purge images after the purgeAfter date but also only keep 100 images *min*
but for OCP images we will switch to retention policy

dev-infrastructure/templates/global-acr.bicep
geoberle: this is now removed

dev-infrastructure/modules/acr/cache.bicep
```diff
@@ -0,0 +1,152 @@
...
+
+  @description('How long ago to purge images (e.g., "2d", "7d")')
+  purgeAfter: string
+
+  @description('Number of images to keep')
+  imagesToKeep: int
+}
+
+@description('Array of purge job configurations')
+param purgeJobs purgeJob[] = []
```
bennerv, line 48: Is this param still being passed now that you switched from purge -> retention?

geoberle: Purge jobs still exist but we don't use them. Not being passed

PR #2178, opened 2025-07-15
Expand test and helpers to make it easy-ish to write new tests

deads2k:
Builds on https://github.com/Azure/ARO-HCP/pull/2125 with additional infrastructure, helpers clean up and the like.

/hold

PR #2302, opened 2025-07-30
use default azure credentials to span multiple environments

deads2k:
Updates auth based on @stevekuznetsov's recommendation in https://github.com/Azure/ARO-HCP/pull/2178#discussion_r2208893410.  This should allow using managed credentials as discussed on the previous call.

@geoberle @bennerv


deads2k:
diff of diffs here: https://github.com/Azure/ARO-HCP/compare/92a66147f7031dee5016bf8d9b2106e02c4d483d..101879a55623fc696f85d864119ce691902dd0b9

PR #2304, opened 2025-07-30
update bicep template to correctly test improper versions

deads2k:
adjusting the same way that we did for the demo script so that the only failure is the banned version.


deads2k:
already merged

PR #2307, opened 2025-07-30
Image management enhancements

mbarnes:
### What

Some enhancements to make ARO-HCP frontend and backend images easier to manage:

- Add a `generate-tag.sh` script to generate an image tag.  This produces the same tag format we've been using but with one important difference: images created for personal development environments will now be tagged as `test-${USER}-${COMMIT}` so they can be easily distinguished from images that are candidates for wider deployment.

- Simplify image digest selection in `get-digest.sh`.  The image tag to match can now be specified directly.  When no image tag is provided, the fallback selection will skip personal development environment images (as described above).

- Add a frontend and backend Makefile goal named `config` that will update all environments in `/config/config.msft.clouds-overlay.yaml` with an appropriate image digest.


mbarnes:
The **ci-yaml / format** failure appears to have been caused by #2199.

tsatam (approved):

PR #2315, opened 2025-07-30
MCE operator bundle image as config reference

geoberle:
### What

having the MCE operate bundle image as configuration allows us to report on it


deads2k:
I'm not entirely clear on the process for us to update to a new level like 2.9.0. Does this change make that painful?  If not, let's do it. If it does make it painful, let's think about how to make it practical.

PR #2321, opened 2025-07-31
Add KeyVault permission to msi mock identity

patilsuraj767:
### What

We need to assign `Azure Red Hat OpenShift KMS Plugin - Dev` role to the MSI mock identity. This is required to enable HyperShift to interact with the Key Vault during the etcd encryption process. The permission is necessary because the MSI Data Plane service is not available in dev or personal environments.

### DIFF
--- a/dev-infrastructure/Makefile
+++ b/dev-infrastructure/Makefile
@@ -138,6 +138,13 @@ create-mock-identities:
 	SUBSCRIPTION_ID=$(shell az account show --query id --output tsv) \
 	./scripts/create-sp-for-rbac.sh
 
+	APPLICATION_NAME=aro-dev-msi-mock2 \
+	KEY_VAULT_NAME=aro-hcp-dev-svc-kv \
+	CERTIFICATE_NAME=msiMockCert2 \
+	ROLE_DEFINITION_NAME='Azure Red Hat OpenShift KMS Plugin - Dev' \
+	SUBSCRIPTION_ID=$(shell az account show --query id --output tsv) \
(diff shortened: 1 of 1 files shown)


dev-infrastructure/Makefile
```diff
@@ -138,6 +138,13 @@ create-mock-identities:
 	SUBSCRIPTION_ID=$(shell az account show --query id --output tsv) \
 	./scripts/create-sp-for-rbac.sh
 
+	APPLICATION_NAME=aro-dev-msi-mock2 \
+	KEY_VAULT_NAME=aro-hcp-dev-svc-kv \
```
janboll, line 142: Is this the right keyvault?

patilsuraj767: This has been copied from the previous role assignment section, where the same identity(`aro-dev-msi-mock2`) is being assigned a role.

patilsuraj767: I am trying to add role assignment, that is assigning role `Azure Red Hat OpenShift KMS Plugin - Dev` to identity `aro-dev-msi-mock2` at subscription level. 
Here keyvault is provided because `create-sp-for-rbac.sh` needs it to store app credential. 

I want to do this because `aro-dev-msi-mock2` is used by hypershift in personal dev environment to communicate with the keyvault(Not the keyvault which it provided here). I want this to test etcd encryption of hosted cluster. 

While we can proceed with a manual role assignment for now(To avoid credential reset), this change is intended to ensure it’s maintained through code going forward.

PR #2324, opened 2025-07-31
E2E Test Suite Fallback Setup and Teardown of cluster

patriksuba:
### What
Add Teardown Validation Test case to run after cluster deletion.
PR introduce Fallback setup to create a cluster and nodepool using bicep template. Fallback creates e2e-setup.json file for a given cluster and nodepool. Added logging of all steps.

### Why
To enable running E2E Test Suite without e2e-setup.json.

Updated Containerfile accordingly. More information in commit messages.


mbukatov (approved):
Thanks for implementing this!

test/e2e/e2e_test.go
```diff
@@ -34,3 +40,26 @@ var _ = BeforeSuite(func() {
...
+	if err != nil {
+		panic(fmt.Sprintf("Cluster deletion should succeed (begin): %v", err))
+	}
+	_, err = poller.PollUntilDone(ctxDel, &runtime.PollUntilDoneOptions{
+		Frequency: 10 * time.Second,
+	})
+	if err != nil {
+		panic(fmt.Sprintf("Cluster deletion should succeed (poll): %v", err))
+	}
+})
```
deads2k, line 65: if you wire creation through a helper, then deletion can be auto-registered with any number of resource groups.  See example here.

https://github.com/Azure/ARO-HCP/blob/main/test/util/framework/per_test_framework.go#L126  with usage here https://github.com/Azure/ARO-HCP/blob/main/test/e2e/complete_cluster_create.go#L52 

```go
			resourceGroup, err := ic.NewResourceGroup(ctx, "basic-create", "uksouth")
			Expect(err).NotTo(HaveOccurred())
```

That block creates a non-conflicting resource group and will auto-clean up when the test is over.

mbukatov (approved):

geoberle (approved):

PR #2330, opened 2025-07-31
add billing object to config schema

gouthamMN:
### What
Add `billing` object to the config schema.

### Why
This is needed to deploy and release billing infra and service.


jonathan34c (approved):
LGTM

PR #2344, opened 2025-08-01
reconfigure HO to use ACR cache

geoberle:
### What

oc-mirror and cached images differ in the used repos. this PR switches the HO over to use the cached ones.

https://issues.redhat.com/browse/ARO-20271


geoberle:
i've created a dedicated personal dev environment, and created a cluster with this.
in addition @jubittajohn was testing this change was well, even leveraging a custom CPO image override, which can only be consumed via the cached images (not the mirrored ones)

bennerv (approved):
lgtm

PR #2347, opened 2025-08-03
helm chart adopt resources

geoberle:
### What

if we move resources between helm-charts (especially non-namespaced ones) we need to have a way to adopt them by the new helm chart. helm is particularly picky when it finds a resources already exists on the cluster but does not have the proper helm ownership annotation.

this functionality can be activated on `hack/helm.sh` usage with the `HELM_ADOPT=true` env variable if needed for dedicated migration scenarios. it is not supposed to be activated permanently. 

concrete usecases:
* we will move the CRDs of MCE to a dedicated helm chart in order to be able to update them
* ingress is currently managed via the frontend chart, but when we bring in the admin api, we would like to move ingress to a dedicated chart


bennerv (approved):
lgtm

PR #2353, opened 2025-08-04
feat: add support for asynchronous cluster updates

ashishmax31:
https://issues.redhat.com/browse/ARO-19963

### What
Introduce a new arm provisions state "updating" non-terminal state when CS returns the state of the cluster resource as "updating".

### Why

This change is needed to support the async updates behaviour in the CS update clusters endpoint.


venkateshsredhat:
LGTM

mociarain (approved):
LGTM

PR #2354, opened 2025-08-04
ARO-20009: update HO in CSPR env to contain Azure breaking changes

miguelsorianod:
Hypershift Operator has introduced API breaking changes that require updating both CS and Hypershift operator to have matching client and server side changes.
This commit updates the Hypershift Operator version in the CSPR environment to contain those new changes that are breaking changes.


miguelsorianod:
The SHA might be different, double checking

miguelsorianod:
> The SHA might be different, double checking

It was correct.

JameelB (approved):

PR #2360, opened 2025-08-04
ARO-20009: update HO and CS images to sync breaking changes

miguelsorianod:
Update Hypershift Operator image and Clusters Service image in the configuration of ARO-HCP environments

Hypershift Operator has introduced API breaking changes that require updating both CS and Hypershift operator to have matching client and server side changes.
This commit updates the Hypershift Operator version in the non MSFT environments to contain those new changes that are breaking changes.

The commit also introduces most recent CS changes.

CS image SHA 1bb9f13def38ff771df1afb7c47496c2aab89915328f8c3066b966f744b2e26d corresponds to CS commit id a0d0ee1.
Hypershift image SHA e470a4ad44f23684d36e1352a36892cffa8cbc604bac76b17a7e9b5d9dff5ed1 corresponds to Hypershift commit a94db93

### DIFF
--- a/cluster-service/cluster-creation.md
+++ b/cluster-service/cluster-creation.md
@@ -265,9 +265,11 @@ cat <<EOF > nodepool-test.json
     "azure_node_pool": {
         "resource_name": "$NAME",
         "vm_size": "Standard_D8s_v3",
-        "os_disk_size_gibibytes": 30,
-        "os_disk_storage_account_type": "StandardSSD_LRS",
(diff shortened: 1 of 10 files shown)


bennerv (approved):

miguelsorianod:
No approver of the cluster-service/cluster-creation.md file is the US timezone and the change that triggers codeowners of CS is simply the small documentation update in cluster-service/cluster-creation.md so I think we can merge it

PR #2362, opened 2025-08-04
collapse test client usage

deads2k:
combines our client initialization into one spot to avoid confusion.


test/util/framework/per_invocation_framework.go
```diff
@@ -87,6 +92,16 @@ func (tc *perBinaryInvocationTestContext) getAzureCredentials() (azcore.TokenCre
 		return tc.azureCredentials, nil
 	}
 
+	if tc.isLocalDevelopmentEnvironment {
```
patriksuba, line 95: Maybe we do not need LOCAL_DEVELOPMENT. AzureDefaultCredentials will accept what is actually present. https://github.com/Azure/ARO-HCP/blob/main/test/util/framework/per_invocation_framework.go#L93

test/util/framework/per_invocation_framework.go
```diff
@@ -164,6 +200,12 @@ func tenantID() string {
 	return os.Getenv("AZURE_TENANT_ID")
 }
 
+// isLocalDevelopmentEnvironment indicates when this environment is development.  This controls client endpoints and disables security
+// when set to development.
+func isLocalDevelopmentEnvironment() bool {
```
patriksuba, line 205: I would not use word Local in this context. Let's just use isDevelopmentEnvironment name and exclude using azidentity.NewAzureCLICredential(nil).

test/util/framework/per_invocation_framework.go
```diff
@@ -99,6 +114,27 @@ func (tc *perBinaryInvocationTestContext) getAzureCredentials() (azcore.TokenCre
 	return tc.azureCredentials, nil
 }
 
+func (tc *perBinaryInvocationTestContext) getClientFactoryOptions() *azcorearm.ClientOptions {
```
patriksuba, line 117: Localhost configuration is crucial for development environment. Agree to include this.

test/e2e/cluster_delete_test.go
```diff
@@ -18,26 +18,25 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/Azure/ARO-HCP/test/util/framework"
```
patriksuba, line 21: make all-tidy from root directory will fix this. https://github.com/Azure/ARO-HCP/blob/main/Makefile#L69

mbukatov (approved):
Thank you for this alignment!

mmazur (approved):

PR #2363, opened 2025-08-04
e2e: test keyvault certs

sanchezl:
https://issues.redhat.com/browse/ARO-18498

### What

Add e2e test to verify that Kube API Server and default Ingress endpoints are using certificates from key vault.

### Why

Feature completion

The tests simply try to verify that the certs are not signed by the internal OpenShift CA. More explicit validation would be great (e.g. actually lookup and compare with cert in keyvalut) but I think this is a good start.

### DIFF
--- a/test/e2e/cluster_tls_endpoints.go
+++ b/test/e2e/cluster_tls_endpoints.go
@@ -132,8 +132,8 @@ var _ = Describe("Customer", func() {
 
 		fmt.Fprintf(GinkgoWriter, "Issuer: %v\n", actualAPICert.Issuer)
 		Expect(actualAPICert.Issuer).To(SatisfyAll(
-			HaveField("CommonName", ContainElements("Microsoft Azure RSA TLS Issuing CA")),
-			HaveField("Organization", ContainElements("Microsoft Corporation")),
+			HaveField("CommonName", MatchRegexp(`Microsoft\sAzure\sRSA\sTLS\sIssuing\sCA\s[0-9]+`)),
+			HaveField("Organization", ContainElement("Microsoft Corporation")),
 		), "expect certificate to be issued by Microsoft")
 
 		By("creating the node pool")
@@ -181,8 +181,8 @@ var _ = Describe("Customer", func() {
 			fmt.Fprintf(GinkgoWriter, "Issuer OU: %v", actualCert.Issuer.OrganizationalUnit)
 			return actualCert, nil
 		}).WithTimeout(2*time.Minute).WithPolling(10*time.Second).To(SatisfyAll(
-			HaveField("CommonName", ContainElements("Microsoft Azure RSA TLS Issuing CA")),
-			HaveField("Organization", ContainElements("Microsoft Corporation")),
+			HaveField("Issuer.CommonName", MatchRegexp(`Microsoft\sAzure\sRSA\sTLS\sIssuing\sCA\s[0-9]+`)),
+			HaveField("Issuer.Organization", ContainElement("Microsoft Corporation")),
 		), "expect certificate to be issued by Microsoft")
 	})
 })


mbukatov:
@sanchezl please link feature JIRA in the PR description

test/e2e/cluster_tls_endpoints.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+)
+
+var _ = Describe("Customer", func() {
+	BeforeEach(func() {
+		// do nothing.  per test initialization usually ages better than shared.
+	})
+
+	It("should be able to create an HCP cluster using bicep templates",
```
mbukatov, line 38: You will need to tweak the description to distinguish this particular test case so that we don't have a test case naming conflict causing a confusion when check the test results.

test/e2e/cluster_tls_endpoints.go
```diff
@@ -0,0 +1,206 @@
...
+				framework.Must(TestArtifactsFS.ReadFile("test-artifacts/generated-test-artifacts/modules/nodepool.json")),
+				map[string]interface{}{
+					"openshiftVersionId": openshiftNodeVersionId,
+					"clusterName":        customerClusterName,
+					"nodePoolName":       customerNodePoolName,
+					"replicas":           2,
+				},
+				45*time.Minute,
+			)
+			Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
jharrington22: Yeah I need the console to test the default ingress certificate

test/e2e/cluster_tls_endpoints.go
jharrington22: Updated, thanks

test/e2e/cluster_tls_endpoints.go
```diff
@@ -0,0 +1,204 @@
...
+				return false
+			}
+			consoleURL = *resp.Properties.Console.URL
+			fmt.Fprintln(GinkgoWriter, "Ingress URL found:", consoleURL)
+			return true
+		}).WithTimeout(15 * time.Minute).WithPolling(10 * time.Second).Should(BeTrue())
+
+		By("examining the server certificate returned by the default ingress when routing the console URL")
+		// Wait for the certificate to be loaded after concole starts
+		time.Sleep(2 * time.Minute)
```
SudoBrendan: nit: It would be best if we added an `Eventually` polling mechanism to fetch the cert rather than waiting a hard-coded time to avoid race conditions.

SudoBrendan (commented):
Things mostly look good here! Can we remove the `Sleep` call in the test?

test/e2e/cluster_tls_endpoints.go
```diff
@@ -0,0 +1,204 @@
...
+			customerVnetSubnetName           = "customer-vnet-subnet1"
+			customerClusterName              = "tls-endpoint-hcp-cluster"
+			customerNodePoolName             = "np-1"
+			openshiftControlPlaneVersionId   = "4.19"
+			openshiftNodeVersionId           = "4.19.7"
+		)
+		tc := framework.NewTestContext()
+
+		By("creating a resource group")
+		resourceGroup, err := tc.NewResourceGroup(ctx, "tls-endpoint-cluster", "uksouth")
```
bennerv:
```suggestion
		resourceGroup, err := tc.NewResourceGroup(ctx, "tls-endpoint-cluster", tc.Location())

test/e2e/cluster_tls_endpoints.go
bennerv: can we not put a hard sleep in here and rather wait for the ingress controllers to restart?

SudoBrendan (dismissed):

SudoBrendan (approved):

PR #2364, opened 2025-08-04
shorten deployment names + introduce geneva admin cert domain

geoberle:
### What

EV2 prepends a random identifier in front of deployment names so we need to make sure our deployment names are short enough to still have room for that


stevekuznetsov:
Any way we can validate this at any point earlier than Ev2 rollout?

tsatam (approved):

bennerv (approved):

PR #2373, opened 2025-08-05
Update demo script with etcd encrption config

patilsuraj767:
### What

- Updated scripts with etcd encrption with customer managed config
- Removed role assignment as `Key Vault Administrator` role is assigned at subscription scope.


venkateshsredhat:
@patilsuraj767  Before updating the demo scripts lets wait for the RP changes to be merged . So lets not merge this now .

demo/03-create-cluster.sh
```diff
@@ -99,6 +99,39 @@ IDENTITY_UAMIS_JSON_MAP=$(echo -n "${IDENTITY_UAMIS_JSON_MAP}" | jq \
...
+initialize_etcd_encryption_json_map() {
+  ETCD_ENCRYPTION_JSON_MAP=$(cat << 'EOF'
+    {
+      "dataEncryption": {
+        "keyManagementMode": "CustomerManaged",
+        "customerManaged": {
+          "encryptionType": "KMS",
+          "kms": {
+            "activeKey": {
+              "vaultName": "string",
```
miguelsorianod, line 111: They can be just empty string too ("")

miguelsorianod:
I think here we should update the cluster tmpl json to make it accurate to what's going to be set. Essentially let's put example with dummy values when using kms with customer managed keys

miguelsorianod:
Let's also make sure we verify that the contents are generated correctly

PR #2376, opened 2025-08-05
 Initial e2e testing for external auth for ARO HCP.

Nanyte25:
[ARO-20092](https://issues.redhat.com/browse/ARO-20092)


mbukatov (changes requested):
Thank you for working on the tests. QE team will reach out to you to help with setup work so that the test can use existing plumbing code and unified way to pass setup information. Without this, every test will be handling setup slightly differently which will be impossible to run in the CI consistently.

PR #2383, opened 2025-08-05
actually verify that the image registry is not present

deads2k:
extending a previous test with kubeconfig creds.


deads2k:
image registry test was green.

PR #2396, opened 2025-08-06
Set the default of enabled for ClusterImageRegistryProfile

mociarain:
I think we just missed it in the past. This PR just sets it as a default of "Enabled".

Note: This property is not updateable, but we only rely on the 'visibility' label for this. I guess that's fine, but we could do something by comparing it to the ResourceDoc and throwing if it's changing. Not for now but just a thought

### DIFF
--- a/frontend/pkg/frontend/ocm.go
+++ b/frontend/pkg/frontend/ocm.go
@@ -46,6 +46,7 @@ const (
 	azureNodePoolNodeDrainGracePeriodUnit string = "minutes"
 	csImageRegistryStateDisabled          string = "disabled"
 	csImageRegistryStateEnabled           string = "enabled"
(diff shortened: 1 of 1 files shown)


frontend/pkg/frontend/ocm.go
```diff
@@ -46,6 +46,7 @@ const (
 	azureNodePoolNodeDrainGracePeriodUnit string = "minutes"
 	csImageRegistryStateDisabled          string = "disabled"
 	csImageRegistryStateEnabled           string = "enabled"
+	csPlatformOutboundType                string = "load_balancer"
```
mociarain, line 49: This is just an unrelated piece of cleanup I did when I saw it

mbarnes (approved):
Nice cleanups.  LGTM.

PR #2402, opened 2025-08-06
Force centraluseuap to be non-zonal

tsatam:
### What

Updates our `_locationAvailabilityZones` array used in Bicep modules to define centraluseuap as a non-zonal region. 

### Why

While the centraluseuap region advertises itself as supporting 2 AZs, these zones are not supported by all resources. 

None

### DIFF
--- a/dev-infrastructure/modules/common.bicep
+++ b/dev-infrastructure/modules/common.bicep
@@ -102,6 +102,7 @@ var _locationAvailabilityZones = {
   eastus2euap: {
     availabilityZones: [
       '1'
+      // '2' Not available in EV2
       '3'
       '4'
     ]


jfchevrette (approved):

dev-infrastructure/modules/common.bicep
```diff
@@ -102,8 +102,8 @@ var _locationAvailabilityZones = {
   eastus2euap: {
     availabilityZones: [
       '1'
```
jfchevrette, line 104:
```suggestion
      '1'
      // '2' Not available in EV2
      '3'
      '4'
```
nit: I would leave 2 commented out with a short explanation, for posterity

PR #2404, opened 2025-08-06
update demo cluster creation bicep

geoberle:
### What

* provide image registry settings (might not be required, investigating pending)
* pass in short x.y.z version on creation


bennerv (changes requested):
can you update them in `test/e2e` as well?

PR #2405, opened 2025-08-06
Gniranjan/billing config

gouthamMN:
### What
1. Add kusto sku to config.
2. Remove ComosDB from billing.

### Why
1. Need to parameterize kusto sku.
2. Refer the cosmosDB name from existing frontend config.


gouthamMN:
> Please squash your commits, explain why we're making the change in the message and we can merge

addressed.

PR #2406, opened 2025-08-06
pipelines: constrain global resource groups to uksouth

stevekuznetsov:
tooling: bump ARO-Tools to get execution constraints

---

pipelines: constrain global resource groups to uksouth

Using the new execution constraint functionality, we can declare that
global resource groups only execute against `uksouth`. Regional resource
groups remain able to refer to the chained outputs from these groups. We
will need to add additional constraints to other clouds as we roll out
to them by adding more entries to the constraint list.

---


weinong (approved):
LGTM

PR #2414, opened 2025-08-07
test: add ExternalAuth E2E tests and supporting artifacts

Nanyte25:
### What
https://issues.redhat.com/browse/ARO-20092


mfreeraro:
@deads2k  and @mociarain could one of you please review this pr

demo/main.sh
```diff
@@ -0,0 +1,261 @@
...
+# Dependencies check
+install_if_missing() {
+  cmd=$1
+  pkg=$2
+  icon=$3
+  if ! command -v "$cmd" &>/dev/null; then
+    echo "$icon Installing $cmd..."
+    sudo dnf install -y "$pkg"
+  fi
+}
```
mbarnes, line 32: I don't think we want to get into package management.  Plenty of people on the team use Mac OS.  Others (including me) may be using Debian or a derivative, which would require an `apt` command instead of `dnf`.

I think it's enough to error out with a message for missing dependencies and let the developer satisfy them.

mbarnes: Long-term, I think this demo code needs to be rewritten in a higher level language.  Python, Go, or whatever.  I appreciate the polish you're trying to add but this is getting way too complex for bash, in my opinion.

mbarnes (changes requested):
I think I'm gonna have to make several passes at this.  It's too much to grok at once and I'm a little out of my depth on the bicep parts.  Suggest getting reviews from others more familiar with bicep like maybe @mbukatov.

Had one request on the `main.sh` script.

test/e2e/external_auth_complete_cluster.go
```diff
@@ -0,0 +1,242 @@
...
+					"customerVnetName":       customerVnetName,
+					"customerVnetSubnetName": customerVnetSubnetName,
+				}, 45*time.Minute)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("creating HCP cluster (injecting MAJOR.MINOR version)")
+			rawClusterTpl := framework.Must(
+				TestArtifactsFS.ReadFile("test-artifacts/generated-test-artifacts/standard-cluster-create/cluster.json"),
+			)
+			ver := strings.TrimSpace(os.Getenv("HCP_OPENSHIFT_VERSION"))
```
deads2k, line 136: Not an env var.  Hard code for now, we can make a loop and run one for every version we support later.

test/e2e/external_auth_complete_cluster.go
```diff
@@ -0,0 +1,242 @@
...
+					"nodePoolName": customerNodePoolName,
+				}, 45*time.Minute)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("applying ExternalAuth config via RP (using Entra app from Step 1)")
+
+			// 1) Load Entra creds JSON (local loader)
+			secretPath := strings.TrimSpace(os.Getenv("ENTRA_E2E_SECRET_PATH"))
+			if secretPath == "" {
+				secretPath = "test/e2e/out/entra_app_secret.json"
```
deads2k, line 172: no defaults.

test/e2e/external_auth_complete_cluster.go
```diff
@@ -0,0 +1,242 @@
...
+			}
+			sf := mustLoadEntraSecretStep2(secretPath)
+
+			// 2) Build issuer/audience/client from Entra
+			issuerURL := fmt.Sprintf("https://login.microsoftonline.com/%s/v2.0", sf.TenantID)
+			externalAuthClientID := sf.ClientID
+			cliAudience := sf.ClientID
+			externalAuthClientSecret := ""
```
deads2k, line 180: clients are defined out of band?  we'll want to create these unless there's a security boundary reason we cannot.

test/e2e/external_auth_complete_cluster.go
```diff
@@ -0,0 +1,242 @@
...
+			body, err := json.Marshal(authPayload)
+			Expect(err).NotTo(HaveOccurred())
+
+			// 4) POST to RP
+			baseURL := strings.TrimSpace(os.Getenv("RP_BASE_URL"))
+			if baseURL == "" {
+				baseURL = "https://127.0.0.1:8443"
+			}
+			rpEndpoint := fmt.Sprintf("%s/api/aro_hcp/v1alpha1/clusters/%s/external_auth_config/external_auths",
+				baseURL, customerClusterName)
```
deads2k, line 217: somewhat surprised our generated client doesn't have these. Double check to see if https://github.com/Azure/ARO-HCP/blob/main/internal/api/v20240610preview/generated/externalauths_client.go#L53 or something near it works.

test/e2e/external_auth_complete_cluster.go
```diff
@@ -0,0 +1,242 @@
...
+			verifyReq, err := http.NewRequestWithContext(ctx, http.MethodGet, rpEndpoint, nil)
+			Expect(err).NotTo(HaveOccurred())
+			verifyResp, err := client.Do(verifyReq)
+			Expect(err).NotTo(HaveOccurred())
+			defer verifyResp.Body.Close()
+			Expect(verifyResp.StatusCode).To(Equal(http.StatusOK))
+
+			respData, err := io.ReadAll(verifyResp.Body)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(respData)).To(ContainSubstring(externalAuthID))
```
deads2k, line 240: once this is supposedly complete, we should verify it.

test/e2e/external_entra_auth_e2e_tests.md
```diff
@@ -0,0 +1,92 @@
...
+ExternalEntra: publish OIDC config to RP builds OIDC config (incl. allow group) and POSTs to RP [step3]
+Customer should be able to create an HCP cluster with Image Registry not present
+
+```
+
+## Environment Variables
+
+| Variable | Required? | Description |
+|----------|-----------|-------------|
+| `HCP_CLUSTER_NAME` | | Name of the HCP cluster to operate on |
```
deads2k, line 37: remove, we want to install a new cluster each time

test/e2e/external_entra_auth_e2e_tests.md
```diff
@@ -0,0 +1,92 @@
...
+| `ENTRA_E2E_SECRET_PATH` | for step 2–3 | Path to JSON file where Step 1 saved the Entra app credentials |
+| `ENTRA_APP_OBJECT_ID` |  for step 4 | Object ID of the Entra application to delete |
+| `EXTERNAL_AUTH_ID` | Optional | ExternalAuth config ID in RP (default: `e2e-hypershift-oidc`) |
+| `INSECURE_SKIP_TLS` | Optional (dev only) | Set to `true` to skip TLS verification when port-forwarding |
+| `RP_BASE_URL` | Optional (stage/prod) | Direct RP endpoint base URL (e.g. `https://<rp-host>:8443`). If unset, tests will port-forward to `aro-hcp-frontend` |
```
deads2k, line 42: for int and stage this should not be needed.

test/e2e/external_entra_auth_e2e_tests.md
```diff
@@ -0,0 +1,92 @@
...
+| `CUSTOMER_SUBSCRIPTION` | Optional | Azure subscription for creating clusters (only needed if test provisions a cluster) |
```
deads2k, line 43: should be required and non-optional

test/e2e/external_entra_create.go
```diff
@@ -0,0 +1,205 @@
...
+				ClientID:        app.AppID,
+				ClientSecret:    addResp.SecretText,
+				DisplayName:     displayName,
+				CreatedAt:       start,
+				SecretExpiresAt: end,
+			})).To(Succeed())
+
+			By("saved Entra app credentials JSON at " + outFile)
+		})
+})
```
deads2k, line 205: it looks like all the pieces are here. We'd like to end up with one test that

1. creates the `entra_app_secret.json` registration
2. creates the cluster that uses that client
3. confirms that the cluster actually honors the external OIDC

We won't be running the three independently, we'll be running them in order.

deads2k (changes requested):
it looks like all the pieces are here. We'd like to end up with one test that

    creates the entra_app_secret.json registration
    creates the cluster that uses that client
    confirms that the cluster actually honors the external OIDC

We won't be running the three independently, we'll be running one test that does all the pieces.

demo/main.sh
mfreeraro: Yeah, I was just tryong polish up the demo script, far enough point though @mbarnes

.gitignore
```diff
@@ -18,6 +18,17 @@ tooling/bin
...
+demo/external-auth-payload.json
+.env
+# Generated Bicep artifacts
+*.sha256
+test/e2e/bicep-templates/
+test/e2e/out/
+<<<<<<< HEAD
+=======
+.openshift-tests-extension/
+>>>>>>> 2b3f0017 (chore: ignore .openshift-tests-extension/)
```
machi1990, line 34: git conflicts

deads2k:
`: Customer should be able to create a cluster with external auth and login via console and external auth` shows success.  Bugged or working?  That woudl be great if working.

mbarnes (approved):
Not an expert on E2E tests yet but it looks reasonable to me.

SudoBrendan (approved):

mociarain (approved):

PR #2419, opened 2025-08-07
add permissions for MIs to e2e-setup bicep

mbukatov:
based on https://github.com/Azure/ARO-HCP/pull/2274 (commit 033a42c)

### What

Implements changes in MI based on 033a42c in test/e2e-setup

### Why

The change is necessary for Staging enviroment.

I executed deployment  `mbuk-0807a-rg` and finished with success in INT.


mgahagan73 (approved):
I can't spot any issues reading through the templates, and it looks like it can make a cluster so lgtm.

PR #2420, opened 2025-08-07
tooling/templatize: honor dependencies during step execution

stevekuznetsov:
Previously, we were doing a naive in-order traversal of the resource groups and steps, which led some piplines to be written in very convoluted ways so as to make sure that this implementation detail did not interfere with execution and input/output chaining.

We now build a graph of steps that need to execute, and traverse them in a depth-first manner. We will be able to add parallelism to this as well, but I did not have time to make that happen right now.

### DIFF
--- a/tooling/templatize/pkg/pipeline/run.go
+++ b/tooling/templatize/pkg/pipeline/run.go
@@ -19,8 +19,12 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
+	"strings"
+	"sync"
 
 	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/Azure/ARO-Tools/pkg/config"
 	"github.com/Azure/ARO-Tools/pkg/types"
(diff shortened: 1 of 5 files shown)


geoberle (dismissed):
LGTM

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -78,56 +82,320 @@ func (o ShellOutput) GetValue(_ string) (*OutPutValue, error) {
...
+			thisLogger := logger.WithValues("routine", fmt.Sprintf("consumer-%d", i))
+			defer func() {
+				wg.Done()
+				thisLogger.Info("Consumer thread shutting down.")
+			}()
+			for {
+				select {
+				case step, open := <-queue:
+					if !open {
+						thisLogger.Info("Queue channel closed.")
```
stevekuznetsov, line 168: With more than one consumer routine, this (and all of the early exits closing `done` and `errs`) will cause writes to closed channel panics as the consumers race to close everything.

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -78,56 +82,320 @@ func (o ShellOutput) GetValue(_ string) (*OutPutValue, error) {
...
+						errs <- err
+						close(errs)
+						return
+					}
+					stepLogger.Info("Finished step.")
+					done <- struct{}{}
+				case <-ctx.Done():
+					thisLogger.Info("Context cancelled.")
+					close(done)
+					close(errs)
```
deads2k, line 187: closing channels more than once causes a panic

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -78,56 +82,320 @@ func (o ShellOutput) GetValue(_ string) (*OutPutValue, error) {
...
+	logger.Info("Generating execution graph.")
+	nodes, err := createExecutionGraph(ctx, pipeline, options.Region, options.SubsciptionLookupFunc)
+	if err != nil {
+		return nil, err
+	}
+
+	lock := &sync.RWMutex{}
+	executed := sets.Set[types.StepDependency]{}
+	queued := sets.Set[types.StepDependency]{}
+	outputs := make(Outputs)
```
deads2k, line 102: I think a struct containing these items so the lock access is closely associated and passed a single unit will likely make life easier.

PR #2421, opened 2025-08-07
oidc storage account PLS location  config

geoberle:
### What

for AFD the oidc storage accounts creates a PLS in the same region as the storage account. but not every region supports this so we need a way to override.

the global defaults will still default to the same region


bennerv (approved):
lgtm

PR #2422, opened 2025-08-08
Add etcd kms e2e

jharrington22:
[ARO-20323](https://issues.redhat.com//browse/ARO-20323)

### What

e2e test to verify creation of clusters with customer managed KMS for etcd encryption


jharrington22:
No ready for merge yet.

test/e2e/cmk_etcd_cluster_create.go
```diff
@@ -0,0 +1,84 @@
...
+			const (
+				customerNetworkSecurityGroupName = "customer-nsg-name"
+				customerVnetName                 = "customer-vnet-name"
+				customerVnetSubnetName           = "customer-vnet-subnet1"
+				customerClusterName              = "cmk-etcd-hcp-cluster"
+			)
+			tc := framework.NewTestContext()
+
+			By("creating a resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "cmk-etcd", "uksouth")
```
deads2k, line 47: nit: we aren't charged extra for vowels: `customer-managed-key-etcd`?

test/e2e/test-artifacts/cmk-etcd/cmk-etcd-cluster.bicep
```diff
@@ -0,0 +1,596 @@
...
+}
+
+resource nsg 'Microsoft.Network/networkSecurityGroups@2022-07-01' existing = {
+  name: nsgName
+}
+
+//
+// K E Y V A U L T
+//
+module clusterEtcdKeyVault '../../../../dev-infrastructure/modules/keyvault/keyvault.bicep' = {
```
deads2k, line 42: today I learned a thing.  This is how we wire up referenced files?

@mbukatov is right, next week we'll need to refactor and make some kind of a rule for which directory we want these to all live in.  It's not a huge deal yet, but a verifier will help us get handle on it.  is it the keyword `module`?

deads2k (changes requested):
Based on CI output

> fail [github.com/Azure/ARO-HCP/test/util/framework/per_invocation_framework.go:214]: Test Panicked: open test-artifacts/generated-test-artifacts/standard-cluster-create/customer-infra.json: file does not exist

Looks like hte bash script modifications failed.

Otherwise looks fine.

test/e2e/test-artifacts/cmk-etcd/cmk-etcd-cluster.bicep
mbukatov: > today I learned a thing. This is how we wire up referenced files?

Yes, via [module keyword](https://learn.microsoft.com/en-us/training/modules/create-composable-bicep-files-using-modules/), you can reference one bicep file from another.

mbukatov: Ok, let's add a KMS module to the e2e-setup/bicep/modules directory which can be used by the e2e tests.

mbukatov: Since etcd customer managed encryption (with keys in customer's key vault) is now mandatory/default (via https://github.com/Azure/ARO-HCP/pull/2464), there is no need to add any bicep files in this PR. You have two options:
- use the default demo.bicep file for now (eg. like @patriksuba is doing in https://github.com/Azure/ARO-HCP/pull/2480/files#diff-806127346576c1be7b26880a1b61f8a5673c93672d691073c3e12f00b3f06f86R74)
- create a new directory in https://github.com/Azure/ARO-HCP/tree/main/test/e2e-setup/bicep for this test (eg. `cmk-etcd`) and copy `cluster.bicep` file from modules dir there without modifying it and use that (like @deads2k does in https://github.com/Azure/ARO-HCP/pull/2434)
I prefer the 1st option right now since it's easier, but 2nd options makes sense if we want to change the default/mandatory status of the feature in the future.

mbukatov (changes requested):
See https://github.com/Azure/ARO-HCP/pull/2422#discussion_r2279535378 how to update the PR after the changes introduced this week.

test/e2e/cmk_etcd_cluster_create.go
jharrington22: Fixed

raelga:
Superseeded by https://github.com/Azure/ARO-HCP/pull/2464

raelga (commented):
Superseeded by https://github.com/Azure/ARO-HCP/pull/2464

/close

PR #2428, opened 2025-08-08
 api: Disallow X.Y.Z cluster versions

mbarnes:
### What

Explicitly disallow `MAJOR.MINOR.PATCH` cluster versions in request bodies.

Also, this reworks #2408 from yesterday.  It turns out the baked-in `semver` validation tag _requires_ a `PATCH` value, so I've restored the custom `openshift_version` tag that parses with `github.com/hashicorp/go-version` which is more lenient.  But we still disallow the `openshift-v` prefix.


bennerv (changes requested):
please update the bicep files to use the shorter version.  

`demo/`

`test/`

bennerv (approved):

PR #2434, opened 2025-08-08
Use e2e setup scripts

deads2k:
This separates e2e from demo scripts.


test/e2e-setup/bicep/image-registry/disabled-image-registry-cluster.bicep
```diff
@@ -0,0 +1,89 @@
+@description('Name of the hypershift cluster')
```
mbukatov, line 1: Since you are moving this to e2e setup, I believe it's worth updating it bit further.

Since this file is basically almost the same as the cluster bicep file from the modules dir:

```
$ cd test/e2e-setup/bicep
$ diff --ignore-space-change -u image-registry/disabled-image-registry-cluster.bicep modules/cluster.bicep 
--- image-registry/disabled-image-registry-cluster.bicep        2025-08-11 13:39:53.237993175 +0200
+++ modules/cluster.bicep       2025-08-11 11:40:56.667148071 +0200
@@ -71,7 +71,7 @@
       visibility: 'Public'
     }
     clusterImageRegistry: {
-      state: 'Disabled'
+      state: 'Enabled'
     }
     platform: {
       managedResourceGroup: managedResourceGroupName
```

I suggest to:

- update the existing modules/cluster.bicep file so that state of the cluster image registry is controlled via parameter with a default value of "Enabled" (eg. `param clusterImageRegistryState string = 'Enabled'`)
- create new bicep file `disabled-image-registry-cluster.bicep` in the `tests/e2e-setup/bicep/` directory, where you will just use the modules as done in the demo.bicep, but set the parameter to disabled
- the the golang code, deploy just that instead of manually deploying the modules one by one

This will simplify the work we do around setup a lot and allow us to quickly review that configurations we are using.

Obviously, if you need to have more control, we can add another high level bicep file with lot of parameters which one can tweak from the test code.

mbukatov (changes requested):
This is a good move. I'm adding further suggestions. Let me know if that makes sense.

deads2k:
Failure is deletion time. Spoke over slack this morning with Martin, we're starting here to get the big change in

PR #2436, opened 2025-08-08
First Party Doc/Diagram Fixes

SudoBrendan:
Fixes provisioning a cluster doc to use correct MISE routing and backend async flows

Came across this while reviewing our onboarding materials in https://issues.redhat.com/browse/ARO-15851, so I guess it's kinda related?

I'd suggest using the `view file` option to see the rendered mermaid - but this is how I see it now in my browser:
image


mociarain (approved):
LGTM

PR #2441, opened 2025-08-11
Make it possible to use socket for audit client

janboll:
### What

Make it possible to use socket for audit client

### Why

Need to use socket for frontend audit logging in MSFT environments


janboll:
superseeded by https://github.com/Azure/ARO-HCP/pull/2476

PR #2450, opened 2025-08-11
tooling/templatize: adapt to new ARO-Tools graph

stevekuznetsov:
tooling/templatize: adapt to new ARO-Tools graph

We can remove our own graph construction code and use the one provided
by the ARO-Tools library.

---

tooling/templatize: hold on to lock for access to queued

---

tooling/templatize: hold variables with lock in struct

Co-locating the lock with the data over which it operates helps to make
sure users correctly hold the lock before touching the data.

---

tooling/templatize: signal consumers to exit with context

Previously, a consumer routine would close the output channels for all
consumers before exiting, but this would cause other consumers to
attempt to send to a closed channel if they exited later. Instead, we
can use a scoped context specifically for consumers to signal to other
consumers that we encountered an error and need to close. Waiting for
all consumers to exit before closing the channels ensures we don't panic.

---

tooling/templatize: allow parallel execution

Simple follow-up to thread through a parallelism option and handle the
cleanups that we needed as a result.

---


stevekuznetsov:
Requires https://github.com/Azure/ARO-Tools/pull/79

tooling/templatize/pkg/pipeline/arm.go
```diff
@@ -207,10 +207,12 @@ func pollAndPrint[T any](ctx context.Context, p *runtime.Poller[T]) error {
...
 }
 
-func doDryRun(ctx context.Context, client *armresources.DeploymentsClient, rgName string, deploymentName string, step *types.ARMStep, pipelineWorkingDir string, cfg config.Configuration, input Outputs) (Output, error) {
+func doDryRun(ctx context.Context, client *armresources.DeploymentsClient, rgName string, deploymentName string, step *types.ARMStep, pipelineWorkingDir string, cfg config.Configuration, state *ExecutionState) (Output, error) {
 	logger := logr.FromContextOrDiscard(ctx)
 
-	inputValues, err := getInputValues(step.Variables, cfg, input)
+	state.RLock()
+	inputValues, err := getInputValues(step.Variables, cfg, state.Outputs)
+	state.RUnlock()
```
deads2k, line 215: prefer getters for things you need.  If they must match exactly, one function with two return values.

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -122,28 +134,25 @@ func RunPipeline(pipeline *types.Pipeline, ctx context.Context, options *Pipelin
 					return
```
deads2k, line 134: above this in the diff, what prevents `close(queue)` from being called twice?

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -82,35 +84,45 @@ func (o ShellOutput) GetValue(_ string) (*OutPutValue, error) {
...
+		Queued:   sets.Set[graph.Dependency]{},
+		Outputs:  make(Outputs),
+	}
 
-	queue := make(chan *stepExecutionNode, len(nodes))
-	done := make(chan struct{}, len(nodes))
-	errs := make(chan error, len(nodes))
-	wg := &sync.WaitGroup{}
-	wg.Add(1)
+	queue := make(chan graph.Dependency, len(executionGraph.Nodes))
```
deads2k, line 116: is this logically, `stepsToExecuteNow`?

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -206,200 +228,68 @@ func RunPipeline(pipeline *types.Pipeline, ctx context.Context, options *Pipelin
...
 	}
 
-	output, err := executor(node.Step.Delegate, ctx, node.Step.Target, options, outputs)
+	output, err := executor(step, ctx, target, options, state)
 	if err != nil {
 		return err
 	}
 
-	lock.Lock()
+	state.Lock()
```
deads2k, line 279: Use `defer unlock` here.  Looks risky for panic.

deads2k:
some comments.  lgtm otherwise.

tooling/templatize/pkg/pipeline/run.go
stevekuznetsov: very short  (7 lines) critical section with no early returns? seems ok

PR #2456, opened 2025-08-12
feat: update CS image to include etcd data encryption with cmk in dev…

miguelsorianod:
We update the CS image to introduce the functionality that now requires creating ARO-HCP Clusters with etcd data encryption enabled with customer managed keys.

The sha 551377f1f37f26e92ef0d4b1228fa42650f78b2cd3194b52ffaf480cd0796dda corresponds to the CS git commit a196446

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -697,7 +697,7 @@ clouds:
       clustersService:
         environment: "arohcpdev"
         image:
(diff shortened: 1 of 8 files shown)


miguelsorianod:
Do not merge yet. We need to coordinate the rollout with the RP as they are breaking changes

PR #2468, opened 2025-08-13
ARO-20528 no CNI e2e setup

mbukatov:
e2e-setup for ARO-20528

### What

Add a no CNI configuration to e2e-setup bicep

### Why

To create a configuration for BYO CNI use case.

### DIFF
--- a/test/e2e-setup/bicep/modules/cluster.bicep
+++ b/test/e2e-setup/bicep/modules/cluster.bicep
@@ -16,6 +16,15 @@ param subnetName string
 @description('OpenShift Version ID to use')
 param openshiftVersionId string = '4.19'
 
+@description('Network configuration of the hosted cluster')
+param networkConfig object = {
+  networkType: 'OVNKubernetes'
+  podCidr: '10.128.0.0/14'
+  serviceCidr: '172.30.0.0/16'
+  machineCidr: '10.0.0.0/16'
+  hostPrefix: 23
+}
+
 @description('Cluster Managed Identities: ')
 param userAssignedIdentitiesValue object
 
@@ -54,13 +63,7 @@ resource hcp 'Microsoft.RedHatOpenShift/hcpOpenShiftClusters@2024-06-10-preview'
       channelGroup: 'stable'
     }
     dns: {}
-    network: {
-      networkType: 'OVNKubernetes'
-      podCidr: '10.128.0.0/14'
-      serviceCidr: '172.30.0.0/16'
-      machineCidr: '10.0.0.0/16'
-      hostPrefix: 23
-    }
+    network: networkConfig
     console: {}
     etcd: {
       dataEncryption: {
--- a/test/e2e-setup/bicep/no-cni.bicep
+++ b/test/e2e-setup/bicep/no-cni.bicep
@@ -1,0 +1,49 @@
+@description('If set to true, the cluster will not be deleted automatically after few days.')
+param persistTagValue bool = false
+
+@description('Name of the hypershift cluster')
+param clusterName string
+
+module customerInfra 'modules/customer-infra.bicep' = {
+  name: 'customerInfra'
+  params: {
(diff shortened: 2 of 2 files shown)


test/e2e-setup/bicep/no-cni.bicep
```diff
@@ -0,0 +1,49 @@
...
+  name: 'cluster'
+  params: {
+    clusterName: clusterName
+    vnetName: customerInfra.outputs.vnetName
+    subnetName: customerInfra.outputs.vnetSubnetName
+    nsgName: customerInfra.outputs.nsgName
+    userAssignedIdentitiesValue: managedIdentities.outputs.userAssignedIdentitiesValue
+    identityValue: managedIdentities.outputs.identityValue
+    networkConfig: {
+      networkType: 'Other'
```
deads2k, line 34: definitely looking forward to seeing this test.  Could the verification for this test including running two deployments that try to communication using pod-to-pod, pod-to-service, and pod-to-kube-apiserver?  We can probably publish some images from OCP to enable such a test.

test/e2e-setup/bicep/modules/cluster.bicep
```diff
@@ -16,6 +16,15 @@ param subnetName string
 @description('OpenShift Version ID to use')
 param openshiftVersionId string = '4.19'
 
+@description('Network configuration of the hosted cluster')
```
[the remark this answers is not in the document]
mbukatov: Yes, this is what we are sending to ARM to deploy ARO HCP hosted cluster. I just moved this section to a parameter so that the test case covering no CNI use case can easily specify related network options as needed.

mbukatov:
> I was wondering how would the CNI plugin be installed so that the nodepools are able to operate?

This PR is concerned with just the 1st step, deploying a cluster in no CNI plugin configuration.

The CNI plugin installation will be covered in a separate test (and a PR, see ARO-20529). I will base it on your demo as well as what we do for ROSA HCP already in [cucushift-hypershift-extended-cilium](https://steps.ci.openshift.org/reference/cucushift-hypershift-extended-cilium). If you want to discuss that part, comment under ARO-20529.

test/e2e/cluster_create_no_cni.go
```diff
@@ -0,0 +1,105 @@
...
+				customerClusterName  = "no-cni-cl"
+				customerNodePoolName = "no-cni-np"
+			)
+			tc := framework.NewTestContext()
+
+			By("creating a resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "e2e-no-cni", tc.Location())
+			Expect(err).NotTo(HaveOccurred())
+
+			By("deploying no-cni bicep file to create no-cni cluster without a node pool")
```
deads2k, line 45: Deploying a cluster without nodes should be the default. Most of our glue tests (at least so far), don't need one

test/e2e-setup/bicep/cluster-only.bicep
```diff
@@ -0,0 +1,37 @@
+@description('If set to true, the cluster will not be deleted automatically after few days.')
+param persistTagValue bool = false
+
+@description('Name of the hypershift cluster')
+param clusterName string
```
deads2k, line 5: Why accept as input instead of reporting the name back as output.

test/e2e/cluster_create_no_cni.go
mbukatov: yes, and this particular test also need to be in control of the network configuration

test/e2e-setup/bicep/cluster-only.bicep
mbukatov: this is the same approach as used in other bicep files in this level - we can add a default value and report the cluster name back as an output there, but that means we will have to change the other bicep files there as well - let's consider that in a separate PR

PR #2476, opened 2025-08-13
Use Domain Socket in INT and PROD to push audit logs to AzSecPack Namespace

rajdeepc2792:
https://issues.redhat.com/browse/ARO-20061

### What
As per the AzSec documentation, to forward the OTEL Audit Logs in AKS clusters use a volumeMount to hostpath on the AKS VMSS instances that has fluentd socket available.
  

### Why
It is Azure S360 Security requirement to forward the Primary Audit logs like the ARM requests/calls received by ARO-HCP frontend.

- Tested the changes in INT where the AzSec AutoConfig is enabled. The DomainSocket got connected successfully.
- The new socketConnect property need to be added in the config.yaml present in sdp-pipeline for Prod and INT
- After the changes are released, validation will be needed to confirm if the audit logs are successfully visible in correct namespace.


frontend/deploy/templates/frontend.deployment.yaml
```diff
@@ -83,5 +90,11 @@ spec:
             port: 8443
           initialDelaySeconds: 5
           periodSeconds: 10
+      {{- if .Values.audit.connectSocket }}
+      volumes:
+        - name: mdsd-asa-run-vol
+          hostPath: 
+            path: /var/run/mdsd/asa
```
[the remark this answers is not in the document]
rajdeepc2792: Updated the type.
So, in case the mdsd directory is not present someday on the AKS(INT/Staging/Prod) instance the frontend pod would fail.

internal/audit/otel-audit.go
```diff
@@ -45,14 +45,19 @@ func (c *AuditClient) Send(ctx context.Context, msg msgs.Msg, options ...base.Se
 	return c.client.Send(ctx, msg)
 }
 
-func NewOtelAuditClient(remoteAddress string, options ...base.Option) (*AuditClient, error) {
+func NewOtelAuditClient(connectSocket bool, remoteAddress string, options ...base.Option) (*AuditClient, error) {
```
stevekuznetsov, line 48: This is an awkward function IMO - just let the user pass in the connection, passing in proto-parameters and doing a `switch` saves nothing

rajdeepc2792: Moved the switch part to a separate function and passed the audit.CreateConn to this function.

PR #2478, opened 2025-08-13
Revert "Initial AI context files"

mbarnes:
This reverts commit f47efca266c67b02cb9a8cf3bf3ce373630ea1f4.

The soft links break Microsoft's CI.  This can be revisited later but unfortunately we're in a crunch.


stevekuznetsov:
@mmazur  we package the directory rooted at `topology.yaml` into a `.tar.gz` for use in Ev2 - there's no sane way to package symlinks into a tarball, and we need to keep deployments running, so we're taking the quick path here and simply reverting your PR. Please consider another approach to your desired end state which does not require symbolic links in this repo.

PR #2480, opened 2025-08-14
E2E Test cases for supported GPU nodepool instance types

patriksuba:
- Update e2e-bicep nodepool file with configurable VMsize
- Created helper functions to delete and get nodepool with timeout

[ARO-20545](https://issues.redhat.com/browse/ARO-20545)

### What

Add E2E Test case for GPU nodepools

### Why

To test creation and deletion of GPU nodepools

### DIFF
--- a/test/e2e/gpu_nodepools_create_delete.go
+++ b/test/e2e/gpu_nodepools_create_delete.go
@@ -21,6 +21,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/util/rand"
 
 	hcpapi20240610 "github.com/Azure/ARO-HCP/internal/api/v20240610preview/generated"
 	"github.com/Azure/ARO-HCP/test/util/framework"
@@ -56,10 +57,8 @@ var _ = Describe("HCP Nodepools GPU instances", func() {
 			labels.Critical,
 			labels.Positive,
 			func(ctx context.Context) {
-				const (
-					customerClusterName = "gpu-nodepool-cluster"
-					location            = "uksouth"
-				)
+				customerClusterName := "gpu-nodepool-cluster-" + rand.String(6)
+				location := "uksouth"
 
 				tc := framework.NewTestContext()
 


deads2k:
looks like the dry-runs were victims of timeouts and the kms problems.  I suggest limiting the list to two and retrying.  We can refine and add more over time.

test/e2e/gpu_nodepools_create_delete.go
```diff
@@ -0,0 +1,151 @@
...
+					ctx,
+					tc.Get20240610ClientFactoryOrDie(ctx).NewNodePoolsClient(),
+					*resourceGroup.Name,
+					customerClusterName,
+					npName,
+					25*time.Minute,
+				)).To(Succeed())
+
+				// Confirm it's gone
+				_, getErr := framework.GetNodePool(ctx,
```
deads2k, line 140: make a list function instead and confirm the output is none.

test/e2e/gpu_nodepools_create_delete.go
```diff
@@ -0,0 +1,151 @@
...
+					5*time.Minute,
+				)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(created.Properties).ToNot(BeNil())
+				Expect(created.Properties.ProvisioningState).ToNot(BeNil())
+				Expect(*created.Properties.ProvisioningState).To(Equal(hcpapi20240610.ProvisioningStateSucceeded))
+				Expect(created.Properties.Platform).ToNot(BeNil())
+				Expect(created.Properties.Platform.VMSize).ToNot(BeNil())
+				Expect(*created.Properties.Platform.VMSize).To(Equal(sku.vmSize))
+
```
deads2k, line 127: before we delete, check to ensure the new node is present using the `adminRESTConfig`

deads2k:
minor comments.  lgtm otherwise.  great test.

test/e2e-setup/bicep/modules/nodepool.bicep
```diff
@@ -10,6 +10,9 @@ param replicas int = 2
 @description('OpenShift Version ID to use')
 param openshiftVersionId string = '4.19.0'
 
+@description('VM size for the nodepool VMs')
+param vmSize string = 'Standard_D8s_v3'
```
deads2k, line 14: rather than add a parameter, how about we create an entirely new file.  The benefit to a new file is that as people write more tests and want more options, we'll never modify what *this* test runs.  That allows us to write a test once and never worry about what changes as we on.  If a test does eventually break or we break an API or instance type, we'll be feeling the same pain as customers.

mbukatov: For this particular test (changing `vmSize` only), updating the parameter here is completely reasonable. I suggest against such duplication here.

mbukatov (approved):
Nice test! The check could be extended to use kubeconfig to confirm node status from within the cluster as David notes, but this is not critical.

mbukatov:
The OpenShift CI runs fail on the same issue as the rest of the tests. I'm merging it to avoid too many open PRs with E2E tests.

PR #2493, opened 2025-08-15
E2E Tests negative and positive test case for updating cluster fields

patriksuba:
[ARO-20542](https://issues.redhat.com/browse/ARO-20542)

### What

Add negative and positive test case for updating simple cluster field

### Why

Expanding test cases in E2E Test Suite

### DIFF
--- a/test/e2e/cluster_update.go
+++ b/test/e2e/cluster_update.go
@@ -45,15 +45,12 @@ var _ = Describe("Update HCPOpenShiftCluster", func() {
 		It("creates a cluster and fails to update its name with a PATCH request",
 			labels.RequireNothing, labels.Medium, labels.Negative,
 			func(ctx context.Context) {
-				const (
-					clusterName = "patch-name-cluster"
-					location    = "uksouth"
-				)
+				const clusterName = "patch-name-cluster"
 
 				tc := framework.NewTestContext()
 
 				By("creating a resource group")
-				resourceGroup, err := tc.NewResourceGroup(ctx, "patch-name", location)
+				resourceGroup, err := tc.NewResourceGroup(ctx, "patch-name", tc.Location())
 				Expect(err).NotTo(HaveOccurred())
 
 				By("deploying demo template (single-step infra + identities + cluster)")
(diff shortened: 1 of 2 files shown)


mbukatov (approved):
Look good. Later on I would rather switch from updating a tag to something more fundamental about the aro hcp cluster, so that it will work in dev as well.

mgahagan73:
{fail [github.com/Azure/ARO-HCP/test/e2e/cluster_update.go:89]: Expected
    <string>: patch https://management.azure.com/subscriptions/64f0619f-ebc2-4156-9d91-c4c781de7e54/resourcegroups/patch-name-kln52k/providers/microsoft.redhatopenshift/hcpopenshiftclusters/patch-name-cluster
    response 400: 400 bad request
    error code: mismatchingresourcename
    {
      "error": {
        "code": "mismatchingresourcename",
        "message": "the provided resource name 'patch-name-cluster-renamed' did not match the name in the url 'patch-name-cluster'."
      }
    }
    
to contain substring
    <string>: field 'name' is read-only  fail [github.com/Azure/ARO-HCP/test/e2e/cluster_update.go:89]: Expected
    <string>: patch https://management.azure.com/subscriptions/64f0619f-ebc2-4156-9d91-c4c781de7e54/resourcegroups/patch-name-kln52k/providers/microsoft.redhatopenshift/hcpopenshiftclusters/patch-name-cluster
    response 400: 400 bad request
    error code: mismatchingresourcename
    {
      "error": {
        "code": "mismatchingresourcename",
        "message": "the provided resource name 'patch-name-cluster-renamed' did not match the name in the url 'patch-name-cluster'."
      }
    }
    
to contain substring
    <string>: field 'name' is read-only}


I think something is amiss here either with error message expectation in the test or we are trying to update the wrong resource.

test/e2e/cluster_update.go
```diff
@@ -0,0 +1,173 @@
...
+					ctx,
+					tc.Get20240610ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient(),
+					*resourceGroup.Name,
+					clusterName,
+					update,
+					10*time.Minute,
+				)
+				Expect(err).To(HaveOccurred())
+				// The server enforces that 'name' is read-only for updates
+				Expect(strings.ToLower(err.Error())).To(ContainSubstring(strings.ToLower("Field 'name' is read-only")))
```
mbukatov, line 89: Looks like updating a cluster name is not a good use case as this seems to be rejected on a different grounds:

```
"error": {
        "code": "mismatchingresourcename",
        "message": "the provided resource name 'patch-name-cluster-renamed' did not match the name in the url 'patch-name-cluster'."
      }
```

mbukatov (changes requested):
We need to either fix  what error we look for or find a different use case.

PR #2498, opened 2025-08-15
demo: use ARM api call instead of key vault data plane to create etcd kms encryption key

bennerv:
Fixes `Caller is not authorized to perform action on resource.` when attempting to generate etcd encryption keys

### Why

`az keyvault key create` uses the data plane APIs (keyvault.azure.com) instead of management.azure.com.  By using a `az rest` call directly to ARM, we can leverage our existing access, which mimics what is done in the bicep customer-infra templates.  

👍


miguelsorianod:
lgtm

miguelsorianod (approved):

PR #2502, opened 2025-08-15
cluster_pullsecret test 

mgahagan73:
https://issues.redhat.com/browse/ARO-20543

### What

create cluster then use kubernetes API to update the global pull secret, verify pull secret was updated. 
### Why

confirm pull secret update functionallity.


mbukatov (changes requested):
Suggesting to simplify deployment phase.

bennerv (changes requested):
The way for a customer to provide a pull secret day2 is by creating a secret in the `kube-system` namespace named `additional-pull-secret`.  It will then combine the `pull-secret` in the `openshift-config` namespace and drop a pull secret named `global-pull-secret` in the `kube-system` namespace and write the global pull secret data into the kubelet config so it can pull from all the referenced repositories.  

https://hypershift-docs.netlify.app/how-to/common/global-pull-secret/

The approach you're taking regarding updating the `pull-secret` in `openshift-config` won't work as it will be written over with the content of the pull secret that's stored in the hosted cluster namespace.

test/util/verifiers/nfd.go
```diff
@@ -0,0 +1,101 @@
...
+					imagePullErrors = append(imagePullErrors, fmt.Sprintf("pod %s: %s", pod.Name, condition.Message))
+				}
+			}
+		}
+
+		// Check container statuses for waiting state with image pull errors
+		for _, containerStatus := range pod.Status.ContainerStatuses {
+			if containerStatus.State.Waiting != nil {
+				reason := containerStatus.State.Waiting.Reason
+				if reason == "ImagePullBackOff" || reason == "ErrImagePull" {
```
mbukatov: Wouldn't it make sense to just log any waiting Reason? There are other possible errors (see [full list](https://github.com/kubernetes/kubernetes/blob/854e67bb51e177b4b9c012928d8271704e9cb80d/pkg/kubelet/images/types.go#L27)), and if these happen, I assume the verifier still fail (because we don't reach imagePulledSuccessfully state), but we don't get the reason behind this logged (if I read this right).

mbukatov (commented):
Proposing small change in the pull image verifier.

mbukatov (dismissed):

mbukatov (dismissed):

test/e2e/cluster_pullsecret.go
```diff
@@ -0,0 +1,323 @@
...
+				pullSecretNamespace,
+				testPullSecretHost,
+				auth,
+				testPullSecretEmail,
+			).Verify(ctx, adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("reading pull-secret file from aro-hcp-qe-pull-secret directory")
+			pullSecretFilePath := filepath.Join("/var/run/aro-hcp-qe-pull-secret", "pull-secret")
+			pullSecretFileData, err := os.ReadFile(pullSecretFilePath)
```
bennerv: should we make this a framework function?  

If someone runs the test suite locally, this will fail.

mbukatov: Moving this into a library function would make more sense if we could reconfigure the location of the aro-hcp-qe-pull-secret for local runs.

mgahagan73: This is in the framework now and we can specify an env var. for local runs

mbukatov (dismissed):

test/e2e/cluster_pullsecret.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("Cluster Pull Secret Management", func() {
+	BeforeEach(func() {
+		// per test initialization
+	})
+
+	It("should be able to create an HCP cluster and manage pull secrets",
```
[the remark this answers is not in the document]
mbukatov: This e2e test covers ARO HCP  feature ARO-21429, as documented in ARO HCP doc draft. And while most of the changes happened in HyperShift as you say, we find it important to be covered by E2E test.

geoberle (dismissed):

PR #2506, opened 2025-08-18
Update CS ARO HCP OCP maxVersion to 4.19.6

kostola:
### What

Update max supported version of OCP to the latest version available in the stable channel which is currently 4.19.6.

### Why

To allow clusters to be created using OCP 4.19.6.


machi1990:
This was done here https://github.com/Azure/ARO-HCP/pull/2497

PR #2531, opened 2025-08-20
Fix casing of "encryptionType" value ("KMS" not "kms")

mbarnes:
### What

The ARO-HCP RP temporarily accepts both `"kms"` and `"KMS"`. Once these corrections are deployed to all environments, the RP will be made to only accept `"KMS"`.

### Why

The OpenAPI specification for ARM only defines the enum value `"KMS"`, and enum values are case-sensitive.

### DIFF
--- a/demo/03-create-cluster.sh
+++ b/demo/03-create-cluster.sh
@@ -105,7 +105,7 @@ initialize_etcd_encryption_json_map() {
       "dataEncryption": {
         "keyManagementMode": "CustomerManaged",
         "customerManaged": {
-          "encryptionType": "kms",
+          "encryptionType": "KMS",
           "kms": {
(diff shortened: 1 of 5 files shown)


tsatam (approved):

mbukatov (approved):

mbarnes:
Well, I was going to hold off on merging this til I figure out why the RP in Staging is behaving like it doesn't have #2508 (even though [ARO HCP Release Status](https://service-status.dptools.openshift.org/#stg) indicates it does).  But at the same time having #2508 deployed is now causing [ARO-20738 - Running a bicep script twice to create an HCP fails](https://issues.redhat.com/browse/ARO-20738).

So screw it.  I'm merging this.

PR #2537, opened 2025-08-20
fix: make sure nodepool name length <= 15

mbukatov:
### What

Drops "np-gpu" prefix from a node pool name in e2e gpu node pool test.

### Why

Max length of a node pool name is 15 characters, which we are breaching with the current prefix.


mbukatov:
> looks good... could we implement some sort of pre-merge validator that could catch non-compliant Azure names in the future?

that would not be straightforward, as the names used in the tests can be constructed dynamically, as is the case here, or in other cases, we want the name to be wrong for a negative test ...

PR #2538, opened 2025-08-20
Handle unknown enum values in CS/RP conversions

mbarnes:
### What

In the course of converting a CS resource to an RP resource (or vice versa), if the RP encounters an unrecognized enum value it would just leave the destination field empty.  With this change we instead fail the conversion with an error.

### Why

Promised followup to review comments in #2508 (particularly [here](https://github.com/Azure/ARO-HCP/pull/2508#discussion_r2283483841) and [here](https://github.com/Azure/ARO-HCP/pull/2508#discussion_r2283503579)).

Please review one commit at a time.  It will be a lot easier.

This forced me to do some extensive refactoring of `ocm_test.go` because some of the tests were _relying_ on empty enum values.  The 3rd commit ("frontend: Update OCM tests") has all that junk.

### DIFF
--- a/frontend/go.mod
+++ b/frontend/go.mod
@@ -3,6 +3,7 @@ module github.com/Azure/ARO-HCP/frontend
 go 1.24.3
 
 require (
+	dario.cat/mergo v1.0.1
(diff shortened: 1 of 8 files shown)


mociarain (approved):
LGTM

PR #2545, opened 2025-08-21
e2e test of gpu nodepool: disable NC4asT4v3

mbukatov:
### What

Disabling NC4asT4v3 instance for a gpu nodepool E2E test.

### Why

In some testing regions, we don't have NC4asT4v3 available. Let's try a single gpu nodepool configuration for now.


geoberle (approved):

UlrichSchlueter (approved):
Not sure keeping these comments around is good practice, but apart from it LGTM

PR #2552, opened 2025-08-21
[WIP] Adding MRG Delete Validation

SudoBrendan:
Adds a test case to verify that in addition to the HCP instance, the Managed Resource Group is also deleted. This ensures that all resources related to the HCP instance are gone from the customer's subscription and that they won't be billed for zombie resources.

### What

New E2E test case for deletion to verify that all resources related to the HCP instance are gone from the customer subscription.

### Why

If implementation now or in the future leaves resources in the MRG, customers will continue to be billed for those resources (locked behind a Deny Assignment, so they can't touch them). This would include things like VMs/IPs/LBs that provide the customer-side infra of the HCP instance.

This also happens to protect us as "customers" here - if our CI jobs fail this test - we may have zombie resources in _our_ subs that need manual cleanup.

WIP: This implementation might be incomplete for cases where the MRG is not provided to bicep, and therefore will default to a generated value (requiring a GET before DELETE in the test flow) - I need to run the test suite to confirm this.

### DIFF
--- a/test/e2e-setup/bicep/modules/cluster.bicep
+++ b/test/e2e-setup/bicep/modules/cluster.bicep
@@ -2,7 +2,7 @@
 param clusterName string
 
 @description('The Hypershift cluster managed resource group name')
-param managedResourceGroupName string = '${clusterName}-rg'
+param managedResourceGroupName string = ''
 
 @description('The Network security group name for the hcp cluster resources')
 param nsgName string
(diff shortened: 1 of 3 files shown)


mbukatov (changes requested):
This is a good point, and we should add this check.

That said, there are few things we need to address:

* the change in test/e2e/complete_cluster_create.go (a test case with labels.RequireNothing label and as such executed during e2e-parallel jobs) is not compatible with the change proposed in the bicep file - here I suggest to drop both changes (see comments incline)
* chage in `test/e2e/cluster_delete_test.go` test case looks good but this won't be executed during e2e-parallel job since this type of test case uses different approach to a setup which we don't have a ci job for, so we will need a change in the teardown code which the parallel tests are using

geoberle:
@SudoBrendan please reopen this if you are ready to follow up

PR #2553, opened 2025-08-21
separate tests that only function in one environment

deads2k:
combined with https://github.com/openshift/release/pull/68369 keeps the test working for integration and skipped for stage.

### DIFF
--- a/test/cmd/aro-hcp-tests/main.go
+++ b/test/cmd/aro-hcp-tests/main.go
@@ -54,7 +54,7 @@ func main() {
 		Qualifiers: []string{
 			// Remember that the label constants are (currently) slices, not items.
 			// TODO we will need per-env markers eventually, but it's ok to start here
-			fmt.Sprintf(`labels.exists(l, l=="%s")`, labels.RequireNothing[0]),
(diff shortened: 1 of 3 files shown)


mgahagan73 (approved):
Ack to running GPU instances only on INT.

deads2k:
Merging. This will skip the test on staging. Alternatives welcome in the form of PRs. I like hard coded lists best from this list:

an individual env var for "VM we should use for this test" is a pattern that ages poorly as each test grows a parameter that people use.  Comparing some options:
1.  a list hardcoded by environment: extremely easy, rarely updated, zero config, pre-merge test shows working or not, owned in repo with the people debugging it, when it fails we notice
2.  configuration option for a list of VMs that are present with a default hard coded: default likely to fail weirdly for people and not notice, updates will be a differnet repo for configuration, comparable tests will be unknown and unknowable
3.  configuration option for a single VM specified for a test: same as #2, with additional problem of encouraging configuration per test

What other cons do we have for 1 or benefits for 2?  Right now the hardcoded list by environment is a clear win for speed, maintainability, and consistency.

PR #2557, opened 2025-08-22
Add Kiota SDK Generation: MSGraph

SudoBrendan:
The Microsoft Graph SDK is massive, this PR introduces an alternative to the full SDK: generating our own from the openapi spec using Kiota. This PR gives us a path forward on migrating out of microsoftgraph/msgraph-sdk-go and into ARO-HCP/internal/graph/graphsdk, which gives us the following savings:

- 191MB SDK -> 27MB SDK
- 10,000+ endpoints -> 7 endpoints
- Faster builds, smaller binaries, reduced security risk

The solution also allows us to follow this pattern for other SDKs in the future as needed, it'll take a few lines of Bash to get there.

As a follow-up, I added some utility functions to wrap common graph calls (e.g. creating an Entra Application). These are UNTESTED, and need validation.

I split this into several commits, which are likely worth reviewing independently.

### DIFF
--- a/tooling/kiota/openapi.yaml
+++ b/tooling/kiota/openapi.yaml
@@ -7,13 +7,13 @@ info:
     toolName: Microsoft.OpenApi.OData
     toolVersion: 1.0.9.0
 servers:
-  - url: https://graph.microsoft.com/v1.0
+- url: https://graph.microsoft.com/v1.0
 paths:
   /applications:
     description: Provides operations to manage the collection of application entities.
     get:
       tags:
-        - applications.application
+      - applications.application
       summary: List applications
       description: Get the list of applications in this organization.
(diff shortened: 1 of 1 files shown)


bennerv (commented):
what's the difference between this and classic?  Are we just generating less?  

https://github.com/Azure/ARO-RP/blob/dbf8d0b7bdf07aa7be2437665df443f14ca92949/hack/graphsdk/openapi.yaml

mociarain (approved):
LGTM. I guess the first E2E to use this will test and make some tweaks but this is good to merge as is IMO

mbarnes (approved):
Shiny.

bennerv (approved):

PR #2561, opened 2025-08-23
Restore support for personal Quay.io registries

mbarnes:
### What

Our aging [frontend documentation](https://github.com/Azure/ARO-HCP/blob/main/frontend/README.md) advises using your own personal [Quay.io](https://quay.io/) registry for test images in personal development environments.  This is sound advice which I haven't been following.  Recently when I decided to mend my ways, I quickly realized our deployment tooling has evolved to _only_ support Azure Container Registry.

We used to support personal Quay.io registries during early development.  This is an attempt to restore it.

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -1,36 +1,17 @@
 # ARO-HCP-FRONTEND
 
-## Build frontend binary for local testing
-```
-make frontend
-```
+## Deploy the frontend container
 
-## Build the frontend container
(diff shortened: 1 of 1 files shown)


venkateshsredhat:
Ahh now i see , I still use Quay but I had to kind of manually edit deployment of Frontend and Backend with new image .(which is not a good practice )

PR #2564, opened 2025-08-25
Fixing interval

janboll:
### What

Make alert groups interval configurable


mbukatov (approved):
Changes in test dir touches go.mod and go.sum files only, while I see that e2e run passed. Looks good from my perspective.

PR #2570, opened 2025-08-25
Update CS script to create kv within name length restrictions

jharrington22:
Fixes issues with key vault name being too long.


geoberle:
@jharrington22 please reopen if this is still needed

PR #2573, opened 2025-08-26
Allow RP to use versions from fast channel

ahitacat:
https://issues.redhat.com/browse/ARO-20740

### What

With this change the RP will allow to create clusters and node pool with fast channel.

### Why

We want to add fast channel support, so OCP versions can be delivered sooner. For that the RP and CS needs to support the use of versions from this channel. 
Additionally 'fast' channel will be the default channel. This will be in a follow-up MR.

**Requirements**:
- CS need the support of creation of node pools https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/10414
- CS needs to support multiple channel configuration https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/10401
- ARO-HCP configuration is spitted for envs (different confs for different envs) 
- Update CS image in aro-hcp, configure fast channel in `cluster-service/deploy/templates/aro-hcp-ocp-versions-config.configmap.yaml`

**Future steps:**
- Make `fast`  the default channel in CS API
- Make `fast` the default channel in the ARM API

For manual test this I have:

1. Configure CS to support fast channels
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: aro-hcp-ocp-versions-config
  namespace: '{{ .Release.Namespace }}'
data:
  aro-hcp-ocp-versions-config.yaml: |
    defaultVersion:
      channelGroupName: fast
      version: 4.19.0
    channelGroups:
      # the URL to cincinnati will be the same across all environments, other URLs also need to be whitelisted
      - url: https://api.openshift.com/api/upgrades_info/graph
        channelGroupName: fast
        # the versions must be in sync with what's defined in dev-infrastructure/templates/global-image-sync.bicep
        # this is to ensure that the enabled release images are synchronized to ACR in lockstep
        minVersion: 4.19.0
        # maxVersion is exclusive, contrary to oc-mirror which defines that boundary as inclusive.
        maxVersion: 4.19.8
```
2. Deploy CS with the change in https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/10414
3. Deploy the frontend with the change in this PR
4. Change the demo templates for cluster and node pool to use `fast` channel

### DIFF
--- a/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
+++ b/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
@@ -148,7 +148,7 @@ model VersionProfile {
 
   /** ChannelGroup is the name of the set to which this version belongs. Each version belongs to only a single set. */
   @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
-  channelGroup?: string = "stable";
+  channelGroup?: string = "fast";
 }
 
 /** DNS contains the DNS settings of the cluster */
@@ -574,7 +574,7 @@ model NodePoolVersionProfile {
 
   /** ChannelGroup is the name of the set to which this version belongs. Each version belongs to only a single set. */
   @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
-  channelGroup?: string = "stable";
+  channelGroup?: string = "fast";
 }
 
 /** Azure node pool platform configuration */
--- a/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
+++ b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json
@@ -2951,7 +2951,7 @@
         "channelGroup": {
           "type": "string",
           "description": "ChannelGroup is the name of the set to which this version belongs. Each version belongs to only a single set.",
-          "default": "stable",
+          "default": "fast",
           "x-ms-mutability": [
             "read",
             "update",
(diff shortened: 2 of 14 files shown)


mbarnes (changes requested):
The changes so far look good, I think.  The channel group suffix on version IDs comes as a surprise to me but it makes sense.

One thing missing here is we declare `"stable"` as the default channel group in the ARM API spec.  That needs changed as well, and it should be fine to just manually patch the Swagger file so you don't have to install all the tooling to generate it.

You'll need to update:
- The TypeSpec cluster model file ([here](https://github.com/Azure/ARO-HCP/blob/c010c6bea2b03cb830f9e61d3e29654aeb57618b/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp#L151) and [here](https://github.com/Azure/ARO-HCP/blob/c010c6bea2b03cb830f9e61d3e29654aeb57618b/api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp#L577))
- The generated Swagger file ([here](https://github.com/Azure/ARO-HCP/blob/c010c6bea2b03cb830f9e61d3e29654aeb57618b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json#L2954) and [here](https://github.com/Azure/ARO-HCP/blob/c010c6bea2b03cb830f9e61d3e29654aeb57618b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json#L3627))

mbarnes (approved):
LGTM

Now that I see how many places the default channel group is spelled out I should really define a constant for it.  I'll take care of that.

PR #2590, opened 2025-08-27
prepare CS image sha update for prod. CS commit id 5c0cfa5

miguelsorianod:
Includes TLS changes


miguelsorianod:
to be merged only when changes have been rolled out to stage, the e2e stage tests pass with the changes rolled out and also manual verification around tls certs is done

PR #2591, opened 2025-08-27
e2e: remove skipping tls verification on kubeconfigs now that we have publicly trusted certs

bennerv:
### What

Remove the tls verification skipping on kubeconfigs.  

### Why

OneCert publicly trusted certs should exist now/soon.  

Green int/stage/prod tests should be our gate here.


bennerv:
I expect this to work in INT.  Stage is currently ongoing a rollout so it'll be progressive and multiple tests before we see success.

bennerv:
Rebased to have latest e2e suite so this should work.  Also stage and prod should pass tests.

PR #2593, opened 2025-08-27
e2e: Test case for cluster & node pool with no version ID

mbarnes:
To verify bug https://issues.redhat.com/browse/ARO-20849

### DIFF
--- a/test/e2e/cluster_create_nodepool_default_version.go
+++ b/test/e2e/cluster_create_nodepool_default_version.go
@@ -1,0 +1,72 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+)
+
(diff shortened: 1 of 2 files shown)


mgahagan73:
ok it is probably best to have this test use its own bicep template(s) like e2e/cluster_create_nodepool_osdisk.go does (e2e-setup/bicep/cluster-nodepool-osdisk.bicep is based on demo.bicep). This way you can be sure that the version for cluster and nodepool are never set by a future MR that might add a paramater for the version with a default value set in the shared bicep template.   I see that e2e-setup/bicep/modules/cluster.bicep  for example has a default version value which I don't think you want for this test if I understand things correctly. This probably means you can't rely on the modules for this test as they are currently written.

test/e2e/cluster_create_nodepool_default_version.go
```diff
@@ -0,0 +1,70 @@
...
+					"nodePoolName":    customerNodePoolName,
+				},
+				45*time.Minute,
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			// FIXME Verify the reported version of the new cluster
+			//       and node pool is the current default version.
+			//
+			//       This is blocked until the hcpOpenShiftVersion
```
mbukatov, line 66: Please reference a JIRA so that we know when this needs to be addressed and the test case is no longer valid without an update.

mbukatov (dismissed):
Looks good.

test/e2e/cluster_create_nodepool_default_version.go
mbarnes: Will do, if I can find it...

mbarnes:
Looks like I have a correction to make yet.
```
ERROR: /go/src/github.com/Azure/ARO-HCP/test/e2e-setup/bicep/cluster-nodepool-default-version.bicep(44,19) : Error BCP057: The name "nodePoolName" does not exist in the current context. [https://aka.ms/bicep/core-diagnostics#BCP057]
```

mbarnes:
Ok good, the test case is failing as I would expect for now.
```
{
    "message": "Version 'openshift-v.0.7' doesn't exist"
}
```
(The node pool would be failing with a similar error if the deployment got that far.)

Once the [bug fix](https://github.com/Azure/ARO-HCP/pull/2592) is rolled out — presumably next week — this test case should pass.

I'd hold off on merging til then.

mbarnes:
/retest

The bug fix should at least be in INT by now.

mbarnes:
Closing since this test case has become obsolete.  Version ID fields are no longer optional.

PR #2595, opened 2025-08-27
e2e: ensure we generate a unique name for the managed resource group to prevent parallel tests causing conflicts

bennerv:
### What

If two e2e tests run at the same time in the same environment, the managed resource group names will conflict, resulting in one of the clusters to fail to create.  Use a unique string suffix on the managed resource group based on where the resourceGroup is. 

This will work since we create a resource group with a suffix when we call the test framework.  

### Why

Prevent e2e test failures when they multiple run at the same time.


bennerv:
```
Cluster with managed resource group name 'hcp-cluster-np-128-rg' in subscription '<sub-id>' already exists, please provide a unique managed resource group name
```

mgahagan73 (approved):
LGTM. There are some other tests which use their own bicep templates, I can find those tomorrow and submit a MR to update them in the morning. I'll also include a note in the readme as well.

PR #2596, opened 2025-08-27
BREAKING CHANGE: Demos Auto-Detect Env

SudoBrendan:
JIRA: N/A - this arose from convos with consulting teams.

### What

TL;DR - **the BREAKING CHANGE: to hit `localhost` Frontend instances when using demos, use `RP_MODE=development`**. All other requests route via ARM, and your resource extensions (e.g. `-int`) are automatically determined based on the Subscription you are logged into (and its registered AFECs).

### Why

Today, we ask our consultants to manually edit `env_vars` in our demos to test our RP. This change gives them a clear path to test without modifying code, but also simplifies our demo naming conventions to auto-detect what environment you are connecting to rather than keying off of hard-coded values.

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -8,6 +8,22 @@
 * (optional but useful) port-forward Maestro running on SC: `kubectl port-forward -n maestro svc/maestro 8002:8000`
 * ensure that the environment variable `$LOCATION` is either unset or set to the integrated dev environment's region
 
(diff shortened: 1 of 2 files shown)


mociarain (dismissed):
Do we still want this? If so, I have no problem with it so leaving it as approved and up to you

PR #2599, opened 2025-08-28
Revise handling of null values in PATCH requests

mbarnes:
### What

This reworks the RP to correctly apply requests using JSON Merge Patch format.

The first step was to actually use a [JSON Merge Patch package](https://github.com/evanphx/json-patch) instead of just unmarshalling the request body directly onto a resource model struct.  This is encapsulated in a new `ApplyRequestBody` helper function.  If the HTTP method is `"PATCH"`, we:

- Marshal a versioned resource struct representing the current resource to a byte array.
- Feed the byte array along with the request body to the [JSON Merge Patch package](https://github.com/evanphx/json-patch).
- Unmarshal the resulting byte array into a "fresh" versioned resource struct.

Because we're using a _versioned_ resource struct, where all the fields are pointers, we can correctly represent fields that were removed by the patch request.

The question then becomes how to _interpret_ field removals in a predefined resource model.  This is where things get thorny: some fields are designed to be removable, but most are not.  What do you do, for example, with a `"PATCH"` request that wants to eliminate the `"properties"` bag entirely?

```
{
    "properties": null
}
```

The solution I landed on was to whitelist which resource fields can be removed in a `"PATCH"` request.  The whitelist is assembled by examining the _internal_ (non-versioned) resource model types.  Fields that meet the following two conditions are whitelisted for removal:

1. The field is declared as a pointer type.  This includes slices and maps.
2. The field is updatable according to it's own `"visibility"` struct tag or through inherited visibility.

The whitelist is recorded in the model type's `VisibilityMap`.

The `VisibilityMap` is a mapping of dot-notated struct field names (e.g. `"Properties.DNS.BaseDomainPrefix"`) to visibility flags (e.g. `VisibilityRead | VisibilityCreate`).  It essentially transfers the `"visibility"` struct tags from the _internal_ (non-versioned) model type to the _versioned_ model type.  (This is necessary because [Autorest won't let us set custom struct tags in its generated Go models](https://github.com/Azure/autorest.go/issues/1614).)

A new _computed_ visibility flag - `VisibilityNullable` - marks which fields can be removed by `null` in a `"PATCH"` request.  Attempting to remove fields without a `VisibilityNullable` flag will result in an error response during visibility validation.

To revisit the earlier example, a `"PATCH"` request with

```
{
    "properties": null
}
```

will now result in this error response:

```
{
    error: {
      code: "InvalidRequestContent",
      message: "Field 'properties' cannot be removed",
      target: "properties"
    }
}
```


internal/api/utils.go
```diff
@@ -171,3 +178,184 @@ func MergeStringPtrMap(src map[string]*string, dst *map[string]string) {
 		}
 	}
 }
+
+// CopyReadOnlyValues copies from src any empty values for read-only fields in dst.
+func CopyReadOnlyValues[T any](src VersionedCreatableResource[T], dst VersionedCreatableResource[T]) {
```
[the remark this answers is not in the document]
stevekuznetsov: Not only do you not even know if the algorithmic approach is correct (see the comments in here about slices) but your claim is simply untrue.

A generator that walks structures to find their members, uses struct tags or comments to decorate them and generate e.g. deep copies, validation, transformation is better in dare I say every way:

 - reflection is often an order of magnitude slower; this is such a well-understood fact I'm not super keen on finding you specific data - please search the internet or run some of your own benchmarks. You can even see the [official JSON benchmarks](https://github.com/go-json-experiment/jsonbench) claim: "For JSONv1 and JSONv2, marshaling from concrete types is significantly limited by the performance of Go reflection." By inspection, consider what the CPU actually has to _do_ in the code you've written (how many cycles, how much allocation, etc) versus the alternative here, which is a generated method that copies some data from A to B
 - reflection-based approaches are hard to write, hard to read, and hard to validate - I claim you know this, and I have your comments on the slices as proof. A generated snippet that handles a few fields explicitly by moving or copying is clearly superior in every instance
 - serialization and deserialization is extremely expensive, as well - doing it twice in and of itself should be an extremely difficult thing to justify in a server hot-path, if not disqualifying. deserialization once then an update in-place for merging should be what we're aiming for
 
 > For maintainability sake I'd like to avoid this path. 

If you have some need to get this API behavior in place quickly, it's a suggestion. Writing the code manually for the few types you manage will help you understand the shape of the problem and unblock you to have some more time to formulate a permanent solution.

frontend/pkg/frontend/node_pool.go
```diff
@@ -105,13 +111,28 @@ func (f *Frontend) CreateOrUpdateNodePool(writer http.ResponseWriter, request *h
...
 		case http.MethodPut:
+			// Initialize versionedRequestNodePool to include both
+			// non-zero default values and current read-only values.
+			reqNodePool := api.NewDefaultHCPOpenShiftClusterNodePool()
+
+			// Some optional create-only fields have dynamic default
+			// values that are determined downstream of this phase of
+			// request processing. To ensure idempotency, add these
+			// values to the target struct for the incoming request.
+			reqNodePool.Properties.Version.ID = hcpNodePool.Properties.Version.ID
```
mociarain, line 130: I'm worried that it's going to be hard to remember to add to this in future. Could we move the dynamic defaulfs into a new function like NewDefaultHCPOpenShiftClusterNodePoolWithDynamicDefaults that takes a HCPNodePoolDynamicDefaults just so it's more self-contained and obvious

internal/api/v20240610preview/external_auth_methods.go
```diff
@@ -79,21 +79,32 @@ func (h *ExternalAuth) Normalize(out *api.HCPOpenShiftClusterExternalAuth) {
...
-		c.ExternalAuth,
-		current.(*ExternalAuth).ExternalAuth,
-		externalAuthStructTagMap, updating)
+	errorDetails = c.ValidateVisibility(current, updating)
+
+	// Proceed with additional validation only if visibility validation has
+	// passed. This avoids running further checks on changes we already know
+	// to be invalid and prevents the response body from becoming overwhelming.
+	if len(errorDetails) == 0 {
+		var normalized api.HCPOpenShiftClusterExternalAuth
```
mociarain, line 101: No blocking but just my 2 cents: I would do the opposite here and give a user all the information rather than "whack-a-mole". Maybe a question for product.

mociarain (approved):
Few small ideas but otherwise LGTM

internal/api/v20240610preview/external_auth_methods.go
mbarnes: I definitely want to avoid "whack-a-mole".

My thinking was just to partition validation into three progressive phases.

1. Are you even allowed to set/change these fields?
2. Are the values you're setting these fields to valid (syntactically, within acceptable limits, etc.)?
3. Do these changes violate invariants across the whole resource (or its parent)?

Each phase can still report multiple errors, but your changes have to pass one phase to get to the next.

But yes, it's debatable if this is better than everything at once.

PR #2601, opened 2025-08-28
Verify workloads work in e2e tests

deads2k:
Adds an optional verifier that deploys a simple webapp and makes sure that we get a response from it with valid cert via the cluster's ingress.


deads2k:
@mbukatov I ended up needing to debug with the cluster still present.  I took the liberty of adding an env var to stop cleanup as you asked for.  It would be fine to refine it (maybe take text and have an option to only save things that failed?), but it's a basic start.

mbukatov:
> @mbukatov I ended up needing to debug with the cluster still present. I took the liberty of adding an env var to stop cleanup as you asked for. It would be fine to refine it (maybe take text and have an option to only save things that failed?), but it's a basic start.

This is good, I wanted to have this from the beginning.

mbukatov (approved):
This is nice.  I was just thinking that we need some interface for deploying k8s resources when working on the cilium setup.

deads2k:
```
failed to create route: the server does not allow this method on the requested resource
    {
        msg: "failed to create route: the server does not allow this method on the requested resource",
        err: <*errors.StatusError | 0xc0002d8460>{
            ErrStatus: {
                TypeMeta: {Kind: "Status", APIVersion: "v1"},
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "the server does not allow this method on the requested resource",
                Reason: "MethodNotAllowed",
                Details: {Name: "", Group: "", Kind: "", UID: "", Causes: nil, RetryAfterSeconds: 0},
                Code: 405,
            },
        },
    }
```

This is slightly worse than it looks. It means that when we're saying the cluster is ready, the openshift apiserver is not working for some reason.

Luckily that is partially visible in the kube-apiserver.

test/util/verifiers/apiservice.go
```diff
@@ -0,0 +1,92 @@
...
+				failureReasonErrs = append(failureReasonErrs, fmt.Errorf("apiservice/%s is not available because it has no status", unavailableAPIService.Name))
+			} else {
+				failureReasonErrs = append(failureReasonErrs, fmt.Errorf("apiservice/%s is not available because: %v", unavailableAPIService.Name, availableCondition.Message))
+			}
+		}
+		return errors.Join(failureReasonErrs...)
+	}
+	if !foundAPIServices.HasAll(requiredAPIServices.UnsortedList()...) {
+		return fmt.Errorf("required apiservices are missing: %v", strings.Join(requiredAPIServices.Difference(foundAPIServices).SortedList(), ", "))
+	}
```
deads2k, line 76: @mbukatov notice how I require certain services and require they have a true.  It's not the only way, but it's a way.

test/util/verifiers/basic.go
```diff
@@ -0,0 +1,110 @@
...
+
+var standardVerifiers = []HostedClusterVerifier{
+	verifyBasicAccess(),
+	verifyAllAPIServicesAvailable(),
+}
+
+func VerifyHCPCluster(ctx context.Context, adminRESTConfig *rest.Config, additionalVerifiers ...HostedClusterVerifier) error {
+	allVerifiers := append(standardVerifiers, additionalVerifiers...)
+
+	// if these start taking a long time, run in parallel
```
stevekuznetsov, line 100: errgroup makes this fairly trivial impl, may as well

test/util/verifiers/helper.go
```diff
@@ -0,0 +1,179 @@
...
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"sigs.k8s.io/yaml"
```
stevekuznetsov, line 26: Has this passed `make lint-fix`? I thought we'd see a different ordering of imports - I don't care about import ordering, but wanted to check that the linter is catching these files.

test/util/verifiers/serving_app.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create namespace: %w", err)
+	}
+
+	dynamicClient, err := dynamic.NewForConfig(adminRESTConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create dynamic client: %w", err)
+	}
+
+	deployment, err := createArbitraryResource(ctx, dynamicClient, namespace.Name, must(staticFiles.ReadFile("artifacts/serving_app/deployment.yaml")))
```
stevekuznetsov, line 77: my assumption is stuff like `must()` exists to make this block more terse, but IMO panic is never a good idea for the poor sod who will hit it, so -

personal preference, feel free to change nothing, but I like a looping pattern that lets you be much more concise and not resort to stuff like `must()` to save chars. such as

```go
    var deployment, service, route *unstructured.Unstructured
	for from, into := range map[string]*unstructured.Unstructured {
		"artifacts/serving_app/deployment.yaml": deployment,
		"artifacts/serving_app/service.yaml": service,
		"artifacts/serving_app/route.yaml": route,
	} {
		data, err := staticFiles.ReadFile("artifacts/serving_app/deployment.yaml")
		if err != nil {
			t.Fatalf("whatever")
		}
		obj, err := createArbitraryResource(ctx, dynamicClient, namespace.Name, data)
		if err != nil {
			t.Fatalf("whatever")
		}
		*into = obj
	}
```

test/util/verifiers/serving_app.go
```diff
@@ -0,0 +1,214 @@
...
+		"deployment", deployment.GetName(),
+		"service", service.GetName(),
+		"route", route.GetName(),
+	)
+
+	// check for route to have hostname for us
+	host := ""
+	var lastErr error
+	err = wait.PollUntilContextTimeout(ctx, 10*time.Second, 5*time.Minute, true, func(ctx context.Context) (done bool, err error) {
+		currRoute, err := dynamicClient.Resource(gvr("route.openshift.io", "v1", "routes")).
```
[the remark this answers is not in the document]
deads2k: > dynamic client why?

It was here and the arbitrary object handling is generally appearing.  People will copy/paste for projects that don't stay as current as OCP and I'd like it to age decently well.

test/util/verifiers/serving_app.go
deads2k: In general, I prefer not-panicking, but in this case we're reading a static embedded file, so the only error mode is when someone breaks it in the code, not at runtime, right?

test/util/verifiers/serving_app.go
```diff
@@ -0,0 +1,214 @@
...
+	case lastErr != nil:
+		return fmt.Errorf("route host was never found: %w", lastErr)
+	case err != nil:
+		return fmt.Errorf("route host was never found: %w", err)
+	}
+
+	return nil
+}
+
+func gvr(group, version, resource string) schema.GroupVersionResource {
```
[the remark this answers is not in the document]
stevekuznetsov: yeah, but it makes me wonder what is going on under the hood since you chose not to use a literal, which then cost me mental overhead during reading - not sure the shortness is worth it but /shrug

test/util/verifiers/helper.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if restMapping.Scope.Name() == meta.RESTScopeNameRoot {
+		return dynamicClient.Resource(restMapping.Resource).Create(ctx, desiredObj, metav1.CreateOptions{})
+	}
+
+	return dynamicClient.Resource(restMapping.Resource).Namespace(namespace).Create(ctx, desiredObj, metav1.CreateOptions{})
+}
+
+// defaultRESTMappings contains enough RESTMappings to have enough of the kube-controller-manager succeed when running
+// against a kube-apiserver that cannot reach aggregated APIs to do a full mapping.  This happens when the OwnerReferencesPermissionEnforcement
```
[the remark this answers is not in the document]
stevekuznetsov: are you saying you'll refactor to use the client to get the restmapping instead of hardcoding it in the tool?

test/util/verifiers/helper.go
deads2k: I ran it, didn't do anything.

PR #2603, opened 2025-08-28
add frontend e2e tests

hbhushan3:
Adds E2E tests for frontend endpoints. CI should pass.

https://issues.redhat.com/browse/ARO-17324

### DIFF
--- a/test/e2e/admin_credential_lifecycle.go
+++ b/test/e2e/admin_credential_lifecycle.go
@@ -1,0 +1,298 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/client-go/rest"
+
+	hcpapi "github.com/Azure/ARO-HCP/internal/api/v20240610preview/generated"
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("Customer", func() {
+	BeforeEach(func() {
+		// do nothing.  per test initialization usually ages better than shared.
+	})
+
+	It("should be able to test admin credentials before cluster ready, then full admin credential lifecycle",
+		labels.RequireNothing, labels.High, labels.Positive,
+		func(ctx context.Context) {
+			const (
+				customerNetworkSecurityGroupName = "admin-cred-nsg"
+				customerVnetName                 = "admin-cred-vnet"
+				customerVnetSubnetName           = "admin-cred-subnet"
+				openshiftControlPlaneVersionId   = "4.19"
+			)
+			clusterName := "admin-cred-lifecycle-" + rand.String(6)
+			tc := framework.NewTestContext()
+
+			By("creating resource group for admin credential lifecycle testing")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "admin-credential-lifecycle-test", "uksouth")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("creating a customer-infra")
+			customerInfraDeploymentResult, err := framework.CreateBicepTemplateAndWait(ctx,
+				tc.GetARMResourcesClientFactoryOrDie(ctx).NewDeploymentsClient(),
+				*resourceGroup.Name,
+				"customer-infra",
+				framework.Must(TestArtifactsFS.ReadFile("test-artifacts/generated-test-artifacts/modules/customer-infra.json")),
+				map[string]interface{}{
+					"persistTagValue":        false,
+					"customerNsgName":        customerNetworkSecurityGroupName,
+					"customerVnetName":       customerVnetName,
+					"customerVnetSubnetName": customerVnetSubnetName,
+				},
+				45*time.Minute,
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("creating a managed identities")
+			keyVaultName, err := framework.GetOutputValue(customerInfraDeploymentResult, "keyVaultName")
+			Expect(err).NotTo(HaveOccurred())
+			managedIdentityDeploymentResult, err := framework.CreateBicepTemplateAndWait(ctx,
+				tc.GetARMResourcesClientFactoryOrDie(ctx).NewDeploymentsClient(),
+				*resourceGroup.Name,
+				"managed-identities",
+				framework.Must(TestArtifactsFS.ReadFile("test-artifacts/generated-test-artifacts/modules/managed-identities.json")),
+				map[string]interface{}{
+					"clusterName":  clusterName,
+					"nsgName":      customerNetworkSecurityGroupName,
+					"vnetName":     customerVnetName,
+					"subnetName":   customerVnetSubnetName,
+					"keyVaultName": keyVaultName,
+				},
+				45*time.Minute,
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("starting cluster deployment asynchronously to test admin credentials on deploying cluster")
+			userAssignedIdentities, err := framework.GetOutputValue(managedIdentityDeploymentResult, "userAssignedIdentitiesValue")
+			Expect(err).NotTo(HaveOccurred())
+			identity, err := framework.GetOutputValue(managedIdentityDeploymentResult, "identityValue")
+			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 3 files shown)


test/e2e/revoke_credentials_test.go
```diff
@@ -0,0 +1,98 @@
...
+				Expect(framework.VerifyHCPCluster(ctx, adminRESTConfig)).To(Succeed())
+
+				By("revoking the cluster credentials via ARM API")
+				poller, err := clusterClient.BeginRevokeCredentials(ctx, *resourceGroup.Name, clusterName, nil)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("waiting for revocation operation to complete")
+				_, err = poller.PollUntilDone(ctx, nil)
+				Expect(err).NotTo(HaveOccurred())
+
```
SudoBrendan, line 83: shouldn't we verify our credentials no longer work here? Subsequent calls with our kubeconfig should get us a 403, right?

test/e2e/admin_credentials_test.go
```diff
@@ -0,0 +1,70 @@
...
+				clusterName,
+				10*time.Minute,
+			)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(adminRESTConfig).NotTo(BeNil())
+
+			By("verifying admin credentials work with cluster")
+			Expect(framework.VerifyHCPCluster(ctx, adminRESTConfig)).To(Succeed())
+		})
+})
```
SudoBrendan, line 70: I think it's a super heavy lift for us to create a cluster here just to get credentials for it. We also duplicate this test below (and rely on _another_ HCP creation) again effectively - because revoking credentials implies creating them. Is there a test we can piggy-back off of for 1 suite for "Create and Revoke Credentials" together to save cores/cycles?

test/e2e/revoke_credentials_test.go
SudoBrendan: another missing test case that's worth a test: revoke credentials is expected to cause _all_ customer created credentials to fail as above. Test case would be:

1. Create cred A, GET on default Service returns 2xx
2. Create cred B, GET on default Service returns 2xx
3. Revoke
4. Cred A GET on default Service returns 403
5. Cred B GET on default Service returns 403

To improve the test further, we could create an arbitrarily long list of credentials, revoke, and test each of them, so it's easy for us to test scalability concerns as well (e.g. 10 or 100 creds would be easy to scale the test to)

SudoBrendan (changes requested):

test/e2e/cluster_versions_test.go
```diff
@@ -0,0 +1,65 @@
...
+			tc := framework.NewTestContext()
+			invalidLocation := "invalidlocation"
+
+			By("attempting to list versions for invalid location")
+			versionsClient := tc.Get20240610ClientFactoryOrDie(ctx).NewHcpOpenShiftVersionsClient()
+			versionsPager := versionsClient.NewListPager(invalidLocation, nil)
+
+			_, err := versionsPager.NextPage(ctx)
+			Expect(err).To(HaveOccurred(), "Should return error for invalid location")
+		})
```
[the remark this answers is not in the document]
SudoBrendan: a TL;DR on that - this test would be validating ARM rather than the RP, I think we can remove it. ARM is responsible for location-based routing of requests, so if that is not working, there's nothing we can really do to affect that; this is effectively a test for ARM not experiencing an outage, so I think we can remove it.

mbukatov (changes requested):
Add a check that revoked cert no longer works as Brendan suggests (in https://github.com/Azure/ARO-HCP/pull/2603/files#r2310493070).

The other Brendan suggestion with multiple credentials to test with could be turned into a separate (sub)test.

I also believe we can drop test/e2e/admin_credentials_test.go (this is covered already) and test/e2e/admin_credentials_test.go (we don't need to test ARM).

test/e2e/revoke_credentials_test.go
hbhushan3: Added. Thanks! Currently, we are using `credentialCount := 3`

hbhushan3: Also added CONSULTANT HYPOTHESIS TEST

test/e2e/admin_credential_lifecycle.go
```diff
@@ -0,0 +1,300 @@
...
+				}
+				Expect(cluster.HcpOpenShiftCluster.ID).NotTo(BeNil())
+				return nil
+			}, 15*time.Minute, 30*time.Second).Should(Succeed(), "Cluster should appear in ARM API within 15 minutes")
+
+			// Get cluster again for subsequent operations
+			cluster, err := clusterClient.Get(ctx, *resourceGroup.Name, clusterName, nil)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Should be in a provisioning state, not succeeded
```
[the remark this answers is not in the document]
stevekuznetsov: Maybe can you force cluster bringup to fail by providing an MSI name that does not yet exist, for instance, doing all this logic while the cluster cannot come up, then creating the MSI and waiting for the HCP system to recognize that and progress?

test/e2e/admin_credential_lifecycle.go
```diff
@@ -0,0 +1,300 @@
...
+				clusterName,
+				10*time.Minute,
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("ensuring the cluster is viable")
+			err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("creating the node pool")
```
stevekuznetsov, line 228: You don't need a node pool for this test

test/util/framework/per_test_framework.go
```diff
@@ -111,14 +111,20 @@ func (tc *perItOrDescribeTestContext) deleteCreatedResources(ctx context.Context
...
+		if err != nil {
+			ginkgo.GinkgoLogr.Error(err, "at least one resource group failed to delete: %w", err)
+		}
 	}
 
-	err = CleanupAppRegistrations(ctx, graphClient, appRegistrations)
-	if err != nil {
-		ginkgo.GinkgoLogr.Error(err, "at least one app registration failed to delete: %w", err)
+	// Only attempt app registration cleanup if there are app registrations to clean up
+	if len(appRegistrations) > 0 {
```
bennerv, line 123: Why would we care if we check length?  Shouldn't the functions themselves loop over the app registrations and resource groups and do it gracefully?  I think this change isn't needed.

hbhushan3: If we remove these checks, the cluster versions test fails (cluster_versions.go) as we will be deleting resources that were never created in the cluster versions file. We get test panic / nil error as seen in this CI build - "should be able to list available HCP OpenShift versions and validate response content" -

https://[storage.googleapis.com/test-platform-results/pr-logs/pull/Azure_ARO-HCP/2603/pull-ci-Azure-ARO-HCP-main-integration-e2e-parallel/1963377109103546368/build-log.txt](https://storage.googleapis.com/test-platform-results/pr-logs/pull/Azure_ARO-HCP/2603/pull-ci-Azure-ARO-HCP-main-integration-e2e-parallel/1963377109103546368/build-log.txt)

```
"result": "failed",
    "output": "  STEP: listing HCP OpenShift versions via ARM API @ 09/03/25 23:27:37.145\n  STEP: validating version response structure and content @ 09/03/25 23:27:37.783\n  STEP: verifying at least one version is available for cluster creation @ 09/03/25 23:27:37.783\n\"level\"=0 \"msg\"=\"collecting debug info\"\n\"level\"=0 \"msg\"=\"finished collecting debug info\"\n\"level\"=0 \"msg\"=\"deleting created resources\"\n  [PANICKED] in [DeferCleanup (Each)] - /usr/lib/golang/src/runtime/panic.go:262 @ 09/03/25 23:27:37.783\n",
    "error": "\ngithub.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources.(*ClientFactory).NewResourceGroupsClient(...)\n\t/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources@v1.2.0/client_factory.go:78\ngithub.com/Azure/ARO-HCP/test/util/framework.(*perItOrDescribeTestContext).deleteCreatedResources(0xc00048e400, {0x7f72cc47d068, 0xc0006ec7e0})\n\t/go/src/github.com/Azure/ARO-HCP/test/util/framework/per_test_framework.go:114 +0x201\nreflect.Value.call({0x1b89fa0?, 0xc00012c8e0?, 0x13?}, {0x1f4b204, 0x4}, {0xc0008b0f60, 0x1, 0x1?})\n\t/usr/lib/golang/src/reflect/value.go:584 +0xca6\nreflect.Value.Call({0x1b89fa0?, 0xc00012c8e0?, 0x0?}, {0xc00048c760?, 0x0?, 0x0?})\n\t/usr/lib/golang/src/reflect/value.go:368 +0xb9\ngithub.com/onsi/ginkgo/v2/internal.NewCleanupNode.func2({0x23079b0?, 0xc0006ec7e0?})\n\t/go/pkg/mod/github.com/openshift/onsi-ginkgo/v2@v2.6.1-0.20250416174521-4eb003743b54/internal/node.go:589 +0x174\ngithub.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()\n\t/go/pkg/mod/github.com/openshift/onsi-ginkgo/v2@v2.6.1-0.20250416174521-4eb003743b54/internal/suite.go:901 +0x7b\ncreated by github.com/onsi/ginkgo/v2/internal.(*Suite).runNode in goroutine 55\n\t/go/pkg/mod/github.com/openshift/onsi-ginkgo/v2@v2.6.1-0.20250416174521-4eb003743b54/internal/suite.go:888 +0xd7b\n\nfail [runtime/panic.go:262]: Test Panicked: runtime error: invalid memory address or nil pointer dereference\nfail [runtime/panic.go:262]: Test Panicked"
  },```

bennerv: > as we will be deleting resources that were never created in the cluster versions file

This feels like a bug in the test case or cleanup code then.  We have negative tests that don't create resource groups already and they clean up just fine without this change.  `cluster_create_test.go` for example.

bennerv: Once this is merged and you rebase on top of main, we shouldn't need the length check anymore: https://github.com/Azure/ARO-HCP/pull/2646

bennerv: Rebase, resolve comment, watch e2e pass (hopefully)

PR #2610, opened 2025-08-31
use cached upstream images for ACM/MCE

geoberle:
### What

PR #2607 prepared an image cache for the ACM/MCE upstream images from quay.io/acm-d to the service ACR under acm-d-cached and now we can leverage this cache for ACM/MCE deployment for 2 benefits

* we don't need oc-mirror anymore to mirror images from registry.redhat.io (because ACM artifact caching works for quay.io)
* we can use ACM/MCE prerelease images from quay.io/acm-d in case we need to

this PR is kind of a noop. the images content won't change because the digests remain the same, but since registry and repo change, there will be rolling upgrades.

https://issues.redhat.com/browse/ARO-20950


bennerv (approved):
lgtm

PR #2613, opened 2025-09-01
feat: make ocp version config configurable based on env

ashishmax31:
https://issues.redhat.com/browse/ARO-20733


config/config.schema.json
```diff
@@ -854,6 +854,48 @@
...
+                    "type": "string"
+                  },
+                  "minVersion": {
+                    "type": "string"
+                  },
+                  "maxVersion": {
+                    "type": "string"
+                  }
+                },
+                "required": ["url", "channelGroupName", "minVersion", "maxVersion"],
```
ahitacat, line 892: maxVersion is optional. 
I.e, for personal dev envs we can configure CS without setting maxVersion so it can get the latest images.

config/config.yaml
```diff
@@ -590,6 +590,16 @@ defaults:
       serviceAccountName: clusters-service
     azureRuntimeConfig:
       tlsCertificatesIssuer: OneCertV2-PublicCA
+    # OCP Versions configuration
```
ahitacat, line 593: If we want to specify this in different envs this can be overwritten in the corresponding env in this file?

ahitacat (changes requested):

config/config.schema.json
ashishmax31: Sure thanks

config/config.yaml
ashishmax31: Yes that is the idea.

config/config.schema.json
```diff
@@ -912,6 +935,39 @@
...
+              "type": "object",
+              "properties": {
+                "stable": {
+                  "$ref": "#/definitions/ocpVersionChannelGroupSpec"
+                },
+                "fast": {
+                  "$ref": "#/definitions/ocpVersionChannelGroupSpec"
+                }
+              },
+              "required": ["stable", "fast"],
```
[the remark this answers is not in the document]
ashishmax31: No it will be the same since we keep the "fast" block's enabled as false in the pipeline values file.

ahitacat (changes requested):
It looks good so far, but we haven't introduce yet the fast channel. And I am not sure if we will introduce it soon. Can you remove the fast channel references? I could add them back once we decide if fast will be part or even other channel groups.

PR #2619, opened 2025-09-03
metrics: fixup titles, correlation IDs, update MSFT copies

stevekuznetsov:
metrics: update MSFT copies

These were missed when the definitions were updated.

---

tooling/prometheus-rules: use require, not assert

The assert library does not exit early if the assertion fails, which is
not the correct semantic almost ever for testing. Change to using the
require library does what we want.

---

tooling/prometheus-rules: allow udpating test fixtures

It is not clear to me why these tests were not written to use the normal
golden fixture approach we have, or why they are all factored to use
implicit data passing using the filesystem. A better refactor would test
a function that takes config and data as input (as opposed to references
to files) and asserts over the output, not over output files. That is a
larger refactor; keeping this surgical, I added the smallest diff to get
us useful updates to test fixtures.

---

tooling/prometheus-rules: add a more complex test case

---

tooling/prometheus-rules: collapse whitespace in multi-line rules

We seemingly require single-line rules in our generated prometheus
configuration, but the previous munging left a lot of whitespace that
only made the rules less readable.

---

tooling/prometheus-rules: expose title, correlation ID to IcM

We want the IcM titles to be more meaningful - right now they contain a
bunch of useless or semi-useless information. By explicitly setting the
field, we can use the nice templated descriptions and highlight the
cluster.

Similarly, the default correlation ID does not take into account the
dimensions along which the alert varies, which are necessary to include
in the correlation ID to make sure that separate instances of the alert
firing (across different clusters, or for different workloads in one
cluster) get separate IcMs.

---

dev-infrastructure: bring MSFT rule groups up to date

---

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/msftPrometheusAlertingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/msftPrometheusAlertingRules.bicep
@@ -13,130 +13,202 @@ resource kubernetesResources 'Microsoft.AlertsManagement/prometheusRuleGroups@20
     interval: 'PT1M'
     rules: [
       {
-        actions: [for g in actionGroups: { actionGroupId: g }]
+        actions: [
+          for g in actionGroups: {
+            actionGroupId: g
+            actionProperties: {
+              'IcM.Title': concat('#$.labels.cluster#', ': ', '#$.annotations.description#')
+              'IcM.CorrelationId': '#$.annotations.correlationId#'
+            }
+          }
+        ]
         alert: 'KubeCPUOvercommit'
         enabled: true
         labels: {
           severity: 'warning'
         }
         annotations: {
+          correlationId: 'KubeCPUOvercommit/{{ $labels.cluster }}'
           description: 'Cluster {{ $labels.cluster }} has overcommitted CPU resource requests for Pods by {{ $value }} CPU shares and cannot tolerate node failure.'
           runbook_url: 'https://runbooks.prometheus-operator.dev/runbooks/kubernetes/kubecpuovercommit'
           summary: 'Cluster has overcommitted CPU resource requests.'
         }
-        expression: 'sum(namespace_cpu:kube_pod_container_resource_requests:sum{}) by (cluster) - (sum(kube_node_status_allocatable{job="kube-state-metrics",resource="cpu"}) by (cluster) - max(kube_node_status_allocatable{job="kube-state-metrics",resource="cpu"}) by (cluster)) > 0 and (sum(kube_node_status_allocatable{job="kube-state-metrics",resource="cpu"}) by (cluster) - max(kube_node_status_allocatable{job="kube-state-metrics",resource="cpu"}) by (cluster)) > 0'
+        expression: 'sum(namespace_cpu:kube_pod_container_resource_requests:sum{}) by (cluster) - (sum(kube_node_status_allocatable{job="kube-state-metrics",resource="cpu"}) by (cluster) - max(kube_node_status_allocatable{job="kube-state-metrics",resource="cpu"}) by (cluster)) > 0and(sum(kube_node_status_allocatable{job="kube-state-metrics",resource="cpu"}) by (cluster) - max(kube_node_status_allocatable{job="kube-state-metrics",resource="cpu"}) by (cluster)) > 0'
         for: 'PT10M'
         severity: 3
       }
       {
-        actions: [for g in actionGroups: { actionGroupId: g }]
+        actions: [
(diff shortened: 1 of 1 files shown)


dev-infrastructure/modules/metrics/rules/generatedHCPPrometheusAlertingRules.bicep
```diff
@@ -11,7 +11,15 @@ resource kasMonitorRules 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-0
     interval: 'PT1M'
     rules: [
       {
-        actions: [for g in actionGroups: { actionGroupId: g }]
+        actions: [
+          for g in actionGroups: {
+            actionGroupId: g
+            actionProperties: {
+              'IcM.Title': concat('#$.labels.cluster#', ': ', '#$.annotations.description#')
```
gouthamMN, line 18: does including a region name in title helps?

dev-infrastructure/modules/metrics/rules/generatedHCPPrometheusAlertingRules.bicep
```diff
@@ -20,14 +28,23 @@ resource kasMonitorRules 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-0
...
         for: 'PT2M'
         severity: 3
       }
       {
-        actions: [for g in actionGroups: { actionGroupId: g }]
+        actions: [
+          for g in actionGroups: {
+            actionGroupId: g
+            actionProperties: {
+              'IcM.Title': concat('#$.labels.cluster#', ': ', '#$.annotations.description#')
```
gouthamMN, line 43: same comment as above and for other instances of ICM title too.

dev-infrastructure/modules/metrics/rules/generatedPrometheusAlertingRules.bicep
```diff
@@ -27,141 +36,222 @@ resource kubernetesApps 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-03
...
         }
         annotations: {
+          correlationId: 'KubePodNotReady/{{ $labels.cluster }}/{{ $labels.namespace }}/{{ $labels.pod }}'
           description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} has been in a non-ready state for longer than 15 minutes.'
           runbook_url: 'https://runbooks.prometheus-operator.dev/runbooks/kubernetes/kubepodnotready'
           summary: 'Pod has been in a non-ready state for more than 15 minutes.'
         }
-        expression: 'sum by (namespace, pod, cluster) (   max by(namespace, pod, cluster) (     kube_pod_status_phase{job="kube-state-metrics", phase=~"Pending|Unknown|Failed"}   ) * on(namespace, pod, cluster) group_left(owner_kind) topk by(namespace, pod, cluster) (     1, max by(namespace, pod, owner_kind, cluster) (kube_pod_owner{owner_kind!="Job"})   ) ) > 0'
+        expression: 'sum by (namespace, pod, cluster) (max by(namespace, pod, cluster) (kube_pod_status_phase{job="kube-state-metrics", phase=~"Pending|Unknown|Failed"}) * on(namespace, pod, cluster) group_left(owner_kind) topk by(namespace, pod, cluster) (1, max by(namespace, pod, owner_kind, cluster) (kube_pod_owner{owner_kind!="Job"}))) > 0'
```
gouthamMN, line 59: Do we need to include region? or aksfqdn? in dimensions.

gouthamMN:
overall lgtm

dev-infrastructure/modules/metrics/rules/generatedHCPPrometheusAlertingRules.bicep
stevekuznetsov: It's a good idea - I cannot figure out what value we can use for that, though. The cluster names themselves have the regional shortname in there, so it's kind of OK? If we can figure out what value contains the region, I will add it

dev-infrastructure/modules/metrics/rules/generatedHCPPrometheusAlertingRules.bicep
stevekuznetsov: (these are autogenerated)

dev-infrastructure/modules/metrics/rules/generatedPrometheusAlertingRules.bicep
stevekuznetsov: `cluster` should be specific enough here

PR #2620, opened 2025-09-03
E2E external auth negative tests cases.

Nanyte25:
### What
E2E external auth negative test cases


raelga (commented):
Closed due to inactivity.

/close

PR #2622, opened 2025-09-03
Demonstrate some basic refactors for allowing separate document schemas.

deads2k:
This is working and it's largely good.  It's heavily invasive because the code is over-abstracted to the point where difference, even useful difference, is fought by the code.

While placing a generic would seem easy, the shape of the interfaces resist generics because you can have a generic on a single method, it must be on the receiver, so every hand coded client resists adding that way.

Given a choice between embracing difference and then putting together a better client or trying to make a better shaped client, the code benefits overall from embracing differences that are switch/cased already.

TODOs

1. somehow tests are passing?  Either the tests are pretty simplistic or something we unexpectedly.
2. confirm it is type compatible in cosmos? We could just yolo in integration if we somehow don't have the ability to use an actual cosmos instance.
3. expose the BaseDocument and TypedDocument so the information is present for internal clients
4. whatever else comes up in review

Thinking through stages of migration in the service
1. prod-release/1 start synchronizing customer owned fields from cluster-service to cosmos.  frontend writes to cluster-service then to cosmos. customer spec reads come from cluster-service.  customer status reads come from cluster-service. This will get all low throughput customer desired state into the instances.  During this phase, cluster-service is the source of truth for customer spec fields.
2. prod-release/2 frontend writes to cosmos then to cluster-service.  customer spec reads come from cosmos.   customer status reads come from cluster-service.  This state ensures that cluster-service is no longer the source of truth.
3. prod-release/3 frontend writes to cosmos.  backend writes to cluster-service. customer reads come from cluster-service.  This state removes the direct cluster-service write.
4. prod-release/4 frontend writes to cosmos, backend writes to cluster-service, sychronize status from cluster-service to cosmos,  customer spec reads come from cosmos, customer status reads come from cluster-service.  This will get all low throughput status into the instances.  During "active" times we read from cluster-service pretty aggressively.
5. prod-release/5 frontend writes to cosmos, backend writes to cluster-service, sychronize status from cluster-service to cosmos,  customer spec reads come from cosmos, customer status reads come from cosmos.  At this point the data model has shifted and we're going to be talking about bringing individual pieces of cluster-service into ARO-HCP.

Bringing individual pieces of cluster-service into ARO-HCP is probably done back to front, so that cluster-service never grows a way to say, "accept as input this thing I have provisioned", but instead cluster-service grows a way to understand, "don't perform step/X" and it's ARO-HCP's problem to stitch the bits.


internal/database/types_hcpcluster.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+// HCPClusterDocument represents a customer desired HCP Cluster.
+// To transition from our current state using cluster-service as half the source of truth to a state where
+// cosmos contains all the desired state and all the observed state, we are basing the schema on ResourceDocument.
+type HCPClusterDocument struct {
+	// do NOT marshal this field.  These are system properties that we will set on read so we can look at them, but the values
+	// in this type are never serialized because the typed document is the wrapper, not the content.
+	TypedDocument    `json:"-"`
```
deads2k, line 55: This is present so we have etag information. Hiding it makes integration with the wider cosmos ecosystem harder and doesn't benefit us since we can remove it from serialization here.

internal/database/typeddocument.go
```diff
@@ -68,30 +67,25 @@ func (td *typedDocument) getPartitionKey() azcosmos.PartitionKey {
...
 	}
 }
 
 // typedDocumentMarshal returns the JSON encoding of typedDoc with innerDoc
 // as the properties value. First, however, typedDocumentMarshal validates
 // the type field in typeDoc against innerDoc to ensure compatibility. If
 // validation fails, typedDocumentMarshal returns a typedDocumentError.
-func typedDocumentMarshal[T DocumentProperties](typedDoc *typedDocument, innerDoc *T) ([]byte, error) {
-	err := typedDoc.validateType(*innerDoc)
+func typedDocumentMarshal(typedDoc *typedDocument, innerDoc any, validResourceTypes ...string) ([]byte, error) {
```
deads2k, line 87: Similar issues about validation inside of the the serialization.

internal/database/database.go
```diff
@@ -409,10 +433,64 @@ func (d *cosmosDBClient) GetResourceDoc(ctx context.Context, resourceID *azcorea
...
+	ret.ResourceID = resourceID
+
+	return typedDoc.ID, ret, nil
+}
+
 func (d *cosmosDBClient) CreateResourceDoc(ctx context.Context, doc *ResourceDocument) error {
 	typedDoc := newTypedDocument(doc.ResourceID.SubscriptionID, doc.ResourceID.ResourceType)
 
-	data, err := typedDocumentMarshal(typedDoc, doc)
+	data, err := typedDocumentMarshal(typedDoc, doc, ResourceDocument{}.GetValidTypes()...)
```
deads2k, line 493: Once we have separate schema types, we won't have to worry about the serialization being incorrect.

internal/database/types_hcpcluster.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// newTypedDocument returns a TypedDocument from a ResourceType.
+func NewTypedDocument(partitionKey string, resourceType azcorearm.ResourceType) *TypedDocument {
+	return &TypedDocument{
+		BaseDocument: NewBaseDocument(),
+		PartitionKey: strings.ToLower(partitionKey),
+		ResourceType: strings.ToLower(resourceType.String()),
+	}
+}
```
deads2k, line 47: These only exist so the diff didn't get ridiculous by exposing the types, but it might have been clearer to do so.  At any rate, I think these should be exposed for consumers to inspect, but NOT persist.  Notice below persistence is disallowed.

frontend/pkg/frontend/frontend.go
```diff
@@ -403,6 +386,114 @@ func (f *Frontend) ArmResourceList(writer http.ResponseWriter, request *http.Req
...
+		options.ContinuationToken = api.Ptr(queryParams.Get("$skipToken"))
+		top, err := strconv.ParseInt(queryParams.Get("$top"), 10, 32)
+		if err == nil && top > 0 {
+			options.PageSizeHint = api.Ptr(int32(top))
+		}
+	}
+	return options
+}
+
+func (f *Frontend) ArmResourceListHCPClusters(writer http.ResponseWriter, request *http.Request) {
```
deads2k, line 407: This is the worst one.  Once cluster-service is gone, a per-item transform would make it about 50% tighter.

internal/database/types_hcpcluster.go
```diff
@@ -0,0 +1,96 @@
...
+		},
+	}
+}
+
+// GetValidTypes returns the valid resource types for a ResourceDocument.
+func (doc HCPClusterDocument) GetValidTypes() []string {
+	return []string{
+		api.ClusterResourceType.String(),
+	}
+}
```
mbarnes, line 79: I think you'll want to remove `api.ClusterResourceType` from `ResourceDocument.GetValidTypes` since `HCPClusterDocument` is taking over.  Might not matter for execution but just for clarity.

PR #2623, opened 2025-09-03
e2e: External Auth Logs Into KAS with entra token

bennerv:
### What

Furthers our e2e test of external auth by logging into the kube-apiserver with an entra token. 

### Why

More tests yay

🚀  Green e2e should mean :shipit:


mociarain (approved):
LGTM. Just a question about whether we want to merge this as is or wait until ARO-20830 is live.

I'm easy with either but a slight preference for just merging this as is

PR #2628, opened 2025-09-04
address IOUs from workload verification

deads2k:
tidying comments from https://github.com/Azure/ARO-HCP/pull/2601

/assign @stevekuznetsov


test/util/verifiers/serving_app.go
```diff
@@ -100,10 +100,6 @@ func (v verifySimpleWebApp) Verify(ctx context.Context, adminRESTConfig *rest.Co
 		currRoute, err := dynamicClient.Resource(gvr("route.openshift.io", "v1", "routes")).
 			Namespace(namespace.GetName()).Get(ctx, route.GetName(), metav1.GetOptions{})
 		if err != nil {
```
stevekuznetsov, line 102: usually i like if err != lastErr and yell then, you still get the sequence of unique errors on the way to failure

stevekuznetsov (approved):

PR #2631, opened 2025-09-04
frontend: Fix bugs in building ExternalAuthClaim for CS

mbarnes:
### What

Always include `GroupsClaim` and `ValidationRules` when building the `ExternalAuthClaim`, even if the they're empty. Otherwise these optional fields cannot be removed once set.

Discovered while testing #2599 on external auth resources.


hbhushan3:
LGTM

PR #2640, opened 2025-09-05
e2e: configure clients for external auth

bennerv:
### What

Configure external auth with clients for cli and console.  

### Why

Should pass once [ARO-20830](https://issues.redhat.com//browse/ARO-20830) has been rolled out to all regions.  If it doesn't pass, i'll trigger a rollout and we can merge on green CI.

### DIFF
--- a/test/e2e/external_auth_create.go
+++ b/test/e2e/external_auth_create.go
@@ -186,25 +186,24 @@ var _ = Describe("Customer", func() {
 							},
 						},
 					},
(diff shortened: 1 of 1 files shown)


bennerv:
Note: this is #2623 + one commit.  Once 2623 merges, i'll rebase on main and it'll only show the one commit.

bennerv:
INT is passing, clients are now allowed.  Going to start a stage release for frontend/backend image updates.

PR #2646, opened 2025-09-06
e2e: fix bug in constructing the armresources client factory

bennerv:
### What

The client factory incorrectly checks the wrong internal var being `nil` in armresources client factory construction.  
### Why

Tests that do not instantiate the armresources client factory before the hcpCluster client factory will fail without this change.


bennerv:
going to merge as it's late on friday and blocking #2603

PR #2651, opened 2025-09-08
Bump helm to latest (v3.18.6)

mmazur:
We've been running a way too old version.

### Why

We're hitting compatibility issues at this point and there's no chance MS is running this old a version in ev2 (which we should be matching).


venkateshsredhat (approved):
lgtm

PR #2652, opened 2025-09-08
frontend: Stop logging /healthz requests

mbarnes:
### What

In the spirit of @petrkotas's LogLight project, let's stop logging every `/healthz` request.  They just clog up frontend logs and force SREs to have to filter them out in their Kusto queries.

We have a Prometheus gauge named `"frontend_health"` that's emitted from the `/healthz` handler.  That should suffice for tracking frontend pod health, at least until we have `kubelet` logs.

I also simplified the `/healthz` handler to correct a [long-standing misbehavior](https://github.com/Azure/ARO-HCP/pull/529): we no longer factor database connectivity into our `/healthz` response.  If we're alive and we're able to respond then we're healthy.  Period.

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -27,7 +27,6 @@ import (
 	"path"
 	"strconv"
 	"strings"
-	"sync/atomic"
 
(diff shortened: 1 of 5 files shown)


mociarain (dismissed):
LGTM, current conversation ongoing

mociarain (approved):

PR #2666, opened 2025-09-10
use deployment stacks in pipelines

geoberle:
### What

leverage deployment stacks instead of deployments for bicep rollouts. this way we can do proper cleanup of orphaned resources

merge this only after deployment stack support has been added to the ev2 generator in sdp-pipelines


stevekuznetsov (approved):

stevekuznetsov:
Just synced with the team, looks like we can onboard next week.

PR #2675, opened 2025-09-11
frontend: Versioned API refactoring

mbarnes:
### What

Before we embark on more API features and _especially_ before we fork the current API version, I wanted to get in some refactoring of the `Version` interface.

### Why

Primary goal of the refactoring is to minimize code repetition across API versions.  We'll always have conversion functions for every API version but I'm trying to factor the rest of it out as much as possible.

Please review one commit at a time and take note of comments in commit messages.

See also #2683

### DIFF
--- a/internal/api/registry.go
+++ b/internal/api/registry.go
@@ -58,21 +58,9 @@ type VersionedCreatableResource[T any] interface {
 	ValidateVisibility(current VersionedCreatableResource[T], updating bool) []arm.CloudErrorBody
 }
 
-type VersionedHCPOpenShiftCluster interface {
-	VersionedCreatableResource[HCPOpenShiftCluster]
-	ValidateStatic(current VersionedHCPOpenShiftCluster, updating bool) *arm.CloudError
-}
-
-type VersionedHCPOpenShiftClusterNodePool interface {
-	VersionedCreatableResource[HCPOpenShiftClusterNodePool]
-	ValidateStatic(current VersionedHCPOpenShiftClusterNodePool, cluster *HCPOpenShiftCluster, updating bool) *arm.CloudError
-}
-
-type VersionedHCPOpenShiftClusterExternalAuth interface {
-	VersionedCreatableResource[HCPOpenShiftClusterExternalAuth]
-	ValidateStatic(current VersionedHCPOpenShiftClusterExternalAuth, updating bool) *arm.CloudError
-}
-
+type VersionedHCPOpenShiftCluster VersionedCreatableResource[HCPOpenShiftCluster]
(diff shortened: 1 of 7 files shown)


mociarain (dismissed):
LGTM. Just left a comment after playing with something for a bit

internal/api/registry.go
```diff
@@ -43,42 +43,124 @@ var (
 	VersionResourceType      = azcorearm.NewResourceType(ProviderNamespace, "locations/"+VersionResourceTypeName)
 )
 
-type VersionedCreatableResource[T any] interface {
-	Normalize(*T)
-	GetVisibility(path string) (VisibilityFlags, bool)
-	ValidateVisibility(current VersionedCreatableResource[T], updating bool) []arm.CloudErrorBody
+type Resource interface {
```
mociarain, line 46: Should this be mixed in to the types?

mociarain (commented):
Sorry, I meant to ask about this

internal/api/registry.go
```diff
@@ -43,42 +43,124 @@ var (
...
 
 	// Response Marshaling
-	MarshalHCPOpenShiftCluster(*HCPOpenShiftCluster) ([]byte, error)
-	MarshalHCPOpenShiftClusterNodePool(*HCPOpenShiftClusterNodePool) ([]byte, error)
-	MarshalHCPOpenShiftClusterExternalAuth(*HCPOpenShiftClusterExternalAuth) ([]byte, error)
 	MarshalHCPOpenShiftClusterAdminCredential(*HCPOpenShiftClusterAdminCredential) ([]byte, error)
-	MarshalHCPOpenShiftVersion(*HCPOpenShiftVersion) ([]byte, error)
+}
+
+func ValidateVersionedHCPOpenShiftCluster(incoming, current VersionedHCPOpenShiftCluster, updating bool) *arm.CloudError {
```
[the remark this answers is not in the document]
mbarnes: I hear ya, I'm taking baby steps with the generics.  Finding the right abstractions is hard.

Possibly we could reduce these down to two functions: one for top-level resource types and another for nested.  But I don't know if we have any more "creatable" resource types coming that would need validation, so I dunno if it's worth further abstraction.

internal/api/registry.go
mociarain: Sorry, "Mixin" is a very OOP term.

AFAIK there is nothing actually using this (I tested this by deleting it and it linted successfully). I would have expected to see it under the "HCPOpenShiftCluster****" structs.

mociarain: Resolved offline

PR #2687, opened 2025-09-14
api: Rename autorest-config.yaml to readme.md

mbarnes:
[ARO-21054 - Frontend: Support Multiple API Versions](https://issues.redhat.com/browse/ARO-21054)

### What

Rename `/api/autorest-config.yaml` to `/api/readme.md`, add support for multiple API versions through tags, and a few other code generation improvements.

### Why

We have a new API version on the horizon and need to make sure our tooling is ready.


SudoBrendan (approved):
While I kinda hate the fact that we're relying on a README.md file as a configuration file for autorest - if that's what MSFT expects us to, then we should follow suit rather than rolling our own.

The CI job will catch any strangeness we may experience (https://github.com/Azure/ARO-HCP/actions/runs/17718888979/job/50347769759?pr=2687) so LGTM.

PR #2701, opened 2025-09-15
add new HCPCluster fields

deads2k:
Minimal change to add fields for HCP Cluster elements.  This adds fields on all `ResourceDocument` deserialization and filters the data on write. This approach keeps the change small, but still allows storing state for HCPClusters.  For other ResourceDocuments (NodePool and ExternalAuth) the fields are force cleared.  For HCPClusters, the unknown fields are removed to prevent squatting.

Once we have this, the next step is to add a loop to the backend that every minute reads the cluster-service clusters and writes customer desired values and serviceprovider values.

The overall flow will be.
1. prod-release/1 start synchronizing customer owned fields from cluster-service to cosmos. frontend writes to cluster-service then to cosmos. customer spec reads come from cluster-service. customer status reads come from cluster-service. This will get all low throughput customer desired state into the instances. During this phase, cluster-service is the source of truth for customer spec fields.
2. prod-release/2 frontend writes to cosmos then to cluster-service. customer spec reads come from cosmos. customer status reads come from cluster-service. This state ensures that cluster-service is no longer the source of truth.
3. prod-release/3 frontend writes to cosmos. backend writes to cluster-service. customer reads come from cluster-service. This state removes the direct cluster-service write.
4. prod-release/4 frontend writes to cosmos, backend writes to cluster-service, sychronize status from cluster-service to cosmos, customer spec reads come from cosmos, customer status reads come from cluster-service. This will get all low throughput status into the instances. During "active" times we read from cluster-service pretty aggressively.
5. prod-release/5 frontend writes to cosmos, backend writes to cluster-service, sychronize status from cluster-service to cosmos, customer spec reads come from cosmos, customer status reads come from cosmos. At this point the data model has shifted and we're going to be talking about bringing individual pieces of cluster-service into ARO-HCP.

### DIFF
--- a/internal/database/typeddocument_test.go
+++ b/internal/database/typeddocument_test.go
@@ -15,10 +15,13 @@
 package database
 
 import (
+	"encoding/json"
 	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/Azure/ARO-HCP/internal/api"
 )
 
 const testResourceType = "test"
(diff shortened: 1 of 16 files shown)


mbarnes:
Can you clarify what you mean by "customer spec reads" and "customer status reads".  I understand you're thinking in Kubernetes terms.  I'm trying to map it to Azure terms.

In Azure terms does "customer spec reads" mean customer-initiated GET requests on an Azure resource
```
GET /subscriptions/.../resourceGroups/.../Microsoft.RedHatOpenShift/hcpOpenShiftClusters/${name}
```

and "customer status reads" mean ARM-initiated GET requests for status on an ongoing async operation?
```
GET /subscriptions/.../Microsoft.RedHatOpenShift/locations/.../hcpOperationStatuses/${uuid}
```

Or have I misunderstood what "status reads" means?

internal/database/typeddocument_test.go
```diff
@@ -109,3 +112,144 @@ func TestTypedDocumentUnmarshal(t *testing.T) {
...
+}
+
+// must is a helper function that takes a value and error, returns the value if no error occurred,
+// or panics if an error occurred. This is useful for test setup where we don't expect errors.
+func must[T any](v T, err error) T {
+	if err != nil {
+		panic(err)
+	}
+	return v
+}
```
mbarnes, line 238: Minor, but it jumped out at me.  Could write this instead as:
```suggestion
func must[T any](t testing.T, v T, err error) T {
    require.NoError(t, err)
	return v
}
```
where `require` is "[github.com/stretchr/testify/require](https://pkg.go.dev/github.com/stretchr/testify/require)".

I thought of writing a helper function like this but couldn't think of a good name for it.  "Must" is perfect.  I might move this to `/internal/api/testhelpers.go` so it's more widely available.

internal/database/types_hcpcluster.go
```diff
@@ -0,0 +1,91 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+)
+
+type CustomerDesiredHCPClusterState struct {
+	// HCPOpenShiftCluster contains the desired state from a customer.  It is filtered to only those fields that customers
+	// are able to set.
+	// We will whitelist specific fields as we go through to prevent conflicts.
```
mbarnes, line 27: I think I'll need you to clarify what you have in mind with the filter/whitelist comments.  It sounds like it might overlap with some validation checks we already do.

internal/database/types_hcpcluster.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal filterObj: %w", err)
+	}
+	filteredMap := map[string]any{}
+	if err := json.Unmarshal(filteredBytes, &filteredMap); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal into filtered map: %w", err)
+	}
+
+	return filteredMap, nil
+}
```
mbarnes, line 91: I'm still getting my head around this and don't fully know what you have in mind for it, but just a thought:

I recently started using [github.com/evanphx/json-patch](https://pkg.go.dev/github.com/evanphx/json-patch) to properly apply customer PATCH request bodies and I wonder if that could be beneficial here.

In other words, maybe write `filterObj` as a JSON Merge Patch and apply it with [jsonpatch.MergePatch](https://pkg.go.dev/github.com/evanphx/json-patch#MergePatch).  Don't know that it would be any less expensive but maybe could simplify the design?

deads2k:
Pushed for comments.  diff of diffs here: https://github.com/Azure/ARO-HCP/compare/949ac6f3df4e83e9cf27d4d611e9659ba930698f..3f2dd79b1929f39b1ffd7047280ec9705d925d92

PR #2714, opened 2025-09-16
Add the mandatory KMS managed identity to the cluster definition in the cluster-creation.md

anshulvermapatel:
### What

The cluster creation doc here [cluster-service/cluster-creation.md](https://github.com/Azure/ARO-HCP/blob/main/cluster-service/cluster-creation.md) is missing the KMS identity creation and its usage in the cluster definition. Having KMS managed identity is mandatory with customer managed etcd encryption.
This PR adds that missing managed identity to the doc.


machi1990 (approved):
lgtm

@patilsuraj767 can you take a look as well?

patilsuraj767 (approved):
LGTM 👍

PR #2722, opened 2025-09-17
tooling:Fix OutputOnly check

rachelvweber:
https://issues.redhat.com/browse/ARO-21143

### What

OutputOnly was failing locally for me on some steps that only have bicep with existing resources when those bicep files use modules. This recursively checks modules to confirm that they have only existing resources.

### DIFF
--- a/tooling/templatize/internal/end2end/e2e_test.go
+++ b/tooling/templatize/internal/end2end/e2e_test.go
@@ -476,7 +476,6 @@ output parameterA string = parameterA`,
 		"",
 		"")
 
-
 	e2eImpl.EnableDryRun()
 
 	persistAndRun(t, e2eImpl)


tooling/templatize/internal/end2end/e2e_test.go
```diff
@@ -410,3 +410,34 @@ param parameterA = 'Hello Bicep'`,
...
+	if !shouldRunE2E() {
+		t.Skip("Skipping end-to-end tests")
+	}
+
+	tmpDir := t.TempDir()
+
+	e2eImpl, err := newE2E(tmpDir, "../../testdata/e2eOutputOnly.yaml")
+	require.NoError(t, err)
+
+	bicepFile := `
```
stevekuznetsov, line 425: You mentioned a nested module being the trigger, but I don't see that here - did you want this e2e to cover that case?

rachelvweber: There's no way to run these locally for us, yeah?

tooling/templatize/internal/end2end/e2e.go
```diff
@@ -170,9 +170,11 @@ func (e *e2eImpl) Persist() (*run.RawRunOptions, error) {
 				return nil, err
 			}
 
-			err = os.WriteFile(e.tmpdir+"/"+b.paramFileName, []byte(b.paramFile), 0644)
-			if err != nil {
-				return nil, err
+			if (len(b.paramFile) != 0) || (len(b.paramFileName) != 0) {
```
stevekuznetsov, line 173: I'm not sure I follow this logic - in what case is either the file name or the content empty? If the content is empty, why would we care? If the file name is empty but the content is non-empty, why should we write an empty file at `e.tempDir+"/"`?

rachelvweber: This is for the modules that don't have parameter files. I can change this to fail here if the filename is empty and the content is not, but I figured we would already return an error on line 176 in that case

PR #2724, opened 2025-09-17
Improve frontend e2e tests

hbhushan3:
• Fix admin_credential_lifecycle.go to use cluster-only.bicep template instead of manual multi-step deployment
• Replace hardcoded "uksouth" locations with tc.Location() in both test files


mbukatov (dismissed):
Thank you for polishing the test.

PR #2730, opened 2025-09-18
test: increase timeout for ext auth access token retrieval

JameelB:
[ARO-21082](https://issues.redhat.com/browse/ARO-21082)

### What
Follows on from https://github.com/Azure/ARO-HCP/pull/2723. Increases time to 5 minutes since the stage e2e parallel failed with the same error.

### Why
There's a delay to propagate aad/msgraph. If we attempt to fetch a token and it fails, it should wait for a bit longer. Fixes [ARO-21082](https://issues.redhat.com/browse/ARO-21082)


JameelB:
Looked at the wrong test, no need to increase to 5m.

PR #2733, opened 2025-09-18
feat: make CS K8s Deployment rolling update strategy configurable

miguelsorianod:
We add the ability to configure the rolling update strategy attributes for the components.

We introduce that configuration explicitly for Clusters Service. We start by introducing a default of maxSurge=25% and maxUnavailable=25% for Clusters Service in all environments, which is what we currently have as it is the default rolling update strategy but we make it explicit.

Additionally, we change the configuration for CS in dev based environments to be in absolute numbers instead of percentage. Specifically, we set maxSurge=1 and maxUnavailable=0. That should ensure that no matter the number of replicas we only deploy a new replica at a time and we only delete an old replica once a new replica has been deployed. This is the same behavior as what we have with maxSurge=25%, maxUnavailable=25% in our envs because of the current number of replicas there at the moment of writing this. Using absolute numbers should ensure that if in the future the number of replicas changes we will still have the same behavior. We intend to change this for all the other envs to the same absolute values.

In the future we can decide to relax this or moving it back to percentage based but when doing it we should carefully evaluate the implications and introduce it gradually to our environments making sure that there are no unintended side effects.

### DIFF
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -17,6 +17,53 @@
       "pattern": "^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$",
       "description": "A valid RFC1123 DNS subdomain. Multiple RFC1123 DNS label devided by a dot."
     },
+    "intOrString": {
+       "description": "IntOrString is a type that can hold an int32 or a string.  When used in JSON or YAML marshalling and unmarshalling, it produces or consumes the inner type. This allows you to have, for example, a JSON field that can accept a name or number.",
+       "format": "int-or-string",
+       "oneOf": [
+         {
+           "type": "integer"
+         },
+         {
+           "type": "string"
+         }
+       ]
+    },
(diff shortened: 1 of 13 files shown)


machi1990 (dismissed):

config/config.schema.json
```diff
@@ -17,6 +17,53 @@
       "pattern": "^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$",
       "description": "A valid RFC1123 DNS subdomain. Multiple RFC1123 DNS label devided by a dot."
     },
+    "intOrString": {
```
geoberle, line 20: can we go with `intOrPercent` and enhance the `string` type to match a regex

miguelsorianod: The reason I used "int or string" was because that's what K8s defines at openapi level for it so I went for that approach.

It does not go into details about the format, ranges, ... even for the rolling strategy values.

geoberle: ok. thanks for looking.

geoberle (approved):

PR #2734, opened 2025-09-18
add new hcpcluster, externalauth, and nodepool types with compatible serialization

deads2k:
This also makes use a new library for fuzzing structs so that we can ensure compatibility of conversion and roundtripping.


deads2k:
> Asking out of ignorance but also curious: I'm aware of but not very familiar with the standard testing package's support for fuzzing (verbosely named [F](https://pkg.go.dev/testing#F)). Is [sigs.k8s.io/randfill](https://pkg.go.dev/sigs.k8s.io/randfill) objectively a better fit for this purpose or was it just a familiarity choice?

The base golang fuzz package doesn't natively do structs (or at least google couldn't find it and AI made a huge diff making it).  Kube's already exists and does what needed, so we reused.

PR #2736, opened 2025-09-18
add a specific client for HCPClusters

deads2k:
Adds and uses a database client for clusters in the frontend.  The `frontend/hack/test-simulation.sh` shows this is byte compatible for reads.

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -109,7 +109,7 @@ type DBClient interface {
 	PatchBillingDoc(ctx context.Context, resourceID *azcorearm.ResourceID, ops BillingDocumentPatchOperations) error
 
 	// GetHCPClusterCRUD retrieves a CRUD interface for managing HCPCluster resources and their nested resources.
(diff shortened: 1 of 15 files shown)


internal/database/database.go
```diff
@@ -53,9 +55,9 @@ func NewPartitionKey(subscriptionID string) azcosmos.PartitionKey {
 	return azcosmos.NewPartitionKeyString(strings.ToLower(subscriptionID))
 }
 
-type DBClientIteratorItem[T DocumentProperties] iter.Seq2[string, *T]
+type DBClientIteratorItem[T any] iter.Seq2[string, *T]
 
-type DBClientIterator[T DocumentProperties] interface {
+type DBClientIterator[T any] interface {
```
deads2k, line 60: @deads2k to be sure this was intentional.

deads2k:
passed `frontend/hack/test-simulation.sh` still working on automated CI runs

PR #2737, opened 2025-09-18
[wip] synchronize customer desired cluster state from cluster-service

deads2k:
This starts a control loop that reads from cluster-service and writes the customer desired state into cosmos.

before merge, we need
1. mechanism to store controller state per key (subscription, resourcegroup, control loop, hcp cluster, subresource)
2. probably break off the database client changes
3. other?

Future, we need
1. mechanism to dump the current state
2. refactor the launch to be more integrate testable
3. make it trivial to add more controllers

### DIFF
--- a/backend/controllers/cs_customer_sync.go
+++ b/backend/controllers/cs_customer_sync.go
@@ -1,0 +1,234 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+
+	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
+
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+)
+
+type Controller interface {
+	Run(ctx context.Context, threadiness int)
+}
+
+type hcpClusterKey struct {
+	subscriptionID  string
+	resourceGroupID string
+	hcpClusterID    string
+}
+
+type clusterServiceToCustomerSync struct {
+	cosmosClient database.DBClient
+
+	clusterServiceClient ocm.ClusterServiceClientSpec
+
+	// queue is where incoming work is placed to de-dup and to allow "easy"
+	// rate limited requeues on errors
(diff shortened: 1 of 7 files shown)


backend/controllers/cs_customer_sync.go
```diff
@@ -0,0 +1,144 @@
...
+	// queue is where incoming work is placed to de-dup and to allow "easy"
+	// rate limited requeues on errors
+	queue workqueue.TypedRateLimitingInterface[hcpClusterKey]
+}
+
+func NewClusterServiceToCustomerSyncController(cosmosClient database.DBClient, clusterServiceClient ocm.ClusterServiceClientSpec) Controller {
+	c := &clusterServiceToCustomerSync{
+		cosmosClient:         cosmosClient,
+		clusterServiceClient: clusterServiceClient,
+		queue: workqueue.NewTypedRateLimitingQueueWithConfig(
```
stevekuznetsov, line 41: For which events do you want to be triggered? CosmosDB has changefeeds, unclear if useful to you.

internal/database/crud_helpers.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// Replace the key field from Cosmos with the given resourceID,
+	// which typically comes from the URL. This helps preserve the
+	// casing of the resource group and resource name from the URL
+	// to meet RPC requirements:
+	//
+	// Put Resource | Arguments
+	//
+	// The resource group names and resource names should be matched
+	// case insensitively. ... Additionally, the Resource Provier must
```
stevekuznetsov, line 85:
```suggestion
	// case insensitively. ... Additionally, the Resource Provider must
```

internal/database/crud_toplevel_resource.go
```diff
@@ -0,0 +1,103 @@
...
+		d.resourceType.Namespace,
+	}
+
+	parts = append(parts, d.resourceType.Type)
+
+	if len(resourceID) > 0 {
+		parts = append(parts, resourceID)
+	}
+
+	return azcorearm.ParseResourceID(path.Join(parts...))
```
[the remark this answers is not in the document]
deads2k: > Can't you just create an arm.ResourceID literal? Why round trip through string formatting & parsing? (applies generally)

As someone coming in, it's very easy to understand and search for the individual items and pass them properly.  The single literal is harder to work through the meaning for and very easy to pass the wrong thing through.  And it's difficult to detect the wrong value coming through during the review.

backend/controllers/cs_customer_sync.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// start up your worker threads based on threadiness.  Some controllers
+	// have multiple kinds of workers
+	for i := 0; i < threadiness; i++ {
+		// runWorker will loop until "something bad" happens.  The .Until will
+		// then rekick the worker after one second
+		go wait.UntilWithContext(ctx, c.runWorker, time.Second)
+	}
+
+	go wait.JitterUntilWithContext(ctx, c.queueAllHCPClusters, time.Minute, 0.1, true)
```
nimrodshn: Why queue _all_ clusters every min? 

Theoretically - you are doing _much_ more work than is required as only a subset of the clusters might have changed over the duration of a single minute. 
Alternatively you could -
1. call `queueAllHcpCluster` - "Sync all world" - _on startup_ of the controller thread. This is required so that the application is crash resilient.
2. Than, allow the frontend to *queue* cluster keys [sync requests] when it receives a PUT on a cluster resource. Thus, you will only sync clusters desired state when you know it has changed. Here is where you need to be careful to queue only after the frontend issued the REST request to CS. [this is what I raised in the meeting yesterday, as I (Naively) imagined 1+2 as you're solution].
3. Both the current solution and the alternative raise in 1+2 raise the following questions - 

- When will the RP manage the desired state? 
- How will we deprecate this mechanism?

nimrodshn: > Given two separate processes, how would you do this?

There are several IPC solutions out there (The maestro team uses Azure Event Grid. In our testing we had used both RabbitMQ, I also have some experience with Kafka, etc.) . In fact, [the solution described here is the solution advocated by Azure!](https://eng.ms/docs/products/arm/rpaas/development_guides/long_running_operations).

PR #2745, opened 2025-09-19
msft env promotion script

geoberle:
### What

as long as we keep the MSFT env settings seperate, we need some way to promote consistent sets of images between environments and report on it.

`make -C config bump-msft-stg`
`make -C config bump-msft-prod`


hack/msft-env-promotion.sh
```diff
@@ -0,0 +1,313 @@
...
+    
+    # Find and display PR information for this digest (only for actually updated digests)
+    echo "      Looking up PR information for digest..."
+    
+    # Find the commit that introduced this digest (with timeout to prevent hanging)
+    echo "      Searching git history..."
+    pr_info=""
+    commit_info=""
+    
+    if commit_hash=$(timeout 15s git log --format="%H" -S "$source_digest" -- "$CONFIG_FILE" 2>/dev/null | head -1); then
```
stevekuznetsov, line 150: 😵

stevekuznetsov:
Didn't we do like 99% of the work to stop having per-env config? What's left to kill this whole process and simplify our life? I'd rather see movement on that front

raelga:
Replaced by https://github.com/Azure/ARO-HCP/pull/3577

PR #2746, opened 2025-09-19
dev-infrastructure: don't use outputs from ARM steps

stevekuznetsov:
We want to allow for incremental releases to skip steps when running them would not yield a different outcome from the last time they successfully ran. However, we have a couple of places where large ARM deployment steps output variables that are chained into other steps. If any of those derivative steps needs to run, the big ARM deployment will be pulled into the execution graph as well to provide the output variables. By using a dedicated output-only step instead, we allow for the large deployment step to be skipped and only require the output step.


dev-infrastructure/templates/svc-infra-lookup.bicep
```diff
@@ -0,0 +1,9 @@
+@description('The name of the service keyvault')
+param serviceKeyVaultName string
+
+resource serviceKeyVault 'Microsoft.KeyVault/vaults@2024-04-01-preview' existing = {
+  name: serviceKeyVaultName
```
geoberle, line 5: there is a config override for the resourcegroup the svc keyvault lives in

https://github.com/Azure/ARO-HCP/blob/ab4df6c10a830a8c9f4f0bf8afe8094f693a6170/dev-infrastructure/configurations/svc-cluster.tmpl.bicepparam#L91

https://github.com/Azure/ARO-HCP/blob/1bc7db417f314c36f428f70ee7b4633891f182ac/dev-infrastructure/templates/svc-infra.bicep#L59

reason: in DEV the service KV is shared

PR #2753, opened 2025-09-19
secret-sync-controller: move to native Helm step

stevekuznetsov:
We can omit the different API version handling logic in the Helm chart as we're only ever going to deploy to new versions of Kubernetes.


secret-sync-controller/deploy/values-azure.yaml
```diff
@@ -1,6 +1,7 @@
+image: '{{ .acr.svc.name }}.azurecr.io/{{ .secretSyncController.image.repository }}@{{ .secretSyncController.image.digest }}'
```
roivaz, line 1: Some values from the deleted values.yaml are required

```suggestion
image: '{{ .acr.svc.name }}.azurecr.io/{{ .secretSyncController.image.repository }}@{{ .secretSyncController.image.digest }}'
imagePullPolicy: IfNotPresent

# Configuration for validating admission policies
validatingAdmissionPolicies:
  applyPolicies: true
  allowedSecretTypes:
  - "Opaque"
  - "kubernetes.io/basic-auth"
  - "bootstrap.kubernetes.io/token"
  - "kubernetes.io/dockerconfigjson"
  - "kubernetes.io/dockercfg"
  - "kubernetes.io/ssh-auth"
  - "kubernetes.io/tls"
  deniedSecretTypes:
  - "kubernetes.io/service-account-token"

# Default values for controller configuration
controllerName: secrets-store-sync-controller-manager
tokenRequestAudience:
- audience: # e.g. api://TokenAudienceExample

# Metrics configuration
metricsPort: 8085
```

secret-sync-controller/deploy/templates/secrets-store-sync-controller.yaml
```diff
@@ -122,7 +122,7 @@ spec:
       {{- toYaml .Values.providerContainer | nindent 6 }}
       {{- end }}
       - name: manager
-        image: {{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}
+        image: {{ .Values.image }}
         imagePullPolicy: '{{ .Values.image.pullPolicy }}'
```
roivaz, line 126:
```suggestion
        imagePullPolicy: '{{ .Values.imagePullPolicy }}'
```

roivaz (changes requested):
Tested the dry-run for this and it seems that some vars from the deleted values.yaml file are required. I added suggestions to fix that.

PR #2769, opened 2025-09-22
Update personal-dev.md

sclarkso:
add yq as required utility

### What

yq is required to setup personal dev environment, but isn't listed as required utility.  Commands complete without error if it is NOT installed, but environment doesn't work.

### DIFF
--- a/docs/personal-dev.md
+++ b/docs/personal-dev.md
@@ -21,6 +21,7 @@ Ensure you have access to the RH Azure tenant:
 The following additional tools are also required:
 
 - `jq`
+- `yq`
 - `make`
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
@sclarkso @geoberle  please, instead of this MD - make the `$(YQ)` target a required pre-requisite of the `make` targets that use it, so it's installed at the right version and updated as necessary without users having to do anything. We should be aiming for a zero-touch dev setup.

Please see this for an example of how a shell target can manage it's own dependencies:

https://github.com/Azure/ARO-HCP/blob/dc08b93fee42c938c90fb64dda5ea911c1bf260e/Makefile#L44-L46

PR #2781, opened 2025-09-23
finish internal validation for HCPClusters

deads2k:
builds on https://github.com/Azure/ARO-HCP/pull/2767 and completes that validation.


frontend/test/simulate/artifacts/ClusterMutation/create-with-illegal-value/expected-errors.txt
```diff
@@ -1 +1,2 @@
-InvalidRequestContent: properties.network.hostPrefix: Invalid value '30' for field 'hostPrefix' (must be at most 26)
+InvalidRequestContent: properties.network.hostPrefix: properties.network.hostPrefix: Invalid value: 30: must be less than or equal to 26
+InvalidRequestContent: properties.dns.baseDomainPrefix: properties.dns.baseDomainPrefix: Invalid value: "": (must be a valid DNS RFC 1035 label)
```
deads2k, line 2: interestingly, it seems that previously this error was not caught in ARO-HCP, but I believe it matches the annotations.

frontend/test/simulate/artifacts/ClusterMutation/change-channel/create.json
```diff
@@ -12,7 +12,9 @@
       "state": "Disabled"
     },
     "console": {},
-    "dns": {},
+    "dns": {
+      "baseDomainPrefix": "example"
+    },
```
deads2k, line 17: previous validation in ARO-HCP seems to have missed this.

frontend/test/simulate/artifacts/ClusterMutation/missing-required-fields/expected-errors.txt
```diff
@@ -1,10 +1,14 @@
...
-InvalidRequestContent: properties.etcd.dataEncryption.customerManaged.kms.activeKey.name: Missing required field 'name'
-InvalidRequestContent: properties.etcd.dataEncryption.customerManaged.kms.activeKey.vaultName: Missing required field 'vaultName'
-InvalidRequestContent: properties.etcd.dataEncryption.customerManaged.kms.activeKey.version: Missing required field 'version'
-InvalidRequestContent: identity.type: Missing required field 'type'
+InvalidRequestContent: properties.platform.subnetId: properties.platform.subnetId: Required value
+InvalidRequestContent: properties.platform.networkSecurityGroupId: properties.platform.networkSecurityGroupId: Required value
+InvalidRequestContent: properties.platform.operatorsAuthentication.userAssignedIdentities.controlPlaneOperators: properties.platform.operatorsAuthentication.userAssignedIdentities.controlPlaneOperators: Required value
+InvalidRequestContent: properties.platform.operatorsAuthentication.userAssignedIdentities.controlPlaneOperators[]: properties.platform.operatorsAuthentication.userAssignedIdentities.controlPlaneOperators[]: Required value
```
deads2k, line 4: message change.  Required and format aren't the same thing.

frontend/test/simulate/artifacts/ClusterMutation/missing-required-fields/expected-errors.txt
```diff
@@ -1,10 +1,14 @@
...
+InvalidRequestContent: properties.platform.operatorsAuthentication.userAssignedIdentities.dataPlaneOperators: properties.platform.operatorsAuthentication.userAssignedIdentities.dataPlaneOperators: Required value
+InvalidRequestContent: properties.platform.operatorsAuthentication.userAssignedIdentities.dataPlaneOperators[]: properties.platform.operatorsAuthentication.userAssignedIdentities.dataPlaneOperators[]: Required value
+InvalidRequestContent: properties.etcd.dataEncryption.customerManaged.kms.activeKey.name: properties.etcd.dataEncryption.customerManaged.kms.activeKey.name: Required value
+InvalidRequestContent: properties.etcd.dataEncryption.customerManaged.kms.activeKey.name: properties.etcd.dataEncryption.customerManaged.kms.activeKey.name: Invalid value: "": must be at least 1 characters long
```
deads2k, line 8: minlength and required were specified.  Really want both messages?

frontend/test/simulate/artifacts/ClusterMutation/some-immutable-field-checks/expected-errors.txt
```diff
@@ -1,12 +1,15 @@
-InvalidRequestContent: properties.platform.networkSecurityGroupId: Field 'networkSecurityGroupId' cannot be updated
-InvalidRequestContent: properties.platform.subnetId: Field 'subnetId' cannot be updated
-InvalidRequestContent: properties.platform.managedResourceGroup: Field 'managedResourceGroup' cannot be updated
-InvalidRequestContent: properties.etcd.dataEncryption.customerManaged.kms.activeKey.name: Field 'name' cannot be updated
-InvalidRequestContent: properties.etcd.dataEncryption.customerManaged.kms.activeKey.vaultName: Field 'vaultName' cannot be updated
-InvalidRequestContent: properties.etcd.dataEncryption.customerManaged.kms.activeKey.version: Field 'version' cannot be updated
```
deads2k: notice 
> InvalidRequestContent: properties.etcd: properties.etcd: Forbidden: field is immutable

in the diff.  We have to decide whether these field are immutable for every type used under dataencryption no matter where the type is used or if the immutability is property of how the type is used here.

Similar for several other fields.  This was conflated before.

frontend/test/simulate/artifacts/ClusterMutation/some-immutable-field-checks/expected-errors.txt
```diff
@@ -1,12 +1,15 @@
...
+InvalidRequestContent: properties.platform.managedResourceGroup: properties.platform.managedResourceGroup: Forbidden: field is immutable
+InvalidRequestContent: properties.etcd: properties.etcd: Forbidden: field is immutable
+InvalidRequestContent: properties.network: properties.network: Forbidden: field is immutable
+InvalidRequestContent: properties.network.hostPrefix: properties.network.hostPrefix: Forbidden: field is immutable
+InvalidRequestContent: properties.network.machineCidr: properties.network.machineCidr: Forbidden: field is immutable
+InvalidRequestContent: properties.network.networkType: properties.network.networkType: Forbidden: field is immutable
+InvalidRequestContent: properties.network.podCidr: properties.network.podCidr: Forbidden: field is immutable
+InvalidRequestContent: properties.network.serviceCidr: properties.network.serviceCidr: Forbidden: field is immutable
+InvalidRequestContent: properties.version.id: properties.version.id: Forbidden: field is immutable
+InvalidRequestContent: properties.version.id: properties.version.id: Invalid value: "illegal-change": Malformed version: illegal-change
```
deads2k, line 15: previous validation skipped checking the values.

internal/validation/validate_cluster.go
```diff
@@ -371,6 +371,12 @@ func validateUserAssignedIdentitiesProfile(ctx context.Context, op operation.Ope
 		newObj.ControlPlaneOperators, safe.Field(oldObj, toPlatformOperatorsAuthenticationUserAssignedIdentitiesControlPlaneOperators),
 		validate.RequiredValue,
 	)...)
+	// even though it's not listed, prior validation had the value required.
```
deads2k, line 374: quirk of old validation.  Having a format for an optional field is definitely a thing in general.

frontend/test/simulate/artifacts/ClusterMutation/change-channel/create.json
deads2k: So baseDomainPrefix is not reuired, so fix the validation code.

mbarnes (commented):
As I'm getting my head around the Kubernetes validation packages I really like the changes here.  Had some questions though before I approve.  I'm still not clear about exactly how we deal with API versions.

Also a few suggestions but they can be deferred.

internal/api/arm/error.go
```diff
@@ -79,6 +80,53 @@ type CloudErrorBody struct {
...
+func cloudErrorBodyFromFieldError(fieldError *field.Error) *CloudErrorBody {
+	if fieldError == nil {
+		return nil
+	}
+	return &CloudErrorBody{
+		Code:    CloudErrorCodeInvalidRequestContent,
+		Message: fieldError.Error(),
+		Target:  fieldError.Field,
+	}
+}
```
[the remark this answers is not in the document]
deads2k: > I understand you're preserving 1:1 compatibility in this PR, but as a future enhancement I'd like to use the more descriptive `field.ErrorType` values as cloud error codes (or at least map them to our own error codes) instead of throwing them away for `CloudErrorCodeInvalidRequestContent`.

This is ok, but it's worth discussing at an arch call.  Since automated remediation of errors is unlikely and humans are going to have to resolve the issues, who benefits from more granularity than "you messed up"?

frontend/test/simulate/artifacts/ClusterMutation/change-channel/TODO
```diff
@@ -1,3 +1,4 @@
 1. allegedly only "stable" is allowed, but there is no error when updating to different value in ARO-HCP.
+2. why does the baseDomainPrefix not need to be set when issuing the update?
```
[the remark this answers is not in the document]
deads2k: > Did we address this in a call or is it still an open question?

We found the line of code that add causes it to happen, but I think we still have an open question of "why".  Mostly likely answer (I think): This field is automatically set by the service and thus the user always submits it empty.  If that is the case, we need better docs.

internal/validation/validators.go
```diff
@@ -0,0 +1,218 @@
...
+		return field.ErrorList{field.Invalid(fldPath, *value, "resource group is required")}
+	}
+	if len(resourceID.Name) == 0 {
+		return field.ErrorList{field.Invalid(fldPath, *value, "resource name is required")}
+	}
+	if !strings.EqualFold(resourceTypeRestriction, resourceID.ResourceType.String()) {
+		return field.ErrorList{field.Invalid(fldPath, *value, fmt.Sprintf("resource ID must reference an instance of type %q", resourceTypeRestriction))}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
deads2k: If we do that, then the resulting function isn't a ValidationFunc per the interface.  I'm open to having them both call a private function.  Followup?

internal/validation/validate_cluster.go
```diff
@@ -0,0 +1,648 @@
...
+
+func ValidateClusterCreate(ctx context.Context, newCluster *api.HCPOpenShiftCluster) field.ErrorList {
+	op := operation.Operation{Type: operation.Create}
+	return validateCluster(ctx, op, newCluster, nil)
+}
+
+func ValidateClusterUpdate(ctx context.Context, newCluster, oldCluster *api.HCPOpenShiftCluster) field.ErrorList {
+	op := operation.Operation{Type: operation.Update}
+	return validateCluster(ctx, op, newCluster, oldCluster)
+}
```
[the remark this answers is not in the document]
deads2k: > The problem is when we add a new field to the next API version ("2025-12-23-preview"), it will be present for that version and all subsequent versions. So a simple check like

Don't later API versions have the option of removing fields and making them unsettable?

PR #2786, opened 2025-09-24
Don't do PR checks for ai context PRs v2

mmazur:
The "let's not break CI this time" edition.


venkateshsredhat (approved):

raelga:
The policy is to test everything all the time, even is there is no relation between the change and the test.

PR #2788, opened 2025-09-24
Spike log redaction

janboll:


tooling/log-redaction/main.go
```diff
@@ -0,0 +1,209 @@
...
+
+	opts := &Options{
+		Stamp: 1,
+	}
+
+	cmd.Flags().StringVar(&opts.ServiceConfigFile, "service-config-file", "config/config.yaml", "Path to the service configuration file.")
+	cmd.Flags().StringVar(&opts.RedactionConfigFile, "redaction-config-file", "config/config.redaction.yaml", "Path to the redaction configuration file.")
+	cmd.Flags().StringVar(&opts.Cloud, "cloud", "dev", "The name of the cloud.")
+	cmd.Flags().StringVar(&opts.Environment, "environment", "dev", "The name of the environment.")
+	cmd.Flags().StringVar(&opts.Region, "region", "eastus2", "The name of the region.")
```
deads2k, line 78: Can we simply include for all regions to reduce variation?

tooling/log-redaction/main.go
```diff
@@ -0,0 +1,209 @@
...
+		},
+	}
+
+	opts := &Options{
+		Stamp: 1,
+	}
+
+	cmd.Flags().StringVar(&opts.ServiceConfigFile, "service-config-file", "config/config.yaml", "Path to the service configuration file.")
+	cmd.Flags().StringVar(&opts.RedactionConfigFile, "redaction-config-file", "config/config.redaction.yaml", "Path to the redaction configuration file.")
+	cmd.Flags().StringVar(&opts.Cloud, "cloud", "dev", "The name of the cloud.")
```
deads2k, line 76: What are the different possible values of cloud?  Fewer options are better.

tooling/log-redaction/main.go
```diff
@@ -0,0 +1,209 @@
...
+	cmd.Flags().StringVar(&opts.Environment, "environment", "dev", "The name of the environment.")
+	cmd.Flags().StringVar(&opts.Region, "region", "eastus2", "The name of the region.")
+	cmd.Flags().IntVar(&opts.Stamp, "stamp", opts.Stamp, "Stamp value to use.")
```
deads2k, line 79: what is a stamp?

deads2k (approved):
Starting anywhere is good.  Having one redaction config for int, one for stg, and one for prod would be ideal to avoid accidents, rather than building one for each region.

config/config.redaction.yaml
```diff
@@ -0,0 +1,28 @@
+keys:
+  - administration.readerGroupId
```
stevekuznetsov, line 2: Crazy thought - why not every UUID?

tooling/log-redaction/main.go
stevekuznetsov: There are about nine clouds, and the count is always growing. If we are successful with this product, it will deploy to them all. Public cloud is the obvious one - security-hardened ones exist, as do sovereigns.

tooling/log-redaction/main.go
stevekuznetsov: When some deployment unit can be replicated in a region many times, the stamp number is used to differentiate between the replicas. I don't imagine this will need to be set by the user of this tool, we could resolve the config for stamp 0 without loss of generality

tooling/log-redaction/main.go
```diff
@@ -0,0 +1,209 @@
...
+	redactionConfigTemplate.Execute(os.Stdout, map[string]interface{}{
+		"redactionKeys": redactionKeys,
+	})
+
+	return nil
+}
+
+func resolveRedactionKeys(cfg config.Configuration, redactionConfig redactionConfig) ([]string, error) {
+	redactionKeys := make([]string, 0)
+	for _, redactionKey := range redactionConfig["keys"] {
```
stevekuznetsov, line 149: Clever, but a flat list with some repetition for keys with common prefixes would be easier to read, only mildly harder to write, and much easier to reason about as opposed to this :)

config/config.redaction.yaml
stevekuznetsov: e.g. use the regex support in the sanitizer to just nuke all UUID-looking things. The approach here will be forever-whack-a-mole.

stevekuznetsov: Also, wouldn't customer subscription IDs, etc, be considered PII?

janboll: Is it only UIDs we want to redact? Let me try figure that out. If so, we can do a uid regex

tooling/log-redaction/main.go
janboll: After sleeping over it, agreed. I'll try to find a simpler approach

config/config.redaction.yaml
janboll: What about log correlation IDs? I think simply filtering all uids won't work in that case

tooling/log-redaction/main.go
janboll: I'll remove this parameter

tooling/log-redaction/main.go
janboll: I'll remove this parameter and loop over all/consider all variations in one go

PR #2792, opened 2025-09-24
make basic frontend test against cosmos

deads2k:
adds a basic starting point for using the cosmos emulator to run a set of golang tests (called simulator tests) with mock cluster-service.

`frontend/hack/test-simulation.sh`

/assign @mbarnes


frontend/hack/stop-all-cosmos-emulators.sh
```diff
@@ -0,0 +1,51 @@
...
+        ${CONTAINER_RUNTIME} logs "$container" > "${TMP_DATA_DIR}/${container_name}.log" 2>&1 || true
+    fi
+done
+
+# Stop and remove all containers
+echo "Stopping containers..."
+echo "$CONTAINERS" | xargs -r ${CONTAINER_RUNTIME} stop
+
+echo "Removing containers..."
+echo "$CONTAINERS" | xargs -r ${CONTAINER_RUNTIME} rm
```
stevekuznetsov, line 48: Are there volumes in the container? Are you keeping the volumes on purpose?

frontend/test/simulate.run/frontend simulations.run.xml
```diff
@@ -0,0 +1,18 @@
+<component name="ProjectRunConfigurationManager">
```
stevekuznetsov, line 1: What is this?

deads2k: makes goland slightly easier to use for launching tests in debug mode.

frontend/test/simulate/utils.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/Azure/ARO-HCP/frontend/pkg/util"
+	"github.com/Azure/ARO-HCP/internal/audit"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/mocks"
+)
+
+type SimulationTestInfo struct {
+	CosmosDatabaseClient     *azcosmos.DatabaseClient
+	DBClient                 database.DBClient
+	MockClusterServiceClient *mocks.MockClusterServiceClientSpec
```
mbarnes, line 45: nit: Could we define this as `ocm.ClusterServiceClientSpec` and not assume it's mocked?

I may want to leverage this framework for basic interaction tests w/ Cluster Service (in demo mode) in the interim before we drop it.

frontend/test/simulate/launch_test.go
```diff
@@ -0,0 +1,47 @@
...
+	require.NoError(t, err)
+	defer testInfo.Cleanup(context.Background())
+
+	stop := make(chan struct{})
+	signalChannel := make(chan os.Signal, 1)
+	signal.Notify(signalChannel, syscall.SIGINT, syscall.SIGTERM)
+	go frontend.Run(ctx, stop)
+
+	// run for a little bit and don't crash
+	time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
deads2k: >Can you elaborate? Are these go func races something that needs fixed in the frontend?

No, they are "natural" races caused by the `go func` and a few thread spawn. Very normal and reasonable for a server initialization.

mbarnes (changes requested):
Just a couple nits.

Also, what do you think about running the Cosmos emulator by way of a Docker compose file instead of `docker/podman run` commands?  It's overkill for what you have here, but I can envision lots of validation and other tests where I would want to run at least the RP frontend and maybe also Cluster Service as containers alongside the Cosmos emulator.

PR #2794, opened 2025-09-25
ARO-21025: do not set version in e2e tests

ahitacat:
[ARO-21025](https://issues.redhat.com/browse/ARO-21025)

### What

Remove hardcoded version specifications from cluster and nodepool templates to allow for more flexible version management.

  - Demo templates: Removed hardcoded OpenShift version 4.19 and channel group stable from cluster and nodepool Bicep templates and JSON templates
  - Test templates: Removed hardcoded version specifications from E2E test Bicep modules for clusters and nodepools

### Why

ARO-HCP should transition to use fast channel group instead of stable. Tests are being use across all the environments, this transition will not be available in all the envs at the same time. Thus, to make the transition less disruptive, tests will not specify the version configuration.

This way, RP and CS will select the default channel group. So e2e will not fail during the transition from 'stable' to 'fast' in all environments (int, staging, prod).


mbukatov (approved):
Assuming the RP in all environments is able to handle the missing channel without error.

PR #2801, opened 2025-09-25
add testing for frontend access

deads2k:
Add test that does an actual frontend read.

/assign @mbarnes

### DIFF
--- a/frontend/test/simulate/frontend_test.go
+++ b/frontend/test/simulate/frontend_test.go
@@ -1,0 +1,71 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package simulate
+
+import (
+	"context"
+	"encoding/json"
+	"io/fs"
+	"os"
+	"testing"
+
+	csarhcpv1alpha1 "github.com/openshift-online/ocm-api-model/clientapi/arohcp/v1alpha1"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+)
+
+func TestFrontendClusterRead(t *testing.T) {
+	if os.Getenv("FRONTEND_SIMULATION_TESTING") != "true" {
+		t.Skip("Skipping test")
+	}
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	frontend, testInfo, err := NewFrontendFromTestingEnv(ctx, t)
+	require.NoError(t, err)
+	defer testInfo.Cleanup(context.Background())
+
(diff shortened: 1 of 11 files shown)


frontend/test/simulate/frontend_test.go
```diff
@@ -0,0 +1,76 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	frontend, testInfo, err := NewFrontendFromTestingEnv(ctx, t)
+	require.NoError(t, err)
+	defer testInfo.Cleanup(context.Background())
+
+	stop := make(chan struct{})
+	signalChannel := make(chan os.Signal, 1)
+	signal.Notify(signalChannel, syscall.SIGINT, syscall.SIGTERM)
```
mbarnes, line 48: Do we need this `os.Signal` channel for tests?  It's not being used here.

frontend/test/simulate/frontend_test.go
```diff
@@ -0,0 +1,76 @@
...
+	go frontend.Run(ctx, stop)
+
+	err = testInfo.CreateInitialCosmosContent(ctx, api.Must(fs.Sub(artifacts, "artifacts/ClusterReadOldData/initial-cosmos-state")))
+	require.NoError(t, err)
+
+	clusterServiceCluster, err := csarhcpv1alpha1.UnmarshalCluster(api.Must(artifacts.ReadFile("artifacts/ClusterReadOldData/initial-cluster-service-state/02-some-cluster.json")))
+	require.NoError(t, err)
+	testInfo.MockClusterServiceClient.EXPECT().GetCluster(gomock.Any(), api.Must(ocm.NewInternalID("/api/clusters_mgmt/v1/clusters/fixed-value"))).Return(clusterServiceCluster, nil)
```
mbarnes, line 56: nit for the future:
We should use `aro_hcp/v1alpha1` instead of `clusters_mgmt/v1` here and in test data.  Only because `ocm.NewInternalID` will likely stop recognizing the `clusters_mgmt` API in the foreseeable future and these tests will break when that day comes.

frontend/test/simulate/testinfo.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
+	"github.com/google/uuid"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	hcpapi20240610 "github.com/Azure/ARO-HCP/internal/api/v20240610preview/generated"
```
mbarnes, line 35: This will break when #2739 merges.  Import this instead:
```suggestion
	hcpsdk "github.com/Azure/ARO-HCP/test/sdk/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
```

mbarnes (dismissed):
Approved with a few nits, but I recommend at least changing the client SDK import.  That may break as early as next week.

frontend/test/simulate/testinfo.go
deads2k: >This will break when https://github.com/Azure/ARO-HCP/pull/2739 merges. Import this instead:

I have deep reservations about this suggestion.  The goal of this test is to ensure that `v20240610preview` works.  And for any test suite, coding against an exact level ensures we never look at it again and if it ever breaks, we've broken compatibility for customers.

I understand the desire to have a single client to use and continuously update and for internal service code, by all means.  But for testing, we need to use a generated external client that matches what customers use and if it ever stops working, then we have a stop-ship bug.

deads2k:
Comments addressed.  Merging in now.

PR #2803, opened 2025-09-26
Geneva Action Enablement

geoberle:
### What

    * create dedicated KV for the geneva actions certificate
    * grant the geneva actions app permissions to act in the KV based on documentation found in https:/
    * create certificate in the KV
    * create secret to allowlist extensions

### DIFF
--- a/config/dev.digests.yaml
+++ b/config/dev.digests.yaml
@@ -3,19 +3,19 @@ clouds:
     environments:
       cspr:
         regions:
-          westus3: 8ceab2b36eb9f8a4857dced92d62442b6a0ad896195779bbeae8aff733974426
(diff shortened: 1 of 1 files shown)


tony-schndr (dismissed):

katherinelc321 (approved):

config/rendered/dev/cspr/westus3.yaml
```diff
@@ -210,6 +210,21 @@ frontend:
     address: ""
     exporter: ""
 geneva:
+  actions:
+    allowedAcisExtensions: AzureRedHatHypershiftExtension
+    certificate:
+      issuer: OneCertV2-PrivateCA
```
bennerv, line 216: Do we want self issuer here?  OneCert wont' work.

bennerv: I guess it's not enabled so it doesn't matter.

PR #2806, opened 2025-09-26
Adding additional field to msi credential refresher to allow early re…

rachelvweber:
### What

Adding field for early refresh msi credential frequency to be used in staging.

### Why

By default, credentials are only eligible for refresh every 45 days. This will allow us to test the entire refresh flow more frequently.

### DIFF
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -1397,6 +1397,10 @@
           "type": "string",
           "description": "The client ID of the first party application that will be used by the refresher"
         },
+        "earlyRefreshFrequency": {
+          "type": "string",
+          "description": "Early refresh interval before credential is eligible for refresh. For testing. 0 means no early refresh. Example: 24h"
+        },
         "k8s": {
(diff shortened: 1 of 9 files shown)


config/config.schema.json
```diff
@@ -1397,6 +1397,10 @@
           "type": "string",
           "description": "The client ID of the first party application that will be used by the refresher"
         },
+        "earlyRefreshFrequency": {
+          "type": "string",
+          "description": "Early refresh interval before credential is eligible for refresh. For testing. 0 means no early refresh. Example: 24h"
```
weinong, line 1402:
```suggestion
          "description": "Early refresh interval before credential is eligible for refresh. For testing, 0 means no early refresh. Example: 24h"
```

rachelvweber: Refresh interval is how frequently the refresher loop runs. By default it's 10 minutes

weinong (approved):
lgtm

PR #2817, opened 2025-09-29
fix: Frontend and Backend alerts

raelga:
### What

- Bump the Frontend 95th percentile latency threshold to 5s
- Updates avalability alerts for Frontend and Backend to take in account the 30s scrape time
- Removes the Backend latency alert, team confirmed that BackendOperationErrorRate and BackendHealthAvailability should be the ones used

### Why

Currently, there are several false alerts and transistent alerts creating ICM incidents due to alert misconfiguraiton.

- https://redhat-internal.slack.com/archives/C07A2CVUV44/p1758628276177509
- https://redhat-internal.slack.com/archives/C07A2CVUV44/p1758293403317739

cc @mbarnes @hbhushan3  @swiencki


mbarnes (approved):
My PromQL is pretty rusty, but I think this looks reasonable based on what you described as the intended changes.  Backend latency checks were probably just copy/pasted from the frontend.

bennerv (approved):

PR #2823, opened 2025-09-29
Adding fpa tenant id to credential refresher

rachelvweber:
### What

Adding fpa tenant id to credential refresher
### Why

Getting auth errors when I try to auth with the first party app. They say I need to specify the tenant when I make the client credential.

### DIFF
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -1397,6 +1397,10 @@
           "type": "string",
           "description": "The client ID of the first party application that will be used by the refresher"
         },
+        "firstPartyAppTenantId": {
+          "type": "string",
+          "description": "The client ID of the first party application that will be used by the refresher"
+        },
         "earlyRefreshFrequency": {
           "type": "string",
           "description": "Early refresh interval before credential is eligible for refresh. For testing, 0 means no early refresh. Example: 24h"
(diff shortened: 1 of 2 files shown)


config/config.schema.json
```diff
@@ -1397,6 +1397,10 @@
           "type": "string",
           "description": "The client ID of the first party application that will be used by the refresher"
         },
+        "firstPartyAppTenantId": {
+          "type": "string",
+          "description": "The client ID of the first party application that will be used by the refresher"
```
bennerv, line 1402: nit: 

```suggestion
          "description": "The tenant ID of the first party application that will be used by the refresher"

rachelvweber:
> @rachelvweber we're running the refresher with a workload identity correct? Should we just use the AZURE_TENANT_ID envvar injected by the workload identity webhook?

I just realized there's already a field for this, my bad, so I'm going to close my pr. But it's a different tenant than the app's primary  identity

bennerv:
> I just realized there's already a field for this, my bad, so I'm going to close my pr. But it's a different tenant than the app's primary identity
Yeah the `AZURE_TENANT_ID` env var would be the local tenant, which is what we'd want to authenticate to talk to MSI-RP

PR #2824, opened 2025-09-29
stop tracking go.work.sum

deads2k:
this file constantly conflicts for me and the general advice appears to be not checking it in:

> The go.work.sum file should generally not be checked into version control.
The go.work file and its accompanying go.work.sum file are primarily intended for local development convenience within a Go workspace. 

/assign @stevekuznetsov


miquelsi:
Thank you! I completely agree.

PR #2849, opened 2025-10-01
add fuzz testing for frontend internal to external to internal types

deads2k:
We cannot start from external yet because nil vs zero-value-pointer doesn't actually roundtrip (existing condition).


deads2k:
wrong branch. opened https://github.com/Azure/ARO-HCP/pull/2850

PR #2850, opened 2025-10-01
add internal external internal fuzz tests.

deads2k:
replaces https://github.com/Azure/ARO-HCP/pull/2849 with right branch.

We cannot start from external yet because nil vs zero-value-pointer doesn't actually roundtrip (existing condition).

### DIFF
--- a/internal/api/v20240610preview/hcpopenshiftclusters_methods_test.go
+++ b/internal/api/v20240610preview/hcpopenshiftclusters_methods_test.go
@@ -1,0 +1,72 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


internal/api/v20240610preview/hcpopenshiftclusters_methods_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/google/go-cmp/cmp"
+	"sigs.k8s.io/randfill"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+)
+
+func TestRoundTripInternalExternalInternal(t *testing.T) {
+	seed := rand.Int63()
```
stevekuznetsov, line 30: Can you make failing tests print their seed and provide an ergonomic way to re-run given a seed, so failures can be squashed and validated? Not sure if it's simple to do, but if you're able to wire this in to Go's built-in fuzzer, you'll get some of this for free.

deads2k: I wasn't able to wire up go's built in fuzzer for the same purpose, I'd end up rebuilding Daniel's impl.  I'll find a way to print the seed as a followup.

deads2k:
IOU promised and ok'd via slack.

PR #2856, opened 2025-10-02
DNS documentation

geoberle:
### What

document the DNS architecture in ARO HCP

https://issues.redhat.com/browse/ARO-18259

### DIFF
--- a/docs/dns.md
+++ b/docs/dns.md
@@ -61,7 +61,7 @@ Example: `uksouthstaging.aroapp-hcp.io`
 
 DNS zones are created as part of the infrastructure deployment process:
 
-- **Parent zones**: Managed in `global-infra.bicep`, delegation from the parent-parent zones is managed through EV2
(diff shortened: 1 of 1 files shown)


docs/dns.md
```diff
@@ -0,0 +1,135 @@
...
+    regionalSubdomain: "{{ .ctx.region }}staging"
+```
+
+Example: `uksouthstaging.aroapp-hcp.io`
+
+### Zone Management and Deployment
+
+DNS zones are created as part of the infrastructure deployment process:
+
+- **Parent zones**: Managed in `global-infra.bicep`, delegation from the parent-parent zones is managed through EV2
```
roivaz, line 64: > delegation from the parent-parent zones is managed through EV2

what's the process involved here? could you add a few words clarifying this?

roivaz (commented):
Great info. Left you just a minor question about parent zones and EV2.

docs/dns.md
geoberle: added some clarification

PR #2858, opened 2025-10-02
update nodepool.bicep template to allow autoscale

mukrishn:
Update nodepool bicep template to enabled autoscale, would be useful for future e2e test case for [ARO-22022](https://issues.redhat.com//browse/ARO-22022).


tony-schndr:
Instead of creating a new template just for auto scale, could you enhance the existing template?
https://github.com/Azure/ARO-HCP/blob/main/demo/bicep/nodepool.bicep

raelga (commented):
@mukrishn Does this still apply?

/retest

mukrishn:
we can close this, autoscaler test is included in this https://github.com/Azure/ARO-HCP/pull/4133

PR #2861, opened 2025-10-02
Make defaulting a per-external version concern

deads2k:
This makes it possible for default values to evolve over time instead of pinning them to the default values of the internal version.  This is also a step in separating the idea of registry (how do create, convert, and validate types) from instances of the types themselves.

I found this while working through mechanisms for per-external version validation path mapping.


bennerv (approved):
lgtm.  Changing default values is a breaking api change, but if we had to do it, we could do it at an API boundary now.

PR #2873, opened 2025-10-03
Remove acrpull role assignment to kubelet identity

bennerv:
### What

We're still granting the kubelet identity permissions over the whole ACRs in question.  We need to start using msi-acrpull where applicable. 

I'm opening this in draft because for the ocp registry on mgmt cluster we may need this role assignment temporarily still.  But this PR should merge at some point with it fully plumbed for svc cluster.

Checks:
 - [ ] All images can be pulled on service cluster
     - [ ]  MISE Missing ACRPullBinding
     - [ ] Backend Missing ACRPullBinding
 - [ ] Method to delete role bindings on kubelet if we're not using complete deployments or a way to remove role assignments on our aks clusters


bennerv:
/close 

Stale, will reopen later as time permits.

PR #2890, opened 2025-10-06
frontend: Temporarily disable the new cluster validation

mbarnes:
Just until we sort out a regression.


deads2k:
/lgtm

but also needs to fix frontend tests since it' snot a revert.

PR #2891, opened 2025-10-06
make batch processs config configurable based on env

mahesh-palkar:
https://issues.redhat.com/browse/ARO-21112


config/config.schema.json
```diff
@@ -1028,6 +1028,13 @@
         "manage"
       ]
     },
+    "batchProcessesDryRun": {
+      "type": "boolean",
```
miguelsorianod, line 1032: If we define this as a boolean, what happens if I don't provide the attribute at all in config.yaml?

miguelsorianod:
Have we verified if this works end to end with a personal svc/mgmt cluster?

PR #2895, opened 2025-10-06
add patch test for frontend simulation

deads2k:
Done in response to the panic, but since systemdata is filled in by the system it was neat, but didn't expose the problem.

### DIFF
--- a/frontend/test/simulate/cluster_mutation_test.go
+++ b/frontend/test/simulate/cluster_mutation_test.go
@@ -98,13 +98,21 @@ func (tt *clusterMutationTest) runTest(t *testing.T) {
 	clusterClient := tt.testInfo.Get20240610ClientFactory(tt.subscriptionID).NewHcpOpenShiftClustersClient()
 	_, mutationErr := clusterClient.BeginCreateOrUpdate(ctx, tt.resourceGroupName, *toCreate.Name, *toCreate, nil)
 
-	if tt.genericMutationTestInfo.isUpdateTest() {
+	if tt.genericMutationTestInfo.isUpdateTest() || tt.genericMutationTestInfo.isPatchTest() {
 		require.NoError(t, mutationErr)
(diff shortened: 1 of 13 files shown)


frontend/test/simulate/cluster_mutation_test.go
```diff
@@ -98,14 +98,22 @@ func (tt *clusterMutationTest) runTest(t *testing.T) {
 	clusterClient := tt.testInfo.Get20240610ClientFactory(tt.subscriptionID).NewHcpOpenShiftClustersClient()
 	_, mutationErr := clusterClient.BeginCreateOrUpdate(ctx, tt.resourceGroupName, *toCreate.Name, *toCreate, nil)
 
-	if tt.genericMutationTestInfo.isUpdateTest() {
+	if tt.genericMutationTestInfo.isUpdateTest() || tt.genericMutationTestInfo.isPatchTest() {
 		require.NoError(t, mutationErr)
 		require.NoError(t, MarkOperationsCompleteForName(ctx, tt.testInfo.DBClient, tt.subscriptionID, ptr.Deref(toCreate.Name, "")))
+	}
 
+	switch {
```
stevekuznetsov, line 106: Doesn't smell like `func (tt *clusterMutationTest) runTest(t *testing.T) {` is factored well, but if this is expedient for you, sure. Probably some other factoring that separates these out would be easier

frontend/test/simulate/mutation_test_utils.go
```diff
@@ -107,6 +114,62 @@ func trivialPassThroughClusterServiceMock(t *testing.T, testInfo *SimulationTest
 	}).AnyTimes()
 }
 
+func addFakeAzureIdentityData(clusterServiceCluster any) (*csarhcpv1alpha1.Cluster, error) {
+	// the API is so hard to work with that we'll make it a map[string]any to manipulate it
```
stevekuznetsov, line 118: :dizzy_face:

PR #2898, opened 2025-10-06
update logging middleware to include subscription, resourcegroup, and hcpcluster

deads2k:
This allows us to grep for these items in a structured when when we're search for messages about a particular resourcegroup, subscription, or related to an hcp cluster.  This will make a log collection tool far more valuable

/assign @mbarnes

### DIFF
--- a/frontend/pkg/frontend/middleware_logging.go
+++ b/frontend/pkg/frontend/middleware_logging.go
@@ -20,8 +20,10 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"strings"
 	"time"
 
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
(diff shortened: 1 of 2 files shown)


frontend/pkg/frontend/middleware_logging.go
```diff
@@ -71,13 +73,45 @@ func MiddlewareLogging(w http.ResponseWriter, r *http.Request, next http.Handler
 	logger = logger.With(
 		"request_method", r.Method,
 		"request_path", r.URL.Path,
+	)
+
+	// make a best attempt at parsing the resourceID. This will often fail because we have non-resource requests.
+	// we do this so that we can add subscription, resourceGroup, and hcpCluster to the logger context for future searching
+	// if possible.
+	// It's important to do before the second panic handler so that panics can be correlated easily.
+	// TODO are the value we find case sensitive or case insensitive.  They used to be case sensitive, so I have left that
```
deads2k, line 82: if they aren't case sensitive, then our search is much easier if we tolower everything

frontend/pkg/frontend/middleware_logging_test.go
```diff
@@ -115,13 +121,15 @@ func TestMiddlewareLoggingPostMux(t *testing.T) {
 				w.WriteHeader(http.StatusOK)
 			}
 
-			MiddlewareLoggingPostMux(writer, req, next)
+			MiddlewareLogging(writer, req, func(w http.ResponseWriter, r *http.Request) {
+				MiddlewareLoggingPostMux(w, r, next)
+			})
```
deads2k, line 126: wrapped because that's what our router.go actually does and the first set of context is added much earlier now.

frontend/pkg/frontend/middleware_logging.go
```diff
@@ -71,13 +73,45 @@ func MiddlewareLogging(w http.ResponseWriter, r *http.Request, next http.Handler
...
+	if resourceID, err := azcorearm.ParseResourceID(r.URL.Path); err == nil {
+		logger = logger.With(
+			"subscription_id", resourceID.SubscriptionID,
+			"resource_group", resourceID.ResourceGroupName,
+		)
```
mbarnes, line 87: Should we have a `"resource_id"` attribute as well?  [We do in the backend](https://github.com/Azure/ARO-HCP/blob/1b060917ac40f43c98b345d6d06c24f671e77311/backend/operations_scanner.go#L440), so that would be a common field name to search.  (Note, backend also has `"subscription_id"` but not `"resource_group"`.)

frontend/pkg/frontend/middleware_logging.go
```diff
@@ -71,13 +73,45 @@ func MiddlewareLogging(w http.ResponseWriter, r *http.Request, next http.Handler
...
 		"request_proto", r.Proto,
 		"request_query", r.URL.RawQuery,
+		// TODO referrer is under the client's control.  Printing it out could be huge.
 		"request_referer", r.Referer(),
 		"request_remote_addr", r.RemoteAddr,
-		"request_user_agent", r.UserAgent())
-
-	logger.Info("read request")
+		// TODO user agent is under the client's control.  Printing it out could be huge.
+		"request_user_agent", r.UserAgent(),
```
mbarnes, line 113: Should we truncate the string if it's larger than some reasonable limit?
(If ARM does this automatically, I couldn't find any docs stating so.)

frontend/pkg/frontend/middleware_logging.go
deads2k: > Should we have a "resource_id" attribute as well? [We do in the backend](https://github.com/Azure/ARO-HCP/blob/1b060917ac40f43c98b345d6d06c24f671e77311/backend/operations_scanner.go#L440), so that would be a common field name to search. (Note, backend also has "subscription_id" but not "resource_group".)

oh, I see. The idea that we could do a best effort here to help earlier in the stack?  I'd say it's a good followup.

frontend/pkg/frontend/middleware_logging.go
```diff
@@ -71,13 +73,45 @@ func MiddlewareLogging(w http.ResponseWriter, r *http.Request, next http.Handler
...
+	}
+
+	// include the context values (logger.With) with every line so we can grep for them.
+	ctx = ContextWithLogger(ctx, logger)
+	r = r.WithContext(ctx)
+
+	logger.Info("read request",
 		"request_proto", r.Proto,
 		"request_query", r.URL.RawQuery,
+		// TODO referrer is under the client's control.  Printing it out could be huge.
```
mbarnes, line 109: It's actually not.  ARM provides the referer header to RPs, overriding the client's value if present.

PR #2902, opened 2025-10-07
cloud specific CS operator role sets in helm chart

geoberle:
### What

declare premade sets of operator roles per deployment cloud (`dev` being the RH tenant, `prod` being AME). config.yaml only specifies the roleSetName to be used for a specific environment.

also the lookup of the operator role IDs has been removed. the IDs are now specified along with the role names in the values.yaml file. reasoning: we need to switch to the native helm step soon where such lookups with shell commands are not supported. the usual way would be lookup in bicep and forward the data to the helm chart, but roles can't be looked up by name via bicep yet. there is a fix in the working on bicep side to support that.

### Why

due to parameter count limitations we need to specify the operator role names within the helm values file instead of the configuration.


cluster-service/Makefile
```diff
@@ -149,24 +148,17 @@ personal-runtime-config:
...
 	helm template deploy \
-	--set ocpVersions.defaultVersion.channelGroupName=${OCP_VERSIONS_DEFAULT_VERSION_CHANNEL_GROUP_NAME} \
 	--set ocpVersions.defaultVersion.version=${OCP_VERSIONS_DEFAULT_VERSION_VERSION} \
-	--set ocpVersions.channelGroups.stable.enabled=${OCP_VERSIONS_CHANNEL_GROUPS_STABLE_ENABLED} \
 	--set ocpVersions.channelGroups.stable.url=${OCP_VERSIONS_CHANNEL_GROUPS_STABLE_URL} \
 	--set ocpVersions.channelGroups.stable.minVersion=${OCP_VERSIONS_CHANNEL_GROUPS_STABLE_MIN_VERSION} \
 	--set ocpVersions.channelGroups.stable.maxVersion=${OCP_VERSIONS_CHANNEL_GROUPS_STABLE_MAX_VERSION} \
-	--set ocpVersions.channelGroups.fast.enabled=${OCP_VERSIONS_CHANNEL_GROUPS_FAST_ENABLED} \
-	--set ocpVersions.channelGroups.fast.url=${OCP_VERSIONS_CHANNEL_GROUPS_FAST_URL} \
-	--set ocpVersions.channelGroups.fast.minVersion=${OCP_VERSIONS_CHANNEL_GROUPS_FAST_MIN_VERSION} \
```
geoberle, line 168: unrelated: leftover from previous fast channel cleanup

PR #2912, opened 2025-10-07
draft: kust-must-gather

janboll:
### What

Init KQL Queries for hcpctl


tooling/hcpctl/cmd/must-gather/cmd.go
```diff
@@ -0,0 +1,259 @@
...
+	validated, err := opts.Validate(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Complete options
+	completed, err := validated.Complete(ctx)
+	if err != nil {
+		return err
+	}
```
deads2k, line 133: Structurally, attaching these to `RawMustGatherOptions` as the method recevier and doing them in the cobra `Run` is more canonical from kube.  Additionally, that makes it more practical to chain commands if we ever need to and unit test them.

So
```
func (o *RawMustGather) Validate() error
func (o *RawMustGather) Complete() error
func (o *RawMustGather) Run() error
```

tooling/hcpctl/cmd/must-gather/cmd.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Create Kusto client
+	client, err := kusto.NewClient(opts.KustoEndpoint)
```
deads2k, line 136: Creating this client in the `Complete` and then using it in the `Run` will make unit testing `.Run` with a mock much easier.

tooling/hcpctl/pkg/kusto/client.go
```diff
@@ -0,0 +1,164 @@
...
+		rowData := reflect.New(reflect.TypeOf(rowType)).Interface()
+		err := row.ToStruct(rowData)
+		if err != nil {
+			return fmt.Errorf("failed to convert row to struct: %w", err)
+		}
+		// from the docs https://pkg.go.dev/github.com/Azure/azure-kusto-go#readme-query-into-structs
+		if row.Replace {
+			data = data[:0]
+		}
+		data = append(data, rowData)
```
deads2k, line 132: As a starting point, by all means let's do this.  Our next iteration needs to take this an instead of holding all the data in memory, we need to stream it out to disk.

mmazur (approved):

PR #2970, opened 2025-10-13
split cert for genevaactions bicep

katherinelc321:
### What

Separates cert creation from kv creation bicep


geoberle:
superseeded by #3003

PR #2971, opened 2025-10-13
Clean up validation messages

deads2k:
This adds map key reporting and eliminates the duplicate field path in the errors, though it's missing from messages now.  All validation tests now use the same verification for fieldpath and messages.  empty strings aren't allowed anymore and all errors must be listed.

builds on https://github.com/Azure/ARO-HCP/pull/2957

### DIFF
--- a/internal/validation/validate_cluster_test.go
+++ b/internal/validation/validate_cluster_test.go
@@ -74,15 +74,6 @@ func verifyErrorsMatch(t *testing.T, expectedErrors []expectedError, errs field.
 	}
 }
 
-func containsError(errs field.ErrorList, expectedErr expectedError) bool {
-	for _, err := range errs {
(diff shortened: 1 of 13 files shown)


internal/validation/validate_cluster_test.go
```diff
@@ -29,13 +29,49 @@ type expectedError struct {
...
+	for _, expectedErr := range expectedErrors {
+		if len(strings.TrimSpace(expectedErr.fieldPath)) == 0 {
+			t.Errorf("expected error with path %s to be non-empty", expectedErr.fieldPath)
+		}
+		if len(strings.TrimSpace(expectedErr.message)) == 0 {
+			t.Errorf("expected error with msg %s to be non-empty", expectedErr.message)
+		}
+
+		found := false
+		for _, err := range errs {
```
stevekuznetsov: Oh, nvm, you scan both

mbarnes (approved):

PR #3006, opened 2025-10-15
configure new aro-hcp cs sandbox quay repo caching

miguelsorianod:
We are going to move the aro-hcp cs sandbox images to a new repository as part of ocm cs split.

### DIFF
--- a/dev-infrastructure/configurations/acr-svc.tmpl.bicepparam
+++ b/dev-infrastructure/configurations/acr-svc.tmpl.bicepparam
@@ -10,12 +10,25 @@ param quayRepositoriesToCache = [
     userIdentifier: 'quay-componentsync-username'
     passwordIdentifier: 'quay-componentsync-password'
   }
+  {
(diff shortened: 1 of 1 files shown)


dev-infrastructure/configurations/acr-svc.tmpl.bicepparam
```diff
@@ -10,6 +10,13 @@ param quayRepositoriesToCache = [
     userIdentifier: 'quay-componentsync-username'
     passwordIdentifier: 'quay-componentsync-password'
   }
+  {
+    ruleName: 'aroHcpCsSandboxImages'
+    sourceRepo: 'quay.io/app-sre/aro-hcp-clusters-service-sandbox'
+    targetRepo: 'app-sre/aro-hcp-clusters-service-sandbox'
+    userIdentifier: 'quay-componentsync-username'
+    passwordIdentifier: 'quay-componentsync-password'
+  }
```
machi1990: Why do we need to do this? The sandboxes images are for cs pr checks only.

PR #3034, opened 2025-10-21
Add the same roles mmazur has to 3 au members

trevorwilliams2025:
Add three new members to all entries where @mmazur appears:
@sclarkso
@trevorwilliams2025
@weherdh


trevorwilliams2025:
closed intentionally, maybe overzealous

PR #3039, opened 2025-10-21
prepare cs image f0568be3327d for rollout in int

miguelsorianod:


miguelsorianod:
Do not merge this MR until the image has been rolled out to shared dev beforehand

miguelsorianod:
The MR is ready as the image has been rolled out to shared dev

PR #3046, opened 2025-10-21
[wip] allow serializing internal objs to storage

deads2k:
This serializes our internal types to storage and is exposing shortcomings where what we store is not roundtrippable back to the internal types.

Clearly must be fixed and tested prior to changing.


deads2k:
corrected in other PRs.

PR #3047, opened 2025-10-21
add inline API type docs

deads2k:
We're well down this path now and the type system has achieved a little more clarity.  I've written up a small section here.

/assign @mbarnes

### DIFF
--- a/internal/README.md
+++ b/internal/README.md
@@ -1,0 +1,57 @@
+# Background Knowledge
+
+## Actors
+
+1. REST Client - typically `az`.  Can be anything able to make the HTTP request.
+2. ARO-HCP/frontend - receives REST Client requests via External API, converts to Internal API, validates, converts to Cosmos API, and writes.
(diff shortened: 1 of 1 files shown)


internal/README.md
```diff
@@ -0,0 +1,57 @@
...
+   They are used inside frontend and backend.
+
+On a REST Client read, the frontend will go through Cosmos API -> Internal API -> External API.
+
+## Major Packages
+
+1. validation - holds all static validation rules written against internal APIs.
+   The only information available is the new and old internal API instance.
+   Validation only runs inside of the frontend apiserver, NOT on the backing cosmos storage.
+   This means the backend can directly modify content is ways that bypass all validation rules.
```
stevekuznetsov:
```suggestion
   This means the backend can directly modify content in ways that bypass all validation rules.
```

raelga:
@deads2k Does this still apply? 

/retest

PR #3048, opened 2025-10-21
set back CS image from quay uhc repository in shared dev

miguelsorianod:
Tests in shared dev to verify the deploy flow can reference the new CS quay repository images have finished.

We revert back to the CS image that was used before the tests.


miguelsorianod:
This is a conceptual revert of https://github.com/Azure/ARO-HCP/pull/3040

bennerv (approved):

PR #3051, opened 2025-10-22
adding sclarkso and wenqih to /config in codeowners

sclarkso:
need added to /config in codeowners to be able to do promotions.


trevorwilliams2025 (approved):
lgtm

PR #3057, opened 2025-10-22
Set batchProcessesDryRun to false for batch job ARO-20736

nimrodshn:


miguelsorianod:
The MR was opened from the fork which doesn't work. I am going to close it and open a new one with the changes.

PR #3058, opened 2025-10-22
clean:  remove cluster service alerts

raelga:
### What

Removes the CS monitoring rules as they are triggerind false alerts and not properly monitoring the service. 

### Why

The current alerts, are triggering most of the time and no reporting properly the SLO.


raelga:
cc @roivaz

roivaz (approved):

PR #3059, opened 2025-10-22
additional api ci validators/linters

tony-schndr:
https://issues.redhat.com/browse/ARO-22151

### What

Adds additional validators and linters for the api.

### Why

To reduce the number of iterations required when upstreaming the api to azure rest specs.


SudoBrendan (approved):
I think if those warnings are relevant, we can issue a follow-up PR - so I'm good to merge. Great work, thanks Tony!

PR #3070, opened 2025-10-23
Add ServiceProvider fields to internal API

deads2k:
This adds CosmosUID for replace/patch operations and ClusterServiceID.  It also aligns types on the internal API so that we don't have to convert to/from ResourceIDs and InternalIDs everywhere we use them.

This is a precursor to being able to use InternalAPI for more direct REST operations instead of generic ResourceDocument.

### DIFF
--- a/internal/api/registry.go
+++ b/internal/api/registry.go
@@ -16,6 +16,7 @@ package api
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -48,10 +49,10 @@ func init() {
 				return a.UTC().Equal(b.UTC())
 			},
 			func(a, b azcorearm.ResourceID) bool {
-				return a.String() == b.String()
+				return strings.EqualFold(a.String(), b.String())
 			},
 			func(a, b azcorearm.ResourceType) bool {
-				return a.String() == b.String()
+				return strings.EqualFold(a.String(), b.String())
 			},
 			func(a, b InternalID) bool {
 				return a.String() == b.String()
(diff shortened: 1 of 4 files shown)


internal/api/registry.go
```diff
@@ -17,12 +17,49 @@ package api
...
 
+var (
+	// we have some private fields that cmp.Diff chokes on. This skips them.
+	// Use only for error messages unless you know what you're doing.
+	CmpDiffOptions = []cmp.Option{
+		cmpopts.IgnoreFields(azcorearm.ResourceID{}, "ResourceType"),
+		cmpopts.IgnoreFields(azcorearm.ResourceID{}, "isChild"),
+		cmpopts.IgnoreFields(azcorearm.ResourceID{}, "stringValue"),
+		cmpopts.IgnoreFields(InternalID{}, "path"),
+		cmpopts.IgnoreFields(InternalID{}, "kind"),
```
mbarnes: Would [cmpopts.IgnoreUnexported](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreUnexported) work here instead of enumerating them all?

internal/api/v20240610preview/external_auth_methods.go
```diff
@@ -68,7 +70,7 @@ func (h *ExternalAuth) GetVersion() api.Version {
 
 func (h *ExternalAuth) Normalize(out *api.HCPOpenShiftClusterExternalAuth) {
 	if h.ID != nil {
-		out.ID = *h.ID
+		out.ID = api.Must(azcorearm.ParseResourceID(*h.ID))
```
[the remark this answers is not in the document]
deads2k: > `Normalize` methods were intended to never fail, which is why I left the internal field type as `string` and deferred the parsing to validation.
> 
> Not necessarily an objection to this change, just explaining why I didn't do it prior.

Thanks for the explanation.  It makes sense, but I tried both ways and this way produces a more readable result.

internal/api/registry.go
deads2k: It doesn't work.  I think it's because the type is aliased to an internal type perhaps?

PR #3071, opened 2025-10-23
fix internal->external->internal conversions for nodepool and external auth

deads2k:
Added fuzz testing and found several bugs.  It is now ready to review/merge.


mbarnes (approved):
LGTM

Thought for the future: most if not all of `conversion_fuzz_test.go` looks ripe for factoring out to the `/internal/api` level so it doesn't have to be copied for each API version.  The version interface could be an input parameter where needed.  Maybe.

PR #3081, opened 2025-10-27
Service monitor creation check for KAS monitoring

weherdh:
[To enhance the RMO readiness work ](https://issues.redhat.com/browse/AROSLSRE-147)
### What
To add a new alert to monitor the `ServiceMonitor`, which is created for HCP 

### Why
RMO was recently updated from upstream, making it break in AKS

There is no kubeapiserver target for the newly created HC in the blackexporter page. We should be notified of this kind of failure

### DIFF
--- a/observability/alerts/HCPkasMonitor-prometheusRule.yaml
+++ b/observability/alerts/HCPkasMonitor-prometheusRule.yaml
@@ -44,3 +44,26 @@ spec:
         short_window: 6h
       annotations:
         description: "High error budget burn for {{ $labels.probe_url }} (current value: {{ $value }})"
+    # ServiceMonitor Creation Monitoring - Detects when HCP namespaces exist but probe_success metrics are missing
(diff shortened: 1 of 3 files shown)


observability/alerts/HCPkasMonitor-prometheusRule.yaml
```diff
@@ -44,3 +44,26 @@ spec:
...
+            count(kube_namespace_status_phase{phase="Active", namespace=~"ocm-aro.*-.*-.*"})
+            -
+            count(probe_success{namespace=~"ocm-aro.*"})
+          )
+          /
+          count(kube_namespace_status_phase{phase="Active", namespace=~"ocm-aro.*-.*-.*"})
+        ) > 0.144
+      for: 15m
+      labels:
+        severity: warning
```
stevekuznetsov: Up to you all - but perhaps consider making this informational for a while. You will learn if the alert is robust.

PR #3083, opened 2025-10-27
feat: remove cs rolling update override config in int

miguelsorianod:
It is already set as a global default with the same value so we can remove the override.


machi1990 (approved):
Thanks for the cleanup.

For reference, the global default is applied here https://github.com/Azure/ARO-HCP/blob/da005bba2910b6b225c678cf55b6d5b65a9de075/config/config.yaml#L651

PR #3086, opened 2025-10-27
prepare cs image from new quay repository app-sre/aro-hcp-clusters-service

miguelsorianod:
The image is 7966fb31c657d9ccd6333c40d30079f8fa9ced7cf1d89fcaf784c6f4ba569d30 and it is functionally equivalent to the image
6af9be4724b345e040f8f1718870fdcc43a8ece6dbf66d26abdefad243938e6e from app-sre/uhc-clusters-service

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -805,7 +805,8 @@ clouds:
       clustersService:
         environment: "arohcpdev"
         image:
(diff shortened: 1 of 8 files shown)


miguelsorianod:
what-if MR is failing due to unrelated changes. As mentioned by lifecycle team, we can ignore it as it is brittle and it is going to be replaced.

PR #3102, opened 2025-10-28
prepare update cs image in stg, prod using new quay repository app-sre/aro-hcp-clusters-service

miguelsorianod:
The change sets the CS registry path global default to app-sre/aro-hcp-clusters-service and removes the overwrites as it is now set globally.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -622,7 +622,7 @@ defaults:
   clustersService:
     image:
       registry: quay.io
-      repository: app-sre/uhc-clusters-service
+      repository: app-sre/aro-hcp-clusters-service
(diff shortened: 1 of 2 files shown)


config/config.yaml
```diff
@@ -622,7 +622,7 @@ defaults:
   clustersService:
     image:
       registry: quay.io
-      repository: app-sre/uhc-clusters-service
+      repository: app-sre/aro-hcp-clusters-service
```
machi1990: This makes https://github.com/Azure/ARO-HCP/blob/49782426c3abd3bd498e94ced2c605eaafed9cba/config/config.yaml#L815 no longer needed

ahitacat (approved):

PR #3113, opened 2025-10-28
*: migrate to new dev flow, use Helm built-in

stevekuznetsov:
Notably, I removed the CURRENT_VERSION from the ConfigMap as I found no users of it, and the image-environment file from the container image build, for the same reason.


stevekuznetsov:
Dry-run is broken on https://github.com/Azure/ARO-Tools/pull/150 but this looks good in local dev.

mbarnes (dismissed):
I can only really speak to the first two commits, but the changes seem ok to me.  Was trying to test it out but got thwarted by Azure outages.

geoberle (dismissed):

PR #3116, opened 2025-10-29
Add testing for helm charts

janboll:
### What

Add tooling to test helm files

### Why

I.e. the arobit helm chart is quite complex with conditionals and environment specific configuration. Using golden test files should make it easier to maintain these files.

### DIFF
--- a/tooling/helmtest/main_test.go
+++ b/tooling/helmtest/main_test.go
@@ -25,6 +25,7 @@ import (
 	"helm.sh/helm/v4/pkg/action"
 	"helm.sh/helm/v4/pkg/chart/loader"
 	"helm.sh/helm/v4/pkg/cli"
+	"helm.sh/helm/v4/pkg/release"
 
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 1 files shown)


tooling/helmtest/main_test.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal .helmtests.yaml: %w", err)
+	}
+	return testConfig, nil
+}
+
+func runHelmTemplate(helmDeploy, helmChartDir, namespace string, setValues []string) (string, error) {
+	args := []string{
+		"helm",
+		"template",
```
stevekuznetsov: Import & use their library :)

janboll: I'd rather not. The template command is not exported, as i.e. the upgrade. I'd have to re-implement the templating here.

janboll: I will revisit this, since using the SDK will also make it easier to create tests and reuse the config/values files!

raelga (dismissed):

PR #3122, opened 2025-10-29
Refactor Kusto Log setup

janboll:
https://issues.redhat.com/browse/AROSLSRE-200

## What

Refactoring Kusto Log Setup:

- Create dedicated instances via dedicated pipelines file.
- Add mapping from region to Geo (from ev2config), to reduce the number of Kusto instances total.
- Arobit, add geo/region mapping and generate ingest url from naming convention.
- Naming convention for instances: hcp-$geoShortId, RG:  hcp-kusto-$geoShortId
- Add toggles to ensure Kusto access/deployment is only attempted where possible/required.

### Deployment to int/prod

- In theory deployment should happen by calling dedicated entrypoint `Microsoft.Azure.ARO.HCP.Log.Infra`. See generated changes https://dev.azure.com/msazure/AzureRedHatOpenShift/_git/sdp-pipelines/pullrequest/14015010.
- But first step after promoting this PR is to ensure nothing in the existing setup breaks. 
- If things continue to work fine, create follow-ups in sdp-pipeline, that configure kusto instances in all regions required (one per geo). 
- After instances have been created, toggle `arobit.kusto.enabled`. This will grant the log mi access to the Kusto databases and toggle the outputs in the config to point to Kusto. Up until we disable mdsd, this will send logs into the old and the new Kusto instances.
- After ensuring logs look fine, disable mdsd for service (keep it for customer logs, cause shoebox...)

### DIFF
--- a/config/dev.digests.yaml
+++ b/config/dev.digests.yaml
@@ -3,22 +3,22 @@ clouds:
     environments:
       cspr:
         regions:
-          westus3: 9d1e465b9fb49af27aa27fca617c8919e72b4f528b0e49efd7fb7464420a75ce
+          westus3: fb0796526d98ea03fd65cc04e6aa036d6f94937e83a73c78c79b6cdaf624de96
       dev:
         regions:
-          westus3: f35680e71c63e14dbda469a20f618dba7ddded820bd0be13e2dc6a4768b55bcf
+          westus3: ca51f1e42c3aa4a4954d785628078187fa70265178a3d3d63278c932e35b9f96
       ntly:
         regions:
-          uksouth: 4c4936aba69db4bfd13c1e3ff93562b986cfe6025ec587a2b63d770bf43e8903
+          uksouth: 44c19f525f0e318b5dd15fc0809383c50d54b0f15eb197ed0d8eff3b5fde8466
       perf:
         regions:
-          westus3: 1dadf86f9d0da12a836688c0c0b468d7dd05c154b96f1b203a67bbc9dd52e834
+          westus3: ad60709430ca3857131410ab9648fb1da9365f7629e1955efe57de661068e911
       pers:
         regions:
-          westus3: 8d3f8fa781762e565de41b2d9e1f43aea63d02305d773bf30888b77197f901a0
+          westus3: 6c9f7566fea55d4358a42c8210736c23e40e412d6c0b3202dc55f5b5c912bf86
       prow:
         regions:
-          westus3: fcedab6f74c62c0d3acf2ddf48390373cf71fe2ca77711dd1260eb563edb5560
+          westus3: 4125e133c2fac20cb66a903f803a12d9a3f35ea9d39168fa32955d8af9a238d9
       swft:
         regions:
-          uksouth: ace05b5022478ea9d3a81098b7ba6aec029c5b95c5beb78b9e01e368e2d849d9
+          uksouth: c2aff65601fbc02da7b24e852feb7cec12946bcd2c9399e76fe0ce2f1f77bb49


dev-infrastructure/modules/logs/kusto/tables/backendContainerLogs.kql
```diff
@@ -0,0 +1,56 @@
...
+  resourceId: string,
+  internalId: string,
+  clusterId: string,
+  clientRequestId: string,
+  correlationRequestId: string,
+  cloudErrorCode: string,
+  cloudErrorMessage: string,
+  subscriptionId: string,
+  tag: string,
+  unmatched: dynamic
```
stevekuznetsov: Do you have any idea how we might catch new log statements being added to the frontend that either typo a field, use some other field (clusterID vs clusterId), etc?

dev-infrastructure/modules/logs/kusto/tables/backendContainerLogs.kql
```diff
@@ -0,0 +1,56 @@
...
+  {"column":"podName","Properties":{"path":"$.log.kubernetes.pod_name"}},
+  {"column":"environment","Properties":{"path":"$.log.environment"}},
+  {"column":"level","Properties":{"path":"$.level"}},
+  {"column":"operation","Properties":{"path":"$.log.operation"}},
+  {"column":"operationId","Properties":{"path":"$.log.operation_id"}},
+  {"column":"resourceId","Properties":{"path":"$.log.resource_id"}},
+  {"column":"internalId","Properties":{"path":"$.log.internal_id"}},
+  {"column":"clusterId","Properties":{"path":"$.log.cluster_id"}},
+  {"column":"clientRequestId","Properties":{"path":"$.log.client_request_id"}},
+  {"column":"correlationRequestId","Properties":{"path":"$.log.correlation_request_id"}},
```
stevekuznetsov: I don't know that 
a) we want to remap the structured fields that users have opted to use in their code (they will expect the logs to look the way they wrote them)
b) we want to allow our components to name things "wrong" - e.g. this is just the `correlationId`, I have never ever heard this called `correlation_request_id`. I imagine we want to allow JOIN across tables given this key, would be nice to get some consistency on it

dev-infrastructure/modules/common.bicep
```diff
@@ -408,6 +408,333 @@ var _locationAvailabilityZones = {
   }
 }
 
+var _regionToGeography = {
```
stevekuznetsov: Why do you need this as opposed to looking up the geo from the config?

dev-infrastructure/kusto-pipeline.yaml
```diff
@@ -0,0 +1,33 @@
...
+  resourceGroup: '{{ .kusto.rg }}'
+  subscription: '{{ .global.subscription.key }}'
+  executionConstraints:
+  - clouds:
+    - public
+    environments:
+    - int
+    regions:
+    - eastus
+  - clouds:
```
stevekuznetsov: stg?

dev-infrastructure/configurations/mgmt-cluster.tmpl.bicepparam
```diff
@@ -87,3 +87,7 @@ param genevaManageCertificates = {{ .geneva.logs.manageCertificates }}
 
 // Log Analytics Workspace ID will be passed from region pipeline if enabled in config
 param logAnalyticsWorkspaceId = '__logAnalyticsWorkspaceId__'
+
+param arobitKustoEnabled = {{ .arobit.kusto.enabled }}
+param serviceLogsDatabase = '{{ .kusto.serviceLogsDatabase }}'
+param customerLogsDatabase = '{{ .kusto.customerLogsDatabase }}'
```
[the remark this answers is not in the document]
janboll: Forgot to rename it to hostedControlPlaneLogsDatabase, done now

dev-infrastructure/modules/logs/kusto/tables/backendLogs.kql
```diff
@@ -0,0 +1,56 @@
...
+  {"column":"podName","Properties":{"path":"$.log.kubernetes.pod_name"}},
+  {"column":"environment","Properties":{"path":"$.log.environment"}},
+  {"column":"level","Properties":{"path":"$.level"}},
+  {"column":"operation","Properties":{"path":"$.log.operation"}},
+  {"column":"operationId","Properties":{"path":"$.log.operation_id"}},
+  {"column":"resourceId","Properties":{"path":"$.log.resource_id"}},
+  {"column":"internalId","Properties":{"path":"$.log.internal_id"}},
+  {"column":"clusterId","Properties":{"path":"$.log.cluster_id"}},
+  {"column":"clientRequestId","Properties":{"path":"$.log.client_request_id"}},
+  {"column":"correlationRequestId","Properties":{"path":"$.log.correlation_request_id"}},
```
janboll: I think this is out of scope for this PR?

dev-infrastructure/modules/logs/kusto/tables/backendLogs.kql
```diff
@@ -0,0 +1,56 @@
...
+  resourceId: string,
+  internalId: string,
+  clusterId: string,
+  clientRequestId: string,
+  correlationRequestId: string,
+  cloudErrorCode: string,
+  cloudErrorMessage: string,
+  subscriptionId: string,
+  tag: string,
+  unmatched: dynamic
```
janboll: No, I think this requires manual rigor

observability/arobit/deploy/templates/forwarder-configmap.yaml
```diff
@@ -1,3 +1,5 @@
+{{- $kustoRegion := get ( .Files.Get "kusto-regions.yaml" | fromYaml ) .Values.forwarder.geoShortId -}}
```
stevekuznetsov: Why do we need to do the reverse lookup? Can't we use the geoShortID as the DNS segment in the Kusto link? I guess it's fine to have this way, but remembering to keep the `kusto-regions.yaml` always up-to-date with every possible cloud & region sounds like a bad idea. If we can remove the need for the reverse lookup, that would be better

stevekuznetsov (approved):

PR #3127, opened 2025-10-30
Quality gate and Filter to support Local test suite

venkateshsredhat:
Run make record-nonlocal-e2e - Generate the list of non-Local test specs
Check for changes - Use git status --short to see if the generated file changed
 Fail if changes found - If the file changed, it means:
    - New tests were added without proper labeling
    - Existing tests had their labels changed
    - The baseline needs to be updated

This is to ensure future tests will support local runs . followup  PR : https://github.com/Azure/ARO-HCP/pull/3089 .

### DIFF
--- a/.github/workflows/ci-go.yml
+++ b/.github/workflows/ci-go.yml
@@ -81,7 +81,7 @@ jobs:
         make record-nonlocal-e2e
         if [[ ! -z "$(git status --short)" ]]
         then
-          echo "E2E test specs have changed. All new tests must have the 'Local' label and support local env run or be included in the non-Local baseline. Run 'make record-nonlocal-e2e' to update the baseline if needed."
+          echo "E2E test specs have changed. All new tests must have the 'ARO-HCP-RP-API-Compatible' label and support local env run or be included in the non-Local baseline. Run 'make record-nonlocal-e2e' to update the baseline if needed."
           git status
           git diff
           exit 1


nonlocal-e2e-specs.txt
```diff
@@ -0,0 +1,113 @@
...
+List HCPOpenShiftCluster Positive Successfully lists clusters filtered by subscription ID [PreLaunchSetup:HappyPathInfra, Importance:Medium, Positivity:Positive]
+/home/ves/Desktop/dev/ARO-HCP/test/e2e/cluster_list_test.go:44
+• [0.000 seconds]
+------------------------------
+List HCPOpenShiftCluster Positive Successfully lists clusters filtered by resource group name [PreLaunchSetup:HappyPathInfra, Importance:Medium, Positivity:Positive]
+/home/ves/Desktop/dev/ARO-HCP/test/e2e/cluster_list_test.go:69
+• [0.000 seconds]
+------------------------------
+Customer should create an HCP cluster and validate TLS certificates [PreLaunchSetup:None, Importance:Critical, Positivity:Positive]
+/home/ves/Desktop/dev/ARO-HCP/test/e2e/cluster_tls_endpoints.go:39
```
venkateshsredhat: we cant have these paths and timing .

stevekuznetsov (dismissed):

mbukatov (changes requested):
Based on yesterday's discussion, I'm rejecting this PR. Please close it for now.

Let's focus on https://github.com/Azure/ARO-HCP/pull/3089 and then we can have a discussion how to enforce gating based on not yet defined testing guidelines.

mbukatov:
Let me explain why I think it's not right time to merge this PR right now.

While we agree that we want to get to the point where we are able to run E2E tests against RP running in dev env directly without ARM integration, there is still few details we need to figure about the details how to get there.

For this reason we don't want to basically stop all E2E test development until the effort is figured out. Merging this PR would prevent adding more tests which are in progress, which is very disruptive.

Moreover this PR contains few details which depends on us to figure the details out, as well as points I raised already. So I didn't want us to waste time arguing here.

As I explained yesterday, I don't want to have test suites named according to an environment, but rather a purpose (what it's testing and why). Since this is "local" approach driven by an effort to run E2E test cases as a final part of infra deployment CI check in dev, it makes sense to establish such test suite instead of just calling it "local". So eg. "dev-infra-check/parallel".

In the same way, the labels should not indicate an environments, but some aspects of the tests. For the purpose of distinguishing test cases which we can run against RP directly without ARM integration, we could adopt a label indicating either that the test case is using RP api only "rp-api-only" instead of "local".

The reason is behind this is that it will allow us to control what tests to run for which purpose. That was not a problem initially where there was only few tests, but eventually, we won't run all the tests in all cases because this won't be good use of our time and resources.

PR #3140, opened 2025-10-31
ARO-21891: Add support for non-stable OpenShift channel groups

ahitacat:
https://issues.redhat.com/browse/ARO-21891

## What

  Add support for creating ARO-HCP clusters using non-stable OpenShift
  channel groups (candidate, nightly) when the
  AllowDevNonStableChannels Azure Feature Exposure Control (AFEC) flag is
  registered on the subscription.

## Why
ARO-HCP developers currently can only provision clusters using stable OpenShift release channels. This limitation prevents early testing of upcoming OpenShift features and fixes. By enabling non-stable channel access for authorized developer subscriptions, the team can develop with early access OpenShift images.

### Key changes
  Key Changes Made:

  1. Validation Layer (internal/validation/validate_cluster.go) static validation
  - Make channelGroup an immutable field (it will not allow to update clusters with a different channel group)
  2.  Admission Layer (internal/admission/admit_cluster.go)
  For cluster creation.
  - Added subscription parameter to admission functions to check for AllowDevNonStableChannels AFEC flag
  - When feature flag is enabled:
    - Allow ANY channel group (not restricted to stable/candidate/nightly/fast)
    - Allow full semver format (X.Y.Z-prerelease) for version.id
  - Without feature flag: Only "stable" channel group allowed with X.Y format
  3. Update frontend to also call admission validation during cluster creation
  4. OCM Conversion Layer (internal/ocm/client.go and convert.go)
  - Updated NewOpenShiftVersionXYZ() to format versions as:
    - For stable: openshift-v4.18.7 (no channel group suffix)
    - For non-stable: openshift-v4.15.1-candidate (with channel group suffix)
  - Applied to both cluster and node pool conversions
  - Fixed edge case where empty inputs would produce "-" instead of ""
  - Only appends default patch ".7" for X.Y format inputs
  5. Test Updates
  - Updated comprehensive validation tests (validate_cluster_comprehensive_test.go)
  - Updated frontend node pool test to use the new version formatting
  - Updated convert_test to add channel-group cases different than stable
  - Created admit_cluster tests to check admission validation.
  
  Behavior:

  With AllowDevNonStableChannels AFEC flag enabled:
  {
    "version": {
      "id": "4.15.1",           // Full semver allowed
      "channelGroup": "candidate"  // Any channel group allowed
    }
  }
  → Sends to OCM as: openshift-v4.15.7-candidate

  Without feature flag (default):
  {
    "version": {
      "id": "4.15",            // X.Y format only
      "channelGroup": "stable"     // Only "stable" allowed
    }
  }
  → Sends to OCM as: openshift-v4.15.7

## Future work:
- fuzz tests (?)
- e2e jobs (?)

(Updated with recent changes https://github.com/Azure/ARO-HCP/pull/3140#issuecomment-3528368060)

### DIFF
--- a/internal/admission/admit_cluster_test.go
+++ b/internal/admission/admit_cluster_test.go
@@ -1,0 +1,336 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package admission
+
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/ptr"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+)
+
+type expectedError struct {
+	message   string // Expected error message (partial match)
+	fieldPath string // Expected field path for the error
+}
+
+func verifyErrorsMatch(t *testing.T, expectedErrors []expectedError, errs field.ErrorList) {
+	if len(expectedErrors) != len(errs) {
+		t.Errorf("expected %d errors, got %d: %v", len(expectedErrors), len(errs), errs)
+		return
+	}
+
+	// Check that each expected error message and field path is found
+	for _, expectedErr := range expectedErrors {
+		if len(strings.TrimSpace(expectedErr.fieldPath)) == 0 {
+			t.Errorf("expected error with path %s to be non-empty", expectedErr.fieldPath)
+		}
+		if len(strings.TrimSpace(expectedErr.message)) == 0 {
+			t.Errorf("expected error with msg %s to be non-empty", expectedErr.message)
+		}
+
+		found := false
+		for _, err := range errs {
+			messageMatch := strings.Contains(err.Detail, expectedErr.message) || strings.Contains(err.Error(), expectedErr.message)
+			fieldMatch := strings.Contains(err.Field, expectedErr.fieldPath)
+			if messageMatch && fieldMatch {
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Errorf("expected error containing message '%s' at field '%s' but not found in: %v", expectedErr.message, expectedErr.fieldPath, errs)
+		}
+	}
+
+	for _, err := range errs {
+		found := false
+		for _, expectedErr := range expectedErrors {
+			messageMatch := strings.Contains(err.Detail, expectedErr.message) || strings.Contains(err.Error(), expectedErr.message)
+			fieldMatch := strings.Contains(err.Field, expectedErr.fieldPath)
+			if messageMatch && fieldMatch {
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Errorf("actual error '%v' but not found in expected", err)
+		}
+	}
+}
+
+// Helper function to create a valid cluster for testing
+func createValidCluster() *api.HCPOpenShiftCluster {
+	cluster := api.NewDefaultHCPOpenShiftCluster(api.Must(azcorearm.ParseResourceID("/subscriptions/0465bc32-c654-41b8-8d87-9815d7abe8f6/resourceGroups/some-resource-group/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/test-cluster")))
+
+	// Set required fields
+	cluster.Location = "eastus"
+	cluster.CustomerProperties.Version.ID = "4.15"
+	cluster.CustomerProperties.Version.ChannelGroup = "stable"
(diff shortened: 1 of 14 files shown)


internal/ocm/client.go
```diff
@@ -545,18 +545,36 @@ func NewOpenShiftVersionXY(v string) string {
...
+		if len(parts) == 2 {
+			parts = append(parts, OpenShift419Patch)
+		}
+
+		csVersion = api.OpenShiftVersionPrefix + strings.Join(parts, ".") + prereleasePart
+
+		// Only append channel group if it's not empty and not "stable"
+		if len(cg) > 0 && cg != "stable" {
+			csVersion = csVersion + "-" + cg
+		}
```
mbarnes: Could you add a comment somewhere in here showing the expected version formats for the various channels we're now supporting?  I gather from the code it's going to be something like:
```
   stable: openshift-vX.Y.Z
candidate: openshift-vX.Y.Z[-prereleasePart]-candidate
  nightly: openshift-vX.Y.Z[-prereleasePart]-nightly
```

But I'd like more detail about how `prereleasePart` is formatted.

In the example you gave in a comment above it looks like "nightly" could appear twice?
e.g. `openshift-v4.19.0-0.nightly-2025-01-01-nightly`?

mbarnes (changes requested):
This is just a first-pass review.  Looks good overall.  Requested a couple small changes.

internal/validation/validate_cluster.go
```diff
@@ -32,14 +32,14 @@ import (
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 )
 
-func ValidateClusterCreate(ctx context.Context, newCluster *api.HCPOpenShiftCluster, validationPathMapper api.ValidationPathMapperFunc) field.ErrorList {
+func ValidateClusterCreate(ctx context.Context, subscription *arm.Subscription, newCluster *api.HCPOpenShiftCluster, validationPathMapper api.ValidationPathMapperFunc) field.ErrorList {
```
deads2k: this change isn't allowed.  Validation methods are static (only the new and old obj).  If non-static validation is needed, add it to the the admission package.  This separation makes it clear which pieces of validation are guaranteed (static) and which are suggestions likely to slip over time and become invalid (admission).

If we expect this option to go away in the future and be permanently on, it can be added to the Operation.Options.

ahitacat:
Key Changes:

- Structural change: major part of the version validation is done in admission as it requires AFEC flag check - as suggested here https://github.com/Azure/ARO-HCP/pull/3140#discussion_r2486542341
- Static validation simplified: now channelGroup is also immutable and it is required - https://github.com/Azure/ARO-HCP/pull/3140#discussion_r2481197293 (It will always be provided as when creating the cluster with the NewDefaultHCPOpenShiftCluster always sets it)
- New tests to reflect these changes

TODO
I need to do manual validation of this.

internal/admission/admit_cluster_test.go
```diff
@@ -0,0 +1,336 @@
...
+	cluster.Location = "eastus"
+	cluster.CustomerProperties.Version.ID = "4.15"
+	cluster.CustomerProperties.Version.ChannelGroup = "stable"
+	cluster.CustomerProperties.DNS.BaseDomainPrefix = "test-cluster"
+	cluster.CustomerProperties.Platform.SubnetID = "/subscriptions/0465bc32-c654-41b8-8d87-9815d7abe8f6/resourceGroups/some-resource-group/providers/Microsoft.Network/virtualNetworks/test-vnet/subnets/test-subnet"
+	cluster.CustomerProperties.Platform.NetworkSecurityGroupID = "/subscriptions/0465bc32-c654-41b8-8d87-9815d7abe8f6/resourceGroups/some-resource-group/providers/Microsoft.Network/networkSecurityGroups/test-nsg"
+	cluster.CustomerProperties.Platform.ManagedResourceGroup = "managed-rg"
+
+	return cluster
+}
```
mbarnes: Not a blocker, but just FYI in `/internal/api/testhelpers.go` there's a function named `MinimumValidClusterTestCase` that does the same as this, except for the explicit version ID.

Could simplify if you want but I'll leave it up to you.

mbarnes (changes requested):
Nice job.  This is basically good to go in my view, I just had a couple final trivial nits.

internal/api/featureflags.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+const (
+	// FeatureFlag AllowDevNonStableChannels is the feature in the subscription that
+	// allows the usage of non stable channels (i.e. candidate, nightly) for creation
+	// of new OpenShift clusters.
+	AllowDevNonStableChannels = "Microsoft.RedHatOpenShift/AllowDevNonStableChannels"
```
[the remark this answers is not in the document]
ahitacat: @mbarnes @deads2k do we want to add the Feature prefix or is it better to move this into a feature package?

internal/validation/validate_cluster.go
```diff
@@ -334,22 +334,38 @@ var (
...
+	if allowNonStableChannels && newObj.ChannelGroup != "stable" {
+		// For non-stable channels with feature flag: allow full semver format (X.Y.Z-prerelease)
+		errs = append(errs, OpenshiftVersionWithOptionalMicro(ctx, op, fldPath.Child("id"), &newObj.ID, nil)...)
+	} else {
+		// For stable or without feature flag: only MAJOR.MINOR format
+		errs = append(errs, OpenshiftVersionWithoutMicro(ctx, op, fldPath.Child("id"), &newObj.ID, nil)...)
+	}
 
 	// ChannelGroup string `json:"channelGroup,omitempty"      visibility:"read create update"`
 	errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("channelGroup"), &newObj.ChannelGroup, nil)...)
```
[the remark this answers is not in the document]
ahitacat: I added the channelGroup as an immutable field, and changed the mutation tests to address this change.

PR #3147, opened 2025-10-31
Add pod antiAffinity rules to deployments to prefer scheduling highly available or on different nodes

bennerv:
### What
Add pod anti affinity rules (preferred, not required) to deployments which run more than one replica.  

### Why

If pods end up scheduling on the same node and that node falls over, the service will go down temporarily while pods reschedule.  

🚢

### DIFF
--- a/observability/prometheus/deploy/templates/prometheus.yaml
+++ b/observability/prometheus/deploy/templates/prometheus.yaml
@@ -15,6 +15,20 @@ spec:
             operator: In
             values:
             - infra
+    podAntiAffinity:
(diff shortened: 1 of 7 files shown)


observability/prometheus/deploy/templates/prometheus.yaml
```diff
@@ -15,6 +15,20 @@ spec:
             operator: In
             values:
             - infra
+    podAntiAffinity:
```
bennerv: Need to check this works on the CR.

PR #3151, opened 2025-10-31
Migrate to topologyConstraints instead of anti affinity

bennerv:
### What

Use topology constraints across our pods instead of pod anti-affinity.  

### Why

As we scale replicas, this will result in more balanced deployments.   It achieves a different goal than anti-affinity.  Rather than focusing on not scheduling, we focus on balancing our workloads to result in an HA deployment.


bennerv:
In draft to test to make sure everything is behaving as expected.

SudoBrendan (approved):

PR #3161, opened 2025-11-03
Update acm chart, add mce testfile

janboll:

### DIFF
--- a/acm/pipeline.yaml
+++ b/acm/pipeline.yaml
@@ -16,6 +16,22 @@ resourceGroups:
   resourceGroup: '{{ .mgmt.rg }}'
   subscription: '{{ .mgmt.subscription.key }}'
   steps:
+  - name: deploy
+    action: Shell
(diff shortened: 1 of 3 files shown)


acm/pipeline.yaml
```diff
@@ -16,6 +16,22 @@ resourceGroups:
   resourceGroup: '{{ .mgmt.rg }}'
   subscription: '{{ .mgmt.subscription.key }}'
   steps:
+  - name: deploy
+    action: Shell
+    command: echo "1"
```
geoberle: you can make such env vars available to the Makefile without a noop deploy step.
switch out `-include ../setup-env.mk` for `-include ../setup-templatize-env.mk` in the Makefile and provide an `/acm/Env.mk` file similar to https://github.com/Azure/ARO-HCP/blob/71252a09ba05025e203ac9a7351318331282c6c2/admin/Env.mk

PR #3181, opened 2025-11-04
New E2E test case: deployment of a node pool is not possible for hosted cluster in failed provisioning state

mvacula02:
[ARO-22116](https://issues.redhat.com/browse/ARO-22116)

### What

New negative E2E test case to ensure customer can't deploy a node pool into a hosted cluster in failed provisioning state

### How

1. Create a resource group
2. Create a cluster using ~~bicep template~~ API* with invalid ~~vnetName~~ subnet ID to force failure
3. Verify that an error does not occur before ARM resource creation
4. Verify the cluster resource exists
5. Verify the cluster provisioning state is failed (polls and logs current provisioning state)
6. Attempt node pool deployment into the cluster via API*
7. Verify the node pool deployment fails with appropriate error code

*using framework as per https://github.com/Azure/ARO-HCP/pull/3273

### DIFF
--- a/test/e2e/nodepool_in_failed_cluster.go
+++ b/test/e2e/nodepool_in_failed_cluster.go
@@ -25,8 +25,6 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/rand"
 
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
-
 	hcpsdk20240610preview "github.com/Azure/ARO-HCP/test/sdk/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
 	"github.com/Azure/ARO-HCP/test/util/framework"
 	"github.com/Azure/ARO-HCP/test/util/labels"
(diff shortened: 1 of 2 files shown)


mgahagan73 (dismissed):
Nice test!

test/e2e/nodepool_in_failed_cluster.go
```diff
@@ -0,0 +1,103 @@
...
+		labels.Negative,
+		labels.Critical,
+		func(ctx context.Context) {
+			tc := framework.NewTestContext()
+
+			By("creating resource group ")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "rg-nodepool-in-failed-cluster", tc.Location())
+			Expect(err).NotTo(HaveOccurred())
+
+			clusterName := "nodepool-in-failed-cluster" + rand.String(6)
```
mbukatov: Could we name this cluster just "failed-cluster"? It's little confusing that something with nodepool prefix is actually a cluster, and the information about the use case "nodepool-in-failed-cluster" is captured in the resource group already. It will make the errors (if any) easier to read in CI.

test/e2e/nodepool_in_failed_cluster.go
```diff
@@ -0,0 +1,103 @@
...
+			nodePool := hcpsdk20240610preview.NodePool{
+				Location: to.Ptr(tc.Location()),
+				Properties: &hcpsdk20240610preview.NodePoolProperties{
+					Platform: &hcpsdk20240610preview.NodePoolPlatformProfile{
+						VMSize: to.Ptr("Standard_D2s_v3"),
+					},
+					Replicas: to.Ptr(int32(1)),
+				},
+			}
+			_, err = nodePoolClient.BeginCreateOrUpdate(ctx, *resourceGroup.Name, clusterName, nodePoolName, nodePool, nil)
```
mbukatov: Would it make sense to enforce some timeout here?

mbukatov (commented):
Adding few minor comments.

patriksuba (approved):

PR #3185, opened 2025-11-04
test: Bump openshift-tests-extension to latest commit

patriksuba:
### What

Update module openshift-tests-extension to latest commit

### Why

Include changes with Deserialization error fix when node timeout occurs


mgahagan73:
PR looks fine, I want to wait on the tests finishing before approving just in case.

PR #3226, opened 2025-11-07
admin api ingress and mise integration

geoberle:
### What

* virtual service and extauth authorization policy for admin api
* mise setup for geneva action app bearer token authentication
* frontend host matcher in virtual service
* testing tool for authenticated admin api http requests
* use ev2 central config values to configure mise top level
* helmtests for admin api and istio (with and without mise)

https://issues.redhat.com/browse/ARO-22400

### DIFF
--- a/config/dev.digests.yaml
+++ b/config/dev.digests.yaml
@@ -3,22 +3,22 @@ clouds:
     environments:
       cspr:
         regions:
-          westus3: 66c4d6df54a2367fb0bc41f4a047c270ff48e8a7a717013aeff251d2b9ded25d
+          westus3: 386213dc4200ce465067b79bc2180e2c4ec6c61f1d98150e18ee4d9c0642c497
       dev:
         regions:
-          westus3: bce4c8ba26364e05cf5aaed888a2b92122b3258235155c7d099ad3256663cfd5
+          westus3: ca79aac0b15b8058fe5baf5dd11cefbf741475e2cc430a46db07feed1b1c0fd2
       ntly:
         regions:
-          uksouth: 08745287f86cd34cb4557ea39cc12f25ddae4616ba73ced5956130ee5c55b85c
+          uksouth: c2d2ceac13deb40a9f6124d97a25d483947f1febbd618c4c4e5852fade130729
       perf:
         regions:
-          westus3: cf75e3dd8d76983d02ccaf560e338720143023e478f7e480fd9d14e6fc8f89af
+          westus3: c92158b134c4815510b9716fd6a29e16553d10288ea02d7540b631a1e5696073
       pers:
         regions:
-          westus3: c20622017cfb13581fc1d4b13fef99b9cdfc426d9f4702533f5c6cf9abf21269
+          westus3: bc69b5a06288316a296b7fdd57d2a2c04af7d221997aa71567a29577af002203
       prow:
         regions:
-          westus3: 2b2ef5178ad3063c6b4447048f5811a64432294a4db0ef3873047ea05a12d3ec
+          westus3: 10f092661182d2ccd95f7ddccd414e69bfad30fe161fff878fd3c4250c82d926
       swft:
         regions:
-          uksouth: 97f4fa7db135902cab22dcdebe9c533017cda753fd0a0c732f6916f3c1857209
+          uksouth: 2a8f6a48854e4380723206d31897a26d9b92527f6cae22f397d34f2b3bb7e7ef
--- a/config/rendered/dev/prow/westus3.yaml
+++ b/config/rendered/dev/prow/westus3.yaml
@@ -515,15 +515,20 @@ miMockCertName: msiMockCert2
 miMockClientId: e8723db7-9b9e-46a4-9f7d-64d75c3534f0
 miMockPrincipalId: d6b62dfa-87f5-49b3-bbcb-4a687c4faa96
 mise:
-  armAppId: ""
-  armInstance: ""
-  azureAdInstance: ""
+  arm:
+    applicationId: e2c2ff5c-e5b4-4e79-8c3e-1da8c48461e7
+    audienceFQDN: management.azure.com
+    authorityFQDN: login.microsoftonline.com
+    policyLabel: ARM Policy
+    tenantId: 33e01921-4d64-4f8c-a055-5bdaffd5e33d
   deploy: false
-  firstPartyAppId: ""
+  genevaActions:
+    audienceFQDN: management.azure.com
+    authorityFQDN: sts.windows.net
+    policyLabel: Geneva Actions
   image:
-    digest: ""
-    repository: ""
-  tenantId: ""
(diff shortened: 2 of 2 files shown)


tony-schndr (commented):
overall lgtm, did you mean to include `Revert "Add mgmt cluster capacity alerts with SL team routing"`

admin/cli/pkg/cli/auth/auth.go
```diff
@@ -0,0 +1,217 @@
...
+				fmt.Printf("    Certificate %d: %s...\n", i+1, certStr[:50])
+			} else {
+				fmt.Printf("    Certificate %d: %s\n", i+1, certStr)
+			}
+		}
+	} else {
+		fmt.Println("  X5C Certificate Chain: Not present (standard auth)")
+	}
+
+	// Print claims information
```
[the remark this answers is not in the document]
geoberle: done

mbukatov (dismissed):
Acking update of ARO-Tools in the dependencies of in `/test` (e2e test runs not affected).

config/config.schema.json
```diff
@@ -1568,23 +1574,30 @@
...
           "description": "The valid app ID for mise, which is the ARM Application Id, this is common across all clouds (MSIT/AME/FF)"
         },
+        "tenantId": {
+          "type": "string",
+          "description": "The tenant ID where the ARM Application (see armAppId property) lives. Usually AME"
+        },
+        "bearerTokenAuthorityFQDN": {
+          "type": "string",
+          "description": "The FQDN of the bearer token authority"
+        },
```
SudoBrendan: nit: Now that we're supporting multiple configurations of MISE and it's established kinds of trusts - is it worth putting them into buckets? e.g. a bucket for the ARM PoP tokens, and another for Geneva Actions? The documentation goes a long way here to be sure - but I wonder if we could make this more self-documenting that we've got exactly two kinds of config we care about?

admin/cli/cmd/base/options.go
```diff
@@ -0,0 +1,161 @@
...
+	Scopes            []string
+	KeyVaultURI       string
+	CertificateSecret string
+	Endpoint          string
+	Host              string
+	Insecure          bool
+}
+
+func (o *RawAuthOptions) BindFlags(cmd *cobra.Command) error {
+	cmd.Flags().StringVar(&o.TenantID, "ga-auth-tenant-id", o.TenantID, "Tenant ID to use for the Geneva Action auth certificate")
```
stevekuznetsov: Perhaps my misunderstanding, but what does it mean to use a tenant ID for a certificate? Or scopes? Is this the tenant in which we use the cert to auth, and are those the OAuth scopes we ask for?

SudoBrendan (dismissed):
Ultimately things look good here :heavy_check_mark: 

A few comments / food for thought you can address or ignore :)

Thanks Gerd!

admin/cli/cmd/base/options.go
```diff
@@ -0,0 +1,161 @@
...
+			Token:    token,
+			Endpoint: o.Endpoint,
+			Host:     o.Host,
+			Insecure: o.Insecure,
+		},
+	}, nil
+}
+
+func (o *AuthOptions) Execute(ctx context.Context) error {
+	logger := logr.FromContextOrDiscard(ctx)
```
stevekuznetsov: prefer to pass logger explicitly, if you must not, at least handle the error. implicit parameters in context + panic is a good way to end up in a bad place

admin/cli/pkg/cli/request/request.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+type Client struct {
+	httpClient  *http.Client
+	hostHeader  string
+	bearerToken string
+}
+
+func NewClient(bearerToken string, hostHeader string, insecureSkipVerify bool) *Client {
+	httpClient := &http.Client{}
```
stevekuznetsov: check out the debugging round tripper https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/sdp-pipelines?path=/tooling/pkg/clients/ev2/client.go&version=GBmain&line=233&lineEnd=234&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

you likely want something like that for ease of testing

stevekuznetsov (dismissed):

admin/api/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/Azure/ARO-HCP/admin/api
+
+go 1.24.3
+
+require github.com/spf13/cobra v1.10.1
+
+require (
+	github.com/inconshreveable/mousetrap v1.1.0 // indirect
```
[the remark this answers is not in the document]
geoberle: yeah it is cobra dep
https://github.com/spf13/cobra/blob/7da941c3547e93b8c9f70bbd3befca79c6335388/go.mod#L7

admin/deploy/testdata/zz_fixture_TestHelmTemplate_admin_api_mise.yaml
```diff
@@ -0,0 +1,223 @@
...
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: admin-api
+        azure.workload.identity/use: "true"
+    spec:
+      serviceAccountName: admin-api
+      affinity:
+        podAntiAffinity:
```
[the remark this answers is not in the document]
geoberle: ben did that for admin api, so i just rebased to fix this

admin/cli/cmd/base/options.go
geoberle: i handle the error. it is for sure better to fail than to silently drop logging and become blind when no logger was passed via the context

admin/cli/pkg/cli/request/request.go
geoberle: i've adopted that pattern

admin/cli/cmd/base/cmd.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	if err := opts.BindFlags(cmd); err != nil {
+		return nil, err
+	}
+
+	return cmd, nil
+}
+
+func runAuthTest(ctx context.Context, opts *RawAuthOptions) error {
```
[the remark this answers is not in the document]
geoberle: there are two things to it.
* the `client/pkg` code is a general purpose client to the admin api that can be used in e2e tests to verify the admin api
* the `client/cmd` code makes the client library available as a cli for SREs to run in lower environments like INT.

the e2e test is not yet part of this PR but will be handled in a followup one and integrate into the Prow based PR checks once they are available.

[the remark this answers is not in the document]
geoberle: i had a talk with @venkateshsredhat and we believe we can add this kind of testing to the local e2e test suit that will be available in the next two weeks.

stevekuznetsov (dismissed):

PR #3231, opened 2025-11-07
hypershift-operator: ocpbugs-64817 enable multi-arch CPO image for 4.19.7

bennerv:
### What

OCPBUGS-64817 - the CPO override in the HyperShift Operator wasn't built for multi-arch support.  This means if a customer configures a global pull secret on arm64 nodes, there will be an image pull backoff.  

### Why

This should fix that.  Needs to be tested, would probably be good to use our global pull secret e2e test Mike is working on and add an ARM node pool with it.  

Not tested yet, just an image bump.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -841,7 +841,7 @@ clouds:
         image:
           registry: quay.io
           repository: acm-d/rhtap-hypershift-operator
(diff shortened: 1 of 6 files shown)


hbhushan3:
LGTM. Waiting for CI to be green.

PR #3236, opened 2025-11-10
Frontend: Support HCP-Level Autoscaling for All NodePools

mbarnes:
[ARO-17695 - Frontend: Support HCP-Level Autoscaling for All NodePools](https://issues.redhat.com/browse/ARO-17695)

### What

This implements updating of Global Autoscaling properties in an HCPOpenShiftCluster resource, and also ensures Global Autoscaling properties are present whenever the frontend returns an HCPOpenShiftCluster resource in a response.

I also tacked on a regression fix: the `Frontend.OperationResult` method was broken for clusters due to [recent database refactoring](https://github.com/Azure/ARO-HCP/pull/2736).  This is the method that returns the resource upon completion of an async operation like cluster creation.  Noticed while testing.

### Why

This is made complicated by the fact that Cluster Service publishes Global Autoscaling properties at a separate endpoint:

See [/api/clusters_mgmt/v1/clusters/{cluster_id}/autoscaler](https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id__autoscaler), although we're of course using the `aro_hcp` API.

The `hcpOpenShiftClusters` ARM model does not split out autoscaling properties into a nested resource.  To compensate for this difference, the frontend now makes two Cluster Service calls when fetching or updating an `hcpOpenShiftCluster`:

1. `GET|PATCH /api/aro_hcp/v1alpha1/clusters/{cluster_id}`
2. `GET|PATCH /api/aro_hcp/v1alpha1/clusters/{cluster_id}/autoscaler`

Cluster Service does not currently track the progress of Global Autoscaling updates.  This is being worked on, though I'm not aware of a timeframe.  For the time being the best the RP can do is track the cluster update to completion and assume the autoscaling update completed as well.

I strongly advise reviewing one commit at a time and read individual commit messages for additional explanation.

### DIFF
--- a/frontend/test/simulate/mutation_test_utils.go
+++ b/frontend/test/simulate/mutation_test_utils.go
@@ -44,6 +44,7 @@ import (
 
 func trivialPassThroughClusterServiceMock(t *testing.T, testInfo *SimulationTestInfo, initialDataDir fs.FS) error {
 	internalIDToCluster := map[string][]any{}
+	internalIDToAutoscaler := map[string][]any{}
 	internalIDToExternalAuth := map[string][]any{}
 	internalIDToNodePool := map[string][]any{}
 
@@ -78,6 +79,7 @@ func trivialPassThroughClusterServiceMock(t *testing.T, testInfo *SimulationTest
 					return fmt.Errorf("duplicate cluster: %s", obj.HREF())
 				}
 				internalIDToCluster[obj.HREF()] = []any{obj}
+				internalIDToAutoscaler[obj.HREF()] = []any{obj}
 
 			case strings.HasSuffix(dirEntry.Name(), "-external-auth.json"):
 			case strings.HasSuffix(dirEntry.Name(), "-node-pool.json"):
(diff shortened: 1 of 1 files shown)


venkateshsredhat:
> Nope, should be fine - check out the `rp-api-compat-all/parallel` suite. Write a test that uses the RP API directly (not through ARM) to exercise it in your personal environment.

Reference : https://github.com/Azure/ARO-HCP/blob/main/test/e2e/complete_cluster_create.go#L42

mbukatov (changes requested):

test/e2e/cluster_autoscaling.go
```diff
@@ -0,0 +1,190 @@
...
+						Autoscaling: &hcpsdk20240610preview.ClusterAutoscalingProfile{
+							MaxNodesTotal: to.Ptr(autoscalingMaxNodesTotal),
+						},
+					},
+				},
+				5*time.Minute,
+			)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(got.Properties.Autoscaling).ToNot(BeNil())
+			Expect(update.Properties.Autoscaling.MaxNodesTotal).To(Equal(to.Ptr(autoscalingMaxNodesTotal)))
```
stevekuznetsov: Did we expect the e2e to actually autoscale? Simply verifying the API is a start, but I think the real user intent here is to have an auto-scaling node pool, not just to have their intent saved in a database :)

mbarnes: I haven't been able to induce any deterministic autoscaling behavior in my test environment by fiddling with node limits.  Not sure if I'm doing something wrong or if it's not fully wired up further down the stack.

Anyway, that can be a followup.  The scope of this pull request is forwarding autoscaling updates over to Cluster Service correctly.

geoberle (approved):
LGTM
is the synced data 
from CS within cosmos already in good enough shape (and current enough at all times) to use it for basic frontend read operations? if yes we should use it as we can save lots of GET calls towards CS (also the additional calls to resolve the autoscaler link + the nodepool links)

PR #3272, opened 2025-11-12
add overrides for prometheus

venkateshsredhat:


raelga:
Implemented in https://github.com/Azure/ARO-HCP/commit/a10db0448f4a0d9d88f9364d3c13794d203afcb1

PR #3273, opened 2025-11-12
test: Expand initial effort on E2E test against RP

patriksuba:
[ARO-22464](https://issues.redhat.com/browse/ARO-22464)

### What

Introduce functions to populate cluster params from deployments of cluster resources.
Update complete cluster create, no cni cluster, image registry cluster test cases.

### Why

Establish a comprehensive E2E testing framework that can be executed against both Azure Resource Manager (ARM) and Resource Provider (RP) environments


venkateshsredhat (approved):
lgtm

mbukatov (approved):
This is nice! Further documentation and validation will be needed, as this creates dependency between the bicep module files and the deployment framework code, but we have a plan for that already. Also we should drop `disabled-image-registry-cluster.bicep` file because this file is there for the purposes of a particular e2e test only, but we can do that during followup cleanup/improvement work.

PR #3274, opened 2025-11-12
adjust hcp kas burn alert to reduce flappiness

sclarkso:
[](https://issues.redhat.com/browse/AROSLSRE-107?filter=-1)

### What

Changes '5m and' condition to '30m and'

### Why

This alert was very flappy.  This change reduces flapping by 35% based on historic alert data.


sclarkso:
@raelga please review as discussed

PR #3287, opened 2025-11-12
feat: add support for sha images

raelga:
[AROSLSRE-235](https://issues.redhat.com/browse/AROSLSRE-235)

## Why

The image-updater tool currently only supports storing image digests in the standard format with the `sha256:` prefix (e.g., `sha256:abc123...`). However, some configuration fields in our codebase use a `.sha` convention that stores only the hash value without the prefix (e.g., just `abc123...`).

The Prometheus operator images in our configuration use `.sha` fields rather than `.digest` fields, requiring the tool to support both formats to enable automated updates for these images.

## What

This PR extends the image-updater to support flexible digest formats based on field naming conventions:

- **Fields ending with `.digest`**: Store the full digest including `sha256:` prefix (existing behavior)
- **Fields ending with `.sha`**: Store only the hash value without the `sha256:` prefix (new behavior)

**Changes include:**
- Updated `updater.go` to detect `.sha` suffixed paths and strip the `sha256:` prefix before writing
- Added comprehensive test coverage in `updater_test.go` with 6 test cases covering both field types
- Updated `config.yaml` to add Prometheus image tracking (prometheus-operator, prometheus, prometheus-config-reloader, kube-state-metrics)
- Updated `README.md` documentation to explain the digest vs sha field conventions
- Added inline comments in config to organize image sections

**Example:**
```yaml
# Using .digest field
jsonPath: defaults.image.digest  # Stores: sha256:abc123...

# Using .sha field
jsonPath: defaults.image.sha     # Stores: abc123...
```

This enables the image-updater to manage Prometheus and other images that use the `.sha` field convention while maintaining backward compatibility with existing `.digest` fields.

### DIFF
--- a/tooling/image-updater/README.md
+++ b/tooling/image-updater/README.md
@@ -18,6 +18,7 @@ The image-updater supports multiple container registry types with optimized clie
 - **Opt-in Authentication**: Explicitly enable authentication only for private registries
 - **Architecture-Aware**: Automatically filters images by architecture (defaults to amd64)
 - **Multi-Registry Client**: Automatically selects the appropriate client based on registry URL
+- **Flexible Digest Format**: Supports both `.digest` fields (with `sha256:` prefix) and `.sha` fields (hash only)
 
 ## Managed Images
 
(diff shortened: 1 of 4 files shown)


avollmer-redhat (approved):
/lgtm

PR #3295, opened 2025-11-13
feat: Add image tag info to digest update

raelga:
[AROSLSRE-236](https://issues.redhat.com/browse/AROSLSRE-236)

### What

This PR enhances the image-updater tool to include image tag information alongside digest updates and adds a force flag for config regeneration:

1. **Tag information in digest updates**: When updating image digests, the tool now appends the corresponding image tag as an inline comment (e.g., `sha256:abc123... #
v1.2.3`), making it easier to identify which version each digest represents
2. **Force flag**: Added a flag to always regenerate config files, regardless of whether changes are detected
3. **Missing target paths**: Fixed missing image-updater target paths for `acrPull` and `prometheus-operator` in the dev cloud configuration

### Why

Previously, when image digests were updated in configuration files, there was no easy way to determine which semantic version (tag) a particular digest corresponded to.
  This made it difficult to:
- Track which version is currently deployed
- Debug version-related issues
- Understand the history of updates when reviewing config changes

Adding the tag as an inline comment provides immediate visibility into the version information without requiring additional lookups or tooling.

The force flag enables regeneration of config files even when the tool detects no changes, which is useful for ensuring consistency or when testing the tool itself.

- The tag is added as a YAML comment (e.g., `# v1.2.3`) at the end of the line containing the digest
- This preserves file formatting and doesn't affect parsing
- The missing target paths for `config.yaml` ensure that dev cloud configurations also get updated with the latest digests

### DIFF
--- a/tooling/image-updater/internal/clients/utils.go
+++ b/tooling/image-updater/internal/clients/utils.go
@@ -21,17 +21,11 @@ import (
 
 // RegistryClient defines the interface for container registry clients
 type RegistryClient interface {
-	GetArchSpecificDigest(ctx context.Context, repository string, tagPattern string, arch string, multiArch bool) (*ImageInfo, error)
+	GetArchSpecificDigest(ctx context.Context, repository string, tagPattern string, arch string, multiArch bool) (*Tag, error)
 }
 
 type Tag struct {
 	Name         string
 	Digest       string
 	LastModified time.Time
-}
-
-// ImageInfo holds the digest and tag information for an image
-type ImageInfo struct {
-	Digest string
-	Tag    string
 }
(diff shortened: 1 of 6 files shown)


tooling/image-updater/internal/clients/utils.go
```diff
@@ -21,11 +21,17 @@ import (
...
 	Name         string
 	Digest       string
 	LastModified time.Time
 }
+
+// ImageInfo holds the digest and tag information for an image
+type ImageInfo struct {
+	Digest string
+	Tag    string
+}
```
hbhushan3: This is confusing and redundant. Should this be 
```
type ImageInfo struct {
	Digest string
	Tag    Tag
}
```

hbhushan3 (approved):

PR #3306, opened 2025-11-17
Bump Cluster Service image digest for the INT environment

oriAdler:
### What

Bump Cluster Service image digest for the INT environment


machi1990 (changes requested):
The latest digest is cdbd84afe623b7cd58e879a6196325519f52b322a054f2355771639ce2f836d3 let's bump to the latest.

PR #3324, opened 2025-11-18
enable CS debug job to set provision shard 1 in maintenance for stg/prod

miguelsorianod:
We have a new shard with a different id and we want to stop using the old shard with id '1'. This commit enables a CS debug job that takes care of setting the provision shard with id '1' to maintenance state. This debug job is enabled for stage and prod environments. In stage we already have the shard in maintenance state so the execution of the job should be a noop there.


patilsuraj767 (approved):
LGTM 👍

PR #3329, opened 2025-11-18
Revert "frontend: Combine cluster updates with autoscaler updates"

mbarnes:
### What

This reverts commit eb21862fb5765e54f0efa4483ea4f2be57d2328e.

### Why

There are some outstanding validation issues to address. Cluster Service appears to be executing validation that should be skipped on a no-change PATCH. There's also a question around whether some validation checks need rethought for ARO-HCP, which unlike ROSA allows for zero node pools in a cluster.


deads2k:
> Cluster Service appears to be executing validation that should be skipped on a no-change PATCH

None of our execution short-circuits on ratcheting values, so validation is expected to always pass.

Merging to try to fix integration.

PR #3335, opened 2025-11-19
fixing regex in dashboard and fixing deployment

sclarkso:
### What

Fixes regex missing error in `make grafana` and adds dashboard to observability.yaml

### Why
regex change limits which clusters are shown in datasource dropdown, and observability.yaml change makes sure dashboard is actually deployed.

### DIFF
--- a/observability/grafana-dashboards/kas-monitor/kas_monitor_api_slo.json
+++ b/observability/grafana-dashboards/kas-monitor/kas_monitor_api_slo.json
@@ -844,16 +844,13 @@
     "list": [
       {
         "allowCustomValue": false,
-        "current": {
-          "text": "Managed_Prometheus_hcps-usw3scla",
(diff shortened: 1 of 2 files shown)


observability/grafana-dashboards/kas-monitor/kas_monitor_api_slo.json
```diff
@@ -844,16 +844,13 @@
...
-          "value": "hcps-usw3scla"
-        },
+        "current": {},
         "includeAll": true,
         "name": "datasource",
         "options": [],
         "query": "prometheus",
         "refresh": 1,
-        "regex": "",
+        "regex": "^Managed_Prometheus_hcps-.*$",
```
stevekuznetsov: What other ones could possibly exist? If we have leftover crud from previous attempts, I would rather see us remove them than hide them.

PR #3354, opened 2025-11-19
Add timestamps to test logger

avollmer-redhat:
[Slack-Directive](https://redhat-external.slack.com/archives/C075PHEFZKQ/p1763577305076369?thread_ts=1763569418.200829&cid=C075PHEFZKQ)

### What

-Add timestamps to loglines for test logger
-Removes %w from log lines as they aren't used for that.

### Why

Timestamps were needed for logging in per_test_framework.go  which uses ginkgo logging and doesn't allow for timestamps. A wrapper was introduced to add them.

A second commit removes %w which isn't used for logging.

A test application was made to inject values into the test. Timestamps were added and the %w was removed.
Test for test was not committed
Example output:
time="2025-11-19T15:21:29-05:00" level=info msg="creating resource group"
resourceGroup=test-rg-abc123

### DIFF
--- a/test/go.mod
+++ b/test/go.mod
@@ -9,7 +9,6 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources v1.2.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions v1.3.0
 	github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azsecrets v1.4.0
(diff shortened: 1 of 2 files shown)


stevekuznetsov (approved):

avollmer-redhat:
Looks like PROW had a hard time parsing my output after adding the timestamp, I'm switching to JSON.

mbukatov:
@stevekuznetsov this shouldn't have been merged without proper validation, it seems to break log reporting in openshift CI

PR #3366, opened 2025-11-20
Identity reuse for e2e test runs

roivaz:
https://issues.redhat.com/browse/AROSLSRE-304

### Summary

This PR introduces a **dual‑mode managed identity workflow** for the E2E suite, allowing tests to either:

- **Reuse pre‑created identities from a pool of resource groups**, or  
- **Create identities scoped to the cluster resource group** and tied to the test’s lifecycle,

with behavior controlled by `POOLED_IDENTITIES` and `LEASED_MSI_CONTAINERS`. In CI, `LEASED_MSI_CONTAINERS` is populated by the Boskos leasing server, providing the CI‑configured number of identity containers for each E2E run. Bicep templates and the Go test framework are refactored so that identity management is explicit and centralized, and so the choice of mode is per‑run rather than per‑test.

Additionally, the cleanup command in `test/cmd/aro-hcp-tests/cleanup` is improved with `--is-development` and `--tracked` flags, making it easier to tailor cleanup behavior for local dev environments and to restrict deletion to explicitly tracked resources.

---

### 1. Subscription‑scoped `managed-identities.bicep`

- **Problem:** We need to support both:
  - Identities created once in **MSI pool RGs** and reused across tests.
  - Identities created per‑cluster in the **cluster RG**.
- **Decision:** Run `managed-identities.bicep` at **subscription scope**, orchestrating:
  - **MSI‑scoped assignments** in the MSI RG (`modules/msi-scoped-assignments.bicep`).
  - **Non‑MSI‑scoped assignments** in the cluster RG (`modules/non-msi-scoped-assignments.bicep`).
- **Parameters:** `msiResourceGroupName`, `clusterResourceGroupName`, `pooledIdentities`, `useMsiPool`.
  - When `useMsiPool=true`, we reuse pre‑created identities in the MSI RG.
  - When `useMsiPool=false`, we call `cluster-identities.bicep` to create identities in the cluster RG and wire its outputs into both assignment submodules.
- **Cleanup robustness:** Role assignments in MSI RGs use stable names; if some cleanup is missed, later runs reuse the same assignments rather than accumulating unbounded state.

### 2. 3‑state leasing model with shared YAML state and history

- **Problem:** Ginkgo parallelism uses **multiple OS processes**, so we need cross‑process coordination to avoid:
  - Over‑leasing the same MSI RG to multiple tests.
  - Deadlocks when tests require more than one identity container.
- **Decision:** Use a **shared YAML state file** with `flock` and a 3‑state leasing model:
  - **State file:** `${SHARED_DIR}/identities-pool-state.yaml`.
  - **Entry structure:** `leasedIdentityPoolEntry` with:
    - `Current` `leaseEntry` (`state: free|assigned|busy`, `leasedBy`, `transitionedAt`).
    - `History []leaseEntry` recording all transitions for traceability.
- **States:**
  - `free`: container is available.
  - `assigned`: reserved for a specific spec (by `specID()`), but not yet “checked out”.
  - `busy`: spec has “checked out” the container and is actively using it.
- **Operations:**
  - `AssignNTo(me, n)`:
    - Acquires a file lock, reads state, and tries to mark **exactly `n` free entries** as `assigned` to `me`.
    - If fewer than `n` free entries exist, returns `ErrNotEnoughFreeIdentityContainers` **without writing** the file.
  - `UseNextAssigned(me, registerFn)`:
    - Finds the next entry in `assigned` state for `me`, transitions it to `busy`, records history, and calls `registerFn` to track the RG for cleanup.
  - `ReleaseByContainerName(rg, cleanupFn)`:
    - Transitions any non‑`free` entry for that RG back to `free`, appends to history, and calls `cleanupFn` (best‑effort).

### 3. Up‑front reservation to avoid deadlocks

- **Problem:** Some specs require more than one identity container; naive “lease as you go” can deadlock when the pool is partially consumed.
- **Decision:** Each test **declares its maximum container demand up front** and reserves them atomically:
  - Tests call `tc.AssignIdentityContainers(ctx, N, 60*time.Second)` at the start of the `It` block, where `N` is the maximum number of containers that spec may consume.
  - `AssignIdentityContainers` loops on `AssignNTo`:
    - If `ErrNotEnoughFreeIdentityContainers`, it **waits and retries**, so tests **block** instead of failing or deadlocking.
    - On success, the spec owns N `assigned` containers; per‑use calls to `ResolveIdentitiesForTemplate` / `DeployManagedIdentities` drive individual transitions to `busy`.
- **Result:** Either a spec reserves all of the containers it needs or it waits until they’re available; no partial reservations remain in the shared file, and no deadlocks arise from multi‑container specs.
---

## Testing status

Verified in personal dev environment

### DIFF
--- a/test/e2e-setup/bicep/modules/msi-scoped-assignments.bicep
+++ b/test/e2e-setup/bicep/modules/msi-scoped-assignments.bicep
@@ -73,11 +73,17 @@ resource serviceManagedIdentity 'Microsoft.ManagedIdentity/userAssignedIdentitie
 //
 // R O L E   D E F I N I T I O N S
 //
+
+// Reader: Grants permission to read Azure resources
+// https://www.azadvertizer.net/azrolesadvertizer/acdd72a7-3385-48ef-bd42-f606fba81ae7.html
 var readerRoleId = subscriptionResourceId(
   'Microsoft.Authorization/roleDefinitions',
   'acdd72a7-3385-48ef-bd42-f606fba81ae7'
 )
 
+// Azure Red Hat OpenShift Federated Credential: Create, update and delete federated credentials on user assigned 
+// managed identities in order to build a trust relationship between the managed identity, OpenID Connect (OIDC), and the service account.
+// https://www.azadvertizer.net/azrolesadvertizer/ef318e2a-8334-4a05-9e4a-295a196c6a6e.html
 var federatedCredentialsRoleId = subscriptionResourceId(
   'Microsoft.Authorization/roleDefinitions',
   'ef318e2a-8334-4a05-9e4a-295a196c6a6e'
--- a/test/e2e-setup/bicep/modules/non-msi-scoped-assignments.bicep
+++ b/test/e2e-setup/bicep/modules/non-msi-scoped-assignments.bicep
@@ -68,7 +68,9 @@ resource clusterApiAzureMi 'Microsoft.ManagedIdentity/userAssignedIdentities@202
   scope: resourceGroup(resourceGroupName)
 }
 
-// Azure Red Hat OpenShift Hosted Control Planes Cluster API Provider
+// Azure Red Hat OpenShift Hosted Control Planes Cluster API Provider: Enables permissions to allow cluster API to 
+// manage nodes, networks and disks for OpenShift cluster.
+// https://www.azadvertizer.net/azrolesadvertizer/88366f10-ed47-4cc0-9fab-c8a06148393e.html
 var hcpClusterApiProviderRoleId = subscriptionResourceId(
   'Microsoft.Authorization/roleDefinitions',
   '88366f10-ed47-4cc0-9fab-c8a06148393e'
@@ -94,7 +96,9 @@ resource controlPlaneMi 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-0
   scope: resourceGroup(resourceGroupName)
 }
 
-// Azure Red Hat OpenShift Hosted Control Planes Control Plane Operator
+// Azure Red Hat OpenShift Hosted Control Planes Control Plane Operator: Enables the control plane operator to read 
+// resources necessary for OpenShift cluster.
+// https://www.azadvertizer.net/azrolesadvertizer/fc0c873f-45e9-4d0d-a7d1-585aab30c6ed.html
 var hcpControlPlaneOperatorRoleId = subscriptionResourceId(
   'Microsoft.Authorization/roleDefinitions',
   'fc0c873f-45e9-4d0d-a7d1-585aab30c6ed'
@@ -130,7 +134,8 @@ resource cloudControllerManagerMi 'Microsoft.ManagedIdentity/userAssignedIdentit
   scope: resourceGroup(resourceGroupName)
 }
 
-// Azure Red Hat OpenShift Cloud Controller Manager
+// Azure Red Hat OpenShift Cloud Controller Manager: Manage and update the cloud controller manager deployed on top of OpenShift.
+// https://www.azadvertizer.net/azrolesadvertizer/a1f96423-95ce-4224-ab27-4e3dc72facd4.html
 var cloudControllerManagerRoleId = subscriptionResourceId(
   'Microsoft.Authorization/roleDefinitions',
   'a1f96423-95ce-4224-ab27-4e3dc72facd4'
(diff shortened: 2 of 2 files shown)


test/e2e-setup/bicep/modules/managed-identities.bicep
```diff
@@ -44,9 +75,9 @@ resource keyVault 'Microsoft.KeyVault/vaults@2024-12-01-preview' existing = {
 // C L U S T E R   A P I   A Z U R E   M I
 //
 
-resource clusterApiAzureMi 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' = {
-  name: '${clusterName}-cp-cluster-api-azure-${randomSuffix}'
```
stevekuznetsov: Yeah, if we have just one set per RG and they're not cluster-dependent, let's just make this called `cp-cluster-api-azure`, etc

test/util/framework/identities_helper.go
```diff
@@ -0,0 +1,627 @@
...
+			},
+		})
+	}
+
+	data, err := yaml.Marshal(entries)
+	if err != nil {
+		return fmt.Errorf("failed to marshal managed identities pool state: %w", err)
+	}
+
+	if err := os.WriteFile(statePath, data, 0644); err != nil {
```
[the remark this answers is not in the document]
roivaz: Done. Now initialization is performed under flock as well.

test/util/framework/identities_helper.go
```diff
@@ -0,0 +1,660 @@
...
+	return deploymentResult, nil
+}
+
+func (tc *perItOrDescribeTestContext) CreateIdentitiesPoolStateFile() error {
+
+	f, err := os.OpenFile(tc.msiPoolStateFilePath(), os.O_RDWR|os.O_CREATE, 0666)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
```
stevekuznetsov: general: log your errors

test/util/framework/identities_helper.go
```diff
@@ -0,0 +1,660 @@
...
+	}
+
+	if err = state.writeUnlocked(); err != nil {
+		return "", fmt.Errorf("failed to write managed identities pool state file: %w", err)
+	}
+
+	return leasedRG, nil
+}
+
+// ReserveContainers attempts to reserve n free identity containers by marking
```
stevekuznetsov: bad godoc

test/util/framework/identities_helper.go
```diff
@@ -0,0 +1,660 @@
...
+	f, err := os.OpenFile(path, os.O_RDWR, 0)
+	if err != nil {
+		return LeasedIdentityPoolState{}, fmt.Errorf("failed to open managed identities pool state file %s: %w", path, err)
+	}
+
+	return LeasedIdentityPoolState{file: f}, nil
+}
+
+func (state *LeasedIdentityPoolState) Close() {
+	_ = state.file.Close()
```
stevekuznetsov: log your errors

test/util/framework/identities_helper.go
```diff
@@ -0,0 +1,660 @@
...
+
+	if err := state.initializeUnlocked(leasedRGs); err != nil {
+		return fmt.Errorf("failed to initialize managed identities pool state: %w", err)
+	}
+
+	return nil
+}
+
+func (tc *perItOrDescribeTestContext) AssignIdentityContainers(ctx context.Context, count uint8, waitBetweenRetries time.Duration) error {
+	state, err := NewLeasedIdentityPoolState(tc.msiPoolStateFilePath())
```
stevekuznetsov: Why do we allow creating many instances of this when we use flock under the hook to limit concurrency anyway? A singleton in the `perInvocation` explains the intent better

test/util/framework/identities_helper.go
```diff
@@ -0,0 +1,660 @@
...
+	}
+	return nil
+}
+
+func (tc *perItOrDescribeTestContext) msiPoolStateFilePath() string {
+	return filepath.Join(tc.perBinaryInvocationTestContext.sharedDir, msiPoolStateFileName)
+}
+
+func (tc *perItOrDescribeTestContext) registerLeasedIdentityContainer(resourceGroupName string) error {
+	tc.contextLock.Lock()
```
stevekuznetsov: Why is this state disjoint from the on-disk state? Prefer one source of truth and one lock

test/util/framework/identities_helper.go
roivaz: Fixed

test/util/framework/identities_helper.go
roivaz: Fixed

test/util/framework/identities_helper.go
roivaz: Moved to a singleton in the perInvocationTestContext. I also removed the state file initialization with `specs.AddBeforeAll()` as that was running once per process anyway. The instance is created once per process in the perInvocationTestContext and state file initialization is multi-process safe.

test/util/framework/identities_helper.go
roivaz: Cleanup now works off the state file without an intermediate state

PR #3367, opened 2025-11-20
Bump CS image digest to the latest - Dev & INT

oriAdler:
To resolve CVE-2025-6965, UBI minimal was updated to the latest version.

PR #3369, opened 2025-11-20
bump cs image in dev & int

ziccardi:


machi1990:
Can we also bump INT at the same time?

This supersedes https://github.com/Azure/ARO-HCP/pull/3367 cc @oriAdler

PR #3378, opened 2025-11-21
remove unnecessary patches from the cluster create transaction

deads2k:
With the separation of create and update, along with the inclusion of typed for the database calls, we no longer have a need to perform create plus patching to create a full view of the object.  it is more clear to create the resource we want to store and store that.  Additionally, we want to validate what we actually store, so avoiding the patch is important to that end.

### DIFF
--- a/frontend/pkg/frontend/cluster.go
+++ b/frontend/pkg/frontend/cluster.go
@@ -226,8 +226,6 @@ func (f *Frontend) createHCPCluster(writer http.ResponseWriter, request *http.Re
 		return err
 	}
 
-	operationRequest := database.OperationRequestCreate
-
 	// Initialize top-level resource fields from the request path.
 	// If the request body specifies these fields, validation should
 	// accept them as long as they match (case-insensitively) values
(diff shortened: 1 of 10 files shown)


frontend/pkg/frontend/cluster.go
```diff
@@ -244,11 +242,24 @@ func (f *Frontend) createHCPCluster(writer http.ResponseWriter, request *http.Re
...
 	}
 	newExternalCluster.Normalize(newInternalCluster)
+
+	// set fields that were not included during the conversion, because the user does not provide them or because the
+	// data is determined live on read.
+	newInternalCluster.SystemData = systemData
+	// Clear the user-assigned identities map since that is reconstructed from Cluster Service data.
+	// TODO we'd like to have the instance complete when we go to validate it.  Right now validation fails if we clear this.
+	// TODO we probably update validation to require this field is cleared.
+	//newInternalCluster.Identity.UserAssignedIdentities = nil
```
mbarnes: I imagine this will get cleaner once the RP itself is fetching client and principal IDs for user-assigned identities instead of depending on Cluster Service to do it.  That might be a good addition to the initial set of tasks for the new backend.

PR #3384, opened 2025-11-21
switch to use internal API types from transaction results

deads2k:
shows refactors to start eliminating extraneous variables from rest handler to align the logic to run on the internal versions.


frontend/pkg/frontend/cluster.go
mbarnes: This comment might be losing its relevance amidst all the refactoring.  I would need to go back and re-read it in its original context from earlier this year.

But I can tell you the motivation for the special handling:

The Tags field is odd in that it's always fully replaced, _**even in PATCH requests**_.  You can't add or remove individual tags through JSON Merge Patch like you might expect.

To quote the Resource Provider Contract:
> **Common scenarios for implementing Patch**
>
> Tags
> 
> A common scenario for the usage of a Patch operation is to update the tags section of a tracked resource. The behavior for a Patch operation of the tags property is to replace the tags property with the keys and values provided with the Patch operation. Eg: if the resource currently has tag1 and tag2, and a Patch request sends tag3, the final resource should have tag3 (and *not* tag1, tag2 and tag3).

PR #3391, opened 2025-11-24
kusto cross sub awareness

geoberle:
### What

the goal of this PR is to make the kusto instance mgmt cross-subscription aware.

challenge 1: the role assignments for the arobit MI to the geo-kusto instance that happens in the SVC and MGMT pipelines, assumes the kusto instance lives in the same subscription as svc or mgmt clusters. this assumption is true in RH DEV and INT, but will fail within STAGE and PROD where dedicated subscriptions are used for all bits and pieces.

challenge 2: the `kusto.staticKustoName` variable acts as a dev-only config option.

challenge 3: kusto details like the ingestion endpoint are calculated based on the kusto-regions.yaml geo->region mapping file. as such, the information about what kusto instance to use for a region is not accessible outside the arobit helm chart.

all three challenges can be adressed with the same solution

* all kusto instances will be created in the global subscription and dedicated resourcegroup `kusto.rg`
* instead of a dev-only `kusto.staticKustoName` property, we have a `kusto.kustoName` property in config.yaml
  * to make sure we share the same kusto instance for everything in DEV, we statically define `kusto.rg` and `kusto.kustoName` in config.yaml under `clouds.dev.defaults`
  * for MSFT envs, we use the `ev2.geoShortId` to form `kusto.kustoName` and `kusto.rg` so every region inherently gets their right kusto instance
* in order to make the correct kusto endpoint available to the components that need it (arobit, admin api, native helm step inline diagnostics https://github.com/Azure/ARO-Tools/pull/165), we will have a kusto-lookup.bicep output-only step that provides there respective data via output chaining. hence the ingestion endpoint logic in the arobit helm chart is not required anymore

alternative for global subscription: we could introduce a dedicate kusto subscription to use for all kusto instances

### DIFF
--- a/acm/deploy/helm/testdata/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_mce_config.yaml
+++ b/acm/deploy/helm/testdata/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_mce_config.yaml
@@ -1,7 +1,6 @@
 ---
 # Source: multicluster-engine-local-cluster/charts/policy/crds/agent.open-cluster-management.io_klusterletaddonconfigs_crd.yaml
 # Copyright (c) 2020 Red Hat, Inc.
-
 # Copyright Contributors to the Open Cluster Management project
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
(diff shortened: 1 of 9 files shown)


config/config.schema.json
```diff
@@ -1551,9 +1551,6 @@
     "kusto": {
       "type": "object",
       "properties": {
-        "subscription": {
```
geoberle: i had this introduced on friday but realized we don't need it when we go for the global subscription

PR #3392, opened 2025-11-24
ARO-17798

oriAdler:
[ARO-17798
](https://issues.redhat.com/browse/ARO-17798)
### What

Create a Contributor role assignment for aro-dev-first-party2 to enable the mock FPA to call the Check Access API in Dev & INT environments.

### Why

Enable user infra permissions inflights epic.

### DIFF
--- a/dev-infrastructure/Makefile
+++ b/dev-infrastructure/Makefile
@@ -165,6 +165,16 @@ create-mock-identities: configurations/mock-identities.bicepparam
 		--scope "/subscriptions/$(shell az account show --query id --output tsv)" \
 		--only-show-errors
 
+# TODO: this is a workaround to enable the mock FPA to call the Check Access API with a built-in role.
+# The permissions should be reduced to match "dev-first-party-mock" role definition (ARO-17798).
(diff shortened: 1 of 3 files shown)


dev-infrastructure/Makefile
```diff
@@ -165,6 +165,16 @@ create-mock-identities: configurations/mock-identities.bicepparam
 		--scope "/subscriptions/$(shell az account show --query id --output tsv)" \
 		--only-show-errors
 
+# TODO: this is a workaround to enable the mock FPA to call the Check Access API with a built-in role.
+# The permissions should be reduced to match "dev-first-party-mock" role definition (ARO-17798).
+.PHONY: assign-role-for-rbac
+
+assign-role-for-rbac:
```
miguelsorianod: Does this need to be incorporated somewhere so in case an env is recreated it is configured correctly? and if so, does it need to be idempotent?

miguelsorianod: The target name and the script name seem very generic whereas the task is highly specific

PR #3407, opened 2025-11-25
admin-api: add base endpoint for hcp related resources

geoberle:
### What

* add resource id based resource paths
* context handler for resource id extraction
* context handler for client principal name extraction from geneva action
* demo /admin/v1/hcp/$resourceid/helloworld endpoint

### DIFF
--- a/admin/server/middleware/hcpresourceid.go
+++ b/admin/server/middleware/hcpresourceid.go
@@ -17,7 +17,6 @@ package middleware
 import (
 	"fmt"
 	"net/http"
-	"path"
 	"strings"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


admin/server/middleware/hcpresourceid.go
```diff
@@ -0,0 +1,109 @@
...
+			subscriptionID,
+			"resourceGroups",
+			resourceGroupName,
+			"providers",
+			strings.ToLower(api.ProviderNamespace),
+			strings.ToLower(api.ClusterResourceTypeName),
+			resourceName,
+		)
+
+		resourceID, err := azcorearm.ParseResourceID(resourceIDPath)
```
stevekuznetsov: It boggles my mind a bit that there's no constructor for a resource ID that is not stringly typed - is formatting into a string and parsing the best we can do?

stevekuznetsov (approved):

admin/server/middleware/hcpresourceid.go
geoberle: there is only an internal constructor and lots of unexported functions :(

PR #3408, opened 2025-11-25
Templatize: add retries to step execution

janboll:
### What

- Add retries to step execution
- Configure retries for kusto ingest

### Why

Steps might fail for certain reasons which can only be solved by retries (i.e. Kusto grant ingest on new managed identity). By adding retries we can ensure certain errors are retried

### DIFF
--- a/config/dev.digests.yaml
+++ b/config/dev.digests.yaml
@@ -15,10 +15,10 @@ clouds:
           westus3: a9cd837a398be8dad7f9594fc2eca1009d39578f5d70db6c3112d731c8b91cf0
       pers:
         regions:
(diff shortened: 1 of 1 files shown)


raelga (dismissed):

tooling/templatize/pkg/pipeline/run.go
```diff
@@ -481,15 +482,28 @@ func executeNode(logger logr.Logger, executor Executor, graphCtx *graph.Graph, n
 	var output Output
```
stevekuznetsov: You probably want to make sure that timing metadata, etc, correctly shows re-tries

janboll: Did not think about that, of course... 
Will look into that

PR #3414, opened 2025-11-26
Fail fast during e2e cleanup if hcpOpenShiftClusters resources still exist after complete deletion

tsatam:
[ARO-22847](https://issues.redhat.com/browse/ARO-22847)

### What

Add an explicit step during e2e test cleanup's `DeleteHCPCluster` to check and confirm the cluster is gone, and fail fast if the cluster resource still exists. While the existing implementation waits for the asynchronous DELETE operation to complete before proceeding, there may be cases where the RP reports a successful deletion before the resource is actually removed.  

Note: Instead of this approach, we may prefer instead to poll and wait for e.g. a cluster in `Deleting` provisioning state to actually be removed before proceeding, but if our RP is reporting a cluster as successfully deleted while the resource still exists we would likely consider that to be a blocking failure. 

### Why

We've observed that the subsequent `DeleteResourceGroup` cleanup step sometimes fails due to the resource group containing an hcpOpenShiftCluster resource still in Deleting. Ideally, all of these resources should be completely gone by the time we delete the resource group.

### DIFF
--- a/test/util/framework/hcp_helper.go
+++ b/test/util/framework/hcp_helper.go
@@ -16,7 +16,9 @@ package framework
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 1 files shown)


raelga (changes requested):
@tsatam https://redhat.atlassian.net/browse/ARO-22847 was closed, is this PR still valid?

raelga:
After reviewing with @tsatam

/close

PR #3438, opened 2025-12-01
test: Check operation result content in helper functions

mbarnes:
### What

This integrates a check into the E2E framework to ensure the RP returns the expected resource model upon completion of an asynchronous create or update operation.

### Why

There was a recent regression in the RP where the operation result endpoint broke for clusters.  This went undetected in our E2E suite.

Operation result endpoints — which have a resource type of `Microsoft.RedHatOpenShift/hcpOperationResults` — are called by ARM (or the [runtime.Poller](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime#Poller) for E2E) when an asynchronous operation completes successfully.  The endpoint is supposed to behave the same as if the completed operation had been synchronous.  So for example, for a resource create or update operation the endpoint responds with **201 Created** or **200 OK** and the current resource model in the body.  For a resource delete operation the endpoint responds with **204 No Content** and no body.

### DIFF
--- a/test/util/framework/hcp_helper.go
+++ b/test/util/framework/hcp_helper.go
@@ -183,7 +183,11 @@ func UpdateHCPCluster(
 
 	switch m := any(operationResult).(type) {
 	case hcpsdk20240610preview.HcpOpenShiftClustersClientUpdateResponse:
-		// Make sure the operation result content matches the current cluster model.
+		// Verify the operationResult content matches the current cluster model.
+		// When an asynchronous operation completes successfully, the RP's result
+		// endpoint for the operation is supposed to respond as though the operation
+		// were completed synchronously. In production, ARM would call this endpoint
+		// automatically. In this context, the poller calls it automatically.
 		expect, err := GetHCPCluster(ctx, hcpClient, resourceGroupName, hcpClusterName)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


test/util/framework/hcp_helper.go
```diff
@@ -36,6 +38,29 @@ import (
 	hcpsdk20240610preview "github.com/Azure/ARO-HCP/test/sdk/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
 )
 
+// CheckOperationResult ensures the result model returned by a runtime.Poller
+// matches the resource model returned from a GET request.
+func CheckOperationResult(expectModel, resultModel any) error {
```
deads2k: nit: private

test/util/framework/hcp_helper.go
```diff
@@ -294,8 +328,15 @@ func CreateOrUpdateExternalAuthAndWait(
 
 	switch m := any(operationResult).(type) {
 	case hcpsdk20240610preview.ExternalAuthsClientCreateOrUpdateResponse:
-		// TODO someone may want this return value.  We'll have to work it out then.
-		//fmt.Printf("#### got back: %v\n", spew.Sdump(m))
+		// Make sure the operation result content matches the current external auth model.
```
deads2k: minor, but still required: please add the details of your slack conversation with me as a comment on each of these to explain what we're testing

PR #3443, opened 2025-12-01
test/util: fix timing output names

stevekuznetsov:
The previous approach did not correctly determine the "fully qualified name" of the test when hashing it to create the output timing file. Who knew it would be *this* hard to figure out the name for a Ginkgo spec.


mgahagan73 (approved):
/lgtm

PR #3453, opened 2025-12-02
add controller type to report status

deads2k:
Adds a controller type to report status.  This is highlighting some pre-work.

### DIFF
--- a/internal/database/types_controller.go
+++ b/internal/database/types_controller.go
@@ -1,0 +1,58 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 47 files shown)


internal/database/types_controller.go
```diff
@@ -0,0 +1,125 @@
...
+	TypedDocument `json:",inline"`
+
+	ControllerProperties ControllerProperties `json:"properties"`
+}
+
+var _ ResourceProperties = &Controller{}
+
+type ControllerProperties struct {
+	// ExternalID is the Azure resource ID of the type this is associated with.
+	ExternalID *azcorearm.ResourceID `json:"externalId,omitempty"`
```
jharrington22: In what cases would the resource ID be empty?

mbarnes: Shouldn't ever be empty, IIUC.  We maybe overuse the `omitempty` JSON tag.

mbarnes (approved):

PR #3459, opened 2025-12-02
api: Remove old validation struct tags from resource models

mbarnes:
### What

Removes lingering validation struct tags that are no longer used.

### Why

I've observed some confusion around whether they're still used or wanted.


internal/validation/validate_externalauth.go
```diff
@@ -181,12 +181,11 @@ var (
 func validateTokenIssuerProfile(ctx context.Context, op operation.Operation, fldPath *field.Path, newObj, oldObj *api.TokenIssuerProfile) field.ErrorList {
 	errs := field.ErrorList{}
 
-	// TODO why didn't we make this a URL?
```
mbarnes: Answer: To do so would have presumed the string was a syntactically valid URL before validation could check it.  Also, conversion from external to internal resource models is supposed to never fail.

PR #3460, opened 2025-12-02
Align NodePool frontend CRUD to cluster CRUD on internal types

deads2k:
This brings our mutation into the same pattern of decode to internal, manipulate, validate, store, then finally convert back to external for encoding.

builds on https://github.com/Azure/ARO-HCP/pull/3379 to convert nodepools.

### DIFF
--- a/frontend/pkg/frontend/cluster.go
+++ b/frontend/pkg/frontend/cluster.go
@@ -436,7 +436,7 @@ func (f *Frontend) updateHCPCluster(writer http.ResponseWriter, request *http.Re
 		return err
 	}
 
-	return f.updateHCPClusterInCosmos(ctx, writer, request, newInternalCluster, oldInternalCluster)
+	return f.updateHCPClusterInCosmos(ctx, writer, request, http.StatusOK, newInternalCluster, oldInternalCluster)
 }
 
 func decodeDesiredClusterPatch(ctx context.Context, oldInternalCluster *api.HCPOpenShiftCluster) (*api.HCPOpenShiftCluster, error) {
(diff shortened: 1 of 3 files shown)


frontend/pkg/frontend/cluster.go
```diff
@@ -563,62 +583,67 @@ func (f *Frontend) updateHCPClusterInCosmos(ctx context.Context, writer http.Res
 		return err
 	}
 
+	// TODO is patch supposed to be status accepted?
```
deads2k: Wondered about this.  Nothing is failing, but should it be different?

frontend/pkg/frontend/cluster.go
```diff
@@ -563,62 +583,67 @@ func (f *Frontend) updateHCPClusterInCosmos(ctx context.Context, writer http.Res
...
 	}
 	return nil
 }
 
-func decodeDesiredClusterPatch(ctx context.Context, oldInternalCluster *api.HCPOpenShiftCluster) (*api.HCPOpenShiftCluster, error) {
-	versionedInterface, err := VersionFromContext(ctx)
+// mergeToInternalCluster renders a CS Cluster object in JSON format, applying
+// the necessary conversions for the API version of the request.
+// TODO this overwrite will transformed into a "set" function as we transition fields to ownership in cosmos
+func mergeToInternalCluster(csCluster *arohcpv1alpha1.Cluster, internalCluster *api.HCPOpenShiftCluster) (*api.HCPOpenShiftCluster, error) {
```
deads2k: next three functions are moved here and delegate to each other so they are consistent. There were two slightly different flows before.

frontend/pkg/frontend/cluster.go
```diff
@@ -563,62 +583,67 @@ func (f *Frontend) updateHCPClusterInCosmos(ctx context.Context, writer http.Res
...
+	if database.IsResponseError(err, http.StatusNotFound) {
+		return nil, arm.NewResourceNotFoundError(resourceID)
+	}
 	if err != nil {
-		return err
+		return nil, err
 	}
-
-	return f.updateHCPClusterInCosmos(ctx, writer, request, newInternalCluster, oldInternalCluster)
+	return f.readInternalClusterFromClusterService(ctx, internalCluster)
```
deads2k: delegated for consistency (found when re-using in nodepool cluster lookup)

frontend/pkg/frontend/error.go
```diff
@@ -53,6 +57,17 @@ func writeError(ctx context.Context, w http.ResponseWriter, err error, args ...i
 
 	logger.Error(fmt.Sprintf("%v", err), args...) // fmt used to handle nil
 
+	var ocmError *ocmerrors.Error
+	if errors.As(err, &ocmError) {
```
deads2k: convenient for error handling when we're able to do this.

frontend/pkg/frontend/cluster.go
mbarnes: Yes, good catch.  Response code should be 202 (Accepted).

Resource Provider Contract quote (emphasis mine):
> **Response**
>
> The response includes an HTTP status code, a set of response headers, and a response body.
>
> **Status Code**
>
> The resource provider must return 200 (OK) to indicate that the synchronous Patch operation completed successfully. **202 (Accepted) must be returned to indicate that the operation will complete asynchronously.**
>
> If the resource group *or* resource does not exist, 404 (NotFound) should be returned.

mbarnes (changes requested):
Had a concern about the top-level error handling, and also that `http.StatusOK` -> `http.StatusAccepted` you spotted needs fixed.  The rest LGTM.

PR #3495, opened 2025-12-04
test: increase external auth list and verify timeout to 90m

raelga:
Since the timeout was fixed in https://github.com/Azure/ARO-HCP/pull/3448 and updated in https://github.com/Azure/ARO-HCP/pull/3466, this test is failing frecuently on all enviroments: int, stage and prod. Increasing the timeout to 90 minutes will allow to verify if that is just that the custer needs more than 45 minutes to be ready or there is another issue. This should be reverted at some point, this is just a test.

### DIFF
--- a/test/e2e/external_auth_list_and_verify.go
+++ b/test/e2e/external_auth_list_and_verify.go
@@ -68,7 +68,7 @@ var _ = Describe("Customer", func() {
 			}
 
 			// Create ARO HCP cluster
-			timeout := 45 * time.Minute
+			timeout := 90 * time.Minute
 			deploymentCtx, deploymentCancel := context.WithTimeoutCause(ctx, timeout, fmt.Errorf("timeout '%f' minutes exceeded during external auth list and verify test", timeout.Minutes()))
 			defer deploymentCancel()
 


sclarkso (approved):

tooling/templatize/pkg/pipeline/shell.go
```diff
@@ -287,9 +287,11 @@ func checkCachedOutput[T any](logger logr.Logger, data any, stepCacheDir string)
 
 		var output T
 		if err := json.Unmarshal(content, &output); err != nil {
-			return "", nil, nil, fmt.Errorf("failed to deserialize output: %w", err)
+			logger.V(2).Info("Failed to deserialize cached output, ignoring cache", "err", err, "content", string(content))
+			// Don't fail, just proceed without cache
```
stevekuznetsov: Please no. Figure out what corrupted it and fix it.

raelga:
Didn't improve the test.

tooling/templatize/pkg/pipeline/shell.go
mbukatov: Agreed, this was identified as a bug.

PR #3498, opened 2025-12-04
internal: divulge errors before swallowing them

stevekuznetsov:
There is no simple way to get a logger (even the default one) into the internal package, so I'm not trying here. A simple JSON record to stdout it is. We _must not_ swallow errors silently.

### DIFF
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -16,6 +16,7 @@ package api
 
 import (
 	"encoding/json"
+	"fmt"
 	"iter"
 	"net/http"
 	"reflect"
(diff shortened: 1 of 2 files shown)


bennerv (dismissed):

internal/api/utils.go
```diff
@@ -193,6 +194,8 @@ func ApplyRequestBody(requestMethod string, body []byte, v any) *arm.CloudError
 	case http.MethodPatch:
 		originalData, err := json.Marshal(v)
 		if err != nil {
+			// there doesn't seem to be any sensible way to get at the logger in here, please forgive me
+			fmt.Printf(`{"level":"ERROR", "msg":"ApplyRequestBody: failed to marshal original data: %q"}\n`, err.Error())
 			return arm.NewInternalServerError()
```
deads2k: /hold

just return err

PR #3499, opened 2025-12-04
track errors by line numbers for logging sources

deads2k:
this will make debugging where an error originates easier to track, while keeping the logging centralized.  It will also track teh stack as it pops up.

### DIFF
--- a/frontend/pkg/frontend/cluster.go
+++ b/frontend/pkg/frontend/cluster.go
@@ -34,6 +34,7 @@ import (
 	"github.com/Azure/ARO-HCP/internal/conversion"
 	"github.com/Azure/ARO-HCP/internal/database"
 	"github.com/Azure/ARO-HCP/internal/ocm"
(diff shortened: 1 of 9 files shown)


stevekuznetsov (approved):

internal/utils/line_tracking_error_test.go
```diff
@@ -0,0 +1,135 @@
...
+func TestReturningNilForNil(t *testing.T) {
+	wrappedErr := TrackError(nil)
+	if wrappedErr != nil {
+		t.Error("expected nil for nil input error")
+	}
+}
+
+func TestLineTrackingError_Error(t *testing.T) {
+	originalErr := errors.New("database connection failed")
+	wrappedErr := TrackError(originalErr) // This is line 55
```
bennerv:
```suggestion
	wrappedErr := TrackError(originalErr) // This is line 43

PR #3528, opened 2025-12-08
Refactor and Extend Timeout of Admin Credential Revocation

bennerv:
### What

[aro-21682](https://issues.redhat.com//browse/aro-21682): refactor the admin credential lifecycle test and extend the timeout until OCPBUGS-62177 is resolved.

### DIFF
--- a/test/e2e/admin_credential_lifecycle.go
+++ b/test/e2e/admin_credential_lifecycle.go
@@ -21,13 +21,15 @@ import (
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mbarnes (commented):
Changes look reasonable to me, but I wonder if this test should really be using `labels.AroRpApiCompatible`.  Conflicts based on provisioning state of tracked resources appears to be something ARM handles directly, and the RP's logic for this is redundant.  So it seems to me like this test should be talking to ARM, not the RP directly.

If we do switch to using ARM, we should verify this in RP logs.  Check if the rejected credential request even made it to the RP.

PR #3534, opened 2025-12-09
feat: add new make target for local CS dev

ahitacat:
### What
Add a new make target that will deploy infra and services, configure the permissions and create the provision shard config for cluster-service.

### Why
This was part of the make infra.all, but it recently change. Will solve minor issues we have found trying to run CS locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -288,6 +288,36 @@ personal-dev-env:
 endif
 .PHONY: personal-dev-env
 
+#
+# Local Cluster Service Development Environment
+#
+ifeq ($(DEPLOY_ENV),pers)
+local-pers-dev-env: personal-dev-env
+	@echo ""
+	@echo "===================================================================="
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -288,6 +288,36 @@ personal-dev-env:
...
+	@echo "Generating local provision-shard config..."
+	@echo ""
+	@cd cluster-service && $(MAKE) local-deploy-provision-shard
+	@echo ""
+	@echo "===================================================================="
+	@echo "Provision shard configuration generated at:"
+	@echo "cluster-service/local/local-provisioning-shards.yml"
+	@echo "===================================================================="
+else
+local-pers-dev-env:
```
machi1990: is this missing https://github.com/Azure/ARO-HCP/blob/deef2f7dda6c8f0c7df78a0aac27129f676743ab/cluster-service/Makefile#L152 ?

PR #3540, opened 2025-12-09
add admin api to dump cosmos data

deads2k:
builds on https://github.com/Azure/ARO-HCP/pull/3536

Adds generic CRUD which allows get and list without inspecting details.  Adds testing for it.  This will be useful later for deletion.

Additionally, adds admin API endpoint that uses the generic CRUD to log to stdout every cosmos record underneath a cluster.  This could be later expanded to be a resource group or a subscription.

Finally, followup work: refactoring parts of the frontend handler chain to provide a mechanism for logging with context in the admin API.

### DIFF
--- a/frontend/pkg/frontend/cluster.go
+++ b/frontend/pkg/frontend/cluster.go
@@ -647,7 +647,7 @@ func (f *Frontend) getInternalClusterFromStorage(ctx context.Context, resourceID
 		return nil, utils.TrackError(err)
 	}
 
-	// Replace the key field from Cosmos with the given resourceID,
+	// Replace the ID field from Cosmos with the given resourceID,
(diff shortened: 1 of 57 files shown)


internal/utils/context.go
```diff
@@ -0,0 +1,77 @@
+package utils
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+
+	"github.com/Azure/ARO-HCP/frontend/pkg/util"
```
stevekuznetsov: maybe pull this one as well so you don't end up adminapi->internal->frontend, I imagine that was not the intent

geoberle (approved):

PR #3541, opened 2025-12-09
frontend: Combine cluster updates with autoscaler updates

mbarnes:
[ARO-17695 - Frontend: Support HCP-Level Autoscaling for All NodePools](https://issues.redhat.com/browse/ARO-17695)

### What

Restoring functionality from commit eb21862fb which had to be temporarily reverted in #3329. See the original commit message for more context about this change.

### Why

Cluster Service semantics have been corrected to better deal with the zero node pools case unique to ARO-HCP, and to more accurately detect changes to the cluster autoscaler parameters.

Leaving this in draft until we have an E2E test that actually induces and verifies autoscaling behavior instead of just tweaking autoscaler parameters and reading them back.


mbarnes:
Taking out of draft with the expectation that @shitaljante will complete E2E for this feature.

PR #3542, opened 2025-12-09
Move resourceID overwrite into the frontend

deads2k:
This avoids trying to track the value to the data storage layer and lets the data storage layer return the exact data that is stored and not manipulate it.  This also makes it easier to add more storage.


mbarnes (changes requested):
A few trivial nits to keep the comment(s) accurate.

PR #3547, opened 2025-12-10
Reduce alert noise and improve signal-to-noise ratio

swiencki:
https://issues.redhat.com/browse/ARO-22283

### What

Addresses top alert volume issues causing non-actionable tickets:

  - KubeDeploymentReplicasMismatch: Add 50% severity threshold and increase
    duration from 15m to 30m to filter single pod failures and transient
    scaling events

  - KubeContainerWaiting: Exclude CrashLoopBackOff (has dedicated alert),
    focus on actionable waiting reasons (ImagePullBackOff, config errors),
    and exclude Job/CronJob workloads that legitimately wait

  - kas-monitor-ErrorBudgetBurn (Variant 1): Increase sustained period
    from 2m to 15m to eliminate tickets for 8-minute transient issues

Enhance the prometheus-rules generator to support proper alert overrides when multiple files define the same rule group. This allows custom alert definitions to replace upstream alerts while preserving all other rules.

Override was created as `make alerts`  pulls from the [upstream prometheus-operator](https://github.com/Azure/ARO-HCP/blob/main/observability/Makefile#L12), wiping out any modifications.

The generator changes are global and affect all alert generated Bicep files, not just the HCP one. The merge logic uses a Go map to store groups, with deterministic ordering.

Files listed in rulesFolders are processed in order. If the same alert group/rule name appears in multiple files, the later file overrides the earlier one with the order-preserving.

In the case of `kubernetesControlPlane-prometheusRule.yaml` and `aro-hcp-kubernetes-overrides-prometheusRule.yaml`, the override is ran last due to the composition of `observability/Makefile`

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/generatedHCPPrometheusAlertingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedHCPPrometheusAlertingRules.bicep
@@ -4,8 +4,8 @@ param azureMonitoring string
 #disable-next-line no-unused-params
 param actionGroups array
 
-resource kubernetesSystem 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-03-01' = {
-  name: 'kubernetes-system'
+resource kubernetesSystemControllerManager 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-03-01' = {
+  name: 'kubernetes-system-controller-manager'
   location: resourceGroup().location
   properties: {
     interval: 'PT1M'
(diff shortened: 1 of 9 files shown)


hawkowl:
@swiencki  I think that downloading the alerts from upstream and merging them with overrides is going to get rapidly confusing and brittle. If we are to mass-import alerts, I think that instead we should import them once, so we can modify them easier (especially descriptions, which we should do custom). If we think upstream k8s might add new alerts that we would find useful as time goes on, perhaps we should have a list of alerts that we've removed for whatever reason, and check that `upstreamIDs = ourIDs U deletedIDs` so that we can check the new alerts and either import them into our config or add them to the deleted IDs list.

hawkowl:
(for alerts for customer control plane resources, I also would prefer we have alerts/SLOs that are all based around our user journeys specifically, rather than generic k8s ones, but that is another discussion :) )

swiencki:
> @swiencki I think that downloading the alerts from upstream and merging them with overrides is going to get rapidly confusing and brittle. If we are to mass-import alerts, I think that instead we should import them once, so we can modify them easier (especially descriptions, which we should do custom). If we think upstream k8s might add new alerts that we would find useful as time goes on, perhaps we should have a list of alerts that we've removed for whatever reason, and check that `upstreamIDs = ourIDs U deletedIDs` so that we can check the new alerts and either import them into our config or add them to the deleted IDs list.

This is a valid point, and I'm unsure of the reasoning of pulling upstream prometheus-operator alerts. If we important once it would reduce some of the complexity and changes to [generator.go](https://github.com/Azure/ARO-HCP/pull/3547/files#diff-a701a2dbd6dac6239cc2d9ae202c856132c7f3364690d512e94bd4d74addbede), however updates from upstream would be manual going forward.

I'll let others comment to get some additional input.

PR #3548, opened 2025-12-10
stop individually patching clusters, nodepools, and externalauth

deads2k:
instead do complete replaces inside of the transaction. This eliminates the indirect setting of fields and ensures that when a user expresses an intent to change a field that change is honored in cosmos instead of losing that information.

Also cleans up some now dead code.


bennerv:
/test e2e-parallel

maestro infra didn't spin up, failed on Eventgrid deployment failure internal server error.

mbarnes (approved):

deads2k:
merged in another PR

PR #3551, opened 2025-12-10
Reduce stg&prod test parallelism even further

mmazur:
Int should be kept higher, so that we don't hide the issues. However failures on stage and prod are toily (they disrupt the release process) and a high percentage of flaky failures increases the risk of someone missing an actual problem during releases. For these reasons, let's keep stage and prod low so we reduce flakes in those environments.

Side note: ci runs against prod will be outliers compared to int and stage because int and stage will always be empty, while prod will be increasingly busy. At some point we'll start hitting scaling failures on prod that do not show up in int/stage. It's worth thinking about how to set up int/stg in such a way that we can simulate the busyness of prod clusters to the highest extent.


mbukatov (approved):
Based on a discussion during today's Quality Status call, I'm ok with the change, since we are looking into the issue and plan to monitor it in parallel.

mmazur:
int and stg has been quite green since the new CS got released. Holding off on this, unless we start hitting timeouts again.

PR #3560, opened 2025-12-10
add data dumps just prior to deletion and for every long running oper…

deads2k:
…ation

We will dump about every five minutes, limited by an LRU cache to prevent explosions in logging.  Once we have a better way to hit the admin API, we can remove the periodic dump.  This will grow in value, the more information we place in cosmos and can be harvested in kusto and must-gather.

### DIFF
--- a/admin/server/handlers/cosmosdump/dump_cosmos.go
+++ b/admin/server/handlers/cosmosdump/dump_cosmos.go
@@ -15,12 +15,11 @@
 package cosmosdump
 
 import (
-	"encoding/json"
-	"errors"
 	"net/http"
 
 	"github.com/Azure/ARO-HCP/internal/database"
 	"github.com/Azure/ARO-HCP/internal/errorutils"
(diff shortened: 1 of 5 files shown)


admin/server/handlers/cosmosdump/dump_cosmos.go
```diff
@@ -38,46 +37,12 @@ func (h *CosmosDumpHandler) ServeHTTP(w http.ResponseWriter, req *http.Request)
...
-		}
-		logger.Info(string(content))
-	}
-	if err := allCosmosRecords.GetError(); err != nil {
-		return utils.TrackError(err)
-	}
-
-	return utils.TrackError(errors.Join(errs...))
-
+	return serverutils.DumpDataToLogger(ctx, h.cosmosClient, resourceID)
```
geoberle: eventually we might want the dump to be returned as payload here. logging and just reporting errors is fine for now

PR #3564, opened 2025-12-10
Return usable error message on attempted creation of a second external auth instance

bennerv:
[ARO-23159](https://issues.redhat.com//browse/ARO-23159)

### What

Return a usable error message when a customer attempts to create a second external auth instance on a cluster.

### Why

cluster service does not support this, so we want to return an error message sooner as well as one that is usable to the customer.  

### TODO:
- [x] : Update e2e tests and mark them in progress until RP is rolled out everywhere.  
- [x] mark tests as informing
    - [ ] Tests should be failing in integration, why is that not the case?  
- [ ] use a database constraint + resourceType (with name & sub) instead
- [ ] CS failing external auth resources counting against the "limit 1"


mbukatov (approved):

geoberle:
can auth providers be in a failed state? if yes, do they count towards that limit?
i guess it depends on how CS is counting here

mbukatov:
Was this addressed in the meantime in a different way? If not, this PR still looks like something we should revisit.

PR #3572, opened 2025-12-11
Use standard CRUD for deletion

deads2k:
This builds on the operation CRUD work to shift deletion to standard CRUD, test that CRUD, and delete nested controllers.

This completes the cleanup necessary to start creating helper/associated resources in cosmos that are cleaned up.  Looks like the backend is missing integration level testing (I updated all the mocks).  But we'll want that integration level testing before we go merging many backend controllers.

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -517,7 +517,7 @@ func (f *Frontend) ArmSubscriptionPut(writer http.ResponseWriter, request *http.
 
 	// Clean up resources if subscription is deleted.
 	if subscription.State == arm.SubscriptionStateDeleted {
-		if err := f.DeleteAllResources(ctx, writer, nil, subscriptionID); err != nil {
+		if err := f.DeleteAllResourcesInSubscription(ctx, subscriptionID); err != nil {
 			return utils.TrackError(err)
 		}
 	}
--- a/frontend/pkg/frontend/helpers.go
+++ b/frontend/pkg/frontend/helpers.go
@@ -139,7 +139,7 @@ func (f *Frontend) CheckForProvisioningStateConflict(ctx context.Context, operat
 	return checkForProvisioningStateConflict(ctx, f.dbClient, operationRequest, doc.ResourceID, doc.ProvisioningState)
 }
 
-func (f *Frontend) DeleteAllResources(ctx context.Context, writer http.ResponseWriter, request *http.Request, subscriptionID string) error {
+func (f *Frontend) DeleteAllResourcesInSubscription(ctx context.Context, subscriptionID string) error {
 	transaction := f.dbClient.NewTransaction(subscriptionID)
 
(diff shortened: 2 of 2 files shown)


frontend/pkg/frontend/cluster.go
```diff
@@ -605,6 +608,142 @@ func (f *Frontend) updateHCPClusterInCosmos(ctx context.Context, writer http.Res
...
+	err = f.clusterServiceClient.DeleteCluster(ctx, cluster.ServiceProviderProperties.ClusterServiceID)
+	var ocmError *ocmerrors.Error
+	if errors.As(err, &ocmError) && ocmError.Status() == http.StatusNotFound {
+		// StatusNotFound means we have stale data in Cosmos DB.
+		// This can happen in test environments if a user bypasses
+		// the RP to delete a resource (e.g. "ocm delete"). It can
+		// also happen if an asynchronous deletion operation fails.
+		// we will fall through and cancel all operations and go through as normal a deletion flow as we can to avoid
+		// leaking data related to the resource, like controller status.
+		logger.Info("clusterService nodepool missing, trying to clean up", "err", err)
```
mbarnes:
```suggestion
		logger.Info("clusterService cluster missing, trying to clean up", "err", err)
```

frontend/pkg/frontend/external_auth.go
```diff
@@ -540,6 +543,120 @@ func (f *Frontend) updateExternalAuthInCosmos(ctx context.Context, writer http.R
...
+		database.OperationRequestDelete,
+		externalAuth.ID,
+		externalAuth.ServiceProviderProperties.ClusterServiceID,
+		"",
+		"",
+		"",
+		correlationData)
+	if request != nil {
+		// these are optional because when this is triggered via the subscription deletion flow, there is no notification URI
+		// so these operations cannot be directly tracked.
```
mbarnes:
```suggestion
		// these are optional because when this is triggered via the subscription deletion flow, there is no
		// deletion request containing these headers so these operations cannot be directly tracked.
```

frontend/pkg/frontend/frontend.go
```diff
@@ -582,7 +517,7 @@ func (f *Frontend) ArmSubscriptionPut(writer http.ResponseWriter, request *http.
 
 	// Clean up resources if subscription is deleted.
 	if subscription.State == arm.SubscriptionStateDeleted {
-		if err := f.DeleteAllResources(ctx, subscriptionID); err != nil {
+		if err := f.DeleteAllResources(ctx, writer, request, subscriptionID); err != nil {
```
mbarnes: I think we can pass a `nil` request here, right?  (Or just the request parameter from `DeleteAllResources`.) That will prevent `Location` and  `Azure-AsyncOperation` headers from being added to the subscription PUT response.
```suggestion
		if err := f.DeleteAllResources(ctx, writer, nil, subscriptionID); err != nil {
```

PR #3574, opened 2025-12-11
Add example do nothing controller

deads2k:
We'll actually merge this when ready so we can refine our example and we'll always have a copy/pasteable example showing our current best practices when writing another controller.


backend/controllers/do_nothing.go
```diff
@@ -0,0 +1,228 @@
...
+			"provisioning_state", cosmosHCPCluster.ServiceProviderProperties.ProvisioningState,
+		)
+	} else {
+		logger.Info("starting work for item",
+			"provisioning_state", cosmosHCPCluster.ServiceProviderProperties.ProvisioningState,
+			"controller_degraded", getCondition(existingController.Status.Conditions, "Degraded"),
+		)
+	}
+
+	// Do your work here.  If anything fails, return an error to have it recorded.
```
stevekuznetsov: Somewhere, explain that this controller enforces the invariant that every AROHCPCluster has an AzureThing, to orient people for what they're about to read.

backend/controllers/do_nothing.go
```diff
@@ -0,0 +1,228 @@
...
+	allSubscriptions := c.cosmosClient.ListAllSubscriptionDocs()
+	for subscriptionID := range allSubscriptions.Items(ctx) {
+		allHCPClusters, err := c.cosmosClient.HCPClusters(subscriptionID, "").List(ctx, nil)
+		if err != nil {
+			logger.Error("unable to list HCP clusters", "error", err, "subscription_id", subscriptionID)
+			continue
+		}
+
+		for _, hcpCluster := range allHCPClusters.Items(ctx) {
+			c.queue.Add(HCPClusterKey{
```
stevekuznetsov: Only thing missing from this example is some secondary/derivative state feeding back parent object's key for re-queue when noticing diffs / using changefeed on another store - do we have any such flows?

backend/controllers/util.go
```diff
@@ -0,0 +1,186 @@
...
+		// TODO we'd prefer a DeepCopy, but we don't have one yet.
+		temp := *existingController
+		desiredController = &temp
+		desiredController.Status.Conditions = slices.Clone(existingController.Status.Conditions)
+	}
+	for _, mutationFn := range mutationFns {
+		mutationFn(desiredController)
+	}
+
+	if equality.Semantic.DeepEqual(existingController, desiredController) {
```
stevekuznetsov: We need struct tags for semantic equality, don't we?

test-integration/utils/controllertesthelpers/basic_controller.go
```diff
@@ -0,0 +1,106 @@
...
+
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// this forces every test to have its own directory and a couple permanent fixtures
+	testDir, err := fs.Sub(tc.ArtifactDir, tc.Name)
+	require.NoError(t, err)
+
+	logger := utils.LoggerFromContext(ctx)
```
stevekuznetsov: Who will put it in there? Consider `testr.New(t)`

backend/controllers/do_nothing.go
deads2k: > Only thing missing from this example is some secondary/derivative state feeding back parent object's key for re-queue when noticing diffs / using changefeed on another store - do we have any such flows?

possibly. Changefeeds don't have client support (yet).  Perhaps when we find one.

test-integration/utils/controllertesthelpers/basic_controller.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (tc *BasicControllerTest) RunTest(t *testing.T) {
+	integrationutils.SkipIfNotSimulationTesting(t)
+
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// this forces every test to have its own directory and a couple permanent fixtures
+	testDir, err := fs.Sub(tc.ArtifactDir, tc.Name)
```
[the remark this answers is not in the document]
deads2k: > Check the `golden` implementation for some string mangling that ensures `tc.Name` is safe to use as an inode name

Eh, they get what they deserve.

PR #3566, opened 2025-12-11
switch to using new Operation CRUD

deads2k:
This brings more CRUD inline with out standard storage.

PR #3583, opened 2025-12-12
provide a notification URI for deletion when triggered via direct delete

deads2k:
reapplies https://github.com/Azure/ARO-HCP/pull/3566, then adds a commit to correct the failure.    
    
>        failed to cleanup resource group: at least one hcp cluster failed to delete: response is missing polling URL",
 
When used in a standard delete request (not a subscription delete) we need to include the notification URI in the operation so that the request is well formed


deads2k:
provisioning failure that is unlikely to be this PR

/retest

PR #3592, opened 2025-12-14
ARO-22851 parametrize ocp versions

mbukatov:
[ARO-22851](https://issues.redhat.com//browse/ARO-22851)

### What

- removing OpenShift version IDs from the test code
- provide a way to redefine it from the outside
- update the default to 4.20 (for testing purposes)

### Why

We want to be able to redefine default OpenShift version we test with easily.

WIP

### DIFF
--- a/test/e2e-setup/bicep/cluster-only.bicep
+++ b/test/e2e-setup/bicep/cluster-only.bicep
@@ -13,7 +13,7 @@ param identities object
 param usePooledIdentities bool = false
 
 @description('ControlPlane OpenShift Version ID')
-param openshiftControlPlaneVersionId string = '4.19'
+param openshiftControlPlaneVersionId string = '4.20'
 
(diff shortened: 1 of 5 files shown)


test/e2e-setup/bicep/cluster-only.bicep
```diff
@@ -4,6 +4,9 @@ param persistTagValue bool = false
 @description('Name of the hypershift cluster')
 param clusterName string
 
+@description('ControlPlane OpenShift Version ID')
+param openshiftControlPlaneVersionId string = '4.19'
```
JameelB: what's the reason for keeping this to 4.19? Is it because we haven't changed the default version yet? 🤔 same with the node pool version below https://github.com/Azure/ARO-HCP/blob/296c62e356645c7670338d0f8b10c0cada813450/test/e2e-setup/bicep/demo.bicep#L11

PR #3600, opened 2025-12-15
Move NonLocal E2E to Test Dir

SudoBrendan:
Jira N/A - artifact of the new Prow process

### What

Move E2E config into ./test dir

### Why

We shouldn't configure E2E test runs with a top-level file; this requires approval from the top-level OWNERS.


bennerv:
There's a diff in the file you're moving that it doesn't like.  
```
diff --git a/test/nonlocal-e2e-specs.txt b/test/nonlocal-e2e-specs.txt
index da6ab8af..6aafd5cd 100644
--- a/test/nonlocal-e2e-specs.txt
+++ b/test/nonlocal-e2e-specs.txt
@@ -19,4 +19,3 @@
   "should create an HCP cluster and validate TLS certificates",
   "should not be able to create a 4.18 HCP cluster"
 ]
-

SudoBrendan:
> I care quite a lot about which tests are allowed to add to the list.

@stevekuznetsov, in your view, what are the biggest risks in giving up control on this file? How could we gain confidence in extending trust to the test OWNERS so they have the same respect for its contents as you do?

PR #3623, opened 2025-12-16
e2e: add SelfSubjectReview request on admin_credential_lifecycle 

bennerv:
[ARO-21682](https://issues.redhat.com//browse/ARO-21682)

I don't expect this to fix the bug, but it will more accurately test the admin credential we're requesting.  Additionally, on revocation we are performing a SelfSubjectReview rather than using the standard verifiers and expecting them to fail.  We're explicitly looking for an unauthorized error from the kube-apiserver.  

Has some smaller refactors of other things as well.


test/e2e/admin_credential_lifecycle.go
```diff
@@ -169,7 +174,7 @@ var _ = Describe("Customer", func() {
 				credentials = append(credentials, adminRESTConfig)
 
 				By(fmt.Sprintf("validating admin credential %d works", i+1))
-				Expect(verifiers.VerifyHCPCluster(ctx, adminRESTConfig)).To(Succeed())
+				Expect(verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyBreakglassAdminAccess())).To(Succeed())
```
stevekuznetsov: Why add a verifier, will anyone else care to run this? IMO inlining it here makes more sense

PR #3628, opened 2025-12-17
refactor: refactor helpers for admit cluster tests

ahitacat:
### What
Refactor the helper method used in admit_cluster_test, now this test will use the methods in testhelpers.

### Why

Follow up the comment https://github.com/Azure/ARO-HCP/pull/3140#discussion_r2585084939

### DIFF
--- a/internal/api/testhelpers.go
+++ b/internal/api/testhelpers.go
@@ -19,9 +19,12 @@ import (
 	"log/slog"
 	"path"
 	"testing"
+	"time"
 
 	"dario.cat/mergo"
 	"github.com/stretchr/testify/require"
+
+	"k8s.io/utils/ptr"
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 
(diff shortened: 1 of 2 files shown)


internal/api/testhelpers.go
```diff
@@ -140,3 +143,30 @@ func Must[T any](v T, err error) T {
...
+// Helper function to create a subscription without AllowDevNonStableChannels feature flag
+func CreateStandardSubscription() *arm.Subscription {
+	return &arm.Subscription{
+		State:            arm.SubscriptionStateRegistered,
+		RegistrationDate: ptr.To(time.Now().Format(time.RFC1123)),
+		Properties: &arm.SubscriptionProperties{
+			RegisteredFeatures: &[]arm.Feature{},
+		},
+	}
+}
```
mbarnes: How about if we combine these and take registered features as a variadic argument list?
```
func CreateTestSubscription(registeredFeatures ...string)
```
That will help future-proof it if we add more subscription features.

mbarnes (approved):

PR #3646, opened 2025-12-17
Add owning team tag value to aks cluster

abiduke612:
https://issues.redhat.com/browse/ARO-22825

### What

Add tag value to aks cluster to indicate owning team

### Why

Tag on aks cluster will be used to identify clusters within AKS ManagedClusterSnapshot table in order to create alert rules on clusters

### DIFF
--- a/config/dev.digests.yaml
+++ b/config/dev.digests.yaml
@@ -3,22 +3,22 @@ clouds:
     environments:
       cspr:
         regions:
-          westus3: 253ece6f973334508bfc97a566b161f93e16686a80856fac229a1227ef3edb17
+          westus3: b52a5cad286b8ae66a25136e4b7c165c5c2bd60efb7d36d777d0c9d3c22845c8
       dev:
         regions:
-          westus3: 12cfdff07c02cd70138251b597f869e631244d7523d3c6254e2c92ff808b4db2
+          westus3: dc0c24be1e9f4d87a45d626a98da9bf16c96cf7473bb49c1185a8dae5afbb47e
       ntly:
         regions:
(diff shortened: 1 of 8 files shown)


config/config.schema.json
```diff
@@ -2316,6 +2316,9 @@
           "pattern": "^[a-z0-9][a-z0-9-]*[a-z0-9]$",
           "description": "The workspace name should include 4-63 letters, digits or '-'. The '-' shouldn't be the first or the last symbol. We limit the name to 44 characters so we have a future option to use NSP: https://learn.microsoft.com/en-us/azure/private-link/network-security-perimeter-concepts"
         },
+        "alertRuleOwningTeamTag": {
+          "type": "string"
+        },
```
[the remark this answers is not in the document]
stevekuznetsov: No, the one in sdp-pipelines just flows from here

dev-infrastructure/modules/aks-cluster-base.bicep
```diff
@@ -294,6 +294,7 @@ resource aksCluster 'Microsoft.ContainerService/managedClusters@2024-10-01' = {
   tags: {
     persist: 'true'
     clusterType: clusterType
+    owningTeam: owningTeamTagValue
```
stevekuznetsov: Will this leave a literal tag called `owningTeam`? Maybe add some namespace/prefix like `aro.azure.us/owner` to make it very unlikely anyone re-uses it.

abiduke612: The tag would be something like 'owningTeam: "aro-msft"', so the actual value of the tag would hopefully be unique to us, and I would check that the value of owningTeam is "aro-msft", not just that the cluster has an "owningTeam" tag. But I can definitely change the tag to be something more specific if that makes more sense to you

PR #3650, opened 2025-12-18
test: allow re-trying APIService verifier for some time

stevekuznetsov:
HyperShift currently doesn't wait for aggregated API servers to be ready before determining that a HostedCluster is finshed provisioning. They have committed to fixing this soon, but we're flaking on our cluster verifiers in the interim. This fix hopefully ameliorates our flakes while not papering over this issue forever.

### DIFF
--- a/test/util/verifiers/apiservice.go
+++ b/test/util/verifiers/apiservice.go
@@ -19,8 +19,12 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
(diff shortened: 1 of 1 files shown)


test/util/verifiers/apiservice.go
```diff
@@ -39,43 +41,67 @@ func (v verifyAllAPIServicesAvailableImpl) Verify(ctx context.Context, adminREST
...
+				if availableCondition == nil {
+					failureReasonErrs = append(failureReasonErrs, fmt.Errorf("apiservice/%s is not available because it has no status", unavailableAPIService.Name))
+				} else {
+					failureReasonErrs = append(failureReasonErrs, fmt.Errorf("apiservice/%s is not available because: %v", unavailableAPIService.Name, availableCondition.Message))
+				}
 			}
+			return false, errors.Join(failureReasonErrs...)
+		}
+		if !foundAPIServices.HasAll(requiredAPIServices.UnsortedList()...) {
+			return false, fmt.Errorf("required apiservices are missing: %v", strings.Join(requiredAPIServices.Difference(foundAPIServices).SortedList(), ", "))
```
deads2k: pretty sure that any error causes the poll to not retry.

PR #3651, opened 2025-12-18
rekey all cosmos data from .id=UUID to .id=resourceID

deads2k:
This change performs a data migration on read to switch from reading based on a cosmos search of fields matching a pattern to direct item reads based on key.  On a failure to lookup by cosmosID and success reading based on existing search, we create a new item and delete the original.
    
Since old frontends use the cosmosUID as fully opaque, the new ID doesn't cause problems for old frontends even on rollback.
    
When the frontend starts, we find all subscriptions, then search for all resources in each subscription using the untyped client listRecursive, then read each item to trigger the migration.

If the create succeeds and the delete fails, then resulting GETs will produce an ambiguous result error (a pre-existing handled error, so old frontends are ok) and we'll have to resolve the conflict manually.

Still needs more test coverage for types, but since we merged https://github.com/Azure/ARO-HCP/pull/3662 we can see this works properly for clusters and operations.  Need to add coverage (in a separate PR first to prove it works) for nodepools and externalauths.

### DIFF
--- a/test-integration/utils/databasemutationhelpers/resource_crud_test_util.go
+++ b/test-integration/utils/databasemutationhelpers/resource_crud_test_util.go
@@ -19,13 +19,17 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/fs"
+	"net/http"
 	"sort"
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/neilotoole/slogt"
 	"github.com/stretchr/testify/require"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
 
@@ -106,6 +110,17 @@ func (tt *ResourceMutationTest) RunTest(t *testing.T) {
 	cleanupCtx = utils.ContextWithLogger(cleanupCtx, slogt.New(t, slogt.JSON()))
 	defer testInfo.Cleanup(cleanupCtx)
 	go frontend.Run(ctx, ctx.Done())
+
+	// wait for migration to complete to eliminate races with our test's second call migrateCosmos and to ensure the server is ready for testing
+	err = wait.PollUntilContextCancel(ctx, 1*time.Second, true, func(ctx context.Context) (bool, error) {
+		_, err := http.Get(testInfo.FrontendURL)
+		if err != nil {
+			t.Log(err)
+			return false, nil
+		}
+		return true, nil
+	})
+	require.NoError(t, err)
 
 	// create anything and round trip anything for cluster-service
 	err = integrationutils.TrivialPassThroughClusterServiceMock(t, testInfo, nil)


deads2k:
ha!  it worked!

frontend/pkg/frontend/cluster.go
```diff
@@ -671,6 +671,7 @@ func (f *Frontend) DeleteCluster(writer http.ResponseWriter, request *http.Reque
 	}
 	_, err = transaction.Execute(ctx, nil)
 	if err != nil {
+		logger.Error("failed executing transaction", "transaction", transaction)
```
bennerv: is it useful to emit the error here in the log as well, or we'll have it tracked as part of `utils.TrackError()`

frontend/pkg/frontend/frontend.go
```diff
@@ -157,6 +163,88 @@ func (f *Frontend) Run(ctx context.Context, stop <-chan struct{}) {
...
+
+	subscriptionIterator, err := cosmosClient.Subscriptions().List(ctx, nil)
+	if err != nil {
+		panic(err)
+	}
+	for _, subscription := range subscriptionIterator.Items(ctx) {
+		if _, err := cosmosClient.Subscriptions().Get(ctx, subscription.ResourceID.Name); err != nil {
+			panic(err)
+		}
+	}
```
bennerv: What's the purpose of listing the subscriptions then getting the individual fields?  afaict the `id` in subscription doc is already the subscription, but also, we're not doing anything here besides listing all subs, then getting the subs individually by the subscription.

bennerv: Also, why do we use `subscription.ResourceID.Name` over the `subscription.ResourceID.SubscriptionID` for keying into cosmosdb?  Fundamentally there's no difference, but wondering if there's a specific reason.

test-integration/utils/databasemutationhelpers/resource_crud_test_util.go
```diff
@@ -193,6 +208,9 @@ func NewStep[InternalAPIType any](indexString, stepType, stepName string, testDi
 	case "completeOperation":
 		return newCompleteOperationStep(stepID, stepDir)
 
+	case "migrateCosmos":
```
bennerv: If I understand correctly, this keys off of the directory name (05-**migrateCosmos**-migration) and then will perform the database migration of cosmosdb by calling the frontend function. Then we use the cosmosCompare step/directory to compare the contents of the docs in cosmos match the rekeying.  

Makes sense, more a mental note to myself^

bennerv (commented):
/hold
/lgtm

Took a few passes and it makes sense to me.  Had a couple open questions, but feel free to remove the hold so it merges.

internal/api/arm/subscription.go
```diff
@@ -53,20 +53,28 @@ type Subscription struct {
 	// LastUpdated is a copy of the Cosmos DB system generated
 	// "_ts" last updated timestamp field for metrics reporting.
 	LastUpdated int `json:"-"`
+
+	// CosmosUID is used to keep track of whether we have transitioned to a new cosmosUID scheme for this item
+	CosmosUID string `json:"-"`
```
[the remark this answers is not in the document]
deads2k: > why do we not marshal this, but do so for operation types and other types?

I think you're right.  Sounds like via slack I can take this an IOU to clear the stack and deliver it.

PR #3654, opened 2025-12-18
Only run authorized cidrs test in integration env since it hasn't landed in production yet

nimrodshn:
### What

The [code to support](https://github.com/Azure/ARO-HCP/pull/2138) this feature has not yet landed in staging and production envs and thus we would like to run this test exclusively in integration.

Merging this will allow our CI in staging and production to be green and thus allow for smooth production rollout which should start today (cc @mmazur ). 

Once this is merged we expect rollout to prod to start right away and than we will be able to remove this label.

The code to support this feature is in the frontend component.


deads2k:
/hold 

needs a comment in the code about why this can only run in integration.

PR #3658, opened 2025-12-18
e2e: loop until context timeout rather than if we encounter an error

bennerv:
Continue to loop even if we encounter an error during e2e admin test requesting a SSR.  It will log out the error already if it differs from last but we expect eventual consistency until OCPBUGS-62177 is implemented.


stevekuznetsov (approved):
/lgtm
/approve

PR #3660, opened 2025-12-18
move to a single integration suite for frontend, backend, and admin API

deads2k:
mostly a straight move and some bash tidying to get our integration testing aligned.


stevekuznetsov (approved):
/lgtm
/approve

PR #3662, opened 2025-12-18
Expand the step-style integration tests to make http frontend testing easier

deads2k:
This adds the ability to easily test the cosmos DB entries that result from frontend requests.  It is a necessary prereq to the data migration in https://github.com/Azure/ARO-HCP/pull/3651 where the records in cosmos will change.


mbarnes (approved):
There's a lot here and I think I'm gonna have to play with this to fully grok it.  Fortunately I've got a couple bugs on my plate where I think I can leverage this test framework you've built.  I didn't go through 100% of the code in detail, but I think I get the gist of it and it looks good to me.

/lgtm
/approve

PR #3688, opened 2025-12-24
increasing max user nodepool count

sclarkso:
### What

increases the max size of the user nodepool

### Why

kube-apiserver replicasets were timing out on e2e tests and the nodepool events had errors claiming no nodes were available, and they could not be scaled because maximum number had been reached.


sclarkso:
@raelga please review

PR #3699, opened 2026-01-05
dev/enable-identity-reuse

roivaz:
https://issues.redhat.com/browse/AROSLSRE-305

* Some tests are missing identity assignment for the identity pool. Add it.
* Adjustments to how the identity pool resource groups are created.

### DIFF
--- a/test/e2e/cluster_autoscaling.go
+++ b/test/e2e/cluster_autoscaling.go
@@ -55,6 +55,11 @@ var _ = Describe("Customer", func() {
 			)
 			tc := framework.NewTestContext()
 
+			if tc.UsePooledIdentities() {
+				err := tc.AssignIdentityContainers(ctx, 1, 60*time.Second)
+				Expect(err).NotTo(HaveOccurred())
+			}
+
 			By("creating a resource group")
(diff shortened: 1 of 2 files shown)


test/hack/apply-msi-pool.sh
```diff
@@ -2,6 +2,6 @@ az stack sub create \
     --name aro-hcp-msi-pool \
     --location westus3 \
     --template-file test/e2e-setup/bicep/msi-pools.bicep \
-    --parameters poolSize=20 \
+    --parameters poolSize=120 resourceGroupBaseName=aro-hcp-test-msi-containers-dev \
```
janboll: This syntax looks odd, shouldn't it be:
```
    --parameters poolSize=120
   --parameters resourceGroupBaseName=aro-hcp-test-msi-containers-dev 
```
https://learn.microsoft.com/en-us/cli/azure/stack/sub?view=azure-cli-latest#az-stack-sub-create

janboll (changes requested):

test/hack/apply-msi-pool.sh
roivaz: Oh you are right, I'll fix that in the next iteration.

PR #3702, opened 2026-01-05
Fix Kusto output

janboll:
[AROSLSRE-245](https://issues.redhat.com/browse/AROSLSRE-245)

### What

Use arobit toggle and not manage instance toggle for output.

### Why

This fixes the conditional kusto deployment. The output should only expect the kusto to be there if we actually want to use it (enable arobit output). The manageinstance flag is there to check if we actually want to deploy kusto.

### DIFF
--- a/dev-infrastructure/configurations/kusto-lookup.tmpl.bicepparam
+++ b/dev-infrastructure/configurations/kusto-lookup.tmpl.bicepparam
@@ -2,4 +2,4 @@ using '../templates/kusto-lookup.bicep'
 
 param kustoName = '{{ .kusto.kustoName }}'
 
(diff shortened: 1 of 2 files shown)


dev-infrastructure/configurations/kusto-lookup.tmpl.bicepparam
```diff
@@ -2,4 +2,4 @@ using '../templates/kusto-lookup.bicep'
 
 param kustoName = '{{ .kusto.kustoName }}'
 
-param manageInstance = {{ .kusto.manageInstance }}
+param kustoEnabled = {{ .arobit.kusto.enabled }}
```
janboll: this is the actual fix, rest is just renaming parameter for clarity

PR #3713, opened 2026-01-06
remove dead functions to make adding database type easier

deads2k:
Showed up when adding `ServiceProviderCluster` for controller coordination without conflicts with user writes.


mbarnes (approved):
/lgtm
/approve

PR #3716, opened 2026-01-06
start controller for subscription cache and prove it works

deads2k:
prove involves using the subscription cache to power a replacement for the cluster creation operation.

/hold

holding I haven't tried this yet.


deads2k:
role assignments

/retest

miguelsorianod:
> Decently significance: https://search.dptools.openshift.org/?search=408+Request+timed+out&maxAge=336h&context=1&type=junit&name=ARO-HCP&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job
> 
> perhaps we need to reduce parallelism

https://learn.microsoft.com/en-us/azure/cosmos-db/emulator?source=recommendations#differences-between-the-emulator-and-cloud-service here differences of the emulator compared to the real service are explained. The following seem relevant:
* The emulator ideally supports up to 10 fixed-size containers at 400 RU/s or 5 unlimited-size containers. Theoretically, you can create more containers, but you could experience performance degradation with the emulator.
* The emulator supports a maximum of five JOIN statements per query.
* The emulator constraints the unique identifier of items to a size of 254 characters.
* The emulator only supports the [Session](https://learn.microsoft.com/en-us/azure/cosmos-db/consistency-levels#session-consistency) and [Strong](https://learn.microsoft.com/en-us/azure/cosmos-db/consistency-levels#strong-consistency) consistency levels. The emulator isn't a scalable service and doesn't actually implement the consistency levels. The emulator only flags the configured consistency level for testing purposes.

backend/controllers/do_nothing.go
```diff
@@ -25,14 +25,16 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/workqueue"
 
+	"github.com/Azure/ARO-HCP/backend/listers"
```
miguelsorianod: can we add the listers in `pkg` ? In the other MR about inflight we are creating some packages within `pkg`

PR #3721, opened 2026-01-07
set E2E default NodePoolVersionId to 4.20.8

mbukatov:
[ARO-23277](https://issues.redhat.com//browse/ARO-23277) Update E2E node pool tests to use 4.20.8

### What

Update E2E test's default NodePoolVersionId to 4.20.8.

### Why

We want to run tests against 4.20.8 by default.

### DIFF
--- a/test/e2e-setup/bicep/cluster-nodepool-osdisk.bicep
+++ b/test/e2e-setup/bicep/cluster-nodepool-osdisk.bicep
@@ -16,7 +16,7 @@ param usePooledIdentities bool = false
 param openshiftControlPlaneVersionId string = '4.20'
 
 @description('NodePool OpenShift Version ID')
-param openshiftNodePoolVersionId string = '4.20.5'
+param openshiftNodePoolVersionId string = '4.20.8'
 
 @description('Node pool osDisk Size in GiB')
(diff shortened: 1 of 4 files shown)


raelga:
INT 

```
❯ az rest --method GET --uri "/subscriptions/64f0....INT....e54/providers/Microsoft.RedHatOpenShift/locations/uksouth/hcpOpenShiftVersions?api-version=2024-06-10-preview" | tail -n10
      "id": "/subscriptions/64f0....INT....e54/providers/Microsoft.RedHatOpenShift/locations/uksouth/hcpOpenShiftVersions/4.20.8",
      "name": "4.20.8",
      "properties": {
        "channelGroup": "stable",
        "enabled": true
      },
      "type": "Microsoft.RedHatOpenShift/locations/hcpOpenShiftVersions"
    }
  ]
}
```

deads2k:
/hold

We can't just drop testing for a version we support installing.  We must at least cover installation of a 4.y, with nodes, and functional cluster-ingress.  We already have a working set of artifacts to do this today, so simply copying them and using them in a separate test is enough to satisfy the need.

deads2k:
/hold cancel

This was apparently broken last year. @davidffrench promised to get the tests created starting next Monday.

PR #3724, opened 2026-01-07
serialize operations with resourceID

deads2k:
This builds on previous integration tests to ensure we keep functioning. It adds serialziation for resourceID in operations so that the standard "get" path will work for them. It also makes the OperationStatus endpoint atttempt to read both eventual new cosmosIDs and current cosmosIDs.

This also includes removing the azure global location because unit tests failed when that wasn't set because the resourceID was no longer proper and valid.

### DIFF
--- a/internal/database/convert_operation.go
+++ b/internal/database/convert_operation.go
@@ -63,6 +63,10 @@ func CosmosToInternalOperation(cosmosObj *Operation) (*api.Operation, error) {
 	internalObj := &tempInternalAPI
 
 	// old records don't serialize this, but we want all readers to be able to depend on it. We can derive it from the operationID
+	// this ID does not include the location because doing so changes the resulting azcorearm.ParseResourceID().ResourceType to be
+	// Microsoft.RedHatOpenShift/locations/hcpOperationStatuses.  This type is not compatible with the current cosmos storage and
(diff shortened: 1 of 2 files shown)


internal/database/convert_operation.go
```diff
@@ -79,5 +79,14 @@ func CosmosToInternalOperation(cosmosObj *Operation) (*api.Operation, error) {
 		}
 	}
 
+	// old records don't serialize this, but we want all readers to be able to depend on it. We can derive it from the operationID
+	if internalObj.ResourceID == nil {
+		internalObj.ResourceID = api.Must(azcorearm.ParseResourceID(path.Join("/",
+			"subscriptions", internalObj.ExternalID.SubscriptionID,
+			"providers", api.ProviderNamespace,
+			api.OperationStatusResourceTypeName, internalObj.OperationID.Name,
```
mbarnes: This is missing a `locations` segment.  Is that intentional?

The actual status endpoint follows the pattern:
```
/subscriptions/${subscriptionId}/providers/Microsoft.RedHatOpenShift/locations/${azureRegion}/hcpOperationStatuses/${operationId}
```

mbarnes (changes requested):
Question below about the format of `ResourceID`.  I forget if you covered this in our call yesterday.  Request changes to either fix the omission or comment why it's omitted.

PR #3727, opened 2026-01-07
frontend: Fix parent resource conflict detection

mbarnes:
### What

Regression slipped through where node pool and external auth requests were not being rejected with 409 Conflict when the parent resource is in a "Provisioning" or "Deleting" state.

This includes a number of new integration test cases to ensure this doesn't regress again.

### DIFF
--- a/test-integration/frontend/artifacts/NodePoolMutation/cluster-creating/expected-errors.txt
+++ b/test-integration/frontend/artifacts/NodePoolMutation/cluster-creating/expected-errors.txt
@@ -1,0 +1,1 @@
+Conflict: /subscriptions/0465bc32-c654-41b8-8d87-9815d7abe8f6/resourceGroups/some-resource-group/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/cluster-creating/nodePools/np1: Cannot create resource while parent resource is provisioning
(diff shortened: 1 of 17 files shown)


test-integration/frontend/artifacts/NodePoolMutation/cluster-creating/expected-errors.txt
```diff
@@ -0,0 +1 @@
+Conflict: /subscriptions/0465bc32-c654-41b8-8d87-9815d7abe8f6/resourceGroups/some-resource-group/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/cluster-creating/nodePools/np1: Cannot create resource while parent resource is provisioning
```
deads2k: this seems weird to me.  Is it an ARM requirement that children cannot be created during provisioning?  why shouldn't a user be able to add nodepools?

test-integration/frontend/artifacts/NodePoolMutation/cluster-deleting/expected-errors.txt
```diff
@@ -0,0 +1 @@
+Conflict: /subscriptions/0465bc32-c654-41b8-8d87-9815d7abe8f6/resourceGroups/some-resource-group/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/cluster-deleting/nodePools/np1: Cannot create resource while parent resource is deleting
```
deads2k: this one makes sense to me.

deads2k:
alright, let's keep cluster-service working for now, but plan on revisiting once we migrate off cluster-service

/lgtm
/approve

PR #3729, opened 2026-01-07
Make the APIRegistry new external functions roundtrippable.

deads2k:
Previously, passing a nil for creating a new external instance got one that alreayd had defaulted values. this was surprising for a couple reasons, primarily because the output doesn't roundtrip and compare to the input, but also because nil and empty were treated differently. This also meant that creating a new instance using "nil" resulted in a value that would have latent, unexpected values after writing or deserializing into it.  While this may be done intentionally, it's a case where a caller can explicitly make that choice.

This change only impacted three callsites associated with preflights becuase the previous refactors have eliminated all other usages.


raelga:
cc @mbarnes @geoberle PTAL

PR #3730, opened 2026-01-08
feat: add support for nodepool changes

ashishmax31:
Changes to support provisioning a new nodepool in an env. 
The idea is to only override the poolcount(in sdp-pipelines) to a non zero value in stage so that we can deploy a new nodepool in stage and leave all other env's untouched.

### DIFF
--- a/config/dev.digests.yaml
+++ b/config/dev.digests.yaml
@@ -3,19 +3,19 @@ clouds:
     environments:
       cspr:
         regions:
-          westus3: a2d622596b905c29652d1da9bde9c430b1a8fb7e908b979440c931badbfcb973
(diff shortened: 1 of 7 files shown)


config/config.yaml
```diff
@@ -440,6 +440,14 @@ defaults:
         poolCount: 2
         zones: ""
         zoneRedundantMode: "Auto"
+      newInfraAgentPool:
+        minCount: 1
+        maxCount: 3
+        vmSize: 'Standard_D4ds_v5'
```
ashishmax31: We can decide the shape of the node. Just kept the existing shape for now.

raelga:
@geoberle Please take a look to the new implementation, is not the best one sadly. We will work on that once we have capacity to do it. 😿

ashishmax31:
Looks good to me @raelga

PR #3733, opened 2026-01-08
Kusto autoscaling

janboll:
### What

Enable the usage of autoscaling for the kusto instances. Set conservative default for dev.

### Why

Autoscaling is configured for prod instances right now. We want to make sure to not run out of capacity for ingestion/query in Prod, and thus enable scaling configuration, as it is configured in classic prod.


venkateshsredhat (approved):
/lgtm

PR #3735, opened 2026-01-08
showcase some cluster inflight validation areas by taking one as an example

miguelsorianod:
To start figuring out how do we want to introduce and organize cluster inflights I open this MR that showcases some (but not all) of the areas and concerns that need to be solved to be able to introduce cluster inflights.

Specifically, it showcases introduction of one of the cluster inflights that CS runs where it checks that the customer's azure subscription does have the set of resource providers registered that are needed so the cluster creation process works correctly.

This MR builds on top of https://github.com/Azure/ARO-HCP/pull/3718.

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -162,11 +162,9 @@ func loadAzureRuntimeConfig(path string) (configdto.AzureRuntimeConfig, error) {
 	return config, nil
 }
 
-func buildAzureConfig(azurRuntimeConfigDTO configdto.AzureRuntimeConfig, tracerProvider trace.TracerProvider) (azureconfig.AzureConfig, error) {
-	builder := config.NewAzureCloudEnvironmentBuilder()
-	builder.CloudEnvironment(azurRuntimeConfigDTO.CloudEnvironment.String())
-	builder.TracerProvider(tracerProvider)
-	cloudEnvironment, err := builder.Build()
+func buildAzureConfig(azureRuntimeConfigDTO configdto.AzureRuntimeConfig, tracerProvider trace.TracerProvider) (azureconfig.AzureConfig, error) {
+
+	cloudEnvironment, err := config.NewAzureCloudEnvironment(azureRuntimeConfigDTO.CloudEnvironment.String(), tracerProvider)
 	if err != nil {
 		return azureconfig.AzureConfig{}, fmt.Errorf("error building azure cloud environment configuration: %w", err)
 	}
--- a/backend/pkg/azure/config/azure_cloud_environment.go
+++ b/backend/pkg/azure/config/azure_cloud_environment.go
@@ -37,23 +37,8 @@ type AzureCloudEnvironmentBuilder struct {
 	tracerProvider   trace.TracerProvider
 }
 
-func NewAzureCloudEnvironmentBuilder() *AzureCloudEnvironmentBuilder {
-	return &AzureCloudEnvironmentBuilder{}
-}
-
-func (a *AzureCloudEnvironmentBuilder) CloudEnvironment(cloudEnvironment string) *AzureCloudEnvironmentBuilder {
-	a.cloudEnvironment = cloudEnvironment
-	return a
-}
-
-func (a *AzureCloudEnvironmentBuilder) TracerProvider(
-	tracerProvider trace.TracerProvider) *AzureCloudEnvironmentBuilder {
-	a.tracerProvider = tracerProvider
-	return a
-}
-
-func (a *AzureCloudEnvironmentBuilder) Build() (AzureCloudEnvironment, error) {
-	if a.cloudEnvironment == "" {
+func NewAzureCloudEnvironment(cloudEnvironment string, tracerProvider trace.TracerProvider) (AzureCloudEnvironment, error) {
+	if cloudEnvironment == "" {
 		return AzureCloudEnvironment{}, errors.Errorf("cloud environment cannot be empty")
 	}
 
@@ -88,21 +73,21 @@ func (a *AzureCloudEnvironmentBuilder) Build() (AzureCloudEnvironment, error) {
 		},
 	}
 
-	configuration, ok := azureCloudEnvironmentConfigurationMapping[a.cloudEnvironment]
+	configuration, ok := azureCloudEnvironmentConfigurationMapping[cloudEnvironment]
 	if !ok {
 		return AzureCloudEnvironment{},
-			errors.Errorf("cloud environment %q is not supported", a.cloudEnvironment)
+			errors.Errorf("cloud environment %q is not supported", cloudEnvironment)
 	}
 
 	clientOptions := policy.ClientOptions{
 		Cloud: configuration.cloud,
 	}
-	if a.tracerProvider != nil {
-		clientOptions.TracingProvider = azotel.NewTracingProvider(a.tracerProvider, nil)
+	if tracerProvider != nil {
+		clientOptions.TracingProvider = azotel.NewTracingProvider(tracerProvider, nil)
 	}
 
 	return AzureCloudEnvironment{
-		id:                       a.cloudEnvironment,
+		id:                       cloudEnvironment,
 		configuration:            configuration.cloud,
 		rdbmsScope:               configuration.rdbmsScope,
 		checkAccessV2Environment: configuration.checkAccessV2Environment,
@@ -116,7 +101,7 @@ func (a AzureCloudEnvironment) PolicyClientOptions() policy.ClientOptions {
 	return a.clientOptions
 }
 
-// ARMClientOptions returns an arm.ClientOptions instance from the current
+// ArmClientOptions returns an arm.ClientOptions instance from the current
 // Azure Cloud environment.
 func (a AzureCloudEnvironment) ARMClientOptions() *arm.ClientOptions {
 	return &arm.ClientOptions{


backend/controllers/cluster_inflights.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// TODO this might not be how we instantiate the validations nor introduce them,
+	// for example we might decide that the clusterInflights type just receive an array of them directly,
+	// but it showcases what kind of dependencies are used and when are they needed.
+	var clusterValidations []ClusterValidation
+	clusterValidations = append(clusterValidations,
+		NewAzureRpRegistrationValidation(
+			// TODO what do we do when we need a logger and we are outside of the context
+			// where a Go context is available? For now a new one is created but this is not
+			// how it is going to be performed
```
deads2k: plumb the context so you have it. Always use the context logger.

deads2k: Never ever pass a direct logger instance.  doing so makes it impossible to build up a series of context values derived from the call chain itself and is strictly worse.  Instanitating with a logger loses that context of the callchain, which is what actually matters, not the instantiation of the instance.

backend/controllers/azure_rp_registration_validation.go
```diff
@@ -0,0 +1,104 @@
...
+
+func NewAzureRpRegistrationValidation(
+	l *slog.Logger,
+	name string,
+	fpaTokenCredRetriever fpa.FirstPartyApplicationTokenCredentialRetriever,
+	azureCloudEnvironment azureconfig.AzureCloudEnvironment,
+) *AzureRpRegistrationValidation {
+	logger := l.With("validation_name", azureRpRegistrationValidationName)
+	return &AzureRpRegistrationValidation{
+		logger:                logger,
```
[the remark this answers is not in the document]
deads2k: Every function should have a context.  Never create a function that lacks a context.  Even if you don't need it for cancellation, anything that might ever log needs to have it.

backend/controllers/azure_rp_registration_validation.go
```diff
@@ -0,0 +1,104 @@
...
+
+const (
+	// RpRegistrationStateRegistered - Resource provider is registered
+	azureRpRegistrationStateRegistered string = "Registered"
+	azureRpRegistrationValidationName  string = "azure-rp-registration-validation"
+)
+
+// The RpRegistrationValidation struct validates the states of several
+// Azure Resource Providers associated with a clusters region, subscription, etc.
+type AzureRpRegistrationValidation struct {
```
deads2k: `getResourceProvidersClient func(sub, tenant, something)`


then in unit testing, you instantiat with 

```
&AzureRpRegistrationValidation{
getRPClient: fun(foo){ return hardcoded mock}
```

backend/controllers/azure_rp_registration_validation.go
```diff
@@ -0,0 +1,115 @@
...
+	logger                           *slog.Logger
```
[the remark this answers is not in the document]
miguelsorianod: yes I forgot to push that leftover change, I did it now

backend/controllers/azure_rp_registration_validation.go
```diff
@@ -0,0 +1,113 @@
...
+// The RpRegistrationValidation struct validates the states of several
+// Azure Resource Providers associated with a clusters region, subscription, etc.
+type AzureRpRegistrationValidation struct {
+	name                             string
+	resourceProvidersClientRetriever azureclient.ResourceProvidersClientRetriever
+	fpaTokenCredRetriever            fpa.FirstPartyApplicationTokenCredentialRetriever
+	azureCloudEnvironment            azureconfig.AzureCloudEnvironment
+}
+
+func NewAzureRpRegistrationValidation(
```
deads2k: Thinking structurally, I'd say that this validation is useful to run, but I woiuldn't block any execution of other controllers based on it.  If this fails, it should be the first item reported to the user in conditions, but every other controller trying to provision should run even if this fails or hasn't run at all. Those that can proceed will work, those that cannot will not work.  But we separately select the condition we provide back and can prioritize this one.

It's a great one to bring in first to work out creds, clients, interfacees.  But I dont' see it as gating other controllers.

backend/controllers/cluster_inflights.go
miguelsorianod: Adding another note just to have it written down:

There are a couple of implications that I can think of about the context approach:
* It makes the logger dependency less explicit/intuitive, as it is now stored in the Go context. That means it is also stored as unstructured attribute. We mitigate this by providing helper functions to retrieve it and convention.
* It impacts the signature of almost all the methods/functions as now a Go context has to be passed almost everywhere, as logging needs could occur everywhere
* Because we don't pass loggers as attribute dependencies we cannot achieve easily something like "all the methods of this type should have this contextual information". We have to instead in each method of the type have a `.With` statement that includes it, which sometimes can be forgotten/not consistent due to human error

When balancing the benefits vs those implications we still favor the benefits of having all the contextual logging information through the call chain as it helps to debug better.

backend/pkg/azure/config/azure_cloud_environment.go
```diff
@@ -0,0 +1,148 @@
...
+			cloud:      cloud.AzureChina,
+			rdbmsScope: "https://ossrdbms-aad.database.chinacloudapi.cn",
+			checkAccessV2Environment: checkAccessV2Environment{
+				domainSuffix: "azure.cn",
+				scope:        "https://authorization.azure.cn/.default",
+			},
+		},
+		"AzurePublicCloud": {
+			cloud:      cloud.AzurePublic,
+			rdbmsScope: "https://ossrdbms-aad.database.windows.net/.default",
```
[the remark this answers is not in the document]
miguelsorianod: As far as I am aware, ServiceConfiguration does not contain the information.

Here we also map a cloud string identifier, which will be provided to the backend service as part of the deployment configuration, to its associated information.

Notice also how we provide cloud.Configuration too.

miguelsorianod:
I added a couple of more commits that iterate on the previous content:
* The validation introduced here now receives a azureclient.ClientBuilder that allows to retrieve the ResourceProviders client. This simplifies the number of dependencies received
* Three new required CLI flags are added:
  * --azure-runtime-config-path: Contains json/yaml configuration associated to some, but not all, Azure related settings that the service needs to receive. We call this "the Azure runtime configuration".
  * --azure-first-party-application-certificate-bundle-path: Path to a file containing an X.509 Certificate based client certificate, consisting of a private key and certificate chain, in a PEM or PKCS#12 format for authenticating clients with a first party application identity
  * --azure-first-party-application-client-id: The Client ID of the first party application identity
* Added backend/pkg/azure/configdto packages that defines the data types needed to serialize/deserializse configuration files related to Azure. In this MR we introduce the AzureRuntimeConfig data type which is used to serialie/deserialize the contents of the config file provided in --azure-runtime-config-path. The types also contain methods to perform validation. Notice how there's also the backend/pkg/azure/config package which also defines data types related to configuration but those are the domain layer ones, which are decoupled
* Added functions (for now in main.go, just to showcase) that load the azure runtime config file contents and build the AzureConfig domain layer type (in backend/pkg/azure/config) receiving the DTO type and other information
* The DTO types have all attributes exported to be friendly to serialization/deserialization. For the types in backend/pkg/azure/config (the domain layer ones) they are unexported

backend/main.go
```diff
@@ -278,6 +376,42 @@ func Run(cmd *cobra.Command, args []string) error {
...
+		// cluster(the tenant of the cluster, the subscription id, ...)
+		fpaTokenCredRetriever, err := fpa.NewFirstPartyApplicationTokenCredentialRetriever(
+			logger,
+			"TODO", // TODO: receive the client ID associated to the FPA via CLI flag
+			certReader,
+			// Notice how we are passing the policy client options from the Azure cloud environment type.
+			azureConfig.RuntimeConfig.CloudEnvironment.PolicyClientOptions(),
+		)
+
+		fpaClientBuilder := azureclient.NewFpaClientBuilder(
```
deads2k: Open another PR that doesn't include any of the controller changes, but does make a call using this new clientbuilder right here.  I'd like to see it succeed or fail.

backend/pkg/azure/configdto/azure_runtime_config.go
```diff
@@ -0,0 +1,185 @@
...
+	"net/url"
+	"slices"
+	"strings"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/azure/resourceid"
+	"github.com/Azure/ARO-HCP/backend/pkg/azure/urlvalidation"
+)
+
+// AzureRuntimeConfig represents user provided Azure related configuration for running the service
+type AzureRuntimeConfig struct {
```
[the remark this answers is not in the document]
deads2k: What is a DTO?

backend/pkg/azure/resourceid/resourceid.go
```diff
@@ -0,0 +1,354 @@
...
+// specified in `rawResourceID` of Azure Resource Type `resourceType`.
+// The Resource ID is assumed to be a Resource Group scoped Azure Resource ID.
+// An error is returned if `rawResourceID` cannot be parsed.
+// The validations that are performed are:
+// - The rawResourceID is a valid Azure Resource ID
+// - The resource type of the rawResourceID matches the provided resourceType
+// - The Azure Subscription ID of rawResourceID can be parsed and it is not empty
+// - The Azure Resource Group Name of rawResourceID can be parsed and it is not empty
+// - The Azure Resource Name of rawResourceID can be parsed and it is not empty
+func ParseResourceGroupScopedResourceID(
```
deads2k: No.  You can add validations for a resourceID, but keep it distinct from parsing to encourage using the correct types in the code and not strings.

backend/pkg/azure/resourceid/resourceid.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+// ParseVnetResourceID parses an Azure VNet Resource ID string
+// specified in `rawVnetResourceID`. An error is returned if `rawVnetResourceID`
+// cannot be parsed.
+// The validations that are performed are:
+//   - The vnet resource id is a valid Azure Resource ID
+//   - The resource type of the vnet resource id is the expected one for Azure
+//     VNets.
+func ParseVnetResourceID(rawVnetResourceID string) (azcorearm.ResourceID, error) {
```
deads2k: all these checks seem extraneous. We validate before we store and if the types are wrong, then I'd expect the lookups to fail later on.  This looks like an unnecessary layer of indirection.

miguelsorianod: An example is the azure runtime config file. A resource ID to an ACR registry is passed.
When we validate the file we check that the provided resource id is valid syntactically and semantically (https://github.com/Azure/ARO-HCP/pull/3735/files#diff-7ca1037ebeacf0254a78c85c3c325507f286a43c4a121a27b80634c9ae8d7805R127). We give an early feedback by validating at that point and we do not wait to let it fail at a potentially later point in time where the connection is not so clear anymore to the config file.

PR #3741, opened 2026-01-08
add frontend testing for nodes to gather exist cosmos data for testing future changes

deads2k:
needed so that we can be sure what we're able to read old node data before the migration.

### DIFF
--- a/test-integration/frontend/artifacts/FrontendCRUD/NodePool/read-old-data/01-load-old-data/cluster-create-with-tags.json
+++ b/test-integration/frontend/artifacts/FrontendCRUD/NodePool/read-old-data/01-load-old-data/cluster-create-with-tags.json
@@ -1,0 +1,83 @@
+{
+	"_attachments": "attachments/",
+	"_etag": "\"00000000-0000-0000-7121-fd067c0501dc\"",
+	"_rid": "WZlNALYurecDAAAAAAAAAA==",
+	"_self": "dbs/WZlNAA==/colls/WZlNALYurec=/docs/WZlNALYurecDAAAAAAAAAA==/",
+	"_ts": 1766174370,
+	"id": "c893be2c-80b0-497f-8490-e3e5fd2b2cc4",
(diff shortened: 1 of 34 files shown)


miguelsorianod:
/retest

e2e-parallel tests failed but it seems unrelated to MR changes

miguelsorianod:
I've taken a look at it and it looks good to me.

I write down notes here that I've taken with my understanding after having read implementation details of the code:

This MR adds a NodePool FrontendCRUD test called "read-old-data" that conceptually does the following:
1. It prepares the test by directly writing into cosmos information about: a cluster (in succeeded state), nodepool1 (in accepted state), nodepool2 (in accepted state), azure subscription, arm operations (3) associated to the cluster and nodepools, with matching operation state to the corresponding resource
2. It sets up GoMock expectations of the RP's CS client interface so it mocks returned data by CS API associated to the cluster, nodepools and autoscaler. Those conceptually match whatever is in cosmos
3. It verifies that doing a GET of nodepool1 against the frontend api returns the same content as the provided expected one at api level
4. It verifies that doing a LIST of nodepools against the frontend api returns the same content as the provided expected one at api level
5. It verifies that doing a LIST activeoperations against the frontend api returns the same content as the provided expected one at api level for the operations associated to nodepool1 and 2 which are nodepools that represent are still being created. Here the operation associated to cluster is not expected because the cluster is in state succeeded and the test here is about listing active operations
6. It performs an UPDATE of nodepool1 against the frontend API, providing a payload representing the update to perform. Specifically, it seems to change an attribute representing a k8s taint
7. It verifies that doing a GET of nodepool1 against the frontend api returns the same content as the provided expected one at api level, which now is the one with the changed k8s taint in the nodepool
8. Finally, it verifies that the contents in cosmosdb after all these actions are the expected ones for cluster, nodepool1, nodepool2, operationcreatecluster, operationcreatenodepool1, operationcreatenodepool2, operationupdatenodepool1, subscription

deads2k:
rebased to update the location

PR #3745, opened 2026-01-09
E2E Test Cluster Creation: Add retry to cluster creation

janboll:
### What

Add a retry to cluster creation to make it resilient against capacity issues. 
### Why

Observed to many clusters created in Integration E2E tests. Since we run tests in parallel, it appears valid to simply retry on certain errors to not fail in such "easy" cases.


stevekuznetsov (changes requested):
/hold

PR #3779, opened 2026-01-13
Use structured logging rather than logging with format strings

bennerv:
### What

Uses structured logging rather than `logger.[Warn|Info|Error](fmt.Sprintf(...`


raelga (changes requested):
@bennerv is this still needed? I can do a followup PR with your changes rebase. lmk

PR #3794, opened 2026-01-15
Use fqdn of ingest service to avoid DNS server issues at cluster creation

pepedocs:
### What

This PR updates the config to use ingest fqdn.

### Why

During cluster creation in my personal dev, I consistently encountered the following error. The DNS server isn't yet reachable at that stage and therefore the cluster creation fails. Using the fqdn avoids the DNS resolution.

```
{"time":"2026-01-15T02:01:42.426118882Z","level":"ERROR","source":{"function":"github.com/Azure/ARO-HCP/internal/tracing.ConfigureOpenTelemetryTracer.func2","file":"/app/internal/tracing/otel_sdk.go","line":88},"msg":"OpenTelemetry.ErrorHandler: traces export: Post \"http://ingest.observability:4318/v1/traces\": dial tcp: lookup ingest.observability on 10.130.0.10:53: no such host"}
```

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1080,7 +1080,7 @@ clouds:
             server:
               mqttClientName: 'maestro-server-{{ .ctx.regionShort }}-dev'
               tracing:
-                address: "http://ingest.observability:4318"
+                address: "http://ingest.observability.svc.cluster.local:4318"
                 exporter: "otlp"
           # Frontend
           frontend:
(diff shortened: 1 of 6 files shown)


raelga (changes requested):
Some changes are also included in https://github.com/Azure/ARO-HCP/pull/3795/changes

Can you review both PRs? They are still needed?

PR #3824, opened 2026-01-19
Warn on subscription document missing from cosmosdb

bennerv:
Ideal state is to update the dump_data controller to include this, but it's not omnipresent since it's periodic, so this will at least tell us that we're missing the subscription from cosmosdb.


bennerv:
/close 

#3826 has more info than this provides

PR #3840, opened 2026-01-20
Draft of the new E2E test: trying to create a cluster with already used managed identities

aklymenk:
[ARO-22504](https://issues.redhat.com/browse/ARO-22504)

### What

Trying to create a cluster with already used managed identities and expecting the creation to fail.


davidffrench:
Closing — this draft is very stale and uses outdated E2E framework patterns. #6141 covers the related positive case (reusing MIs within the same cluster). If the cross-cluster MI reuse negative test is still desired, it should be written fresh against the current framework. Thank you for the original work, Andrii!

PR #3846, opened 2026-01-20
Add fipsEnabled property: the cluster would be using fips validated libs

mzazrivec:
https://issues.redhat.com/browse/ARO-17753

### What

This PR adds `fipsEnabled` property to ARO-HCP API specs.

### Why

This change is for the ARO-HCP product is to support scenarios and use cases where the control plane and its workers need to be fips enabled (i.e. running on systems using FIPS-enabled cryptographic libraries).


raelga (commented):
/ok-to-test

mzazrivec:
Closing in favor of https://github.com/Azure/ARO-HCP/pull/5935

PR #3897, opened 2026-01-26
Add 20251223 to testsdk.md

mzazrivec:
### What

Update to `api/testsdk.md`, adding section for the `v20251223preview` tag.

### Why

So that we can now run `make testsdk`.


raelga (commented):
/ok-to-test

PR #3921, opened 2026-01-28
[WIP] Add Managed Resource Group cluster provisioning controller

miguelsorianod:
This MR starts conversation around introducing a cluster provisioning controller that takes care of provisioning the ARO-HCP Cluster's Managed Resource Group when the cluster as part of the creation process.

The Managed Resource Group (MRG) is an Azure Resource Group associated to the ARO-HCP Cluster that is created and it is used by the ARO-HCP service (RP/CS/Hypershift/OCP, ...) to put Azure resources associated to the cluster within it.
In environments that are in the AME tenant, after the MRG is created a next step is that we create an Azure DenyAssignment that prevents the end-users from (easily) modifying/deleting the Managed Resource Group nor its contents within it directly. In lower environments Microsoft's DenyAssignment integration is not available so no prevention occurs.

Other relevant notes:
* MRG names (and RG names in general) are unique within a Subscription (Azure does not allow creating two with the same name within a Subscription)
* Once the ARO-HCP Cluster is deleted one of the deprovisioning steps is the deletion of the MRG
* The MRG must not previously exist
* MRG name cannot be equal to the RG name of the aro-hcp cluster
* MRG is located in the same Azure location as the cluster's location
* Not to be confused with `resource_group_name`, which is the Azure

### DIFF
--- a/backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
+++ b/backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
@@ -1,0 +1,250 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clusterprovisioningcontrollers
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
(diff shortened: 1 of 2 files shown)


backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
```diff
@@ -0,0 +1,214 @@
...
+func (c *managedResourceGroupProvisioningSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	existingCluster, err := c.cosmosClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
+	if database.IsResponseError(err, http.StatusNotFound) {
+		return nil // cluster doesn't exist, no work to do
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
+	}
+
+	// TODO do we want to store that the managed resource group was created in
```
deads2k: isn't the name always on the cluster itself in cosmos?

miguelsorianod:
Closing this as we can now create PRs with branches coming from our own forks. I will open a new PR with an up to date proposal around this.

PR #3925, opened 2026-01-28
API spec: add vnet integration subnet

tsatam:
[ARO-17758](https://issues.redhat.com/browse/ARO-17758)

### What

Adds a dedicated `vnetIntegrationSubnetId` field to HCPOpenShiftCluster's PlatformProfile. This field is required, and will thus be a breaking change to the API spec. It is implemented exclusively on the v20251223preview API. 

### Why

This field will be used for vnet injection, in order to run control plane components in user vnets via SWIFT.

### DIFF
--- a/internal/api/types_cluster.go
+++ b/internal/api/types_cluster.go
@@ -125,6 +125,7 @@ type ServiceProviderAPIProfile struct {
 type CustomerPlatformProfile struct {
 	ManagedResourceGroup    string                         `json:"managedResourceGroup,omitempty"`
 	SubnetID                *azcorearm.ResourceID          `json:"subnetId,omitempty"`
+	VnetIntegrationSubnetID *azcorearm.ResourceID          `json:"vnetIntegrationSubnetId,omitempty"`
 	OutboundType            OutboundType                   `json:"outboundType,omitempty"`
 	NetworkSecurityGroupID  *azcorearm.ResourceID          `json:"networkSecurityGroupId,omitempty"`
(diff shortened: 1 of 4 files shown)


api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -274,6 +275,15 @@ model PlatformProfile {
   @visibility(Lifecycle.Read, Lifecycle.Create)
   subnetId: SubnetResourceId;
 
+  /** The Azure resource ID of the subnet used to inject hosted
+   * control plane components into the cluster's vnet.
+   * Note that this subnet cannot be shared with the default worker subnet
+   * or any nodepools' subnets.
+   */
```
SudoBrendan: Maybe keep the customer-facing stuff more generic / high level?

`The Azure resource ID of the subnet used to establish Control Plane to Node communication. Cannot be shared with any NodePool subnets.`

SudoBrendan (approved):
LGTM - I think we could clean up the comment, which will show up in SDKs if you'd like to reword it. I was using https://learn.microsoft.com/en-us/azure/aks/api-server-vnet-integration#api-server-connectivity as a guide for what a comment could look like

SudoBrendan (approved):
/lgtm

bennerv:
Should we have validation on the field in `validate_cluster.go` ?

jharrington22 (changes requested):

raelga (commented):
https://issues.redhat.com/browse/ARO-17758 is now closed.

/close

PR #3975, opened 2026-02-03
Add a Route Monitor controller

venkateshsredhat:
### What

Current Route monitor Operator was built for ROSA and this setup required many hacks to make it work in ARO env .

### Why

RM controller will be a light weight controller to replace the Route Monitor Operator .

We will be adding tests to this and deployment yamls . This is just a draft .


route-monitor-controller/controllers/hostedcluster_controller.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (r *HostedClusterReconciler) reconcileServiceMonitor(ctx context.Context, hostedCluster *hypershiftv1beta1.HostedCluster) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	desired := r.buildServiceMonitor(hostedCluster)
+
+	//Set OwnerReference for automatic garbage collection
+	if err := controllerutil.SetControllerReference(hostedCluster, desired, r.Scheme); err != nil {
+		logger.Error(err, "Failed to set OwnerReference")
+		return ctrl.Result{}, err
+	}
```
roivaz: We can't just rely on garbage collection for deletion of ServiceMonitors. Garbage collection only kicks in when the owner resource is gone. The HostedCluster deletion process is a long one as the HostedCluster resource has several resource finalizers that need to be managed by its controller first and these take a while to complete -- we don't want the ServiceMonitor up and likely reporting API failures during this.
So the reconcile loop should look like:
* Get the HostedCluster resource
* Check if the `DeletionTimestamp` is set. If it is, the HostedCluster is being deleted and we need to ensure that the ServiceMonitor is deleted as well. 
* If the HostedCluster resource is not being deleted, then progress the normal reconcile flow.

route-monitor-controller/controllers/hostedcluster_controller.go
```diff
@@ -0,0 +1,205 @@
...
+	if err := r.Get(ctx, req.NamespacedName, &hostedCluster); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("HostedCluster not found, was deleted")
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "Failed to get HostedCluster")
+		return ctrl.Result{}, err
+	}
+
+	return r.reconcileServiceMonitor(ctx, &hostedCluster)
```
roivaz: Shouldn't we be checking here if the HostedCluster has actually finished provision? You should check the status and requeue with a delay (`RequeueAfter`) in case it isn't.

route-monitor-controller/controllers/hostedcluster_controller.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	logger.Info("ServiceMonitor is up to date")
+	return ctrl.Result{}, nil
+}
+
+func (r *HostedClusterReconciler) buildServiceMonitor(hostedCluster *hypershiftv1beta1.HostedCluster) (*monitoringv1.ServiceMonitor, error) {
+	serviceMonitorName := fmt.Sprintf("%s-route-monitor", hostedCluster.Name)
+	namespace := hostedCluster.Namespace
+
```
venkateshsredhat: @roivaz @ashishmax31  Any changes/suggestions on naming and namespace ?/

route-monitor-controller/controllers/hostedcluster_controller.go
```diff
@@ -0,0 +1,230 @@
...
+					Scheme:        "http",
+					Params:        params,
+					MetricRelabelConfigs: []monitoringv1.RelabelConfig{
+						{
+							Replacement: &routeURL,
+							TargetLabel: "probe_url",
+						},
+						{
+							Replacement: &hostedCluster.Spec.ClusterID,
+							TargetLabel: "_id",
```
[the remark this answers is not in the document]
venkateshsredhat: gotcha

route-monitor-controller/controllers/hostedcluster_controller.go
```diff
@@ -0,0 +1,230 @@
...
+	return &monitoringv1.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceMonitorName,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"app.kubernetes.io/name":       hostedCluster.Name,
+				"app.kubernetes.io/managed-by": "route-monitor-controller",
+			},
+		},
+		Spec: monitoringv1.ServiceMonitorSpec{
```
venkateshsredhat: @roivaz what about the namespace selector etc https://github.com/Azure/ARO-HCP/blob/2a008fc2bb9a84f334a8786ee175f1219595c6ef/route-monitor-operator/deploy/helm/route-monitor-operator/templates/kas-monitoring-hack.policy.yaml#L67-L68

route-monitor-controller/controllers/hostedcluster_controller.go
```diff
@@ -0,0 +1,230 @@
...
+	if !hostedCluster.DeletionTimestamp.IsZero() {
+		logger.Info("HostedCluster is being deleted, cleaning up ServiceMonitor")
+		return r.handleDeletion(ctx, &hostedCluster)
+	}
+
+	availableCondition := meta.FindStatusCondition(hostedCluster.Status.Conditions, string(hypershiftv1beta1.HostedClusterAvailable))
+	if availableCondition == nil {
+		logger.Info("HostedCluster is still being Provisioned")
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
venkateshsredhat: @roivaz  yes we need this :
```
if availableCondition.Status != metav1.ConditionTrue {
//blah  blah
    return ctrl.Result{}, nil
}

route-monitor-controller/controllers/hostedcluster_controller.go
venkateshsredhat: hcp-kas-monitor isnt this the existing name? Dont want it to be difficult to locate that's all .

PR #4046, opened 2026-02-09
[WIP] Create new Deny Assignment backend controller

rogbas:


backend/pkg/controllers/deny_assignment.go
```diff
@@ -0,0 +1,867 @@
...
+
+	return allCreated, nil
+}
+
+// fetchPrincipalIDs fetches principal IDs for all operator managed identities
+func (c *denyAssignmentReconciler) fetchPrincipalIDs(ctx context.Context, cluster *api.HCPOpenShiftCluster, subscriptionID string) (map[string]string, error) {
+	logger := utils.LoggerFromContext(ctx)
+	principalIDs := make(map[string]string)
+
+	msiClient, err := armmsi.NewUserAssignedIdentitiesClient(subscriptionID, c.credential, nil)
```
machi1990: What credential will be used to create this client? is it the FPA  or the SMI (service managed identity)? 

If it is the former, I do not think if we can retrieve the client id directly via ARM - there could be permissions issues. 

CS uses the FPA to retrieve control plane MIs info from MSI RP directly and the SMI credential to retrieve the principal id of the data plane clients directly from ARM.

The way I see this flow is;
(1) FPA is used to retrieve credential of the SMI
(2) From which a credential retrieve in step(1) a client is created similar to what's done here 
(3) The client is used to fetch the principal id of those CP + DP.

backend/pkg/controllers/deny_assignment.go
```diff
@@ -0,0 +1,867 @@
...
+		logger.Info("some deny assignments are still being created")
+	}
+
+	return nil
+}
+
+// isClusterReadyForDenyAssignments checks if a cluster is in a state where
+// deny assignments should be managed
+func isClusterReadyForDenyAssignments(cluster *api.HCPOpenShiftCluster) bool {
+	// Only process clusters that are Succeeded or Updating
```
miguelsorianod, line 148: This doesn't seem correct. We shouldn't depend on the provisioning state. We should create it as long as it's not there and the cluster hasn't been marked as deleted

machi1990:
Superseded by https://github.com/Azure/ARO-HCP/pull/6269

geoberle:
closing as stale
reopen if you still needed

PR #4069, opened 2026-02-11
custom-link-tools: add cluster/kusto args

roivaz:
* Add CLI arguments to be able to derive cluster names and kusto url in persistent environments where steps.yaml is not available.
* Add discovery of mgmt cluster name from steps.yaml


janboll (commented):
/lgtm

raelga (approved):

mmazur:
/hold

Lemme do it with the full config.yaml in SHARED_DIR

PR #4157, opened 2026-02-19
remove old resourceID serialization

deads2k:
After https://github.com/Azure/ARO-HCP/pull/5113 gets to production we can do this.  We need to be sure the backend doesn't rely on the field we're removing before we remove it.


deads2k:
rebased on deletion fix https://github.com/Azure/ARO-HCP/pull/4395

deads2k:
test again with the latest controller code and then check status to try to figure out which call isn't being made

deads2k:
done

/hold cancel

PR #4192, opened 2026-02-24
[WIP] feat: add controller that retrieves information about MSI based identities

miguelsorianod:
This PR builds on top of https://github.com/Azure/ARO-HCP/pull/3820.

We add a controller that retrieves the Client ID and Principal ID associated to the following identities associated to an ARO-HCP Cluster:
* The Control Plane operators identities
* The Service Managed Identity

We leverage Microsoft's Managed Identities Data Plane service to retrieve the information. When the service is not available (outside of AME tenants) the fake managed identities data plane client is leveraged which returns the information associated to the mock msi identity for all requests/responses to it. We do not directly use Azure Go SDK's UserAssignedIdentities client because otherwise we would return the information of clientid+principalid of the passed identities in the payload instead of the actual clientid+principalid that ends up being used in the management cluster side.


miguelsorianod:
Superseded by https://github.com/Azure/ARO-HCP/pull/6301

PR #4193, opened 2026-02-24
[WIP] feat: add controller that calculates Cluster Data Plane Identities extra information

miguelsorianod:
We add a controller that retrieves the Client ID and Principal ID associated to the Data Plane operators identities associated to the ARO-HCP Cluster.

We leverage the Service Managed Identity associated to the ARO-HCP Cluster to retrieve the Data Plane operators identities information. We use Azure Go SDK's UserAssignedIdentities API to retrieve it. This is a different method than what's done for MSI based identities where the Managed Identities Data Plane service is used instead (https://github.com/Azure/ARO-HCP/pull/3820). This is because for the control plane identities. on the environments where the managed identities data plane service is not available we use the mi mock identity instead, which includes its clientid+principalid instead of the ones associated to the identities passed in the cluster payload. By using the mock managed identities data plane client we retrieve that transparently.


backend/pkg/controllers/fetch_data_plane_operators_managed_identities_info.go
```diff
@@ -0,0 +1,189 @@
...
+			continue
+		}
+
+	}
+	if len(identitiesToSync) == 0 {
+		return nil
+	}
+
+	smiResourceID := existingCluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ServiceManagedIdentity
+	uaisClient, err := c.smiClientBuilder.UserAssignedIdentitiesClient(ctx, existingCluster.ServiceProviderProperties.ManagedIdentitiesDataPlaneIdentityURL, smiResourceID, existingCluster.ID.SubscriptionID)
```
deads2k:
```suggestion
	userAssignedIdentitiesClient, err := c.smiClientBuilder.UserAssignedIdentitiesClient(ctx, existingCluster.ServiceProviderProperties.ManagedIdentitiesDataPlaneIdentityURL, smiResourceID, existingCluster.ID.SubscriptionID)
```

don't be stingy.

miguelsorianod:
Superseded by https://github.com/Azure/ARO-HCP/pull/6300

PR #4211, opened 2026-02-25
Add cosmosmetadata to internal types.

deads2k:
built on https://github.com/Azure/ARO-HCP/pull/4198

after https://github.com/Azure/ARO-HCP/pull/4103 makes it to production, we can merge this PR. We cannot do it before because if revert from the level containing https://github.com/Azure/ARO-HCP/pull/4103 to the level just before, we need the old level to be able to understand new data that was written.

This puts cosmosmetadata into the internal API types so we can collapse down more easily once https://github.com/Azure/ARO-HCP/pull/4198 makes it to prod.  We don't have to wait on https://github.com/Azure/ARO-HCP/pull/4198 for this change.

not in [9fc81a5](https://github.com/Azure/ARO-HCP/commit/9fc81a50bac83d869ed4b68c88accbd344507656)


deads2k:
taking a different approach

PR #4272, opened 2026-03-02
record cosmos resource counts and controller conditions as metrics

geoberle:
### What

expose metrics for cosmos objects
    
* backend_resource_provider_objects (by provisioning_state)
* backend_controller_status_conditions (by resource_type, controller_name, condition, status)
    
add a BackendControllerDegraded alert that fires per controller_name and resource_type when >5% of instances report Degraded=True for 5m.

this PR also makes sure to bring direct backend metrics under one prometheus registry to enable proper testing.

### DIFF
--- a/backend/pkg/informers/metrics/metrics.go
+++ b/backend/pkg/informers/metrics/metrics.go
@@ -1,0 +1,286 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


backend/pkg/informers/metrics/metrics.go
```diff
@@ -0,0 +1,286 @@
...
+	resourcesObjectCounts            *prometheus.GaugeVec
+	controllerConditionsObjectCounts *prometheus.GaugeVec
+}
+
+// NewStorageMetricsObserver creates a StorageMetricsObserver that periodically
+// polls backendInformers and emits statistics as metrics.
+func NewStorageMetricsObserver(registerer prometheus.Registerer, backendInformers informers.BackendInformers) *StorageMetricsObserver {
+	return &StorageMetricsObserver{
+		backendInformers: backendInformers,
+		pollPeriod:       1 * time.Minute,
```
stevekuznetsov: Is this the pattern from kube-state-metrics? I guess naively I would have expected the processing to just react to informer events and, given the object being processed, update the in-memory state of the metric, and let the scraper do whatever it needs to at whatever interval it uses.

PR #4286, opened 2026-03-02
Proposed: Automatically fix yaml if make command not run

hlipsig:
### What
Just an idea, fix PRs where make yamlfmt were not run automatically instead of waiting on prow to fail.

### Why

Saved a developer loop over something solved pretty easily.

Totally acceptable to reject. Just an idea.


raelga:
@hlipsig Thanks for this, but this may add unnecesary extra commits for wip PRs and so far this has not been an issue as is part of the workflow.

/close

PR #4344, opened 2026-03-06
our current clean misses a few azure labels

venkateshsredhat:
Currently our cleaned logs have these labels and we should not be having these in our cleaned output .


raelga (approved):
/retest

raelga (changes requested):
Is this still needed?

PR #4370, opened 2026-03-09
only read new keys

deads2k:
only read new key format

### DIFF
--- a/internal/database/crud_helpers.go
+++ b/internal/database/crud_helpers.go
@@ -47,20 +47,7 @@ func oldResourceIDStringToCosmosID(resourceID string) (string, error) {
 	return strings.ReplaceAll(strings.ToLower(resourceID), "/", "|"), nil
 }
 
-// TODO this will eventually be the standard GET, but until we rewrite all records with new `id` values, it must remain separate and specifically called.
-func getByItemID[InternalAPIType, CosmosAPIType any](ctx context.Context, containerClient *azcosmos.ContainerClient, partitionKeyString string, cosmosID string) (*InternalAPIType, error) {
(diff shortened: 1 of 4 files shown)


deads2k:
/hold

the integration test will fail again.  I need to look at ahigher priority thing. Weird that validation broke though....

deads2k:
/hold cancel

will be slightly easier to read after https://github.com/Azure/ARO-HCP/pull/5004

internal/database/crud_helpers.go
```diff
@@ -75,57 +62,29 @@ func getByItemID[InternalAPIType, CosmosAPIType any](ctx context.Context, contai
...
+	return responseItemToInternalObj[InternalAPIType, CosmosAPIType](ctx, cosmosID, responseItem)
+}
 
-	logger.Info("migrating item", "newCosmosID", newExactCosmosID, "oldCosmosID", originalCosmosID)
-	if _, err := containerClient.CreateItem(ctx, partitionKey, newBytes, nil); err != nil {
-		return nil, utils.TrackError(err)
-	}
-	if _, err = containerClient.DeleteItem(ctx, partitionKey, originalCosmosID, nil); err != nil {
+func get[InternalAPIType, CosmosAPIType any](ctx context.Context, containerClient *azcosmos.ContainerClient, partitionKeyString string, completeResourceID *azcorearm.ResourceID) (*InternalAPIType, error) {
+	// try to see if the cosmosID we've passed is also the exact resource ID.  If so, then return the value we got.
```
miguelsorianod, line 83: nitpick: I think this code comment is not needed anymore, because if we can't we are now returning an error

PR #4403, opened 2026-03-11
Managed Identities Role Definition Getter 

shitaljante:
### What

This PR introduces changes that will be used to fetch the role definitions to filter the actions and Cache the actions with permissions in a map 

This reference comes from https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/9596


backend/pkg/azure/client/mock_role_definitions_client.go
```diff
@@ -0,0 +1,96 @@
+// Code generated by MockGen. DO NOT EDIT.
```
patilsuraj767: The generated mocks in `mock_role_definitions_client.go` don't appear to be used anywhere currently. We should either remove the file if it is not needed, or add unit tests for getter.go that actually utilize these mocks.

shitaljante:
@mbarnes can you please approve this PR?

backend/pkg/azure/roledefinitions/getter.go
```diff
@@ -0,0 +1,232 @@
...
+// The 6-hour TTL was decided arbitrarily, to find a balance between frequent calls to fetch the role definition
+// and having the latest role definition.
+// This value could change in the future if needed.
+const roleDefinitionResourceIdCacheKeyTtl = 6 * time.Hour
+
+// cachedRoleDefinition holds an Azure role definition along with the timestamp of
+// when it was last updated in the cache.
+type cachedRoleDefinition struct {
+	roleDefinition armauthorization.RoleDefinition
+	lastUpdate     time.Time
```
miguelsorianod: let's document that it's in UTC

backend/pkg/azure/roledefinitions/getter.go
```diff
@@ -0,0 +1,232 @@
...
+	return dataActionsUnion.UnsortedList(), nil
+}
+
+func (s *CachedRoleDefinitionsGetter) isStale(roleDefinition cachedRoleDefinition) bool {
+	return time.Since(roleDefinition.lastUpdate) > roleDefinitionResourceIdCacheKeyTtl
+}
+
+func (s *CachedRoleDefinitionsGetter) extractRoleDefinitionActions(roleDefinition armauthorization.RoleDefinition) ([]string, error) {
+	if roleDefinition.Properties == nil || roleDefinition.Properties.Permissions == nil {
+		return nil, fmt.Errorf("role definition '%s' doesn't contain permissions", stringValue(roleDefinition.ID))
```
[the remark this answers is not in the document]
shitaljante: Fixed

backend/pkg/azure/roledefinitions/getter.go
```diff
@@ -0,0 +1,244 @@
...
+//
+// The 6-hour TTL was decided arbitrarily, to find a balance between frequent calls to fetch the role definition
+// and having the latest role definition.
+// This value could change in the future if needed.
+const roleDefinitionResourceIDCacheKeyTTL = 6 * time.Hour
+
+// RoleDefinitionsGetter interface is thread-safe, future implementations of this interface should ensure thread-safety.
+
+// RoleDefinitionsGetter defines cached access to Azure role definition permissions.
+type RoleDefinitionsGetter interface {
```
miguelsorianod, line 45: We are going to approach the design differently in the RP. There are a couple of design characteristics we want:
* We want to make the consumer code be aware that what is being used is cached information
* We want to decouple the retrieval and caching of information from Azure to higher level operations performed on it

Because of this, the new design would look like the following:
* In backend/pkg/azure/client we keep the RoleDefinitions client
* We have a new package backend/pkg/azure/cachedreader that has a new interface `RoleDefinitions` that complies 1:1 with the azure go sdk client methods. The cached information is whatever the sdk returns, The implementation of this interface contains the cache, and it receives a backend/pkg/azure/client.RoleDefinitions client
* We have a new package backend/pkg/azure/azurehelpers that provides package-level functions that operate on pure azure go sdk data. This would include the higher level methods like the one that acts on multiple role definitions etc.

backend/pkg/app/backend_identity_wiring.go
```diff
@@ -53,8 +54,14 @@ func NewBackendIdentityAzureClients(ctx context.Context, azureConfig *azureconfi
 		return nil, utils.TrackError(fmt.Errorf("failed to create dataplane identities OIDC configuration blob storage client: %w", err))
 	}
 
+	roleDefinitionsClient, err := armauthorization.NewRoleDefinitionsClient(defaultAzureCredential, azureConfig.CloudEnvironment.ARMClientOptions())
```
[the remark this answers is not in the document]
miguelsorianod: We should be fine using our tenant

backend/pkg/azure/roledefinitions/getter.go
shitaljante: Thanks for the context on design approach. I have one question reg "We want to make the consumer code be aware that what is being used is cached information"
- How will the consumer explicitly know whether the data returned by the cachedreader is coming from cache versus a fresh call to Azure, and does this affect how they use the data?

shitaljante: Updated the changes based on the suggested design.

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,128 @@
...
+		return armauthorization.RoleDefinitionsClientGetByIDResponse{}, err
+	}
+	c.roleDefinitionsCacheLock.RLock()
+	entry := c.roleDefinitionsCache[roleDefinitionResourceID]
+	c.roleDefinitionsCacheLock.RUnlock()
+	return entry.response, nil
+}
+
+// NewListPager forwards to the underlying client without caching.
+func (c *CachedRoleDefinitions) NewListPager(scope string, options *armauthorization.RoleDefinitionsClientListOptions) *runtime.Pager[armauthorization.RoleDefinitionsClientListResponse] {
```
miguelsorianod: That doesn't seem correct. If the client is cached all interactions to the client should be pulling from the same cache and complying with the definition of the ttl of the cache

backend/pkg/app/backend_identity_wiring.go
```diff
@@ -54,8 +55,14 @@ func NewBackendIdentityAzureClients(ctx context.Context, azureConfig *azureconfi
...
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to create role definitions client: %w", err))
+	}
+
 	clients := &azureclient.BackendIdentityAzureClients{
```
miguelsorianod, line 64: Here the question is, how do we pass both the cached and uncached clients, as they will probably be both needed at some point.

The same question applies when for the other identities (FPA, SMI, ...), but being the difference there that those provide ability to build the clients deferred instead of containing instantated ones (like the backend identity does in this PR)

backend/pkg/azure/client/backend_identity_clients.go
```diff
@@ -23,4 +23,7 @@ type BackendIdentityAzureClients struct {
 	// OIDC configuration associated to the ARO-HCP Clusters' Data Plane Operators
 	// Azure Identities
 	DataplaneIdentitiesOIDCConfigurationBlobStorageClient BlobStorageClient
+	// RoleDefinitionsClient is the client used to get Azure RBAC role definitions
+	// (e.g. for resolving role definition IDs to allowed actions).
+	RoleDefinitionsClient RoleDefinitionsClient
```
miguelsorianod, line 28: Here we would eventually end up needing both the cached and uncached variants of some clients. However, this would cause a package cyclic import.
A similar problem would occur with the other identities that use instead client builder interfaces. We need to figure out how do we want to design that.

backend/pkg/azure/client/role_definitions_client.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2"
+)
+
+// RoleDefinitionsClient is an interface that mirrors armauthorization.RoleDefinitionsClient.
+// (https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/resourcemanager/authorization/armauthorization/v2).
+// If the Azure Go SDK adds methods to RoleDefinitionsClient, add them here to keep parity.
+type RoleDefinitionsClient interface {
```
miguelsorianod, line 29: Here because this interface and the one in cachedreader have the same methods in theory one could use them interchangeably but that's probably not what we want because we don't want to pass accidentally a cached client to a place that expects to receive an uncached one and the other way around

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,128 @@
...
+
+// NewCachedRoleDefinitions returns a RoleDefinitions implementation that caches GetByID.
+func NewCachedRoleDefinitions(inner azureclient.RoleDefinitionsClient) *CachedRoleDefinitions {
+	return &CachedRoleDefinitions{
+		inner:                inner,
+		roleDefinitionsCache: make(map[string]cachedGetByIDResponse),
+	}
+}
+
+// Get forwards to the underlying client without caching.
```
deads2k: why would it be this way?

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,128 @@
...
+	if err := c.ensureCachedGetByID(ctx, roleDefinitionResourceID, options); err != nil {
+		return armauthorization.RoleDefinitionsClientGetByIDResponse{}, err
+	}
+	c.roleDefinitionsCacheLock.RLock()
+	entry := c.roleDefinitionsCache[roleDefinitionResourceID]
+	c.roleDefinitionsCacheLock.RUnlock()
+	return entry.response, nil
+}
+
+// NewListPager forwards to the underlying client without caching.
```
deads2k: if it's not cached, then it's not in the interface.

deads2k (changes requested):
I've refined interfaces in https://github.com/Azure/ARO-HCP/pull/5046 and tried to demonstrate: we never want to abstract away "is it a cache or not".  Because of that, misaligning interfaces is useful.

Additionally, everywhere that wants to use a cache or live data will take the interface (or both) that they actually want.

Also, naming something cached and then having it actually be not-cached is an anti-pattern.

shitaljante:
@deads2k @miguelsorianod Thank you, I agree with the comments and feedback. I’ve incorporated the changes from https://github.com/Azure/ARO-HCP/pull/5046 into this PR. 
I see a few comments from Copilot, I’ll address those, and then this PR should be ready to merge.

test-integration/go.mod
```diff
@@ -27,6 +27,7 @@ require (
 require (
 	cloud.google.com/go/compute/metadata v0.9.0 // indirect
 	github.com/Azure/azure-kusto-go v0.16.1 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2 v2.2.0 // indirect
```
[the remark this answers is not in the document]
shitaljante: Ran "go mod tidy" in test-integration/ and fixed

backend/pkg/azure/cachedreader/backend_identity_cache.go
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cachedreader
+
+// BackendIdentityAzureClients is a type that contains the Azure clients that
+// are used to interact with the Azure platform as the backend identity. The
+// backend identity is used to interact with Red Hat side Azure infrastructure.
+type BackendIdentityAzureCache struct {
```
miguelsorianod: Let's call this type `BackendIdentityAzureCachedReaders`

backend/pkg/azure/cachedreader/backend_identity_cache.go
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cachedreader
+
+// BackendIdentityAzureClients is a type that contains the Azure clients that
```
miguelsorianod: the Azure clients -> the azure cached readers

backend/pkg/app/backend.go
```diff
@@ -77,6 +78,7 @@ type BackendOptions struct {
 	MaestroSourceEnvironmentIdentifier string
 	FPAClientBuilder                   azureclient.FirstPartyApplicationClientBuilder
 	BackendIdentityAzureClients        *azureclient.BackendIdentityAzureClients
+	BackendIdentityAzureCache          *cachedreader.BackendIdentityAzureCache
```
miguelsorianod: Let's call the attribute BackendIdentityAzureCachedReaders

backend/pkg/azure/cachedreader/backend_identity_cache.go
```diff
@@ -0,0 +1,24 @@
...
+// are used to interact with the Azure platform as the backend identity. The
+// backend identity is used to interact with Red Hat side Azure infrastructure.
+type BackendIdentityAzureCache struct {
+	// RoleDefinitionsCache is the client used to get Azure RBAC role definitions
+	// (e.g. for resolving role definition IDs to allowed actions).
+	RoleDefinitionsCache CachedRoleDefinitions
```
miguelsorianod: Let's call the instance RoleDefinitionsCachedReader

backend/cmd/root.go
```diff
@@ -333,6 +333,11 @@ func (f *BackendRootCmdFlags) ToBackendOptions(ctx context.Context, cmd *cobra.C
 		return nil, utils.TrackError(fmt.Errorf("failed to create backend identity azure clients: %w", err))
 	}
 
+	backendIdentityAzureCache, err := app.NewBackendIdentityAzureCache(ctx, backendIdentityAzureClients)
```
miguelsorianod: backendIdentityAzureCache -> backendIdentityAzureCachedReaders

backend/cmd/root.go
```diff
@@ -333,6 +333,11 @@ func (f *BackendRootCmdFlags) ToBackendOptions(ctx context.Context, cmd *cobra.C
...
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to create backend identity azure cache: %w", err))
```
miguelsorianod: failed to create backend identity azure cache -> failed to create backend identity azure cached readers

backend/pkg/app/backend_identity_wiring.go
```diff
@@ -54,8 +56,24 @@ func NewBackendIdentityAzureClients(ctx context.Context, azureConfig *azureconfi
...
 		DataplaneIdentitiesOIDCConfigurationBlobStorageClient: blobStorageClient,
+		RoleDefinitionsClient: roleDefinitionsClient,
+	}
+
+	return clients, nil
+}
+
+func NewBackendIdentityAzureCache(ctx context.Context, backendIdentityClients *azureclient.BackendIdentityAzureClients) (*cachedreader.BackendIdentityAzureCache, error) {
```
miguelsorianod: NewBackendIdentityAzureCache -> NewBackendIdentityAzureCachedReaders

backend/pkg/app/backend_identity_wiring.go
```diff
@@ -54,8 +56,24 @@ func NewBackendIdentityAzureClients(ctx context.Context, azureConfig *azureconfi
...
+	roleDefinitionsCache := cachedreader.NewCachedRoleDefinitions(backendIdentityClients.RoleDefinitionsClient)
+
+	clients := &cachedreader.BackendIdentityAzureCache{
```
miguelsorianod: clients -> cachedReaders

backend/pkg/azure/cachedreader/backend_identity_cache.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: change filename to backend_identity_cached_readers.go

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,110 @@
...
+
+// roleDefinitionResourceIDCacheKeyTTL defines how long a cached role definition is considered valid.
+// After this TTL (6 hours), the cached value is considered stale and will be refreshed
+// on the next GetByID request via the Azure Role Definitions API.
+const roleDefinitionResourceIDCacheKeyTTL = 6 * time.Hour
+
+// RoleDefinitions mirrors the method set of armauthorization.RoleDefinitionsClient.
+// Implementations may add caching; callers should use a type from this package when they
+// accept cached Azure reads.
+type CachedRoleDefinitions interface {
```
miguelsorianod: CachedRoleDefinitions -> RoleDefinitionsCachedReader. This propagates to constructor, implementor type, ...

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,110 @@
...
+	azureclient "github.com/Azure/ARO-HCP/backend/pkg/azure/client"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// roleDefinitionResourceIDCacheKeyTTL defines how long a cached role definition is considered valid.
+// After this TTL (6 hours), the cached value is considered stale and will be refreshed
+// on the next GetByID request via the Azure Role Definitions API.
+const roleDefinitionResourceIDCacheKeyTTL = 6 * time.Hour
+
+// RoleDefinitions mirrors the method set of armauthorization.RoleDefinitionsClient.
```
miguelsorianod: outdated comment. Update

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// CachedRoleDefinitions implements RoleDefinitions by delegating to azureclient.RoleDefinitionsClient.
+// GetByID responses are cached in memory with a TTL and singleflight deduplication; Get and
+// NewListPager are forwarded without caching.
+type cachedRoleDefinitions struct {
+	inner azureclient.RoleDefinitionsClient
+
+	roleDefinitionsCacheLock sync.RWMutex
+	roleDefinitionsCache     map[string]cachedGetByIDResponse
```
miguelsorianod: Document the attributes. For the specific case of the cache itself, document what's the key and an example to showcase what's its format

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,110 @@
...
+// accept cached Azure reads.
+type CachedRoleDefinitions interface {
+	GetCachedByID(ctx context.Context, roleDefinitionResourceID string, options *armauthorization.RoleDefinitionsClientGetByIDOptions) (armauthorization.RoleDefinitionsClientGetByIDResponse, error)
+}
+
+// CachedRoleDefinitions implements RoleDefinitions by delegating to azureclient.RoleDefinitionsClient.
+// GetByID responses are cached in memory with a TTL and singleflight deduplication; Get and
+// NewListPager are forwarded without caching.
+type cachedRoleDefinitions struct {
+	inner azureclient.RoleDefinitionsClient
```
miguelsorianod, line 49: When documenting, mention it's an uncached client

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,110 @@
...
+
+	roleDefinitionsCacheLock sync.RWMutex
+	roleDefinitionsCache     map[string]cachedGetByIDResponse
+	sfGroup                  singleflight.Group
+}
+
+type cachedGetByIDResponse struct {
+	response   armauthorization.RoleDefinitionsClientGetByIDResponse
+	lastUpdate time.Time
```
miguelsorianod, line 66: Let's document `lastUpdate`. In what timezone is it stored?

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,110 @@
...
+	if err := c.ensureCachedGetByID(ctx, roleDefinitionResourceID, options); err != nil {
+		return armauthorization.RoleDefinitionsClientGetByIDResponse{}, err
+	}
+	c.roleDefinitionsCacheLock.RLock()
+	entry := c.roleDefinitionsCache[roleDefinitionResourceID]
+	c.roleDefinitionsCacheLock.RUnlock()
+	return entry.response, nil
+}
+
+func (c *cachedRoleDefinitions) ensureCachedGetByID(ctx context.Context, roleDefinitionResourceID string, options *armauthorization.RoleDefinitionsClientGetByIDOptions) error {
```
miguelsorianod: Let's document usage of singleFlight

backend/pkg/azure/azurehelpers/role_definitions.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"fmt"
+
+	"k8s.io/utils/set"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2"
+
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
```
miguelsorianod, line 26: Let's add unit tests for all of these

shitaljante:
@miguelsorianod updated unit test cases and documentation as per above comments. PTAL

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/stretchr/testify/require"
+
+	"k8s.io/utils/ptr"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2"
+)
+
+// fakeRoleDefinitionsClient implements azureclient.RoleDefinitionsClient for tests.
+type fakeRoleDefinitionsClient struct {
```
miguelsorianod: Instead of using this we can generate a gomock. Similar to how we generate maestro mock client (make mocks etc...)

backend/pkg/azure/azurehelpers/role_definitions_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"k8s.io/utils/ptr"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2"
+)
+
```
miguelsorianod, line 27: For each test in this file, all its subtests can be converted to table-based tests as they seem similar enough

shitaljante: looking into this

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,88 @@
...
+		RoleDefinition: armauthorization.RoleDefinition{
+			ID: ptr.To(id),
+			Properties: &armauthorization.RoleDefinitionProperties{
+				RoleName: ptr.To("Reader"),
+			},
+		},
+	}
+}
+
+func TestNewRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
```
miguelsorianod: We could be a bit better with the tests. I find these ones are missing:
* Check that once cache is expired we end up calling the real thing
* Check that if we have multiple cached entries the correct one is returned

With the current design, the role definitions cached reader is not testable on that side because of hardcoded time usage. Define an unexported attribute of type                 utilsclock.PassiveClock (k8s)  that you initialize in the constructor function, and in the tests where a fake one is needed you can pass clocktesting.NewFakePassiveClock . For tests where that's not important real time can be used using utilsclock.RealClock{}

Think also if the subtests are similar enough that can be collapsed into table-based tests.

shitaljante: is the expectation to add utilsclock.PassiveClock in the cached reader struct ?

miguelsorianod: yes, in roleDefinitionsCachedReader

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,126 @@
...
+	// It is compared with time.Since for TTL / staleness checks (same interpretation as time.Now).
+	lastUpdate time.Time
+}
+
+// NewRoleDefinitionsCachedReader returns a RoleDefinitionsCachedReader that caches GetByID results from inner.
+func NewRoleDefinitionsCachedReader(inner azureclient.RoleDefinitionsClient) RoleDefinitionsCachedReader {
+	return newRoleDefinitionsCachedReader(inner, utilsclock.RealClock{})
+}
+
+func newRoleDefinitionsCachedReader(inner azureclient.RoleDefinitionsClient, clock utilsclock.PassiveClock) *roleDefinitionsCachedReader {
```
miguelsorianod: No need for this intermediate function. Let's do it all in NewRoleDefinitionsCachedReader

backend/pkg/azure/azurehelpers/role_definitions_test.go
```diff
@@ -0,0 +1,226 @@
...
+			roleDefinition:    armauthorization.RoleDefinition{},
+			expectError:       true,
+			errorContainsText: "doesn't contain permissions",
+		},
+		{
+			name: "collects data actions",
+			roleDefinition: armauthorization.RoleDefinition{
+				Properties: &armauthorization.RoleDefinitionProperties{
+					Permissions: []*armauthorization.Permission{
+						{DataActions: []*string{ptr.To("Microsoft.Storage/storageAccounts/blobServices/containers/blobs/read")}},
```
miguelsorianod: This test is not exhaustive enough. The value of the function is that it can traverse actions cross permissions but this test only contains one permissions value. It should contain more for the test to be useful

miguelsorianod: I would even go further and also add actions cross permissions so we can know that the result doesn't contain them

backend/pkg/azure/azurehelpers/role_definitions_test.go
```diff
@@ -0,0 +1,226 @@
...
+			},
+			expectError: true,
+		},
+		{
+			name: "collects actions from all permission entries",
+			roleDefinition: armauthorization.RoleDefinition{
+				ID: ptr.To("/subscriptions/sub/resource"),
+				Properties: &armauthorization.RoleDefinitionProperties{
+					Permissions: []*armauthorization.Permission{
+						{Actions: []*string{ptr.To("Microsoft.Compute/*/read")}},
```
miguelsorianod: To make this test more useful we could also add dataactions cross permissions so we know that the result does not include them

backend/pkg/azure/azurehelpers/role_definitions_test.go
```diff
@@ -0,0 +1,226 @@
...
+				return
+			}
+
+			require.NoError(t, err)
+			assert.ElementsMatch(t, tt.expectedActions, actions)
+		})
+	}
+}
+
+func TestDataActionsFromRoleDefinition(t *testing.T) {
```
miguelsorianod, line 109: In this test we check properties is nil but not permissions is nil, however in the previous one we do. This is inconsistent. Either we deem checking both appropriate or not

backend/pkg/azure/azurehelpers/role_definitions_test.go
```diff
@@ -0,0 +1,226 @@
...
+func TestUnionActions(t *testing.T) {
+	tests := []struct {
+		name            string
+		roleDefinitions []armauthorization.RoleDefinition
+		expectedActions []string
+		expectError     bool
+	}{
+		{
+			name: "unions and deduplicates",
+			roleDefinitions: []armauthorization.RoleDefinition{
```
miguelsorianod, line 189: Similar to comments before:
* Let's have more than one permissions value in the entries
* Let's add data actions (also cross permissions) to make sure they are not included

backend/pkg/azure/azurehelpers/role_definitions_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+func TestUnionDataActions(t *testing.T) {
+	tests := []struct {
+		name                string
+		roleDefinitions     []armauthorization.RoleDefinition
+		expectedDataActions []string
+		expectError         bool
+	}{
+		{
+			name: "unions and deduplicates",
```
miguelsorianod, line 264: Similar to https://github.com/Azure/ARO-HCP/pull/4403/changes#r3208085822

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,162 @@
...
+		name string
+		run  func(t *testing.T, ctx context.Context, rid string, otherRID string)
+	}{
+		{
+			name: "caches successful GetByID",
+			run: func(t *testing.T, ctx context.Context, rid string, _ string) {
+				ctrl := gomock.NewController(t)
+				want := roleDefByIDResponse(rid)
+				mockClient := azureclient.NewMockRoleDefinitionsClient(ctrl)
+				mockClient.EXPECT().GetByID(gomock.Any(), rid, nil).Return(want, nil).Times(1)
```
miguelsorianod: Let's add a comment here that we verify by making sure that the call to the Azure API is only performed once. If that's not the case then the test would fail because of the mocking expectation erroring

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+				for range 2 {
+					got, err := r.GetCachedByID(ctx, rid, nil)
+					require.NoError(t, err)
+					assert.Equal(t, want, got)
+				}
+			},
+		},
+		{
+			name: "propagates inner error",
```
miguelsorianod: The name of this test is not very clear. "propagates azure api interaction error when it occurs"

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,162 @@
...
+				mockClient.EXPECT().GetByID(gomock.Any(), rid, nil).Return(initial, nil).Times(1)
+				mockClient.EXPECT().GetByID(gomock.Any(), rid, nil).Return(refreshed, nil).Times(1)
+
+				start := time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)
+				fakeClock := clocktesting.NewFakePassiveClock(start)
+				r := newRoleDefinitionsCachedReader(mockClient, fakeClock)
+
+				got, err := r.GetCachedByID(ctx, rid, nil)
+				require.NoError(t, err)
+				assert.Equal(t, initial, got)
```
miguelsorianod: Is this really doing what we think it's doing? the contents of both initial and refreshed are the same. What is it being compared?

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+				fakeClock.SetTime(start.Add(roleDefinitionResourceIDCacheKeyTTL + time.Second))
+
+				got, err = r.GetCachedByID(ctx, rid, nil)
+				require.NoError(t, err)
+				assert.Equal(t, refreshed, got)
```
miguelsorianod: Same question as https://github.com/Azure/ARO-HCP/pull/4403/changes#r3208121225

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func TestNewRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
+	ctx := context.Background()
+	rid := "/subscriptions/11111111-1111-1111-1111-111111111111/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c"
+	otherRID := "/subscriptions/11111111-1111-1111-1111-111111111111/providers/Microsoft.Authorization/roleDefinitions/acdd72a7-3385-48ef-bd42-f606fba81ae7"
+
+	tests := []struct {
+		name string
+		run  func(t *testing.T, ctx context.Context, rid string, otherRID string)
```
miguelsorianod: This way of designing this table based tests defeats its purpose: if every test needs of its own very specific test logic there's no value on the table based test. Compare it with the table based tests of role definitions helpers

backend/pkg/app/backend_identity_wiring.go
```diff
@@ -54,9 +58,28 @@ func NewBackendIdentityAzureClients(ctx context.Context, azureConfig *azureconfi
...
+		backendIdentityClients.RoleDefinitionsClient,
+		utilsclock.RealClock{},
+	)
+
+	cachedReaders := &cachedreader.BackendIdentityAzureCachedReaders{
+		RoleDefinitionsCachedReader: roleDefinitionsCachedReader,
+	}
+
+	return cachedReaders, nil
+}
```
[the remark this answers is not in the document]
miguelsorianod: Modify NewBackendIdentityAzureCachedReaders to not return an error

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -0,0 +1,123 @@
...
+type cachedGetByIDResponse struct {
+	response armauthorization.RoleDefinitionsClientGetByIDResponse
+	// lastUpdate is the wall-clock instant (UTC) when this cache entry was written, from time.Now().UTC().
+	// It is compared with time.Since for TTL / staleness checks (same interpretation as time.Now).
+	lastUpdate time.Time
+}
+
+// NewRoleDefinitionsCachedReader returns a RoleDefinitionsCachedReader that caches GetByID results from inner.
+// clock controls cache staleness and entry timestamps (use utilsclock.RealClock{} in production).
+func NewRoleDefinitionsCachedReader(inner azureclient.RoleDefinitionsClient, clock utilsclock.PassiveClock) RoleDefinitionsCachedReader {
```
miguelsorianod: Do not make clock configurable in the constructor. Initialize it directly inside.
For tests the struct can be initialized directly without constructor and the appropriate desired clock

backend/pkg/azure/azurehelpers/role_definitions_test.go
miguelsorianod: It's better but i still miss multiple actions within a permission

backend/pkg/azure/azurehelpers/role_definitions_test.go
miguelsorianod: This has not been updated. It still has a single permission.

It's also missing the same as https://github.com/Azure/ARO-HCP/pull/4403#discussion_r3266351463

backend/pkg/azure/azurehelpers/role_definitions_test.go
miguelsorianod: Here I still miss that actions/dataactions only have a single element

backend/pkg/azure/azurehelpers/role_definitions_test.go
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/4403#discussion_r3266370471

backend/pkg/azure/cachedreader/role_definitions_test.go
miguelsorianod: This has not been addressed

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,202 @@
...
+		RoleDefinition: armauthorization.RoleDefinition{
+			ID: ptr.To(id),
+			Properties: &armauthorization.RoleDefinitionProperties{
+				RoleName: ptr.To("Reader"),
+			},
+		},
+	}
+}
+
+func TestNewRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
```
miguelsorianod: This seems overly complicated.

I would expect the organization to be something like this:

* Struct of test cases. It has the following attrs:
  * azure client factory function (we set up mocks there etc..) (receives gomock controller)
  * clock
  * firstCall
    * expectedResult
    * expectedErr
    * clock
  * *secondCall (can be nil)
    * expectedResult
    * expectedErr
    * advanceClockBy (duration)

or similar

Additionally, no need of its own defined struct type as its own type. See other tests where there are test cases where there's no dedicated type defined at the package level. Also let's not separate the test logic into its own function separated from the test function.

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,202 @@
...
+	if tt.useFakeClock {
+		start := time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)
+		fakeClock = clocktesting.NewFakePassiveClock(start)
+		clock = fakeClock
+	}
+
+	r := NewRoleDefinitionsCachedReader(mockClient, clock)
+
+	for i, step := range tt.getCachedByIDSteps {
+		if fakeClock != nil && i > 0 && tt.advanceClockAfter == i-1 {
```
miguelsorianod: This is not good because it is coupling tests to each others. Tests should be fully independent to each other.

See for a proposed design https://github.com/Azure/ARO-HCP/pull/4403/changes#r3266689127

backend/pkg/azure/cachedreader/role_definitions_test.go
shitaljante: updated

backend/pkg/azure/cachedreader/role_definitions_test.go
shitaljante: Thanks for feedback , kept the test case struct local to the test, removed the test logic from helper function and consolidated these tests in table based.

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	tests := []struct {
+		name        string
+		setupClient func(ctrl *gomock.Controller) azureclient.RoleDefinitionsClient
+		clock       utilsclock.PassiveClock
+		firstCall   struct {
+			roleDefinitionID string
+			wantResponse     armauthorization.RoleDefinitionsClientGetByIDResponse
+			wantError        bool
+			wantErrIs        error
```
miguelsorianod: wantErrIs seems unnecessary

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,305 @@
...
+			wantErrContains  string
+		}
+		secondCall *struct {
```
miguelsorianod: At this point, what's the differece between the different calls? let's have a unified array of call expectations and that's it. The advance clock would be optional

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,305 @@
...
+					if call.wantErrIs != nil {
+						assert.ErrorIs(t, err, call.wantErrIs)
+					}
+					if call.wantErrContains != "" {
+						assert.ErrorContains(t, err, call.wantErrContains)
+					}
+					return
+				}
+				require.NoError(t, err)
+				assert.Equal(t, call.wantResponse, got)
```
miguelsorianod, line 237: What does this compare exactly? (pointers, values, deep, shallow, ...) how do we test that it's really the cached entry?

backend/pkg/azure/cachedreader/role_definitions_test.go
shitaljante: actually the contents of initial and refreshed are different , initial contains the Reader role name, while refreshed contains Contributor.

backend/pkg/azure/cachedreader/role_definitions_test.go
shitaljante: Addressed in the latest role_definitions_test.go refactor, I can re-check if anything’s still unclear.

backend/pkg/azure/cachedreader/role_definitions_test.go
shitaljante: We are doing a deep compare on the struct(pointer fields as well) , but that alone does not prove caching. For the cache case, we rely on gomock with Times(1) that means the second GetCachedByID call should not make another Azure API call.

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,241 @@
...
+				advanceClockBy   time.Duration
+				roleDefinitionID string
+				wantResponse     armauthorization.RoleDefinitionsClientGetByIDResponse
+				wantError        bool
+				wantErrContains  string
+			}{
+				{
+					roleDefinitionID: testRoleDefinitionRID,
+					wantError:        true,
+					wantErrContains:  testRoleDefinitionRID,
```
miguelsorianod: Checking for the resourceid in the error is a bit odd. I would checking the part that helps identify that particular error (that doesn't change)

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+	azureclient "github.com/Azure/ARO-HCP/backend/pkg/azure/client"
+)
+
+const (
+	testRoleDefinitionRID      = "/subscriptions/11111111-1111-1111-1111-111111111111/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c"
+	testOtherRoleDefinitionRID = "/subscriptions/11111111-1111-1111-1111-111111111111/providers/Microsoft.Authorization/roleDefinitions/acdd72a7-3385-48ef-bd42-f606fba81ae7"
+)
+
+func roleDefByIDResponse(id string) armauthorization.RoleDefinitionsClientGetByIDResponse {
```
miguelsorianod: No need to make this a package level function. It can be inside the test

backend/pkg/azure/cachedreader/role_definitions_test.go
shitaljante: Switched to asserting error message as "failed to get role definition"

backend/pkg/azure/cachedreader/role_definitions_test.go
shitaljante: Moved roleDefByIDResponse into the test

PR #4544, opened 2026-03-19
Add metrics for storage access to catch hotloops

deads2k:
AI generated. Add metrics for cosmosClient access per resource.

### DIFF
--- a/internal/database/metrics.go
+++ b/internal/database/metrics.go
@@ -1,0 +1,57 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


internal/database/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	operationLabelCreate  = "create"
+	operationLabelReplace = "replace"
+	operationLabelDelete  = "delete"
+	operationLabelGet     = "get"
+	operationLabelList    = "list"
+)
+
+var (
+	// DatabaseOperationTotal counts the total number of database operations attempted.
+	DatabaseOperationTotal = promauto.NewCounterVec(
```
geoberle, line 32: in backend we expose metrics from the component-base legacyregistry
but i don't think we do that for the frontend already

fix - https://github.com/Azure/ARO-HCP/pull/5052

geoberle:
superseded by https://github.com/Azure/ARO-HCP/pull/6929

PR #4563, opened 2026-03-20
fix: rename tests correctly

jharrington22:
Rename tests correctly


deads2k:
will probably fail units.

mbukatov (changes requested):
These are [test cases using per-run cluster](https://github.com/Azure/ARO-HCP/blob/main/test/e2e/README.md#test-cases-with-per-run-cluster), are not part of e2e-paralles suite and the `_test.go` suffix is expected. See also [File Structure](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md#file-structure).
 
This means that these files should not be renamed.

That said, we likely want to rename another file(s): `nodepool_ephemeral_osdisk_test.go`, which already is part of [#4473](https://github.com/Azure/ARO-HCP/pull/4473/changes#diff-15416023b601968e3bf62421e29a458376d5150e4502631533be58a01926865e) PR.

/hold

raelga (commented):
Superseeded by https://github.com/Azure/ARO-HCP/pull/4473/

/close

PR #4599, opened 2026-03-24
feat: add nodepool scoped maestro readonly bundles controllers

miguelsorianod:
We add nodepool-scoped controllers in backend that are able to manage maestro readonly bundles.
With it, we introduce the ability to read the Hypershift's NodePool CRs associated to the node pools and persist it in Cosmos.

The Cosmos ManagementClusterContent type is a direct child of the node pool Cosmos type. Documents within the ManagementClusterContent Cosmos type are different readonly bundles internal references. A new ManagementClusterContent document per each Node Pool of the cluster is added. The maestro bundle internal name references associated to the hypershift nodepools is `readonlyHypershiftNodePool`.


backend/pkg/maestro/maestro_api_maestro_bundle_name_generator.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/google/uuid"
+
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// MaestroAPIMaestroBundleNameGenerator is an interface that defines a method to generate a new Maestro API Maestro Bundle name.
+// The generated name must be globally unique within a given Maestro Consumer Name and Maestro Source ID.
+// It can be used to generate a new Maestro API Maestro Bundle name for a new Maestro Bundle reference.
+type MaestroAPIMaestroBundleNameGenerator interface {
```
deads2k: Don't like this indirection and passing into a controller.  The controller can have a private field for choosing a name so you can more easily write a unit test, but it should never be exposed.

miguelsorianod:
I pushed a commit to rebase from main resolving conflicts.

I haven't merged this MR yet even though it was already approved because we found a race condition in the maestro readonly bundle delete orphan controller that could occur in some scenarios and I want the fix to land before introducing this. The PR with the race condition fix is https://github.com/Azure/ARO-HCP/pull/4856, and once it is merged I will rebase this PR with those changes.

miguelsorianod:
/hold cancel

I rebased the PR from main which contains the race condition mentioned in https://github.com/Azure/ARO-HCP/pull/4599#issuecomment-4236887333 (https://github.com/Azure/ARO-HCP/pull/4856).

> However, with the refactor it doesn't seem possible to do further filtering of received events anymore, which means that now the controllers receive managementclustercontent events but they cannot filter further to only process cluster-scoped or nodepool-scoped ones. We need to decide if there are more changes needed around this

About this. What we ended up doing is adding the ability to register an informer and deciding the max parent depth to look for when filtering the resources.

miguelsorianod:
I added a new commit that updates the orphan controller to take into account that eventually the CS ID won't always be set when controllers run, as we will move the creation of resources against CS in the backend and not in frontend.

Details here on what's been done:

We do not want to make the whole sync fail when there are some cosmos
resources that do not have the ClusterServiceID attribute set yet. This
could be a common occurrence when there's enough amount of creation
activity of resources.

It should be safe to skip those because if a maestro bundle in the
maestro API exists it means that there should be a corresponding Cosmos resource
with a maestro bundle reference. If for some reason during the orphan calculation inbetween the first read of cosmos
resources and the read in the Maestro api there's a new bundle in maestro, the
second read of cosmos resources will catch that and prevent accidental deletion.

miguelsorianod:
I also added another commit that deals with the same scenario of CSID not being set for both the create and read maestro readonly bundles.
For the cluster-scoped maestro controllers, that is being added in https://github.com/Azure/ARO-HCP/pull/4752

PR #4610, opened 2026-03-24
unrevert of read from cosmos

deads2k:
/hold

need to add fixes from https://github.com/Azure/ARO-HCP/pull/4412


machi1990:
@deads2k The way https://github.com/Azure/ARO-HCP/blob/428d5acda06ee272961298e9e99621f21bd67e05/internal/ocm/convert.go#L311 is written will get us in a spot where we still can set identity with empty clientId and principalId

This https://github.com/Azure/ARO-HCP/blob/428d5acda06ee272961298e9e99621f21bd67e05/internal/ocm/convert.go#L318 needs to 
be something like 

`clientID, ok := operatorIdentity.GetClientID()` and we only set the cliendID to the returned response when ok is true.

We also need to ensure that the migration identity controller can handle the case of only some values being partially set https://github.com/Azure/ARO-HCP/blob/428d5acda06ee272961298e9e99621f21bd67e05/backend/pkg/controllers/clusterpropertiescontroller/identity_migration.go#L80 and resync anything.

i.e NeedsWork re-written in a way that it says work is needed if any of the identity elements have a nil clientId or principalId

internal/api/types_cluster.go
```diff
@@ -295,6 +295,44 @@ func (cluster *HCPOpenShiftCluster) EnsureDefaults() {
...
+			cluster.Identity = &arm.ManagedServiceIdentity{}
+		}
+		if cluster.Identity.UserAssignedIdentities == nil {
+			cluster.Identity.UserAssignedIdentities = make(map[string]*arm.UserAssignedIdentity)
+		}
+
+		if _, ok := cluster.Identity.UserAssignedIdentities[operatorIdentityResourceID.String()]; !ok {
+			cluster.Identity.UserAssignedIdentities[operatorIdentityResourceID.String()] = &arm.UserAssignedIdentity{}
+		}
+	}
```
machi1990: It's not needed for data plane; only ControlPlane + SMI identities should be in the cluster.Identity. S
Something like https://github.com/Azure/ARO-HCP/blob/978e85c5a789ab5e5c20b41320b48f2e30210ff7/demo/bicep/cluster.bicep#L574 is what the user will provide which contains only the SMI + CP ones as those are treated differently from the DP ones.
For DP we don't request credentials directly from the MI RP but rather perform federations on them so they don't need to be in this list. However for the CP + SMI, do interact with the MI RP and hence the need for the .Identity stanza which makes ARM give the FPA credential ability to retrieve credentials for these identities from the MI RP

frontend/pkg/frontend/cluster.go
```diff
@@ -257,6 +223,10 @@ func decodeDesiredClusterCreate(ctx context.Context, azureLocation string, reque
 	if err != nil {
 		return nil, utils.TrackError(err)
 	}
+
+	// Clear the user-assigned identities map since that is built by a controller. The defaults will be set next and valid until we know the actual values.
+	newInternalCluster.Identity.UserAssignedIdentities = nil
```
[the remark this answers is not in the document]
deads2k: Ok, I think the summary of all that is, "we must validate the user input as provided. we must return frontend values with all the keys and either accurate or empty values."  Doable.  Will update.

machi1990 (changes requested):
Took another round of reviews. 
Just 1 one comment and bumping 2 old  related comments that I think will be critical to address

machi1990 (approved):
/lgtm

machi1990:
/lgtm

Doing it again since the e2e failure pointed out in https://github.com/Azure/ARO-HCP/pull/4610#discussion_r3027679955 is gauged as flake and not related to the PR. 
However, there are some prow job failing if a fix to them removes the lgtm label, it can be re-added

test-integration/frontend/artifacts/FrontendCRUD/Cluster/patch-identity/05-httpGet-cluster/cluster.json
machi1990, line 1: This is failing the formatting validation
```
 		},
 		"provisioningState": "Accepted",
 		"version": {
-			"id": "4.20",
-			"channelGroup": "stable"
+			"channelGroup": "stable",
+			"id": "4.20"
 		}
 	},
-	"type": "Microsoft.RedHatOpenShift/hcpOpenShiftClusters",
 	"systemData": {
 		"createdBy": "Unknown-ARO-HCP-frontend",
 		"createdByType": "Application",
 		"lastModifiedBy": "Unknown-ARO-HCP-frontend",
 		"lastModifiedByType": "Application"
-	}
+	},
+	"type": "Microsoft.RedHatOpenShift/hcpOpenShiftClusters"
 }
```

deads2k:
/hold

need to first update the cluster create operation to require the URL.


found in error
>{fail [github.com/Azure/ARO-HCP/test/e2e/cluster_authorized_cidrs_connectivity.go:145]: cluster Properties.API.URL was nil
Expected
    <*string | 0x0>: ...
not to be nil  fail [github.com/Azure/ARO-HCP/test/e2e/cluster_authorized_cidrs_connectivity.go:145]: cluster Properties.API.URL was nil
Expected
    <*string | 0x0>: ...
not to be nil}

deads2k:
https://github.com/Azure/ARO-HCP/pull/4695 must get to prod before we can merge this.

deads2k:
precondition met

/hold cancel

internal/ocm/convert.go
```diff
@@ -503,258 +331,37 @@ func convertCIDRBlockAllowAccessRPToCS(in api.CustomerAPIProfile) (*arohcpv1alph
...
-				if internalCluster.Identity.UserAssignedIdentities == nil {
-					internalCluster.Identity.UserAssignedIdentities = make(map[string]*arm.UserAssignedIdentity)
-				}
-
 				clientID, _ := operatorIdentity.GetClientID()
 				principalID, _ := operatorIdentity.GetPrincipalID()
-				internalCluster.Identity.UserAssignedIdentities[operatorIdentity.ResourceID()] = &arm.UserAssignedIdentity{
-					ClientID:    &clientID,
-					PrincipalID: &principalID,
+				if len(clientID) > 0 && len(principalID) > 0 {
```
miguelsorianod, line 344: I am unsure that this is correct: doesn't that mean that if from CS we get the identity but it hasn't calculated the clientid/principalid yet we would unset that identity fully in the  identity migration controller because it's not returned as a result from here? Shouldn't we simply return an entry that has empty clientid and principalid?

internal/ocm/convert.go
```diff
@@ -503,258 +331,37 @@ func convertCIDRBlockAllowAccessRPToCS(in api.CustomerAPIProfile) (*arohcpv1alph
...
-				if internalCluster.Identity.UserAssignedIdentities == nil {
-					internalCluster.Identity.UserAssignedIdentities = make(map[string]*arm.UserAssignedIdentity)
-				}
-
 				clientID, _ := mi.ServiceManagedIdentity().GetClientID()
 				principalID, _ := mi.ServiceManagedIdentity().GetPrincipalID()
-				internalCluster.Identity.UserAssignedIdentities[mi.ServiceManagedIdentity().ResourceID()] = &arm.UserAssignedIdentity{
-					ClientID:    &clientID,
-					PrincipalID: &principalID,
+				if len(clientID) > 0 && len(principalID) > 0 {
```
miguelsorianod, line 356: Same comment as https://github.com/Azure/ARO-HCP/pull/4610/changes#r3180404768

frontend/pkg/frontend/cluster.go
```diff
@@ -344,9 +302,13 @@ func (f *Frontend) createHCPCluster(writer http.ResponseWriter, request *http.Re
 		return utils.TrackError(err)
 	}
 
-	// Now that validation is done we clear the user-assigned identities map since that is reconstructed from Cluster Service data
-	// TODO this is bad, see above TODOs. We want to validate what we store.
-	newInternalCluster.Identity.UserAssignedIdentities = nil
+	// we must validate using user provided .Identity.UserAssignedIdentities because that is the intent expressed by the user to allow
+	// us to use these identities. The information contained in those key is not trusted to be accurate, so we clear this field and set to
```
miguelsorianod, line 305: Could you expand on this? I thought the end-user needs to provide both .Identity.UserAssignedIdentities (as it's a requirement by ARM) as well as the .platform.operatorsAuthentication.* data and we would just limit what we do to validate all the expected data is there and consistent between them instead of modifying .Identity with something different than they have provided.

internal/ocm/convert.go
deads2k: we separately set that value for REST calls in `completeClusterIdentity`.  This only returns complete information for storage in cosmos.

frontend/pkg/frontend/cluster.go
deads2k: We validate the identity in the call above.  We clear the input before storage and set it to an empty valid value.

internal/ocm/convert.go
miguelsorianod: With the latest commit this has been fixed

internal/ocm/convert.go
miguelsorianod: With the latest commit this has been fixed

internal/ocm/convert.go
```diff
@@ -503,258 +331,41 @@ func convertCIDRBlockAllowAccessRPToCS(in api.CustomerAPIProfile) (*arohcpv1alph
 	return arohcpv1alpha1.NewCIDRBlockAccess().Allow(cidrBlockAllowAccess), nil
 }
 
-// LegacyCreateInternalClusterFromClusterService this exists only for clusters that were created before we held all
-// customer desired state in cosmos.
-func LegacyCreateInternalClusterFromClusterService(resourceID *azcorearm.ResourceID, azureLocation string, cluster *arohcpv1alpha1.Cluster) (*api.HCPOpenShiftCluster, error) {
```
miguelsorianod, line 508: Do we want to remove this and all the methods it calls now? although it's not being used anymore because of us stop calling it from Frontend and the removal of the the cluster customer properties migration controller, wouldn't it be possible that in the future we end up with the desire/need to convert a CS Cluster to a RP Cluster type in some controller that we end up adding? asking given that it's a non trivial amount of code.

deads2k: We'll never add it back. This is the moment we switch

miguelsorianod: In Frontend yes. But my comment is about new backend controllers we add where we get the data from CS

miguelsorianod: As discussed through other channels: in the case of backend controllers that need the data from CS, we don't expect to need to use the entire type, rather more concrete specific fields. If for some reason it turns out that we end up needing the whole type or a big part we should remember we could get it from git history.

PR #4611, opened 2026-03-24
E2e add poll timeout to clusters_sharing_resgroup PollUntilDone calls

twolff-gh:
### What

Add 45-minute timeout to cluster creation polling in e2e test.
Prevents indefinite polling in the clusters_sharing_resgroup e2e test by using a context with a 45-minute timeout for PollUntilDone calls.

### Why

The PollUntilDone calls in the e2e tests have no context timeout.
When cluster provisioning stalls for any reason, the test blocks indefinitely until CI's external hard kill terminates the process before it can write output, producing a misleading "deserialization error" instead of a clear timeout message.


bennerv (approved):
/lgtm

PR #4673, opened 2026-03-28
fix: error failing to get node count

jharrington22:
Failing to get node count will silently error.


raelga (approved):
/restest
/lgtm
/approve

PR #4682, opened 2026-03-30
Export image tags for prow image mirror

inbharajmani:
https://redhat.atlassian.net/browse/AROSLSRE-523

### What

Export image registry variables so that prow job can use it to push the images using `oc image mirror <src> <dest>`

### Why

Podman build and push is not supported in prow. Hence we need to rely on the images which are built by ci-operator and use `oc image mirror` to push images to ACR. For which prow job needs the acr image registry and image name with tag

### Testing

Tested locally by running `make export-image-vars`

Example usage in prow: https://github.com/openshift/release/pull/76898

### DIFF
--- a/admin/Makefile
+++ b/admin/Makefile
@@ -78,6 +78,14 @@ build-and-push: image $(ORAS)
 	$(CONTAINER_ENGINE) push ${ADMIN_API_TAGGED_IMAGE}
 .PHONY: build-and-push
 
+export-image-vars:
+ifndef IMAGE_VARS_FILE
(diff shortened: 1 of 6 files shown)


raelga (approved):
/lgtm

inbharajmani:
/hold 

Wont be needed since image names are already rendered using shell script in the prow job https://github.com/openshift/release/pull/76898/changes#diff-7f136228cc04b636cf48ae6862db75ea19da3ca1258b402fcc4220e4f45cd09e

PR #4688, opened 2026-03-30
[FSI] Override HyperShift shared ingress HAProxy image to pull from ACR

avollmer-redhat:
## Summary

- Override the HyperShift shared ingress HAProxy image to pull from ACR via `--additional-operator-env-vars IMAGE_SHARED_INGRESS_HAPROXY=...` instead of the hardcoded `quay.io` default, for FSI/SFI compliance
- Use ORAS on-demand mirror (ImageMirror pipeline step) instead of pull-through cache, consistent with how all other service images are sourced

## Context

The upstream HyperShift `GetSharedIngressHAProxyImage()` in `support/images/envvars.go` reads the `IMAGE_SHARED_INGRESS_HAPROXY` env var directly and falls back to a hardcoded `quay.io/redhat-user-workloads/...` default. This code path does **not** go through `--registry-overrides` string substitution, so the only way to override it is via the env var.

### Changes

- `config/config.yaml` — `sharedIngressImage` block under `hypershift:` with `registry`, `repository`, and `digest` fields for ORAS mirroring
- `config/config.schema.json` — `sharedIngressImage` property added to `hypershift` definition (optional, references `containerImage`)
- `hypershiftoperator/pipeline.yaml` — `mirror-shared-ingress-image` ImageMirror step added (copies image from quay.io to SVC ACR at deploy time), deploy step depends on both mirror steps
- `hypershiftoperator/values.yaml` — `sharedIngressImage` templatized value under `operatorEnvVars`
- `hypershiftoperator/deploy/templates/installer.job.yaml` — conditional `--additional-operator-env-vars IMAGE_SHARED_INGRESS_HAPROXY=...` block
- `tooling/image-updater/config.yaml` — `hypershift-shared-ingress` entry for automated digest updates

### Why ORAS mirror instead of pull-through cache?

Pull-through cache (`quay-cache/...` prefix) has drawbacks for FSI compliance:
- Cached images can be evicted by the ACR retention policy
- Cache availability depends on upstream registry being reachable at pull time
- ORAS mirror copies the image into the ACR as a first-class artifact, pinned by digest, at deploy time — the same mechanism used for 22 other service images

### Follow-up required

A corresponding change in `sdp-pipelines/hcp/` is needed to add the `sharedIngressImage` field to the higher-environment config and schema (which has `additionalProperties: false`).

## Related

- JIRA: ARO-22152, [ARO-24834](https://redhat.atlassian.net/browse/ARO-24834)
- PR #4687 — ACR pull-through cache rules (this PR no longer depends on `quay-redhat-user-workloads` cache rule)

## Test plan

- [x] `make -C config materialize` passes (schema validation, config render, 42 helm template tests, 11 ACR value tests)
- [x] Fixture shows `--additional-operator-env-vars IMAGE_SHARED_INGRESS_HAPROXY=arohcpsvcdev.azurecr.io/redhat-user-workloads/...` in rendered output
- [x] ORAS mirror step copies image from `quay.io` to `arohcpsvcdev.azurecr.io/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main@sha256:1af59b7a...`
- [x] Deploy to personal dev — hypershift operator env var `IMAGE_SHARED_INGRESS_HAPROXY` set correctly
- [x] Shared ingress router pods (`hypershift-sharedingress` namespace) use ACR-mirrored image path (not `quay-cache/`)
- [x] Image registry policy E2E tests pass on both SVC (3/3) and MGMT (3/3) clusters

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -287,6 +287,9 @@ defaults:
     # By default we set it to empty as this tag is only required for msft environments. We override
     # this value in sdp pipelines for msft environments.
     sharedIngressIPTag: ''
+    sharedIngressImage:
+      repository: quay-cache/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
+      digest: sha256:1af59b7a29432314bde54e8977fa45fa92dc48885efbf0df601418ec0912f472
     limitClusterSizes: false
   # Log settings
   logs:
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -288,7 +288,8 @@ defaults:
     # this value in sdp pipelines for msft environments.
     sharedIngressIPTag: ''
     sharedIngressImage:
-      repository: quay-cache/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
+      registry: quay.io
+      repository: redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
       digest: sha256:1af59b7a29432314bde54e8977fa45fa92dc48885efbf0df601418ec0912f472
     limitClusterSizes: false
   # Log settings
--- a/config/rendered/dev/cspr/westus3.yaml
+++ b/config/rendered/dev/cspr/westus3.yaml
@@ -411,7 +411,8 @@ hypershift:
   sharedIngressIPTag: ""
   sharedIngressImage:
     digest: sha256:1af59b7a29432314bde54e8977fa45fa92dc48885efbf0df601418ec0912f472
-    repository: quay-cache/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
+    registry: quay.io
+    repository: redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
 imageSync:
   environmentName: aro-hcp-image-sync
   ocMirror:
--- a/config/rendered/dev/cspr/westus3.yaml
+++ b/config/rendered/dev/cspr/westus3.yaml
@@ -409,6 +409,9 @@ hypershift:
   limitClusterSizes: true
   namespace: hypershift
   sharedIngressIPTag: ""
+  sharedIngressImage:
+    digest: sha256:1af59b7a29432314bde54e8977fa45fa92dc48885efbf0df601418ec0912f472
+    repository: quay-cache/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
 imageSync:
   environmentName: aro-hcp-image-sync
   ocMirror:
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -2222,6 +2222,9 @@
         },
         "limitClusterSizes": {
           "type": "boolean"
+        },
+        "sharedIngressImage": {
+          "$ref": "#/definitions/containerImage"
         }
       },
       "additionalProperties": false,
--- a/config/rendered/dev/dev/westus3.yaml
+++ b/config/rendered/dev/dev/westus3.yaml
@@ -411,7 +411,8 @@ hypershift:
   sharedIngressIPTag: ""
   sharedIngressImage:
     digest: sha256:1af59b7a29432314bde54e8977fa45fa92dc48885efbf0df601418ec0912f472
-    repository: quay-cache/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
+    registry: quay.io
+    repository: redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
 imageSync:
   environmentName: aro-hcp-image-sync
   ocMirror:
--- a/config/rendered/dev/dev/westus3.yaml
+++ b/config/rendered/dev/dev/westus3.yaml
@@ -409,6 +409,9 @@ hypershift:
   limitClusterSizes: true
   namespace: hypershift
   sharedIngressIPTag: ""
+  sharedIngressImage:
+    digest: sha256:1af59b7a29432314bde54e8977fa45fa92dc48885efbf0df601418ec0912f472
+    repository: quay-cache/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
 imageSync:
   environmentName: aro-hcp-image-sync
   ocMirror:
--- a/config/rendered/dev/perf/westus3.yaml
+++ b/config/rendered/dev/perf/westus3.yaml
@@ -411,7 +411,8 @@ hypershift:
   sharedIngressIPTag: ""
   sharedIngressImage:
     digest: sha256:1af59b7a29432314bde54e8977fa45fa92dc48885efbf0df601418ec0912f472
-    repository: quay-cache/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
+    registry: quay.io
+    repository: redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
 imageSync:
   environmentName: aro-hcp-image-sync
   ocMirror:
--- a/config/rendered/dev/perf/westus3.yaml
+++ b/config/rendered/dev/perf/westus3.yaml
@@ -409,6 +409,9 @@ hypershift:
   limitClusterSizes: true
   namespace: hypershift
   sharedIngressIPTag: ""
+  sharedIngressImage:
+    digest: sha256:1af59b7a29432314bde54e8977fa45fa92dc48885efbf0df601418ec0912f472
+    repository: quay-cache/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main
 imageSync:
   environmentName: aro-hcp-image-sync
   ocMirror:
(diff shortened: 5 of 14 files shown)


avollmer-redhat:
AI generated review: 

## Review: PR #4688 — Override HyperShift shared ingress HAProxy image to pull from ACR

  ### Overall

  Clean, well-scoped change. The ORAS ImageMirror approach is consistent with how 22+ other
service images are sourced and is the right call over pull-through cache for this use case.
The PR description is excellent — thorough context, clear rationale for ORAS vs. cache, and
honest about follow-up work needed.

  LGTM with a few observations below (none blocking).

  ### Observations

  **1. `--additional-operator-env-vars` used twice (non-issue, just documenting)**

  The installer template renders two separate `--additional-operator-env-vars` flags when
both `sharedIngressIPTag` and `sharedIngressImage` are set. I initially thought this would
cause the second invocation to overwrite the first, but pflag's `StringToStringVar.Set()`
uses a `changed` bool — the first call replaces the default empty map, and subsequent calls
**merge** into it. So this works correctly.

  That said, in dev environments `sharedIngressIPTag` is always empty, so CI only exercises
the single-flag path. Worth a manual sanity check in a Microsoft environment (int) where
both conditionals render, to confirm the merge behavior works end-to-end in a real
`hypershift install` invocation.

  **2. `acrDNSSuffix` vs hardcoded `.azurecr.io` — minor inconsistency (not introduced by
this PR)**

  The new `sharedIngressImage` in `values.yaml` correctly uses `{{ .acr.svc.name }}.{{ .
acrDNSSuffix }}` for sovereign cloud portability. The existing `image` and
`registryOverrides` on lines 5 and 7 still hardcode `.azurecr.io`. Not something to fix in
this PR, but worth a follow-up to align the existing fields if sovereign cloud support is
on the roadmap.

  **3. Schema — `sharedIngressImage` is optional (correct)**

  The field is not in the `required` array for the hypershift definition, which is correct
for backwards compatibility. I confirmed the sdp-pipelines schema at `hcp/config.schema.
json:2256` already has `sharedIngressImage` declared under `properties`, so it won't be
rejected by `additionalProperties: false`. The follow-up is just adding the actual image
values in the int/stg/prod config files.

  **4. Image-updater `tagPattern` for shared ingress**

  ```yaml
  tagPattern: "^[a-f0-9]{40}$"

This matches 40-char hex strings (git commit SHAs), which is the tagging convention for
redhat-user-workloads  builds. Looks correct and consistent with how the upstream image is
tagged.

### Summary

• Architecture: Sound — ORAS mirror at deploy time gives deterministic, eviction-proof
image availability
• Config/schema: Correct —  containerImage  type, digest format with  sha256:  prefix,
optional field
• Pipeline: Correct — mirror step + dependency wiring matches existing  mirror-image
pattern exactly
• Helm template: Correct — conditional block, env var override via  --additional-operator-
env-vars
• Image-updater: Correct — proper group, source, tagPattern, jsonPath target

/lgtm

config/rendered/dev/cspr/westus3.yaml
```diff
@@ -418,6 +418,10 @@ hypershift:
   limitClusterSizes: true
   namespace: hypershift
   sharedIngressIPTag: ""
```
hbhushan3, line 411: This should be in schema too right? `sharedIngressIPTag`

config/config.yaml
```diff
@@ -284,6 +284,10 @@ defaults:
     # By default we set it to empty as this tag is only required for msft environments. We override
     # this value in sdp pipelines for msft environments.
```
hbhushan3, line 286: Do we have a PR or JIRA card for the change in sdp-pipelines ready?

avollmer-redhat:
The sdp-pipelines PR has been created for this: [fsi-hypershift-shared-ingress-image](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/sdp-pipelines/pullrequestcreate?sourceRef=fsi-hypershift-shared-ingress-image&targetRef=main)

It adds `sharedIngressImage` to `hcp/config.schema.json`, `hcp/config.yaml`, and `hcp/config.publish.yaml`. The sdp-pipelines PR should merge **before** this one gets bumped in, since `values.yaml` references `{{ .hypershift.sharedIngressImage.repository }}` unconditionally and the templatize engine uses `missingkey=error`.

Related: [ARO-24834](https://redhat.atlassian.net/browse/ARO-24834)

avollmer-redhat:
Correction — the sdp-pipelines PR link: [PR 15403118](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/sdp-pipelines/pullrequest/15403118)

raelga (approved):
/lgtm
/approve

janboll (commented):
/lgtm

avollmer-redhat:
@copilot resolve the merge conflicts in this pull request

avollmer-redhat:
## Mirror shared ingress HAProxy image to OCP ACR instead of SVC ACR

### Problem: Cilium e2e test failures

The previous commit (`8c3b5bb6c`) mirrored the shared ingress HAProxy image to **SVC ACR** and set `IMAGE_SHARED_INGRESS_HAPROXY` to point there. This caused 2 Cilium e2e test failures:

1. `Customer should be able to create a HCP cluster and use cilium CNI plugin`
2. `Customer should be able to create a no-CNI private cluster with a private key vault, a nodepool and install cilium CNI successfully`

Both failed with `unauthorized: authentication required` when customer worker nodes tried to pull the image from SVC ACR.

### Root cause

Upstream HyperShift uses `IMAGE_SHARED_INGRESS_HAPROXY` for **two different purposes** via a single env var:

1. **Shared ingress Deployment** on the management cluster (`hypershift-sharedingress` namespace) — runs HAProxy 3.0 with Proxy Protocol v2 TLV for SNI-based KAS routing
2. **`kube-apiserver-proxy` static pod** on customer worker nodes — a simple L4 proxy to the KAS endpoint, baked into ignition config

The `resolveHAProxyImage()` function in the nodepool controller (`nodepool_controller.go`) returns `GetSharedIngressHAProxyImage()` whenever `UseSharedIngress()` is true, which is always the case for ARO-HCP (`MANAGED_SERVICE=ARO-HCP`). There is no separate env var for the customer-side image.

### Why SVC ACR doesn't work for customer worker nodes

| ACR | Management cluster nodes | Customer worker nodes |
|---|---|---|
| **SVC ACR** (`arohcpsvc*`) | ✅ AcrPull RBAC on kubelet identity | ❌ No credentials |
| **OCP ACR** (`arohcpocp*`) | ✅ AcrPull RBAC on kubelet identity | ✅ Scope map tokens (`_repositories_pull`) created by CS |

Customer worker nodes receive ACR tokens scoped to OCP ACR's built-in `_repositories_pull` scope map, which grants `repositories/*/content/read` — pull access to **all** repositories in OCP ACR. They have no access to SVC ACR at all.

### Solution: Mirror to OCP ACR

The fix mirrors the shared ingress HAProxy image to **OCP ACR** instead of SVC ACR. This way both the management-side shared ingress Deployment and the customer-side `kube-apiserver-proxy` can pull the image:

- **Management cluster nodes**: Have AcrPull RBAC on OCP ACR (same as SVC ACR)
- **Customer worker nodes**: Have `_repositories_pull` scope map tokens granting pull access to all OCP ACR repositories

Changes:
- `hypershiftoperator/pipeline.yaml`: `targetACR` changed from `acr.svc.name` to `acr.ocp.name`
- `hypershiftoperator/values.yaml`: Image ref template changed from `acr.svc.name` to `acr.ocp.name`
- `dev-infrastructure/templates/global-acr.bicep`: Added `globalMSIOcpAcrAccess` module granting globalMSI push+pull access to OCP ACR (previously only had SVC ACR access)
- Fixture regenerated

### Options considered and ruled out

| Approach | Why ruled out |
|---|---|
| **Keep SVC ACR** | Customer worker nodes can't pull — no SVC ACR credentials, causes e2e failures and would break every subsequent deployment |
| **Add SVC ACR credentials to customer worker node pull secrets** | Requires CS and/or backend code changes to provision SVC ACR tokens per cluster — much larger scope |
| **Use separate env vars for management vs customer image** | Upstream HyperShift only has one env var (`IMAGE_SHARED_INGRESS_HAPROXY`); no per-NodePool override exposed through ARO-HCP API |
| **Use NodePool annotation `hypershift.openshift.io/haproxy-image`** | Highest priority override exists upstream, but ARO-HCP API doesn't expose annotations on NodePool resources — would need API surface changes |
| **Don't set `IMAGE_SHARED_INGRESS_HAPROXY` at all** | Management-side shared ingress Deployment falls back to hardcoded `quay.io` default — defeats the FSI compliance goal |
| **Skip the 2 Cilium e2e tests** | Would unblock this PR but the failure would recur in every upper environment deployment — not sustainable |
| **Use `image-sync` MSI instead of `globalMSI`** | `image-sync` MSI already has push to both ACRs, but would require exposing its resource ID as a Bicep output and changing the pipeline pattern — more invasive than granting `globalMSI` OCP ACR access |

### Why OCP ACR is appropriate for this image

The separation between SVC ACR and OCP ACR is:
- **SVC ACR**: Service component images (RP, CS, maestro, HyperShift operator, etc.) — consumed only by service/management infrastructure
- **OCP ACR**: Images consumed by hosted cluster infrastructure, including customer data plane nodes

The shared ingress HAProxy image is consumed by **both** management infrastructure (shared ingress Deployment) and customer data plane nodes (`kube-apiserver-proxy` static pod). Since it must be pullable by customer nodes, OCP ACR is the correct target. The `globalMSI` push access to OCP ACR is a minimal infrastructure change (one new ACR role assignment module, same pattern as the existing SVC ACR block).

### Infrastructure change note

The `globalMSIOcpAcrAccess` module in `global-acr.bicep` must be deployed before the first pipeline run that mirrors to OCP ACR. This happens automatically in dev environments via the global pipeline. For production, the corresponding `sdp-pipelines` change (already noted as a follow-up) must include this infrastructure update.

PR #4691, opened 2026-03-30
align all controllers to use the same style of constructor

deads2k:
Tidying up to use the currently standard constructor technique.

/assign @ahitacat


deads2k:
hold was a primitive "block queue" that wasn't released.

/hold cancel

PR #4744, opened 2026-04-02
docs: add 'Running pipelines locally' section to cspr.md

hbhushan3:
Adds a new "Running pipelines locally" section to the CSPR environment documentation with:

- A caution callout noting that CI/CD is the primary deployment path and local runs should only be used in emergencies
- `entrypoint/Region` as the command for full region deployment
- Dry-run usage example with `DRY_RUN=true`
- Nuke-and-rebuild workflow using `pipeline/Management.Delete` followed by `entrypoint/Region` for broken environments


raelga (changes requested):
@hbhushan3 Is this UP to date? The CSPR was recreated.

PR #4753, opened 2026-04-03
Validate Azure-AsyncNotificationUri header to prevent SSRF

gouthamMN:
### What
Add `ValidateNotificationURI()` which enforces HTTPS and restricts the host to known ARM management endpoints (management.azure.com, management.usgovcloudapi.net).

Validation is applied in `NewOperation()` and in the delete handlers for clusters, node pools, and external auth configs where the header value is set directly on the operation document. Invalid URIs are silently dropped since this header is injected by ARM, not by end users.

### Why
We were [alerted](https://portal.microsofticm.com/imp/v5/incidents/details/31000000578371/summary) by a MSFT internal security team about a potential security vulnerability in our source code.

The `NotificationURI` from the `Azure-AsyncNotificationUri` header is used by the backend to make outbound HTTP POST requests when async operations complete. Without validation, a crafted URI could be used to target internal services or attacker-controlled endpoints.

### Testing
unit testing


raelga (changes requested):
@gouthamMN Can you take a look to https://github.com/Azure/ARO-HCP/pull/4753#issuecomment-4184845510 feedback?

geoberle:
closing as stale
reopen if you still needed

PR #4765, opened 2026-04-06
Changing functions to have non-pointer method receievers

deads2k:
This caused an interface check to fail, which caused us to lose our logging context.  Added a unit test to be sure we dont' lose it again for at least the value.


stevekuznetsov (approved):
/lgtm
/approve

PR #4774, opened 2026-04-07
Enable testing of versions > 4.20 and calculate latest z-stream for non-stable

miquelsi:
### What

Remove the filter that prevents testing cluster install for OCP versions greater than 4.20

Note there's a new filter to skip candidate builds of OCP 5.0 until https://redhat.atlassian.net/browse/OCPBUGS-83564 is fixed.

We are also calculating the latest z-stream version of OCP when using non-stable channels.

### Why

This filter was added because of 2 issues breaking versions greater than 4.20. 

- https://redhat.atlassian.net/browse/ARO-25490
- https://redhat.atlassian.net/browse/ARO-24955

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ report.json
 _artifacts
 must-gather-*
 /hcpctl
+timing-metadata-*
--- a/test/util/framework/version_helper.go
+++ b/test/util/framework/version_helper.go
@@ -24,6 +24,7 @@ import (
 	"net/url"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/blang/semver/v4"
 	"github.com/google/uuid"
@@ -38,7 +39,10 @@ var (
 	ErrNightlyReleaseStreamNotFound = errors.New("nightly release stream not found")
 	ErrNoAcceptedNightlyTags        = errors.New("no accepted nightly tags found")
 	ErrNoParseableNightlyTags       = errors.New("no parseable nightly tags found")
+	ErrVersionNotFound              = errors.New("no graph nodes found")
 )
+
+const graphAPIRequestTimeout = 30 * time.Second
 
 // GetInstallVersionForZStreamUpgrade returns the version to install the cluster with when testing
 // a z-stream upgrade, and whether that version has an available z-stream upgrade path. It uses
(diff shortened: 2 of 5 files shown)


test/util/framework/version_helper.go
```diff
@@ -198,6 +199,86 @@ func GetUpgradeCandidatesInMaxMinorFromCincinnati(ctx context.Context, channelGr
...
+	case "candidate":
+		return GetLatestInstallVersionForGraphChannel("candidate", version)
+	case "fast":
+		return GetLatestInstallVersionForGraphChannel("fast", version)
+	default:
+		return "", fmt.Errorf("invalid channel group: %s", channelGroup)
+	}
+}
+
+func GetLatestInstallVersionForGraphChannel(channelGroup string, version string) (string, error) {
```
JakobGray: Does this function the same as GetLatestVersionInMinor() and if so we could use that instead

miquelsi: It does the same, but `GetLatestVersionInMinor()` requires a context parameter, and here we need this to be called from deployment_params.go, with no context.

We could follow up on this and use a common approach, but it would need to pass the context to all `NewDefaultClusterParams`.

JakobGray: Would there be an issue in passing context.Background() to satisfy the parameter?

miquelsi: Following up on this. The reason to add a different method is because `GetLatesstVersionInMinor()` won't return engineering candidate (-ec.x) releases, and we want to start testing those.

.gitignore
```diff
@@ -33,3 +33,4 @@ report.json
 _artifacts
 must-gather-*
 /hcpctl
+timing-metadata-*
```
miquelsi, line 36: Ignoring these files generated when running tests locally.

test/util/framework/version_helper.go
```diff
@@ -198,6 +202,86 @@ func GetUpgradeCandidatesInMaxMinorFromCincinnati(ctx context.Context, channelGr
...
+			latestVersion = nodeVersion
+			latestVersionID = node.Version
+		}
+	}
+
+	if latestVersionID == "" {
+		return "", fmt.Errorf("no %s versions found in %s for requested minor %s", channelGroup, channel, version)
+	}
+
+	// For stable channel, return the latest version in the minor
```
JakobGray: This comment sounds misleading because it's not returning the latest version in the minor. Do we need this 'if' statement at all? We already avoid calling this function for stable with `if channelGroup != "stable"`

miquelsi: You are right, this is no longer necessary.

JakobGray (approved):

PR #4819, opened 2026-04-09
validation: Require tenantId in subscriptions

mbarnes:
Followup to [a comment in #4784](https://github.com/Azure/ARO-HCP/pull/4784#discussion_r3051165754) 

### What

This changes subscription validation to require the `TenantId` field.

### Why

In a subscription PUT request, the `TenantId` field is optional according to the [Resource Provider Contract](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/subscription-lifecycle-api-reference.md) but the ARO-HCP backend relies on it for cluster creation. Therefore we want to make sure we never write a subscription document to Cosmos without a `TenantId` value.  The frontend tries to ensure its presence before validating, filling it in from the `X-Ms-Home-Tenant-Id` request header if necessary.

It's unclear whether a subscription registration without a `TenantId` value would ever actually occur outside of test environments.  This may be purely academic, but it ensures an invariant in our Cosmos DB data.

### Testing

Several existing integration tests were updated to reflect the automatic addition of a tenant ID value from request headers.

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -562,6 +562,20 @@ func (f *Frontend) ArmSubscriptionPut(writer http.ResponseWriter, request *http.
 	}
 	requestSubscription.ResourceID = requestSubscription.CosmosMetadata.ResourceID
 
+	// Ensure a tenant ID value is present in SubscriptionProperties. If it was omitted
+	// in the request body then get it from the request headers. The tenant ID field is
+	// optional according to the Resource Provider Contract, but the backend depends on
+	// it for cluster creation so it's effectively a required field for ARO-HCP.
+	if requestSubscription.Properties == nil {
+		requestSubscription.Properties = &arm.SubscriptionProperties{}
+	}
+	if requestSubscription.Properties.TenantId == nil || len(*requestSubscription.Properties.TenantId) == 0 {
+		tenantID := request.Header.Get(arm.HeaderNameHomeTenantID)
+		if len(tenantID) > 0 {
+			requestSubscription.Properties.TenantId = api.Ptr(tenantID)
+		}
+	}
+
 	validationErrs := validation.ValidateSubscriptionCreate(ctx, &requestSubscription)
 	if err := arm.CloudErrorFromFieldErrors(validationErrs); err != nil {
 		return utils.TrackError(err)
--- a/internal/validation/validate_subscription.go
+++ b/internal/validation/validate_subscription.go
@@ -18,12 +18,17 @@ import (
 	"context"
 
 	"k8s.io/apimachinery/pkg/api/operation"
+	"k8s.io/apimachinery/pkg/api/safe"
 	"k8s.io/apimachinery/pkg/api/validate"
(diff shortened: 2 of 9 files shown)


internal/validation/validate_subscription.go
```diff
@@ -50,7 +55,20 @@ func validateSubscription(ctx context.Context, op operation.Operation, newObj, o
...
 	return errs
 }
+
+func validateSubscriptionProperties(ctx context.Context, op operation.Operation, fldPath *field.Path, newObj, oldObj *arm.SubscriptionProperties) field.ErrorList {
+	errs := field.ErrorList{}
+
+	// TenantId is optional according to the Resource Provider Contract but the
+	// ARO-HCP backend relies on it for cluster creation. The frontend tries to
+	// ensure its presence, cherry-picking from request headers if necessary.
+	errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("tenantId"), &newObj.TenantId, nil)...)
```
geoberle, line 71: are there ways newObj can be null?
i know you populate it in frontend.go but other paths might lead to here eventually via `ValidateSubscriptionCreate`

frontend/pkg/frontend/frontend.go
```diff
@@ -562,6 +562,20 @@ func (f *Frontend) ArmSubscriptionPut(writer http.ResponseWriter, request *http.
...
 	requestSubscription.ResourceID = requestSubscription.CosmosMetadata.ResourceID
 
+	// Ensure a tenant ID value is present in SubscriptionProperties. If it was omitted
+	// in the request body then get it from the request headers. The tenant ID field is
+	// optional according to the Resource Provider Contract, but the backend depends on
+	// it for cluster creation so it's effectively a required field for ARO-HCP.
+	if requestSubscription.Properties == nil {
+		requestSubscription.Properties = &arm.SubscriptionProperties{}
+	}
+	if requestSubscription.Properties.TenantId == nil || len(*requestSubscription.Properties.TenantId) == 0 {
```
geoberle, line 572: so the request data can send various invalid states? properties nil but also properties not nil and tenant nil or tenant empty? this this a known ARM inconsistency?

internal/validation/validate_subscription.go
geoberle: check for RequiredPointer as well

frontend/pkg/frontend/frontend.go
mbarnes: The RPC states that `Properties` is a required field in subscription payloads but `Properties.TenantId` is optional.  For better or worse [we have the subscription model defined entirely as pointer fields](https://github.com/Azure/ARO-HCP/blob/main/internal/api/arm/subscription.go), so I'm just being careful here to avoid a panic.  Such inconsistencies would be more likely, I think, in a test scenario than an actual request from ARM.

internal/validation/validate_subscription.go
deads2k, line 71: required pointer, then nil check, then requirevalue

PR #4821, opened 2026-04-09
feat: create clusters service control plane in backend

JakobGray:
### What

Move CS cluster creation out of OperationClusterCreate into a dedicated
ClusterClusterServiceCreate controller. The new controller waits for
ControlPlaneDesiredVersion to resolve the install version, creates the CS
cluster (or reuses one matched by Azure metadata), and persists
ClusterServiceID on the cluster document.

OperationClusterCreate now polls CS provisioning status via that stored ID
instead of operation.InternalID. BuildCSCluster prefers the resolved
DesiredVersion from ServiceProviderCluster when building the CS cluster.

Async cluster installation will not take affect until cluster creation
is removed from the frontend.

### Why

The clusters service control plane is currently created synchronously during the create flow and uses hard coded versions derived from the customer desired version. By moving to an asynchronous approach we reduce create time and have clusters service deployment managed in the background. This will also allow us time to lookup the desired version from Cincinnati.

### Testing

Includes unit tests. Integration and end-to-end testing with the controller functional is done in https://github.com/Azure/ARO-HCP/pull/5709.

### DIFF
--- a/backend/pkg/app/backend.go
+++ b/backend/pkg/app/backend.go
@@ -38,6 +38,7 @@ import (
 	azureclient "github.com/Azure/ARO-HCP/backend/pkg/azure/client"
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers"
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/billingcontrollers"
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/clustercreation"
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/clusterdeletion"
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/clusterpropertiescontroller"
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
@@ -774,6 +775,13 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context, 
 		backendInformers,
 	)
 
+	clusterClusterServiceCreateController := clustercreation.NewClusterClusterServiceCreateController(
+		b.options.ResourcesDBClient,
+		b.options.ClustersServiceClient,
+		activeOperationLister,
+		backendInformers,
+	)
+
 	clusterDeletionClusterServiceDeleteDispatchController := clusterdeletion.NewClusterClusterServiceDeleteDispatchController(
 		utilsclock.RealClock{},
 		b.options.ResourcesDBClient,
@@ -829,6 +837,7 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context, 
 				go doNothingController.Run(ctx, 20)
 				go dispatchRequestCredentialController.Run(ctx, 20)
 				go dispatchRevokeCredentialsController.Run(ctx, 20)
+				go clusterClusterServiceCreateController.Run(ctx, 20)
 				go nodePoolClusterServiceCreateController.Run(ctx, 20)
 				go externalAuthClusterServiceCreateController.Run(ctx, 20)
 				go operationClusterCreateController.Run(ctx, 20)
--- a/backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
+++ b/backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
@@ -1,0 +1,254 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clustercreation
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/backend/pkg/informers"
+	"github.com/Azure/ARO-HCP/backend/pkg/listers"
+	"github.com/Azure/ARO-HCP/internal/api"
+	controllerutil "github.com/Azure/ARO-HCP/internal/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+type clusterClusterServiceCreateSyncer struct {
+	cooldownChecker       controllerutil.CooldownChecker
+	resourcesDBClient     database.ResourcesDBClient
+	clusterLister         listers.ClusterLister
+	subscriptionLister    listers.SubscriptionLister
+	clustersServiceClient ocm.ClusterServiceClientSpec
+}
+
+var _ controllerutils.ClusterSyncer = (*clusterClusterServiceCreateSyncer)(nil)
+
+func NewClusterClusterServiceCreateController(
+	resourcesDBClient database.ResourcesDBClient,
+	clustersServiceClient ocm.ClusterServiceClientSpec,
+	activeOperationLister listers.ActiveOperationLister,
+	backendInformers informers.BackendInformers,
+) controllerutils.Controller {
+	_, clusterLister := backendInformers.Clusters()
+	_, subscriptionLister := backendInformers.Subscriptions()
+	syncer := &clusterClusterServiceCreateSyncer{
+		cooldownChecker:       controllerutils.DefaultActiveOperationPrioritizingCooldown(activeOperationLister),
+		resourcesDBClient:     resourcesDBClient,
+		clusterLister:         clusterLister,
+		subscriptionLister:    subscriptionLister,
+		clustersServiceClient: clustersServiceClient,
+	}
+
+	return controllerutils.NewClusterWatchingController(
+		"ClusterClusterServiceCreate",
+		resourcesDBClient,
+		backendInformers,
+		nil,
+		time.Minute,
+		syncer,
+	)
+}
+
+func (c *clusterClusterServiceCreateSyncer) needsWork(cluster *api.HCPOpenShiftCluster) bool {
+	return cluster.ServiceProviderProperties.DeletionTimestamp == nil &&
+		(cluster.ServiceProviderProperties.ClusterServiceID == nil ||
+			len(cluster.ServiceProviderProperties.ClusterServiceID.String()) == 0)
+}
+
+func (c *clusterClusterServiceCreateSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	logger := utils.LoggerFromContext(ctx)
+
+	// Quick cache lookup first to see if work is needed
+	cluster, err := c.clusterLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if !c.needsWork(cluster) {
+		return nil
+	}
+
+	// Confirm against the live document to make sure the cluster hasn't been deleted or modified since we last checked
+	cluster, err = c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if !c.needsWork(cluster) {
+		return nil
+	}
+
+	existingServiceProviderCluster, err := database.GetOrCreateServiceProviderCluster(ctx, c.resourcesDBClient, cluster.ID)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get or create ServiceProviderCluster: %w", err))
+	}
+
+	ready, err := c.createPreconditionDesiredVersionResolved(ctx, existingServiceProviderCluster)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	if !ready {
+		return nil
+	}
+
+	subscription, err := c.subscriptionLister.Get(ctx, key.SubscriptionID)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	if subscription.Properties == nil || subscription.Properties.TenantId == nil {
+		return utils.TrackError(fmt.Errorf("subscription %s has no tenantId", key.SubscriptionID))
+	}
+	tenantID := *subscription.Properties.TenantId
+	mrg := cluster.CustomerProperties.Platform.ManagedResourceGroup
+
+	csCluster, err := c.findAROHCPClusterByAzureInfo(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, tenantID, mrg)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if csCluster == nil {
+		csCluster, err = c.createClusterServiceCluster(ctx, cluster, existingServiceProviderCluster, tenantID)
+		if err != nil {
+			return utils.TrackError(fmt.Errorf("failed to create cluster in CS: %w", err))
+		}
+	}
+
+	csInternalID, err := api.NewInternalID(csCluster.HREF())
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	logger.Info("Storing ClusterServiceID on cluster document", "clusterServiceID", csInternalID.String())
+	cluster.ServiceProviderProperties.ClusterServiceID = &csInternalID
+	_, err = c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Replace(ctx, cluster, nil)
+	if database.IsPreconditionFailedError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to replace Cluster: %w", err))
+	}
+
+	return nil
+}
+
+// createPreconditionDesiredVersionResolved reports whether the ControlPlaneDesiredVersion
+// controller has written the Cincinnati-resolved desired version to the ServiceProviderCluster.
+// Returns (false, nil) when this controller should wait and retry.
+func (c *clusterClusterServiceCreateSyncer) createPreconditionDesiredVersionResolved(ctx context.Context, serviceProviderCluster *api.ServiceProviderCluster) (bool, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	if serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion != nil {
+		return true, nil
+	}
+	logger.Info("DesiredVersion not yet set, waiting for ControlPlaneDesiredVersion controller")
+	return false, nil
+}
+
+// findAROHCPClusterByAzureInfo returns the Cluster Service cluster whose Azure
+// metadata matches the given subscription, resource group, ARM resource name,
+// tenant ID, and managed resource group name (MRG).
+// It returns (nil, nil) when no such cluster exists.
+// An error is returned if more than one cluster is returned matching the Azure metadata, as it should be unique.
+func (c *clusterClusterServiceCreateSyncer) findAROHCPClusterByAzureInfo(ctx context.Context, subscriptionID, resourceGroupName, resourceName, tenantID, managedResourceGroupName string) (*arohcpv1alpha1.Cluster, error) {
+	// Subscription ID, resource group, and cluster name are lowercased when building the Cluster Service
+	// cluster (see withImmutableAttributes in convert.go).
+	wantSub := strings.ToLower(subscriptionID)
+	wantRG := strings.ToLower(resourceGroupName)
+	wantName := strings.ToLower(resourceName)
+	// Tenant ID and managed resource group are not lowercased in the OCM CS
+	// builder (see withImmutableAttributes in convert.go), we keep the casing as it is.
+	wantTenant := tenantID
+	wantMRG := managedResourceGroupName
+	search := c.clustersServiceClusterByAzureInfoSearchString(wantSub, wantRG, wantName, wantTenant, wantMRG)
+	matches, err := c.csClustersMatchingClusterByAzureInfo(ctx, c.clustersServiceClient.ListClusters(search), wantSub, wantRG, wantName, wantTenant, wantMRG)
+	if err != nil {
+		return nil, err
+	}
+	if len(matches) > 1 {
+		return nil, fmt.Errorf(
+			"cluster service returned %d clusters for one Azure resource (expected exactly 1): "+
+				"subscription_id=%q resource_group=%q resource_name=%q tenant_id=%q managed_resource_group=%q",
+			len(matches), wantSub, wantRG, wantName, wantTenant, wantMRG,
+		)
+	}
+	if len(matches) == 1 {
+		return matches[0], nil
+	}
+	return nil, nil
+}
+
+func (c *clusterClusterServiceCreateSyncer) clustersServiceClusterByAzureInfoSearchString(wantSub, wantRG, wantName, wantTenant, wantMRG string) string {
+	return fmt.Sprintf(
+		"azure.subscription_id = '%s' and azure.resource_group_name = '%s' and azure.resource_name = '%s' and "+
+			"azure.tenant_id = '%s' and azure.managed_resource_group_name = '%s'",
+		wantSub, wantRG, wantName, wantTenant, wantMRG,
+	)
+}
+
+func (c *clusterClusterServiceCreateSyncer) csClustersMatchingClusterByAzureInfo(ctx context.Context, it ocm.ClusterListIterator, wantSub, wantRG, wantName, wantTenant, wantMRG string) ([]*arohcpv1alpha1.Cluster, error) {
+	var res []*arohcpv1alpha1.Cluster
+	for csCluster := range it.Items(ctx) {
+		az := csCluster.Azure()
+		if az == nil {
+			continue
+		}
+		if az.SubscriptionID() != wantSub ||
+			az.ResourceGroupName() != wantRG ||
+			az.ResourceName() != wantName ||
+			az.TenantID() != wantTenant ||
+			az.ManagedResourceGroupName() != wantMRG {
+			continue
+		}
+		res = append(res, csCluster)
+	}
+	if err := it.GetError(); err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+func (c *clusterClusterServiceCreateSyncer) createClusterServiceCluster(ctx context.Context, cluster *api.HCPOpenShiftCluster, serviceProviderCluster *api.ServiceProviderCluster, tenantID string) (*arohcpv1alpha1.Cluster, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	csClusterBuilder, csAutoscalerBuilder, err := ocm.BuildCSCluster(cluster.ID, tenantID, cluster, nil, nil, serviceProviderCluster)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to build CS cluster: %w", err))
+	}
+
+	logger.Info("Creating cluster in Cluster Service", "version", serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion.String())
+	result, err := c.clustersServiceClient.PostCluster(ctx, csClusterBuilder, csAutoscalerBuilder)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("PostCluster failed: %w", err))
+	}
+
+	return result, nil
+}
+
+func (c *clusterClusterServiceCreateSyncer) CooldownChecker() controllerutil.CooldownChecker {
+	return c.cooldownChecker
+}
--- a/backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
+++ b/backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
@@ -1,0 +1,514 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clustercreation
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/blang/semver/v4"
+	"github.com/go-logr/logr/testr"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	"k8s.io/utils/ptr"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+
+	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/backend/pkg/listertesting"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	"github.com/Azure/ARO-HCP/internal/databasetesting"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// Test constants
+const (
+	testSubscriptionID      = "00000000-0000-0000-0000-000000000000"
+	testResourceGroupName   = "test-rg"
+	testClusterName         = "test-cluster"
+	testClusterServiceIDStr = "/api/aro_hcp/v1alpha1/clusters/abc123"
+	testTenantID            = "11111111-1111-1111-1111-111111111111"
+	testClusterUID          = "00000000-0000-0000-0000-000000000000"
+	// testManagedResourceGroup must match what api.MinimumValidClusterTestCase() sets.
+	testManagedResourceGroup = "testManagedResourceGroup"
+)
+
+// testClusterResourceID builds the ARM resource ID for the test cluster.
+func testClusterResourceID() *azcorearm.ResourceID {
+	return api.Must(azcorearm.ParseResourceID(
+		"/subscriptions/" + testSubscriptionID +
+			"/resourceGroups/" + testResourceGroupName +
+			"/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/" + testClusterName,
+	))
+}
+
+// newTestCluster returns an HCPOpenShiftCluster based on MinimumValidClusterTestCase with
+// test-constant IDs. Callers can further customize it via functional opts.
+// MinimumValidClusterTestCase is used as the base because createClusterServiceCluster
+// calls ocm.BuildCSCluster, which requires a fully-populated cluster (version, DNS, subnet, etc.).
+func newTestCluster(opts ...func(*api.HCPOpenShiftCluster)) *api.HCPOpenShiftCluster {
+	rid := testClusterResourceID()
+	cluster := api.MinimumValidClusterTestCase()
+	cluster.CosmosMetadata = arm.CosmosMetadata{
+		ResourceID:   rid,
+		PartitionKey: strings.ToLower(rid.SubscriptionID),
+	}
+	cluster.ID = rid
+	cluster.Name = testClusterName
+	cluster.Type = rid.ResourceType.String()
+	cluster.ServiceProviderProperties.ClusterServiceID = nil
+	cluster.ServiceProviderProperties.ClusterUID = testClusterUID
+	for _, opt := range opts {
+		opt(cluster)
+	}
+	return cluster
+}
+
+// newTestSubscription returns a minimal Subscription with tenant ID set.
+func newTestSubscription() *arm.Subscription {
+	rid := api.Must(azcorearm.ParseResourceID("/subscriptions/" + testSubscriptionID))
+	return &arm.Subscription{
+		CosmosMetadata: api.CosmosMetadata{
+			ResourceID:   rid,
+			PartitionKey: strings.ToLower(rid.SubscriptionID),
+		},
+		ResourceID: rid,
+		Properties: &arm.SubscriptionProperties{TenantId: ptr.To(testTenantID)},
+	}
+}
+
+// newTestSPC returns a ServiceProviderCluster for the test cluster.
+// Callers can customize it via functional opts.
+func newTestSPC(opts ...func(*api.ServiceProviderCluster)) *api.ServiceProviderCluster {
+	resourceID := api.Must(azcorearm.ParseResourceID(fmt.Sprintf("%s/%s/%s",
+		testClusterResourceID().String(),
+		api.ServiceProviderClusterResourceTypeName,
+		api.ServiceProviderClusterResourceName,
+	)))
+	spc := &api.ServiceProviderCluster{
+		CosmosMetadata: api.CosmosMetadata{ResourceID: resourceID},
+		Spec:           api.ServiceProviderClusterSpec{},
+	}
+	spc.SetPartitionKey(testSubscriptionID)
+	for _, opt := range opts {
+		opt(spc)
+	}
+	return spc
+}
+
+func TestClusterClusterServiceCreate_SyncOnce(t *testing.T) {
+	desiredVersion := ptr.To(semver.MustParse("4.20.0"))
+	clusterInternalID := api.Must(api.NewInternalID(testClusterServiceIDStr))
+
+	tests := []struct {
+		name                           string
+		listCluster                    *api.HCPOpenShiftCluster    // cluster seeded into the lister (nil = not found)
+		dbCluster                      *api.HCPOpenShiftCluster    // cluster stored in the DB
+		existingServiceProviderCluster *api.ServiceProviderCluster // nil = not pre-seeded; controller get-or-creates
+		setupMockCS                    func(ctrl *gomock.Controller) ocm.ClusterServiceClientSpec
+		expectError                    bool
+		verifyDB                       func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient)
+	}{
+		{
+			name:        "successful sync records cluster service ID on cluster",
+			listCluster: newTestCluster(),
+			dbCluster:   newTestCluster(),
+			existingServiceProviderCluster: newTestSPC(func(spc *api.ServiceProviderCluster) {
+				spc.Spec.ControlPlaneVersion.DesiredVersion = desiredVersion
+			}),
+			setupMockCS: func(ctrl *gomock.Controller) ocm.ClusterServiceClientSpec {
+				mockCS := ocm.NewMockClusterServiceClientSpec(ctrl)
+				mockCS.EXPECT().
+					ListClusters(gomock.Any()).
+					Return(ocm.NewSimpleClusterListIterator(nil, nil))
+				csCluster, err := arohcpv1alpha1.NewCluster().
+					HREF(testClusterServiceIDStr).
+					Build()
+				require.NoError(t, err)
+				mockCS.EXPECT().
+					PostCluster(gomock.Any(), gomock.Any(), gomock.Any()).
+					Return(csCluster, nil)
+				return mockCS
+			},
+			expectError: false,
+			verifyDB: func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient) {
+				cluster, err := db.HCPClusters(testSubscriptionID, testResourceGroupName).Get(ctx, testClusterName)
+				require.NoError(t, err)
+				require.NotNil(t, cluster.ServiceProviderProperties.ClusterServiceID)
+				assert.Equal(t, testClusterServiceIDStr, cluster.ServiceProviderProperties.ClusterServiceID.String())
+			},
+		},
+		{
+			name: "skip when cluster already has ClusterServiceID",
+			listCluster: newTestCluster(func(c *api.HCPOpenShiftCluster) {
+				c.ServiceProviderProperties.ClusterServiceID = &clusterInternalID
+			}),
+			dbCluster: newTestCluster(func(c *api.HCPOpenShiftCluster) {
+				c.ServiceProviderProperties.ClusterServiceID = &clusterInternalID
+			}),
+			setupMockCS: func(ctrl *gomock.Controller) ocm.ClusterServiceClientSpec {
+				return ocm.NewMockClusterServiceClientSpec(ctrl)
+			},
+			expectError: false,
+			verifyDB: func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient) {
+				cluster, err := db.HCPClusters(testSubscriptionID, testResourceGroupName).Get(ctx, testClusterName)
+				require.NoError(t, err)
+				require.NotNil(t, cluster.ServiceProviderProperties.ClusterServiceID)
+				assert.Equal(t, testClusterServiceIDStr, cluster.ServiceProviderProperties.ClusterServiceID.String())
+			},
+		},
+		{
+			name:        "desired version not set waits without dispatching",
+			listCluster: newTestCluster(),
+			dbCluster:   newTestCluster(),
+			setupMockCS: func(ctrl *gomock.Controller) ocm.ClusterServiceClientSpec {
+				return ocm.NewMockClusterServiceClientSpec(ctrl)
+			},
+			expectError: false,
+			verifyDB: func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient) {
+				cluster, err := db.HCPClusters(testSubscriptionID, testResourceGroupName).Get(ctx, testClusterName)
+				require.NoError(t, err)
+				assert.Nil(t, cluster.ServiceProviderProperties.ClusterServiceID)
+			},
+		},
+		{
+			name:        "adopts existing Cluster Service cluster for Azure resource",
+			listCluster: newTestCluster(),
+			dbCluster:   newTestCluster(),
+			existingServiceProviderCluster: newTestSPC(func(spc *api.ServiceProviderCluster) {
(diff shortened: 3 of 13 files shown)


backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
```diff
@@ -0,0 +1,202 @@
...
+	provisionShard   string
+	noopProvision    bool
+	noopDeprovision  bool
+}
+
+var _ controllerutils.ClusterSyncer = (*clusterServiceCreateSyncer)(nil)
+
+// NewClusterServiceCreateController creates a controller that registers clusters
+// with Cluster Service once their desired control plane version is computed.
+func NewClusterServiceCreateController(
```
mbarnes, line 52: Probably want to clarify that this is for _cluster_ creation.  I assume nested resource types like node pools will have separate controllers from this one.
```suggestion
func NewClusterServiceCreateClusterController(
```

frontend/pkg/frontend/cluster.go
```diff
@@ -599,6 +588,14 @@ func (f *Frontend) patchHCPCluster(writer http.ResponseWriter, request *http.Req
 func (f *Frontend) updateHCPClusterInCosmos(ctx context.Context, writer http.ResponseWriter, request *http.Request, httpStatusCode int, newInternalCluster, oldInternalCluster *api.HCPOpenShiftCluster) error {
 	logger := utils.LoggerFromContext(ctx)
 
+	if oldInternalCluster.ServiceProviderProperties.ClusterServiceID.String() == "" {
```
machi1990, line 591: len(s) == 0

frontend/pkg/frontend/cluster.go
```diff
@@ -599,6 +588,14 @@ func (f *Frontend) patchHCPCluster(writer http.ResponseWriter, request *http.Req
...
+		return arm.NewCloudError(
+			http.StatusConflict,
+			arm.CloudErrorCodeConflict,
+			oldInternalCluster.ID.String(),
+			"The cluster is still being registered with Cluster Service. Please retry shortly.")
```
machi1990, line 596: How can this happen i.e the update to be triggered before the cluster reached terminal state

frontend/pkg/frontend/cluster.go
```diff
@@ -902,8 +903,14 @@ func legacyMergeToInternalCluster(csCluster *arohcpv1alpha1.Cluster, internalClu
 
 // readInternalClusterFromClusterService takes an internal Cluster read from cosmos, retrieves the corresponding cluster-service data,
 // merges the states together, and returns the internal representation.
+// If ClusterServiceID is not yet set (cluster creation in CS is pending), the
+// cluster is returned as-is from Cosmos without CS-only fields.
 // TODO remove the header it takes and collapse that to some general error handling.
 func (f *Frontend) readInternalClusterFromClusterService(ctx context.Context, oldInternalCluster *api.HCPOpenShiftCluster) (*api.HCPOpenShiftCluster, error) {
+	if oldInternalCluster.ServiceProviderProperties.ClusterServiceID.String() == "" {
```
machi1990, line 910: len(s)==0

backend/pkg/app/backend.go
```diff
@@ -74,6 +75,9 @@ type BackendOptions struct {
 	FPAMIDataplaneClientBuilder        azureclient.FPAMIDataplaneClientBuilder
 	SMIClientBuilder                   azureclient.ServiceManagedIdentityClientBuilder
 	CheckAccessV2ClientBuilder         azureclient.CheckAccessV2ClientBuilder
+	ClusterServiceProvisionShard       string
+	ClusterServiceNoopProvision        bool
+	ClusterServiceNoopDeprovision      bool
```
machi1990, line 80: Are these things we should continue to honor? I think we are way past that "noop" phase now that we've everything wired up

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -76,7 +76,26 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
 		return nil // no work to do
 	}
 
-	clusterStatus, err := c.clusterServiceClient.GetClusterStatus(ctx, operation.InternalID)
+	// When CS cluster creation is deferred to a backend controller, the
+	// operation is initially created without an InternalID. Look it up
+	// from the cluster document's ClusterServiceID.
+	internalID := operation.InternalID
+	if internalID.String() == "" {
```
machi1990, line 83: len(s)==0

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -76,7 +76,26 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
...
+		cluster, err := c.cosmosClient.HCPClusters(
+			operation.ExternalID.SubscriptionID,
+			operation.ExternalID.ResourceGroupName,
+		).Get(ctx, operation.ExternalID.Name)
+		if err != nil {
+			return utils.TrackError(fmt.Errorf("failed to get cluster: %w", err))
+		}
+		internalID = cluster.ServiceProviderProperties.ClusterServiceID
+		if internalID.String() == "" {
```
machi1990, line 92: len(s)==0

backend/pkg/controllers/mismatchcontrollers/cluster_service_cluster_matching.go
```diff
@@ -80,11 +80,15 @@ func (c *clusterServiceClusterMatching) getAllCosmosObjs(ctx context.Context) (m
 
 		for _, cluster := range allHCPClusters.Items(ctx) {
 			ret = append(ret, cluster)
-			existingCluster, exists := clusterServiceIDToCluster[cluster.ServiceProviderProperties.ClusterServiceID.String()]
+			csID := cluster.ServiceProviderProperties.ClusterServiceID.String()
+			if csID == "" {
```
machi1990, line 84: len(s)==0

backend/pkg/controllers/mismatchcontrollers/cluster_service_cluster_matching.go
```diff
@@ -80,11 +80,15 @@ func (c *clusterServiceClusterMatching) getAllCosmosObjs(ctx context.Context) (m
 
 		for _, cluster := range allHCPClusters.Items(ctx) {
 			ret = append(ret, cluster)
-			existingCluster, exists := clusterServiceIDToCluster[cluster.ServiceProviderProperties.ClusterServiceID.String()]
+			csID := cluster.ServiceProviderProperties.ClusterServiceID.String()
```
machi1990, line 83:
```suggestion
			clusterServiceID := cluster.ServiceProviderProperties.ClusterServiceID.String()
```

backend/pkg/controllers/mismatchcontrollers/cosmos_cluster_matching.go
```diff
@@ -60,6 +60,11 @@ func (c *cosmosClusterMatching) synchronizeClusters(ctx context.Context, keyObj
 		return utils.TrackError(err)
 	}
 
+	// Skip clusters that don't have a ClusterServiceID yet (CS creation pending).
+	if cosmosCluster.ServiceProviderProperties.ClusterServiceID.String() == "" {
```
machi1990, line 64: len(s)==0

throughout

backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
```diff
@@ -0,0 +1,202 @@
...
+// clusterServiceCreateSyncer creates the cluster in Cluster Service after
+// the CP desired version controller has computed the initial desired version.
+type clusterServiceCreateSyncer struct {
+	cooldownChecker      controllerutils.CooldownChecker
+	cosmosClient         database.DBClient
+	clusterServiceClient ocm.ClusterServiceClientSpec
+
+	provisionShard   string
+	noopProvision    bool
+	noopDeprovision  bool
```
machi1990, line 45: I don't think we need these anymore; we are now able to create clusters end to end and I don't see the merit of continue to carry this NOOP provisioning/deprovisioning logic

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -108,16 +114,21 @@ func (c *controlPlaneDesiredVersionSyncer) SyncOnce(ctx context.Context, key con
...
-
-	// Create Cincinnati client with cluster UUID from Cluster Service
-	clusterUUID, err := uuid.Parse(clusterServiceCluster.ExternalID())
-	if err != nil {
-		return utils.TrackError(fmt.Errorf("invalid cluster UUID from Cluster Service: %w", err))
+	// When ClusterServiceID is set, read the cluster UUID from CS. Otherwise use a
+	// shared default UUID. Cincinnati's upgrade graph is deterministic regardless of
+	// UUID so this is safe for initial version computation before CS creation.
+	var clusterUUID uuid.UUID
+	if existingCluster.ServiceProviderProperties.ClusterServiceID.String() != "" {
```
machi1990, line 121:
```suggestion
	if len(existingCluster.ServiceProviderProperties.ClusterServiceID.String()) > 0 {
```

machi1990, line 121: let's also do the same here https://github.com/Azure/ARO-HCP/blob/c331d211031b1b4dfbd6136cc81e3cccececca7e/backend/pkg/controllers/upgradecontrollers/trigger_control_plane_upgrade_controller.go#L86

+ adjusting the logic there to;
- not trigger the upgrade if CS id isn't set
- not trigger the upgrade if CS' version == desired version (not strictly needed but we can do it to avoid to avoid creating a policy for nothing)

backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
```diff
@@ -0,0 +1,202 @@
...
+	clusterCRUD := c.cosmosClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName)
+	existingCluster, err := clusterCRUD.Get(ctx, key.HCPClusterName)
+	if database.IsResponseError(err, http.StatusNotFound) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
+	}
+
+	if existingCluster.ServiceProviderProperties.ClusterServiceID.String() != "" {
```
machi1990, line 97:
```suggestion
	if len(existingCluster.ServiceProviderProperties.ClusterServiceID.String()) > 0 {
```

backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if existingServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion == nil {
+		logger.Info("DesiredVersion not yet set, waiting for version controller")
+		return nil
+	}
+
+	// Search for an existing CS cluster that matches this Azure resource.
+	// This handles the case where CS creation succeeded but we failed to
+	// persist the CS ID in Cosmos.
+	csCluster, err := c.findExistingCSCluster(ctx, existingCluster)
```
machi1990, line 114:
```suggestion
	existingClusterServiceCluster, err := c.findExistingCSCluster(ctx, existingCluster)
```

backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
```diff
@@ -0,0 +1,202 @@
...
+	existingCluster.ServiceProviderProperties.ClusterServiceID = clusterServiceID
+	_, err = clusterCRUD.Replace(ctx, existingCluster, nil)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to update cluster with ClusterServiceID: %w", err))
+	}
+
+	return nil
+}
+
+func (c *clusterServiceCreateSyncer) findExistingCSCluster(ctx context.Context, cluster *api.HCPOpenShiftCluster) (*arohcpv1alpha1.Cluster, error) {
```
machi1990, line 141:
```suggestion
func (c *clusterServiceCreateSyncer) findExistingClusterServiceCluster(ctx context.Context, cluster *api.HCPOpenShiftCluster) (*arohcpv1alpha1.Cluster, error) {
```

backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
```diff
@@ -0,0 +1,202 @@
...
+
+	subscription, err := c.cosmosClient.Subscriptions().Get(ctx, cluster.ID.SubscriptionID)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to get subscription: %w", err))
+	}
+
+	var tenantID string
+	if subscription.Properties != nil && subscription.Properties.TenantId != nil {
+		tenantID = *subscription.Properties.TenantId
+	}
```
machi1990, line 170: @mbarnes don't we have the tenantId stored somewhere in cosmos already?

backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
```diff
@@ -0,0 +1,202 @@
...
+	initialProperties := map[string]string{}
+	if c.provisionShard != "" {
+		initialProperties[ocm.CSPropertyProvisionShardID] = c.provisionShard
+	}
+	if c.noopProvision {
+		initialProperties[ocm.CSPropertyNoopProvision] = ocm.CSPropertyEnabled
+	}
+	if c.noopDeprovision {
+		initialProperties[ocm.CSPropertyNoopDeprovision] = ocm.CSPropertyEnabled
+	}
```
machi1990, line 181: In my opinion, we don't need to carry these and we can remove them

internal/ocm/convert.go
machi1990, line 806: @mbarnes isn't the tenantID stored already? Can we store it? Or we safe to assume that we'll have it from the subscription always?

backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// Use the Cincinnati-resolved desired version instead of the
+	// customer's minor version so CS gets the exact patch release.
+	clusterCopy := *cluster
+	clusterCopy.CustomerProperties.Version.ID = desiredVersion.String()
+
+	csClusterBuilder, csAutoscalerBuilder, err := ocm.BuildCSCluster(
+		clusterCopy.ID, tenantID, &clusterCopy, initialProperties, nil,
+	)
```
machi1990, line 190: I wonder if it'll have been less confusing to pass the desired x.y.z version as a parameter as well?

machi1990 (changes requested):
Did an initial review, left some comments.

machi1990 (changes requested):
@JakobGray I see that some changes to allow the cluster id to be missing are in here; let's sync those with the changes in https://github.com/Azure/ARO-HCP/pull/4752 as well

backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
mbarnes: +1 to dropping this baggage.

mbarnes (commented):
In order for the frontend and backend images to stay compatible with a +/-1 version skew, this probably needs to be split into multiple pull requests.

Consider if we introduce this as is but the frontend and backend images don't get deployed simultaneously for some reason.  We could potentially be in a situation where neither the frontend nor backend pods are making the CS call for cluster creation.

The first pull request should introduce the new backend controller but leave in place the CS call in the frontend.  So the new controller will initially be dormant.

Once that's fully deployed, a second pull request can remove the CS call in the frontend, at which point the backend controller will take over.

backend/pkg/controllers/createcontrollers/cluster_service_create_controller.go
```diff
@@ -0,0 +1,201 @@
...
+func (c *clusterServiceCreateClusterSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	logger := utils.LoggerFromContext(ctx)
+
+	cluster, err := c.cosmosClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
+	}
```
machi1990: If the provisioning state of the existing cluster is not `Accepted` we can ignore it.

backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
```diff
@@ -0,0 +1,283 @@
...
+		// Recovery: cluster document was updated with ClusterServiceID but the operation
+		// write failed or lagged. Only patch the operation when it still has no InternalID.
+		if len(operation.InternalID.String()) > 0 {
+			if strings.EqualFold(operation.InternalID.String(), csInternalIDFromCluster.String()) {
+				return nil
+			}
+			return fmt.Errorf("cluster create dispatch: operation internalId %q does not match cluster clusterServiceID %q",
+				operation.InternalID.String(), csInternalIDFromCluster.String())
+		}
+		operation.InternalID = *csInternalIDFromCluster
```
machi1990: Why is it important that we've to do this?

JakobGray: To recover from a partial write where the cluster already has ClusterServiceID but the operation doesn’t. This stops the dispatch controller from redoing work.

backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
```diff
@@ -0,0 +1,283 @@
...
+	if operation.ExternalID == nil || !strings.EqualFold(operation.ExternalID.ResourceType.String(), api.ClusterResourceType.String()) {
+		return false
+	}
+	if len(operation.InternalID.String()) > 0 {
+		return false
+	}
+	return true
+}
+
+func (c *dispatchClusterCreate) SynchronizeOperation(ctx context.Context, key controllerutils.OperationKey) error {
```
miguelsorianod: At the end we decided that the asynchronous work will be based on the resource and not on the operation. Let's change the controllers to honor that (excluding the operation_* ones)

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,283 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: nitpick: let's use the same directory and naming conventions as we are doing for nodepool async create: https://github.com/Azure/ARO-HCP/pull/5125

backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
machi1990: Okay. I didn't anticipate that we are also tracking the cluster service id in the operation obj. In my opinion, we could have lived without that and read it from the cluster obj wherever it is needed. The need to dual write complicates things e.g having code that reconciles on the partial write etc
 
But it predates this PR, so I am fine with that. Do you know why this is tracked in the operation as well?

machi1990: >But it predates this PR, so I am fine with that. 

@miguelsorianod mention to me that we are moving away from needing to track the CS' internal id for this. @miguelsorianod do you think we need to handle that here or it'll be dealt with when that move is made happen? I leaning towards the later but wanted to confirm the expectations

miguelsorianod: We can deal with it here. The idea is we change the operation create controller to stop using the operation InternalID and we query the cosmos resource instead to get that information.

See https://github.com/Azure/ARO-HCP/pull/4821/ which contains an implementation of moving nodepool creation in its end state. That PR just showcases the end state, but this change I am mentioning can be done initially as it shouldn't have an effect to the behavior even in old code.

backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
miguelsorianod: To give a bit more context on this:

This is because we want to continue acting independently on the operation there and independently on the operation state.

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -130,13 +130,20 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
 		return nil // no work to do
 	}
 
-	if len(operation.InternalID.String()) == 0 {
-		// we cannot proceed: yet.
-		// TODO when we update to make clusterserice creation async, we need https://github.com/Azure/ARO-HCP/pull/4695 or similar
-		// and we need to wire up a fail-safe where if we have no ID and we time out, we report the best failure we can.
-		return nil
+	clusterServiceID := operation.InternalID
```
machi1990: Can we start reading from `cluster.ServiceProviderProperties.ClusterServiceID` from the get go?

miguelsorianod: This controller shouldn't rely on operation for that anymore. See the https://github.com/Azure/ARO-HCP/pull/5125 PR on how we approach it

machi1990: I agree.

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	if subscription.Properties == nil || subscription.Properties.TenantId == nil || *subscription.Properties.TenantId == "" {
+		return utils.TrackError(fmt.Errorf("subscription %s has no tenant id", key.SubscriptionID))
+	}
+	tenantID := *subscription.Properties.TenantId
+
+	mrg := cluster.CustomerProperties.Platform.ManagedResourceGroup
+	if mrg == "" {
```
miguelsorianod: This check is not strictly needed because we should be guaranteed the attribute to be set correctly. If not, that's a programming error

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	if existingServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion == nil {
+		logger.Info("DesiredVersion not yet set, waiting for desiredversion controller")
+		return nil
+	}
+
+	subscription, err := c.resourcesDBClient.Subscriptions().Get(ctx, key.SubscriptionID)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	if subscription.Properties == nil || subscription.Properties.TenantId == nil || *subscription.Properties.TenantId == "" {
```
miguelsorianod: This check is not strictly needed. If the subscription doesn't have a tenant id that's a programming error

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+		key.HCPClusterName,
+		tenantID,
+		mrg,
+	)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	var csCluster *arohcpv1alpha1.Cluster
+	if existingCSCluster != nil {
```
miguelsorianod: To avoid if-else we can flip the condition:


if existingCSCluster == nil {
  buildCS
  existingCSCluster = postCS
}

// from this point on, assume existingCSCluster has been found, so we only need to replace

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	if err := it.GetError(); err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+func (c *clusterClusterServiceCreateSyncer) createClusterServiceCluster(ctx context.Context, cluster *api.HCPOpenShiftCluster, desiredVersion *semver.Version) (*arohcpv1alpha1.Cluster, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	subscription, err := c.resourcesDBClient.Subscriptions().Get(ctx, cluster.ID.SubscriptionID)
```
miguelsorianod: This is doing an extra subscriptions get. We already have it from syncOnce. We can pass it here

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to get subscription: %w", err))
+	}
+
+	var tenantID string
+	if subscription.Properties != nil && subscription.Properties.TenantId != nil {
```
miguelsorianod: Similar comment where we don't really need to check this. Not being set would be a programming error

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+		tenantID = *subscription.Properties.TenantId
+	}
+
+	// Use the Cincinnati-resolved desired version instead of the
+	// customer's minor version so CS gets the exact patch release.
+	clusterCopy := *cluster
```
miguelsorianod: This is not really doing a full copy. Use DeepCopy instead

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	if !c.needsWork(cluster) {
+		return nil
+	}
+
+	existingServiceProviderCluster, err := database.GetOrCreateServiceProviderCluster(ctx, c.resourcesDBClient, cluster.ID)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get or create ServiceProviderCluster: %w", err))
+	}
+
+	if existingServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion == nil {
```
miguelsorianod: non blocker: We could extract this into a function making it clear it's a gate/precondition to continue. This will signal future developers on how to add this and will give a semantic meaning.

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
JakobGray: Updated

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
JakobGray: Updated with latest commit. We lose the comment about adopting an existing cluster but it does improve the readability

machi1990 (commented):
lgtm

leaving the final /lgtm tagging and approval to @miguelsorianod

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
JakobGray: I've framed it as a helper method similar to `deletePreconditionCosmosChildResourcesDeleted`

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	csInternalID, err := api.NewInternalID(csCluster.HREF())
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	logger.Info("Storing ClusterServiceID on cluster document", "clusterServiceID", csInternalID.String())
+	cluster.ServiceProviderProperties.ClusterServiceID = &csInternalID
+	_, err = c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Replace(ctx, cluster, nil)
```
miguelsorianod, line 148: In our controllers, recently we started to change the behavior where when the replace fails because of etag mismatch (precondition failed) we return nil instead of erroring. Let's change it here too. It would look like:

```
_, err = Replace(...)
	if database.IsPreconditionFailedError(err) {
		// if we have a conflict error, then we're guaranteed that our informer will eventually see an update and trigger us again.
		return nil
	}
	if err != nil {
		return utils.TrackError(err)
	}
```

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if !c.needsWork(cluster) {
+		return nil
+	}
+
+	cluster, err = c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
```
miguelsorianod, line 94: Let's add a code comment here where we mention why we retrieve from cosmos instead of uniquely relying on the lister one. Something like:

For the Cluster, we retrieve from the actual database because we are about to use its data to interact with cluster-service.

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// createPreconditionDesiredVersionResolved reports whether the ControlPlaneDesiredVersion
+// controller has written the Cincinnati-resolved desired version to the ServiceProviderCluster.
+// Returns (false, nil) when this controller should wait and retry.
+func (c *clusterClusterServiceCreateSyncer) createPreconditionDesiredVersionResolved(ctx context.Context, serviceProviderCluster *api.ServiceProviderCluster) (bool, error) {
+	if serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion != nil {
+		return true, nil
+	}
+	utils.LoggerFromContext(ctx).Info("DesiredVersion not yet set, waiting for ControlPlaneDesiredVersion controller")
```
miguelsorianod: nitpick: in our codebase it's more idiomatic to create a variable for the logger first at the top of the function and then use it:

...
logger.info(...)

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
```diff
@@ -0,0 +1,339 @@
...
+	clusterInternalID := api.Must(api.NewInternalID(testClusterServiceIDStr))
+
+	tests := []struct {
+		name           string
+		listCluster    *api.HCPOpenShiftCluster // cluster seeded into the lister (nil = not found)
+		dbCluster      *api.HCPOpenShiftCluster // cluster stored in the DB
+		desiredVersion *semver.Version          // nil = no ServiceProviderCluster desiredVersion set
+		setupMockCS    func(ctrl *gomock.Controller) ocm.ClusterServiceClientSpec
+		expectError    bool
+		verify         func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient)
```
miguelsorianod: nitpick: rename to verifyDB as it is restricted to DB level checks

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
```diff
@@ -0,0 +1,339 @@
...
+
+		s := &clusterClusterServiceCreateSyncer{clustersServiceClient: mock}
+		got, err := s.findAROHCPClusterByAzureInfo(ctx, sub, rg, resName, tenant, mrg)
+		require.NoError(t, err)
+		require.Nil(t, got)
+	})
+
+	t.Run("multiple matches error", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
```
miguelsorianod: no need call ctrl.Finish in any test

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
```diff
@@ -0,0 +1,339 @@
...
+	rg := "my-rg"
+	resName := "MyCluster"
+	tenant := "bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb"
+	mrg := "arohcp-mycluster-uuid"
+
+	wantSearch := "azure.subscription_id = '" + strings.ToLower(sub) + "' and azure.resource_group_name = '" + strings.ToLower(rg) + "' and azure.resource_name = '" + strings.ToLower(resName) + "'" +
+		" and azure.tenant_id = '" + tenant + "'" +
+		" and azure.managed_resource_group_name = '" + mrg + "'"
+
+	t.Run("found on primary search", func(t *testing.T) {
```
miguelsorianod: The N t.Run here can be redesigned to be table-based like with the other test. the test args would be something like: name, setupMockCS, want, wantErr or similar

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
```diff
@@ -0,0 +1,339 @@
...
+			ctx := context.Background()
+			ctx = utils.ContextWithLogger(ctx, testr.New(t))
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			subscription := newTestSubscription()
+			mockDB, err := databasetesting.NewMockResourcesDBClientWithResources(ctx, []any{subscription, tt.dbCluster})
+			require.NoError(t, err)
+
+			if tt.desiredVersion != nil {
```
miguelsorianod: Having this boolean works now but it won't escalate well if in the future this controller gets increased complexity or responsibilities, if we need to set different data for the serviceProviderCluster, etc.... By keeping this boolean then in the future other people will add other booleans, strings, etc... which will end up with a tangled set of condition checks, making the test body hard to reason and the tests hard to maintain

Conceptually, what we want here is to set the serviceprovidercluster. Have a existingServiceProviderCluster attribute for each test case, and then here in the body just add it into the mockDB when not nil.

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
```diff
@@ -0,0 +1,339 @@
...
+	return &arm.Subscription{
+		CosmosMetadata: api.CosmosMetadata{ResourceID: rid},
+		ResourceID:     rid,
+		Properties:     &arm.SubscriptionProperties{TenantId: ptr.To(testTenantID)},
+	}
+}
+
+// addTestServiceProviderCluster inserts a ServiceProviderCluster document with
+// the given desiredVersion into the mock DB.
+func addTestServiceProviderCluster(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient, desiredVersion *semver.Version) {
```
miguelsorianod: See https://github.com/Azure/ARO-HCP/pull/4821/changes#r3427899405. That will probably make this function disappear, and we end up with a helper  that gives you a base serviceprovidercluster with ability to modify it (similar to newTestCluster)

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -114,9 +114,13 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
 	if !c.ShouldProcess(ctx, operation) {
 		return nil // no work to do
 	}
-	if len(operation.InternalID.String()) == 0 {
-		// we cannot proceed: yet.
-		// TODO when we update to make clusterserice creation async, we need to handle this correctly.
+
+	cluster, err := getClusterForOperation(ctx, c.resourcesDBClient, operation)
```
miguelsorianod: I don't think we need a new method for just the get.
Let's just call here the cluster get

backend/pkg/controllers/operationcontrollers/utils.go
```diff
@@ -427,10 +427,28 @@ func PostAsyncNotification(ctx context.Context, notificationClient *http.Client,
 	return nil
 }
 
+func getClusterForOperation(ctx context.Context, db database.ResourcesDBClient, operation *api.Operation) (*api.HCPOpenShiftCluster, error) {
+	if operation.ExternalID == nil {
```
miguelsorianod: No need to check this because it's already checked in the ShouldProcess method

backend/pkg/controllers/operationcontrollers/utils.go
```diff
@@ -427,10 +427,28 @@ func PostAsyncNotification(ctx context.Context, notificationClient *http.Client,
...
+		return nil, fmt.Errorf("operation has no ExternalID")
+	}
+	return db.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).Get(ctx, operation.ExternalID.Name)
+}
+
+func clusterServiceIDFromCluster(cluster *api.HCPOpenShiftCluster) (api.InternalID, bool) {
```
miguelsorianod: Follow similar approach to https://github.com/Azure/ARO-HCP/pull/5125/ with the shouldReconcileOperationAndResourceStatus method when checking for CSID. This will allow us to put other gating conditions in the future and structure it similarly

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -123,19 +120,14 @@ func (c *controlPlaneDesiredVersionSyncer) SyncOnce(ctx context.Context, key con
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
```
miguelsorianod, line 137: Asking just in case:

Is it safe to remove this check? How do we control that this controller does a noop if it doesn't have the needed data?

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -130,13 +130,21 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
...
+	cluster, err := c.clusterLister.Get(ctx, operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName, operation.ExternalID.Name)
+	if database.IsNotFoundError(err) {
+		logger.Info("cluster not yet in cache, waiting")
 		return nil
 	}
-	clusterStatus, err := c.clusterServiceClient.GetClusterStatus(ctx, operation.InternalID)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get cluster to resolve ClusterServiceID: %w", err))
+	}
+	clusterServiceID, ok := clusterServiceIDFromCluster(cluster)
```
miguelsorianod: Similar comment to the update operation controller

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
miguelsorianod: Actually, I think none of the changes in this operation controller are strictly needed for the cluster create functionality. They can be addressed in the corresponding cluster update functionality, which is already underway, and contains those changes: https://github.com/Azure/ARO-HCP/pull/5593/

Let me know if that's not the case for some reason and they must be included with the create cluster functionality

backend/pkg/controllers/operationcontrollers/operation_cluster_delete_legacy.go
```diff
@@ -17,6 +17,7 @@ package operationcontrollers
 import (
```
miguelsorianod, line 17: The legacy one shouldn't be changed as we want to keep the previous behavior, which is its purpose.

The problem comes from trying to change the convertClusterStatus reused function and reuse it everywhere

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
miguelsorianod: ah, I see why you did it. It's because you are trying to reuse the convertClusterStatus and change its signature everywhere. This is problematic because the cluster delete legacy code shouldn't change its behavior (https://github.com/Azure/ARO-HCP/pull/4821/changes#r3453691500)

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+
+	ready, err := c.createPreconditionDesiredVersionResolved(ctx, existingServiceProviderCluster)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	if !ready {
+		return nil
+	}
+
+	subscription, err := c.resourcesDBClient.Subscriptions().Get(ctx, key.SubscriptionID)
```
miguelsorianod: Use a subscription lister. This information is unlikely to change.

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	tenantID := *subscription.Properties.TenantId
+	mrg := cluster.CustomerProperties.Platform.ManagedResourceGroup
+
+	csCluster, err := c.findAROHCPClusterByAzureInfo(ctx,
```
miguelsorianod: No need to newline. We accept long lines in the aro-hcp project

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -131,13 +131,21 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
...
 
-	if len(operation.InternalID.String()) == 0 {
-		// we cannot proceed: yet.
-		// TODO when we update to make clusterserice creation async, we need https://github.com/Azure/ARO-HCP/pull/4695 or similar
-		// and we need to wire up a fail-safe where if we have no ID and we time out, we report the best failure we can.
+	cluster, err := c.clusterLister.Get(ctx, operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName, operation.ExternalID.Name)
+	if database.IsNotFoundError(err) {
+		logger.Info("cluster not yet in cache, waiting")
 		return nil
 	}
```
miguelsorianod, line 141: Let's add an additional check that if the ActiveOperationID attribute of the cluster doesn't match the operation's operation we log that occurence and we return without error

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -131,13 +131,21 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
 		return nil // no work to do
 	}
 
-	if len(operation.InternalID.String()) == 0 {
-		// we cannot proceed: yet.
-		// TODO when we update to make clusterserice creation async, we need https://github.com/Azure/ARO-HCP/pull/4695 or similar
-		// and we need to wire up a fail-safe where if we have no ID and we time out, we report the best failure we can.
+	cluster, err := c.clusterLister.Get(ctx, operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName, operation.ExternalID.Name)
+	if database.IsNotFoundError(err) {
+		logger.Info("cluster not yet in cache, waiting")
```
miguelsorianod: nitpick: change "cluster not yet in cache, waiting" for "cluster not found in cache, waiting".

Let's not make assumptions about whether it will be eventually there or not.

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
miguelsorianod: Use active operation lister to read the operation information. Reserve the resourceDBClient only to perform write on it

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
miguelsorianod: something like:

```
		logger.Info("cluster active operation id mismatch, returning early", "synchronizedActiveOperationID", operation.ResourceID.Name, "clusterActiveOperationID", cluster.ServiceProviderProperties.ActiveOperationID)
```

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	if !ready {
+		return nil
+	}
+
+	subscription, err := c.subscriptionLister.Get(ctx, key.SubscriptionID)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	if subscription.Properties == nil || subscription.Properties.TenantId == nil {
```
miguelsorianod, line 123: No need to check this. Subscriptions should be created correctly in Cosmos. Otherwise it's a programming error

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -131,13 +134,28 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
...
+	cluster, err := c.clusterLister.Get(ctx, operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName, operation.ExternalID.Name)
+	if database.IsNotFoundError(err) {
+		logger.Info("cluster not found in cache, waiting")
 		return nil
 	}
-	clusterStatus, err := c.clusterServiceClient.GetClusterStatus(ctx, operation.InternalID)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get cluster to resolve ClusterServiceID: %w", err))
+	}
+	if operation.OperationID != nil &&
```
miguelsorianod: no need to check  operation.OperationID != nil . operations should always have the ID

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -131,13 +134,28 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
...
+		cluster.ServiceProviderProperties.ActiveOperationID != operation.OperationID.Name {
+		logger.Info("cluster active operation id mismatch, returning early",
```
miguelsorianod, line 146: We can have long line

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -131,13 +134,28 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
...
+	if database.IsNotFoundError(err) {
+		logger.Info("cluster not found in cache, waiting")
 		return nil
 	}
-	clusterStatus, err := c.clusterServiceClient.GetClusterStatus(ctx, operation.InternalID)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get cluster to resolve ClusterServiceID: %w", err))
+	}
+	if operation.OperationID != nil &&
+		cluster.ServiceProviderProperties.ActiveOperationID != operation.OperationID.Name {
```
miguelsorianod: nitpick: put operation.OperationID.Name on the left side of the comparison, to be consistent with how it's being written in the other equivalent operation create controllers

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
JakobGray: I removed it but added it back because Copilot kept complaining

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
JakobGray: We don't want the controller to noop if ClusterServiceID is not set. It will run against the cluster before that is set

backend/pkg/controllers/operationcontrollers/operation_cluster_delete_legacy.go
```diff
@@ -115,7 +115,7 @@ func (c *operationClusterDelete) legacySynchronizeOperation(ctx context.Context,
 		return utils.TrackError(err)
 	}
 
-	newOperationStatus, newOperationError, err := convertClusterStatus(ctx, c.clusterServiceClient, operation, clusterStatus)
+	newOperationStatus, newOperationError, err := convertClusterStatus(ctx, c.clusterServiceClient, operation, clusterStatus, operation.InternalID)
```
miguelsorianod, line 118: 👍 . That's good enough to make sure there's no change in behavior in the legacy approach.

backend/pkg/controllers/operationcontrollers/operation_cluster_create_test.go
```diff
@@ -55,16 +55,21 @@ func TestOperationClusterCreate_SynchronizeOperation(t *testing.T) {
...
 	tests := []struct {
-		name         string
-		clusterState arohcpv1alpha1.ClusterState
-		createdAt    *time.Time
-		expectError  bool
-		verify       func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient, fixture *clusterTestFixture)
+		name                      string
+		clusterState              arohcpv1alpha1.ClusterState
+		createdAt                 *time.Time
+		clearOperationInternalID  bool
```
miguelsorianod: non blocker but to improve: adding N booleans in the test case args won't escalate well. As new needs and responsibilities appear for the controller more and more would be added, making the tests brittle and hard to reason about.

At the end, what the test needs is:
existingCluster
existingOperation
setupCSMock

which changes depending on the test case.

This makes it scalable and removes the N args and convoluted test body.

Also nitpick: let's rename verify to verifyDB following how other tests name it. We make specific that that's only about DB specific checks.

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
```diff
@@ -0,0 +1,371 @@
...
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// Test constants
+const (
+	testSubscriptionID      = "00000000-0000-0000-0000-000000000000"
+	testResourceGroupName   = "test-rg"
+	testClusterName         = "test-cluster"
+	testClusterServiceIDStr = "/api/clusters_mgmt/v1/clusters/abc123"
```
miguelsorianod: The CS id str doesn't adjust to reality. We use aro-hcp CS endpoints. To fix

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	ctx := context.Background()
+	sub := "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
+	rg := "my-rg"
+	resName := "MyCluster"
+	tenant := "bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb"
+	mrg := "arohcp-mycluster-uuid"
+
+	wantSearch := "azure.subscription_id = '" + strings.ToLower(sub) + "' and azure.resource_group_name = '" + strings.ToLower(rg) + "' and azure.resource_name = '" + strings.ToLower(resName) + "'" +
```
miguelsorianod: Use clustersServiceClusterByAzureInfoSearchString. No need to duplicate.

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
```diff
@@ -0,0 +1,368 @@
...
+			}
+
+			if tt.verifyDB != nil {
+				tt.verifyDB(t, ctx, mockDB)
+			}
+		})
+	}
+}
+
+func TestClusterClusterServiceCreate_findAROHCPClusterByAzureInfo(t *testing.T) {
```
miguelsorianod, line 279: I think testing this a bit deeper might be worth it:
* the lowercasing part on different attrs
* some other combinations that end up in not found (cs returns info but the resourcename differs, or the rgname differs etc...

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
miguelsorianod, line 185: Before this error check add a preconditionfailed check and return nil in that case. See the opreation nodepool create to see how it's been done

miguelsorianod:
rebased solving conflicts

PR #4827, opened 2026-04-10
clusters-service: bump liveness and readiness probe timeouts

stevekuznetsov:
We see p95 latencies for clusters-service under load often getting close to 10 seconds, and we see chronic probe failures for clusters-service during end-to-end tests. We need to give these queries more time so the service is less likely to be killed during a run from lack of liveness or readiness.


geoberle:
/override "Ensure PR is submitted from Azure/ARO-HCP"

PR #4828, opened 2026-04-10
cfg: updating acm to 216 and 211

sclarkso:
### What

Update acm to minor versions 2.16 and 2.11.

### Why

https://github.com/Azure/ARO-HCP/pull/4595


raelga:
Will be implemented in https://github.com/Azure/ARO-HCP/pull/4595

PR #4830, opened 2026-04-10
backend: additional structured logging

geoberle:
### What

* add `internal_id` field to all operations related logging


geoberle:
closing. apparently internalId is not always set

PR #4833, opened 2026-04-10
Switch demo cluster to swift

mmazur:
Default should be in line with what we use. 

Tested, I get a cluster up with `hypershift.openshift.io/swift-pod-network-instance` annotation.


bennerv:
closing in favor of #4874

PR #4836, opened 2026-04-10
Upgrade to aks 1.33.8

tmstff:
https://redhat.atlassian.net/browse/ARO-25883

### What

Upgrade AKS version to 1.33.8 for svc and mgmt.
Our current approach was not upgrading nodepool, but this PR adresses it with the `az aks upgrade` step in both mgmt and svc pipeline. This is a preliminary solution until we have AKS Fleet in place.

### Why

We want to upgrade to AKS 1.35 to close possible gaps to the latest version - and we need to do it step by step.
1.33.x (x>8) will have a fix we need for swift

### Testing

manual tests confirmed the functionality works.
prow E2E tests will not trigger real upgrades because the install version matches already the target version enforced by the upgrade script. in any case it is a noop

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -406,7 +406,7 @@ defaults:
       vnetAddressPrefix: "10.128.0.0/14"
       subnetPrefix: "10.128.8.0/21"
       podSubnetPrefix: "10.128.64.0/18"
-      kubernetesVersion: 1.33.7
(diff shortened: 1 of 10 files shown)


geoberle:
/test e2e-parallel
`RoleAssignmentLimit` issue

PR #4854, opened 2026-04-13
ACR caching of operator catalog images

geoberle:
### What

* introduce a pull secret to operator catalog images stored on quay.io/redhat-prod instead of registry.redhat.io
* introduce caching rules for the individual images and mapping them to the redhat-prod registry in the ACR

ACR has a limitation of defining caching rules for existing repositories. this requires us to change the repository we use from `redhat` to `rrio/redhat` (needs to be different from`redhat`).

in order to leverage this, there CS need to adapt to it and until then we can't remove the existing `redhat` repository nor the mirror job in the azure container instance.

breadcrumbs:
* pull secret - https://gitlab.cee.redhat.com/exd/quay/quay-access-management/-/merge_requests/496


geoberle:
/hold
image pull fails with these creds
```
quay.io/redhat-prod/redhat/redhat-operator-index:v4.20
```
investigating if the image ref is valid

geoberle:
/unhold

correct image ref `quay.io/redhat-prod/redhat----redhat-operator-index:v4.21`

ahitacat:
I have tested this with a change in CS using the new olmCatalog registry `rrio/redhat` and it worked well. 
LGTM :tada:

PR #4862, opened 2026-04-13
Fix: return 404 when attempting invalid listing in dev environment

mvacula02:
fixes [ARO-25089](https://redhat.atlassian.net/browse/ARO-25089) and enables [ARO-22570](https://redhat.atlassian.net/browse/ARO-22570), [ARO-22571](https://redhat.atlassian.net/browse/ARO-22571)

### What

Implement a fix to error handling after invalid listing. Currently, an internal server occurs instead of the expected ResourceNotFound or similar in the dev environment. Investigation shows that frontend's writeError couldn't extract a resource ID from list endpoint URLs. ~~PR currently only enables the test cases to observe the CI behaviour first. Afterwards a bug fix will be implemented.~~

~~The fix adds a ResourceGroupChecker that queries Azure for resource group existence. In the dev environment, there is no ARM proxy layer in front of the RP to validate resource groups before requests reach the frontend. When a list handler encounters a "not found" condition (empty results for clusters, or Cosmos 404 for the parent cluster in node pool/external auth cases), it now checks whether the resource group actually exists. If not, it returns ResourceGroupNotFoundError (404). If the resource group exists but the parent cluster doesn't (node pool and external auth listing), it returns ParentResourceNotFoundError (404). These are proper *CloudError values that the error handler (internal/errorutils/error.go) handles directly, avoiding the ResourceIDFromContext fallback that fails on list endpoints (where no single resource ID exists in the context) and causes an internal server error. The behaviour should now be consistent with higher environments.~~

Front end changes will not be implemented, instead The PR adds an armResourceGroupValidationPolicy to the E2E test harness that simulates ARM's resource group validation in development environment. For every request targeting the RP frontend, the policy checks whether the resource group exists in Azure and returns a 404 if it doesn't, the same behavior ARM provides in production. This enables the [ARO-22570](https://redhat.atlassian.net/browse/ARO-22570) and [ARO-22571](https://redhat.atlassian.net/browse/ARO-22571) test cases for listing clusters and node pools in nonexistent resource groups, without any changes to the frontend.

Scenario removed: listing a resource with non-existent parent. Removed as it can't be implemented without significant front end changes. 

Note: ~~tests will be enabled in #4876 as the frontend changes need to populate first~~ #4876 will be closed as the tests can be enabled within the PR as the fix now

### Why

Implement missing error handling and unblock negative listing cases both for clusters and node pools. 

### Testing

Tests were initially blocked by the bug. Enabling the tests again and having the run succeed should warrant the bug fix to work. 

Bug fix was verified through successful listing test runs in personal dev environment ~~with a locally built frontend~~.


mvacula02:
- [x] Remove MustFilter after test validation in Stage

internal/api/arm/error.go
```diff
@@ -272,6 +273,27 @@ func WriteResourceNotFoundError(w http.ResponseWriter, resourceID *azcorearm.Res
...
+}
+
+// NewParentResourceNotFoundError creates a CloudError for when a child resource
+// operation fails because the parent resource does not exist.
+func NewParentResourceNotFoundError(parentResourceID *azcorearm.ResourceID, childResourceType azcorearm.ResourceType) *CloudError {
+	return NewCloudError(
+		http.StatusNotFound,
+		CloudErrorCodeParentResourceNotFound,
+		parentResourceID.String(),
+		"Failed to perform 'read' on resource(s) of type '%s', because the parent resource '%s' could not be found.",
```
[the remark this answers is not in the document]
mvacula02: the point is to match the exact error messages that ARM uses in higher environments.

mbarnes (changes requested):
In your pull request description you say:
> In the dev environment, there is no ARM proxy layer in front of the RP to validate resource groups before requests reach the frontend.

I understand that to mean `azureResourceGroupChecker` is only applicable to development environments, not to environments where ARM is present because a collection GET request for a non-existent resource group would never even get routed to the RP.

I think this is important to capture in the source code so it doesn't get removed down the road by someone thinking "this can never happen in production, why do we need it?"

Could you add a comment somewhere explaining this?  Wherever you think appropriate.

geoberle:
i don't feel comfortable with this change. we are moving something that ARM is doing into the RP and it is painful for the things where we do it.
have you considered replicating ARMs behavior in our "AroRpApiCompatible" test harness? (`armSystemDataPolicy`) 

@bennerv @mbarnes wdyt?

mvacula02:
> have you considered replicating ARMs behavior in our "AroRpApiCompatible" test harness? (`armSystemDataPolicy`)

I believe that would be possible only for the non-existent listings. The case, where a parent resource is non-existent would need to be dropped. That is fine with me if you think the current approach is too much

mbarnes:
>  have you considered replicating ARMs behavior in our "AroRpApiCompatible" test harness? (armSystemDataPolicy)

If the negative test cases you've added to #4876 are deemed worth keeping, maybe just split them out in such a way that they don't have the `AroRpApiCompatible` label since they're not actually compatible without ARM present.

Those cases would then be testing more ARM behavior than RP behavior, so maybe it's appropriate to drop them entirely.

test/util/framework/azure_client_policies.go
```diff
@@ -51,6 +54,75 @@ func (p *armSystemDataPolicy) Do(req *policy.Request) (*http.Response, error) {
...
+		return nil, err
+	}
+
+	return req.Next()
+}
+
+// parseResourceGroupFromPath extracts the subscription ID and resource group
+// name from a URL path like /subscriptions/{subId}/resourceGroups/{rgName}/...
+func parseResourceGroupFromPath(path string) (subscriptionID, resourceGroupName string) {
+	parts := strings.Split(strings.ToLower(path), "/")
```
mgahagan73, line 114: Consider using strings.EqualFold() here. You will be able to drop the ToLower() step as it is case insensitive. For example:

parts := strings.Split(path, "/")                                                                                                                                                          
  for i, part := range parts {                                                                                                                                                               
      if strings.EqualFold(part, "subscriptions") && i+1 < len(parts) {                                                                                                                      
          subscriptionID = parts[i+1]                                                                                                                                                        
      }                                                                                                                                                                                      
      if strings.EqualFold(part, "resourcegroups") && i+1 < len(parts) {                                                                                                                     
          resourceGroupName = parts[i+1]                                                                                                                                                     
      }                                                                                                                                                                                      
  }

test/util/framework/azure_client_policies.go
```diff
@@ -51,6 +54,75 @@ func (p *armSystemDataPolicy) Do(req *policy.Request) (*http.Response, error) {
...
+
+	client, err := armresources.NewResourceGroupsClient(subID, p.cred, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create resource group client: %w", err)
+	}
+
+	_, err = client.Get(req.Raw().Context(), rgName, nil)
+	if err != nil {
+		var respErr *azcore.ResponseError
+		if errors.As(err, &respErr) && respErr.ErrorCode == "ResourceGroupNotFound" {
```
mgahagan73, line 89: Please replace the literal "ResourceGroupNotFound" with arm.CloudErrorCodeResourceGroupNotFound to be consistent with the rest of the code.

mgahagan73:
In addition to previous comments I think might want to add a unit test for the ARM simulator bits in parseResourceGroupFromPath() and other functions in azure_client_policies.go . Adding it in separate PR would be fine.

miquelsi (approved):
/lgtm

PR #4863, opened 2026-04-13
chore: add e2e test for eventhub as a destination for shoebox diagnostic logs

ashishmax31:
This PR modifies the existing shoebox e2e test to also include azure eventhub as a destination for shoebox diagnostic logs

### DIFF
--- a/test/e2e/shoebox_logs.go
+++ b/test/e2e/shoebox_logs.go
@@ -22,10 +22,14 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	"golang.org/x/sync/errgroup"
+
 	"k8s.io/apimachinery/pkg/util/rand"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/eventhub/armeventhub"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/monitor/armmonitor"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage"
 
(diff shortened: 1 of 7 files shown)


ashishmax31:
Keeping it draft because i'm waiting for stage e2e to pass.

test/cmd/aro-hcp-tests/main.go
```diff
@@ -200,7 +200,7 @@ func setupCli() *cobra.Command {
 	// 	specs = specs.AddLabel("SLOW")
 
 	// Specs can be globally filtered...
-	// specs = specs.MustFilter([]string{`name.contains("filter")`})
+	specs = specs.MustFilter([]string{`name.contains("shoebox")`})
```
ashishmax31: Adding this filter for quick feedback from stage e2e. Will remove once stage e2e passes.

test/e2e/shoebox_logs.go
```diff
@@ -34,8 +38,171 @@ import (
...
+	}
+
+	keysResp, err := nsClient.ListKeys(ctx, resourceGroupName, namespaceName, authRuleName, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list Event Hub namespace keys: %w", err)
+	}
+
+	return &eventHubResult{
+		AuthorizationRuleID: *authRuleResp.ID,
+		ConnectionString:    *keysResp.PrimaryConnectionString,
```
janboll, line 199: check for nil

test/e2e/shoebox_logs.go
```diff
@@ -34,8 +38,171 @@ import (
...
+		"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/%s",
+		subscriptionID, resourceGroupName, clusterName,
+	)
+}
+
+// pollVerifier polls a verifier at 60-second intervals until it succeeds or the timeout is reached.
+func pollVerifier(ctx context.Context, name string, verifier shoeboxLogVerifier, timeout time.Duration) bool {
+	ticker := time.NewTicker(60 * time.Second)
+	defer ticker.Stop()
+	timer := time.After(timeout)
```
janboll, line 84: The timer is never cancelled if the context completes first. Use context.WithTimeout instead.

test/e2e/shoebox_logs.go
```diff
@@ -34,8 +38,171 @@ import (
...
+type eventHubResult struct {
+	AuthorizationRuleID string
+	ConnectionString    string
+	EventHubName        string
+}
+
+func createEventHub(ctx context.Context, subscriptionID string, creds azcore.TokenCredential, resourceGroupName, location string) (*eventHubResult, error) {
+	const (
+		namespaceName = "shoebox-eh-ns"
+		hubName       = "shoebox-eh"
```
janboll, line 145: should this be randomized?

test/e2e/shoebox_logs.go
ashishmax31: That is unecessary from go 1.23. From go doc:
"Before Go 1.23, this documentation warned that the underlying [Timer](https://pkg.go.dev/time#Timer) would not be recovered by the garbage collector until the timer fired, and that if efficiency was a concern, code should use NewTimer instead and call [Timer.Stop](https://pkg.go.dev/time#Timer.Stop) if the timer is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers."

test/e2e/shoebox_logs.go
ashishmax31: There is no need to randomize this as they are scoped under a RG (and the RG is randomized)

test/e2e/shoebox_logs.go
ashishmax31: Is that necessary? There are error checks before which catch and return the errors if either fetching `authRuleResp` or 
`keysResp`fails.

janboll: In general, pointer de referencing should be checked. There could be a bug in the sdk or other weird conditions. Perhaps for this case it's okay?
Please check and dismiss if you think it's okay

janboll (commented):
/lgtm

PR #4864, opened 2026-04-13
istio: bump istio for dev envs to asm-1-28

bennerv:
### What

Broken swift/dev environments which don't have istio asm-1-28 revision defined.

```
az aks mesh get-revisions -l uksouth | jq -r '.[].[].revision'
asm-1-27
asm-1-28
asm-1-29
```
### Why

We need to continuously upgrade istio to ensure it's not out of support.  This should not impact any INT/Stage/Prod envs as they have their own defaults defined in config in sdp-pipelines.  

### Testing

Ran `DEPLOY_ENV=swft make entrypoint/Region` locally.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1161,8 +1161,8 @@ clouds:
         nsp:
           accessMode: 'Learning'
         istio:
-          targetVersion: "asm-1-26"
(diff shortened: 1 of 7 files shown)


bennerv:
ugh i gotta submit from a non-fork...

PR #4845, opened 2026-04-10
backend: improve visibility into NodePool state

stevekuznetsov:
backend/controllers: dump CS node pool state

As a minimal amount of insight into the CS state behind node pools, we
can query for them when we're dumping CS cluster state. Of course, a
node-pool-watching controller might be more appropriate, but a minimum
amount of visibility is still better than what we have now (none).

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

backend: add Maestro read-only bundles for NodePools

We need to have visibility into the NodePools that HyperShift handles
for us just as much as we do for the HostedClusters.

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

PR #4866, opened 2026-04-14
backend/controllers: dump CS node pool state

geoberle:
### What

As a minimal amount of insight into the CS state behind node pools, we can query for them when we're dumping CS cluster state. Of course, a node-pool-watching controller might be more appropriate, but a minimum amount of visibility is still better than what we have now (none).

this is the CS datadump enhancement part of https://github.com/Azure/ARO-HCP/pull/4845

### DIFF
--- a/backend/pkg/controllers/datadumpcontrollers/cs_state_dump.go
+++ b/backend/pkg/controllers/datadumpcontrollers/cs_state_dump.go
@@ -110,6 +110,50 @@ func (c *csStateDump) SyncOnce(ctx context.Context, key controllerutils.HCPClust
 		"csCluster", clusterData,
 	)
 
+	// Fetch and dump node pools
+	allNodePools, err := c.cosmosClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).NodePools(key.HCPClusterName).List(ctx, nil)
+	if err != nil {
+		logger.Error(err, "failed to list node pools from cosmos for CS state dump")
+		// best effort, don't fail
+		return nil
+	}
+
+	for _, nodePool := range allNodePools.Items(ctx) {
+		npCSID := nodePool.ServiceProviderProperties.ClusterServiceID
+		if len(npCSID.String()) == 0 {
+			// No ClusterServiceID yet, node pool hasn't been registered with CS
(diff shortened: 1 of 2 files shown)


backend/pkg/controllers/datadumpcontrollers/cs_state_dump.go
```diff
@@ -110,6 +110,50 @@ func (c *csStateDump) SyncOnce(ctx context.Context, key controllerutils.HCPClust
...
+					"nodePoolClusterServiceID", npCSID.String(),
+				)
+				continue
+			}
+		}
+
+		logger.Info("cluster-service node pool state dump",
+			"clusterServiceID", csID.String(),
+			"nodePoolClusterServiceID", npCSID.String(),
+			"csNodePool", nodePoolData,
```
bennerv, line 150: do you want to update the kql queries, or are you fine with these not being pulled out as individual fields for the time being?

backend/pkg/controllers/datadumpcontrollers/cs_state_dump.go
```diff
@@ -110,6 +110,50 @@ func (c *csStateDump) SyncOnce(ctx context.Context, key controllerutils.HCPClust
 		"csCluster", clusterData,
 	)
 
+	// Fetch and dump node pools
+	allNodePools, err := c.cosmosClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).NodePools(key.HCPClusterName).List(ctx, nil)
```
bennerv, line 114: Do you care about the log ordering / consistency of the nodepools?  If not, use a `NewNodePoolWatchingController` to utilize the informers and look at the CS state from there.  Then you change this to a `cs_cluster_state_dump.go` and introduce another `cs_nodepool_state_dump.go` controller.


Or.... 

recursively list all objects and if they have corresponding CS types, dump them as well in this controller, similar to what we're doing for cosmos state.  Any reason why we shouldn't be dumping external auth representations as well?

bennerv:
/lgtm
/approve 

Some follow-up items, but getting the data dumped to logs trumps waiting for follow-up.

PR #4867, opened 2026-04-14
Removed unnecessary config for OCP versions

miquelsi:
### What

Removed OCP versions property from configuration.

### Why

As a follow up to #4717 we're removing `ocpVersions` propery, since the service relies only in data from `aro-hcp-ocp-versions-config.configmap.yaml` file.


machi1990:
Let's also remove references of these config in https://github.com/Azure/ARO-HCP/blob/efd3f894ced0385e6cd4abbfe38969e7f2e4c9c0/cluster-service/pipeline.yaml#L171

machi1990:
As well as this https://github.com/Azure/ARO-HCP/blob/efd3f894ced0385e6cd4abbfe38969e7f2e4c9c0/cluster-service/helm-charts/cluster-service/values.yaml#L325

ahitacat:
Also these in the makefile 
https://github.com/Azure/ARO-HCP/blob/main/cluster-service/Makefile#L95

raelga (approved):

PR #4872, opened 2026-04-14
Embed nodepool listing test cases into nodepool_update_nodes.go

aklymenk:
[ARO-21295](https://redhat.atlassian.net/browse/ARO-21295)

### What

Adds node pool listing validations into the existing `nodepool_update_nodes` E2E test. The listing is verified at four points in the test:

1. By resource group before the cluster is created - expect 0 node pools
2. By cluster after the cluster is created with 0 node pools - expect 0 node pools
3. By cluster and by resource group after all 3 node pools are created - expect all 3 listed
4. By cluster and by resource group after all node pools are updated - expect all 3 still listed

### Why

Node pool listing was identified as a use case that needs E2E test coverage (listing by cluster and by resource group). Instead of creating a dedicated test that deploys a new cluster solely for listing, the validations are embedded into the existing `nodepool_update_nodes` test which already creates a cluster and 3 node pools. This avoids adding cluster and nodepool creation time to the test suite.

Subscription-level listing is intentionally excluded, following the precedent set in [#4520](https://github.com/Azure/ARO-HCP/pull/4559): the subscription-level list fans out across all regions and can fail due to availability issues in unrelated regions, and the underlying code path is identical to the resource-group-level list.

### Testing

The change itself is the test. The listing helpers use `NodePoolsClient.NewListByParentPager` (by cluster) and `HcpOpenShiftClustersClient.NewListByResourceGroupPager` + `NewListByParentPager` (by resource group), both from the HCP SDK, so the `AroRpApiCompatible` label is preserved.

### DIFF
--- a/test/e2e/nodepool_update_nodes.go
+++ b/test/e2e/nodepool_update_nodes.go
@@ -16,6 +16,7 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -32,7 +33,7 @@ import (
 )
 
 var _ = Describe("Customer", func() {
-	It("should be able to update nodepool replicas and autoscaling",
+	It("should be able to update nodepool replicas and autoscaling as well as list node pools by cluster and resource group",
 		labels.RequireNothing,
 		labels.High,
 		labels.Positive,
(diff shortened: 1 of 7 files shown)


test/e2e/nodepool_update_nodes.go
```diff
@@ -69,10 +70,66 @@ var _ = Describe("Customer", func() {
 				Expect(err).NotTo(HaveOccurred())
 			}
 
+			verifyNodePoolListByCluster := func(ctx context.Context, rgName, clusterName string, expectedNodePoolNames []string) {
```
mbukatov, line 73: Would it make sense to turn this function into a verifier?

aklymenk: I didn't think we could potentially reuse it in some other E2E tests. Do you think we may want to be able to do it?

test/e2e/nodepool_update_nodes.go
```diff
@@ -69,10 +70,66 @@ var _ = Describe("Customer", func() {
...
+						if np.Name != nil {
+							listedNodePoolNames = append(listedNodePoolNames, *np.Name)
+						}
+					}
+				}
+
+				Expect(listedNodePoolNames).To(ConsistOf(expectedNodePoolNames), "expected %v node pools listed by cluster %s, got %v", expectedNodePoolNames, clusterName, listedNodePoolNames)
+			}
+
+			verifyNodePoolListByRG := func(ctx context.Context, rgName string, expectedNodePoolNames []string) {
```
mgahagan73, line 93: Please add this as a verifier as well.

test/e2e/nodepool_update_nodes.go
mgahagan73: I think it makes sense to add this one to verifiers as well as the RG based one below.

raelga (commented):
@aklymenk In the past few weeks the test had several iterations. DOues this still apply? Can you review the conflicts?

aklymenk:
> @aklymenk In the past few weeks the test had several iterations. DOues this still apply? Can you review the conflicts?

Yes, this PR is still relevant. But firstly I need another PR to merge before this one, and I need to make some changes to this PR.

aklymenk:
Explanation why this PR is in draft now: [https://redhat.atlassian.net/browse/ARO-21295?focusedCommentId=16919022](https://redhat.atlassian.net/browse/ARO-21295?focusedCommentId=16919022)

davidffrench:
Closing — the listing test approach was reworked by mvacula02 in #4816 (merged 2026-04-09), which moved listing coverage into the golden path test structure. Thank you Andrii!

raelga (commented):
/close

PR #4873, opened 2026-04-14
e2e resource group cleanup: use rg-ordered workflow and make "no-rp" mode best-effort

roivaz:
Replace the ad-hoc `DeleteResourceGroup` calls in e2e cleanup with the `cleanup-sweeper`'s `rg-ordered` workflow, and let callers control whether deletion errors are fatal.

**Changes:**

- **Both `standard` and `no-rp` cleanup modes** now use `runOrderedResourceGroupCleanup`, which runs `ResourceGroupOrderedCleanupWorkflow` from the `cleanup-sweeper` engine. This removes direct use of `DeleteResourceGroup`, `resourceClientFactory`, and `networkClientFactory` from the cleanup paths.
- **`ContinueOnError` added to `WorkflowOptions`**: Steps in the ordered workflow now use `opts.ContinueOnError` instead of hardcoding `true`, letting callers choose between strict and best-effort behavior.
- **`standard` mode** (`ContinueOnError: false`, `Wait: true`): Deletion failures abort the workflow and fail the job — correct for int/stg/prod where cleanup issues need immediate visibility.
- **`no-rp` mode** (`ContinueOnError: true`, `Wait: false`): Per-target deletion failures are logged and skipped inside the engine. Discovery errors still propagate since they indicate broken code. Deletions are fire-and-forget.
- Existing `cleanup-sweeper` and `templatize` callers explicitly pass `ContinueOnError: true` to preserve their best-effort behavior.
- Added `cleanup-sweeper` as a dependency of the `test` module.
- **Grant release bot Graph directory read permission** for orphaned role assignment discovery, and treat per-target delete failures as best-effort so one failed deletion doesn't fail the whole sweep.


raelga (approved):
/lgtm

PR #4875, opened 2026-04-14
Calculate latest OCP version for e2e tests

miquelsi:
### What

Calculate latest OCP version for e2e tests


miquelsi:
Addressed in #4774

PR #4884, opened 2026-04-15
Move install-tools to base-ci image and update e2e tests image

patriksuba:
[ARO-23919](https://redhat.atlassian.net/browse/ARO-23919)

### What

The PR moves install-tools from aro-hcp-e2e-tools image to base-ci image. This change affects e2e tests (aro-hcp-e2e-tests) image.

### Why

To not install 'install-tools' on every CI execution that require this tools.

Related changes in PROW CI https://github.com/openshift/release/pull/75802

### DIFF
--- a/test/Containerfile.e2e
+++ b/test/Containerfile.e2e
@@ -3,7 +3,8 @@ USER root
 WORKDIR /opt/app-root/src/github.com/Azure/ARO-HCP
 COPY . .
 ENV GOFLAGS='-mod=readonly'
-RUN make -C tooling/templatize templatize && \
+RUN make build-hcpctl && \
+    make -C tooling/templatize templatize && \
     make -C test/ build && \
     go clean -cache -modcache && rm -f go.work.sum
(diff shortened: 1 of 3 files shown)


test/Containerfile.e2e
```diff
@@ -3,6 +3,8 @@ USER root
 WORKDIR /opt/app-root/src/github.com/Azure/ARO-HCP
 COPY . .
```
janboll, line 4: perhaps add  a .dockerignore to avoid sending i.e. .git or other unrelated files as build context

patriksuba: Good point. We need the .git folder to get the commit and version for the hcpctl build.

Besides size, is there another concern? The repo is only 100 MB. Unless we need to exclude something for security, I'd prefer to keep it as-is to ensure we don't break future dependencies.

mbukatov (commented):
/lgtm

stevekuznetsov (approved):
/lgtm
/approve

PR #4889, opened 2026-04-15
Turning off local auth for audit logs eventhub

rachelvweber:
[ARO-26129](https://redhat.atlassian.net/browse/ARO-26129)

### What

Turns off local auth for audit logs eventhub

### Why

Required by Azure policy. The kusto data connection uses aad auth, so it will be unaffected

### Testing

I ran the e2e test with this flag flipped, and with the eventhub manageInstance flag set to true. It updated the eventhub to not use local auth, and the logs continued to flow to kusto in dev


janboll (commented):
/lgtm

PR #4891, opened 2026-04-15
backend: Add controllers for dispatching credential requests

mbarnes:
[ARO-24384 - Move Cluster Service CRUD calls to ARO-HCP backend](https://redhat.atlassian.net/browse/ARO-24384)

### What

This is **stage two** of a three-stage deployment plan described in #4901 to move the Cluster Service calls for requesting and revoking break-glass credentials to the RP backend.

This pull request introduces two new backend controllers – `DispatchRequestCredential` and `DispatchRevokeCredentials` – but they are dormant for now because the RP frontend is still making the Cluster Service calls.

Once this pull request reaches production, a subsequent pull request will remove these two Cluster Service calls from the RP frontend, at which point the new backend controllers will take over.

### Why

This is part of a plan to move _all_ Cluster Service calls to the RP backend, but I'm starting with break-glass credentials because I believe it will have the largest impact in meeting the [Control Plane Latency KPI](https://redhat.atlassian.net/browse/ARO-21536).

### Testing

Relevant unit tests have been introduced or updated and I have performed manual testing in a personal dev environment (with frontend CS calls disabled) to verify the new backend controllers.

This change should be fully transparent to E2E tests, so no changes are needed there.


backend/pkg/controllers/operationcontrollers/dispatch_revoke_credentials.go
```diff
@@ -0,0 +1,147 @@
+// Copyright 2025 Microsoft Corporation
```
machi1990, line 1: 2026

backend/pkg/controllers/operationcontrollers/dispatch_request_credential.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2025 Microsoft Corporation
```
machi1990, line 1: 2026

machi1990:
> .. but they are inert for now because the RP frontend is still making the Cluster Service calls.

I am trying to think of a reason we'd prefer to do a 2 phased rollout: Is that if we do a one phased, during the rollout to a region, it could be that the frontend is rolled out but not the backend in which case we'll have issues?

mbarnes:
> I am trying to think of a reason we'd prefer to do a 2 phased rollout: Is that if we do a one phased, during the rollout to a region, it could be that the frontend is rolled out but not the backend in which case we'll have issues?

Correct.  If the backend had to be held back for some reason then we could be in a situation where neither component is making the CS call.

backend/pkg/controllers/operationcontrollers/dispatch_revoke_credentials.go
mbarnes: Thanks, fixed.

backend/pkg/controllers/operationcontrollers/dispatch_request_credential.go
mbarnes: Thanks, fixed.

mbarnes:
Taking this out of draft since I reworked some of the backend logic and testing is faring better now.

Working on backend integration tests, but they don't necessarily need to be in this PR since the dispatch controllers are dormant for now.

backend/pkg/controllers/operationcontrollers/dispatch_request_credential.go
```diff
@@ -0,0 +1,119 @@
...
+	// The frontend cancels all active RequestCredential operations when
+	// handling a revocation request, but it cannot do so atomically. So
+	// there is a slim chance of a straggler slipping through. This is a
+	// second line of defense.
+
+	if len(cluster.ServiceProviderProperties.RevokeCredentialsOperationID) > 0 {
+		logger.Info("revocation in progress, canceling operation",
+			"revoke_credentials_operation_id", cluster.ServiceProviderProperties.RevokeCredentialsOperationID)
+
+		operation.Cancel(localClock.Now())
```
deads2k: clock should be an instance variable on the controller struct.

backend/pkg/controllers/operationcontrollers/dispatch_request_credential.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Dispatch the credential request to Clusters Service.
+
+	logger.Info("dispatching operation")
+	csBreakGlassCredential, err := opsync.clustersServiceClient.PostBreakGlassCredential(ctx, cluster.ServiceProviderProperties.ClusterServiceID)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	csBreakGlassCredentialID, err := api.NewInternalID(csBreakGlassCredential.HREF())
```
deads2k, line 129: if we crash here, we can get two `PostBreakGlassCredential` calls for the same operation.  I don't think anything bad happens (just a cred we lose track of and can still revoke), but please confirm and note it in the code as a comment.

backend/pkg/controllers/operationcontrollers/dispatch_revoke_credentials.go
```diff
@@ -0,0 +1,137 @@
...
+
+	cluster, err := opsync.cosmosClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).Get(ctx, operation.ExternalID.Name)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	if cluster.ServiceProviderProperties.RevokeCredentialsOperationID != operation.OperationID.Name {
+		logger.Info("cluster RevokeCredentialsOperationID mismatch",
+			"revoke_credentials_operation_id", cluster.ServiceProviderProperties.RevokeCredentialsOperationID)
+
+		operation.Cancel(localClock.Now())
```
deads2k: same clock comment

backend/pkg/controllers/operationcontrollers/dispatch_revoke_credentials.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			return utils.TrackError(err)
+		}
+
+		return nil
+	}
+
+	// Dispatch the revocation request to Clusters Service.
+
+	logger.Info("dispatching operation")
```
deads2k: here and above give us a better string.

backend/pkg/controllers/operationcontrollers/dispatch_request_credential.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// Cancel the operation if a revocation is in progress.
+	//
+	// The frontend cancels all active RequestCredential operations when
+	// handling a revocation request, but it cannot do so atomically. So
+	// there is a slim chance of a straggler slipping through. This is a
+	// second line of defense.
+
+	if len(cluster.ServiceProviderProperties.RevokeCredentialsOperationID) > 0 {
```
deads2k, line 107: I'm feeling slow.  What sets this value empty string once the revocation is complete so that a flow like revoke, then create works?

internal/api/types_operation.go
```diff
@@ -76,6 +76,17 @@ type Operation struct {
 	Error *arm.CloudErrorBody `json:"error,omitempty"`
 }
 
+// Cancel updates relevant fields to indicate the operation is canceled.
+func (o *Operation) Cancel(now time.Time) *Operation {
```
deads2k: no functions in the API types. Place in apihelpers.

deads2k:
Create a second PR built on this PR that uses these backend controllers to demonstrate they work. Also, unit tests for the controllers are required even if they dormant for now.

backend/pkg/controllers/operationcontrollers/dispatch_request_credential.go
mbarnes: That was introduced recently in the [OperationRevokeCredentials controller](https://github.com/Azure/ARO-HCP/blob/0ac752486951dcdee3af38c2bd68db9a2608cc3a/backend/pkg/controllers/operationcontrollers/operation_revoke_credentials.go#L143-L148).

#4796 for more context.

backend/pkg/controllers/operationcontrollers/dispatch_request_credential.go
mbarnes: Correct. On the failed attempt, CS allocates a credential that never makes it back to the client.  The lost credential eventually expires or gets revoked.  Will add a comment.

backend/pkg/controllers/operationcontrollers/dispatch_request_credential.go
mbarnes: Fixed in a new commit.

backend/pkg/controllers/operationcontrollers/dispatch_request_credential.go
mbarnes: Comment added in a new commit.

backend/pkg/controllers/operationcontrollers/dispatch_revoke_credentials.go
mbarnes: Fixed in a new commit.

backend/pkg/controllers/operationcontrollers/dispatch_revoke_credentials.go
mbarnes: Hopefully gave a satisfactory string in a new commit.

internal/api/types_operation.go
mbarnes: Reworked the 1st commit to put it in apihelpers.

deads2k:
proof in #5007 works. Nice job

/lgtm
/approve

PR #4895, opened 2026-04-16
Fixes for aro-hcp exporter

janboll:
https://redhat.atlassian.net/browse/AROSLSRE-481

### What/Why

- Add missing parameters conditionally if kusto is enabled
- Add limit to queries to not go out of bounds
- Fix log message, remove stale 30 minutes message
- Add kusto_logs_age_in_seconds to expected metrics list


raelga (approved):
/lgtm

PR #4898, opened 2026-04-16
chore: remove azure login check gh action

JameelB:
### What

Remove the azure login check Github action which checks that the source repository used is always Azure/ARO-HCP and not a fork of the repo. 

### Why
Removing this check since:
- Tide doesn't care whether the branch is coming from a fork or not.
- After Tide was introduced, the source repo requirement was no longer enforced. There have been instances where PRs have been merged manually with this check failing

If this is not enforced any longer, we should remove this check to avoid failing actions when it no longer matters. 

### Testing
N/A - Github action configuration change only

N/A


raelga:
Rebased

PR #4901, opened 2026-04-16
frontend: Initialize credential revocation operations to "Deleting"

mbarnes:
[ARO-24384 - Move Cluster Service CRUD calls to ARO-HCP backend](https://redhat.atlassian.net/browse/ARO-24384)

### What

Moving the Clusters Service DELETE call for break-glass credentials to the backend is proving to be tricky.

Because there's no single Clusters Service resource to poll for status, the backend controllers for dispatching the CS call and polling CS for revocation completion can't reliably coordinate.

Imagine a credential revocation when there's no present break-glass credentials but there _are_ pending requests in Cosmos DB that haven't yet been dispatched to Clusters Service. The controller for polling status on this operation will see there's no credentials pending revocation and immediately mark the revocation operation as successful. If that happens before the controller for dispatching the revocation to Clusters Service has a chance to run, then the Clusters Service call will never be made and the pending credential requests will complete despite the client requesting revocation.

To resolve this, for the `RevokeCredentials` operation type only I'm imbuing new meaning into some of the operation's non-terminal status values:

  - **Accepted**
    - The client's request for credential revocation is underway. An operation document exists in Cosmos DB but the Clusters Service call has not yet been made.
  - **Deleting**
    - The Clusters Service call has been made and status polling should commence.

Eventually, the dispatch controller for credential revocations will update the operation status from **Accepted** to **Deleting** after calling Clusters Service. Meanwhile, the controller for polling status will disregard credential revocation operations with an **Accepted** status.

First, however, the frontend must initialize credential revocation operations to status **Deleting** instead of **Accepted**, since it's still making the Clusters Service call immediately for now.

### Testing

The relevant frontend integration test has been updated to confirm `RevokeCredentials` operations now begin with a **Deleting** status.

In order to keep the frontend and backend images compatible with a +/-1 version skew, moving break-glass credential calls to the backend is going to have to be a three-stage deployment.

#### Stage 1

This pull requests changes the initial state of credential revocation operations to **Deleting**, since the frontend is making the CS call.

#### Stage 2

The status polling controller for `RevokeCredentials` will begin disregarding operations with status **Accepted**.

At the same time, introduce a dispatch controller for `RevokeCredentials` that makes the CS call when the operation has an **Accepted** status, and then updates the status to **Deleting**.

Note, at this stage the frontend is still making the CS call itself and initializing the operation to **Deleting**, so the new dispatch controller is dormant.

#### Stage 3

Remove the CS call from the frontend and change the initial state of credential revocation operations back to **Accepted**.

### DIFF
--- a/frontend/pkg/frontend/frontend.go
+++ b/frontend/pkg/frontend/frontend.go
@@ -517,6 +517,19 @@ func (f *Frontend) ArmResourceActionRevokeCredentials(writer http.ResponseWriter
 		request.Header.Get(arm.HeaderNameClientObjectID),
 		request.Header.Get(arm.HeaderNameAsyncNotificationURI),
 		correlationData)
+
+	// XXX TEMPORARY: This must be removed along with the Clusters Service call.
+	//
+	//     For this operation type, because there is no single Clusters Service
+	//     resource to poll for completion, the operation's status field is used
+	//     for backend controller coordination. "Accepted" means the revocation
+	//     has not yet been dispatched to Clusters Service. Once dispatched, the
+	//     operation status becomes "Deleting" and is ready for status polling.
+	//
+	//     Because the credentials revocation has already been dispatched here,
+	//     set the initial operation status to "Deleting".
+	operationDoc.Status = arm.ProvisioningStateDeleting
+
 	transaction.OnSuccess(addOperationResponseHeaders(writer, request, operationDoc.NotificationURI, operationDoc.OperationID))
 	_, err = f.dbClient.Operations(operationDoc.OperationID.SubscriptionID).AddCreateToTransaction(ctx, transaction, operationDoc, nil)
 	if err != nil {
--- a/test-integration/frontend/artifacts/FrontendCRUD/AdminCredentials/revoke/03-listActiveOperations-revoke-credentials/operation-revoke-credentials.json
+++ b/test-integration/frontend/artifacts/FrontendCRUD/AdminCredentials/revoke/03-listActiveOperations-revoke-credentials/operation-revoke-credentials.json
@@ -8,6 +8,6 @@
 	"request": "RevokeCredentials",
 	"resourceId": "/subscriptions/0465bc32-c654-41b8-8d87-9815d7abe8f6/providers/Microsoft.RedHatOpenShift/hcpOperationStatuses/00000000-0000-0000-0000-000000000000",
 	"startTime": "2025-01-01T00:00:00Z",
-	"status": "Accepted",
+	"status": "Deleting",
 	"tenantId": "00000000-0000-0000-0000-000000000000"
 }


geoberle:
the plan is sound
/approve
/lgtm

machi1990:
@mbarnes 
If the operation controller was modified to handle it like:
- if revocation complete for all the credentials return "ProvisioningStateSucceeded"
- if all expired, consider the revocation done and return "ProvisioningStateSucceeded"
- if any revocation ongoing (awaiting revocation or active) return "ProvisioningStateDeleting"

This leaves a question of: How/When does the revocation controller perform the revocation which I think can be achieved via:
- always revoke if revocation needs to be done; e.g there is a revocation id & no ongoing revocation (if there is CS returns 400)

Will that help bring the number of steps down to 2 from 3?

mbarnes:
That works when there's break-glass credentials in CS to monitor.

What I'm trying to address is a corner case where there are no break-glass credentials in CS but there _are_ pending credential requests in Cosmos DB that haven't been dispatched to CS yet.

Keep in mind that moving CS calls to the backend introduces significant delay between the frontend creating a new operation in Cosmos DB and the backend noticing it and making the requisite CS call(s).  A lot can happen in that interval.  (Integrating the Cosmos DB change feed will mitigate the delay somewhat.)

In this scenario, if the dispatch controller runs first then all is fine.  The controller can be smart enough to skip the CS call but still cancel the pending requests only in Cosmos DB.  It could even mark the revocation operation as done.

If the operation controller runs first, it finds no break-glass credentials in CS and immediately marks the operation as done.  The dispatch controller never processes the operation since it's already done, and therefore never cancels the pending requests in Cosmos DB.  So they get dispatched to CS and produce a valid admin kubeconfig despite the client's revocation request.

I don't know, there may well be a way to achieve this in only two stages but it hurts my brain.  There's a lot of potential races here.  Leveraging the operation's status value for coordination is just easier to reason about.

machi1990:
Thanks for the explanation.

> Keep in mind that moving CS calls to the backend introduces significant delay between the frontend creating a new operation in Cosmos DB and the backend noticing it and making the requisite CS call(s). A lot can happen in that interval. (Integrating the Cosmos DB change feed will mitigate the delay somewhat.)

> If the operation controller runs first, it finds no break-glass credentials in CS and immediately marks the operation as done. The dispatch controller never processes the operation since it's already done, and therefore never cancels the pending requests in Cosmos DB. So they get dispatched to CS and produce a valid admin kubeconfig despite the client's revocation request.

In my mind(speaking from 1000 feet view) this is the last case where the revocation https://github.com/Azure/ARO-HCP/pull/4901#issuecomment-4260802463 (copied below) i.e the the bgc is `"active"` (whatever that means )

> if any revocation ongoing (awaiting revocation or active) return "ProvisioningStateDeleting"

> I don't know, there may well be a way to achieve this in only two stages but it hurts my brain. There's a lot of potential races here. Leveraging the operation's status value for coordination is just easier to reason about.

I've not thought it deep as you are and I see your point, reasoning about it could be a bit harder or more involving than doing a one off phased rolled out.

PR #4903, opened 2026-04-16
stage mise v2 config in prep for cutover

tony-schndr:
[ARO-26217](https://redhat.atlassian.net/browse/ARO-26217)

### What

Updates config.schema.json to add misev2 container in preparation for https://github.com/Azure/ARO-HCP/pull/4886.  Rollout will be in 3 stages.

1. Update schema (this PR)
2. Add image details to sdp-pipelines
3. Merge #4886

### Why

To not break sdp-pipelines and deployments since the misev2 image will not be set in sdp-pipelines and we don't manage that image in this repository.

### DIFF
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -2053,6 +2053,9 @@
         "image": {
           "$ref": "#/definitions/containerImage"
         },
+        "imageV2": {
+          "$ref": "#/definitions/containerImage"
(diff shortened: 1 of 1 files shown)


mbarnes:
/lgtm
... for what little it's worth since I'm not a `/config` owner.

PR #4908, opened 2026-04-16
gather-observability: add chart about frontend request latency

geoberle:
### What

"control plane request latency" becomes important to us, so lets make it visible in e2e tests


bennerv:
/lgtm
/approve

The latency will look slightly different since it's measured from a different perspective, but should be representative.

PR #4912, opened 2026-04-16
ARO-26227: backend metrics controller followups

swiencki:
[ARO-26227](https://redhat.atlassian.net/browse/ARO-26227)

## What

This PR is intentionally narrowed to a small backend correctness fix:

- wire `MetricsRegisterer` and `MetricsGatherer` explicitly from `backend/cmd/root.go` into `BackendOptions`
- make `NewBackend` return an error and fail fast when the metrics registerer/gatherer pairing is not wired correctly
- serve `/healthz` and `/metrics` from separate `http.ServeMux` / `http.Server` instances
- create the metrics listener inside `Run` so no socket is bound when `Run` is never called
- add tests for nil options and the required metrics registry pairing

## Why

This keeps `#4912` focused on the narrow backend correctness work that surfaced in review:

- avoid silently gathering `/metrics` from one registry while controllers write to another
- avoid sharing the global `DefaultServeMux` between `/healthz` and `/metrics`
- make backend construction fail fast before any goroutine, tracer, or listener is allocated when metrics wiring is invalid

## Not in scope

- no operation metrics label-schema changes
- no `operation_id`
- no `resource_id` / `subscription_id` redesign
- no metrics controller or informer refactors

## Validation

- `go test -run '^$' ./backend/cmd/...`
- `go test ./backend/pkg/app ./test-integration/backend/launch`

### DIFF
--- a/backend/pkg/app/backend.go
+++ b/backend/pkg/app/backend.go
@@ -32,7 +32,6 @@ import (
 	k8sutilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
-	"k8s.io/component-base/metrics/legacyregistry"
 	utilsclock "k8s.io/utils/clock"
 
 	azureclient "github.com/Azure/ARO-HCP/backend/pkg/azure/client"
@@ -90,7 +89,10 @@ func (o *BackendOptions) RunBackend(ctx context.Context) error {
 	ctx, cancel := context.WithCancelCause(ctx)
 	defer cancel(fmt.Errorf("function returned"))
 
-	backend := NewBackend(o)
+	backend, err := NewBackend(o)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to construct backend: %w", err))
+	}
 	runErrCh := make(chan error, 1)
 	go func() {
 		runErrCh <- backend.Run(ctx)
@@ -109,24 +111,32 @@ func (o *BackendOptions) RunBackend(ctx context.Context) error {
 	return nil
 }
 
-func (o *BackendOptions) metricsRegisterer() prometheus.Registerer {
-	if o.MetricsRegisterer != nil {
-		return o.MetricsRegisterer
+func NewBackend(options *BackendOptions) (*Backend, error) {
+	if options == nil {
+		return nil, errors.New("BackendOptions must not be nil")
 	}
-	return legacyregistry.Registerer()
-}
-
-func (o *BackendOptions) metricsGatherer() prometheus.Gatherer {
-	if o.MetricsGatherer != nil {
-		return o.MetricsGatherer
+	if err := options.validate(); err != nil {
+		return nil, err
 	}
-	return legacyregistry.DefaultGatherer
-}
-
-func NewBackend(options *BackendOptions) *Backend {
 	return &Backend{
 		options: options,
+	}, nil
+}
+
+// validate checks BackendOptions for invariants that must hold before Run.
+// Any failure here is a programmer error in the calling code (flag wiring or
+// test setup), not a user-facing condition — we fail fast before any goroutine,
+// tracer, or leader-election resource is allocated.
+func (o *BackendOptions) validate() error {
+	// Registerer and Gatherer must both be explicitly wired by the caller.
+	// The production path sets them in cmd/root.go; tests must inject their
+	// own. A single half-configured field would silently expose metrics from
+	// one registry while populating another, so we refuse to start.
+	if o.MetricsRegisterer == nil || o.MetricsGatherer == nil {
+		return fmt.Errorf("BackendOptions.MetricsRegisterer and BackendOptions.MetricsGatherer must both be set (MetricsRegisterer set=%t, MetricsGatherer set=%t)",
+			o.MetricsRegisterer != nil, o.MetricsGatherer != nil)
 	}
+	return nil
 }
 
 func (b *Backend) Run(ctx context.Context) error {
@@ -176,57 +186,12 @@ func (b *Backend) Run(ctx context.Context) error {
 	errCh := make(chan error, 3)
 	wg := sync.WaitGroup{}
 
-	// Handle requests directly for /healthz endpoint
-	if b.options.HealthzServerListenAddress != "" {
-		backendHealthGauge := promauto.With(b.options.metricsRegisterer()).NewGauge(prometheus.GaugeOpts{Name: "backend_health", Help: "backend_health is 1 when healthy"})
+	healthzServer = b.serveHealthz(ctx, electionChecker, &wg, errCh)
 
-		http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
-
-			if err := electionChecker.Check(r); err != nil {
-				logger.Error(err, "Readiness probe failed")
-				http.Error(w, "lease not renewed", http.StatusServiceUnavailable)
-				backendHealthGauge.Set(0.0)
-				return
-			}
-			w.WriteHeader(http.StatusOK)
-			backendHealthGauge.Set(1.0)
-		})
-
-		healthzServer = &http.Server{Addr: b.options.HealthzServerListenAddress}
-
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			logger.Info(fmt.Sprintf("Healthz server listening on %s", b.options.HealthzServerListenAddress))
-			errCh <- healthzServer.ListenAndServe()
-		}()
-	}
-
-	if b.options.MetricsServerListenAddress != "" || b.options.MetricsServerListener != nil {
-		http.Handle("/metrics", promhttp.InstrumentMetricHandler(
-			b.options.metricsRegisterer(),
-			promhttp.HandlerFor(
-				prometheus.Gatherers{b.options.metricsGatherer()},
-				promhttp.HandlerOpts{},
-			),
-		))
-
-		metricsAddr := b.options.MetricsServerListenAddress
-		if b.options.MetricsServerListener != nil {
-			metricsAddr = b.options.MetricsServerListener.Addr().String()
-		}
-		metricsServer = &http.Server{Addr: metricsAddr}
-
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			logger.Info(fmt.Sprintf("metrics server listening on %s", metricsAddr))
-			if b.options.MetricsServerListener != nil {
-				errCh <- metricsServer.Serve(b.options.MetricsServerListener)
-				return
-			}
-			errCh <- metricsServer.ListenAndServe()
-		}()
+	var err error
+	metricsServer, err = b.serveMetrics(ctx, &wg, errCh)
+	if err != nil {
+		return fmt.Errorf("failed to start metrics server: %w", err)
 	}
 
 	wg.Add(1)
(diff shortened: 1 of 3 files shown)


backend/pkg/informers/informers_test.go
```diff
@@ -835,3 +836,68 @@ func controllerInformerTestCase() informerTestCase {
...
+
+	require.Eventually(t, func() bool {
+		return len(informer.GetIndexer().ListKeys()) == seedCount
+	}, 5*time.Second, 50*time.Millisecond, "cache size should equal seeded operations count")
+
+	// After several relist cycles, the cache size must still equal the
+	// number of operations actually present in the DB — repeated relists
+	// must not cause the cache to grow unbounded.
+	time.Sleep(1 * time.Second)
+	require.Equal(t, seedCount, len(informer.GetIndexer().ListKeys()), "cache size must not grow across relists")
```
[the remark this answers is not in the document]
swiencki: Addressed in 17ed38068: replaced the fixed time.Sleep + require.Equal with require.Never over a 1s/50ms window so a slow-CI relist/replace race can't false-positive the invariant.

backend/pkg/controllers/metricscontrollers/operation_phase_metrics_controller.go
```diff
@@ -17,82 +17,180 @@ package metricscontrollers
...
+		Help: "Unix timestamp when the operation last changed phase.",
+	}, operationMetricLabelNames))
+	if err != nil {
+		return nil, err
+	}
+	return &operationPhaseMetricsHandler{
+		phaseInfo:          phaseInfo,
+		startTime:          startTime,
+		lastTransitionTime: lastTransitionTime,
+		lastLabels:         map[string]operationKey{},
```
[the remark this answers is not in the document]
swiencki: Addressed in 17ed38068: registerOrAdopt now returns (C, adopted bool, error). When NewOperationPhaseMetricsHandler adopts existing GaugeVecs, it .Reset()s all three so the new handler's (empty) lastLabels map stays consistent with collector contents. Covered by TestNewOperationPhaseMetricsHandler_AdoptedCollectorsAreReset.

backend/pkg/app/backend.go
```diff
@@ -123,10 +126,29 @@ func (o *BackendOptions) metricsGatherer() prometheus.Gatherer {
...
+// validate checks BackendOptions for invariants that must hold before Run.
+// Any failure here is a programmer error in the calling code (flag wiring or
+// test setup), not a user-facing condition — we fail fast before any goroutine,
+// tracer, or leader-election resource is allocated.
+func (o *BackendOptions) validate() error {
+	// Registerer and Gatherer must either both be injected (e.g. by tests)
+	// or both be left unset (production path, which falls back to the
+	// k8s legacy registry). A single half-configured field silently causes
+	// metrics to be exposed from one registry and populated to another.
+	if (o.MetricsRegisterer == nil) != (o.MetricsGatherer == nil) {
```
geoberle: validation is good but i prefer that both are always set and we fail here if they are not. no fallback.
root.go needs to fill in the registerer and gatherer when creating the BackendOptions

swiencki: Done in 7fe0feb02: dropped the legacyregistry fallback entirely. `BackendOptions.validate()` now fails fast when either `MetricsRegisterer` or `MetricsGatherer` is nil, and `cmd/root.go` wires `legacyregistry.Registerer()` + `legacyregistry.DefaultGatherer` explicitly when building `BackendOptions`. Test table updated: both-unset is now a rejected case.

backend/pkg/controllers/metricscontrollers/operation_phase_metrics_controller.go
```diff
@@ -17,82 +17,192 @@ package metricscontrollers
...
 		return
 	}
-	subscriptionID := subscriptionIDMetricLabel(op.GetResourceID())
+
+	storeKey, err := resourceIDStoreKey(op)
+	if err != nil || storeKey == "" {
+		// Without a store key we cannot track which series we wrote, so
+		// we must not emit new series (we'd leak them forever on update).
+		return
+	}
```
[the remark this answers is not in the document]
swiencki: Addressed in 790d4279e: Sync now computes storeKey first and calls deleteStored(storeKey) before returning when ExternalID is empty, so an op that loses its omitempty ExternalID on a later patch can't leak previously-emitted series. Covered by TestOperationPhaseMetricsHandler_ExternalIDRemovedClearsStaleSeries.

backend/pkg/controllers/metricscontrollers/metrics_controller.go
```diff
@@ -144,7 +144,7 @@ func (c *Controller[T]) syncResource(ctx context.Context, key string) error {
 	typed, ok := obj.(T)
 	if !ok {
 		logger := utils.LoggerFromContext(ctx)
-		logger.Info("unexpected object type in indexer", "key", key, "controller", c.name, "type", fmt.Sprintf("%T", obj))
+		logger.Error(nil, "unexpected object type in indexer", "key", key, "controller", c.name, "type", fmt.Sprintf("%T", obj))
```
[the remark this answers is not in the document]
swiencki: Addressed in 790d4279e: build a real `fmt.Errorf("unexpected object type in indexer: %T", obj)` and pass it to logger.Error so structured logs no longer carry error=null.

backend/pkg/app/backend.go
```diff
@@ -109,24 +111,29 @@ func (o *BackendOptions) RunBackend(ctx context.Context) error {
...
-	if o.MetricsRegisterer != nil {
-		return o.MetricsRegisterer
-	}
-	return legacyregistry.Registerer()
-}
-
-func (o *BackendOptions) metricsGatherer() prometheus.Gatherer {
-	if o.MetricsGatherer != nil {
-		return o.MetricsGatherer
+func NewBackend(options *BackendOptions) (*Backend, error) {
```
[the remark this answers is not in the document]
swiencki: Addressed in 790d4279e: NewBackend now nil-guards options and returns a clear error before calling validate(). Covered by TestNewBackend_NilOptionsReturnsError.

backend/pkg/controllers/metricscontrollers/register.go
```diff
@@ -0,0 +1,53 @@
...
+// incompatible series.
+func registerOrAdopt[C prometheus.Collector](r prometheus.Registerer, c C) (C, bool, error) {
+	if err := r.Register(c); err != nil {
+		var are prometheus.AlreadyRegisteredError
+		if errors.As(err, &are) {
+			existing, ok := are.ExistingCollector.(C)
+			if !ok {
+				var zero C
+				return zero, false, fmt.Errorf("metric already registered with incompatible type %T: %w", are.ExistingCollector, err)
+			}
```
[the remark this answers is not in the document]
swiencki: Addressed in 790d4279e: docstring updated to say the AlreadyRegisteredError is wrapped with type context (matching the implementation) rather than returned as-is. Wrapping is preferable since it preserves the underlying type via %w while annotating with the incompatible-type context.

backend/pkg/app/backend.go
```diff
@@ -109,24 +111,32 @@ func (o *BackendOptions) RunBackend(ctx context.Context) error {
 	return nil
 }
 
-func (o *BackendOptions) metricsRegisterer() prometheus.Registerer {
-	if o.MetricsRegisterer != nil {
-		return o.MetricsRegisterer
+func NewBackend(options *BackendOptions) (*Backend, error) {
```
deads2k: method receiver to be backendoptions.

backend/pkg/app/backend.go
```diff
@@ -176,57 +186,12 @@ func (b *Backend) Run(ctx context.Context) error {
...
-		healthzServer = &http.Server{Addr: b.options.HealthzServerListenAddress}
-
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			logger.Info(fmt.Sprintf("Healthz server listening on %s", b.options.HealthzServerListenAddress))
-			errCh <- healthzServer.ListenAndServe()
-		}()
-	}
+	healthzServer = b.serveHealthz(ctx, electionChecker, &wg, errCh)
```
deads2k: create a separate type for your separate server.

backend/pkg/app/backend.go
```diff
@@ -176,57 +186,12 @@ func (b *Backend) Run(ctx context.Context) error {
...
-			defer wg.Done()
-			logger.Info(fmt.Sprintf("metrics server listening on %s", metricsAddr))
-			if b.options.MetricsServerListener != nil {
-				errCh <- metricsServer.Serve(b.options.MetricsServerListener)
-				return
-			}
-			errCh <- metricsServer.ListenAndServe()
-		}()
+	var err error
+	metricsServer, err = b.serveMetrics(ctx, &wg, errCh)
```
deads2k: create a separate type for your separate server.

backend/pkg/app/backend.go
```diff
@@ -263,6 +228,96 @@ func (b *Backend) Run(ctx context.Context) error {
 	return errors.Join(errs...)
 }
 
+// serveHealthz starts the /healthz HTTP server on its own ServeMux, in a dedicated
+// goroutine, if the option is configured. It returns the *http.Server so the caller
+// can shut it down. Returns nil when HealthzServerListenAddress is empty.
+func (b *Backend) serveHealthz(ctx context.Context, electionChecker *leaderelection.HealthzAdaptor, wg *sync.WaitGroup, errCh chan<- error) *http.Server {
```
deads2k: passing a wait group smells weird. Same with an error channel

backend/pkg/app/backend.go
```diff
@@ -263,6 +228,96 @@ func (b *Backend) Run(ctx context.Context) error {
...
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+		backendHealthGauge.Set(1.0)
+	})
+
+	server := &http.Server{Addr: b.options.HealthzServerListenAddress, Handler: mux}
+
+	wg.Add(1)
+	go func() {
```
deads2k: yeah, don't this from a `serve` funcition. Can be done from a `Start` function, but realistically, you probably want a `Run` that calls `ListenAndServe` and then handle threading a level up.

backend/pkg/app/backend.go
```diff
@@ -109,24 +111,32 @@ func (o *BackendOptions) RunBackend(ctx context.Context) error {
 	return nil
 }
 
-func (o *BackendOptions) metricsRegisterer() prometheus.Registerer {
-	if o.MetricsRegisterer != nil {
-		return o.MetricsRegisterer
+func NewBackend(options *BackendOptions) (*Backend, error) {
+	if options == nil {
+		return nil, errors.New("BackendOptions must not be nil")
 	}
```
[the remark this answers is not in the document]
swiencki: Addressed in ce8067e37: error string lowercased.

backend/pkg/app/backend.go
```diff
@@ -109,24 +111,32 @@ func (o *BackendOptions) RunBackend(ctx context.Context) error {
...
+// test setup), not a user-facing condition — we fail fast before any goroutine,
+// tracer, or leader-election resource is allocated.
+func (o *BackendOptions) validate() error {
+	// Registerer and Gatherer must both be explicitly wired by the caller.
+	// The production path sets them in cmd/root.go; tests must inject their
+	// own. A single half-configured field would silently expose metrics from
+	// one registry while populating another, so we refuse to start.
+	if o.MetricsRegisterer == nil || o.MetricsGatherer == nil {
+		return fmt.Errorf("BackendOptions.MetricsRegisterer and BackendOptions.MetricsGatherer must both be set (MetricsRegisterer set=%t, MetricsGatherer set=%t)",
+			o.MetricsRegisterer != nil, o.MetricsGatherer != nil)
```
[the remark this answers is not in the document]
swiencki: Addressed in ce8067e37: lowercased to "metrics registerer and gatherer must both be set (registerer set=..., gatherer set=...)".

backend/pkg/app/backend.go
swiencki: Already a receiver on BackendOptions — `(o *BackendOptions) NewBackend()` since 7f5531d80.

backend/pkg/app/backend.go
swiencki: Addressed in ce8067e37: backendHealthzServer and backendMetricsServer are now self-contained types with their own Run() that creates listeners, serves, and shuts down on ctx.Done.

backend/pkg/app/backend.go
swiencki: Same as above — addressed in ce8067e37.

backend/pkg/app/backend.go
swiencki: Addressed in ce8067e37: removed sync.WaitGroup entirely. Goroutines counted upfront, each sends to errCh, collected via ranged receive.

backend/pkg/app/backend.go
swiencki: Addressed in ce8067e37: each server type creates its own listener inside Run(). Factory methods and close() methods removed. Backend.Run just constructs the server struct and launches Run in a goroutine.

backend/pkg/app/backend.go
```diff
@@ -240,20 +237,10 @@ func (b *Backend) Run(ctx context.Context) error {
 
 	<-ctx.Done()
 
-	// always attempt a graceful shutdown, a double ctrl+c exits the process
-	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 31*time.Second)
-	defer shutdownCancel()
-	_ = b.shutdownHTTPServer(shutdownCtx, metricsServer, "metrics server")
-	_ = b.shutdownHTTPServer(shutdownCtx, healthzServer, "healthz server")
-
-	wg.Wait()
```
deads2k, line 249: Wait group wait, was more clear. Please put it back and properly close errCh

deads2k:
It's a really weird refactor.  I guess it solves the Run issue, but is it really the best way?  The coupling is wild.  I don't have time to rewrite myself, I'm pretty sure this works-ish, but something has still gone weird 

/lgtm

PR #4947, opened 2026-04-20
fix for az acr login error in personal-dev-env

sclarkso:
### What

az acr login without --expose-token tries to store credentials via the Docker CLI, which fails when using Podman. The record-override targets already used the token-based approach correctly, but the build-and-push targets were missed during the Podman migration (066f02d41). This switches all build-and-push targets to use --expose-token with $(CONTAINER_ENGINE) login, consistent with the rest of the codebase. Also fixes a hardcoded docker push in tooling/tenant-quota/Makefile.

### DIFF
--- a/tooling/tenant-quota/Makefile
+++ b/tooling/tenant-quota/Makefile
@@ -95,8 +95,9 @@ image: Dockerfile $(shell find . -name '*.go' -o -name 'go.mod' -o -name 'go.sum
 .PHONY: image
 
 build-and-push: image
-	az acr login --name ${ARO_HCP_IMAGE_ACR}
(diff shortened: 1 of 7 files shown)


tooling/tenant-quota/Makefile
mmazur, line 1:
```suggestion
TENANT_QUOTA_COLLECTOR_DIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
CONTAINER_ENGINE ?= $(shell command -v podman 2>/dev/null || echo docker)
```

raelga (changes requested):
@sclarkso Does this still applies?

PR #4949, opened 2026-04-20
feat: add external auth resource id name validation

miguelsorianod:
Enforce what CS currently enforces, which is: minimum 1 characters
and the charset of RFC1035 DNS label. A difference compared to what
CS enforces is that CS mandates only lowercase characters whereas here
we allow uppercase ones. The reason for this is that the resource id
in azure are case insensitive. To comply with CS enforcement when we
pass the resource name to CS on external auth creation we lowercase all
the characters

### DIFF
--- a/internal/validation/validate_externalauth.go
+++ b/internal/validation/validate_externalauth.go
@@ -54,6 +54,10 @@ func validateExternalAuth(ctx context.Context, op operation.Operation, newObj, o
 	//arm.ProxyResource
 	errs = append(errs, validateProxyResource(ctx, op, field.NewPath("trackedResource"), &newObj.ProxyResource, safe.Field(oldObj, toExternalAuthProxyResource))...)
 	errs = append(errs, RestrictedResourceIDWithResourceGroup(ctx, op, field.NewPath("id"), newObj.ID, nil, api.ExternalAuthResourceType.String())...)
+	if newObj.ID != nil {
+		errs = append(errs, MaxLen(ctx, op, field.NewPath("id"), &newObj.ID.Name, nil, 15)...)
+		errs = append(errs, MatchesRegex(ctx, op, field.NewPath("id"), &newObj.ID.Name, nil, externalAuthResourceNameRegex, externalAuthResourceNameErrorString)...)
+	}
 
 	//Properties HCPOpenShiftClusterExternalAuthProperties `json:"properties"`
 	errs = append(errs, validateExternalAuthProperties(ctx, op, field.NewPath("properties"), &newObj.Properties, safe.Field(oldObj, toExternalAuthProperties))...)
--- a/internal/validation/validate_externalauth_comprehensive_test.go
+++ b/internal/validation/validate_externalauth_comprehensive_test.go
@@ -343,6 +343,120 @@ func TestValidateExternalAuth(t *testing.T) {
 			},
 		},
 		{
+			name: "invalid external auth resource name - empty",
+			newObj: func() *api.HCPOpenShiftClusterExternalAuth {
+				obj := createValidExternalAuth()
+				obj.ID.Name = ""
+				obj.Name = ""
+				return obj
+			}(),
+			op: operation.Operation{Type: operation.Create},
+			expectErrors: []utils.ExpectedError{
+				{FieldPath: "trackedResource.resource.id", Message: "resource name is required"},
+				{FieldPath: "id", Message: "resource name is required"},
+			},
+		},
+		{
+			name: "invalid external auth resource name - special character",
+			newObj: func() *api.HCPOpenShiftClusterExternalAuth {
+				obj := createValidExternalAuth()
+				obj.ID.Name = "$"
+				obj.Name = "$"
+				return obj
+			}(),
+			op: operation.Operation{Type: operation.Create},
+			expectErrors: []utils.ExpectedError{
+				{FieldPath: "id", Message: "must be a valid DNS RFC 1035 label"},
+			},
+		},
+		{
+			name: "invalid external auth resource name - starts with hyphen",
+			newObj: func() *api.HCPOpenShiftClusterExternalAuth {
+				obj := createValidExternalAuth()
+				obj.ID.Name = "-abcde"
+				obj.Name = "-abcde"
+				return obj
+			}(),
+			op: operation.Operation{Type: operation.Create},
+			expectErrors: []utils.ExpectedError{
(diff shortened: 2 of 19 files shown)


api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
```diff
@@ -49,7 +49,7 @@ model NodePool is TrackedResource<NodePoolProperties> {
 model ExternalAuth is ProxyResource<ExternalAuthProperties> {
   ...ResourceNameParameter<
     ExternalAuth,
-    NamePattern = "^[a-zA-Z][-a-zA-Z0-9]{1,15}$"
+    NamePattern = "^[a-zA-Z][-a-zA-Z0-9]{1,13}[a-zA-Z0-9]$"
```
miguelsorianod, line 52: I also updated the typespec and openapi to adjust to what we enforce now. I am unsure about the implications of this and the rollout.

If we leave it unchanged it won't reflect reality. Even before this PR, providing something that complied with the old regex wouldn't end up in a successful creation of an external auth on cs side: when the request to create the external auth in cs arrived the validation failed on that side.

miguelsorianod, line 52: For completeness purposes, this is the current regex in CS:
```
regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])?$`)
```

Notice how a difference is that CS only allows lowercase and in our validation we allow uppercase characters too. We keep the RP side as lowercase and uppercase because it is mentioned resource ids are case insensitive. We solve this discrepancy by lowercasing before sending to CS.

miguelsorianod, line 52: Notice how adding this validation also changes the minimum character length from 1 to 3. The same reasoning as https://github.com/Azure/ARO-HCP/pull/4949/changes#r3110660555 applies here.

bennerv: Is this a CS restriction or an external auth / hypershift restriction?  If it's the former, then we shouldn't update our API and relax it once we move away from CS.

We can't update our existing API versions (technically we could in private preview) but these won't work anyway so 🤷

bennerv: Technically we're supposed to honor whatever casing the user issued the request in.  I think we're doing that today, @mbarnes can say if we're not.

miguelsorianod: > Is this a CS restriction or an external auth / hypershift restriction? If it's the former, then we shouldn't update our API and relax it once we move away from CS.

CS expects the following regex:
```
regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])?$`)
```

The difference from the RP validation is:
1. It accepts lowercase only
 * If I recall correctly this is because the name is used by CS as part of K8s resource names and many K8s resource names require all lowercase characters https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names. To comply with CS enforcement when we
pass the resource name from RP to CS CS on external auth creation we lowercase all the characters
3. It requires a minimum of 2 characters
  * I can't think of any reason on particular other than whoever wrote the regex tried to define the regex but didn't realize that that made the first and last character mandatory and/or wanted to keep the regex simple
  * This makes me think we might want to change the regex in this PR to {0,13} instead of {1,13}

miguelsorianod: > It requires a minimum of 2 characters

Actually now I noticed that CS has a parentheses so the minimum is effectively 1 character. I wonder if we should also add that for externalauth **and** for cluster and nodepool in the RP relax it too because I see they are more restrictive than CS

internal/validation/validate_externalauth.go
```diff
@@ -54,6 +54,10 @@ func validateExternalAuth(ctx context.Context, op operation.Operation, newObj, o
 	//arm.ProxyResource
 	errs = append(errs, validateProxyResource(ctx, op, field.NewPath("trackedResource"), &newObj.ProxyResource, safe.Field(oldObj, toExternalAuthProxyResource))...)
 	errs = append(errs, RestrictedResourceIDWithResourceGroup(ctx, op, field.NewPath("id"), newObj.ID, nil, api.ExternalAuthResourceType.String())...)
+	if newObj.ID != nil {
+		errs = append(errs, MaxLen(ctx, op, field.NewPath("id"), &newObj.ID.Name, nil, 15)...)
```
swiencki, line 58: This tightens the ExternalAuth naming contract, but the shared “minimum valid” fixture was not updated to match. `internal/api/testhelpers.go` still defines `TestExternalAuthName = "testExternalAuth"`, and `internal/validation/hcpopenshiftclusterexternalauth_test.go` still uses that helper as the valid baseline. On the PR head, `go test ./internal/validation` now fails with both max-length and RFC1035 validation errors for that name. Please carry the new constraint through the shared helper/fixtures and dependent tests in the same change so the validation corpus stays aligned with the updated contract.

miguelsorianod: updated

internal/validation/validators.go
```diff
@@ -327,6 +327,10 @@ var (
 	nodePoolResourceNameRegex       = regexp.MustCompile(nodePoolResourceName)
 	nodePoolResourceNameErrorString = `(must be a valid DNS RFC 1035 label)`
 
+	externalAuthResourceName            = `^[a-zA-Z]([-a-zA-Z0-9]{0,13}[a-zA-Z0-9])?$`
```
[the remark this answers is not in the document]
miguelsorianod: This should be fixed now

internal/validation/validate_externalauth_comprehensive_test.go
```diff
@@ -342,6 +342,120 @@ func TestValidateExternalAuth(t *testing.T) {
...
+			name: "valid external auth resource name - minimum length",
+			newObj: func() *api.HCPOpenShiftClusterExternalAuth {
+				obj := createValidExternalAuth()
+				obj.ID.Name = "a"
+				obj.Name = "a"
+				return obj
+			}(),
+			op:           operation.Operation{Type: operation.Create},
+			expectErrors: nil,
+		},
```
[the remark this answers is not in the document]
miguelsorianod: This should be fixed now

bennerv (approved):
/lgtm

PR #4950, opened 2026-04-20
fix: update cluster resource id name regex to match what CS enforces

miguelsorianod:
The cluster resource id name regex was allowing a hyphen as the last
character. This didn't match what CS currently enforces nor the
documented typespec associated to it. This commit fixes that.

### DIFF
--- a/internal/validation/validators.go
+++ b/internal/validation/validators.go
@@ -236,7 +236,7 @@ var (
 	clusterResourceNameRegex       = regexp.MustCompile(clusterResourceName)
 	clusterResourceNameErrorString = `(must be a valid DNS RFC 1035 label)`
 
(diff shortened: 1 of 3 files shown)


internal/validation/validators.go
```diff
@@ -236,7 +236,7 @@ var (
 	clusterResourceNameRegex       = regexp.MustCompile(clusterResourceName)
 	clusterResourceNameErrorString = `(must be a valid DNS RFC 1035 label)`
 
-	nodePoolResourceName            = `^[a-zA-Z][-a-zA-Z0-9]{1,13}[a-z-A-Z0-9]$`
+	nodePoolResourceName            = `^[a-zA-Z][-a-zA-Z0-9]{1,13}[a-zA-Z0-9]$`
```
miguelsorianod, line 239: I checked the typespec and the openapi files and they were already specifying the new regex

stevekuznetsov (approved):
/lgtm
/approve

PR #4955, opened 2026-04-20
test: add built-in retry for transient cluster update errors

adalrsjr1:
https://redhat.atlassian.net/browse/ARO-25884

### What

Add built-in retry with exponential backoff to both `UpdateHCPCluster` variants (`v20240610` and `v20251223`) for transient errors. All cluster update callers now benefit automatically.

### Why

The IDMS E2E test fails with `400: Cluster is in state pending_update` due to a race condition during cluster updates:

1. Frontend calls cluster-service synchronously, cluster-service commits state change to its database
2. Frontend attempts Cosmos DB update, fails with 412 Precondition Failed (ETag conflict)
3. Frontend returns HTTP 500 to client
4. No rollback occurs, cluster stuck in "Updating" state
5. Client retry gets HTTP 400 "can't update cluster in Updating state"

### Testing

- `go build ./test/...` passes
- IDMS E2E test no longer fails with `pending_update` 400 errors (validated in prior periodic and triggered runs)

**Retry config:** 5 attempts, 1m initial delay with factor 2.0 (~15 min total coverage). Covers the worst case of ~10 minutes for a cluster to leave a transitional state.

**Retryable errors:**
- HTTP 500: e.g. Cosmos DB ETag conflict
- HTTP 409: Conflict
- HTTP 400: only when matching the CS error regex `is in state '<state>', can't update` (not a broad substring match)

**Context-aware retry:** uses `wait.ExponentialBackoffWithContext` instead of `retry.OnError` to respect cancellation.

**checkOperationResult:** the v20251223 variant does not call `checkOperationResult` because the v20251223 SDK does not yet have a matching Get response type. The v20240610 variant does this verification.

**Follow-up:** [ARO-26422](https://redhat.atlassian.net/browse/ARO-26422) for applying the same retry pattern to nodepool update helpers.

### DIFF
--- a/test/e2e/idms_lifecycle.go
+++ b/test/e2e/idms_lifecycle.go
@@ -167,13 +167,9 @@ var _ = Describe("Customer", func() {
 					},
 				},
 			}
+
 			updateAddResp, err := framework.UpdateHCPCluster20251223(
-				ctx,
-				hcpClient,
-				*resourceGroup.Name,
-				customerClusterName,
-				updateAdd,
-				10*time.Minute,
+				ctx, hcpClient, *resourceGroup.Name, customerClusterName, updateAdd, 10*time.Minute,
 			)
 			Expect(err).NotTo(HaveOccurred())
 
@@ -220,13 +216,9 @@ var _ = Describe("Customer", func() {
 					},
 				},
 			}
+
 			updateRemoveResp, err := framework.UpdateHCPCluster20251223(
-				ctx,
-				hcpClient,
-				*resourceGroup.Name,
-				customerClusterName,
-				updateRemove,
-				10*time.Minute,
+				ctx, hcpClient, *resourceGroup.Name, customerClusterName, updateRemove, 10*time.Minute,
 			)
 			Expect(err).NotTo(HaveOccurred())
 
--- a/test/util/framework/hcp_helper.go
+++ b/test/util/framework/hcp_helper.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
 
@@ -339,15 +340,20 @@ func UpdateHCPCluster(
 
 // stateConflictBackoff is the retry config for transient state conflicts (ARO-25884).
 var stateConflictBackoff = wait.Backoff{
-	Steps:    3,               // up to 3 attempts total
+	Steps:    5,               // up to 5 attempts total
 	Duration: 1 * time.Minute, // initial wait before first retry
-	Factor:   2.0,             // double the wait each retry (1m, 2m)
+	Factor:   2.0,             // double the wait each retry (1m, 2m, 4m, 8m)
 	Jitter:   0.1,             // ±10% randomization to avoid thundering herd
 }
+
+// csStateConflictPattern matches the cluster-service error message format:
+// "Cluster '<id>' is in state '<state>', can't update"
+var csStateConflictPattern = regexp.MustCompile(`is in state '[^']+', can't update`)
 
 // isTransientUpdateError detects errors worth retrying during cluster updates:
 //   - HTTP 500: e.g. Cosmos DB ETag conflict after cluster-service commit
 //   - HTTP 409: Conflict
+//   - HTTP 400: cluster-service state conflict when cluster is in a transitional state
 func isTransientUpdateError(err error) bool {
 	if err == nil {
 		return false
@@ -356,8 +362,15 @@ func isTransientUpdateError(err error) bool {
 	if !errors.As(err, &responseError) {
 		return false
 	}
-	return responseError.StatusCode == http.StatusInternalServerError ||
-		responseError.StatusCode == http.StatusConflict
+	if responseError.StatusCode == http.StatusInternalServerError ||
+		responseError.StatusCode == http.StatusConflict {
+		return true
+	}
+	if responseError.StatusCode == http.StatusBadRequest &&
+		csStateConflictPattern.MatchString(responseError.Error()) {
+		return true
+	}
+	return false
 }
 
 // UpdateHCPCluster20251223 updates an HCP cluster using the v20251223preview SDK and waits for the operation to complete.
--- a/test/util/framework/hcp_helper.go
+++ b/test/util/framework/hcp_helper.go
@@ -41,6 +41,7 @@ import (
 	v1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -261,6 +262,8 @@ func waitForHCPClusterDeletion(
 
 // UpdateHCPCluster sends a PATCH (BeginUpdate) request for an HCP cluster and waits for completion
 // within the provided timeout. It returns the final update response or an error.
+// UpdateHCPCluster updates an HCP cluster using the v20240610preview SDK and waits for the operation to complete.
+// Transient 500 and 409 errors are retried automatically with exponential backoff.
 func UpdateHCPCluster(
 	ctx context.Context,
 	hcpClient *hcpsdk20240610preview.HcpOpenShiftClustersClient,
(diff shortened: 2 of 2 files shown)


test/e2e/idms_lifecycle.go
```diff
@@ -36,6 +37,107 @@ import (
 	"github.com/Azure/ARO-HCP/test/util/verifiers"
 )
 
+// updateClusterWithRetry attempts to update a cluster up to 3 times, handling state conflicts.
+// It checks if updates already applied when encountering 4XX errors.
+func updateClusterWithRetry(
```
deads2k: how about a construct similar to https://github.com/kubernetes/client-go/blob/v0.35.4/util/retry/util.go#L103 and change the framework.UpdateHCPCluster20251223 to require the pattern

adalrsjr1: @deads2k I made changes on https://github.com/Azure/ARO-HCP/pull/4955/changes/9a83bbda9ed1383211433a35bd7dfe0f2dd4dc17 so the `UpdateHCPCluster20251223WithRetry` works similar to `RetryOnConflict`

raelga:
As discussed with @adalrsjr1, some improvements have been added to this PR.

test/util/framework/hcp_helper.go
```diff
@@ -320,25 +322,118 @@ func UpdateHCPCluster20251223(
...
+		responseError.StatusCode == http.StatusConflict
+}
 
-	operationResult, err := poller.PollUntilDone(ctx, &runtime.PollUntilDoneOptions{
-		Frequency: StandardPollInterval,
-	})
-	if err != nil {
-		if errors.Is(err, context.DeadlineExceeded) {
-			return nil, fmt.Errorf("failed waiting for hcpCluster=%q in resourcegroup=%q to finish updating, caused by: %w, error: %w", hcpClusterName, resourceGroupName, context.Cause(ctx), err)
+func UpdateHCPCluster20251223WithRetry(
```
machi1990: Why a different method vs making the existing Updates withones retrieable on the 5xx  and 429, 400 errors?
And for 400 - we check that it is an error coming from CS and only retry it there.
I am saying this because other e2e can benefit from this pattern of built in retry mechanism as the flakiness wasn't a property of the IDMS lifecycle e2e test itself but it could happen for other updates perform for other test for clusters/nodepool resources

machi1990 (changes requested):

test/util/framework/hcp_helper.go
```diff
@@ -320,25 +322,118 @@ func UpdateHCPCluster20251223(
...
+			"error", err.Error())
 	}
 
-	return &operationResult.HcpOpenShiftCluster, nil
+	return hcpOpenShiftCluster, err
+}
+
+// waitForClusterReady polls until the cluster's provisioningState is no longer
+// a transitional state (Updating, Creating), or until the context is cancelled.
+func waitForClusterReady(
```
machi1990: Why is this needed?

test/util/framework/hcp_helper.go
raelga: Baked retry into `UpdateHCPCluster20251223` directly, all callers now benefit automatically. Removed the `WithRetry` variant. Fixed in 86c20c9

test/util/framework/hcp_helper.go
raelga: The exponential backoff (1m, 2m with jitter) already provides sufficient delay. The waitForClusterReady poll was adding unnecessary complexity and could hang up to 5 minutes per attempt. [54793ef](https://github.com/Azure/ARO-HCP/pull/4955/commits/54793ef1caa4c2c8e0786fbed33c8af5fa31ccc5)

raelga:
cc @copilot

raelga (approved):
/lgtm

test/util/framework/hcp_helper.go
```diff
@@ -272,46 +275,93 @@ func UpdateHCPCluster(
...
+func isTransientUpdateError(err error) bool {
+	if err == nil {
+		return false
+	}
+	var responseError *azcore.ResponseError
+	if !errors.As(err, &responseError) {
+		return false
+	}
+	return responseError.StatusCode == http.StatusInternalServerError ||
+		responseError.StatusCode == http.StatusConflict
```
machi1990: Let's also include

400 as well when the error contains `"Cluster '%s' is in state '%s', can't update"` which CS returns when the cluster isn't ready: 

To error can look like 
```
 PATCH https://management.azure.com/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/idms-wlx748/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/idms-e2e-hcp-cluster
    RESPONSE 400: 400 Bad Request
    {
      "error": {
        "message": "Cluster <foo-bar> is in state 'pending_update', can't update"
      }
    }

```

or 

```
 PATCH https://management.azure.com/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/idms-wlx748/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/idms-e2e-hcp-cluster
    RESPONSE 400: 400 Bad Request
    {
      "error": {
        "message": "Cluster <> is in state 'validating_update', can't update"
      }
    }
```

When CS just accepted the update and is processing it.
It comes from https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/cmd/clusters-service/service/azure/aro_hcp_cluster_service.go#L535

test/util/framework/hcp_helper.go
```diff
@@ -261,6 +262,8 @@ func waitForHCPClusterDeletion(
 
 // UpdateHCPCluster sends a PATCH (BeginUpdate) request for an HCP cluster and waits for completion
 // within the provided timeout. It returns the final update response or an error.
+// UpdateHCPCluster updates an HCP cluster using the v20240610preview SDK and waits for the operation to complete.
+// Transient 500 and 409 errors are retried automatically with exponential backoff.
 func UpdateHCPCluster(
```
machi1990, line 266: This is awesome, let's update node pool one too (in a separate PR) because it can have the same issue.

The corresponding CS error comes from here https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/cmd/clusters-service/service/azure/aro_hcp_node_pool_service.go#L468

Let's cater for the same 500, 429 and 400 errors: the 400 is conditioned on the CS error linked above

machi1990 (changes requested):
Thank you for the update. Left two comments
- To cater for 400 with the specified error code from CS
- To also modify NodePool update helper to handle replies, in a separate PR so that we can get the current PR one in and followup on top of it

test/util/framework/hcp_helper.go
```diff
@@ -272,46 +275,93 @@ func UpdateHCPCluster(
...
 }
 
-// UpdateHCPCluster20251223 updates an HCP cluster using the v20251223preview SDK and waits for the operation to complete
+// stateConflictBackoff is the retry config for transient state conflicts (ARO-25884).
+var stateConflictBackoff = wait.Backoff{
+	Steps:    3,               // up to 3 attempts total
+	Duration: 1 * time.Minute, // initial wait before first retry
+	Factor:   2.0,             // double the wait each retry (1m, 2m)
+	Jitter:   0.1,             // ±10% randomization to avoid thundering herd
+}
```
machi1990, line 345: I am thinking if the 3 attempts are going to be enough;
In the situations where the request has been made to CS; the cluster becoming ready again in CS for the next attempt to be accepted by it will be depend on how fast the request is processed in the management cluster; it's in order of minutes. Experience tells me it's capped at 10 minutes and can be up to that. Can we increase the retry attempts so that we cater for that?

test/util/framework/hcp_helper.go
raelga: Agreed, opened https://redhat.atlassian.net/browse/ARO-26422 for the nodepool retry as a follow-up PR.

test/util/framework/hcp_helper.go
raelga: Bumped to 5 attempts in `f66d1f0`. With factor 2.0 that gives ~15 minutes of total backoff (1m + 2m + 4m + 8m), which covers the 10-minute worst case for the cluster to leave the transitional state.

test/util/framework/hcp_helper.go
```diff
@@ -272,46 +275,101 @@ func UpdateHCPCluster(
...
+	var responseError *azcore.ResponseError
+	if !errors.As(err, &responseError) {
+		return false
 	}
+	if responseError.StatusCode == http.StatusInternalServerError ||
+		responseError.StatusCode == http.StatusConflict {
+		return true
+	}
+	if responseError.StatusCode == http.StatusBadRequest &&
+		strings.Contains(responseError.Error(), "can't update") {
```
machi1990: can we use a regex to match the "Cluster '%s' is in state '%s', can't update" substring? An open question

test/util/framework/hcp_helper.go
raelga: Good call. Tightened it in 770233676 to use a regex matching the exact CS format: `is in state '<state>', can't update` instead of a broad substring match.

test/util/framework/hcp_helper.go
raelga: Done in 770233676. Using a compiled regex `is in state '[^']+', can't update` that matches the exact CS error format.

test/util/framework/hcp_helper.go
```diff
@@ -261,6 +262,8 @@ func waitForHCPClusterDeletion(
 
 // UpdateHCPCluster sends a PATCH (BeginUpdate) request for an HCP cluster and waits for completion
 // within the provided timeout. It returns the final update response or an error.
+// UpdateHCPCluster updates an HCP cluster using the v20240610preview SDK and waits for the operation to complete.
+// Transient 500 and 409 errors are retried automatically with exponential backoff.
```
tuxerrante: **Nit:** Looks like there are now two doc comments stacked on `UpdateHCPCluster` — the original ("sends a PATCH (BeginUpdate) request...") and the new one ("updates an HCP cluster using the v20240610preview SDK..."). One of them should be removed to keep the godoc clean.

test/util/framework/hcp_helper.go
tuxerrante: **Suggestion:** Using `strings.Contains(responseError.Error(), "can't update")` for detecting the 400 case is a bit fragile — error message wording could change on the cluster-service side without notice.

A more robust approach would be to check the ARM error code via `responseError.ErrorCode`:
```go
if responseError.StatusCode == http.StatusBadRequest &&
    responseError.ErrorCode == "InvalidRequestContent" {
    return true
}
```

For reference, ARO-RP already defines this as a constant ([`CloudErrorCodeInvalidRequestContent`](https://github.com/Azure/ARO-RP/blob/787f2cfad5f0ec594f777aa54a2ff7c9090cc392/pkg/api/error.go#L67)) and uses `responseError.ErrorCode` throughout its error classification layer (e.g., [`azureerrors/error.go:92`](https://github.com/Azure/ARO-RP/blob/787f2cfad5f0ec594f777aa54a2ff7c9090cc392/pkg/util/azureerrors/error.go#L92)).

It might also be worth defining `"InvalidRequestContent"` as a constant in a shared package (or in the test framework) rather than using a raw string — that way both this check and any future error classification can reference it without risk of typos.

test/util/framework/hcp_helper.go
```diff
@@ -272,46 +275,101 @@ func UpdateHCPCluster(
...
+		strings.Contains(responseError.Error(), "can't update") {
+		return true
+	}
+	return false
 }
 
-// UpdateHCPCluster20251223 updates an HCP cluster using the v20251223preview SDK and waits for the operation to complete
+// UpdateHCPCluster20251223 updates an HCP cluster using the v20251223preview SDK and waits for the operation to complete.
+// Transient 500 and 409 errors are retried automatically with exponential backoff.
 func UpdateHCPCluster20251223(
```
tuxerrante, line 376: **Observation:** The retry logic in `UpdateHCPCluster20251223` is nearly identical to `UpdateHCPCluster` above (~45 lines each). The only differences are the SDK type parameters and the `checkOperationResult` verification in the v20240610 variant.

If more SDK versions are added or the retry config changes (e.g., adding new transient error codes), both copies need updating. It might be worth extracting the retry-with-backoff wrapper into a shared generic helper to keep maintenance lighter.

Also — is the missing `checkOperationResult` call in this variant intentional? The v20240610 variant above verifies the operation result against a fresh GET, but this one doesn't. If intentional, a short comment explaining why would help future readers.

test/util/framework/hcp_helper.go
raelga: Fixed in 8a8c726f9. Removed the stale doc comment, kept the updated one.

test/util/framework/hcp_helper.go
raelga: The ErrorCode approach would match any 400 with InvalidRequestContent, which is too broad. Validation errors from ARM also use that code. The regex in 770233676 matches the specific CS error pattern `is in state '<state>', can't update` so it only retries the state conflict case.

test/util/framework/hcp_helper.go
raelga: Agree on the duplication being a future cleanup candidate. For now both variants are small enough that keeping them separate is simpler than a generic wrapper.

The missing checkOperationResult in v20251223 is intentional. Added a comment explaining why in da7979111.

machi1990 (commented):
Will add a lgtm once @tuxerrante is happy about his reviews and once the commits are squashed (not a blocker if a clear reasoning is given but highly recommended). I feel like these 18 commits are not atomic enough on themselves when landed alone in the main branch; it's an all or nothing at this state so we can just squash them together when the reviews have been addressed and we are at a point we can merge: we are there, so let's squash.

test/util/framework/hcp_helper.go
tuxerrante: Maybe this could be a nice follow-up, regex doesn't fill right as a permanent solution, but it is not a blocker for sure!

tuxerrante (approved):
LGTM

raelga:
@machi1990 Squashed to 3 (logical) commits.

machi1990 (approved):
/lgtm

raelga:
/test e2e-parallel

`"No more role assignments can be created."`

PR #4957, opened 2026-04-20
Fix joined error masking in RG cleanup error handling

mgahagan73:
Found by Claude and Copilot while working on https://redhat.atlassian.net/browse/ARO-24036 . This also includes a unit test. 

### What

isIgnorableResourceGroupCleanupError used errors.As which matches any sub-error in a joined error tree. This meant errors.Join(realErr, notFoundErr) was treated as entirely ignorable, silently masking real cleanup failures. Now recursively checks that all sub-errors are ignorable before returning true.

### Why

We may be missing the proper cleanup failures, so this should ensure the correct error(s) make it into the logs to make debugging easier in the future.


mvacula02 (approved):
/lgtm

mvacula02 (approved):
/lgtm

PR #4874, opened 2026-04-14
Remove swft deploy environment

bennerv:
### What

Remove the dedicated `swft` deploy environment and make the demo script unconditionally create the Swift subnet and delegation.

### Why

All `pers` clusters now have Swift V2 enabled by default, making the separate `swft` environment redundant.

### Testing

- `cd config; make materialize` passes validation and rendering
- Helm fixture tests pass
- No remaining references to `DEPLOY_ENV=swft` in the codebase

### Special notes for your reviewer

Changes:
- Removed `swft` environment definition from `tooling/templatize/settings.yaml`
- Removed `swft` environment config section from `config/config.yaml`
- Removed rendered config `config/rendered/dev/swft/uksouth.yaml`
- Updated `demo/` to only support creating production clusters, allow private key vault support too
- Updated `dev-infrastructure/swift/README.md` to reflect that pers environments have Swift V2 enabled by default

PR #4958, opened 2026-04-21
istio: phase 2 - target asm-1-28 in dev rollout overrides

trevorwilliams2025:
### What

Move the remaining dev-cloud Istio rollout overrides to target `asm-1-28`.

This updates `targetVersion` to `asm-1-28` for the temporary `svc.istio`
overrides in `dev`, `cspr`, and `perf`, while keeping
`versions: "asm-1-26,asm-1-28"` for now.

### Why

`asm-1-26` is no longer available.

This change advances the rollout by making `asm-1-28` the active target in the
remaining canary environments without yet removing the older revision from the
declared version list.

### Testing

* `make -C config validate`
* `make -C config materialize`

Changes:
* kept the shared dev default at `targetVersion: "asm-1-28"` and `versions: "asm-1-28"`
* updated the temporary `svc.istio` overrides in `dev`, `cspr`, and `perf` to target `asm-1-28`
* left `versions: "asm-1-26,asm-1-28"` in place for those overrides as a temporary rollout state
* added comments in the override blocks to make the follow-up cleanup explicit
* swft environment is ephemeral and have adjusted for that.
* will most likely need to rebase as swft components have been removed before this will be merged.
* https://github.com/Azure/ARO-HCP/pull/4874

A follow-up will remove `asm-1-26` from `versions` and drop the redundant
overrides once the upgrade is complete.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1149,8 +1149,8 @@ clouds:
         nsp:
           accessMode: 'Learning'
         istio:
-          targetVersion: "asm-1-26"
-          versions: "asm-1-26,asm-1-28"
+          targetVersion: "asm-1-28"
+          versions: "asm-1-28"
         aks:
           etcd:
             softDelete: false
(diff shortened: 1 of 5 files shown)


tuxerrante (commented):
**Config consolidation suggestion:** The temporary istio override block (`targetVersion: "asm-1-28"` / `versions: "asm-1-26,asm-1-28"`) is repeated in dev, cspr, and perf. Would it be simpler to set the dual-version config once at `clouds.dev.defaults.svc.istio` and only override in swft? That way phase 3 cleanup is a single-location change. Not blocking — just a thought if it fits the rollout strategy.

PR #4960, opened 2026-04-21
Lower upload_file_size

janboll:
https://redhat.atlassian.net/browse/AROSLSRE-680

### What/Why

We have multiple Kusto outputs, all using 48 MByte as flush peak memory, where given the gzip implementation it requires more than that. With i.e. 3 concurrent flushes this could worst case mean around 330MByte for flushes.

Reducing the flush trigger (lower upload_file_size), this triggers uploads faster, reducing the concurrent memory to around 110 MByte.

### DIFF
--- a/observability/arobit/values-mgmt.yaml
+++ b/observability/arobit/values-mgmt.yaml
@@ -69,7 +69,7 @@ forwarder:
     buffer_dir: /var/kusto
     buffering_enabled: On
     upload_timeout: 1m
-    upload_file_size: 48M
+    upload_file_size: 16M
     buffer_file_delete_early: Off
     unify_tag: On
     store_dir_limit_size: 1.8GB
--- a/observability/arobit/values-svc.yaml
+++ b/observability/arobit/values-svc.yaml
@@ -69,7 +69,7 @@ forwarder:
     buffer_dir: /var/kusto
     buffering_enabled: On
     upload_timeout: 1m
-    upload_file_size: 48M
+    upload_file_size: 16M
     buffer_file_delete_early: Off
     unify_tag: On
     store_dir_limit_size: 1.8GB
--- a/dev-infrastructure/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_arobit.yaml
+++ b/dev-infrastructure/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_arobit.yaml
@@ -188,7 +188,7 @@ data:
         buffer_dir "/var/kusto"
         buffering_enabled           true 
         upload_timeout              1m
-        upload_file_size            48M
+        upload_file_size            16M
         buffer_file_delete_early    false
         unify_tag                   true
         store_dir_limit_size        1.8GB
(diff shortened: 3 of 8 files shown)


observability/arobit/values-svc.yaml
```diff
@@ -69,7 +69,7 @@ forwarder:
     buffer_dir: /var/kusto
     buffering_enabled: On
     upload_timeout: 1m
-    upload_file_size: 48M
+    upload_file_size: 16M
```
tuxerrante, line 72: This is a Kusto-only knob, but the chart still renders `output.conf` unconditionally and the DaemonSet hashes the whole `ConfigMap`, so this changes the pod template even when `kusto.enabled` is false. That means the disabled path still rolls `arobit-forwarder` for a setting that is never consumed there. Because the rollout strategy uses [`maxUnavailable: 100%`](https://github.com/Azure/ARO-HCP/blob/8fad44f4028356ce521ff041b5853740c53ae4f6/observability/arobit/deploy/templates/forwarder-daemonset.yaml#L17-L20), this can become a full forwarder restart in Kusto-disabled environments.

Refs:
- [`fluent-bit.conf` only includes `output.conf` when Kusto is enabled](https://github.com/Azure/ARO-HCP/blob/8fad44f4028356ce521ff041b5853740c53ae4f6/observability/arobit/deploy/templates/forwarder-configmap.yaml#L23-L35)
- [`output.conf` is still rendered unconditionally, including `upload_file_size`](https://github.com/Azure/ARO-HCP/blob/8fad44f4028356ce521ff041b5853740c53ae4f6/observability/arobit/deploy/templates/forwarder-configmap.yaml#L322-L430)
- [`checksum/configmap` is derived from the full rendered ConfigMap](https://github.com/Azure/ARO-HCP/blob/8fad44f4028356ce521ff041b5853740c53ae4f6/observability/arobit/deploy/templates/forwarder-daemonset.yaml#L21-L29)

observability/arobit/values-mgmt.yaml
```diff
@@ -69,7 +69,7 @@ forwarder:
     buffer_dir: /var/kusto
     buffering_enabled: On
     upload_timeout: 1m
-    upload_file_size: 48M
+    upload_file_size: 16M
```
tuxerrante, line 72: Operational question: have we validated the request-rate tradeoff for this value under sustained load? Dropping from `48M` to `16M` should reduce per-flush peak memory, but it also increases the number of Kusto uploads per stream while the forwarder memory ceiling is still `1Gi` ([source config](https://github.com/Azure/ARO-HCP/blob/8fad44f4028356ce521ff041b5853740c53ae4f6/config/config.yaml#L105-L117), [rendered into the DaemonSet here](https://github.com/Azure/ARO-HCP/blob/8fad44f4028356ce521ff041b5853740c53ae4f6/observability/arobit/deploy/templates/forwarder-daemonset.yaml#L81-L87)). I only see render/fixture updates in this PR, so it would help to link the runtime data or soak result that says `16M` is the right steady-state value.

raelga (approved):
/lgtm

observability/arobit/values-mgmt.yaml
janboll: Before this we were running without buffering, meaning chunks were sent right away. That did not cause issue, and this config still uses larger batches of data.

observability/arobit/values-svc.yaml
janboll: Was not aware we have kusto disabled anywhere :thinking: 
I think this needs to be removed

PR #4966, opened 2026-04-21
Moving k8s audit logs to separate table

rachelvweber:
### What

Pushing aks audit logs to a separate table

### Why

This parses the audit log schema into columns for easier querying

### Testing

Ran the script against dev kusto cluster to confirm that it works and that logs are directed correctly

### DIFF
--- a/dev-infrastructure/modules/logs/kusto/tables/aksEvents.kql
+++ b/dev-infrastructure/modules/logs/kusto/tables/aksEvents.kql
@@ -24,13 +24,39 @@ with (docstring = "Temporary table for storing events from aks cluster")
 )
 with (docstring = "Stores events from aks cluster")
 
+.create-merge table kubeAudit (
+    category:string,
+    operationName:string,
+    containerId:string,
+    stream:string,
+    pod:string,
+    resourceId:string,
+    timestamp:datetime,
+    objectKind:string,
+    apiVersion:string,
+    level:string,
+    auditID:string,
+    stage:string,
+    requestURI:string,
+    verb:string,
+    user:dynamic,
+    sourceIPs:dynamic,
+    userAgent:string,
+    objectRef:dynamic,
+    responseStatus:dynamic,
+    requestReceivedTimestamp:datetime,
+    stageTimestamp:datetime,
+    annotations:dynamic
+)
+with (docstring = "Stores kube-audit events from aks cluster")
+
 .create-or-alter function 
 with (docstring = 'process aks events', skipvalidation = "true")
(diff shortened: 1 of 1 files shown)


dev-infrastructure/modules/logs/kusto/tables/aksEvents.kql
```diff
@@ -40,4 +68,52 @@ rawAksEventsTransform()  {
   | project-away records
 }
 
-.alter table aksEvents policy update @'[{"IsEnabled": true, "Source": "rawAksEvents", "Query": "rawAksEventsTransform()", "IsTransactional": true, "PropagateIngestionProperties": false}]'
\ No newline at end of file
+.create-or-alter function
```
stevekuznetsov, line 69: Do we store fields from the parsed log we didn't pick out specifically, or are they lost?

rachelvweber: This should be everything, but in theory they could change schemas. I can keep the log column if you prefer

rachelvweber: In addition to the other columns

stevekuznetsov: There's some ingest parameter in Kusto to dump the unconsumed fields as a separate dynamic column, if we can use that it would be cool. Seems unlikely to change schema in the future, though, so no big deal.

dev-infrastructure/modules/logs/kusto/tables/aksEvents.kql
```diff
@@ -24,6 +24,34 @@ with (docstring = "Temporary table for storing events from aks cluster")
...
+    timestamp:datetime,
+    objectKind:string,
+    apiVersion:string,
+    level:string,
+    auditID:string,
+    stage:string,
+    requestURI:string,
+    verb:string,
+    user:dynamic,
+    sourceIPs:dynamic,
```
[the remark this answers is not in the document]
rachelvweber: I'm matching the real column names from the audit log schema

dev-infrastructure/modules/logs/kusto/tables/aksEvents.kql
rachelvweber: Do you mind if we leave as is? we can use bag unpack, but I think we have less control over the resulting types. I think it's unlikely they'll change the schema as well.

stevekuznetsov (approved):
/lgtm
/approve

PR #4969, opened 2026-04-22
Fix ci /go/pkg/mod failures

sclarkso:
### What

creates /go/pkg/mod in ci image and sets permissions

### Why

PR 4884 added `go clean -cache -modcache` which removes /go/pkg/mod, as such all PR's are failing.


janboll:
This needs to be force merged, as nothing is passing CI checks in presubmits. See PR comment: `PR 4884 added go clean -cache -modcache which removes /go/pkg/mod, as such all PR's are failing.`

PR #4973, opened 2026-04-22
Fix must-gather-clean

janboll:
https://redhat.atlassian.net/browse/AROSLSRE-537

### What/why
Add all matched uuids to the obfuscation config and not only the first match


raelga (approved):
/lgtm

PR #4976, opened 2026-04-22
sessiongate: generate OpenAPI schema for applyconfiguration field management

geoberle:
### What

The generated fake clientset's NewClientset() requires OpenAPI schema
definitions for FieldManagedObjectTracker to resolve Session CRD fields
during server-side apply. Without this, NewTypeConverter returns empty
type models and SSA field management silently degrades to untyped.

this is a prepwork PR for #4926


raelga (approved):
/lgtm

PR #4986, opened 2026-04-22
Add JUnit output for fired alerts and fix chart gap rendering

geoberle:
### What

Alerts fired during e2e runs are now written as JUnit XML to junit_alerts.xml alongside existing artifacts. This enables Prow to display alert results inline and Sippy to track alert regressions across runs via pass-rate monitoring.

Known-issue alerts are marked <skipped> (invisible to Sippy), unknown alerts are marked <failure> (tracked for regression detection). Alerts are grouped by (alertname, stable-labels) to produce stable test names across Prow jobs.

Additionally, time series charts now break lines across data gaps instead of drawing misleading straight lines between distant points.

### DIFF
--- a/observability/alerts/prometheus-prometheusRule.yaml
+++ b/observability/alerts/prometheus-prometheusRule.yaml
@@ -14,16 +14,16 @@ spec:
     rules:
     - alert: PrometheusJobUp
       expr: group by (cluster) (up{job="kube-state-metrics"}) unless on(cluster) group by (cluster) (up{job="prometheus/prometheus",namespace="prometheus"} == 1)
-      for: 5m
+      for: 10m
       labels:
         severity: critical
       annotations:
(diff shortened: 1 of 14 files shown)


observability/alerts/prometheus-prometheusRule.yaml
```diff
@@ -14,16 +14,16 @@ spec:
     rules:
     - alert: PrometheusJobUp
       expr: group by (cluster) (up{job="kube-state-metrics"}) unless on(cluster) group by (cluster) (up{job="prometheus/prometheus",namespace="prometheus"} == 1)
-      for: 5m
+      for: 10m
```
geoberle, line 17: 5m was too flakey during e2e tests

geoberle:
change on gather-observability works

PR #4988, opened 2026-04-22
hypershift: use the SRE metrics set, opt into everything

stevekuznetsov:
Previously, we seem to have been using the telemetry metrics set for hosted clusters, which leads us to miss most of the useful metrics we need from the control planes. The SRE metrics set mechanism exists in HyperShift to support this exact use case - we opt into it with the install flag and supply the required ConfigMap to configure which metrics we need. For now, we ask for all metrics from the components most likely to matter for us. If we run into cardinality issues we can always winnow these down later.


bennerv (approved):
/lgtm
/approve

PR #4990, opened 2026-04-22
RG Cleanup Job Improvements

sclarkso:
<!-- Link to Jira issue -->

### What

test/util/framework/per_test_framework.go                                                                                                                                                                                                                                  
  - New isResourceGroupExpiredBeyond() - checks if a resource group's deleteAfter.aro-hcp-ci.redhat.com tag is expired by more than a given duration
  - Modified cleanupResourceGroup() - when managed RG deletion times out and the parent's deleteAfter tag is >7 days expired, falls back to the ordered cleanup workflow (same approach as cleanupResourceGroupNoRP): deletes resources inside each managed RG in dependency order (private endpoints, DNS, VNets, NSGs, etc.) then deletes the RG itself. ContinueOnError is set because some platform-managed resources (e.g. serviceAssociationLinks) may not be deletable. For RGs expired <7 days, the existing error behavior is preserved so stuck managed RGs remain visible as a signal of service issues. 

### Why
There are instances where we have mrg and rgs that are well expired (>10 in int currently), that was caused by code issues.  Those issues have been found and fixed, but we have no way of cleaning up the resources.

### Testing

```
09:06:03 ~/repos/ARO-HCP(rg-cleanup-features*) $ AZURE_TOKEN_CREDENTIALS=dev CUSTOMER_SUBSCRIPTION='ARO SRE Team - INT (EA Subscription 3)' /tmp/aro-hcp-tests cleanup resource-groups --expired --dry-run
cluster-listing-2s9rvc
cluster-listing-2wqbzd
idms-2n6xb2
idms-t88mxz
idms-mclzt7
idms-8r7q26
idms-2p7bxl
idms-vnl9cl
rg-cluster-nsg-subnet-reuse-fdfcnq
cluster-listing-xk5f98
rg-np-version-upgrade-v69w84-hxx6l8
rg-np-version-upgrade-6rp87p-kgwjfc
```

<!-- optional -->

### DIFF
--- a/test/util/cleanup/resourcegroups/run.go
+++ b/test/util/cleanup/resourcegroups/run.go
@@ -89,6 +89,15 @@ func (o *Options) Run(ctx context.Context) error {
 		for _, rg := range resourceGroupsToDelete {
 			fmt.Println(rg)
 		}
+		if o.DeleteExpired {
+			orphanedManagedRGs, err := framework.ListOrphanedManagedResourceGroups(ctx, resourceGroupsClient)
+			if err != nil {
+				return fmt.Errorf("failed to list orphaned managed resource groups: %w", err)
+			}
+			for _, rg := range orphanedManagedRGs {
+				fmt.Printf("%s (orphaned managed)\n", *rg.Name)
+			}
+		}
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


sclarkso:
@raelga @roivaz 

A previous PR to clean up orphaned MRG's was closed without merging (I had that change in this code as well, but removed it after finding that PR).   Right now we have RG's in INT (and above) where cleaning up the mrg times out, and cleanup fails.  This changes the cleanup job to attempt to cleanup the MRG (if they're old) instead of just erroring and failing.

PR #4983, opened 2026-04-22
Test framework custom 429 retry policy

roivaz:
Add a custom `throttleRetryPolicy` that handles 429 (Too Many Requests) responses with conservative exponential backoff, and remove 429 from the SDK's built-in retry status codes.

The Azure SDK's default retry policy uses `Retry-After` as the delay **verbatim** -- it does not apply exponential backoff when `Retry-After` is present. With `Retry-After: 1s` and `MaxRetries: 6`, the SDK fires 7 rapid requests in ~6 seconds. 

The new `throttleRetryPolicy` is an outer `PerCallPolicy` that sits before the SDK retry loop. 429 is excluded from the SDK's inner retry `StatusCodes`, so throttled responses bubble up immediately (1 request, no burst). The outer policy applies `max(exponentialBackoff, retryAfter)` -- the server's `Retry-After` is a floor, not a ceiling.

### Before vs After: sustained 429, assuming `Retry-After: 1s`

**Before (SDK default):**

| Attempt | Wait | Elapsed | Total requests |
|---------|------|---------|----------------|
| 1 | -- | 0s | 1 |
| 2 | 1s | ~1s | 2 |
| 3 | 1s | ~2s | 3 |
| 4 | 1s | ~3s | 4 |
| 5 | 1s | ~4s | 5 |
| 6 | 1s | ~5s | 6 |
| 7 | 1s | ~6s | 7 |
| **FAIL** | | **~6s** | **7 requests** |

**After (throttleRetryPolicy):**

| Attempt | Backoff | Retry-After | Actual delay | Elapsed | Total requests |
|---------|---------|-------------|-------------|---------|----------------|
| 1 | -- | -- | -- | 0s | 1 |
| 2 | ~5s | 1s | ~5s | ~5s | 2 |
| 3 | ~9s | 1s | ~9s | ~14s | 3 |
| 4 | ~17s | 1s | ~17s | ~31s | 4 |
| 5 | ~33s | 1s | ~33s | ~64s | 5 |
| 6 | ~65s | 1s | ~65s | ~129s | 6 |
| **FAIL** | | | | **~129s** | **6 requests** |

| | Before | After |
|---|---|---|
| Requests sent | 7 | 6 |
| Time span | ~6s | ~2min |
| Tokens consumed rate | 1.2/s | 0.05/s |
| Recoverable if throttle lifts after 10s? | No (already failed) | Yes (attempt 3) |

PR #4991, opened 2026-04-23
Az SDK policies - use k8s.io/apimachinery/pkg/util/wait

roivaz:
Follow up to address https://github.com/Azure/ARO-HCP/pull/4983#discussion_r3125873668

### DIFF
--- a/test/util/framework/azure_client_policies.go
+++ b/test/util/framework/azure_client_policies.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math/rand"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -28,6 +27,8 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/onsi/ginkgo/v2"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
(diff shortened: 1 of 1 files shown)


test/util/framework/azure_client_policies.go
```diff
@@ -158,18 +159,20 @@ func (p *correlationRequestIDPolicy) Do(req *policy.Request) (*http.Response, er
...
-		MaxBackoff:     10 * time.Second,
-		MaxRetryWindow: 90 * time.Second,
+		backoff: wait.Backoff{
+			Duration: 2 * time.Second,
+			Factor:   2,
+			Jitter:   0.5,
+			Steps:    5,
+			Cap:      10 * time.Second,
+		},
+		maxRetryWindow: 90 * time.Second,
```
bennerv: Why max retry window when we have a `Cap` field on wait.Backoff?

test/util/framework/azure_client_policies.go
```diff
@@ -183,48 +186,37 @@ func (p *lroPollerRetryDeploymentNotFoundPolicy) Do(req *policy.Request) (*http.
...
-
-	for {
+	backoff := p.backoff
 
-		resp, err, retry := func(req *policy.Request) (resp *http.Response, err error, retry bool) {
-			retryReq := req.Clone(req.Raw().Context())
-			if err := retryReq.RewindBody(); err != nil {
-				return nil, err, false
-			}
+	for attempt := 0; ; attempt++ {
```
stevekuznetsov, line 191: Why are we using the raw backoff struct and implementing the rest ourselves versus just using the top-level functions?

PR #4992, opened 2026-04-23
deps: bump k8s.io/* to v0.35.1

raelga:
https://redhat.atlassian.net/browse/AROSLSRE-699

### What

Bump all `k8s.io/*` dependencies from v0.34.3 to v0.35.1 across the entire Go workspace, with all required code migrations.

### Why

k8s.io/apimachinery v0.35.1 is a prerequisite for bumping helm.sh/helm to fix CVE-2026-35206 ([AROSLSRE-667](https://redhat.atlassian.net/browse/AROSLSRE-667)) and tektoncd/pipeline to fix 5 CVEs.

### Testing

- [x] `go build` passes for all modules (internal, frontend, backend, sessiongate, admin, test, test-integration, tooling)
- [x] `go mod tidy` + `go work sync` clean
- [x] `TestImmutableByCompare` and `TestImmutableByReflect` pass
- [x] `TestListWatchWithoutWatchListSemantics` passes
- [ ] CI passes

**Breaking changes in k8s.io/apimachinery v0.35.1:**
- `validate.ImmutableByReflect`/`ImmutableByCompare` removed, replaced with `validate.Immutable` (uses k8s ratcheting framework). Local helpers added preserving the old comparison semantics. Source documented, tests ported.
- `validate.Enum` gained an exclusions parameter.
- `fake.NewSimpleClientset` deprecated in favor of `fake.NewClientset`.
- `WatchListClient` feature gate enabled by default. CosmosDB-backed informers need explicit opt-out since `ExpiringWatcher` doesn't support the watch-list protocol. Wrapper renamed to `listWatchWithoutWatchListSemantics` to align with client-go's internal type, verified against `DoesClientNotSupportWatchListSemantics`.
- sessiongate generated code regenerated with code-generator v0.35.1.


internal/validation/validators.go
```diff
@@ -40,6 +41,33 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils/apihelpers"
 )
 
+func immutableByCompare[T comparable](_ context.Context, op operation.Operation, fldPath *field.Path, value, oldValue *T) field.ErrorList {
```
geoberle, line 50: lets document where we copied them from
i guess these funcs also had tests, we should bring them over

backend/pkg/informers/informers.go
```diff
@@ -35,6 +35,16 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils/armhelpers"
 )
 
+// noWatchListLW wraps a cache.ListWatch to opt out of WatchList semantics.
+// Client-go v0.35.1 enables WatchListClient by default, which requires the
+// watch stream to send bookmark events. Our CosmosDB-backed informers use
+// ExpiringWatcher which does not support this protocol.
+type noWatchListLW struct {
```
geoberle: lets align with the type name client-go uses internally for their own (sadly unexported wrapper) `listWatcherWithWatchListSemanticsWrapper` 

also since the interface `unSupportedWatchListSemantics` is not exported and we can't type assert, lets at least test our wrapper against the `DoesClientNotSupportWatchListSemantics` func in `client-go/util/watchlist`

internal/validation/validators.go
raelga: Added source documentation and tests in a55dbabfa. The helpers are local copies from k8s.io/apimachinery v0.34.3 pkg/api/validate/immutable.go. Tests cover create (always allowed), update same/different values, nil handling for Compare, and zero-value handling for Reflect.

backend/pkg/informers/informers.go
raelga: Renamed to listWatchWithoutWatchListSemantics in bb2439c83 to align with the internal type name. Also added a test asserting that DoesClientNotSupportWatchListSemantics from client-go/util/watchlist returns true for our wrapper.

raelga:
Rebased due to Go Mod conflicts.

raelga:
Failed for unrelated reasons: `Run pipeline step Microsoft.Azure.ARO.HCP.RP.HypershiftOperator/management/deploy`:

/test e2e-parallel

PR #4993, opened 2026-04-23
Fix metric/add alert

janboll:
https://redhat.atlassian.net/browse/AROSLSRE-481

### What
Fix the metric, so it is actually seconds.
Add an alert that checks if logs are missing for more than 60 minutes


raelga (approved):
/lgtm

PR #4999, opened 2026-04-23
feat: configure the AZURE_TOKEN_CREDENTIALS for cs deployment and set it to 'WorkloadIdentityCredential'

machi1990:
### What

feat: configure the AZURE_TOKEN_CREDENTIALS for cs deployment and set it to 'WorkloadIdentityCredential'

### Why

CS is going to require RequireAzureTokenCredentials and set it to true. Follows up https://github.com/Azure/ARO-HCP/pull/4259 but applied for CS

CS change to require RequireAzureTokenCredentials are being worked on as part of https://redhat.atlassian.net/browse/ARO-26335; we need this merged first and rolled out all the way to prod before those CS changes can be merged

### DIFF
--- a/cluster-service/helm-charts/cluster-service/templates/deployment.yaml
+++ b/cluster-service/helm-charts/cluster-service/templates/deployment.yaml
@@ -100,6 +100,11 @@ spec:
       - name: clusters-service-init
         image: '{{ .Values.imageRegistry }}/{{ .Values.imageRepository }}@{{ .Values.imageDigest }}'
         imagePullPolicy: IfNotPresent
+        env:
+        # We set AZURE_TOKEN_CREDENTIALS to WorkloadIdentityCredential to force workload identity because this pod uses
+        # Azure Workload Identity (azure.workload.identity/use); that effectively restricts other credential chains from being tried.
+        - name: AZURE_TOKEN_CREDENTIALS
+          value: "WorkloadIdentityCredential"
         volumeMounts:
         - name: rds
           mountPath: /secrets/rds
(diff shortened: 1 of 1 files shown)


cluster-service/helm-charts/cluster-service/templates/deployment.yaml
```diff
@@ -100,6 +100,9 @@ spec:
       - name: clusters-service-init
         image: '{{ .Values.imageRegistry }}/{{ .Values.imageRepository }}@{{ .Values.imageDigest }}'
         imagePullPolicy: IfNotPresent
+        env:
+        - name: AZURE_TOKEN_CREDENTIALS
+          value: "WorkloadIdentityCredential"
```
miguelsorianod, line 107: non blocking for this MR but it would be good to add a comment describing what does this do and a reference for further detail, as well as why are we setting this.

cluster-service/helm-charts/cluster-service/templates/deployment.yaml
```diff
@@ -133,6 +136,8 @@ spec:
         env:
         - name: DENYASSIGNMENTS
           value: {{ .Values.denyAssignments }}
+        - name: AZURE_TOKEN_CREDENTIALS
```
miguelsorianod, line 143: Same as https://github.com/Azure/ARO-HCP/pull/4999#discussion_r3130493261

cluster-service/helm-charts/cluster-service/templates/deployment.yaml
machi1990: added

cluster-service/helm-charts/cluster-service/templates/deployment.yaml
machi1990: added

PR #5010, opened 2026-04-23
skip mise routing test in higher envs

mbukatov:
https://redhat.atlassian.net/browse/ARO-26431

### What

Skipping MISE routing E2E test case in staging or prod environments

### Why

The E2E test case was implemented in https://github.com/Azure/ARO-HCP/pull/4886 together with the feature code, which is suggested, but the code doesn't try to avoid running the tests in higher environments. For this reason, this test will fail there until we rollout new version of the service there as well.

This is happening because of know testing and release engineering gaps on our side.

### Testing

When merged, we should create a follow up PR which will revert this change, and there we can schedule the test runs in both environments to check that it works there.

Verify that the assumptions here are correct.

### DIFF
--- a/test/e2e/mise_routing.go
+++ b/test/e2e/mise_routing.go
@@ -17,6 +17,8 @@ package e2e
 import (
 	"context"
 	"net/http"
+	"os"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


sclarkso:
This has passed in stage: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-Azure-ARO-HCP-main-periodic-stage-e2e-parallel/2047379801026596864

raelga (commented):
Time bombed in https://github.com/Azure/ARO-HCP/pull/5079

/close

mbukatov:
> Time bombed in https://github.com/Azure/ARO-HCP/pull/5079

Ack

PR #5011, opened 2026-04-23
bumping ARO-Tools 20260423

rachelvweber:
[ARO-26427](https://redhat.atlassian.net/browse/ARO-26427)

### What

bumping ARO-Tools version

### Why

Pulling over dry run option for prow job step for classic ff environments

### Testing

Tested by e2e test


raelga:
## sdp-pipelines bumper failure: ARO-Tools version mismatch

### Error

```
json: cannot unmarshal object into Go struct field ProwJobValidationStep.ProwJobStep.dryRun of type bool
```

Affects `TestClassicHelmTemplate`, `validate-ado-pipelines`, and `validate-pipelines` targets.

### Root cause

ARO-HCP PR [#5011](https://github.com/Azure/ARO-HCP/pull/5011) bumped ARO-Tools to `0.0.0-20260423220125-ef00889b1671`, which adds the `DryRun` field (as a struct) to `ProwJobStep`. The sdp-pipelines bumper picked up this new ARO-HCP commit, but sdp-pipelines' own ARO-Tools dependency is still on `0.0.0-20260420131316-06a82d6f5207` (which doesn't have `DryRun` on `ProwJobStep`).

Meanwhile, `classic/rp/e2e/pipeline.yaml` already uses `dryRun` as an object:

```yaml
dryRun:
  variables:
  - name: DRY_RUN
    value: "true"
```

The old ARO-Tools version can't deserialize this object into the missing/bool field, causing the unmarshal failure.

### Fix

Bump all ARO-Tools dependencies in sdp-pipelines from `0.0.0-20260420131316-06a82d6f5207` to `0.0.0-20260423220125-ef00889b1671` and run `go mod tidy` across all workspace modules.

PR #5013, opened 2026-04-24
kusto: pass location to audit-logs data connection

weherdh:
https://redhat.atlassian.net/browse/AROSLSRE-714

### Why

The eastus2euap canary region deploys its Kusto cluster in eastus2 (via the location override from #4894), but data connections still used `resourceGroup().location` which resolves to eastus2euap. Since `Microsoft.Kusto/clusters/databases/dataConnections` is not available in eastus2euap, the deployment fails.
This follows the same pattern already applied to `database.bicep` and
`cluster.bicep` in #4894. 

### Testing

- No impact on existing regions: `location` defaults to `resourceGroup().location` for backward compatibility


janboll (commented):
/lgtm

PR #5020, opened 2026-04-24
update hypershift to current so we can use new APIs

deads2k:
this requires also     Switch sessiongate to use dynamic clients to avoid hypershift dep
    
hypershift and openshift/api disagree on k8s.io/api levels.  Because    hypershift also requires a transitive karpenter dependency and it does a    replace on that dependency, we cannot use newer API levels and    hypershift. We need the hypershift/api, but we can do without the    client, so we'll eliminate the usage.

/assign @geoberle 

you look like the primary author here.  Are you confident enough in tests that passing tests mean passing code?


geoberle:
> Are you confident enough in tests that passing tests mean passing code?

yes there is an e2e test covering the entire functionality

PR #5019, opened 2026-04-24
Revert "Merge pull request #4986 from Azure/gather-observability-junit"

stevekuznetsov:
This reverts commit fc11c720c29e5adef3ba973bd7d810c6ea51a929, reversing changes made to a3b2179409e0eb89294db6999c5a7c9e77b5a571.

Failed alerts seem to be causing quite a lot of issues, most e2e that run long enough to actually run tests fail with it

https://search.dptools.openshift.org/?search=should+not+have+fired&maxAge=48h&context=1&type=junit&name=pull-ci-Azure-ARO-HCP-main-e2e-parallel&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

PR #5021, opened 2026-04-24
skip test that fails after the revert

deads2k:
this fails on master after merging https://github.com/Azure/ARO-HCP/pull/5019


deads2k:
will merge on green unit, verify, and lint to unbreak the unbreak.

PR #5025, opened 2026-04-26
backend: Add missing operation controller unit tests

mbarnes:
### What

Noticed a couple operation controllers were missing unit tests.

Small amount of refactoring included as well.

### DIFF
--- a/backend/pkg/controllers/operationcontrollers/operation_request_credential.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_request_credential.go
@@ -33,21 +33,21 @@ import (
 )
 
 type operationRequestCredential struct {
-	cosmosClient         database.DBClient
-	clusterServiceClient ocm.ClusterServiceClientSpec
-	notificationClient   *http.Client
+	cosmosClient          database.DBClient
+	clustersServiceClient ocm.ClusterServiceClientSpec
+	notificationClient    *http.Client
 }
 
 func NewOperationRequestCredentialController(
 	cosmosClient database.DBClient,
-	clusterServiceClient ocm.ClusterServiceClientSpec,
+	clustersServiceClient ocm.ClusterServiceClientSpec,
 	notificationClient *http.Client,
 	activeOperationInformer cache.SharedIndexInformer,
 ) controllerutils.Controller {
 	syncer := &operationRequestCredential{
-		cosmosClient:         cosmosClient,
-		clusterServiceClient: clusterServiceClient,
-		notificationClient:   notificationClient,
+		cosmosClient:          cosmosClient,
+		clustersServiceClient: clustersServiceClient,
+		notificationClient:    notificationClient,
 	}
 
 	controller := NewGenericOperationController(
@@ -86,7 +86,7 @@ func (opsync *operationRequestCredential) SynchronizeOperation(ctx context.Conte
 		return nil // no work to do
 	}
 
-	breakGlassCredential, err := opsync.clusterServiceClient.GetBreakGlassCredential(ctx, oldOperation.InternalID)
+	breakGlassCredential, err := opsync.clustersServiceClient.GetBreakGlassCredential(ctx, oldOperation.InternalID)
 	if err != nil {
 		return utils.TrackError(err)
 	}
--- a/backend/pkg/controllers/operationcontrollers/operation_request_credential_test.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_request_credential_test.go
@@ -1,0 +1,118 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package operationcontrollers
+
+import (
+	"context"
+	"testing"
+
+	"github.com/go-logr/logr/testr"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+
+	"github.com/Azure/ARO-HCP/internal/api/arm"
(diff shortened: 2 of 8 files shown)


backend/pkg/controllers/operationcontrollers/operation_request_credential.go
```diff
@@ -33,21 +33,21 @@ import (
 )
 
 type operationRequestCredential struct {
```
tmstff, line 35: disclaimer: it seems like it is not common practice here - and it is not part of the PR changes - so I'm not gonna insist, but I'm still gonna mention it :-D Feel free to ignore this one ;-)

Personally I like a (short) comment on the intention of a piece of code: what is it used for? What problem does it solve? Of course only, if the names do not already reveal it. But (for me, with my "beginner knowhow" of ARO) this is not obvious :-)

backend/pkg/controllers/operationcontrollers/operation_request_credential_test.go
```diff
@@ -0,0 +1,118 @@
...
+	tests := []struct {
+		name                       string
+		breakGlassCredentialStatus cmv1.BreakGlassCredentialStatus
+		expectError                bool
+		verify                     func(t *testing.T, ctx context.Context, db *databasetesting.MockDBClient, fixture *clusterTestFixture)
+	}{
+		{
+			name:                       "created credential updates operation status to provisioning",
+			breakGlassCredentialStatus: cmv1.BreakGlassCredentialStatusCreated,
+			expectError:                false,
```
tmstff, line 45: maybe remove `expectError`, as it is `false` in all cases?

backend/pkg/controllers/operationcontrollers/operation_request_credential_test.go
```diff
@@ -0,0 +1,118 @@
...
+			verify: func(t *testing.T, ctx context.Context, db *databasetesting.MockDBClient, fixture *clusterTestFixture) {
+				op, err := db.Operations(testSubscriptionID).Get(ctx, testOperationName)
```
tmstff, line 47: as for each verification the same code is executed (`db.Operations(testSubscriptionID).Get...`), this code could move to the for loop which iterates over the test cases.

In all cases `require.NoError(t, err)` is called. You would just need to pass Status and Error code. The latter maybe as pointer, so you could indicate "no error code" with nil.

backend/pkg/controllers/operationcontrollers/operation_request_credential_test.go
```diff
@@ -0,0 +1,118 @@
...
+				Return(breakGlassCredential, nil)
+
+			controller := &operationRequestCredential{
+				cosmosClient:          mockDB,
+				clustersServiceClient: mockCSClient,
+			}
+
+			err = controller.SynchronizeOperation(ctx, fixture.operationKey())
+
+			if tt.expectError {
```
tmstff, line 107: could be removed (see above)

Unless, of course, it would make sense to test an error case - in this case I would suggest to do so ;-)

backend/pkg/controllers/operationcontrollers/operation_revoke_credentials_test.go
```diff
@@ -0,0 +1,174 @@
...
+				})
+
+			controller := &operationRevokeCredentials{
+				cosmosClient:          mockDB,
+				clustersServiceClient: mockCSClient,
+			}
+
+			err = controller.SynchronizeOperation(ctx, fixture.operationKey())
+
+			if tt.expectError {
```
tmstff, line 163: could be removed, as all test cases expect no error.

Unless, of course, it would make sense to test an error case - in this case I would suggest to do so ;-)

backend/pkg/controllers/operationcontrollers/operation_revoke_credentials_test.go
```diff
@@ -0,0 +1,174 @@
...
+				cluster, err := db.HCPClusters(testSubscriptionID, testResourceGroupName).Get(ctx, testClusterName)
+				require.NoError(t, err)
+				assert.Empty(t, cluster.ServiceProviderProperties.RevokeCredentialsOperationID)
+			},
+		},
+		{
+			name:                         "mismatched RevokeCredentialsOperationID is left intact",
+			breakGlassCredentialStatuses: []cmv1.BreakGlassCredentialStatus{},
+			revokeCredentialsOperationID: "not-our-operation-id",
+			verify: func(t *testing.T, ctx context.Context, db *databasetesting.MockDBClient, fixture *clusterTestFixture) {
```
tmstff, line 116: All other test cases seem to follow the pattern first `db.Operations(testSubscriptionID).Get(ctx, testOperationName)` then `db.HCPClusters(testSubscriptionID, testResourceGroupName).Get(ctx, testClusterName)`.

If the former does not hurt here, all could be unified and move to the for loop that iterates over the test cases, as mentioned before. It should suffice to pass the Status - and for the operationId, you might need two parameters (like a bool `operationIdPresent` and a pointer to a specific operationId - which is ignored when `nil`.

backend/pkg/controllers/operationcontrollers/operation_request_credential.go
mbarnes: Fair.  I can add those comments for all the operation controllers in a followup if that's ok.

backend/pkg/controllers/operationcontrollers/operation_revoke_credentials_test.go
mbarnes: Addressing this comment and the ones above, this is following a pattern I've seen in other controller unit tests that I think @deads2k probably wrote.

I'm torn on whether to simplify these unit tests down to just what's needed for the particular test cases I've added.  The cases here are not exhaustive; they just exercise the non-error-handling paths to prove basic functionality.

There's something to be said for having a consistent unit test pattern across all the controllers - even if it's overkill for simpler controllers - such that once you learn the pattern it's relatively easy to add new cases to any controller.  The `verify` callback in particular allows for case-specific logic in close proximity to the test case itself, and it keeps the main test loop simple.  I find it more intuitive despite the repetitiveness

I'm slightly inclined to keep it as is but I could be talked into simplifying if there's benefits I've overlooked.

backend/pkg/controllers/operationcontrollers/operation_request_credential.go
tmstff: sure :-)

backend/pkg/controllers/operationcontrollers/operation_revoke_credentials_test.go
tmstff: I already imagined something like this, and was myself debating if it is good advise to do this simplification :-D I'd say let's keep it like that, especially if you already expect more test cases to follow (which might then require the current freedom) :-)

PR #5037, opened 2026-04-27
pers-dev: fix occasional override config path issues

geoberle:
### What

* consistent timestamp for override image file
* adapt image build target of sessiongate to match others

### DIFF
--- a/admin/Makefile
+++ b/admin/Makefile
@@ -78,7 +78,9 @@ build-and-push: image $(ORAS)
 	$(CONTAINER_ENGINE) push ${ADMIN_API_TAGGED_IMAGE}
 .PHONY: build-and-push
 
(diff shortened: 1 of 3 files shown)


tuxerrante (commented):
Nice catch on the `$(date +%s)` → `$(shell date +%s)` fix — that's a real bug, not just a style issue (Make was interpreting it as a variable reference, so the timestamp was empty). Also glad to see the `amin-api` typo fixed.

Minor note: the `touch $@` removal from sessiongate's `image` target means it'll always rebuild on `make image` even if nothing changed. Looks intentional to match admin/frontend — just confirming that's the intent.

geoberle:
> Looks intentional

yes it was causing troubles. i'm working on an improvement to address speed

PR #5039, opened 2026-04-27
Fix E2E cleanup: Force-delete managed RGs stuck in Deleting state

ziccardi:
TEMPORARY MITIGATION for resource groups getting permanently stuck in Azure "Deleting" state, causing 100% failure rate in cleanup jobs.

When cleanupResourceGroup() times out waiting for managed resource groups to be deleted (after 10 minutes), it now attempts to force-delete them instead of returning an error immediately. This prevents RG accumulation and unblocks the periodic cleanup jobs while we investigate the root cause.

Uses the same force-deletion pattern as cleanupResourceGroupNoRP() with Azure ForceDeletionTypes for VMs/VMSSs.

This is a workaround that treats the symptom. The underlying issue of why Azure fails to complete RG deletion within 10 minutes needs separate investigation (blocked resources, permission issues, Azure API problems, etc).

[AROSLSRE-642](https://redhat.atlassian.net/browse/AROSLSRE-642)

### DIFF
--- a/test/util/framework/per_test_framework.go
+++ b/test/util/framework/per_test_framework.go
@@ -507,9 +507,24 @@ func (tc *perItOrDescribeTestContext) cleanupResourceGroup(ctx context.Context, 
 		managedResourceGroups, err = tc.waitForManagedResourceGroupsDeletion(ctx, resourceGroupName, 10*time.Minute)
 		if err != nil {
 			if len(managedResourceGroups) > 0 {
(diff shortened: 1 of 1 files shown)


bennerv (commented):
The deny assignment is going to prevent all these managed resource group deletion calls from failing.

stevekuznetsov:
Shouldn't this be a fix in the RP backend?

ziccardi:
@bennerv @stevekuznetsov  You are both right. Closing the PR. Thank you!

PR #5040, opened 2026-04-27
Add unit tests for Azure client policies in test framework

mvacula02:
[ARO-26228](https://redhat.atlassian.net/browse/ARO-26228)

### What

Adds unit tests for each of the ARM-simulating client policies in test/util/framework/azure_client_policies.go. 

- armResourceGroupValidationPolicy - checks for existence of RG, 3 cases 
- armSystemDataPolicy - system data header injection for frontend-bound requests, 2 cases
- correlationRequestIDPolicy - correlation ID generation when header is absent, verify existing header is not overwritten, 3 cases
- sanitizeAuthHeaderPolicy - Authorization header redaction on responses, 3 cases
- lroPollerRetryDeploymentNotFoundPolicy - retry logic for transient 404 DeploymentNotFound during LRO polling, 7 cases

Adds unit tests for select helper functions, whose full functionality is not implicitly tested by the client policy unit tests.
- parseResourceGroupFromPath - helper function, path parsing with case variations, 12 cases
- ~~parseRetryAfter - retry after header parsing, 9 cases~~ removed as the helper function itself was removed in main

### Why

Unit tests were missing for Azure client policies, some of which simulate functionality necessary for execution of E2E test cases in the dev environment (e.g. negative cluster listing is possible only due to armResourceGroupValidationPolicy and its helper), others handling a role in all environments (e.g. auth header redaction). 

### Testing

The PR adds unit tests for pre-existing Azure client policies. Tests will run automatically in `test-unit` pre-submit Prow job.

### DIFF
--- a/test/util/framework/azure_client_policies_test.go
+++ b/test/util/framework/azure_client_policies_test.go
@@ -61,99 +61,6 @@ func okResponse() (*http.Response, error) {
 	}, nil
 }
 
-func TestParseResourceGroupFromPath(t *testing.T) {
-	t.Parallel()
-
-	tests := []struct {
-		name       string
-		path       string
-		wantSubID  string
-		wantRGName string
-	}{
-		{
-			name:       "standard ARM path",
-			path:       "/subscriptions/sub-id/resourceGroups/rg-name/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/my-cluster",
-			wantSubID:  "sub-id",
-			wantRGName: "rg-name",
-		},
-		{
-			name:       "uppercase segments",
-			path:       "/SUBSCRIPTIONS/sub-id/RESOURCEGROUPS/rg-name/providers/foo",
-			wantSubID:  "sub-id",
-			wantRGName: "rg-name",
-		},
-		{
-			name:       "mixed case segments",
-			path:       "/Subscriptions/sub-id/ResourceGroups/rg-name",
-			wantSubID:  "sub-id",
-			wantRGName: "rg-name",
-		},
-		{
-			name:       "no resourceGroups segment",
-			path:       "/subscriptions/sub-id/providers/Microsoft.RedHatOpenShift",
-			wantSubID:  "sub-id",
-			wantRGName: "",
-		},
-		{
-			name:       "no subscriptions segment",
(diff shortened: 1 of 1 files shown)


test/util/framework/azure_client_policies_test.go
```diff
@@ -0,0 +1,610 @@
...
+			t.Parallel()
+			subID, rgName := parseResourceGroupFromPath(tt.path)
+			assert.Equal(t, tt.wantSubID, subID)
+			assert.Equal(t, tt.wantRGName, rgName)
+		})
+	}
+}
+
+func TestArmSystemDataPolicy(t *testing.T) {
+	const frontendHost = "my-frontend.example.com:8443"
```
mgahagan73: Since you are defining/using frontendHost with the same value 3 times, just make this one a file-level constant by declaring it after the imports are done.

test/util/framework/azure_client_policies_test.go
```diff
@@ -0,0 +1,610 @@
...
+
+		resp, err := pipeline.Do(req)
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+		assert.Empty(t, capturedHeaders.Get(arm.HeaderNameARMResourceSystemData))
+		assert.Empty(t, capturedHeaders.Get(arm.HeaderNameIdentityURL))
+	})
+}
+
+func TestArmResourceGroupValidationPolicy(t *testing.T) {
```
mgahagan73, line 210: Since we only change part of the URL here you can probably make this function shorter by converting to a table driven test instead, would be nice to have but not anything major. I think it would age better if we plan on adding additional cases in the future. Either way I would not block approval on it.

mbukatov:
Would it make sense to refactor the `armResourceGroupValidationPolicy` to accept ARM resource group client from the outside, so that we can mock it in the test and test a negative case as well?

mbukatov:
Github PR UI is somewhat broken :sadcat:

aklymenk:
sorry for accidentally running prod e2e job :see_no_evil:

mbukatov:
Let's do the suggested refactor of `armResourceGroupValidationPolicy` as a follow up.

/lgtm

PR #5041, opened 2026-04-27
Skip missing clusters during cleanup.  Clusters that don't exist don't need maestro bundles.

deads2k:
Found by inspecting Backend Metrics for https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/5020/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2048471908730540032 This failure is happening during cleanup of delete orphaned maestro readonly bundles.

### DIFF
--- a/backend/pkg/controllers/delete_orphaned_maestro_readonly_bundles_controller.go
+++ b/backend/pkg/controllers/delete_orphaned_maestro_readonly_bundles_controller.go
@@ -232,6 +232,9 @@ func (c *deleteOrphanedMaestroReadonlyBundles) mapServiceProviderClustersByProvi
 			// resource with a maestro bundle reference. If for some reason during the orphan calculation inbetween the first read of cosmos
 			// resources and the read in the Maestro api there's a new bundle in maestro, the second read of cosmos resources will catch that
 			// and prevent accidental deletion.
+			// We should also be able to skip the ServiceProviderCluster if the cluster associated with it does not exist anymore
+			// between the time the sync iteration started and this point. In that case it's correct that we eliminate the maestro readonly bundle
+			// in the API because the cluster is no longer there.
 			continue
 		}
 		if _, ok := maestroClientsByShard[shardID]; !ok {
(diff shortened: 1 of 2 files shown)


backend/pkg/controllers/delete_orphaned_maestro_readonly_bundles_controller.go
```diff
@@ -451,6 +451,10 @@ func (c *deleteOrphanedMaestroReadonlyBundles) clusterProvisionShardIDForService
 		return "", false, utils.TrackError(fmt.Errorf("ServiceProviderCluster %s has no parent resource ID", spc.ResourceID.String()))
 	}
 	cluster, err := c.cosmosClient.HCPClusters(clusterResourceID.SubscriptionID, clusterResourceID.ResourceGroupName).Get(ctx, clusterResourceID.Name)
+	if database.IsNotFoundError(err) {
```
miguelsorianod, line 460: I think we need the same check for clusterProvisionShardIDForServiceProviderNodePool

miguelsorianod, line 460: Let's also update the documentation of the clusterProvisionShardIDForServiceProviderCluster method on what cases return skip, as well as for clusterProvisionShardIDForServiceProviderNodePool (related to https://github.com/Azure/ARO-HCP/pull/5041/changes#r3148593634).

And also the comment in the skip check in both mapServiceProviderClustersByProvisionShard and mapServiceProviderNodePoolsByProvisionShard and their corresponding docs, to account for the new case.

machi1990 (commented):
/lgtm

backend/pkg/controllers/delete_orphaned_maestro_readonly_bundles_controller.go
miguelsorianod: I added the check

miguelsorianod: updated

PR #5042, opened 2026-04-27
Mark the cluster X.Y update operation as completed once the desired version is within the target minor and cluster service reports the cluster as ready

machi1990:
### What

Mark the cluster X.Y update operation as completed once the desired version is within the target minor

### Why

Before this patch, wt the moment we mark the version update operation as completed almost instantly after the operation is created we change that with the following proposal:

- We mark the operation as succeeded when the desired version is reflected in the ServiceProviderCluster 
i.e for cluster it'll be that the desired x.y+1.z version is in the same desired minor

The benefit of this approach are:
- Marking the operation as failed if there is no version available for the chosen X.Y release line or there are no upgrade path from the current version; so the know instantly that the upgrade can't be served. This is also particularly useful so that the user gets a near instant feedback
- Operation wise, we'll closely move the marking of operation succeeded to reflect that the Service is able to honor the desired and it going to trigger the upgrade operation

1. Related to https://redhat.atlassian.net/browse/ARO-26048.
For node pool this is going to be addressed in a separate effort https://redhat.atlassian.net/browse/ARO-26304
2. We'll still check for cluster readiness before marking the overall operation as succeeded

### DIFF
--- a/backend/pkg/controllers/controllerutils/util.go
+++ b/backend/pkg/controllers/controllerutils/util.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"runtime/debug"
-	"slices"
 
 	"github.com/go-logr/logr"
 
@@ -32,6 +31,7 @@ import (
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/database"
 	"github.com/Azure/ARO-HCP/internal/utils"
+	"github.com/Azure/ARO-HCP/internal/utils/armhelpers"
 )
 
 type Controller interface {
@@ -219,9 +219,11 @@ func ReportSyncError(syncErr error) controllerMutationFunc {
 	}
 }
 
-type initialControllerFunc func(controllerName string) *api.Controller
+// InitialControllerFunc builds a new api.Controller for the given logical controller name
+// (for example HCPClusterKey.InitialController).
+type InitialControllerFunc func(controllerName string) *api.Controller
 
-func DegradedControllerPanicHandler(ctx context.Context, controllerCRUD database.ResourceCRUD[api.Controller], controllerName string, initialControllerFn initialControllerFunc) func(interface{}) {
+func DegradedControllerPanicHandler(ctx context.Context, controllerCRUD database.ResourceCRUD[api.Controller], controllerName string, initialControllerFn InitialControllerFunc) func(interface{}) {
 	return func(panicVal interface{}) {
 		stack := debug.Stack()
 		err := WriteController(ctx, controllerCRUD, controllerName, initialControllerFn, ReportSyncError(fmt.Errorf("panic caught:\n%v\n\n%s", panicVal, stack)))
@@ -232,42 +234,118 @@ func DegradedControllerPanicHandler(ctx context.Context, controllerCRUD database
 	}
 }
 
+func controllerCRUDForParent(dbClient database.DBClient, parentResourceID *azcorearm.ResourceID) (database.ResourceCRUD[api.Controller], error) {
+	subscriptionID := parentResourceID.SubscriptionID
+	resourceGroupName := parentResourceID.ResourceGroupName
+	hcp := dbClient.HCPClusters(subscriptionID, resourceGroupName)
+
+	switch {
+	case armhelpers.ResourceTypeEqual(parentResourceID.ResourceType, api.ClusterResourceType):
+		return hcp.Controllers(parentResourceID.Name), nil
+	case armhelpers.ResourceTypeEqual(parentResourceID.ResourceType, api.NodePoolResourceType):
+		if parentResourceID.Parent == nil {
+			return nil, fmt.Errorf("node pool resource ID is missing parent cluster ID")
+		}
+		clusterName := parentResourceID.Parent.Name
+		return hcp.NodePools(clusterName).Controllers(parentResourceID.Name), nil
+	case armhelpers.ResourceTypeEqual(parentResourceID.ResourceType, api.ExternalAuthResourceType):
+		if parentResourceID.Parent == nil {
+			return nil, fmt.Errorf("external auth resource ID is missing parent cluster ID")
+		}
+		clusterName := parentResourceID.Parent.Name
+		return hcp.ExternalAuth(clusterName).Controllers(parentResourceID.Name), nil
+	default:
+		return nil, fmt.Errorf("unsupported parent resource type for controllers: %s", parentResourceID.ResourceType)
+	}
+}
+
+// getOrCreateControllerDocument returns the controller document from Cosmos, creating it with
+// initialControllerFn if missing. On create conflict (HTTP 409), it re-reads and returns the
+// existing document (same pattern as database.GetOrCreateServiceProviderCluster).
+func getOrCreateControllerDocument(
+	ctx context.Context,
+	controllerCRUD database.ResourceCRUD[api.Controller],
+	controllerName string,
+	initialControllerFn InitialControllerFunc,
+) (*api.Controller, error) {
+	if initialControllerFn == nil {
+		return nil, fmt.Errorf("initialControllerFn is required")
+	}
+
+	existingController, err := controllerCRUD.Get(ctx, controllerName)
+	if err == nil && existingController != nil {
+		return existingController, nil
+	}
+	if err == nil {
+		err = database.NewNotFoundError()
+	}
+
+	if !database.IsNotFoundError(err) {
+		return nil, fmt.Errorf("failed to get existing controller state: %w", err)
+	}
+
+	existingController, err = controllerCRUD.Create(ctx, initialControllerFn(controllerName), nil)
+	if err == nil {
+		if existingController == nil {
+			return nil, fmt.Errorf("create returned success with nil controller")
+		}
+		return existingController, nil
+	}
+
+	if !database.IsConflictError(err) {
+		return nil, fmt.Errorf("failed to create existing controller state: %w", err)
+	}
+
+	existingController, err = controllerCRUD.Get(ctx, controllerName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get existing controller state after create conflict: %w", err)
+	}
+	if existingController == nil {
+		return nil, fmt.Errorf("failed to get existing controller state after create conflict: document missing")
+	}
+
+	return existingController, nil
+}
+
+// GetOrCreateController gets the named Controller document under the given parent resource
+// (cluster, node pool, or external auth). If it does not exist, it creates one using initialControllerFn.
+// On create conflict (HTTP 409), it re-reads and returns the existing document (same pattern as
+// database.GetOrCreateServiceProviderCluster).
+func GetOrCreateController(
+	ctx context.Context, dbClient database.DBClient, parentResourceID *azcorearm.ResourceID,
+	controllerName string, initialControllerFn InitialControllerFunc,
+) (*api.Controller, error) {
+	controllerCRUD, err := controllerCRUDForParent(dbClient, parentResourceID)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+
+	doc, err := getOrCreateControllerDocument(ctx, controllerCRUD, controllerName, initialControllerFn)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+	return doc, nil
+}
+
 // WriteController will read the existing value, call the mutations in order, then write the result.  It only tries *once*.
 // If it fails, then the an error is returned.  This detail is important, it doesn't even retry conflicts.  This is so that
 // if a failure happens the control-loop will re-run and restablish the information it was trying to write as valid.
 // This prevents accidental recreation of controller instances in cosmos during a delete.
-func WriteController(ctx context.Context, controllerCRUD database.ResourceCRUD[api.Controller], controllerName string, initialControllerFn initialControllerFunc, mutationFns ...controllerMutationFunc) error {
+func WriteController(ctx context.Context, controllerCRUD database.ResourceCRUD[api.Controller], controllerName string, initialControllerFn InitialControllerFunc, mutationFns ...controllerMutationFunc) error {
 	logger := utils.LoggerFromContext(ctx)
 
-	existingController, err := controllerCRUD.Get(ctx, controllerName)
-	if err != nil && !database.IsNotFoundError(err) {
-		return fmt.Errorf("failed to get existing controller state: %w", err)
+	existingController, err := getOrCreateControllerDocument(ctx, controllerCRUD, controllerName, initialControllerFn)
+	if err != nil {
+		return err
 	}
 
-	var desiredController *api.Controller
-	if existingController == nil { // fill in for conveniently avoiding NPEs
-		desiredController = initialControllerFn(controllerName)
-	} else {
-		// TODO we'd prefer a DeepCopy, but we don't have one yet.
-		temp := *existingController
-		desiredController = &temp
-		desiredController.Status.Conditions = slices.Clone(existingController.Status.Conditions)
-	}
+	desiredController := existingController.DeepCopy()
 	for _, mutationFn := range mutationFns {
 		mutationFn(desiredController)
 	}
 
 	if equality.Semantic.DeepEqual(existingController, desiredController) {
 		// nothing to report.
-		return nil
-	}
-
-	if existingController == nil {
-		_, createErr := controllerCRUD.Create(ctx, desiredController, nil)
-		if createErr != nil {
-			logger.Error(createErr, "failed to create")
-			return fmt.Errorf("failed to create existing controller state: %w", createErr)
-		}
 		return nil
 	}
 
--- a/backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
@@ -28,6 +28,8 @@ import (
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/utils/clock"
+	"k8s.io/utils/lru"
 
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
 	"github.com/Azure/ARO-HCP/internal/api"
@@ -38,9 +40,11 @@ import (
 )
 
 type operationClusterUpdate struct {
-	cosmosClient         database.DBClient
-	clusterServiceClient ocm.ClusterServiceClientSpec
-	notificationClient   *http.Client
+	cosmosClient                    database.DBClient
+	clusterServiceClient            ocm.ClusterServiceClientSpec
+	notificationClient              *http.Client
+	clock                           clock.PassiveClock
+	desiredVersionMismatchFirstSeen *lru.Cache
 }
 
 // NewOperationClusterUpdateController periodically lists all clusters and for each out when the cluster was created and its state.
@@ -51,9 +55,11 @@ func NewOperationClusterUpdateController(
 	activeOperationInformer cache.SharedIndexInformer,
 ) controllerutils.Controller {
 	syncer := &operationClusterUpdate{
-		cosmosClient:         cosmosClient,
-		clusterServiceClient: clusterServiceClient,
-		notificationClient:   notificationClient,
+		cosmosClient:                    cosmosClient,
+		clusterServiceClient:            clusterServiceClient,
+		notificationClient:              notificationClient,
+		clock:                           clock.RealClock{},
+		desiredVersionMismatchFirstSeen: lru.New(100000),
 	}
 
 	controller := NewGenericOperationController(
@@ -176,6 +182,7 @@ func (c *operationClusterUpdate) desiredVersionResolutionOperationState(ctx cont
 
 	if customerDesiredVersion.Major == resultingDesiredVersion.Major &&
 		customerDesiredVersion.Minor == resultingDesiredVersion.Minor {
+		c.desiredVersionMismatchFirstSeen.Remove(operation.ResourceID.String())
 		return newOperationState(arm.ProvisioningStateSucceeded, ""), nil
 	}
 	clusterKey := controllerutils.HCPClusterKey{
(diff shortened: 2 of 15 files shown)


backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,6 +87,44 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
+	if existingServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion == nil {
+		return nil
+	}
+
+	customerDesiredVersion, err := semver.ParseTolerant(existingCluster.CustomerProperties.Version.ID)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	// If the customer desired major.minor version does not match the resolved desired major.minor version,
```
machi1990: See and apply the pattern from https://github.com/Azure/ARO-HCP/pull/4695

machi1990: done

machi1990: I'll need to rebase once https://github.com/Azure/ARO-HCP/pull/4695 merges

machi1990: rebased

internal/api/service_provider_cluster_conditions.go
```diff
@@ -0,0 +1,30 @@
...
+
+package api
+
+// ServiceProviderClusterConditionType identifies a ServiceProviderCluster status condition Type value
+// (see ServiceProviderClusterStatus.Conditions).
+type ServiceProviderClusterConditionType string
+
+const (
+	// ServiceProviderClusterConditionTypeDegraded is True when the cluster is in a degraded state.
+	ServiceProviderClusterConditionTypeDegraded ServiceProviderClusterConditionType = "Degraded"
```
deads2k: we're switching to metav1, so just a string.

Also, don't be fancy, this is just Degraded

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,20 +100,107 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
 	}
+	logger.Info("picked cluster update operation status", "provisioningState", picked.provisioningState, "message", picked.message)
+	return picked, nil
+}
 
-	return nil
+func (c *operationClusterUpdate) serviceProviderClusterUpdateOperationState(ctx context.Context, operation *api.Operation) (*operationState, error) {
+	existingCluster, err := c.cosmosClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).Get(ctx, operation.ExternalID.Name)
+	if err != nil {
+		return nil, err
```
deads2k: `utils.TrackError(err)` throughout

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,20 +100,107 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
+	}
+	existingServiceProviderCluster, err := c.cosmosClient.ServiceProviderClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName, operation.ExternalID.Name).Get(ctx, api.ServiceProviderClusterResourceName)
+	if err != nil {
+		return nil, err
+	}
+	desired := existingServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion
```
deads2k: we need a better name: `resultingDesiredVersion`?  We'll eventually have an `sreDesiredVersion` I suspect.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,20 +100,107 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
 	if err != nil {
-		return utils.TrackError(err)
+		return nil, utils.TrackError(err)
 	}
+	logger.Info("picked cluster update operation status", "provisioningState", picked.provisioningState, "message", picked.message)
+	return picked, nil
+}
 
-	return nil
+func (c *operationClusterUpdate) serviceProviderClusterUpdateOperationState(ctx context.Context, operation *api.Operation) (*operationState, error) {
```
deads2k: name this better.  Is this specific for version updates?

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,20 +100,107 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
+	}
+
+	if customerDesiredVersion.Major == desired.Major &&
+		customerDesiredVersion.Minor == desired.Minor {
+		return newOperationState(arm.ProvisioningStateSucceeded, ""), nil
+	}
+	existingDegraded := apimeta.FindStatusCondition(existingServiceProviderCluster.Status.Conditions,
+		string(api.ServiceProviderClusterConditionTypeDegraded))
+	if existingDegraded != nil && existingDegraded.Status == metav1.ConditionTrue &&
+		existingDegraded.Reason == api.ServiceProviderClusterConditionReasonVersionUpgradeNotAccepted {
```
deads2k: This won't age well because it only allows one to "win".  I like the idea though.  Find a way to narrow it down to a particular controller. Perhaps having an `IntentFailed` condition on controllers that can set the value the controller hasn't failed and there's no reason for a rapid retry, but we want to indicate a status here?  If we did that, we would

1. have natural multiplicity with every set of fields we write since it is per controller
2. avoid retry logic on the degraded (I could do my work) flow
3. re-evaluate the controller sync every time the input changed
4. use this technique when the result is "sticky"

Once we have that, we'll need to build logic around hysteresis for reasons and status, but we wouldn't have to start there.

deads2k (changes requested):
good idea.  Some API comments.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
machi1990: Yes. It is specific for upgrade. 
I'll go with `desiredVersionResolutionOperationState()`

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
machi1990: I've added this in the newest commit.

What hasn't been added is:
- an overall controller that reads all the controller docs for a given cluster and set the ServiceProviderCluster degraded condition to True if IntentFailed=True for at least one of the controller.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,20 +100,120 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+	existingServiceProviderCluster, err := c.cosmosClient.ServiceProviderClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName, operation.ExternalID.Name).Get(ctx, api.ServiceProviderClusterResourceName)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+	resultingDesiredVersion := existingServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion
+	if resultingDesiredVersion == nil {
+		return nil, utils.TrackError(fmt.Errorf("service provider cluster has no desired version"))
```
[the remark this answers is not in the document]
machi1990: This is desired to trigger a fast retry

backend/pkg/controllers/controllerutils/util.go
```diff
@@ -232,27 +234,113 @@ func DegradedControllerPanicHandler(ctx context.Context, controllerCRUD database
...
 // This prevents accidental recreation of controller instances in cosmos during a delete.
-func WriteController(ctx context.Context, controllerCRUD database.ResourceCRUD[api.Controller], controllerName string, initialControllerFn initialControllerFunc, mutationFns ...controllerMutationFunc) error {
+func WriteController(ctx context.Context, controllerCRUD database.ResourceCRUD[api.Controller], controllerName string, initialControllerFn InitialControllerFunc, mutationFns ...controllerMutationFunc) error {
 	logger := utils.LoggerFromContext(ctx)
 
-	existingController, err := controllerCRUD.Get(ctx, controllerName)
-	if err != nil && !database.IsNotFoundError(err) {
-		return fmt.Errorf("failed to get existing controller state: %w", err)
+	existingController, err := getOrCreateControllerDocument(ctx, controllerCRUD, controllerName, initialControllerFn)
+	if err != nil {
```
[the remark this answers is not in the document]
machi1990: The comment is still valid.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,20 +100,120 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
+	if operationState, err := c.desiredVersionResolutionOperationState(ctx, operation); err != nil {
+		errs = append(errs, utils.TrackError(err))
+	} else {
+		operationStates = append(operationStates, operationState)
+	}
+	if operationState, csErr := c.clusterServiceUpdateOperationState(ctx, operation); csErr != nil {
+		errs = append(errs, utils.TrackError(csErr))
+	} else {
+		operationStates = append(operationStates, operationState)
+	}
```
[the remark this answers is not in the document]
machi1990: This is desired.
If we want to return the worst and in this case, we want to return updating.

internal/api/service_provider_cluster_conditions.go
```diff
@@ -0,0 +1,26 @@
...
+const (
+	// DegradedCondition is True when the cluster is in a degraded state.
+	DegradedCondition = "Degraded"
+)
+
+// Known ServiceProviderCluster status condition Reason values (metav1.Condition.Reason).
+const (
+	VersionUpgradeNotAcceptedReason string = "VersionUpgradeNotAccepted"
+	NoErrorsReason                  string = "NoErrors"
+)
```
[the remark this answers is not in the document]
machi1990: For now, this is used for cluster, so I am keeping it as is. Once we do the same for node pool, appropriate one can be used

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,20 +100,120 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
 
-	err = UpdateOperationStatus(ctx, c.cosmosClient, operation, newOperationStatus, opError, postAsyncNotificationFn(c.notificationClient))
+	if err := errors.Join(errs...); err != nil {
+		return nil, err
+	}
+	if len(operationStates) == 0 {
+		return nil, utils.TrackError(fmt.Errorf("no operation states"))
```
ahitacat: Why we change to track errors? For create that is not done.

backend/pkg/controllers/operationcontrollers/operation_state.go
```diff
@@ -0,0 +1,88 @@
...
+		return -1
+	}
+	if provisioningStatePriority[lhs.provisioningState] > provisioningStatePriority[rhs.provisioningState] {
+		return 1
+	}
+	return strings.Compare(lhs.message, rhs.message)
+}
+
+// pickWorstOperationState expects states pre-sorted and returns the worst state with merged messages.
+func pickWorstOperationState(states []*operationState) (*operationState, error) {
```
ahitacat, line 72: Can we have some unit tests for this?

backend/pkg/controllers/controllerutils/util.go
```diff
@@ -262,15 +349,6 @@ func WriteController(ctx context.Context, controllerCRUD database.ResourceCRUD[a
...
-		_, createErr := controllerCRUD.Create(ctx, desiredController, nil)
-		if createErr != nil {
-			logger.Error(createErr, "failed to create")
-			return fmt.Errorf("failed to create existing controller state: %w", createErr)
-		}
-		return nil
-	}
-
 	_, replaceErr := controllerCRUD.Replace(ctx, desiredController, nil)
 	if replaceErr != nil {
```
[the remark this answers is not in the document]
deads2k: For ease of understanding, don't do what this comment wants.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,20 +100,120 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
 
-	newOperationStatus, opError, err := convertClusterStatus(ctx, c.clusterServiceClient, operation, clusterStatus)
-	if err != nil {
+	var persistErr *arm.CloudErrorBody
+	if operationalState.provisioningState == arm.ProvisioningStateFailed {
+		persistErr = &arm.CloudErrorBody{
+			Code:    arm.CloudErrorCodeInvalidRequestContent,
+			Message: operationalState.message,
+		}
+	}
```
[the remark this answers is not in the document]
deads2k: Yeah, don't do this one either. Too much change from current

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -93,20 +100,120 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
+		operation.ExternalID,
+		"ControlPlaneDesiredVersion",
+		clusterKey.InitialController,
+	)
+	if getControllerErr != nil {
+		return nil, utils.TrackError(getControllerErr)
+	}
+	intentFailedCondition := apimeta.FindStatusCondition(controllerDoc.Status.Conditions, api.ControllerConditionTypeIntentFailed)
+	if intentFailedCondition == nil || intentFailedCondition.Status != metav1.ConditionTrue || intentFailedCondition.Reason != api.VersionUpgradeNotAcceptedReason {
+		return nil, utils.TrackError(fmt.Errorf("customer desired version does not match resolved desired version"))
```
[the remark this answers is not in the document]
deads2k: This one is good though. When we get here we should return the state.  Add a layer that can convert non-success into operation failure after it fails for 29s

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -150,30 +154,63 @@ func (c *controlPlaneDesiredVersionSyncer) SyncOnce(ctx context.Context, key con
...
-		return nil
+	if err != nil {
+		// Persist IntentFailed on the controller document for Cincinnati VersionNotFound or any non-Cincinnati resolution error.
+		// Other Cincinnati errors are treated as transient graph or transport issues.
+		var cincinnatiErr *cincinnati.Error
+		persistIntentFailed := cincinatti.IsCincinnatiVersionNotFoundError(err) || !errors.As(err, &cincinnatiErr)
+		if persistIntentFailed {
+			logger.Error(err, "desired version resolution failed, persisting IntentFailed condition")
+			controllerCRUD := c.cosmosClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Controllers(key.HCPClusterName)
+			if writeErr := controllerutils.WriteController(ctx, controllerCRUD, controlPlaneDesiredVersionControllerName, key.InitialController,
```
deads2k, line 167: I'll accept it, but are we better off with a factorization that does this at the tail?  If you remind me on Thursday, I can show a possible refactor later

internal/utils/line_tracking_error_test.go
```diff
@@ -133,3 +135,23 @@ func TestLineTrackingError_MultipleWrapping(t *testing.T) {
...
+	}{
+		{name: "nil", err: nil, want: ""},
+		{name: "unwraps single TrackError", err: TrackError(errors.New("root cause")), want: "root cause"},
+		{name: "unwraps nested TrackError", err: TrackError(TrackError(errors.New("base"))), want: "base"},
+		{name: "plain error unchanged", err: errors.New("plain"), want: "plain"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			tt := tt
```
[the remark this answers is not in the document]
deads2k: Should be fixed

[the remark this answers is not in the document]
machi1990: Good catch. This slipped in

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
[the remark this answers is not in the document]
machi1990: Yes, make sense.

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
machi1990: I'll like the idea. 
I'll ping on Thursday for reminder

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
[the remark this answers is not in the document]
machi1990: Done.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -191,13 +199,40 @@ func (c *operationClusterUpdate) desiredVersionResolutionOperationState(ctx cont
...
 	intentFailedCondition := apimeta.FindStatusCondition(controllerDoc.Status.Conditions, api.ControllerConditionTypeIntentFailed)
 	if intentFailedCondition == nil || intentFailedCondition.Status != metav1.ConditionTrue || intentFailedCondition.Reason != api.VersionUpgradeNotAcceptedReason {
-		return nil, utils.TrackError(fmt.Errorf("customer desired version does not match resolved desired version"))
+		acceptedState := newOperationState(arm.ProvisioningStateAccepted, "customer desired version does not match resolved desired version")
+		// Customer desired minor differs from the service provider resolved version, and the
+		// ControlPlaneDesiredVersion controller has not yet set IntentFailed (VersionUpgradeNotAccepted).
+		// Stay Accepted while resolution runs; the returned predicate fails the operation after
+		// 29s from operation StartTime if the mismatch is still unresolved.
+		return acceptedState, func(operation *api.Operation) (bool, string) {
+			elapsed := c.clock.Now().Sub(operation.StartTime)
```
deads2k: this time needs to be, "time since this running binary first observed this item in the loop".  Otherwise a crash and restart here will mass update every operation to failed.

Also, I think the logic here wants to be "accepted right now, but failed if I see it again after 29s".  I suspect the simplest factorization is an LRU of operations with a first observed timestamp.  Then this code here doesn't grow a predicate, but instead just checks right here: "have I see this before and was i tmore than 29s, if so, return fail, if not return accepted"

machi1990: Thank you. The initial version was raw. Thank you for catching that and providing useful suggestion.
I've incorporated them and commit-amended the earlier commit.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -16,24 +16,35 @@ package operationcontrollers
...
 type operationClusterUpdate struct {
-	cosmosClient         database.DBClient
-	clusterServiceClient ocm.ClusterServiceClientSpec
-	notificationClient   *http.Client
+	cosmosClient                    database.DBClient
+	clusterServiceClient            ocm.ClusterServiceClientSpec
+	notificationClient              *http.Client
+	clock                           clock.PassiveClock
+	desiredVersionMismatchFirstSeen *lru.Cache
 }
```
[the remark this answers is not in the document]
deads2k: Likely correct comment

backend/pkg/controllers/operationcontrollers/operation_cluster_update_test.go
```diff
@@ -33,93 +43,195 @@ import (
...
+
+				cluster, err := db.HCPClusters(testSubscriptionID, testResourceGroupName).Get(ctx, testClusterName)
+				require.NoError(t, err)
+				assert.Equal(t, arm.ProvisioningStateFailed, cluster.ServiceProviderProperties.ProvisioningState)
+				assert.Empty(t, cluster.ServiceProviderProperties.ActiveOperationID)
+			},
+		},
 	}
 
 	for _, tt := range tests {
```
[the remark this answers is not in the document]
deads2k: Mind the races. Keep it simple

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -150,30 +154,63 @@ func (c *controlPlaneDesiredVersionSyncer) SyncOnce(ctx context.Context, key con
...
+	logger := utils.LoggerFromContext(ctx)
 
-	// Check if there's a new desired version to set
-	if desiredVersion == nil {
-		return nil
+	if err != nil {
+		// Persist IntentFailed on the controller document for Cincinnati VersionNotFound or any non-Cincinnati resolution error.
+		// Other Cincinnati errors are treated as transient graph or transport issues.
+		var cincinnatiErr *cincinnati.Error
+		persistIntentFailed := cincinatti.IsCincinnatiVersionNotFoundError(err) || !errors.As(err, &cincinnatiErr)
```
tuxerrante, line 163: Kind suggestion: this predicate now persists `IntentFailed` for every non-Cincinnati error, which seems broader than "the requested version cannot be honored". Some of the helpers called from `desiredControlPlaneZVersion()` can also fail because Cosmos/list/get failed, and those would then be surfaced back through the update operation as a user-facing version rejection instead of retrying as a transient backend failure. Would it make sense to narrow this to explicitly user-correctable cases (for example `VersionNotFound`, downgrade/minor-skew validation failures, etc.) and let internal read/validation failures return an error?

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller_test.go
```diff
@@ -892,3 +900,198 @@ func assertVersionResult(t *testing.T, result *semver.Version, err error, expect
...
+			wantDesiredVersion:  nil,
+			wantIntentFailed: &metav1.Condition{
+				Type:    api.ControllerConditionTypeIntentFailed,
+				Status:  metav1.ConditionTrue,
+				Reason:  api.VersionUpgradeNotAcceptedReason,
+				Message: "invalid next y-stream upgrade path from 4.20.0 to 4.19.0: only upgrades to the next minor version are allowed, no downgrades",
+			},
+		},
+		{
+			name:                "Cincinnati upstream error does not persist IntentFailed or desired version",
```
tuxerrante, line 1000: Kind suggestion: would it be worth adding a case for a non-Cincinnati internal failure here as well (for example a DB/list failure while checking node-pool skew)? That seems like the most important classification edge for this controller now, because it decides whether we retry or persist `IntentFailed` and later surface a user-facing `VersionUpgradeNotAccepted` failure.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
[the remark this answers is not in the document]
machi1990: The cache is a thread safe https://github.com/kubernetes/utils/blob/master/lru/lru.go#L28

Am I missing anything?

backend/pkg/controllers/operationcontrollers/operation_cluster_update_test.go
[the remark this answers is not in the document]
machi1990: I don't understand this bot comment; the loop-variable capture race has been fixed long ago https://go.dev/blog/loopvar-preview in go1.22. And we are in go1.25.

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
machi1990: This is a fair point. 
I did it this way since I didn't want to go down the refactoring route. I'll evaluate how far that is and see what can be done now vs deferring that and require a user retrying in the meanwhile if we fail to read from DB

PR #5043, opened 2026-04-27
Add datadump to git information for subscriptions and cluster-service content

deads2k:
This makes debugging slightly easier when trying to relate fields about cluster-service.

Found while digging around into why metrics are doing  badly.


stevekuznetsov (commented):
/lgtm
/approve

PR #5045, opened 2026-04-27
Update cincinatti interaction to use maestro for clusterID

deads2k:
This disconnects cluster-service from another piece of our controllers.
    
We should consider a backup flow that uses an value when missing and see    if we can supply the value to cluster-service and generate it ourselves.

Addresses the first cause of the nodepoolversion controller retry in https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/5020/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2048471908730540032

### DIFF
--- a/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
+++ b/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
@@ -36,6 +36,7 @@ import (
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
 	"github.com/Azure/ARO-HCP/backend/pkg/informers"
 	"github.com/Azure/ARO-HCP/backend/pkg/listers"
+	"github.com/Azure/ARO-HCP/backend/pkg/maestrohelpers"
 	"github.com/Azure/ARO-HCP/internal/admission"
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/cincinatti"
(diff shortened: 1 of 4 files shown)


backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -123,16 +126,14 @@ func (c *controlPlaneDesiredVersionSyncer) SyncOnce(ctx context.Context, key con
 		return utils.TrackError(fmt.Errorf("failed to get or create ServiceProviderCluster: %w", err))
 	}
 
-	// TODO bring the cluster uuid into serviceprovidercluster
-	clusterServiceCluster, err := c.clusterServiceClient.GetCluster(ctx, *existingCluster.ServiceProviderProperties.ClusterServiceID)
+	// Resolve the cluster UUID from the cached HostedCluster so we can build the Cincinnati client.
+	clusterUUID, found, err := maestrohelpers.GetCachedHostedClusterUUIDForCluster(ctx, c.clusterManagementClusterContentLister, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
```
machi1990, line 130: This is okay. For when we move CS cluster creation call to the backend https://github.com/Azure/ARO-HCP/pull/4477 we need to conditionally do this and use a generated uuid for the cincinatti calls

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -123,16 +126,14 @@ func (c *controlPlaneDesiredVersionSyncer) SyncOnce(ctx context.Context, key con
...
 	if err != nil {
-		return utils.TrackError(fmt.Errorf("failed to get cluster from Cluster Service: %w", err))
+		return err
 	}
-
-	// Create Cincinnati client with cluster UUID from Cluster Service
-	clusterUUID, err := uuid.Parse(clusterServiceCluster.ExternalID())
-	if err != nil {
-		return utils.TrackError(fmt.Errorf("invalid cluster UUID from Cluster Service: %w", err))
+	if !found {
```
machi1990, line 134: Same comment as https://github.com/Azure/ARO-HCP/pull/5045/changes#r3154775568

machi1990 (approved):
/lgtm

I left a comment in the cluster desired version controller about using a default UUID one for the case of install when we move the cluster creation call to the backend, that will be handled in the dedicated PR by @JakobGray

PR #5052, opened 2026-04-28
frontend: switch metrics to component-base legacyregistry

geoberle:
### What

The backend already uses legacyregistry for Go runtime and workqueue metrics. Aligning the frontend enables shared internal packages to register metrics once via legacyregistry and have them exposed by both services. This also fixes panic_total (registered via legacyregistry in internal/utils) being invisible on the frontend's /metrics endpoint.


bennerv:
/lgtm

Recalling my previous conversations around this with Gerd... we don't use promauto anywhere without passing an explicit registry; we explicitly define prom metrics, and at this point the same thing as the current code is the legacyregisterer.

raelga (approved):
/lgtm
/approve

PR #4513, opened 2026-03-18
CS management clusters into cosmosdb

geoberle:
### What

mirror CS management clusters into cosmosdb

* introduce new cosmosdb types
** Stamp - the infra deployment unit for a management cluster (think capi Machine)
** ManagementCluster - the provision shards for HCPs (think k8s Node)
* includes API types, CRUD interface, database lister, informer, OCM conversion logic
* ... and a periodic sync controller for importing them from CS
* CS is still the inventory for management clusters and source of truth
* includes a controller to sync the mgmt cluster placement information from CS into the serviceprovidercluster document

```json

"cosmosMetadata": {
    "resourceID": "/providers/microsoft.redhatopenshift/stamps/1"
},
"resourceId": "/providers/microsoft.redhatopenshift/stamps/1",
"spec": {},
"status": {
    "conditions": [
        {
            "type": "Approved",
            "status": "True",
            "lastTransitionTime": "2026-05-11T00:44:23Z",
            "reason": "AutoApproved",
            "message": "Synced from Cluster Service provision shard"
        }
    ]
}

{
  "cosmosMetadata": {
    "resourceID": /providers/microsoft.redhatopenshift/stamps/1/managementclusters/default
  },
  "spec": {
    "schedulingPolicy": "Schedulable"
  },
  "status": {
    "aksResourceID": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/hcp-underlay-westus3/providers/Microsoft.ContainerService/managedClusters/hcp-underlay-westus3-mgmt-1",
    "publicDNSZoneResourceID": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/global/providers/Microsoft.Network/dnsZones/westus3.aroapp.io",
    "hostedClustersSecretsKeyVaultURL": "https://hcp-underlay-westus3-mgmt-1-cx.vault.azure.net",
    "hostedClustersManagedIdentitiesKeyVaultURL": "https://hcp-underlay-westus3-mgmt-1-msi.vault.azure.net",
    "hostedClustersSecretsKeyVaultManagedIdentityClientID": "00000000-0000-0000-0000-000000000001",
    "maestroConsumerName": "hcp-underlay-westus3-mgmt-1",
    "maestroRESTAPIURL": "http://maestro.maestro.svc.cluster.local:8000",
    "maestroGRPCTarget": "maestro-grpc.maestro.svc.cluster.local:8090",
[...]

PR #5055, opened 2026-04-28
Phase 2: ManagementClusterDeployment type, fleet container, and FleetDBClient

geoberle:
⚠️ _hint for reviewer - ignore the first three commits - only review commit prefixed with "phase 2"
they other commits belong to the phase 1 PR https://github.com/Azure/ARO-HCP/pull/4513 and only here because phase 2 builds on it_

Introduces the ManagementClusterDeployment data type and supporting
infrastructure for Phase 2 of management cluster fleet management.
    
ManagementClusterDeployment is the infrastructure definition of a
management cluster, created by the provisioning pipeline. It lives in a
dedicated Fleet CosmosDB container, separate from the RP's Resources
container. A ManagementClusterPromotionController (future PR) will watch
deployments and promote them to ManagementCluster documents.
    
- Add ManagementClusterDeployment type with status fields for infrastructure coordinates (AKS resource ID, key vaults, Maestro config) and conditions (Approved)
- Add ManagementClusterDeploymentCRUD for provider-scoped resources
- Add FleetDBClient and FleetGlobalListers interfaces for the fleet management component, enforcing a clean boundary: the RP's DBClient and GlobalListers remain unchanged and unaware of the fleet container
- Add Fleet CosmosDB container to bicep with autoscale configuration
- Add Phase 2 conditions to ManagementCluster (ClustersServiceRegistered, MaestroRegistered, aggregate Ready)
- Add HostPort validator for maestroGRPCTarget validation

Enhancement: https://github.com/openshift-online/aro-enhancements/pull/1


internal/api/types_serviceprovider_cluster.go
```diff
@@ -129,6 +129,10 @@ type ServiceProviderClusterStatus struct {
 	// The reference contains a mapping between the logical name we give to the Maestro bundle internally
 	// and the Maestro Bundle Name and ID at the Maestro API level.
 	MaestroReadonlyBundles MaestroBundleReferenceList `json:"maestroReadonlyBundles,omitempty"`
+	// ManagementClusterID is the ID of the management cluster
```
deads2k, line 132: needs example of the ID

internal/database/database.go
```diff
@@ -219,6 +221,10 @@ func (d *cosmosDBClient) ServiceProviderNodePools(subscriptionID, resourceGroupN
 		d.resources, nodePoolResourceID, api.ServiceProviderNodePoolResourceType)
 }
 
+func (d *cosmosDBClient) ManagementClusters(subscriptionID, resourceGroupName string) ManagementClusterCRUD {
```
deads2k, line 224: Create a different interface: not DBClient. Sounds like these resources are going to be touched by other processes that probably don't need functions for seeing externalAuths for instance.

internal/database/fleet_database.go
```diff
@@ -0,0 +1,84 @@
...
+
+var _ FleetDBClient = &fleetDBClient{}
+
+type fleetDBClient struct {
+	fleet *azcosmos.ContainerClient
+}
+
+// NewFleetDBClient creates a FleetDBClient that accesses the Fleet container.
+func NewFleetDBClient(ctx context.Context, database *azcosmos.DatabaseClient) (FleetDBClient, error) {
+	fleet, err := database.NewContainer(fleetContainer)
```
deads2k, line 44: What are we going to partition on in this container?  Subscription doesn't seem to make much sense.  Just one partition with a fixed name?  Something else?

internal/database/database.go
deads2k: Additionally, the parition in the resource container by subscription doesn't seem logical for these, right?

deads2k (commented):
API looks largely good. Some comments.

internal/database/database.go
geoberle: right. let me move the `ManagementCluster` to the `fleet` container as well

PR #5056, opened 2026-04-28
bump aro-tools: CEL validation support for config schema

geoberle:
### What

https://github.com/Azure/ARO-Tools/pull/221 - cel validation code
https://github.com/Azure/ARO-HCP/pull/5003 - cel validations added to config schema

Exclude gen-crd-api-reference-docs v0.3.1-0.20260316152250-6bbddc29119c
in test/go.mod. tektoncd/pipeline@v1.11.1 requires this pseudo-version
but the referenced commit only exists in their fork
(tektoncd/ahmetb-gen-crd-api-reference-docs), not the original repo.
Their go.mod replace directive handles this locally but is invisible
to consumers, causing go get to fail on graph resolution.


raelga:
The broken `gen-crd-api-reference-docs` pseudo-version is currently blocking the sdp-pipelines bumper — every bump attempt fails at `go get github.com/Azure/ARO-HCP/tooling/templatize@<sha>` because Go can't resolve the transitive dependency. This is blocking all HCP deployments from progressing past the current pinned commit. Force-merging to unblock the release pipeline.

raelga (approved):
/lgtm

PR #5058, opened 2026-04-28
stop using cluster-service for frontend nodepool reads

deads2k:
this simplifies the read path and allows us to later disconnect the  write path.

### DIFF
--- a/frontend/pkg/frontend/node_pool.go
+++ b/frontend/pkg/frontend/node_pool.go
@@ -83,14 +83,6 @@ func (f *Frontend) ArmResourceListNodePools(writer http.ResponseWriter, request 
 	resourceGroupName := request.PathValue(PathSegmentResourceGroupName)
 	clusterName := request.PathValue(PathSegmentResourceName)
 
-	internalCluster, err := f.dbClient.HCPClusters(subscriptionID, resourceGroupName).Get(ctx, clusterName)
-	if err != nil {
-		return utils.TrackError(err)
-	}
-	if internalCluster.ServiceProviderProperties.ClusterServiceID == nil {
-		return utils.TrackError(fmt.Errorf("cluster %s has no ClusterServiceID", internalCluster.ID))
-	}
-
 	pagedResponse := arm.NewPagedResponse()
 
 	internalNodePoolIterator, err := f.dbClient.HCPClusters(subscriptionID, resourceGroupName).NodePools(clusterName).List(ctx, dbListOptionsFromRequest(request))
--- a/frontend/pkg/frontend/node_pool.go
+++ b/frontend/pkg/frontend/node_pool.go
@@ -83,6 +83,12 @@ func (f *Frontend) ArmResourceListNodePools(writer http.ResponseWriter, request 
 	resourceGroupName := request.PathValue(PathSegmentResourceGroupName)
 	clusterName := request.PathValue(PathSegmentResourceName)
 
+	// Verify the parent cluster exists so we return 404 instead of an empty
+	// list for a non-existent cluster (Cosmos List is prefix-based).
+	if _, err := f.dbClient.HCPClusters(subscriptionID, resourceGroupName).Get(ctx, clusterName); err != nil {
+		return utils.TrackError(err)
+	}
+
 	pagedResponse := arm.NewPagedResponse()
 
 	internalNodePoolIterator, err := f.dbClient.HCPClusters(subscriptionID, resourceGroupName).NodePools(clusterName).List(ctx, dbListOptionsFromRequest(request))
@@ -214,12 +220,7 @@ func decodeDesiredNodePoolCreate(ctx context.Context, azureLocation string) (*ap
 	return newInternalNodePool, nil
 }
 
-func (f *Frontend) newNodePoolAdmissionContext(ctx context.Context, nodePoolResourceID *azcorearm.ResourceID) (*admission.NodePoolAdmissionContext, error) {
-	cluster, err := f.getInternalClusterFromStorage(ctx, nodePoolResourceID.Parent)
-	if err != nil {
-		return nil, utils.TrackError(err)
-	}
-
+func (f *Frontend) newNodePoolAdmissionContext(ctx context.Context, cluster *api.HCPOpenShiftCluster) (*admission.NodePoolAdmissionContext, error) {
 	return &admission.NodePoolAdmissionContext{
 		Cluster: cluster,
 	}, nil
@@ -262,7 +263,7 @@ func (f *Frontend) createNodePool(writer http.ResponseWriter, request *http.Requ
 		return utils.TrackError(fmt.Errorf("cluster %s has no ClusterServiceID", cluster.ID))
 	}
 
-	admissionContext, err := f.newNodePoolAdmissionContext(ctx, resourceID)
+	admissionContext, err := f.newNodePoolAdmissionContext(ctx, cluster)
 	if err != nil {
 		return utils.TrackError(err)
 	}
@@ -542,7 +543,7 @@ func (f *Frontend) updateNodePoolInCosmos(ctx context.Context, writer http.Respo
 		return utils.TrackError(err)
 	}
 
-	admissionContext, err := f.newNodePoolAdmissionContext(ctx, oldInternalNodePool.ID)
+	admissionContext, err := f.newNodePoolAdmissionContext(ctx, cluster)
 	if err != nil {
 		return utils.TrackError(err)
 	}
@@ -550,7 +551,7 @@ func (f *Frontend) updateNodePoolInCosmos(ctx context.Context, writer http.Respo
 		Type:    operation.Update,
 		Options: validation.AFECsToValidationOptions(subscription.GetRegisteredFeatures()),
 	}
-	if mutationErrs := admission.MutateNodePool(ctx, admissionContext, restOperation, newInternalNodePool, nil); len(mutationErrs) > 0 {
+	if mutationErrs := admission.MutateNodePool(ctx, admissionContext, restOperation, newInternalNodePool, oldInternalNodePool); len(mutationErrs) > 0 {
 		return utils.TrackError(arm.CloudErrorFromFieldErrors(mutationErrs))
 	}
 
--- a/frontend/pkg/frontend/node_pool.go
+++ b/frontend/pkg/frontend/node_pool.go
@@ -214,6 +214,17 @@ func decodeDesiredNodePoolCreate(ctx context.Context, azureLocation string) (*ap
 	return newInternalNodePool, nil
 }
 
+func (f *Frontend) newNodePoolAdmissionContext(ctx context.Context, nodePoolResourceID *azcorearm.ResourceID) (*admission.NodePoolAdmissionContext, error) {
+	cluster, err := f.getInternalClusterFromStorage(ctx, nodePoolResourceID.Parent)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+
+	return &admission.NodePoolAdmissionContext{
+		Cluster: cluster,
+	}, nil
+}
+
 func (f *Frontend) createNodePool(writer http.ResponseWriter, request *http.Request) error {
 	ctx := request.Context()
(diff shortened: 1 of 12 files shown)


frontend/pkg/frontend/node_pool.go
miguelsorianod: Is this check needed after the changes?

internal/api/types_nodepool.go
```diff
@@ -83,11 +83,13 @@ type NodePoolVersionProfile struct {
 // NodePoolPlatformProfile represents a worker node pool configuration.
 // Visibility for the entire struct is "read create".
 type NodePoolPlatformProfile struct {
+	// Subnet can be nil and when it is nil, that means "use the cluster subnet field"
 	SubnetID               *azcorearm.ResourceID `json:"subnetId,omitempty"`
 	VMSize                 string                `json:"vmSize,omitempty"`
 	EnableEncryptionAtHost bool                  `json:"enableEncryptionAtHost"`
 	OSDisk                 OSDiskProfile         `json:"osDisk"`
-	AvailabilityZone       string                `json:"availabilityZone,omitempty"`
+	// AvailabilityZone can be empty, but we don't currently know what it means when it is empty.
```
miguelsorianod: What I wrote back down in CS is:
```
	// * If AvailabilityZone is not specified during creation, then the Node Pool
	//   is not part of any Availability Zone and the Node Pool's
	//   virtual machines are included in an Availability Set.
	// * If AvailabilityZone is specified, then the virtual machines are deployed
	//   as standalone instances in the specified Availability Zone.
	// * It must not be specified for Node Pools whose parent cluster is in an
	//   Azure location that does not support Availability Zones
	// * It is a logical Availability Zone
	// * It is optional during creation and immutable
```

At the end, it is whatever behavior Hypershift has as that value is simply propagated there. Unfortunately the documentation on the Hypershift side was/is detailed enough so that's what I ended up writing after investigating. The Hypershift Go doc around that attribute is:
```
	// availabilityZone is the failure domain identifier where the VM should be attached to. This must not be specified
	// for clusters in a location that does not support AvailabilityZone because it would cause a failure from Azure API.
	// +kubebuilder:validation:XValidation:rule="self in ['1', '2', '3']"
	// +optional
	// +kubebuilder:validation:MaxLength=255
	AvailabilityZone string `json:"availabilityZone,omitempty"`
```

If we really want to be sure about it I recommend contacting the Hypershift team to be sure what I wrote back then is accurate and it hasn't changed since then.

frontend/pkg/frontend/node_pool.go
```diff
@@ -363,12 +331,6 @@ func (f *Frontend) createNodePool(writer http.ResponseWriter, request *http.Requ
 	if !ok {
 		return fmt.Errorf("unexpected type %T", resultingUncastInternalNodePool)
 	}
-
-	// TODO this overwrite will transformed into a "set" function as we transition fields to ownership in cosmos
-	resultingInternalNodePool, err = mergeToInternalNodePool(csNodePool, resultingInternalNodePool, f.azureLocation)
```
miguelsorianod, line 368: During create I left a comment through other channels around potential concerns on CS side defaulting not moved to the RP.

miguelsorianod: As discussed, we are going to put a hold on this PR until those are addressed

deads2k:
> /hold see [#5058 (comment)](https://github.com/Azure/ARO-HCP/pull/5058#discussion_r3155526710)

now addressed.  Subnet is set via admission.

frontend/pkg/frontend/node_pool.go
```diff
@@ -254,6 +214,17 @@ func decodeDesiredNodePoolCreate(ctx context.Context, azureLocation string) (*ap
...
 
+func (f *Frontend) newNodePoolAdmissionContext(ctx context.Context, nodePoolResourceID *azcorearm.ResourceID) (*admission.NodePoolAdmissionContext, error) {
+	cluster, err := f.getInternalClusterFromStorage(ctx, nodePoolResourceID.Parent)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+
+	return &admission.NodePoolAdmissionContext{
+		Cluster: cluster,
+	}, nil
```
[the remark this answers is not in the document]
miguelsorianod: I am also interested in this. Why do we re-read the cluster from the DB if it's available already in-memory? can't that also cause inconsistency in the contents in the same request

frontend/pkg/frontend/node_pool.go
```diff
@@ -569,15 +542,22 @@ func (f *Frontend) updateNodePoolInCosmos(ctx context.Context, writer http.Respo
...
-	validationOp := operation.Operation{
+	admissionContext, err := f.newNodePoolAdmissionContext(ctx, oldInternalNodePool.ID)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	restOperation := operation.Operation{
 		Type:    operation.Update,
 		Options: validation.AFECsToValidationOptions(subscription.GetRegisteredFeatures()),
 	}
+	if mutationErrs := admission.MutateNodePool(ctx, admissionContext, restOperation, newInternalNodePool, nil); len(mutationErrs) > 0 {
```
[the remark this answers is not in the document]
miguelsorianod: I am also interested in understanding this better, why are we passing nil and not the old one?

internal/admission/admit_nodepool_test.go
```diff
@@ -15,19 +15,120 @@
...
+
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/validation"
 )
 
+func TestMutateNodePool(t *testing.T) {
+	const (
+		clusterSubnet  = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg/providers/Microsoft.Network/virtualNetworks/cluster-vnet/subnets/cluster-subnet"
+		nodePoolSubnet = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg/providers/Microsoft.Network/virtualNetworks/np-vnet/subnets/np-subnet"
```
miguelsorianod, line 40: While reading this it made me think that we should have this validation, which is being performed on CS side:
" The subnet resource ID must belong to the same VNet as the parent cluster's subnet".

I think we are also missing API documentation about it at ARM level, because it's being enforced by CS but it's not documented at ARM API level.

miguelsorianod: There is additionally another restriction (this one is documented at ARM API level already) that we should enforce:

"Note that a subnet cannot be reused between ARO-HCP Clusters, however the same subnet can be used for NodePools of the same cluster"

frontend/pkg/frontend/node_pool.go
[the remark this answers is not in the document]
deads2k: > I am also interested in this. Why do we re-read the cluster from the DB if it's available already in-memory? can't that also cause inconsistency in the contents in the same request

more generic for future usage.  I don't care enough to fight a bot.

frontend/pkg/frontend/node_pool.go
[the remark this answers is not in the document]
deads2k: > I am also interested in understanding this better, why are we passing nil and not the old one?

this one was just a bug.

internal/admission/admit_nodepool_test.go
deads2k: > There is additionally another restriction (this one is documented at ARM API level already) that we should enforce:

> "Note that a subnet cannot be reused between ARO-HCP Clusters, however the same subnet can be used for NodePools of the same cluster"

Another PR.  It likely requires a reservation type at the subscription level in the controllers and best effort in the frontend admission validation.  How is clsuter-service ensuring it?  Transaction lock?

deads2k: > While reading this it made me think that we should have this validation, which is being performed on CS side: " The subnet resource ID must belong to the same VNet as the parent cluster's subnet".
> 
> I think we are also missing API documentation about it at ARM level, because it's being enforced by CS but it's not documented at ARM API level.

Already present.  pushed some tests showing it.

miguelsorianod: > Another PR. It likely requires a reservation type at the subscription level in the controllers and best effort in the frontend admission validation. How is clsuter-service ensuring it? Transaction lock?

Yes. A postgresql AdvisoryLock is used where the key is the subnet resource id. Also it's important to notice that the check occurs when both creating a cluster and when creating a nodepool, and it needs to check whether other clusters already use it, but also whether other clusters' nodepools use it.

frontend/pkg/frontend/node_pool.go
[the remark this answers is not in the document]
miguelsorianod: non blocking but I still want to get further context: why do we need to re-read the cluster? can we not pass it to newNodePoolAdmissionContext?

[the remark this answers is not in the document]
miguelsorianod: ah, I see now that the PR has been updated since to precisely do that.

internal/admission/admit_nodepool.go
```diff
@@ -78,17 +83,25 @@ func AdmitNodePool(newNodePool, oldNodePool *api.HCPOpenShiftClusterNodePool, cl
...
-		if !strings.EqualFold(nodePoolVNet, clusterVNet) {
-			errs = append(errs, field.Invalid(
-				field.NewPath("properties", "platform", "subnetId"),
-				newNodePool.Properties.Platform.SubnetID,
-				fmt.Sprintf("must belong to the same VNet as the parent cluster VNet '%s'", clusterVNet),
-			))
+	// Check only if it is a creating nodepool or a change in the Subnet.
+	// Compare by string value (not pointer identity) so equal-but-distinct
+	// *azcorearm.ResourceID values aren't treated as a change.
+	if newNodePool.Properties.Platform.SubnetID != nil && cluster.CustomerProperties.Platform.SubnetID != nil {
```
miguelsorianod, line 89: Is the `cluster.CustomerProperties.Platform.SubnetID != nil` check needed? The cluster's subnet is a required attribute on creation.

internal/admission/admit_nodepool.go
```diff
@@ -43,17 +83,25 @@ func AdmitNodePool(newNodePool, oldNodePool *api.HCPOpenShiftClusterNodePool, cl
...
+		var oldSubnetID string
+		if oldNodePool != nil && oldNodePool.Properties.Platform.SubnetID != nil {
+			oldSubnetID = oldNodePool.Properties.Platform.SubnetID.String()
+		}
+		newSubnetID := newNodePool.Properties.Platform.SubnetID.String()
```
miguelsorianod, line 94: When checking this, nodepool admission seems to be executed after nodepool mutation. If that's what we want, at this point subnetID can be nil as it's an optional attribute during nodepool creation and mutation hasn't occurred yet. This would cause a nil pointer exception in that scenario.

internal/admission/admit_nodepool.go
deads2k: The bot wants it. We can keep arguing with it or deal with the cruft it leaves everywhere. I have no interest in bothering to argue.

internal/admission/admit_nodepool.go
deads2k: I don't get it. Doesn't the if block around this check this is non-nil?

miguelsorianod: > When checking this, nodepool admission seems to be executed after nodepool mutation. If that's what we want, at this point subnetID can be nil as it's an optional attribute during nodepool creation and mutation hasn't occurred yet. This would cause a nil pointer exception in that scenario.

I mixed the order when reviewing. Mutation occurs before admission validation so we should be fine here.

internal/admission/admit_nodepool.go
miguelsorianod: I understand this only occurs when the content is changed so it shouldn't impact future PRs.

Seeing the nil check though implies this could be nil for some reason which makes it confusing to reason about.

I don't have a strong preference so non blocker for me.

PR #5059, opened 2026-04-28
Remove cluster-service to cosmos nodepool migration

deads2k:
These controllers have done their work and the frontend keeps track of    all existing data.

/hold

https://github.com/Azure/ARO-HCP/pull/5058 must go to prod before merging


backend/pkg/controllers/nodepoolpropertiescontroller/node_pool_properties_sync.go
```diff
@@ -1,142 +0,0 @@
-// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: Should we keep this one for now because of defaulting on CS side occurring right now? we are getting it on nodepool creation on the frontend but if at some point we remove the reads from there and we haven't moved defaulting to the RP we would stop getting it

deads2k:
/hold

https://github.com/Azure/ARO-HCP/pull/5058 must go to prod before merging

deads2k:
simple rebase

/label lgtm

PR #5060, opened 2026-04-28
gather-observability: add JUnit output for alerts - attemp 2

geoberle:
### What

Generate junit XML file containing all the alerts defined in the azure monitoring workspaces. fired alerts will be tracked as failed "alert foo did not fire" tests in the junit file.

Unknown alerts will not yet fail the gather step. For now only the junit file with the failures will be exposed.


raelga (approved):
/lgtm
/approve

PR #5061, opened 2026-04-28
label tests with high MI usage, run them first

mgahagan73:
https://redhat.atlassian.net/browse/ARO-26090

### What

Check to see if we can run the 2 tests with high MI usage first to mitigate resource starvation issues. 

### Why

Temporarily mitigate resource starvation issues. 

Stop-gap measure to reduce resource starvation while we work on a more comprehensive solution.

### DIFF
--- a/test/util/labels/labels.go
+++ b/test/util/labels/labels.go
@@ -54,6 +54,13 @@ var (
 	AroRpApiCompatible = ginkgo.Label("ARO-HCP-RP-API-Compatible")
 )
 
+// Managed identity container demand for resource-aware scheduling.
+// Tests without these labels default to needing 1 container.
+var (
+	MIDemandHigh   = ginkgo.Label("MIDemand:High")
+	MIDemandMedium = ginkgo.Label("MIDemand:Medium")
+)
+
 // Environments this test can be used in.
 var (
 	RequireNothing        = ginkgo.Label("PreLaunchSetup:None")
--- a/test/cmd/aro-hcp-tests/main.go
+++ b/test/cmd/aro-hcp-tests/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"sort"
 	"time"
 
 	// If using ginkgo, import your tests here
@@ -27,6 +28,7 @@ import (
 
 	"github.com/openshift-eng/openshift-tests-extension/pkg/cmd"
 	e "github.com/openshift-eng/openshift-tests-extension/pkg/extension"
+	et "github.com/openshift-eng/openshift-tests-extension/pkg/extension/extensiontests"
 	g "github.com/openshift-eng/openshift-tests-extension/pkg/ginkgo"
 
 	"github.com/Azure/ARO-HCP/internal/api"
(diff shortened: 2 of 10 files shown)


mgahagan73:
In the previous e2e parallel run only the cillium test failed (waiting on admin creds) so the run is fine for purposes of testing this PR. Here is the result:

The entire test run took 2 hours, 13 minutes, and 16 seconds (from 19:46:46Z to 22:00:02Z).

Tests requiring multiple containers:
- Customer should not be able to reuse subnets and NSGs between clusters (requires 2 MI containers)
- Customer should be able to create several HCP clusters in their customer resource group, but not in the same managed resource group (requires 3 MI containers)
Neither test waited any significant amount of time indicating they both ran first, consuming 5 containers total while there were 20 available.

long waiting tests (each requiring a single output container)
| Test Name | Wait Time | Waiting For |
|---|---|---|
| Customer should upgrade and update a nodepool from 4.21.z to 4.21.zLatest | **6m 10s** | Identity Container |
| Customer should upgrade and update a nodepool from 4.20.z to 4.20.zLatest | **8m 2s** | Identity Container |
| Nodepool Ephemeral OS Disk should create a nodepool with ephemeral OS disk when autoRepair is enabled | **15m 5s** | Identity Container |
| ARO-HCP should be able to perform a control plane and node pool install with OCP candidate channel for 4.22 | **20m 1s** | Identity Container |
| Customer should be able to create a cluster with an external auth config and get the external auth config | **21m 7s** | Identity Container |
| Service Provider should upgrade the control plane z-stream automatically on behalf of the customer for 4.20 | **25m 0s** | Identity Container |
| ARO-HCP should be able to perform a control plane and node pool install with OCP candidate channel for 4.21 | **25m 1s** | Identity Container |
| Service Provider should upgrade the control plane z-stream automatically on behalf of the customer for 4.21 | **31m 4s** | Identity Container |
| Service Provider should upgrade the control plane z-stream automatically on behalf of the customer for 4.19 | **31m 5s** | Identity Container |
| Customer should be able to successfully upgrade control plane minor version from 4.20 minor to 4.21 minor | **32m 6s** | Identity Container |

test/util/labels/labels.go
```diff
@@ -54,6 +54,13 @@ var (
 	AroRpApiCompatible = ginkgo.Label("ARO-HCP-RP-API-Compatible")
 )
 
+// Managed identity container demand for resource-aware scheduling.
+// Tests without these labels default to needing 1 container.
+var (
+	MIDemandHigh   = ginkgo.Label("MIDemand:High")
+	MIDemandMedium = ginkgo.Label("MIDemand:Medium")
```
mbukatov, line 61: Would it make sense to add a comment about where is a threshold between high and medium demand?

mbukatov (commented):
This is a good idea.

/lgtm

test/util/labels/labels.go
mgahagan73: For the purposes of this temporary measure MIDemandMedium is for tests using 2 MI containers, MiDemandHIgh is for 3 or more.  This will only buy us time until we get a better scheduler in place, If we add a few more medium/high MI demand tests before scheduler work is completed we are going to have the same resource starvation problem.

PR #5062, opened 2026-04-28
backend: Pass a PassiveClock to operation controllers

mbarnes:
### What

This allows the [RealClock instance in utils.go](https://github.com/Azure/ARO-HCP/blob/efb6799f1ef037a621b7e18f188ca0b20d0b3467/backend/pkg/controllers/operationcontrollers/utils.go#L46) to be dropped since all the operation controllers now possess a `PassiveClock` interface to pass to utility functions.

### Why

Light refactoring for controller consistency.

### Testing

Existing unit and integration tests should suffice.

### DIFF
--- a/backend/pkg/app/backend.go
+++ b/backend/pkg/app/backend.go
@@ -58,6 +58,7 @@ import (
 )
 
 type Backend struct {
+	clock   utilsclock.PassiveClock
 	options *BackendOptions
 }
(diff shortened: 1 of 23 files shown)


mbarnes:
Moving these changes to #5393 since our forking policy changed and there's too many merge conflicts here.

PR #5068, opened 2026-04-29
Rename e2e test helper files to drop _test.go suffix

raelga:
[AROSLSRE-737](https://redhat.atlassian.net/browse/AROSLSRE-737)

### What

Renames e2e test helper files to drop the `_test.go` suffix:
- `cluster_create_test.go` -> `cluster_create.go`
- `cluster_delete_test.go` -> `cluster_delete.go`
- `cluster_get_test.go` -> `cluster_get.go`
- `cluster_list_test.go` -> `cluster_list.go`
- `nodepool_get_test.go` -> `nodepool_get.go`

Also regenerates test suite fixtures (picks up a new test case that was added to main but not reflected in fixtures).

### Why

These files contain Ginkgo specs and helpers, not standard Go tests. The `_test.go` suffix is misleading and causes `go test` to discover them incorrectly. Most other spec files in the `test/e2e/` directory already use the `.go` suffix.

Originally proposed in #4563 which went stale due to fixture drift.

### Testing

- `go build -tags E2Etests ./e2e/...` passes
- `TestMainListSuitesForEachSuite` passes with regenerated fixtures


mbukatov (changes requested):
My original comment this fully applies here: https://github.com/Azure/ARO-HCP/pull/4563#pullrequestreview-3992228015

/hold

raelga:
Thanks @mbukatov 

/close

PR #5072, opened 2026-04-29
add negative test cases for requesting and revoking break class crede…

tmstff:
…ntials

### What

add negative test cases for requesting and revoking break class credentials

### Why

to have those conditions covered by automated tests and see when the behavior breaks.


backend/go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8
 	github.com/google/uuid v1.6.0
 	github.com/openshift-online/maestro v0.0.0-20260213014104-081c1f6df17b
-	github.com/openshift-online/ocm-sdk-go v0.1.498
+	github.com/openshift-online/ocm-sdk-go v0.1.499
```
mbarnes, line 20: This may have slipped in accidentally but extra test cases don't appear to warrant a version bump.  Suggest reverting this.
```suggestion
	github.com/openshift-online/ocm-sdk-go v0.1.498
```

backend/go.sum
```diff
@@ -189,6 +189,8 @@ github.com/openshift-online/ocm-api-model/model v0.0.453 h1:CsOyRhrpxzXcn4aaobDg
 github.com/openshift-online/ocm-api-model/model v0.0.453/go.mod h1:PQIoq6P8Vlb7goOdRMLK8nJY+B7HH0RTqYAa4kyidTE=
 github.com/openshift-online/ocm-sdk-go v0.1.498 h1:XOPq5eh2UZpv/21jlULAZ3oh7xCHv8wf7Sl3T3wsRUs=
 github.com/openshift-online/ocm-sdk-go v0.1.498/go.mod h1:V1upeXpkgHmyRy0cA2e+i9X/kUBEnv7sBYC6UdRXfzE=
+github.com/openshift-online/ocm-sdk-go v0.1.499 h1:QDr980dOSV0tFNrvOGiorno4l7iMq8lNxU7PY2rVqzY=
+github.com/openshift-online/ocm-sdk-go v0.1.499/go.mod h1:eKqLMWEeZ6vadCHjlRPsePdqap6c34aM8TGxjMgZwKY=
```
mbarnes, line 175: Same... looks unnecessary.
```suggestion
```

mbarnes (dismissed):
Remove the unnecessary `go.mod` / `go.sum` change.

Apart from that I'm wondering if all the test files in this directory should split out a separate function for testing `ShouldProcess`, as some of these new test cases are aiming more at that than `SynchronizeOperation` logic.  Take it up if you want but I won't insist on it here.

backend/pkg/controllers/operationcontrollers/operation_revoke_credentials_test.go
```diff
@@ -151,7 +220,7 @@ func TestOperationRevokeCredentials_SyncrhonizeOperation(t *testing.T) {
 						objs = append(objs, breakGlassCredential)
 					}
 					return ocm.NewSimpleBreakGlassCredentialsListIterator(objs, nil)
-				})
+				}).MaxTimes(1)
```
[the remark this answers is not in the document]
tmstff: It's OK as is, as the behavior already implies that the method returns the desired values - which only happens if it is called.

tmstff:
> Remove the unnecessary `go.mod` / `go.sum` change.
> 
> Apart from that I'm wondering if all the test files in this directory should split out a separate function for testing `ShouldProcess`, as some of these new test cases are aiming more at that than `SynchronizeOperation` logic. Take it up if you want but I won't insist on it here.

Hey @mbarnes ! Thanks for the input! I removed that go.* changes :-) Concerning `ShouldProcess` - I think you're right. I will put it on my ToDo list as a further enhancement - and leave it for now as it is, if that's ok!

tmstff:
moved to https://github.com/Azure/ARO-HCP/pull/5213

PR #5077, opened 2026-04-29
Updating credential refresher alerts

rachelvweber:
### What

Splitting credential refresher alert into expiring in less than 30 days and already expired.

### Why

The mitigation steps for these two scenarios are different.

### Testing

Tested by prometheus rules test file and e2e build gate.


stevekuznetsov (approved):
/lgtm
/approve

PR #5078, opened 2026-04-29
fetch vm serial console logs

mbukatov:
https://redhat.atlassian.net/browse/ARO-26562

### What

- implement ability to fetch vm serial console logs via `DownloadAllVirtualMachineConsoleLogs` and `GetVirtualMachineConsoleLog` functions in framework module
- add `LogDirPath` into test context (so that we can store VM console log files there)
- function `CreateNodePoolFromParam` is extended (now it requires logger in the same was as `CreateHCPClusterFromParam` and managed resource group) so that it will fetch VM console logs in case of a node pool deployment failure
- updates E2E tests to align with new API of CreateNodePoolFromParam function

### Why

To improve ability to debug node pool deployment timeouts.

### Testing

During development, introduce temporary testing commits to:

- fetch the console logs every time (even if the node pool deploys with success)
- make node deployment fail

To make sure the functionality is able to fetch the logs into expected location as expected.

See results in [2051720498206740480](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/5078/pull-ci-Azure-ARO-HCP-main-stage-e2e-parallel/2051720498206740480) run:

*Customer should be able to create a HCP cluster and use cilium CNI plugin* test failed, and the VM console logs were fetched:

```
  STEP: creating the node pool @ 05/05/26 18:22:13.794
"ts"="2026-05-05 19:07:15.699253" "level"=0 "msg"="VM console log fetched" "vmName"="cilium-cl-cilium-np-xbl5c-j4j6k" "targetPath"="/logs/artifacts/Customer_should_be_able_to_create_a_HCP_cluster_and_use_cilium_CNI_plugin/cilium-cl-cilium-np-xbl5c-j4j6k-console.log"
"ts"="2026-05-05 19:07:16.530894" "level"=0 "msg"="VM console log fetched" "vmName"="cilium-cl-cilium-np-xbl5c-lv7dp" "targetPath"="/logs/artifacts/Customer_should_be_able_to_create_a_HCP_cluster_and_use_cilium_CNI_plugin/cilium-cl-cilium-np-xbl5c-lv7dp-console.log"
  STEP: checking that cilium is running and nodes are in Ready state @ 05/05/26 19:07:16.53
```

And stored in artifacts directory of the test [2051720498206740480/artifacts/stage-e2e-parallel/aro-hcp-test-persistent/artifacts/Customer_should_be_able_to_create_a_HCP_cluster_and_use_cilium_CNI_plugin](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/5078/pull-ci-Azure-ARO-HCP-main-stage-e2e-parallel/2051720498206740480/artifacts/stage-e2e-parallel/aro-hcp-test-persistent/artifacts/Customer_should_be_able_to_create_a_HCP_cluster_and_use_cilium_CNI_plugin/)

- In a separate commit, I moved RunVMCommand and GetVirtualMachinesInResourceGroup into new file dedicated to vm functions, to avoid adding these all over the place, and to avoid having everything in hcp helpers.
- The functionality is added to `CreateNodePoolFromParam` so that it will be available for most of E2E test cases. We can consider moving the functionality to a lower level, to cover more tests though.

### DIFF
--- a/test/e2e/cluster_create_complex_cilium_kv.go
+++ b/test/e2e/cluster_create_complex_cilium_kv.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5"
 
 	operatorclient "github.com/openshift/client-go/operator/clientset/versioned"
 
@@ -188,9 +189,22 @@ var _ = Describe("Customer", func() {
 			// We delay checking the error on purpose to get more details
 			// about the issue by running the verifiers.
 
+			var consoleLogErr error = nil
+			if nodePoolErr != nil {
+				var computeFactory *armcompute.ClientFactory
+				computeFactory, consoleLogErr = tc.GetARMComputeClientFactory(ctx)
+				if consoleLogErr == nil {
+					consoleLogErr = framework.DownloadAllVirtualMachineConsoleLogs(
+						ctx,
+						computeFactory,
+						clusterParams.ManagedResourceGroupName,
+						tc.LogDirPath)
+				}
+			}
+
 			By("verifying nodes become Ready with Cilium CNI")
 			err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyNodesReady(), verifiers.VerifyCiliumOperational("kube-system", "k8s-app=cilium"))
-			Expect(errors.Join(err, nodePoolErr)).NotTo(HaveOccurred())
+			Expect(errors.Join(err, nodePoolErr, consoleLogErr)).NotTo(HaveOccurred())
 
 			By("verifying a simple web app can run with cilium")
 			err = verifiers.VerifySimpleWebApp().Verify(ctx, adminRESTConfig)
--- a/test/util/framework/vm_helper.go
+++ b/test/util/framework/vm_helper.go
@@ -16,7 +16,6 @@ package framework
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -113,43 +112,39 @@ func RunVMCommand(ctx context.Context, tc interface {
 }
 
 // GetVirtualMachinesInResourceGroup lists all VMs in the given resource group
-// with a polling loop to handle ARM replication delays.
+// if the VM list contains at least expectedMinimumCount items
 func GetVirtualMachinesInResourceGroup(
 	ctx context.Context,
 	computeClientFactory *armcompute.ClientFactory,
 	resourceGroupName string,
 	expectedMinimumCount int,
-	timeout time.Duration,
 ) ([]*armcompute.VirtualMachine, error) {
-	ctx, cancel := context.WithTimeoutCause(ctx, timeout,
-		fmt.Errorf("timed out waiting for %d VMs in resource group %q", expectedMinimumCount, resourceGroupName))
-	defer cancel()
-
 	vmClient := computeClientFactory.NewVirtualMachinesClient()
-	const pollInterval = 10 * time.Second
-
-	for {
-		var vms []*armcompute.VirtualMachine
-		pager := vmClient.NewListPager(resourceGroupName, nil)
-		for pager.More() {
-			page, err := pager.NextPage(ctx)
-			if err != nil {
-				if errors.Is(err, context.DeadlineExceeded) {
-					return nil, fmt.Errorf("caused by: %w, error: %w", context.Cause(ctx), err)
-				}
-				return nil, fmt.Errorf("failed to list VMs in resource group %q: %w", resourceGroupName, err)
-			}
-			vms = append(vms, page.Value...)
+	var vms []*armcompute.VirtualMachine
+	pager := vmClient.NewListPager(resourceGroupName, nil)
+	for pager.More() {
+		page, err := pager.NextPage(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to list VMs in resource group %q: %w", resourceGroupName, err)
 		}
-
-		if len(vms) >= expectedMinimumCount {
-			return vms, nil
-		}
-
-		select {
-		case <-ctx.Done():
-			return nil, fmt.Errorf("caused by: %w, error: %w", context.Cause(ctx), ctx.Err())
-		case <-time.After(pollInterval):
-		}
+		vms = append(vms, page.Value...)
 	}
+
+	if len(vms) < expectedMinimumCount {
+		vmNames := make([]string, 0, len(vms))
+		for _, vm := range vms {
+			if vm.Name != nil {
+				vmNames = append(vmNames, *vm.Name)
+			}
+		}
+		return nil, fmt.Errorf(
+			"not enough VMs in resource group %q: expecting %d, got %d (%v)",
+			resourceGroupName,
+			expectedMinimumCount,
+			len(vms),
+			vmNames,
+		)
+	}
+
+	return vms, nil
 }
--- a/test/util/framework/vm_helper.go
+++ b/test/util/framework/vm_helper.go
@@ -1,0 +1,155 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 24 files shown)


mbukatov:
Nice, looks like it works, console logs for both vms were fetched and are available in artifacts dir: [artifacts/stage-e2e-parallel/aro-hcp-test-persistent/artifacts/cilium-cl-cilium-np-wrq65-f42b4-console.log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/5078/pull-ci-Azure-ARO-HCP-main-stage-e2e-parallel/2049587490066534400/artifacts/stage-e2e-parallel/aro-hcp-test-persistent/artifacts/cilium-cl-cilium-np-wrq65-f42b4-console.log)

I can go on to polish the PR a bit.

test/e2e/cluster_create_complex_cilium_kv.go
```diff
@@ -188,9 +189,22 @@ var _ = Describe("Customer", func() {
 			// We delay checking the error on purpose to get more details
 			// about the issue by running the verifiers.
 
+			var consoleLogErr error = nil
+			if nodePoolErr != nil {
```
stevekuznetsov, line 193: I think it would be a more natural fit to put this in the node pool verifier so every caller gets it for free - I imagine I'll see that diff when I scroll down, so then I wonder, why is this here explicitly?

test/util/framework/vm_helper.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+
+		return strings.TrimSpace(output), nil
+	}
+
+	return "", nil
+}
+
+// GetVirtualMachinesInResourceGroup lists all VMs in the given resource group
+// with a polling loop to handle ARM replication delays.
```
stevekuznetsov: This comment is insufficiently precise - in what cases do we want to wait? What is an "ARM replication delay"?

test/util/framework/vm_helper.go
```diff
@@ -0,0 +1,270 @@
...
+) error {
+	logger := ginkgo.GinkgoLogr
+
+	// Create target directory if it doesn't exist
+	if err := os.MkdirAll(targetDirectory, 0755); err != nil {
+		return fmt.Errorf("failed to create target directory %q: %w", targetDirectory, err)
+	}
+
+	// List all VMs in the resource group
+	// Use 0 as expectedMinimumCount to return immediately with whatever VMs exist,
```
stevekuznetsov, line 218: I think you might be better off just calling the VM list API directly, instead of (ab)using an unrelated method that was written for some other purpose.

test/util/framework/vm_helper.go
```diff
@@ -0,0 +1,270 @@
...
+	// as we expect to run this after a node pool deployment failures
+	vms, err := GetVirtualMachinesInResourceGroup(ctx, computeClientFactory, resourceGroupName, 0, 5*time.Minute)
+	if err != nil {
+		return fmt.Errorf("failed to list VMs: %w", err)
+	}
+
+	if len(vms) == 0 {
+		return fmt.Errorf("no VMs found in resource group %q", resourceGroupName)
```
stevekuznetsov: Unclear if this is really an error?

test/util/framework/vm_helper.go
```diff
@@ -0,0 +1,270 @@
...
+		logFilePath := filepath.Join(targetDirectory, fmt.Sprintf("%s-console.log", *vm.Name))
+		logFile, err := os.Create(logFilePath)
+		if err != nil {
+			logReader.Close()
+			downloadErrors = append(downloadErrors, fmt.Sprintf("VM %q: failed to create file: %v", *vm.Name, err))
+			continue
+		}
+
+		_, err = io.Copy(logFile, logReader)
+		logFile.Close()
```
stevekuznetsov: these calls can error, no?

test/e2e/cluster_create_complex_cilium_kv.go
mbukatov: Yes. This test doesn't use `tc.CreateNodePoolFromParam` (which was extended in the way you suggest), but `framework.CreateNodePoolAndWait20251223`.

As I note in the description of the PR, we should consider adding the functionality into lower level of these create node pool functions, but I decided to leave it at `tc.CreateNodePoolFromParam` for the initial version of the PR.

test/util/framework/vm_helper.go
mbukatov: I just moved the function without touching it.

test/e2e/cluster_create_complex_cilium_kv.go
```diff
@@ -188,9 +189,22 @@ var _ = Describe("Customer", func() {
...
+				var computeFactory *armcompute.ClientFactory
+				computeFactory, consoleLogErr = tc.GetARMComputeClientFactory(ctx)
+				if consoleLogErr == nil {
+					consoleLogErr = framework.DownloadAllVirtualMachineConsoleLogs(
+						ctx,
+						computeFactory,
+						clusterParams.ManagedResourceGroupName,
+						tc.LogDirPath)
+				}
```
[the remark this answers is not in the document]
mbukatov: Hmm, I didn't try the local runs (since this is not directly possible with any but my personal dev instance which I would have to deploy first). But I would expect that if `ARTIFACT_DIR` is empty, the path would then resolve to whatever is the current working directory plus the extra paths. Skipping the download if `tc.LogDirPath == ""` doesn't make sense. If this is a problem, it needs to be fixed in the framework.

test/util/framework/vm_helper.go
```diff
@@ -0,0 +1,270 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch console log from blob storage for VM %q: %w", vmName, err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		resp.Body.Close()
+		return nil, fmt.Errorf("failed to fetch console log for VM %q, HTTP status: %d", vmName, resp.StatusCode)
+	}
+
+	return resp.Body, nil
```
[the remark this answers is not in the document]
mbukatov: I will check this one.

test/util/framework/vm_helper.go
mbukatov: Ah, right. This should be just logged.

test/util/framework/vm_helper.go
stevekuznetsov: I know - you touch it, you improve it :D

I recommend reaching out to the original author(s).

test/util/framework/vm_helper.go
mbukatov: yes, and we can log it

raelga (approved):
/lgtm
/approve

test/util/framework/vm_helper.go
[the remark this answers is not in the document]
mbukatov: addressed in 5991f4638046dc38fbf9889b82045a3bb35d0d5b

test/util/framework/vm_helper.go
mbukatov: done via df898e39627da07730e7a2f58e7181a586dec2be, I simplified the function a bit

test/util/framework/vm_helper.go
mbukatov: done in 5991f4638046dc38fbf9889b82045a3bb35d0d5b

test/util/framework/vm_helper.go
mbukatov: done in 5991f4638046dc38fbf9889b82045a3bb35d0d5b

mbukatov: but the error reporting on close is quite interesting, if it ever happens, you have bigger problems (like filesystem gets full or were remounted read only out of sudden) ... that said, if I to believe the go doc for this function, it reports error only when it's called more than once

mbukatov:
I addressed most comments. Leaving out:

- notes about `ARTIFACT_DIR`, I would like to have a better story about local runs than we currently have, so I don't want to get into this here, since I would most likely ended up with additional changes elsewhere
- the logs are automatically fetched in `tc.CreateNodePoolFromParam` when node pool deployment fails, which means that any test using lower fuctions such as  `framework.CreateNodePoolAndWait20251223` won't get this for free, which is why the complex test have extra functionality for this ...  @miquelsi is currently exploring this area (how to work with different API versions) via [ARO-25892](https://redhat.atlassian.net/browse/ARO-25892) refinement

stevekuznetsov (approved):
/lgtm
/approve

PR #5086, opened 2026-04-30
E2E test: Improve logging and timeouts in Authorized CIDRs Connectivity test case

mvacula02:
[ARO-24668](https://redhat.atlassian.net/browse/ARO-24668)

## What

Improve logging to include command and output details of assertion error messages.
Increase timeout (2 -> 5 minutes) for Eventually blocks

## Why

Previously the test case was returning generic timeout error messages, with no useful information for debugging.
The Eventually timeout should be strictly longer than the inner RunVMCommand timeout. Increasing to 5 minutes allows for a retry, which lets the CIDR rules to propagate as intended.

## Testing
This is an improvement to an existing test.

### DIFF
--- a/test/e2e/cluster_authorized_cidrs_connectivity.go
+++ b/test/e2e/cluster_authorized_cidrs_connectivity.go
@@ -161,10 +161,9 @@ var _ = Describe("Authorized CIDRs", func() {
 					output, err := framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, connectivityTest, 2*time.Minute)
 					g.Expect(err).NotTo(HaveOccurred(), "RunVMCommand for connectivity test failed (command: %s)", connectivityTest)
 
-					// Should get HTTP response (likely 401 or 200, but not connection refused)
 					httpCode = strings.TrimSpace(output)
 					By(fmt.Sprintf("VM received HTTP status code: %s", httpCode))
-					g.Expect(httpCode).To(MatchRegexp("^[2-5][0-9][0-9]$"), "Should receive valid HTTP status code from authorized IP, got: %s", httpCode)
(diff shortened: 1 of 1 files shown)


mbukatov (commented):
/lgtm

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -159,13 +159,13 @@ var _ = Describe("Authorized CIDRs", func() {
...
 					output, err := framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, connectivityTest, 2*time.Minute)
-					g.Expect(err).NotTo(HaveOccurred())
+					g.Expect(err).NotTo(HaveOccurred(), "RunVMCommand for connectivity test failed (command: %s)", connectivityTest)
 
 					// Should get HTTP response (likely 401 or 200, but not connection refused)
 					httpCode = strings.TrimSpace(output)
 					By(fmt.Sprintf("VM received HTTP status code: %s", httpCode))
-					g.Expect(httpCode).To(MatchRegexp("^[2-5][0-9][0-9]$"), "Should receive valid HTTP status code from authorized IP")
-				}, 2*time.Minute, 10*time.Second).Should(Succeed())
+					g.Expect(httpCode).To(MatchRegexp("^[2-5][0-9][0-9]$"), "Should receive valid HTTP status code from authorized IP, got: %s", httpCode)
```
mgahagan73: If we are intending to treat any http status code as valid because we just want to check the connection. I think we can replace the regex with a check for "000" because curl should return that in the event it can't make a TCP connection at all (DNS error, connect refused etc.) We could use something like:
 g.Expect(httpCode).NotTo(Equal("000"), "Should receive HTTP response from authorized IP, got curl error code 000 (no response)")

mgahagan73:
Nice job!

/approve

miquelsi (approved):
/lgtm

PR #5088, opened 2026-04-30
chore: scale maestro server to 1

machi1990:
### What

chore: scale maestro server to 1

### Why

We don't need 3, 1 is fine and it avoid us onto getting onto some leadership/ownership issues with > 1 replicas


machi1990:
Not needed, there is https://github.com/Azure/ARO-HCP/pull/5087

PR #5089, opened 2026-04-30
Retry AddPassword on untyped transient Graph SDK errors

miquelsi:
https://redhat.atlassian.net/browse/ARO-26567

### What

Retry AddPassword on untyped transient Graph SDK errors

### Why

This fix updates the Graph AddPassword retry behavior so transient failures no longer fail fast when the SDK returns an untyped error shape: instead of only retrying recognized OData transient codes (404, 429, 5xx) and immediately aborting everything else, it now only stops early for clearly non-transient typed OData errors and retries unknown/untyped errors within the existing polling window, which makes app-registration password creation resilient to Microsoft Graph eventual-consistency delays that were causing the external-auth e2e test to fail after a single attempt.

### DIFF
--- a/internal/graph/util/applications.go
+++ b/internal/graph/util/applications.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


internal/graph/util/applications.go
```diff
@@ -103,16 +103,18 @@ func (c *Client) AddPassword(ctx context.Context, appID, displayName string, sta
...
+			// error shapes returned by the Graph SDK, also retry to tolerate
+			// eventual-consistency propagation delays.
 			var odataErr *odataerrors.ODataError
 			if errors.As(err, &odataErr) {
 				code := odataErr.ResponseStatusCode
-				if code == 404 || code == 429 || code >= 500 {
-					return false, nil
+				if code != 404 && code != 429 && code < 500 {
+					// Non-transient typed OData error, stop retrying.
+					return false, err
```
mbarnes: Bare minimum, use the `http.Status*` constants please.

I thought the Graph SDK has an automatic retry policy for some of these error codes like [azure-sdk-for-go does](https://github.com/Azure/azure-sdk-for-go/blob/b87aaa53f9af4cf7b3c6eff36e1d5e7484eeea69/sdk/azcore/policy/policy.go#L111-L120).  `StatusNotFound` won't be among them but `StatusTooManyRequests` should be.  It's buried deep in one of Microsoft's kiota packages.  I've tracked it down before; may have to do so again to verify.

miquelsi: Done

PR #5091, opened 2026-05-01
Allowing pipeline runs with multiple topologies

rachelvweber:
### What

Allow pipeline runs that reference multiple topologies

### Why

Allows us to combine the topologies in here and in sdp-pipelines into real global entrypoint.

### Testing

- Tested pipeline runs locally that reference ARO-HCP/topology.yaml and sdp-pipelines/hcp/topology.yaml
- Confirmed that git diffing for changes to services works by running locally
- pointed my local sdp-pipelines at this version to confirm it still builds and `make validate` still runs correctly

### DIFF
--- a/tooling/templatize/cmd/configuration/validate/cmd.go
+++ b/tooling/templatize/cmd/configuration/validate/cmd.go
@@ -26,13 +26,13 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 )
 
-func NewCommand(centralRemoteUrl string) (*cobra.Command, error) {
+func NewCommand(centralRemoteUrls ...string) (*cobra.Command, error) {
 	scratchDir := filepath.Join(os.TempDir(), "config-"+rand.String(8))
 	if err := os.MkdirAll(scratchDir, os.ModePerm); err != nil {
 		return nil, fmt.Errorf("failed to create scratch directory: %w", err)
 	}
 
-	opts := DefaultOptions(scratchDir, centralRemoteUrl)
+	opts := DefaultOptions(scratchDir, centralRemoteUrls)
 	cmd := &cobra.Command{
 		Use:           "validate",
 		Short:         "Validate rendered configurations for all clouds, environments, and regions.",
--- a/tooling/templatize/cmd/configuration/validate/options.go
+++ b/tooling/templatize/cmd/configuration/validate/options.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -39,10 +40,10 @@ import (
 	"github.com/Azure/ARO-HCP/tooling/templatize/pkg/settings"
 )
 
-func DefaultOptions(outputDir string, url string) *RawOptions {
+func DefaultOptions(outputDir string, urls []string) *RawOptions {
 	return &RawOptions{
 		OutputDir:        outputDir,
-		CentralRemoteUrl: url,
+		CentralRemoteUrl: urls,
 	}
 }
 
@@ -51,7 +52,7 @@ func BindOptions(opts *RawOptions, cmd *cobra.Command) error {
 	cmd.Flags().StringVar(&opts.DevSettingsFile, "dev-settings-file", opts.DevSettingsFile, "Validate only the combinations present in the settings file, using public production Ev2 contexts.")
 	cmd.Flags().StringVar(&opts.DigestFile, "digest-file", opts.DigestFile, "File holding digests of previously-rendered configurations to validate with.")
 	cmd.Flags().StringVar(&opts.OutputDir, "output-dir", opts.OutputDir, "Directory to output rendered configurations to.")
-	cmd.Flags().StringVar(&opts.CentralRemoteUrl, "central-remote-url", opts.CentralRemoteUrl, "Git URL for the central remote, used to calculate merge-base.")
+	cmd.Flags().StringArrayVar(&opts.CentralRemoteUrl, "central-remote-url", opts.CentralRemoteUrl, "Git URL for the central remote, used to calculate merge-base. Can be specified multiple times.")
 	cmd.Flags().BoolVar(&opts.Update, "update", opts.Update, "Update the digest file.")
 
 	for _, flag := range []string{
@@ -73,7 +74,7 @@ func BindOptions(opts *RawOptions, cmd *cobra.Command) error {
 type RawOptions struct {
 	ServiceConfigFile string
 	DevSettingsFile   string
-	CentralRemoteUrl  string
+	CentralRemoteUrl  []string
 
 	DigestFile string
 	OutputDir  string
@@ -96,7 +97,7 @@ type completedOptions struct {
 	Digests       *Digests
 	DevSettings   *settings.Settings
 
-	CentralRemoteUrl  string
+	CentralRemoteUrl  []string
 	OutputDir         string
 	ServiceConfigFile string
 	DigestFile        string
@@ -259,7 +260,7 @@ func ValidateServiceConfig(
 	outputDir string,
 	update bool,
 	digestFile string,
-	centralRemoteUrl string,
+	centralRemoteUrls []string,
 ) error {
 	logger, err := logr.FromContext(ctx)
 	if err != nil {
@@ -405,7 +406,7 @@ func ValidateServiceConfig(
 							ctx,
 							cloud, environment, *regionCtx,
 							serviceConfigFile, jsonSchemaPath,
-							centralRemoteUrl, outputDir, scratchDir,
+							centralRemoteUrls, outputDir, scratchDir,
 						); err != nil {
 							logger.WithValues("cloud", cloud, "environment", environment, "region", region).Error(err, "Failed to render diff.")
 						}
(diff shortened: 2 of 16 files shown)


tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -478,60 +487,134 @@ func DetermineChangedServices(ctx context.Context, centralRemoteUrl, topologyDir
...
+		}
 	}
+	return w.changed, nil
+}
 
-	// these will be relative to the <repo-root-path>
+// New version of DetermineChangedServices that works with CombinedTopology, which can have services from multiple topology files and therefore multiple git roots.
+// The algorithm is the same, but we first determine the set of git roots across all topology files, then collect changed files from each, and compare against each service's pipeline dir as an absolute path.
+// Separate function to maintain backwards compatibilty until we can update all call sites to use the new version.
+func DetermineChangedServicesCombined(ctx context.Context, centralRemoteUrl string, topologyFiles []string, t *topology.CombinedTopology) (sets.Set[string], error) {
```
rachelvweber: sdp-pipelines calls DetermineChangedServices

tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -478,60 +487,134 @@ func DetermineChangedServices(ctx context.Context, centralRemoteUrl, topologyDir
...
 
-	serviceDir := filepath.Join(w.topologyDirRelPath, filepath.Dir(service.PipelinePath))
+	topoDir, err := w.getTopoDirForService(service.ServiceGroup)
+	if err != nil {
+		return err
+	}
+	serviceDir := filepath.Join(topoDir, filepath.Dir(service.PipelinePath))
 
 	for _, file := range w.files {
 		if strings.HasPrefix(file, serviceDir) {
```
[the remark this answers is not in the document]
rachelvweber: This is a best-effort check. I'm not looking to change the functionality of it with this PR

tooling/templatize/cmd/configuration/validate/cmd.go
```diff
@@ -26,13 +26,13 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 )
 
-func NewCommand(centralRemoteUrl string) (*cobra.Command, error) {
```
stevekuznetsov, line 29: So I think because we import and use this in sdp-pipelines, this is the bit that will require a manual update during the bump. You could try modeling it as a variadic argument to avoid the compile time error as long as the behavior is correct when only one remote is available

tooling/templatize/cmd/configuration/validate/options.go
```diff
@@ -564,7 +565,7 @@ func renderDiff(
 }
 
 // DetermineMergeBase determines the merge base between HEAD and what we think the upstream target branch is.
-func DetermineMergeBase(ctx context.Context, dir, centralRemoteUrl string) (string, error) {
+func DetermineMergeBase(ctx context.Context, dir string, centralRemoteUrls []string) (string, error) {
```
stevekuznetsov, line 568: I am confused as to why this is needed - can you add some motivation?

tooling/templatize/cmd/entrypoint/entrypointutils/load.go
```diff
@@ -24,9 +24,13 @@ import (
 )
 
 func LoadPipelines(
```
stevekuznetsov, line 26: I also wouldn't be surprised if we imported stuff like this - so a manual bump may just be required, which is ok.

tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -442,96 +451,171 @@ func handleService(logger logr.Logger, context string, group *errgroup.Group, ba
...
+	}
+	for _, service := range t.Services {
+		if err := w.walk(service); err != nil {
+			return nil, err
+		}
+	}
+	return w.changed, nil
+}
+
+// New version of DetermineChangedServices that works with CombinedTopology, which can have services from multiple topology files and therefore multiple git roots.
```
stevekuznetsov: At every point in the lifetime of this method except for this PR review, nobody will know or care that there was an old version, or that this is a new version - let's focus the GoDoc on what it does and when to use it (and remember to use proper GoDoc by starting the comment with the function name).

tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -442,96 +451,171 @@ func handleService(logger logr.Logger, context string, group *errgroup.Group, ba
...
+// The algorithm is the same, but we first determine the set of git roots across all topology files, then collect changed files from each, and compare against each service's pipeline dir as an absolute path.
```
stevekuznetsov: Nit: the rest of the file has manual line-breaks in the comments, let's try to match existing style

tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -442,96 +451,171 @@ func handleService(logger logr.Logger, context string, group *errgroup.Group, ba
...
+// Separate function to maintain backwards compatibility until we can update all call sites to use the new version.
```
stevekuznetsov: Temporally-limited comments like this age poorly - if you forget to update, or we end up needing it forever, this just ends up stale. Omit it

tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -442,96 +451,171 @@ func handleService(logger logr.Logger, context string, group *errgroup.Group, ba
...
+func DetermineChangedServicesCombined(ctx context.Context, centralRemoteUrls []string, topologyFiles []string, t *topology.CombinedTopology) (sets.Set[string], error) {
+	// Map each unique topology dir to its git root, deduplicating.
```
stevekuznetsov: Prefer not to leave "what we're doing" comments - the code is sufficient.

tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -442,96 +451,171 @@ func handleService(logger logr.Logger, context string, group *errgroup.Group, ba
...
+		gitRoots.Insert(gitRoot)
 	}
 
-	// these will be relative to the <repo-root-path>
+	// For each git root, collect changed files as absolute paths.
 	var files []string
-	{
-		cmd := exec.CommandContext(ctx, "git", "diff", "--name-only", fmt.Sprintf("%s..HEAD", mergeBase))
-		out, err := cmd.Output()
+	for _, gitRoot := range sets.List(gitRoots) {
```
stevekuznetsov: Unless you need a deterministic iteration order, you don't need the allocations from `sets.List`, and you can just iterate through `gitRoots` directly

tooling/templatize/cmd/configuration/validate/options.go
rachelvweber: Why what is needed? Like you mean people won't want to use DetermineChangedServices with topologies that live in different repos?

tooling/templatize/cmd/configuration/validate/cmd.go
rachelvweber: Should be backwards compatible now

tooling/templatize/cmd/entrypoint/entrypointutils/load.go
rachelvweber: I pointed my sdp-pipelines at this version of ARO-HCP, and it compiles fine. I think we shouldn't need a manual bump

tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -442,96 +449,171 @@ func handleService(logger logr.Logger, context string, group *errgroup.Group, ba
...
+		}
 	}
 	return w.changed, nil
 }
 
+func getGitRoot(ctx context.Context, path string) (string, error) {
+	cmd := exec.CommandContext(ctx, "git", "-C", path, "rev-parse", "--show-toplevel")
+	out, err := cmd.Output()
+	if err != nil {
+		return "", fmt.Errorf("failed to run git rev-parse in %s: %w", path, err)
```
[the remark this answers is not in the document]
rachelvweber: I'm not looking to change the logic of this as part of this PR. Just moving the existing code into a separate function for now

tooling/templatize/cmd/pipeline/validate/options.go
```diff
@@ -442,96 +449,171 @@ func handleService(logger logr.Logger, context string, group *errgroup.Group, ba
...
+	}
+	return strings.TrimSpace(string(out)), nil
+}
+
+func getGitDiff(ctx context.Context, path string, mergeBase string) ([]string, error) {
+	cmd := exec.CommandContext(ctx, "git", "-C", path, "diff", "--name-only", fmt.Sprintf("%s..HEAD", mergeBase))
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, fmt.Errorf("failed to run git diff in %s: %w; output: %s", path, err, string(out))
+	}
```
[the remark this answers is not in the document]
rachelvweber: I'm not looking to change the logic of this as part of this PR. Just moving the existing code into a separate function for now

tooling/templatize/cmd/configuration/validate/options.go
stevekuznetsov: It wasn't clear until way later into the diff why config validation needed to know about multiple central remotes

PR #5113, opened 2026-05-04
update deletion controller to handle removal of old resourceID

deads2k:


backend/pkg/controllers/operationcontrollers/utils.go
```diff
@@ -689,11 +688,15 @@ func SetDeleteOperationAsCompleted(ctx context.Context, cosmosClient database.DB
 			continue
 		}
 
-		resourceInfo := database.ResourceDocument{}
-		if err := json.Unmarshal(childResource.Properties, &resourceInfo); err != nil {
-			return utils.TrackError(err)
+		// Use childResource.ResourceID directly. Earlier code parsed childResource.Properties
```
miguelsorianod, line 691: Does that mean that as of now (before this PR) our delete operations cannot be completed because of the deletion of child resources not working because of not being able to get a correct resource id?

deads2k: > Does that mean that as of now (before this PR) our delete operations cannot be completed because of the deletion of child resources not working because of not being able to get a correct resource id?

Current state, all the types include a ResourceID field in the right spot to deserialize into ResourceDocument.  https://github.com/Azure/ARO-HCP/pull/4157 (not merged) changes that and is how I found this.  So right now, things delete ok.

miguelsorianod: Answering myself here:

As of now (before this PR), deletion works because all resources have a ResourceID field in properties' top level. This code comment just assumes that https://github.com/Azure/ARO-HCP/pull/4157 is done and merged. 4157 is not merged yet but this PR already prepares the code comment for when that occurs

PR #5114, opened 2026-05-04
Adding Microsoft.Quota Resource Provider

rachelvweber:
### What

Adding Microsoft.Quota Resource Provider to Service and Management Clusters

### Why

This resource provider is required to use the Group Quota API to automatically request quota as demand increases.

### Testing

Tested by e2e parallel PR gate


weinong (approved):
LGTM

PR #5119, opened 2026-05-05
Extract shared Cincinnati ClientCache from upgrade controllers

ahitacat:
[ARO-25620](https://redhat.atlassian.net/browse/ARO-25620)

### What

Refactors Cincinnati client caching by extracting shared logic into a reusable `ClientCache` interface and fixes the package name typo (`cincinatti` → `cincinnati`).

Changes:
- Extract `ClientCache` interface and `CachingClient` implementation from duplicated controller code into `internal/cincinnati/cache.go`
- Rename package from `cincinatti` to `cincinnati` (fixing the typo)
- Rename `GetClient` to `GetOrCreateClient` to clarify the method creates on cache miss
- Add scenario-based cache tests following 

### Why

Both `control_plane_desired_version_controller.go` and `nodepool_version_controller.go` had duplicated Cincinnati client caching logic. Extracting this into a shared component:
- Reduces code duplication
- Makes the caching behavior testable in isolation
- Provides a mockable interface for controller unit tests

### Testing
Added unit tests in `internal/cincinnati/cache_test.go`

This is a refactoring PR with no behavioral changes. The cache logic extracted is identical to what was previously inline in the controllers.


internal/cincinnati/cache.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// GetOrCreateClient returns a Cincinnati client for the given cluster, using a cache.
+// The clusterUUID is the external ID of the parent cluster from Cluster Service.
+func (c *CachingClient) GetOrCreateClient(key controllerutils.HCPClusterKey, clusterUUID uuid.UUID) Client {
+	// Fast path: check cache with read lock
+	c.lock.RLock()
+	client, ok := c.cache.Get(key)
+	c.lock.RUnlock()
+
```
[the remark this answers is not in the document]
ahitacat: lru cache is thread safe https://github.com/kubernetes/utils/blob/master/lru/lru.go#L28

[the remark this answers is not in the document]
machi1990: The one being used here is `github.com/golang/groupcache/lru` can you check if that's thread safe?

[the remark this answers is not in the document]
ahitacat: My bad, don't know why I thought is used the same as your PR. 
It is not thread safe as per: https://github.com/golang/groupcache/blob/master/lru/lru.go#L22

Not sure if it is best to address this concern by adding what Copilot suggest or change to use the one for kubernetes that we use in other places of this project.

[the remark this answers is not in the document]
machi1990: >  Consider using an exclusive Lock() for all Get/Add operations (and switching to sync.Mutex), or replacing the cache with an implementation that supports concurrent reads safely.

This suggestion over here is essentially fixed by using the k8s one and leave the thread safety concerns to that lib.

internal/cincinnati/cache_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	uuid1 := uuid.New()
+	uuid2 := uuid.New()
+
+	client1 := cache.GetOrCreateClient(uuid1)
+	client2 := cache.GetOrCreateClient(uuid2)
+
+	if client1 == client2 {
+		t.Error("Expected different client instances for different UUIDs")
+	}
```
machi1990: can use assert.Equal(

internal/cincinnati/cache_test.go
```diff
@@ -0,0 +1,49 @@
...
+	clusterUUID := uuid.New()
+
+	// First call creates client
+	client1 := cache.GetOrCreateClient(clusterUUID)
+	// Second call should return cached client
+	client2 := cache.GetOrCreateClient(clusterUUID)
+
+	if client1 != client2 {
+		t.Error("Expected same client instance on cache hit")
+	}
```
machi1990: can use assert.Equal

internal/cincinnati/cache.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// ClientCache provides cached Client instances keyed by cluster UUID.
+type ClientCache interface {
+	GetOrCreateClient(clusterUUID uuid.UUID) Client
+}
+
+// CachingClient implements ClientCache with LRU caching.
+// Uses k8s.io/utils/lru which is thread-safe.
+type CachingClient struct {
```
machi1990: CincinnatiClientCache

machi1990 (changes requested):
minor comments otherwise lgtm; will tag once addressed

internal/cincinnati/cache.go
```diff
@@ -0,0 +1,64 @@
...
+// The clusterUUID is the external ID of the cluster from Cluster Service.
+// The k8s.io/utils/lru cache is thread-safe.
+func (c *CachingClient) GetOrCreateClient(clusterUUID uuid.UUID) Client {
+	if client, ok := c.cache.Get(clusterUUID); ok {
+		return client.(Client)
+	}
+
+	newClient := cincinnati.NewClient(clusterUUID, c.transport, c.userAgent, NewAlwaysConditionRegistry())
+	c.cache.Add(clusterUUID, newClient)
+	return newClient
```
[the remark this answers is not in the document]
machi1990: Has this been addressed? @ahitacat

internal/cincinnati/cache.go
```diff
@@ -0,0 +1,68 @@
...
+		transport: http.DefaultTransport.(*http.Transport).Clone(),
+		userAgent: "ARO-HCP",
+	}
+}
+
+// GetOrCreateClient returns a Cincinnati client for the given cluster UUID, using a cache.
+// The clusterUUID is the external ID of the cluster from Cluster Service.
+func (c *CachingClient) GetOrCreateClient(clusterUUID uuid.UUID) Client {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
machi1990, line 65: This lock doesn't need to be a global one;

can done somewhere around line 65 when we've a cache miss;

We do a get first before creating and adding..

internal/cincinnati/cache.go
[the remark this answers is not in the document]
ahitacat: yes, using a mutex to lock the Add()

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
```diff
@@ -74,7 +68,7 @@ func NewNodePoolVersionController(
 		clusterManagementClusterContentLister: clusterManagementClusterContentLister,
 		cosmosClient:                          cosmosClient,
 		clusterServiceClient:                  clusterServiceClient,
-		clusterToCincinnatiClient:             lru.New(100000),
+		cincinnatiClientCache:                 cincinnati.NewCachingClient(),
```
nimrodshn: nit: `cincinnatiClientCache` -> `cincinnatiCachedClient` makes more sense to me.

internal/cincinnati/cache.go
nimrodshn: I prefer `CachedClient`

client cache to me sounds like a cache for some client which this is not..

machi1990: it is a cache of the cincinnati client...anyway naming :-D

machi1990 (approved):
/lgtm

Thank you for this.

We could in the future evaluate the possibility of also caching with a TTL the cincinatti results for a given version and channel. Can you capture that piece of work somewhere for future improvements?

ahitacat:
/hold cancel
There was a conflict with main, I have rebased.

PR #5120, opened 2026-05-05
feat: move cluster creation from frontend to backend

miguelsorianod:
This commit moves cluster creation from frontend to backend. Frontend now just creates the corresponding ARM Operation, with an empty internalID, and it delegates all the actual cluster creation work to backend.

On backend now we check whether internalID is empty or not to assess whether the cluster create request to Clusters Service has been submitted. Once that occurs, we check the state.

The logic that performs cluster creation against Clusters Service has been moved to a backend operationcontroller named DispatchClusterCreate. The controller attempts to be resilient to failures.

This MR cannot be merged until cluster reads have been moved to backend (https://github.com/Azure/ARO-HCP/pull/4610)

### DIFF
--- a/backend/pkg/app/backend.go
+++ b/backend/pkg/app/backend.go
@@ -327,6 +327,11 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context, 
 		b.options.ClustersServiceClient,
 		activeOperationInformer,
 	)
+	dispatchClusterCreateController := operationcontrollers.NewDispatchClusterCreateController(
+		b.options.CosmosDBClient,
+		b.options.ClustersServiceClient,
+		activeOperationInformer,
+	)
 	operationClusterCreateController := operationcontrollers.NewOperationClusterCreateController(
 		b.options.CosmosDBClient,
 		b.options.ClustersServiceClient,
@@ -545,6 +550,7 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context, 
 				go doNothingController.Run(ctx, 20)
 				go dispatchRequestCredentialController.Run(ctx, 20)
 				go dispatchRevokeCredentialsController.Run(ctx, 20)
+				go dispatchClusterCreateController.Run(ctx, 20)
 				go operationClusterCreateController.Run(ctx, 20)
 				go operationClusterUpdateController.Run(ctx, 20)
 				go operationClusterDeleteController.Run(ctx, 20)
--- a/backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
+++ b/backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
@@ -1,0 +1,241 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package operationcontrollers
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+type dispatchClusterCreate struct {
+	cosmosClient          database.DBClient
+	clustersServiceClient ocm.ClusterServiceClientSpec
+}
+
+func NewDispatchClusterCreateController(
+	cosmosClient database.DBClient,
+	clustersServiceClient ocm.ClusterServiceClientSpec,
+	activeOperationInformer cache.SharedIndexInformer,
+) controllerutils.Controller {
+	syncer := &dispatchClusterCreate{
+		cosmosClient:          cosmosClient,
+		clustersServiceClient: clustersServiceClient,
+	}
+
+	return NewGenericOperationController(
+		"DispatchClusterCreate",
+		syncer,
+		10*time.Second,
+		activeOperationInformer,
+		cosmosClient,
+	)
+}
+
+func (c *dispatchClusterCreate) ShouldProcess(ctx context.Context, operation *api.Operation) bool {
+	if operation.Status.IsTerminal() {
+		return false
+	}
+	if operation.Request != database.OperationRequestCreate {
+		return false
+	}
+	if operation.ExternalID == nil || !strings.EqualFold(operation.ExternalID.ResourceType.String(), api.ClusterResourceType.String()) {
+		return false
+	}
+	if len(operation.InternalID.String()) > 0 {
+		return false
+	}
+	return true
+}
+
+func (c *dispatchClusterCreate) SynchronizeOperation(ctx context.Context, key controllerutils.OperationKey) error {
+	logger := utils.LoggerFromContext(ctx)
+	logger.Info("checking operation")
+
+	operation, err := c.cosmosClient.Operations(key.SubscriptionID).Get(ctx, key.OperationName)
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get active operation: %w", err)
+	}
+	if !c.ShouldProcess(ctx, operation) {
+		return nil
+	}
+
+	cluster, err := c.cosmosClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).Get(ctx, operation.ExternalID.Name)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if cluster.ServiceProviderProperties.ActiveOperationID != "" &&
+		cluster.ServiceProviderProperties.ActiveOperationID != operation.OperationID.Name {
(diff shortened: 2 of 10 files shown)


frontend/pkg/frontend/cluster.go
```diff
@@ -343,44 +342,13 @@ func (f *Frontend) createHCPCluster(writer http.ResponseWriter, request *http.Re
 	// TODO this is bad, see above TODOs. We want to validate what we store.
 	newInternalCluster.Identity.UserAssignedIdentities = nil
 
-	var tenantID string
-	if subscription.Properties != nil && subscription.Properties.TenantId != nil {
-		tenantID = *subscription.Properties.TenantId
-	}
-
-	initialClusterProperties := map[string]string{}
-	if len(f.clusterServiceProvisionShard) != 0 {
```
miguelsorianod, line 352: The functionality of the CSPropertyProvisionShardID, CSPropertyNoopProvision, CSPropertyNoopDeprovision was added in the beginning of the project. This PR removes the ability to set that completely. No permanent aro-hcp environments leverage it, the helm template of the frontend deployment does not even pass the corresponding frontend CLI flags. Technically some developer could be using it. If for some reason we want to keep that we would need to move it to backend.

frontend/cmd/cmd.go
```diff
@@ -101,9 +98,6 @@ func NewRootCmd() *cobra.Command {
 
 	rootCmd.Flags().StringVar(&opts.clustersServiceURL, "clusters-service-url", "https://api.openshift.com", "URL of the OCM API gateway.")
 	rootCmd.Flags().BoolVar(&opts.insecure, "insecure", false, "Skip validating TLS for clusters-service.")
-	rootCmd.Flags().StringVar(&opts.clusterServiceProvisionShard, "cluster-service-provision-shard", "", "Manually specify provision shard for all requests to cluster service")
```
miguelsorianod, line 104: Related to https://github.com/Azure/ARO-HCP/pull/5120/changes#r3187900975. If we decide we don't want to support that anymore the CLI flags can be removed safely from frontend because the helm deployment doesn't pass them.

backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
```diff
@@ -0,0 +1,241 @@
...
+	if subscription.Properties == nil || subscription.Properties.TenantId == nil || *subscription.Properties.TenantId == "" {
+		return utils.TrackError(fmt.Errorf("subscription %s has no tenant id", operation.ExternalID.SubscriptionID))
+	}
+	tenantID := *subscription.Properties.TenantId
+
+	mrg := cluster.CustomerProperties.Platform.ManagedResourceGroup
+	if mrg == "" {
+		return utils.TrackError(fmt.Errorf("cluster %s has no managed resource group", cluster.Name))
+	}
+	existing, err := c.findAROHCPClusterByAzureInfo(ctx,
```
miguelsorianod, line 133: A challenge I found was that it could occur that inbetween the CS create request is received and we store the CS ID in the Cluster doc something fails. Then the challenge is: what do we do at that point? to interact with the CS API you leverage the CS Cluster ID.
What this PR does is to do a list in CS with a search string using the cluster's azure resource information: the azure subscription, tenant id, resource group name, and resource name. This assumes that no two clusters will be created with that information.

**Although the PR uses that, CS currently does not support filtering by some of those fields, we would need to add it. Currently CS list only supports filtering by subscription id and resource group"**

backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
```diff
@@ -0,0 +1,241 @@
...
+	if !c.ShouldProcess(ctx, operation) {
+		return nil
+	}
+
+	cluster, err := c.cosmosClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).Get(ctx, operation.ExternalID.Name)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if cluster.ServiceProviderProperties.ActiveOperationID != "" &&
```
miguelsorianod, line 93: is this check needed and returning a non error the correct approach?

backend/pkg/app/backend.go
```diff
@@ -327,6 +327,11 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
 		b.options.ClustersServiceClient,
 		activeOperationInformer,
 	)
+	dispatchClusterCreateController := operationcontrollers.NewDispatchClusterCreateController(
```
miguelsorianod, line 330: Although the PR implements create, I wonder whether we need to change something somewhere else. What happens if a delete occurs and the CS ID hasn't been stored for some reason?

backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
```diff
@@ -0,0 +1,241 @@
...
+	// Tenant ID and managed resource group are not lowercased in the OCM CS
+	// builder (see withImmutableAttributes in convert.go)), we keep the casing as it is.
+	wantTenant := tenantID
+	wantMRG := managedResourceGroupName
+	search := c.clustersServiceClusterByAzureInfoSearchString(wantSub, wantRG, wantName, wantTenant, wantMRG)
+	matches, err := c.csClustersMatchingClusterByAzureInfo(ctx, c.clustersServiceClient.ListClusters(search), wantSub, wantRG, wantName, wantTenant, wantMRG)
+	if err != nil {
+		return nil, err
+	}
+	if len(matches) > 1 {
```
miguelsorianod, line 200: If we get more than one result we treat it as an error

backend/pkg/controllers/operationcontrollers/operation_cluster_create.go
```diff
@@ -107,12 +110,8 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
 		return nil // no work to do
 	}
 
-	if len(operation.InternalID.String()) == 0 {
-		// we cannot proceed: yet.
-		// TODO when we update to make clusterserice creation async, we need https://github.com/Azure/ARO-HCP/pull/4695 or similar
-		// and we need to wire up a fail-safe where if we have no ID and we time out, we report the best failure we can.
-		return nil
-	}
+	// TODO we need to wire up a fail-safe where if we have no ID and we time out, we report the best failure we can.
```
miguelsorianod, line 113: See.

Additionally, what happens if for some reason the creation has failed in the CS side? and what if for some reason it's not filled in Cosmos in a timely manner?

frontend/pkg/frontend/cluster.go
```diff
@@ -343,44 +342,13 @@ func (f *Frontend) createHCPCluster(writer http.ResponseWriter, request *http.Re
...
-	newInternalCluster.ServiceProviderProperties.ClusterServiceID = &csID
-
 	transaction := f.dbClient.NewTransaction(newInternalCluster.ID.SubscriptionID)
 
 	// TODO extract to straight instance creation and then validation.
 	clusterCreateOperation := database.NewOperation(
 		database.OperationRequestCreate,
 		newInternalCluster.ID,
-		ptr.Deref(newInternalCluster.ServiceProviderProperties.ClusterServiceID, api.InternalID{}),
+		api.InternalID{},
```
miguelsorianod, line 351: In frontend we now pass an empty internal ID as now the responsibility is fully on backend

frontend/pkg/frontend/cluster.go
```diff
@@ -419,6 +387,7 @@ func (f *Frontend) createHCPCluster(writer http.ResponseWriter, request *http.Re
 		return fmt.Errorf("unexpected type %T", resultingUncastInternalCluster)
 	}
 
+	var resultingClusterServiceCluster *arohcpv1alpha1.Cluster // TODO remove this once we moved read from CS
```
miguelsorianod, line 390: temporary to silence the compiler. This depensd on the MR that moves reads out of frontend

backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	subscription, err := c.cosmosClient.Subscriptions().Get(ctx, operation.ExternalID.SubscriptionID)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	if subscription.Properties == nil || subscription.Properties.TenantId == nil || *subscription.Properties.TenantId == "" {
+		return utils.TrackError(fmt.Errorf("subscription %s has no tenant id", operation.ExternalID.SubscriptionID))
+	}
+	tenantID := *subscription.Properties.TenantId
```
miguelsorianod, line 127: We get the tenant id from the cluster's subscription. The operation itself has a tenant but on cluster creation I observed the operation id tenant comes from a http header whereas the cluster tenant comes from the subscription. Because of this I retrieved the tenant id using the cluster's subscription (requires one db query more as we need to retrieve the susbcription)

backend/pkg/controllers/operationcontrollers/dispatch_cluster_create.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+type dispatchClusterCreate struct {
+	cosmosClient          database.DBClient
+	clustersServiceClient ocm.ClusterServiceClientSpec
+}
+
+func NewDispatchClusterCreateController(
```
miguelsorianod, line 38: This dispatches CS cluster creation in a controller. At some point I am guessing we might need to do some resources precreation or calculation in RP itself before sending the call to CS. In that case, how would we coordinate this controller and the others?

machi1990:
@miguelsorianod , @JakobGray had created https://github.com/Azure/ARO-HCP/pull/4821 does this PR supersedes it?

miguelsorianod:
> @miguelsorianod , @JakobGray had created #4821 does this PR supersedes it?

I wasn't aware of that one. It's possible there's some overlap.

backend/pkg/app/backend.go
```diff
@@ -327,6 +327,11 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
 		b.options.ClustersServiceClient,
 		activeOperationInformer,
 	)
```
miguelsorianod, line 329: The introduction of the changes here need to be actually split into two:

2. Then introduce the frontend removal logic

I think it should be safe to introduce 1 because it will only be executed when the operation exists and it doesn't have an internal id, and until 2 is introduced the cluster create operation should always have an internal id as the operation is created together with it set

miguelsorianod:
Superseded by #6082 and #6121. Closing.

PR #5121, opened 2026-05-05
feat: ensure maestro readonly bundle is recreated if it is deleted externally

miguelsorianod:
We update the cluster-scoped and nodepool-scoped maestro readonly bundle creation controllers to ensure the bundle is recreated if it has been deleted in the Maestro API side directly for some reason.

The controllers have also been updated to return an error if the retrieved bundle from the Maestro API has a different UID and it does not have the aro-hcp.azure.com/readonly-bundle-managed-by K8s label to the corresponding scope of the controller. This ensures that we don't start considering a bundle created by something or someone else that happens to have the same maestro bundle name.

### DIFF
--- a/backend/pkg/controllers/create_cluster_scoped_maestro_readonly_bundles_controller.go
+++ b/backend/pkg/controllers/create_cluster_scoped_maestro_readonly_bundles_controller.go
@@ -124,30 +124,9 @@ func (c *createClusterScopedMaestroReadonlyBundlesSyncer) SyncOnce(ctx context.C
 		api.MaestroBundleInternalNameReadonlyHypershiftHostedCluster,
 	}
 
-	var maestroBundlesToSync []api.MaestroBundleInternalName
-	// We first check if there's any recognized Maestro Bundle reference that needs to be synced.
-	for _, maestroBundleInternalName := range recognizedMaestroBundles {
-		currentMaestroBundleReference, err := existingServiceProviderCluster.Status.MaestroReadonlyBundles.Get(maestroBundleInternalName)
-		if err != nil {
-			return utils.TrackError(fmt.Errorf("failed to get Maestro Bundle reference: %w", err))
-		}
-
-		if currentMaestroBundleReference == nil {
-			maestroBundlesToSync = append(maestroBundlesToSync, maestroBundleInternalName)
-			continue
-		}
-		if len(currentMaestroBundleReference.MaestroAPIMaestroBundleName) == 0 {
-			maestroBundlesToSync = append(maestroBundlesToSync, maestroBundleInternalName)
-			continue
-		}
-		if len(currentMaestroBundleReference.MaestroAPIMaestroBundleID) == 0 {
-			maestroBundlesToSync = append(maestroBundlesToSync, maestroBundleInternalName)
-			continue
-		}
-	}
-	if len(maestroBundlesToSync) == 0 {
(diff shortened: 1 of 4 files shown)


miguelsorianod:
The main tradeoff of adding this is that now we need to interact with Maestro to see if the bundle actually exists. This means each time the controller is triggered (changes on clusters and whatever ends up triggering SyncOnce calls) all the readonly bundles associated to the resource are checked. This puts more load on Maestro.

backend/pkg/controllers/create_cluster_scoped_maestro_readonly_bundles_controller.go
```diff
@@ -267,8 +246,15 @@ func (c *createClusterScopedMaestroReadonlyBundlesSyncer) syncMaestroBundle(
 		return lastPersistedSPC, utils.TrackError(fmt.Errorf("failed to get or create Maestro Bundle: %w", err))
 	}
 
-	// If the Maestro API MaestroBundle ID is not set we store the returned Maestro Bundle ID in the corresponding Maestro Bundle reference of the ServiceProviderCluster in Cosmos.
-	if len(existingMaestroBundleRef.MaestroAPIMaestroBundleID) == 0 {
+	// If the stored Maestro Bundle ID differs from the one returned by the Maestro API, update it.
+	// This handles initial creation (empty -> new UID) and recreation after external deletion (old UID -> new UID).
+	if existingMaestroBundleRef.MaestroAPIMaestroBundleID != string(resultMaestroBundle.UID) {
+		// If the Maestro Bundle we retrieved from the Maestro API is not managed by the controller, we return an error.
+		if resultMaestroBundle.Labels[readonlyBundleManagedByK8sLabelKey] != readonlyBundleManagedByK8sLabelValueClusterScoped {
```
miguelsorianod, line 253: We also add a check that if a maestro bundle with the same name exists with a different maestro bundle id and it doesn't have the controller managed by label of this resource then we permanently error. This could technically happen if for some reason someone creates a maestro bundle with the same maestro bundle name than what this controller created originally and it does so without setting the k8s label. If for some reason someone where to create a new maestro bundle with the same name and adds the k8s label then this controller assumes it as if it was the one that created it

backend/pkg/controllers/create_cluster_scoped_maestro_readonly_bundles_controller.go
```diff
@@ -124,30 +124,9 @@ func (c *createClusterScopedMaestroReadonlyBundlesSyncer) SyncOnce(ctx context.C
 		api.MaestroBundleInternalNameReadonlyHypershiftHostedCluster,
 	}
```
miguelsorianod, line 126: I think another implication of this change is that now the create controllers and the delete orphan controllers could race and we could have a creation after the orphan deleter has deleted the bundle from the maestro api. If that occurs a next reconcile in the orphan deleter would delete it again eventually and the create one would stop attempting to do so because the Cosmos cluster would stop to exist.

Is that assessment correct, and if so, is that approach what we want to occur?

deads2k: I won't stress out over it.

deads2k:
/lgtm

you can decide about your hold

miguelsorianod:
Closing as we are not creating maestro readonly bundles anymore

PR #5122, opened 2026-05-05
test: fix version resolution and improve reliability of fetching OCP versions

mvacula02:
[ARO-26732](https://redhat.atlassian.net/browse/ARO-26732)
[ARO-26733](https://redhat.atlassian.net/browse/ARO-26733)

### What

- Control plane version is cached with `sync.Once` and resolved only once per process 
- `DefaultOpenshiftNodePoolVersionId()` now reuses the cached control plane version when channel group matches the cluster's
- Add generic retry helper with exponential back-off - `retryOnTransientError()` 
- Enable the retry for `GetLatestInstallVersion()`, `GetAllVersionsInMinorStartingWith()` and `GetUpgradeCandidatesInMaxMinorFromCincinnati()`
- Make `GetLatestInstallVersionForGraphChannel()` and `GetLatestInstallVersionForNightlyChannel()` private. To be called only through the retry wrapper
- Propagate context to nightly channel function for context cancellation support

### Why

This is a fix for 2 bugs:

1. Version race condition: NewDefaultClusterParams() and NewDefaultNodePoolParams() each independently query Cincinnati for the latest OCP version. Since cluster creation can take more than 30 minutes, a new candidate version can be published between the two queries, causing node pool creation to fail.
2. Transient HTTP failures: The second, and in this case redundant, Cincinnati query can fail due to various network errors, failing the test even though the first query succeeded. 

### Testing

PR modifies framework helper functions used broadly across tests.

### DIFF
--- a/test/util/framework/deployment_params.go
+++ b/test/util/framework/deployment_params.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


test/util/framework/deployment_params.go
```diff
@@ -117,6 +136,9 @@ func DefaultOpenshiftNodePoolVersionId() string {
 	version := os.Getenv("ARO_HCP_OPENSHIFT_NODEPOOL_VERSION")
 	if len(version) == 0 {
 		channelGroup := DefaultOpenshiftNodePoolChannelGroup()
+		if channelGroup == DefaultOpenshiftChannelGroup() {
+			return DefaultOpenshiftControlPlaneVersionId()
+		}
```
mbukatov, line 141: What is this check doing?

mvacula02: when the node pool and control plane have matching channel group, return cached version instead of making a new Cincinnati query. `DefaultOpenShiftControlPlaneVersionId` doesn't actually make a new call, but uses the value cached from previous call due to the `sync.Once`

miquelsi (approved):
/approve
/lgtm

PR #5115, opened 2026-05-04
initial commit of pko orphaned resources cleanup

sclarkso:
<!-- Link to Jira issue -->

### What

Manually cleans up orphaned PKO resources

- Discover all CRDs in the package-operator.run API group (including subgroups like manifests.package-operator.run), extracting plural name, group, and scope from each CRD spec                                                                                           
- Delete all CRs for each CRD, handling cluster-scoped and namespaced resources appropriately
- Poll up to 3 minutes for cascading deletion (via blockOwnerDeletion: true owner refs) to clean up child resources                                                                                                                                                        
- If any CRs are stuck, force-remove their finalizers — the PKO operator is already uninstalled so nothing will process them                                                                                                                                               
- Once all CRs are gone, delete the CRDs themselves                                                                                                                                                                                                                        
- Idempotent and safe to run on clusters that never had PKO installed                                                                                                                                                                                                      
- Wired into mgmt-pipeline.yaml as a new cleanup-pko-resources step, running independently of the existing cleanup-pko step

### Why

Orphaned resources are blocking deletion of clusters that were provisioned while PKO was installed.

### Testing

Wrote a dry-run version of the script that only echo'd and didn't delete and dry-ran it against INT cluster which has orphaned resources:

```
=== Package Operator cleanup — DRY RUN ===

CRDs that would be deleted:
  clusterobjectdeployments.package-operator.run (Cluster)
  clusterobjectsetphases.package-operator.run (Cluster)
  clusterobjectsets.package-operator.run (Cluster)
  clusterobjectslices.package-operator.run (Cluster)
  clusterobjecttemplates.package-operator.run (Cluster)
  clusterpackages.package-operator.run (Cluster)
  objectdeployments.package-operator.run (Namespaced)
  objectsetphases.package-operator.run (Namespaced)
  objectsets.package-operator.run (Namespaced)
  objectslices.package-operator.run (Namespaced)
  objecttemplates.package-operator.run (Namespaced)
  packages.package-operator.run (Namespaced)

--- clusterobjectdeployments.package-operator.run (Cluster) ---
  package-operator  finalizers=

--- clusterobjectsetphases.package-operator.run (Cluster) ---
  (none)

--- clusterobjectsets.package-operator.run (Cluster) ---
  package-operator-7798f7865  finalizers=["package-operator.run/cached","package-operator.run/metrics"]

--- clusterobjectslices.package-operator.run (Cluster) ---
  (none)

--- clusterobjecttemplates.package-operator.run (Cluster) ---
[...]

PR #5109, opened 2026-05-04
Upgrade to AKS 1.35

tmstff:
### What

Upgrade to AKS version 1.35

### Why

To reduce the gap to HyperShift - who test on 1.35

### Testing

- taken care of by the e2e tests

PR #5123, opened 2026-05-05
dev-infrastructure: fix template crash for optional serviceKeyVault.subscription

raelga:
https://redhat.atlassian.net/browse/AROSLSRE-773

### What

Use `index` function instead of direct map access for the optional `serviceKeyVault.subscription` key in `svc-pipeline.yaml`.

### Why

The INT bumper is blocked. Commit 68e4da8c4 added:

```yaml
subscription: '{{ if .serviceKeyVault.subscription.key }}...'
```

Go templates with `missingkey=error` crash on `.serviceKeyVault.subscription.key` when `subscription` doesn't exist as a map key — even inside an `if`. The `serviceKeyVault.subscription` key only exists in the dev cloud config overlay, not in public (INT/STG/PROD).

### Fix

```yaml
# Before (crashes when subscription key missing)
subscription: '{{ if .serviceKeyVault.subscription.key }}{{ .serviceKeyVault.subscription.key }}{{ else }}{{ .svc.subscription.key }}{{ end }}'

# After (index returns nil for missing key, with treats nil as false)
subscription: '{{ with index .serviceKeyVault "subscription" }}{{ .key }}{{ else }}{{ $.svc.subscription.key }}{{ end }}'
```

`index` is a function call — `missingkey=error` only applies to dot-access, not to `index`. Returns nil for missing keys, `with` treats nil as false, falls through to `$.svc.subscription.key`.

### Testing

- Template-only change, no functional impact when key exists (dev cloud)
- Unblocks INT/STG/PROD manifest generation where key is absent

### DIFF
--- a/dev-infrastructure/svc-pipeline.yaml
+++ b/dev-infrastructure/svc-pipeline.yaml
@@ -62,7 +62,7 @@ resourceGroups:
     outputOnly: true
 - name: svcKV
   resourceGroup: '{{ .serviceKeyVault.rg }}'
-  subscription: '{{ if .serviceKeyVault.subscription.key }}{{ .serviceKeyVault.subscription.key }}{{ else }}{{ .svc.subscription.key }}{{ end }}'
+  subscription: '{{ with index .serviceKeyVault "subscription" }}{{ .key }}{{ else }}{{ $.svc.subscription.key }}{{ end }}'
   steps:
   - name: output
     action: ARM


raelga:
PR #5123 needs a force merge — it's blocking the INT bumper ([AROSLSRE-773](https://redhat.atlassian.net/browse/AROSLSRE-773)).

The PR's own E2E already passed:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/5123/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2051627087999012864

Tide is retesting because the base SHA moved after the batch merge of #5102, #5109, #5111, #5114, #5115 — but the change is a single-line template fix in `svc-pipeline.yaml` (no code, no config, no functional change in E2E). The retest won't validate anything new.

Without this fix, no INT deployments can proceed — EV2 manifest generation crashes on `{{ .serviceKeyVault.subscription.key }}` for all public configs.

stevekuznetsov:
You can't do logic like that in an Ev2 pipeline

/close

PR #5125, opened 2026-05-05
feat: move nodepool creation from frontend to backend

miguelsorianod:
Note: this MR showcases the end-state, but it can't be merged. different set of PRs will need to be created to roll this out appropriately. https://github.com/Azure/ARO-HCP/pull/5125#issuecomment-4622800073 for how I think it can be done.

Moves node pool Cluster Service creation out of the frontend and into the backend, and updates integration tests so frontend-only runs still work.
1. **Backend: async node pool CS create (initially gated)** (`3da7b4c60`)
   - Adds `NodePoolClusterServiceCreateController` to POST/adopt the node pool in Cluster Service and persist `clusterServiceID` on the Cosmos document.
   - **Updates the node pool create operation to read `clusterServiceID` from the node pool resource instead of the operation document.** This is a change that is not gated but that is needed already to stop relying on the operation internal id and it shouldn't have an impact.
   - Controller only runs when `clusterServiceID` is missing (safe while frontend still sets it during rollout).
2. **Frontend: stop CS create on node pool PUT** (`1ddc67f19`)
   - Removes `PostNodePool` and related logic from the frontend create path; create writes Cosmos + operation only.
   - Node pool create operations no longer carry a Cluster Service `InternalID` on the operation.
3. **Integration tests: `setClusterServiceID` step** (`894a23cbd`)
   - Frontend integration tests do not run the backend controller, so update/delete would fail without a persisted `clusterServiceID`.
   - Adds a `setClusterServiceID` databasemutation step (clusters, node pools, external auths) with backend-aligned derivation and optional `cluster-service-id.json`.
   - Wires affected FrontendCRUD artifacts and cross-version roundtrip tests; adds `ClusterServiceMock.FindClusterHREF`.


backend/pkg/controllers/operationcontrollers/dispatch_node_pool_create.go
```diff
@@ -0,0 +1,185 @@
...
+	ocmerrors "github.com/openshift-online/ocm-sdk-go/errors"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
```
miguelsorianod: The introduction of the changes here need to be actually split into two:

    First introduce the backend controller. It should never execute because as soon as
    Then introduce the frontend removal logic

I think it should be safe to introduce 1 because it will only be executed when the operation exists and it doesn't have an internal id, and until 2 is introduced the nodepool create operation should always have an internal id as the operation is created together with it set

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_create.go
```diff
@@ -0,0 +1,185 @@
...
+	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
+	ocmerrors "github.com/openshift-online/ocm-sdk-go/errors"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
```
miguelsorianod: This dispatches CS nodepool creation in a controller. At some point I am guessing we might need to do some resources precreation or calculation in RP itself before sending the call to CS. In that case, how would we coordinate this controller and the others?

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_create.go
```diff
@@ -0,0 +1,185 @@
...
+	if operation.ExternalID == nil || !strings.EqualFold(operation.ExternalID.ResourceType.String(), api.NodePoolResourceType.String()) {
+		return false
+	}
+	if len(operation.InternalID.String()) > 0 {
+		return false
+	}
+	return true
+}
+
+func (c *dispatchNodePoolCreate) SynchronizeOperation(ctx context.Context, key controllerutils.OperationKey) error {
```
miguelsorianod: This is an operation controller and not a nodepool one. Is that what we want?

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_create.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil
+	}
+
+	ext := operation.ExternalID
+	nodePool, err := c.cosmosClient.HCPClusters(ext.SubscriptionID, ext.ResourceGroupName).NodePools(ext.Parent.Name).Get(ctx, ext.Name)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if nodePool.ServiceProviderProperties.ActiveOperationID != "" &&
```
miguelsorianod: Is checking this the appropriate approach, as well as returning a non error?

backend/pkg/app/backend.go
```diff
@@ -599,6 +551,7 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
 				go doNothingController.Run(ctx, 20)
 				go dispatchRequestCredentialController.Run(ctx, 20)
 				go dispatchRevokeCredentialsController.Run(ctx, 20)
+				go dispatchNodePoolCreateController.Run(ctx, 20)
```
miguelsorianod: Although this PR moves nodepool creation, I wonder if we need to move at the same time the deletion. If we don't do that I am thinking we could end up in a place where the nodepool is created, creation on CS side for some reason never occurs, and then a delete is issued but never succeeds because on the frontend layer we error if clusterServiceID has never been set in the first place.

backend/pkg/controllers/operationcontrollers/operation_node_pool_create.go
```diff
@@ -83,6 +69,9 @@ func (c *operationNodePoolCreate) ShouldProcess(ctx context.Context, operation *
 	if operation.ExternalID == nil || !strings.EqualFold(operation.ExternalID.ResourceType.String(), api.NodePoolResourceType.String()) {
 		return false
 	}
+	if len(operation.InternalID.String()) == 0 {
```
miguelsorianod: Aside from this new check added in operation nodepool create, I notice that in the cluster one there's the following comment:

```
		// TODO when we update to make clusterserice creation async, we need https://github.com/Azure/ARO-HCP/pull/4695 or similar
		// and we need to wire up a fail-safe where if we have no ID and we time out, we report the best failure we can.
```

Do we need to think about something around this at this point?

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_create.go
```diff
@@ -0,0 +1,185 @@
...
+	if operation.Status.IsTerminal() {
+		return false
+	}
+	if operation.Request != database.OperationRequestCreate {
+		return false
+	}
+	if operation.ExternalID == nil || !strings.EqualFold(operation.ExternalID.ResourceType.String(), api.NodePoolResourceType.String()) {
+		return false
+	}
+	if len(operation.InternalID.String()) > 0 {
```
miguelsorianod: This is the check that ends up controlling whether we are fully done with the work in this controller or not

miguelsorianod:
I updated the PR to show how that would look like. The set of changes are:

Moves node pool Cluster Service creation out of the frontend and into the backend, and updates integration tests so frontend-only runs still work.
1. **Backend: async node pool CS create (initially gated)** (`3da7b4c60`)
   - Adds `NodePoolClusterServiceCreateController` to POST/adopt the node pool in Cluster Service and persist `clusterServiceID` on the Cosmos document.
   - **Updates the node pool create operation to read `clusterServiceID` from the node pool resource instead of the operation document.** This is a change that is not gated but that is needed already to stop relying on the operation internal id and it shouldn't have an impact.
   - Controller only runs when `clusterServiceID` is missing (safe while frontend still sets it during rollout).
2. **Frontend: stop CS create on node pool PUT** (`1ddc67f19`)
   - Removes `PostNodePool` and related logic from the frontend create path; create writes Cosmos + operation only.
   - Node pool create operations no longer carry a Cluster Service `InternalID` on the operation.
3. **Integration tests: `setClusterServiceID` step** (`894a23cbd`)
   - Frontend integration tests do not run the backend controller, so update/delete would fail without a persisted `clusterServiceID`.
   - Adds a `setClusterServiceID` databasemutation step (clusters, node pools, external auths) with backend-aligned derivation and optional `cluster-service-id.json`.
   - Wires affected FrontendCRUD artifacts and cross-version roundtrip tests; adds `ClusterServiceMock.FindClusterHREF`.

miguelsorianod:
This PR shows the end-result but we should not merge it and we should instead rollout in different stages:
but we would be able to roll it out in stages:

1. Introduce the backend side changes disabled. This includes updating the nodepool create operation controller to stop relying on the operation's internal id. This should be fine because even though it changes code the result should be the same even with old code
2. Enable it by modifying it in frontend

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if !c.needsWork(ctx, nodePool) {
+		return nil
+	}
+
+	nodePool, err = c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).NodePools(key.HCPClusterName).Get(ctx, key.HCPNodePoolName)
```
deads2k, line 97: Add comment that this is only necessary because we're about to use cluster-service, not cosmos.

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if !c.needsWork(ctx, nodePool) {
+		return nil
+	}
+
+	cluster, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
```
deads2k: cache lookup, before the live get.

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	if cluster.ServiceProviderProperties.ClusterServiceID == nil || len(cluster.ServiceProviderProperties.ClusterServiceID.String()) == 0 {
```
deads2k, line 117: part of needswork.

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		logger.Info("performing POST node pool to Cluster Service", "cs_node_pool_href", csNodePoolHREF, "node_pool_resource_id", nodePool.ID.String())
+		_, err = c.clustersServiceClient.PostNodePool(ctx, clusterCSInternalID, csNodePoolBuilder)
+		if err != nil {
+			return utils.TrackError(err)
+		}
+	}
+
+	logger.Info("setting node pool cluster service ID in node pool cosmos document", "node_pool_cluster_service_id", nodePoolCSInternalID.String())
+	nodePool.ServiceProviderProperties.ClusterServiceID = nodePoolCSInternalID.DeepCopy() // DeepCopy() to avoid referencing the original pointer
```
deads2k: follow a pattern like 


```suggestion
    replacement := nodePool.DeepCopy()
	replacement.ServiceProviderProperties.ClusterServiceID = nodePoolCSInternalID.DeepCopy() // DeepCopy() to avoid referencing the original pointer
		_, err = c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).NodePools(key.HCPClusterName).Replace(ctx, nodePool, nil)
		if preconditionfail{
		  return nil
		  }
	if err != nil{
	return utilsTrackError(err)
	}
```

backend/pkg/controllers/operationcontrollers/operation_node_pool_create.go
```diff
@@ -105,5 +111,31 @@ func (c *operationNodePoolCreate) SynchronizeOperation(ctx context.Context, key
 		return nil // no work to do
 	}
 
-	return pollNodePoolStatus(ctx, c.clock, c.resourcesDBClient, c.clusterServiceClient, operation, c.notificationClient)
+	nodePool, err := c.resourcesDBClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).NodePools(operation.ExternalID.Parent.Name).Get(ctx, operation.ExternalID.Name)
```
deads2k: cached should be good enough I think.

deads2k:
Some comments. broadly lgtm.

deads2k:
take it out of draft, let's see if it works.

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
miguelsorianod: I kept this separate from NeedsWork because it's something that shouldn't occur and that if it occurs we return an error.

The reasoning around "it shouldn't occur" is because at the point of creation of the nodepool the cluster should exist and have a CSID. If not, something unexpected occurred and it's worth logging.

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
miguelsorianod: For the cosmos cluster resource, even though we reference its `cluster.ServiceProviderProperties.ClusterServiceID` attribute, which is then used to interact with CS, it's not a value that should change over time so I think we can change it so the controller uniquely uses the lister. If in the future some "fresh" data is needed then we would need to change it to use live data, but for now it doesn't seem necessary.

I updated the controller to only use the lister for the cluster resource. I also added a comment indicating why for now we can just use the lister for the cluster, and that if other data is used in the future that can change it should be updated to retrieve from the DB.

backend/pkg/controllers/operationcontrollers/operation_node_pool_create.go
miguelsorianod: Given that it interacts with CS with data, I preferred to keep live data to ensure the most freshness of it.

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
miguelsorianod: updated :+1:

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
miguelsorianod: done

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
deads2k: >I kept this separate from NeedsWork because it's something that shouldn't occur and that if it occurs we return an error.

Why must it fail instead of simply retrying later?

backend/pkg/controllers/operationcontrollers/operation_node_pool_create.go
deads2k: > Given that it interacts with CS with data, I preferred to keep live data to ensure the most freshness of it.

it's reading, not writing.

backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
```diff
@@ -0,0 +1,186 @@
...
+			return utils.TrackError(err)
+		}
+	}
+
+	logger.Info("setting node pool cluster service ID in node pool cosmos document", "node_pool_cluster_service_id", nodePoolCSInternalID.String())
+	replacement := nodePool.DeepCopy()
+	replacement.ServiceProviderProperties.ClusterServiceID = nodePoolCSInternalID.DeepCopy() // DeepCopy() to avoid referencing the original pointer
+	_, err = c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).NodePools(key.HCPClusterName).Replace(ctx, replacement, nil)
+	if database.IsPreconditionFailedError(err) {
+		// if we have a conflict error, then we're guaranteed that our informer will eventually see an update and trigger us again.
```
deads2k, line 155: note this is only safe because we search by list too

backend/pkg/controllers/operationcontrollers/operation_node_pool_create.go
```diff
@@ -87,23 +103,111 @@ func (c *operationNodePoolCreate) ShouldProcess(ctx context.Context, operation *
...
 	}
 
-	return pollNodePoolStatus(ctx, c.clock, c.resourcesDBClient, c.clusterServiceClient, operation, c.notificationClient)
+	nodePool, err := c.nodePoolLister.Get(ctx, operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName, operation.ExternalID.Parent.Name, operation.ExternalID.Name)
+	if err != nil {
+		return fmt.Errorf("failed to get node pool: %w", err)
+	}
+
+	if operation.ResourceID.Name != nodePool.ServiceProviderProperties.ActiveOperationID {
+		return nil
```
deads2k, line 132: followup: log message I think.

test/e2e/simple_negative_cases.go
```diff
@@ -143,7 +143,7 @@ var _ = Describe("Customer", func() {
 				nodePoolParamsInvalidInstance,
 				5*time.Minute,
 			)
-			checkExpectedError(&errs, "node pool creation with invalid instance type", err, "machine type not supported")
+			checkExpectedError(&errs, "node pool creation with invalid instance type", err, "properties.platform.vmSize Unsupported value Standard_A1_v2")
```
deads2k, line 146: now tested in integration and unit. remove test.

miguelsorianod:

Closing this.

PR #5128, opened 2026-05-05
test: VAP Deny->Audit to capture image violations without blocking

raelga:
### What

Changes `validationActions` from `Deny` to `Audit` in the dev cloud override.

### Why

Diagnostic PR to determine whether the VAP is the root cause of the 77 consecutive e2e-parallel failures on #4690 ([AROSLSRE-768](https://redhat.atlassian.net/browse/AROSLSRE-768)).

`Audit` mode logs violations in the API server audit logs without blocking pod creation. If we see audit annotations for non-compliant images, that proves the `Deny` policy was blocking them and causing the cluster creation timeouts. If we see none, the failures were purely infra.

### How to verify

Once the mgmt cluster is provisioned, connect and check for audit violations:

```bash
# Check API server audit logs for VAP violations
kubectl logs -n kube-system -l component=kube-apiserver | grep 'image-registry-allowlist'

# Or query Kusto kubeAudit table
kubeAudit
| where timestamp > ago(2h)
| where annotations has "image-registry-allowlist"
| project timestamp, objectRef, annotations
```

### Expected outcomes

- **Audit violations found**: the VAP was blocking images in `Deny` mode, causing the 77 failures. Need to either mirror those images to ACR or exclude the affected namespaces.
- **No audit violations**: all images come from ACR. The failures were infra saturation (as described in [AROSLSRE-768](https://redhat.atlassian.net/browse/AROSLSRE-768)) and the rebase onto newer main resolved them.

### Parallel test

- #4690 runs with `Deny` (no blanket allowlist) - will it fail again?
- This PR runs with `Audit` - captures what would be denied without blocking

### DIFF
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -40,6 +40,8 @@ ignore:
 - 'backend/deploy/templates/backend.service-key-vault.secretproviderclass.yaml'
 - 'backend/deploy/templates/backend.deployment.yaml'
 - 'tooling/aro-hcp-exporter/deploy/templates/deployment.yaml'
+- 'image-registry-policy/deploy/templates/validatingadmissionpolicybinding.yaml'
+- 'image-registry-policy/values.yaml'
 rules:
   brackets: enable
   colons: enable
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -2392,11 +2392,19 @@
           "items": {
             "type": "string"
           }
+        },
+        "validationActions": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": ["Deny", "Audit", "Warn"]
+          }
         }
       },
       "additionalProperties": false,
       "required": [
-        "extraAllowedRegistries"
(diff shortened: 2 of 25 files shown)


raelga:
### Findings

This diagnostic PR confirmed the root cause: the VAP with `Deny` blocks `capi-provider` pods that use `quay.io/redhat-user-workloads/crt-redhat-acm-tenant/control-plane-operator-4-20`, which is not mirrored to ACR and not covered by `registryOverrides`.

With `Audit` mode (this PR), the same image runs successfully and 40/41 tests pass. With `Deny` mode (#4690), the `capi-provider` ReplicaSet shows `FailedCreate` events with the VAP denial message.

### Fix

The fix is in [AROSLSRE-777](https://redhat.atlassian.net/browse/AROSLSRE-777) / #5137: mirror CPO images to ACR via oc-mirror and add the registry override to HyperShift config.

Closing this diagnostic PR.

PR #5129, opened 2026-05-05
Check for changes that may affect existing cluster nodes

mgahagan73:
https://redhat.atlassian.net/browse/ARO-26612

### What

Adds a pre-merge check for hypershift flag changes that may affect node changes on existing clusters. 

### Why

Address issues described in https://redhat.atlassian.net/browse/ARO-26610

### DIFF
--- a/docs/node-rollout-premerge-check.md
+++ b/docs/node-rollout-premerge-check.md
@@ -1,0 +1,174 @@
+# Pre-Merge Node Rollout Check
+
+## Motivation
+
+On April 28, 2026, a production rollout to EASTUS2 triggered an unexpected
+rolling replacement of customer worker nodes across hosted clusters
+([ARO-26610](https://redhat.atlassian.net/browse/ARO-26610)). The root cause
+was a combination of changes to `hypershift install` flags (`--registry-overrides`
+and `IMAGE_SHARED_INGRESS_HAPROXY`) that altered the rendered MachineConfig hash,
+causing Kubernetes to roll all MachineDeployments.
+
+The pre-merge node rollout check (`TestNodeRolloutConfig`) was introduced to
+prevent this class of incident by making node-affecting flag changes visible
+during code review.
+
+## How the Check Works
+
+The test lives in
+[`tooling/helmtest/testrunner/node_rollout_config.go`](../tooling/helmtest/testrunner/node_rollout_config.go)
+and runs as part of `make test` in the `tooling/helmtest` module.
+
+1. **Render** the HyperShift operator Helm chart for each management cluster in
+   the topology.
+2. **Extract** the shell script from the rendered `install-hypershift` Job
+   ([`hypershiftoperator/deploy/templates/installer.job.yaml`](../hypershiftoperator/deploy/templates/installer.job.yaml)).
+3. **Parse** every flag and `--additional-operator-env-vars` value from the
+   `hypershift install` command.
+4. **Classify** each flag/env var against the registries in
+   `flagCategories` and `envVarCategories`.
+5. **Write** a `NodeRolloutConfig` YAML struct containing the node-affecting
+   values and any unclassified flags to a golden fixture file
+   (`zz_fixture_TestNodeRolloutConfig_*.yaml`).
+6. **Compare** the output to the committed fixture. Any diff fails the test.
+
+### Flag Classification
+
+Every flag is placed into one of three buckets:
+
+| Bucket | Effect | Example |
+|--------|--------|---------|
+| **`flagSafe`** | Operator-level only; no impact on worker nodes | `--enable-conversion-webhook`, `--platform-monitoring` |
+| **`flagNodeAffecting`** | Changes propagate into MachineConfig/MachineDeployment hashes and trigger worker node replacement | `--registry-overrides`, `IMAGE_SHARED_INGRESS_HAPROXY` |
+| **Unclassified** | Flag is not in the registry; surfaces in `additionalInstallArgs` in the fixture, forcing explicit review | Any new flag |
+
+The "unclassified" default is intentionally the safest: unknown flags block the
+PR until someone classifies them.
+
+## How Initial Values Were Gathered
+
+The initial flag classification was determined by tracing the HyperShift
+operator source code from each `hypershift install` flag down to the
+MachineDeployment template hash computation. The key question for each flag:
+**does changing this value alter the hash that names the bootstrap
+DataSecret, which in turn triggers a MachineDeployment rolling update?**
+
+### The HAProxy / Registry-Overrides Chain (root cause of ARO-26610)
+
+The investigation ([ARO-26610 comment](https://redhat.atlassian.net/browse/ARO-26610?focusedCommentId=16862616))
+traced the following path through HyperShift:
+
+1. `--registry-overrides` flag is parsed in `hypershift-operator/main.go`
+2. `RegistryOverrides` is applied to **all** ImageStream tags via string
+   replacement (`support/releaseinfo/registry_mirror_provider.go`)
+3. The HAProxy image is extracted with overrides already applied
+   (`controllers/nodepool/nodepool_controller.go`)
+4. The image URL is embedded in a static pod spec
+   (`controllers/nodepool/apiserver-haproxy/haproxy.go`)
+5. The pod spec is serialized into a MachineConfig (`haproxyRawConfig`)
+6. `haproxyRawConfig` is included in `mcoRawConfig`, which is hashed
+   (`controllers/nodepool/config.go`)
+7. The hash determines the bootstrap Secret name
+   (`controllers/nodepool/token.go`)
+8. A Secret name change causes a MachineDeployment `.spec.template` update,
+   triggering a rolling replacement
+   (`controllers/nodepool/capi.go`)
+
+**Why HAProxy but not other data-plane images?** HAProxy is deployed as a
+static pod baked into the ignition config (it must run without API server
+access for bootstrapping). Other data-plane images (kube-proxy, OVN, etc.) use
+IDMS applied by CRI-O at runtime, so registry override changes do not alter
+ignition content.
+
+### Documented Rollout Triggers in HyperShift API
+
+The HyperShift API
+([`api/hypershift/v1beta1/hostedcluster_types.go`](https://github.com/openshift/hypershift/blob/7aa62352dc5c63ebeebff1b83ce92cd6df748e43/api/hypershift/v1beta1/hostedcluster_types.go#L753))
+marks fields that trigger node rollouts with a `+rollout` tag. At least 6
+`HostedCluster.Spec` fields are explicitly documented:
+
+| Field | Why It Triggers Rollout |
+|-------|----------------------|
+| `PullSecret` | Name change alters ignition secret references |
+| `SSHKey` | Baked into ignition MachineConfig |
+| `ImageContentSources` | Becomes IDMS in ignition payload |
+| `Configuration.Proxy` | Alters MachineConfig proxy settings |
+| `Configuration.Image` | Affects image references in ignition |
+| `AdditionalTrustBundle` | Injected into ignition trust store |
+
+### Full List of Known Rollout Triggers
+
+Combining the `+rollout`-tagged API fields with the code-traced operator-level
+flags, there are **9 confirmed triggers**:
+
+**Operator-level (3):**
+- `--registry-overrides` -- when HAProxy config is managed (the default unless
+  CPO has a skip label)
+- `--control-plane-operator-image` -- when HTTP proxy is configured
+- `IMAGE_SHARED_INGRESS_HAPROXY` -- when shared ingress is enabled
+
+**HostedCluster.Spec (6):**
+- `ImageContentSources`
+- `Configuration.Proxy`
+- `Configuration.Image`
+- `AdditionalTrustBundle`
+- `PullSecret` (name change)
+- `Release.Image`
+
+**Requires further analysis (2):**
+- `--feature-gates` and `HYPERSHIFT_FEATURESET` -- impact depends on which
+  specific gate is toggled; per-gate analysis is needed.
+
+## Maintaining the Check
+
+### When a new flag is added to `hypershift install`
+
+1. The test will **fail** because the new flag appears in
+   `additionalInstallArgs` in the golden fixture.
+2. Determine whether the flag is safe or node-affecting:
+   - Trace the flag through the HyperShift operator source to see if it
+     reaches the MachineDeployment template hash.
+   - Check whether the flag's value is embedded in any MachineConfig,
+     ignition payload, or bootstrap secret.
+   - If the flag only affects operator-level behavior (webhooks, CRD
+     management, monitoring), it is safe.
+3. Add the flag to `flagCategories` (or `envVarCategories` for env vars) in
+   `node_rollout_config.go` with the appropriate classification.
+4. If the flag is `flagNodeAffecting`, add a corresponding field to
+   `NodeRolloutConfig` and extraction logic, so the value is tracked
+   explicitly in the fixture.
+5. Update the fixture:
+   ```
+   UPDATE_NODE_ROLLOUT_FIXTURE=true go test -run TestNodeRolloutConfig -count=1 ./...
+   ```
+6. Document the impact analysis in the PR description.
+
+### When a node-affecting value changes
+
+For example, adding a new `--registry-overrides` entry:
+
+1. The test will **fail** because the fixture diff shows the changed value.
+2. Document in the PR description:
+   - What changed and why.
+   - Expected impact: rolling replacement of all worker nodes across all
+     hosted clusters in affected regions.
+   - Rollout coordination plan with SRE.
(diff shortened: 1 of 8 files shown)


tooling/helmtest/testrunner/node_rollout_config.go
```diff
@@ -0,0 +1,263 @@
...
+type NodeRolloutConfig struct {
+	RegistryOverrides         []string `json:"registryOverrides" yaml:"registryOverrides"`
+	HypershiftImage           string   `json:"hypershiftImage" yaml:"hypershiftImage"`
+	SharedIngressHAProxyImage string   `json:"sharedIngressHAProxyImage,omitempty" yaml:"sharedIngressHAProxyImage,omitempty"`
+	AdditionalInstallArgs     []string `json:"additionalInstallArgs,omitempty" yaml:"additionalInstallArgs,omitempty"`
+}
+
+const nodeRolloutWarningHeader = `# WARNING: Changes to this file indicate HyperShift operator config changes
+# that WILL trigger customer worker node replacements across all hosted
+# clusters in affected regions.
```
adalrsjr1: This header says "config changes that WILL trigger customer worker node replacements" but the test only covers operator install-time flags. Per the investigation, there are at least other `HostedCluster.Spec` fields that also trigger rollouts which are not covered here because they're set per-cluster, not at install time.

The header should explicitly state the scope — something like "HyperShift operator install flags" — so reviewers don't assume this test covers all rollout-triggering changes and develop false confidence.

tooling/helmtest/testrunner/node_rollout_config.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+func collectUnknownFlags(script string) []string {
+	lines := strings.Split(script, "\n")
+	var unknown []string
+
+	knownNodeAffecting := map[string]bool{
+		"--registry-overrides": true,
+		"--hypershift-image":   true,
+	}
```
adalrsjr1: For consistence, this map could be defined as in lines 32-44. Consider a single source of truth — e.g. a struct or map that defines each flag's category (safe, node-affecting, etc) in one place. Even better if all lists come from a external file to facilitate its edition

adalrsjr1:
I want to confirm the behavior of this test. Whenever changes are applied to this [file](hypershiftoperator/zz_fixture_TestNodeRolloutConfig_dev_westus3_mgmt_1_hypershift.yaml) the test will fail. To pass, the user must explictly run `UPDATE=true make -C tooling/helmtest update`. Is my understand correct?

hypershiftoperator/zz_fixture_TestNodeRolloutConfig_dev_westus3_mgmt_1_hypershift.yaml
```diff
@@ -0,0 +1,18 @@
+# WARNING: Changes to this file indicate HyperShift operator config changes
+# that WILL trigger customer worker node replacements across all hosted
+# clusters in affected regions.
+#
+# Before updating:
+# 1. Document impact analysis in your PR description
+# 2. Coordinate rollout scheduling with SRE
+# 3. Run: UPDATE=true make -C tooling/helmtest update
```
stevekuznetsov: I would recommend that you use a special env var, not `$UPDATE`, here, since many people (me included) just have a habit to run the update and we want this to be obvious and exceptional. See sdp-pipelines repo for an example of how we use special env var for golden file update.

hypershiftoperator/zz_fixture_TestNodeRolloutConfig_dev_westus3_mgmt_1_hypershift.yaml
```diff
@@ -0,0 +1,18 @@
...
+additionalInstallArgs:
+- --enable-cpo-overrides
+hypershiftImage: arohcpsvcdev.azurecr.io/redhat-services-prod/crt-redhat-acm-tenant/hypershift/hypershift-operator@sha256:1234567890
```
stevekuznetsov: In the abstract, no, the HyperShift version should _never_ cause a node rollout, and they have ample testing to ensure it. Remove this

hypershiftoperator/zz_fixture_TestNodeRolloutConfig_dev_westus3_mgmt_1_hypershift.yaml
```diff
@@ -0,0 +1,18 @@
...
+registryOverrides:
+- quay.io/openshift-release-dev/ocp-release-nightly=arohcpocpdev.azurecr.io/openshift-release-dev/ocp-release-nightly
+- quay.io/openshift-release-dev/ocp-release=arohcpocpdev.azurecr.io/openshift-release-dev/ocp-release
+- quay.io/openshift-release-dev/ocp-v4.0-art-dev=arohcpocpdev.azurecr.io/openshift-release-dev/ocp-v4.0-art-dev
+- quay.io/openshift-release-dev/ocp-v5.0-art-dev=arohcpocpdev.azurecr.io/openshift-release-dev/ocp-v5.0-art-dev
+- registry.redhat.io/redhat=arohcpocpdev.azurecr.io/redhat
+sharedIngressHAProxyImage: arohcpocpdev.azurecr.io/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-shared-ingress-main@sha256:1234567890
```
stevekuznetsov, line 18: I don't see the metrics-set here, even though we know it caused a rollout. Why?

tooling/helmtest/testrunner/golden.go
```diff
@@ -76,16 +76,21 @@ func CompareWithFixture(t *testing.T, output interface{}, opts ...option) {
 		t.Fatalf("failed to read testdata file: %v", err)
 	}
 	if diff := cmp.Diff(string(expected), string(serializedOutput)); diff != "" {
-		t.Errorf("got diff between expected and actual result:\nfile: %s\ndiff:\n%s\n\nIf this is expected, re-run the test with `UPDATE=true go test ./...` to update the fixtures.", golden, diff)
+		if options.DiffMessage != "" {
```
stevekuznetsov: Why is this not just using the shared library for golden fixtures? Please refactor to import it from ARO-Tools.

tooling/helmtest/testrunner/golden.go
```diff
@@ -76,16 +76,21 @@ func CompareWithFixture(t *testing.T, output interface{}, opts ...option) {
...
-	GoldenDir string
+	Prefix      string
+	Suffix      string
+	Extension   string
+	SubDir      string
+	GoldenDir   string
+	DiffMessage string
 }
 
 type option func(*options)
```
stevekuznetsov: Oh, my. This nonsense has it's own copy of the whole library? No. Delete this please. I know you didn't add it, but this is just not acceptable.

tooling/helmtest/testrunner/node_rollout_config.go
```diff
@@ -0,0 +1,263 @@
...
+
+	"github.com/Azure/ARO-HCP/tooling/helmtest/internal"
+)
+
+var knownSafeFlags = map[string]bool{
+	"--enable-conversion-webhook":         true,
+	"--managed-service":                   true,
+	"--aro-hcp-key-vault-users-client-id": true,
+	"--platform-monitoring":               true,
+	"--metrics-set":                       true,
```
stevekuznetsov: Didn't we specifically identify this as causing a rollout? Document *how* you came up with this list, so when/if the list needs to change, it's clear.

tooling/helmtest/testrunner/node_rollout_config.go
```diff
@@ -0,0 +1,263 @@
...
+	settings, err := internal.LoadSettings(settingsPath)
+	assert.NoError(t, err)
+	assert.NotNil(t, settings)
+
+	helmSteps, err := internal.FindHelmSteps(settings.TopologyDir, settings.ConfigPath)
+	assert.NoError(t, err)
+	assert.NotNil(t, helmSteps)
+
+	for _, helmStep := range helmSteps {
+		if !strings.Contains(helmStep.PipelinePath, "hypershiftoperator/") {
```
stevekuznetsov: This seems wrong, likely better to factor `internal.FindHelmSteps` to let you look one up by name.

tooling/helmtest/testrunner/node_rollout_config.go
```diff
@@ -0,0 +1,263 @@
...
+	for _, doc := range docs {
+		doc = strings.TrimSpace(doc)
+		if doc == "" {
+			continue
+		}
+		if !strings.Contains(doc, "kind: Job") || !strings.Contains(doc, "name: install-hypershift") {
+			continue
+		}
+
+		var job struct {
```
stevekuznetsov: Import the k8s job types. We don't just ad-lib well-known APIs like this.

tooling/helmtest/testrunner/node_rollout_config.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+func extractInstallScript(manifest string) (string, error) {
+	docs := strings.Split(manifest, "\n---\n")
+	for _, doc := range docs {
+		doc = strings.TrimSpace(doc)
+		if doc == "" {
+			continue
+		}
+		if !strings.Contains(doc, "kind: Job") || !strings.Contains(doc, "name: install-hypershift") {
```
stevekuznetsov: Unmarshal into a runtime.Object and look for a Job. This is not correct - we don't use string contains on structured data. See Helm step code in ARO-Tools for an example of how to inspect a bag of YAMLs for individual objects.

tooling/helmtest/testrunner/golden.go
```diff
@@ -76,55 +113,17 @@ func CompareWithFixture(t *testing.T, output interface{}, opts ...option) {
...
-func WithGoldenDir(goldenDir string) option {
-	return func(opts *options) {
-		opts.GoldenDir = goldenDir
+		t.Errorf("got diff between expected and actual result:\nfile: %s\ndiff:\n%s\n\nIf this is expected, re-run the test with `%s=true go test ./...` to update the fixtures.", golden, diff, o.updateEnv)
 	}
 }
 
+// sanitizeFilename is based on the unexported function in github.com/Azure/ARO-Tools/testutil
+// but uses inclusive upper bounds (<= 'z', <= 'Z') to correctly include all letters.
+// Upstream uses exclusive bounds (< 'z', < 'Z') which excludes 'z' and 'Z'.
```
mgahagan73, line 122: Is this a bug in upstream? I couldn't 100% tell if we treat "Zz" as special characters for fixtures. If upstream is buggy I can open a PR to fix it as well as export the function and refactor golden.go to use it.

hypershiftoperator/zz_fixture_TestNodeRolloutConfig_dev_westus3_mgmt_1_hypershift.yaml
mgahagan73: - --metrics-set=SRE is now here, under additionalInstallArgs

adalrsjr1:
the implementation is sane; will leave the final lgtm to @stevekuznetsov who made most of comments.

docs/node-rollout-premerge-check.md
```diff
@@ -0,0 +1,174 @@
...
+### When a node-affecting value changes
+
+For example, adding a new `--registry-overrides` entry:
+
+1. The test will **fail** because the fixture diff shows the changed value.
+2. Document in the PR description:
+   - What changed and why.
+   - Expected impact: rolling replacement of all worker nodes across all
+     hosted clusters in affected regions.
+   - Rollout coordination plan with SRE.
```
adalrsjr1, line 155: If possible I would like to enforce the "openshift-ci bot" to request approval from SRE/Lifecycle team. This will ensure SRE are indeed aware of this change.

adalrsjr1:
the new document is a accurate and factual. lgtm from me, It's better you reach out someone form SLC to get a final approval on this. They should be aware of these changes. I guess @janboll is a person to go.

docs/node-rollout-premerge-check.md
mgahagan73: I will address this in a separate PR.

tooling/helmtest/testrunner/node_rollout_config.go
mgahagan73: Header at line 98 now says "HyperShift operator install flag changes"

tooling/helmtest/testrunner/node_rollout_config.go
mgahagan73: flagCategories and envVarCategories maps serve as the single source of truth with documented methodology

hypershiftoperator/zz_fixture_TestNodeRolloutConfig_dev_westus3_mgmt_1_hypershift.yaml
mgahagan73: Now uses UPDATE_NODE_ROLLOUT_FIXTURE (line 156 of node_rollout_config.go)

hypershiftoperator/zz_fixture_TestNodeRolloutConfig_dev_westus3_mgmt_1_hypershift.yaml
mgahagan73: removed

tooling/helmtest/testrunner/golden.go
mgahagan73: fixed, should be using ARO-tools bits wherever needed.

tooling/helmtest/testrunner/golden.go
mgahagan73: done

tooling/helmtest/testrunner/node_rollout_config.go
mgahagan73: see docs/node-rollout-premerge-check.md

tooling/helmtest/testrunner/node_rollout_config.go
mgahagan73: fixed

tooling/helmtest/testrunner/node_rollout_config.go
mgahagan73: fixed

tooling/helmtest/testrunner/node_rollout_config.go
mgahagan73: fixed.. now using utilyaml.NewYAMLOrJSONDecoder decoding into batchv1.Job

tooling/helmtest/testrunner/node_rollout_config.go
```diff
@@ -0,0 +1,283 @@
...
+
+func extractInstallScript(manifest string) (string, error) {
+	decoder := utilyaml.NewYAMLOrJSONDecoder(bytes.NewReader([]byte(manifest)), 4096)
+	for {
+		var job batchv1.Job
+		if err := decoder.Decode(&job); err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			continue
```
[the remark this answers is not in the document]
mgahagan73: This should not be an issue, non-EOF errors hit continue. We attempt batchv1.Job unmarshaling on each documents so we would hit decode errors because the schema does not match. Tracking every error at the end of the loop is just going to add noise.

tooling/helmtest/testrunner/golden.go
```diff
@@ -24,46 +23,84 @@ import (
...
+		if err != nil {
+			t.Fatalf("failed to yaml marshal output of type %T: %v", output, err)
 		}
+		serializedOutput = serialized
 	}
 
-	golden := filepath.Join(options.GoldenDir, sanitizeFilename(options.Prefix+t.Name()+options.Suffix)) + options.Extension
-	if os.Getenv("UPDATE") != "" {
+	golden := filepath.Join(o.goldenDir, sanitizeFilename(t.Name()+o.suffix)) + o.extension
+	if os.Getenv(o.updateEnv) != "" {
```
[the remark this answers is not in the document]
mgahagan73: callers are all using .yaml files so this should not be needed.

mgahagan73:
/assign @janboll 

Hi Jan, would you please review this one?

tooling/helmtest/testrunner/node_rollout_config.go
```diff
@@ -0,0 +1,284 @@
...
+	lines := strings.Split(script, "\n")
+	var unknown []string
+
+	for _, line := range lines {
+		line = strings.TrimSpace(line)
+		line = strings.TrimSuffix(line, `\`)
+		line = strings.TrimSpace(line)
+		if line == "" || line == "hypershift install" {
+			continue
+		}
```
janboll: This looks fragile... is it guaranteed, there is one flag per line?
Perhaps use regex to check each and use a match group instead

janboll (changes requested):
looks good overal!
please review this one comment

tooling/helmtest/testrunner/node_rollout_config.go
mgahagan73: This has been moved to a regex approach, and also added a unit test.

PR #5132, opened 2026-05-05
Revert "Prow provision shard ci01"

roivaz:
Reverts Azure/ARO-HCP#5084


raelga:
PR #5084 ("Prow provision shard ci01") introduced two issues that this revert fixes:

**Issue 1: CI blocked — `config-change-detection` failing on all PRs**

#5084 added `Microsoft.Quota` as a resource provider in `config/config.yaml` but the rendered config for `ci01` (`config/rendered/dev/ci01/westus3.yaml`) was not fully regenerated. The `config-change-detection` CI step runs `make -C config/ materialize` and diffs the output — since the committed rendered config doesn't match what `materialize` produces, every PR against main now fails this check. This has been blocking all PR merges since #5084 was merged yesterday.

**Issue 2: INT deployments blocked — template crash on optional `serviceKeyVault.subscription`**

#5084 also modified `svc-pipeline.yaml` adding a template expression `{{ if .serviceKeyVault.subscription.key }}` that crashes with Go's `missingkey=error` template option when `subscription` doesn't exist as a map key — even inside an `if` guard. This breaks the INT bumper pipeline because INT's config doesn't have `serviceKeyVault.subscription` defined. #5123 has a targeted fix using the `index` function, but reverting #5084 unblocks INT immediately without waiting for that fix.

**Why revert instead of fix-forward:**

Both issues are blocking CI and INT deployments right now. The revert is the fastest path to unblock, and the ci01 shard provisioning can be re-landed with the fixes applied.

PR #5141, opened 2026-05-06
feat: move nodepool replace and patch from frontend to backend

miguelsorianod:
This moves nodepool replace and patch from frontend to backend. Frontend now just creates the corresponding ARM Operation, a nodepool update operation, and it delegates all the actual nodepool replace and patch work to backend.

 On backend now we check whether internalID is empty or not to assess
 whether the nodepool patch request to Clusters Service has been
 submitted. Once that occurs, we check the state.
    
 We introduce a controller DispatchNodePoolUpdate that takes care of
 making sure that a PATCH call is sent to CS when a nodepool replace
 or patch has been issued through the Frontend resulting in a nodepool
 update active operation.


backend/pkg/controllers/operationcontrollers/dispatch_node_pool_update.go
```diff
@@ -0,0 +1,136 @@
+// Copyright 2026 Microsoft Corporation
+//
```
miguelsorianod: The introduction of the changes here need to be actually split into two:

1. Then introduce the frontend removal logic

I think it should be safe to introduce 1 because it will only be executed when the operation exists and it doesn't have an internal id, and until 2 is introduced the nodepool update operation should always have an internal id as the operation is created together with it set.

This PR has been created so we can review how the whole picture looks like.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_update.go
```diff
@@ -0,0 +1,136 @@
...
+	//    call would fail be if CS returns the NodePool is not in ready state, until the update completes
+	// In that case what would occur is that the update call would fail because of the CS nodepool not being in ready state
+	// It's unavoidable that after CS update call succeeds there might be a failure on the cosmos operation update. In that case, the next UpdateNodePool
+	// call would keep failing until the nodepool becomes ready again (update fully completes, ...), which means the
+	// operation would remain active until that occurs.
+	// We could add a similar check to what request credential controller does to check if we get a specific 400 with an error message
+	// that indicates that the nodepool is in "updating" state on CS side but that technically doesn't cover case 1 if somehow CS can end up in a
+	// 'updating' state not triggered by this controller.
+	logger.Info("dispatching PATCH node pool to Cluster Service", "cs_node_pool_href", csIDFromNodePool.String(), "node_pool_resource_id", nodePool.ID.String())
+	_, err = c.clustersServiceClient.UpdateNodePool(ctx, csIDFromNodePool, csNodePoolBuilder)
```
miguelsorianod: This is potentially the biggest contention point in the PR.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_update.go
deads2k: Don't you need a PR to make all users of `nodePool.ServiceProviderProperties.ClusterServiceID` tolerate a nil first?

deads2k: switching to a pointer makes sure all new code will check for nil.  It's a bit of a pain to sweep, but it's a very sure way to ensure the codebase evolves and fails obviously.

miguelsorianod: Done in https://github.com/Azure/ARO-HCP/pull/5169. Once that one is merged I'll rebase this one from it

mbarnes:
Maybe let's focus on #5142 first and get that right.  I made some comments there that are also applicable here.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_update.go
mbarnes: Yeah, you're gonna have to check CS error messages.  [See this for an example](https://github.com/Azure/ARO-HCP/blob/eafade977c97a4132241ff5fb1b6b3af4fb676f3/backend/pkg/controllers/operationcontrollers/dispatch_revoke_credentials.go#L137-L147).

miguelsorianod:
As an important insight, we should move delete before and/or at the same time than create because of https://github.com/Azure/ARO-HCP/pull/5142#discussion_r3210546381

miguelsorianod:
Superseded by #6067

PR #5142, opened 2026-05-06
feat: move nodepool delete from frontend to backend

miguelsorianod:
This moves nodepool delete from frontend to backend. Frontend now just creates the corresponding ARM Operation, a nodepool delete operation, and it delegates all the actual nodepool delete work to backend.

On backend now we check whether internalID is empty or not to assess
whether the nodepool delete request to Clusters Service has been
submitted. Once that occurs, we check the state.

We introduce a controller DispatchNodePoolDelete that takes care of
making sure that a DELETE call is sent to CS when a nodepool delete
has been issued through the Frontend resulting in a nodepool
delete active operation.

### DIFF
--- a/backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
+++ b/backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
@@ -61,7 +61,7 @@ func (c *dispatchNodePoolDelete) ShouldProcess(ctx context.Context, operation *a
 	if operation.Status.IsTerminal() {
 		return false
 	}
-	if operation.Request != database.OperationRequestUpdate {
+	if operation.Request != database.OperationRequestDelete {
 		return false
 	}
 	if operation.ExternalID == nil || !strings.EqualFold(operation.ExternalID.ResourceType.String(), api.NodePoolResourceType.String()) {
--- a/backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
+++ b/backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
@@ -1,0 +1,130 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package operationcontrollers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+
+	ocmerrors "github.com/openshift-online/ocm-sdk-go/errors"
+)
+
+type dispatchNodePoolDelete struct {
+	cosmosClient          database.DBClient
+	clustersServiceClient ocm.ClusterServiceClientSpec
+}
+
+func NewDispatchNodePoolDeleteController(
+	cosmosClient database.DBClient,
+	clustersServiceClient ocm.ClusterServiceClientSpec,
+	activeOperationInformer cache.SharedIndexInformer,
+) controllerutils.Controller {
+	syncer := &dispatchNodePoolDelete{
+		cosmosClient:          cosmosClient,
+		clustersServiceClient: clustersServiceClient,
+	}
+
+	return NewGenericOperationController(
+		"DispatchNodePoolDelete",
+		syncer,
+		10*time.Second,
+		activeOperationInformer,
+		cosmosClient,
+	)
+}
+
+func (c *dispatchNodePoolDelete) ShouldProcess(ctx context.Context, operation *api.Operation) bool {
+	if operation.Status.IsTerminal() {
+		return false
+	}
+	if operation.Request != database.OperationRequestUpdate {
+		return false
+	}
+	if operation.ExternalID == nil || !strings.EqualFold(operation.ExternalID.ResourceType.String(), api.NodePoolResourceType.String()) {
+		return false
+	}
+	if len(operation.InternalID.String()) > 0 {
+		return false
+	}
+
+	return true
+}
+
+func (c *dispatchNodePoolDelete) SynchronizeOperation(ctx context.Context, key controllerutils.OperationKey) error {
+	logger := utils.LoggerFromContext(ctx)
+	logger.Info("checking operation")
+
+	operation, err := c.cosmosClient.Operations(key.SubscriptionID).Get(ctx, key.OperationName)
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get active operation: %w", err)
+	}
+	if !c.ShouldProcess(ctx, operation) {
+		return nil
+	}
+
+	ext := operation.ExternalID
+	nodePool, err := c.cosmosClient.HCPClusters(ext.SubscriptionID, ext.ResourceGroupName).NodePools(ext.Parent.Name).Get(ctx, ext.Name)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if nodePool.ServiceProviderProperties.ActiveOperationID != "" &&
+		nodePool.ServiceProviderProperties.ActiveOperationID != operation.OperationID.Name {
+		logger.Info("skipping node pool delete dispatch: active operation mismatch",
+			"nodepool_active_operation_id", nodePool.ServiceProviderProperties.ActiveOperationID,
+			"operation_name", operation.OperationID.Name)
+		return nil // TODO should this be return error or nil?
+	}
+
+	csIDFromNodePool := nodePool.ServiceProviderProperties.ClusterServiceID
+	if len(csIDFromNodePool.String()) == 0 {
+		return utils.TrackError(fmt.Errorf("node pool %s has no ClusterServiceID", ext.Name)) // TODO should this be return error or nil?
+	}
+
+	// If the CS nodepool delete is reexecuted and the nodepool is already in CS uninstalling state, the delete call will be successful.
+	// If the CS nodepool's parent cluster is being deleted, the delete call will return an error.
+	logger.Info("dispatching DELETE node pool to Cluster Service", "cs_node_pool_href", csIDFromNodePool.String(), "node_pool_resource_id", nodePool.ID.String())
+	err = c.clustersServiceClient.DeleteNodePool(ctx, csIDFromNodePool)
+	var ocmError *ocmerrors.Error
+	if errors.As(err, &ocmError) && ocmError.Status() == http.StatusNotFound {
+		err = nil
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	operation.InternalID = csIDFromNodePool
+	_, err = c.cosmosClient.Operations(key.SubscriptionID).Replace(ctx, operation, nil)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	return nil
+}
--- a/frontend/pkg/frontend/node_pool.go
+++ b/frontend/pkg/frontend/node_pool.go
@@ -17,7 +17,6 @@ package frontend
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"maps"
 	"net/http"
@@ -29,8 +28,6 @@ import (
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
-
-	ocmerrors "github.com/openshift-online/ocm-sdk-go/errors"
 
 	"github.com/Azure/ARO-HCP/internal/admission"
 	"github.com/Azure/ARO-HCP/internal/api"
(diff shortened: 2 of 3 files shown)


backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
```diff
@@ -0,0 +1,130 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: I wonder if a delete should have some kind of effect to other active operations

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
```diff
@@ -0,0 +1,130 @@
...
+//
```
miguelsorianod, line 2: The introduction of the changes here need to be actually split into two:

1. Then introduce the frontend removal logic

I think it should be safe to introduce 1 because it will only be executed when the operation exists and it doesn't have an internal id, and until 2 is introduced the nodepool delete operation should always have an internal id as the operation is created together with it set.

This PR has been created so we can review how the whole picture looks like.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
```diff
@@ -0,0 +1,130 @@
...
+	if nodePool.ServiceProviderProperties.ActiveOperationID != "" &&
+		nodePool.ServiceProviderProperties.ActiveOperationID != operation.OperationID.Name {
+		logger.Info("skipping node pool delete dispatch: active operation mismatch",
+			"nodepool_active_operation_id", nodePool.ServiceProviderProperties.ActiveOperationID,
+			"operation_name", operation.OperationID.Name)
+		return nil // TODO should this be return error or nil?
+	}
+
+	csIDFromNodePool := nodePool.ServiceProviderProperties.ClusterServiceID
+	if len(csIDFromNodePool.String()) == 0 {
```
miguelsorianod, line 107: I think here we could have the scenario where we receive a delete but the nodepool was never created in clusters service for some reason.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
```diff
@@ -0,0 +1,130 @@
...
+		return utils.TrackError(fmt.Errorf("node pool %s has no ClusterServiceID", ext.Name)) // TODO should this be return error or nil?
+	}
+
+	// If the CS nodepool delete is reexecuted and the nodepool is already in CS uninstalling state, the delete call will be successful.
+	// If the CS nodepool's parent cluster is being deleted, the delete call will return an error.
```
miguelsorianod, line 112: Checking current CS delete logic, there's another case where we would get an error from CS:

If we submit the delete but the node pool's parent cluster is in a non ready state (uninstalling, updating, error, ...) CS will return an error, until the cluster is ready. This means if the CS parent cluster doesn't end up being ready this will fail permanently until the issue is fixed with the parent cluster.

frontend/pkg/frontend/node_pool.go
```diff
@@ -673,20 +670,6 @@ func (f *Frontend) DeleteNodePool(writer http.ResponseWriter, request *http.Requ
...
-		// the RP to delete a resource (e.g. "ocm delete"). It can
-		// also happen if an asynchronous deletion operation fails.
-		// we will fall through and cancel all operations and go through as normal a deletion flow as we can to avoid
-		// leaking data related to the resource, like controller status.
-		logger.Info("clusterService nodepool missing, trying to clean up", "err", err)
-	} else if err != nil {
-		return utils.TrackError(err)
-	}
-
 	transaction := f.dbClient.NewTransaction(nodePool.ID.SubscriptionID)
```
deads2k, line 673: I bet we need to have a marker for "this nodepool  is being deleted" on the nodepool in addition to the operation so that we can fail the operation and still continue deleting.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
```diff
@@ -0,0 +1,130 @@
...
+		return false
+	}
+	if len(operation.InternalID.String()) > 0 {
+		return false
+	}
+
+	return true
+}
+
+func (c *dispatchNodePoolDelete) SynchronizeOperation(ctx context.Context, key controllerutils.OperationKey) error {
```
deads2k, line 77: we need to trigger off a nodepool state for this, not an operation. Perhaps that's actually a common thread for all our create/update/delete.  The changes being requested are distinct from the operations and can live longer than the operations.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
mbarnes: There's a [database.CancelActiveOperations](https://github.com/Azure/ARO-HCP/blob/bfb7f26826848a86ac11d45630649b7792b735f9/frontend/pkg/frontend/node_pool.go#L717-L723) call in the frontend delete handler.

I tend to think the frontend is where that belongs so cancellation is immediate, but I know @deads2k doesn't like transactions.  But that's how I handled credential revocations when I was moving that to the backend: frontend still cancels any ongoing credential requests immediately.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
mbarnes: Historically when the frontend initiates a cluster deletion in CS, it creates deletion operations in Cosmos for all the nested resources.  But IIRC, CS doesn't update the state of its nested resources to indicate they're being deleted.

So maybe pull the cluster resource from CS here.  If it's in an uninstalling state then we can skip the `DeleteNodePool` call since it's effectively already dispatched.  Operation controller just has to wait for the 404 response.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
```diff
@@ -0,0 +1,130 @@
...
+		activeOperationInformer,
+		cosmosClient,
+	)
+}
+
+func (c *dispatchNodePoolDelete) ShouldProcess(ctx context.Context, operation *api.Operation) bool {
+	if operation.Status.IsTerminal() {
+		return false
+	}
+	if operation.Request != database.OperationRequestUpdate {
```
mbarnes:
```suggestion
	if operation.Request != database.OperationRequestDelete {
```

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
```diff
@@ -0,0 +1,130 @@
...
+		return false
+	}
+	if operation.ExternalID == nil || !strings.EqualFold(operation.ExternalID.ResourceType.String(), api.NodePoolResourceType.String()) {
+		return false
+	}
+	if len(operation.InternalID.String()) > 0 {
```
mbarnes, line 70: Check here is wrong.  [`operation.InternalID` is set by the frontend](https://github.com/Azure/ARO-HCP/blob/bfb7f26826848a86ac11d45630649b7792b735f9/frontend/pkg/frontend/node_pool.go#L728-L736).
Either this suggestion or keep the check in `SynchronizeOperation` that logs an error.
```suggestion
	if len(operation.InternalID.String()) == 0 {
```

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
mbarnes: `operation.InternalID` and `nodePool.ServiceProviderProperties.ClusterServiceID` should be equal.

You could just validate `operation.InternalID` in `ShouldProcess` like I suggested above and use that for `csIDFromNodePool`.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	operation.InternalID = csIDFromNodePool
+	_, err = c.cosmosClient.Operations(key.SubscriptionID).Replace(ctx, operation, nil)
+	if err != nil {
+		return utils.TrackError(err)
+	}
```
mbarnes, line 127: This looks unnecessary for a delete.  [`operation.InternalID` is already set by the frontend](https://github.com/Azure/ARO-HCP/blob/bfb7f26826848a86ac11d45630649b7792b735f9/frontend/pkg/frontend/node_pool.go#L728-L736).

mbarnes (changes requested):

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
mbarnes: Yes, this is what @machi1990 suggested in [ARO-24384](https://redhat.atlassian.net/browse/ARO-24384?focusedCommentId=16548992): use the CS resource status.  Initially I was going to add a `dispatched` flag to operations, but he's right.  It can be inferred.

Also, a deletion operation for a nested resource is a little tricky because it could be an [_implicit_ operation created by a cluster deletion](https://github.com/Azure/ARO-HCP/blob/main/frontend/asynchronous-operations.md#explicit-vs-implicit-operations).  I suggested in another comment we'll also want to check if the CS cluster is being uninstalled, in which case the node pool deletion is effectively already dispatched and the CS delete call can be skipped here.

frontend/pkg/frontend/node_pool.go
mbarnes: Node pool's provisioning state will indicate it's being deleted.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
miguelsorianod: Here I was missing the frontend side to send an empty operation InternalIID. At that point this would be checking that. I had it in the patch and in the create MRs but I forgot to commit it here.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
miguelsorianod: Here I was missing the frontend side to send an empty operation InternalID. At that point this would be checking that. I had it in the patch and in the create MRs but I forgot to commit it here.

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
miguelsorianod: :+1: . Copy-paste leftover from update one. Changed

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
miguelsorianod: Check now to see what I mean. I pushed the changes with what was missing (setting empty internalid for nodepool delete operation in nodepool)

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
miguelsorianod: Check now to see what I mean. I pushed the changes with what was missing (setting empty internalid for nodepool delete operation in nodepool)

frontend/pkg/frontend/node_pool.go
```diff
@@ -728,7 +703,7 @@ func (f *Frontend) addDeleteNodePoolToTransaction(ctx context.Context, writer ht
 	operationDoc := database.NewOperation(
 		database.OperationRequestDelete,
 		nodePool.ID,
-		nodePool.ServiceProviderProperties.ClusterServiceID,
+		api.InternalID{},
```
miguelsorianod, line 706: This is like this now but after https://github.com/Azure/ARO-HCP/pull/5169  is merged it we would set it to null here

backend/pkg/controllers/operationcontrollers/dispatch_node_pool_delete.go
miguelsorianod: Comment based on the MR that was missing setting the internalid as empty/null from the frontend side. Review again based on that

frontend/pkg/frontend/node_pool.go
miguelsorianod: This is the key point in the MR, which is that we always set the internalid as empty/null when creating the delete operation and the delete operation and dispatch delete worker use that being set as the marker of "we've sent the delete to cs", where the dispatch controller is the one setting it when it's sure the work it needs to do is done

miguelsorianod: As a relevant note, as I mentioned in https://github.com/Azure/ARO-HCP/pull/5142#discussion_r3207999163, it's impossible to guarantee on the frontend to have a ClusterServiceID at the point of the delete request. The following scenario could occur:

1. HCP NodePool is created via frontend request
2. Before the backend side has had time to create the CS Cluster in CS, a HCP NodePool delete request is performed through the frontend. At that point there's no ClusterServiceID available

Another example scenario would be:
1. HCP NodePool is created via frontend request
2. The backend work to create the cluster using CS is not able to progress for some reason (cs unavailable, some sort of issue or prerequisites to perform the cs cluster create call prevent the backend side from issuing the cs cluster create request, or simply the CS create call permanently fails for some reason 
4. A HCP NodePool delete request arrives to the frontendBefore the backend side has had time to create the CS Cluster in CS, a HCP NodePool delete request is performed through the frontend. At that point there's no ClusterServiceID available

Which means having an alternative design in the frontend side that assumes ClusterServiceID will always be there on Delete and/or Update,Replace operations is incorrect.

miguelsorianod:
As an important insight, we should move delete before and/or at the same time than create because of https://github.com/Azure/ARO-HCP/pull/5142#discussion_r3210546381

frontend/pkg/frontend/node_pool.go
mbarnes: This is where cancellation of ongoing node pool tasks is key.  Today that happens just above with the `database.CancelActiveOperations` call, which is where I think it should stay.

So to revise:

1. HCP NodePool create operation is created via frontend request
2. Before the backend side has had time to create the CS Cluster in CS, a HCP NodePool delete request is performed through the frontend.  The frontend immediately cancels the previous create operation.
3. Because there is no ClusterServiceID available in the NodePool document from Cosmos DB, that informs the frontend that the resource was never actually created.  Instead of initiating an async delete operation, the frontend simply responds with 200.  (Still executing the transaction with the cancellations, though.)
4. Because the NodePool create operation now has a terminal status (Canceled), the backend controllers won't touch it.

Similarly for the second scenario.

There's probably a non-zero chance of some race causing Clusters Service to create an orphaned cluster in this scenario. If the "create node pool" dispatch controller already has the operation at the same time it's being cancelled, it might manage to call Clusters Service.  But orphaned resources have always been possible.  The mismatch controllers can take care of that.

mbarnes: Correction:
> 3. Because there is no ClusterServiceID available in the NodePool document from Cosmos DB, that informs the frontend that the resource was never actually created.  Instead of initiating an async delete operation, the frontend simply responds with 200.  (Still executing the transaction with the cancellations, though.)

Frontend would also delete the NodePool document straight away here.

miguelsorianod:
Closing as it's superseded by other MRs

PR #5145, opened 2026-05-06
Revert "initial commit of pko orphaned resources cleanup"

raelga:
https://redhat.atlassian.net/browse/AROSLSRE-781

Reverts Azure/ARO-HCP#5115

### Why

The `cleanup-pko-resources` shell step failed during INT EV2 rollout (Management.Infra.management.cleanup-pko-resources-uksouth-1). The script crashed while patching finalizers on `clusterpackages.package-operator.run/package-operator`.

EV2 rollout: https://ra.ev2portal.azure.net/#/rollouts/Test/b8e9ef87-cd63-4085-ab14-1c637806568c/Microsoft.Azure.ARO.HCP.GlobalBuildout/ab4ed86d-cfb4-438d-ad7d-876922d90c21/list/Medium.uksouth/Management.Infra.management.cleanup-pko-resources-uksouth-1

Reverting to unblock INT rollout.

PR #5147, opened 2026-05-06
dev-infrastructure: make PKO cleanup script best-effort (AROSLSRE-782)

raelga:
https://redhat.atlassian.net/browse/AROSLSRE-784

### What

Make the PKO orphaned resources cleanup script best-effort so it never blocks EV2 rollouts.

### Why

The original script (#5115) was reverted (#5145) because a failure during finalizer patching caused the shell step to exit non-zero, blocking the INT EV2 rollout. The cleanup is a housekeeping task, not a deployment prerequisite — it should log errors and move on.

### Changes from the original script

| Original | Best-effort |
|---|---|
| `set -o errexit` | Removed — errors don't abort |
| `return 1` in `count_crs()` | Returns `0` and logs error |
| `exit 1` on stuck CRs | Logs error, continues |
| Silent failures | `log_error()` counts and reports all errors |
| No stderr suppression on CRD discovery | Added `2>/dev/null` |
| Exit code reflects success/failure | Always `exit 0` |

### Summary output

On success:
```
=== PKO resource cleanup complete ===
```

On partial failure:
```
=== PKO cleanup completed with 3 error(s) (best-effort, not blocking rollout) ===
```

### Testing

- Script always exits 0 regardless of kubectl errors
- Errors are logged to stderr with `ERROR:` prefix for easy grep in EV2 logs
- Error count tracked and reported in summary line

### DIFF
--- a/dev-infrastructure/scripts/cleanup-pko-resources.sh
+++ b/dev-infrastructure/scripts/cleanup-pko-resources.sh
@@ -59,15 +59,15 @@ count_crs() {
 
   if [[ "${scope}" == "Namespaced" ]]; then
     output=$(kubectl get "${resource}" --all-namespaces -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null) || {
-      log_error "kubectl get ${resource} --all-namespaces failed"
+      echo "ERROR: kubectl get ${resource} --all-namespaces failed" >&2
       echo 0
-      return
+      return 1
     }
   else
     output=$(kubectl get "${resource}" -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null) || {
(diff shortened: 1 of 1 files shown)


hbhushan3 (commented):
count_crs is called via command substitution (count=$(count_crs ...)), which runs the function in a subshell. Any ERRORS=$((ERRORS + 1)) inside that subshell is thrown away when the substitution returns — so kubectl-get failures during the polling loop are logged to stderr but never counted, and the final summary line falsely reports success: "=== PKO resource cleanup complete ===".   Maybe capture the exit status of the function instead of trying to modify a global variable inside a subshell and then in the calling loop, handle the error code to modify the global ERRORS variable in the main shell context?

raelga:
Superseeded by https://github.com/Azure/ARO-HCP/pull/5149

/close

PR #5150, opened 2026-05-06
Bump aro-tools: adding optional commit field to ProwJobSteps

Mustafa-Ali-code:
### What

Add an optional `commit` field to `ProwJobStep` in the pipeline types and JSON schema. sdp-pipelines will use this value as `--base-sha` in the `prowjobexecutor`, allowing Prow to check out a specific source commit instead of HEAD

### Why

Classic E2E pipeline run Prow jobs against ARO-RP master HEAD, which can diverge from the RP version deployed in the environment. This causes test failures when the E2E test expects features or flags that the deployed RP doesn't have yet


dev-infrastructure/modules/postgres/postgres.bicep
rachelvweber, line 85: Was this change added inadvertently?

PR #5153, opened 2026-05-06
Remove Cincinnati upgrade-graph validation for nodepool upgrades

adalrsjr1:
Fixes: [ARO-26792](https://redhat.atlassian.net/browse/ARO-26792)

(updated)

 ### What

  Replace Cincinnati upgrade-edge validation with a version existence check in the nodepool version controller. The backend now verifies the desired version exists in
  Cincinnati but no longer requires an upgrade edge from the current version. Static Kubernetes version skew constraints remain enforced (nodepool ≤ CP version, no downgrade,
   within minor+1).

  Per @deads2k suggestion, this work is split into two PRs:
  1. **(this PR)** Remove edge requirement, keep version existence check
  2. Remove downgrade restriction for nodepool versions (#5184)

  ### Why

  HCP nodepools use the Replace upgrade strategy — nodes are destroyed and recreated at the target version, not upgraded in-place. Cincinnati upgrade edges are irrelevant for
   this strategy.

  The controller was querying Cincinnati for every nodepool upgrade and requiring an edge from the current to the desired version. When no edge existed (e.g. 4.20.8 →
  4.21.1), it entered a permanent error-retry loop. ARM reported `Succeeded` while the upgrade never progressed, leaving customers with no visibility into the failure.

  ### Testing

  - **Unit tests**: Replaced edge-check mocks with existence-check mocks. Added `SucceedsWithoutCincinnatiEdge` (version exists but no edge → succeeds) and
  `VersionNotInCincinnatiFails` (version doesn't exist → fails). Updated `DesiredVersionUnchangedOnFailure` to test VersionNotFound instead of missing edges.
  - **E2E**: Existing y-stream upgrade test (4.20.z → 4.21.zLatest) validates the fix end-to-end. Verified locally against a dev cluster — nodepool upgraded from 4.20.20 to
  4.21.14 successfully.

  **Depends on CS [MR !322](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/322)** — CS also validates upgrade edges via `AvailableUpgrades`
  in `ValidateNodePoolUpgradeVersion()`. Without the CS change, the backend accepts the version but CS blocks the upgrade policy. Safe to deploy in any order — neither change
   alone creates a regression.

  A rationale comment in `validateDesiredNodePoolVersion` notes that if an InPlace upgrade strategy is introduced in the future, strategy-specific upgrade validation should
  be added.

### DIFF
--- a/backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
+++ b/backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
@@ -260,7 +260,8 @@ func prependActiveVersionIfChanged(currentVersions []api.HCPNodePoolActiveVersio
 //   - The desired version exists in Cincinnati
 //
 // Cincinnati upgrade-edge validation is intentionally skipped — HCP nodepools use the Replace
-// strategy (destroy + recreate), so only version existence matters. See OCM-16815, ARO-26792.
+// strategy (destroy + recreate), so only version existence matters.
+// See https://hypershift.pages.dev/reference/nodepool-rollouts/#upgrade-types
 //
 // Returns nil if the desired version is valid, or an error describing why it's invalid.
 func (c *nodePoolVersionSyncer) validateDesiredNodePoolVersion(
--- a/backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
+++ b/backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
@@ -100,7 +100,7 @@ func NewNodePoolVersionController(
 //   - Not less than the highest active version (no downgrades)
 //   - Not greater than the lowest control plane version (node pools cannot exceed CP)
 //   - Not skip minor versions (only z-stream or +1 minor allowed)
-//   - Have a valid upgrade path in Cincinnati from the current version
+//   - Exist as a known version in Cincinnati
 //   - If valid, stores it in ServiceProviderNodePool.Spec.NodePoolVersion.DesiredVersion
 //
 // If the desired version is already among the active versions, validation is skipped
@@ -251,13 +251,16 @@ func prependActiveVersionIfChanged(currentVersions []api.HCPNodePoolActiveVersio
 	return newVersions
 }
 
-// validateDesiredNodePoolVersion checks that the desired node pool version is a valid upgrade path.
+// validateDesiredNodePoolVersion checks that the desired node pool version is a valid upgrade.
 // It validates:
 //   - The desired version is not less than the highest active node pool version (no downgrades)
 //   - The desired version is not greater than the lowest control plane version
 //   - No major version changes (unless FeatureExperimentalReleaseFeatures is registered)
 //   - No minor versions are skipped
-//   - An upgrade path exists from the current version (all the activeVersions) to the desired version (via Cincinnati)
+//   - The desired version exists in Cincinnati
+//
+// Cincinnati upgrade-edge validation is intentionally skipped — HCP nodepools use the Replace
+// strategy (destroy + recreate), so only version existence matters. See OCM-16815, ARO-26792.
 //
 // Returns nil if the desired version is valid, or an error describing why it's invalid.
 func (c *nodePoolVersionSyncer) validateDesiredNodePoolVersion(
@@ -290,22 +293,19 @@ func (c *nodePoolVersionSyncer) validateDesiredNodePoolVersion(
 		return err
 	}
 
-	// Validate upgrade path exists in Cincinnati for ALL active node pool versions
-	for _, activeVersion := range nodePoolActiveVersions {
-		if activeVersion.Version != nil && !desiredVersion.EQ(*activeVersion.Version) {
-			if err := c.validateUpgradePathAvailable(ctx, activeVersion.Version, desiredVersion, channelGroup, clusterUUID); err != nil {
-				return fmt.Errorf("no valid upgrade path from active version %s: %w", activeVersion.Version.String(), err)
-			}
-		}
+	// Validate the desired version exists in Cincinnati (not that an edge exists from the current version).
+	if err := c.validateVersionExistsInCincinnati(ctx, desiredVersion, channelGroup, clusterUUID); err != nil {
+		return err
 	}
 
 	return nil
 }
 
-// validateUpgradePathAvailable checks that an upgrade path exists from current to desired version.
-func (c *nodePoolVersionSyncer) validateUpgradePathAvailable(
+// validateVersionExistsInCincinnati checks that the desired version exists as a node in the
+// Cincinnati update graph.
+func (c *nodePoolVersionSyncer) validateVersionExistsInCincinnati(
 	ctx context.Context,
-	currentVersion, desiredVersion *semver.Version,
+	version *semver.Version,
 	channelGroup string,
 	clusterUUID uuid.UUID,
 ) error {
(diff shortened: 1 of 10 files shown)


test/e2e/nodepool_version_upgrade.go
```diff
@@ -176,12 +179,12 @@ var _ = Describe("Customer", func() {
 			})
 			Expect(parseableVersions).NotTo(BeEmpty(), "no clusterversion status.history entry with valid parseable version")
 
-			By("computing nodepool desired version from Cincinnati (lowest <= lowest control plane version with upgrade path from nodepool)")
+			By("selecting nodepool upgrade target from available versions (<= lowest control plane version)")
 			candidates, err := framework.GetUpgradeCandidatesInMaxMinorFromCincinnati(ctx,
```
swiencki, line 184: This still picks the target via `GetUpgradeCandidatesInMaxMinorFromCincinnati`, which requires a Cincinnati edge to exist. The bug we're fixing is the no-edge case, so the test will Skip when it matters most. ARO-26792 AC #2 specifically asks for a no-edge test. Could we add one?

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
```diff
@@ -214,13 +180,13 @@ func (c *nodePoolVersionSyncer) SyncOnce(ctx context.Context, key controllerutil
...
 
-	subscription, err := c.cosmosClient.Subscriptions().Get(ctx, cluster.ID.SubscriptionID)
+	subscription, err := c.cosmosClient.Subscriptions().Get(ctx, key.SubscriptionID)
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("failed to get subscription: %w", err))
 	}
 
 	// Validate the customer's desired version before setting it
-	if err := c.validateDesiredNodePoolVersion(ctx, &customerDesiredVersion, existingServiceProviderNodePool, existingServiceProviderCluster, subscription, nodePool.Properties.Version.ChannelGroup, clusterKey, clusterUUID); err != nil {
+	if err := c.validateDesiredNodePoolVersion(ctx, &customerDesiredVersion, existingServiceProviderNodePool, existingServiceProviderCluster, subscription); err != nil {
```
swiencki: Cincinnati was also implicitly checking the version exists in the channel. After this PR a non-existent version like 4.20.999 passes the static gate. Is CS the authoritative availability check now? Worth being explicit so we don't regress into the same "ARM Succeeded, never progresses" shape for a different reason.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
```diff
@@ -258,13 +224,18 @@ func prependActiveVersionIfChanged(currentVersions []api.HCPNodePoolActiveVersio
...
 //   - The desired version is not less than the highest active node pool version (no downgrades)
 //   - The desired version is not greater than the lowest control plane version
 //   - No major version changes (unless FeatureExperimentalReleaseFeatures is registered)
 //   - No minor versions are skipped
-//   - An upgrade path exists from the current version (all the activeVersions) to the desired version (via Cincinnati)
+//
+// Cincinnati upgrade-graph validation is intentionally not performed here. HCP nodepools use the
+// Replace upgrade strategy — nodes are destroyed and recreated at the target version, not upgraded
+// in-place — so Cincinnati upgrade edges are irrelevant. Only Kubernetes version skew constraints
+// matter. If an InPlace upgrade strategy is introduced in the future, Cincinnati validation should
```
swiencki: Small thing: the comment implies Cincinnati can just be added back here for InPlace. But this PR also drops the cluster UUID plumbing from #5045, so it'd need both. Maybe phrase as "strategy-specific validation should be added for InPlace"?

swiencki (commented):
Three comments inline. Substance-only review across rubber-duck + GPT-5.5 + Opus 4.7 1M; the headline finding is the test-coverage gap vs ARO-26792 AC #2.

machi1990 (dismissed):
1. I personally still on the fence about this change and I'd have kept the validation and give customer two options:
- Either they recreate their node pool if they don't care about "validated" upgradeability
- Or they do hops from v1, v2...vDesired

I understand that these might not always be better UX but I personally feel like they are better options that doesn't bypass upgrades gates. 
If this use case gets an approval from BU & @deads2k I won't stand on the way of it. Please let's get that approval from them. Also I am looping in @ahitacat for review as well.

2. CS component is also involved in the upgrade and has the upgradeability check; can you double check if there is anything that needs to be done on the CS component as well?

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
ahitacat: CS checks that the version exists but we so also check it in here. Otherwise we will have the same issue, as this validation will pass for a well formed but not existed version (4.20.999) but fail "silently" to the user in CS. 

So, we need to make a call here to Cincinnati to check that the version exists.

ahitacat (commented):
In addition to @machi1990 review https://github.com/Azure/ARO-HCP/pull/5153#pullrequestreview-4241486250

In CS when creating the upgrade policy we also check that the version has an upgradepath. That should be change before this, so otherwise we could have the same issue, as the controller that creates these policies (`trigger_nodepool_upgrade_controller`) will fail.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
jboutaud: Not sure what we do for deployment/install but why not just use the same validation as we do there? Is that different from this flow? I would expect that this would prevent both of those situations

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
```diff
@@ -48,13 +40,9 @@ import (
 // from CS and internal and helps selecting a valid desiredVersion within the user's
 // desired
 type nodePoolVersionSyncer struct {
-	cooldownChecker                       controllerutils.CooldownChecker
-	clusterManagementClusterContentLister listers.ManagementClusterContentLister
-	cosmosClient                          database.DBClient
-	clusterServiceClient                  ocm.ClusterServiceClientSpec
-
-	cincinnatiClientLock      sync.RWMutex
-	clusterToCincinnatiClient *lru.Cache
```
deads2k: this can't be right. We must still contact cincinnati to determin
1. does the version exist
2. eventually: does the version have a warning against usage on ARO.

test/e2e/nodepool_version_upgrade.go
deads2k: simplest no-edge is a downgrade. Let's do that.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
ahitacat: We don't have this for install. The creation of the nodepool is not yet done in the backend, and adding the cincinnati call to the frontend will increase our time response. Let's add it here, and once we have the node pool creation call in the backend we can add the same validation there.

adalrsjr1:
per @deads2k suggestion this change was split into two parts (part 2 in #5184)

1. pr/1 (this one) remove the requirement for an edge, but keep the requirement for "version exists in cincinnati"
2. pr/2 (#5184) remove the restriction about moving nodepool versions backwards and update testing to include one example of doing this (no need to check this for every version)

The previous comments were addressed in this PR. Both PRs are pending on the update on CS ([MR 322](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/322)) FYI @machi1990 @ahitacat

swiencki (commented):
Reviewed, code looks good, no blockers from my side. Concerns in https://github.com/Azure/ARO-HCP/pull/5153#pullrequestreview-4241486250 should be addressed. Reviewed the CS MR too (https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/322) with one non-blocking note.

deads2k:
/approve

I'll leave lgtm to another reviewer.

ahitacat (commented):
After this is merged we can remove the check of the upgrade path. I.e. so we don't skip this.
https://github.com/Azure/ARO-HCP/blob/main/test/e2e/nodepool_version_upgrade.go#L184

Also, I think is worth it to test this with an e2e. Right now from 4.20.8 -> 4.20.9 there is no upgrade path. We could have an e2e that checks in cincinnati a case like this and upgrades succesfully. Can you add it (it should be skipped before this gets merge but we can test it)

adalrsjr1:
> After this is merged we can remove the check of the upgrade path. I.e. so we don't skip this. https://github.com/Azure/ARO-HCP/blob/main/test/e2e/nodepool_version_upgrade.go#L184
> 
> Also, I think is worth it to test this with an e2e. Right now from 4.20.8 -> 4.20.9 there is no upgrade path. We could have an e2e that checks in cincinnati a case like this and upgrades succesfully. Can you add it (it should be skipped before this gets merge but we can test it)

@ahitacat 
* added the e2e test and marked to skip for now; to be allowed in the upcoming PR #5184 
* https://github.com/Azure/ARO-HCP/blob/main/test/e2e/nodepool_version_upgrade.go#L184; to be removed in the upcoming PR #5184 

I also addressed your comments on [CS 322](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/322); pending on your approval.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
```diff
@@ -246,13 +246,16 @@ func prependActiveVersionIfChanged(currentVersions []api.HCPNodePoolActiveVersio
...
 // It validates:
 //   - The desired version is not less than the highest active node pool version (no downgrades)
 //   - The desired version is not greater than the lowest control plane version
 //   - No major version changes (unless FeatureExperimentalReleaseFeatures is registered)
 //   - No minor versions are skipped
-//   - An upgrade path exists from the current version (all the activeVersions) to the desired version (via Cincinnati)
+//   - The desired version exists in Cincinnati
+//
+// Cincinnati upgrade-edge validation is intentionally skipped — HCP nodepools use the Replace
+// strategy (destroy + recreate), so only version existence matters. See OCM-16815, ARO-26792.
```
machi1990: Can we link the hypershift docs on what Replace strategy is?

machi1990: https://hypershift.pages.dev/reference/nodepool-rollouts/#:~:text=without%20replacing%20Nodes-,Upgrade%20Types,-The%20upgrade%20type

machi1990 (commented):
lgtm

As @swiencki as been actively reviewing it, I'll leave the final /lgtm to @swiencki

adalrsjr1:
Dependency on CS done, the changes has been [merged](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/322#note_21353514)

swiencki:
LGTM

geoberle:
/hold
CS fix 322 will merge in with https://github.com/Azure/ARO-HCP/pull/5170, we can remove the hold then

adalrsjr1:
failing because this PR got stale for too long.. need rebase

adalrsjr1:
/retest 

53 minutes waiting for a free identity container

PR #4887, opened 2026-04-15
Revert "Merge pull request #4645 from Azure/aklymenk-aro-23768"

mbukatov:
This reverts commit 33aff3985bd73ca395cf4361fe7efbf733dec333, reversing changes made to 4474d75962290d7c91cd237128c42bd4ff4894d2.

Preventing non deterministic failures, follow fixes in error reporting or service quality tracked in [ARO-26082](https://redhat.atlassian.net/browse/ARO-26082)

https://redhat.atlassian.net/browse/ARO-26082

### What

Reverting PR https://github.com/Azure/ARO-HCP/pull/4645

### Why

Changes introduced in https://github.com/Azure/ARO-HCP/pull/4645 caused nodepool_update_nodes.go test case to fail on the following error every now and then:

```
nodepool_update_nodes.go:262: Expected success, but got an error: expected 4 nodes, found 5
```

This is happening because now the changes are happening in parallel, and the test doesn't assume that some node deployment or teardowns can take more time, nor it reports the status properly, so it's not possible to quickly understand the nature of the problem.

See [ARO-26082](https://redhat.atlassian.net/browse/ARO-26082) for more context.

### Special notes for your reviewer

Based on discussion during ARO HCP Daily Quality Status Review.

PR #5168, opened 2026-05-07
test: re-parallelize nodepool scaling operations in nodepool_update_nodes E2E with node status diagnostics on failure

aklymenk:
[ARO-26082](https://redhat.atlassian.net/browse/ARO-26082)

### What

Re-applies the parallel node pool scaling approach from #4645 (reverted in #4887)
with improved failure diagnostics to investigate the non-deterministic failures
tracked in [ARO-26082](https://redhat.atlassian.net/browse/ARO-26082).

### Why

Changes:
- **Parallel mutations restored**: scale-down (2→1 replicas), scale-up (1→2 replicas), and
  enable-autoscaling (1→0 replicas, min=1, max=2) run concurrently
  on 3 dedicated nodepools (`np-scale-down`, `np-scale-up`, `np-autoscale`).
- **Node status dump on count mismatch**: when the final `VerifyNodeCount` fails,
  the test now lists every node with its name, Ready/NotReady condition, and
  Schedulable/Unschedulable state. This makes it possible to identify which
  node pool contributed the extraneous node.
- Removed `labels.Slow` label.

### Testing

`nodepool_update_nodes.go` test was executed 11 times in personal development environment in order to catch the same error described in [ARO-26082](https://redhat.atlassian.net/browse/ARO-26082) in order to understand which node pool has extraneous node, but the test was passing all the time.

### DIFF
--- a/test/e2e/nodepool_update_nodes.go
+++ b/test/e2e/nodepool_update_nodes.go
@@ -16,12 +16,18 @@ package e2e
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	"golang.org/x/sync/errgroup"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 
@@ -37,12 +43,29 @@ var _ = Describe("Customer", func() {
 		labels.High,
 		labels.Positive,
 		labels.AroRpApiCompatible,
-		labels.Slow,
 		func(ctx context.Context) {
 			const (
-				customerClusterName  = "np-update-nodes-hcp-cluster"
-				customerNodePoolName = "np-update-nodes"
-				oneNodePoolName      = "np-one-node"
+				customerClusterName = "hcp-cluster"
+
+				scaleDownNodePoolName = "np-scale-down"
+				scaleUpNodePoolName   = "np-scale-up"
+				autoscaleNodePoolName = "np-autoscale"
+
+				deployedNodePoolsNumber = 3
+
+				scaleDownNodePoolInitialReplicas = 2
+				scaleDownNodePoolUpdatedReplicas = scaleDownNodePoolInitialReplicas - 1
+
+				scaleUpNodePoolInitialReplicas = 1
+				scaleUpNodePoolUpdatedReplicas = scaleUpNodePoolInitialReplicas + 1
+
+				autoscaleNodePoolInitialReplicas = 1
+				autoscaleNodePoolUpdatedReplicas = 0
+				autoscaleNodePoolMinReplicas     = 1
+				autoscaleNodePoolMaxReplicas     = 2
+
+				clusterInitialNodeCount = scaleDownNodePoolInitialReplicas + scaleUpNodePoolInitialReplicas + autoscaleNodePoolInitialReplicas
+				clusterFinalNodeCount   = scaleDownNodePoolUpdatedReplicas + scaleUpNodePoolUpdatedReplicas + autoscaleNodePoolMinReplicas
 			)
 
 			tc := framework.NewTestContext()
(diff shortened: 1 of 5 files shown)


mvacula02 (approved):
/lgtm

please give a heads up to the wider team once this gets merged as it's expected to cause failures

mvacula02:
/lgtm cancel
sorry for that, I'm just testing if my repo access works

mvacula02 (approved):
/lgtm

test/e2e/nodepool_update_nodes.go
```diff
@@ -37,12 +43,29 @@ var _ = Describe("Customer", func() {
 		labels.High,
 		labels.Positive,
 		labels.AroRpApiCompatible,
-		labels.Slow,
 		func(ctx context.Context) {
 			const (
-				customerClusterName  = "np-update-nodes-hcp-cluster"
-				customerNodePoolName = "np-update-nodes"
-				oneNodePoolName      = "np-one-node"
+				customerClusterName = "hcp-cluster"
```
mgahagan73, line 48: Why change the cluster name here? hcp-cluster sounds very generic and I think the old name or something similar  like update-nodes-hcp-cluster will make debugging easier whenever we have to dig through kusto queries to root cause issues.

test/e2e/nodepool_update_nodes.go
```diff
@@ -117,106 +141,166 @@ var _ = Describe("Customer", func() {
...
+							schedulableStr := "Schedulable"
+							if node.Spec.Unschedulable {
+								schedulableStr = "Unschedulable"
+							}
+							nodeStatuses = append(nodeStatuses, fmt.Sprintf("%s (%s, %s)", node.Name, readyStr, schedulableStr))
+						}
+						err = fmt.Errorf("%w; node details: %s", err, strings.Join(nodeStatuses, "; "))
+					}
+				}
+				Expect(err).NotTo(HaveOccurred())
```
mbukatov, line 300: The additional listing, in case of a failure, should be done in the verifier code, so that any verifier which fails on this will include these details, not just this particular case.

mbukatov (changes requested):
Requesting a minor change.

test/e2e/nodepool_update_nodes.go
mbukatov: Agreed, I would also rather keep the old name unless we have a good reason to do otherwise.

test/e2e/nodepool_update_nodes.go
```diff
@@ -117,106 +141,166 @@ var _ = Describe("Customer", func() {
...
-			for createErr := range errCh {
-				creationErrors = append(creationErrors, createErr)
+			_ = nodePoolCreateErrGroup.Wait() // ignoring error here, will collect all node pool creation errors in the nodePoolCreationErrCh channel
+			close(nodePoolCreationErrCh)
+			var nodePoolCreationErrors []error
+			for nodePoolCreationErr := range nodePoolCreationErrCh {
+				nodePoolCreationErrors = append(nodePoolCreationErrors, nodePoolCreationErr)
 			}
-			Expect(creationErrors).To(BeEmpty(), "nodepool creation errors: %v", creationErrors)
+			Expect(nodePoolCreationErrors).To(BeEmpty(), "node pool creation errors: %v", nodePoolCreationErrors)
```
mbukatov, line 155: Could you just combine the errors in the list into a single error using [errors.Join](https://pkg.go.dev/errors#Join) and then just check `Expect(err).NotTo(HaveOccurred())`. This will make sure all the errors are reported in the same way we usually report errors.

mbukatov (commented):
Few more comments as I promised yesterday.

I see you note that you haven't seen the issue with unexpected number of nodes again, but I wonder have you looked into the possible reasons for this even when you haven't observed it again? In the jira I noted:

> The reasons for non determinism of the test should be investigated, with a possible further checks for the actual behaviour of the service (just in case):
> * the autoscaler may still be handling the scaling after the node pool update is declared finished (eg. node may still be deprovisioning)
> * the autoscaler may scale up to 2 instead of 1 as the test case expects

Were you able to rule out any of these options?

davidffrench:
Closing — while the revert (#4887) is still in effect and scaling remains sequential, this PR needs a significant rebase and the investigation into the flakiness ([ARO-26082](https://redhat.atlassian.net/browse/ARO-26082)) would need to be revisited. If someone wants to re-parallelize scaling, it would be easier to start fresh. Thank you Andrii!

raelga (commented):
/close

PR #5173, opened 2026-05-07
feat: make internal externalauth ClusterServiceID attribute a pointer

miguelsorianod:
We change the ServiceProviderProperties.ClusterServiceID attribute of the internal/api.HCPOpenShiftClusterExternalAuth Go type from a InternalID to a *InternalID. When we move the writes to CS from frontend to backend this will be initially unset. By making it a pointer we make it easier for readers to understand it could be unset as opposed to an empty string.

The commit also changes the cosmos-internal two way conversion to be transparent.


frontend/pkg/frontend/external_auth.go
```diff
@@ -470,23 +471,25 @@ func (f *Frontend) updateExternalAuthInCosmos(ctx context.Context, writer http.R
...
-	csExternalAuthBuilder, err := ocm.BuildCSExternalAuth(ctx, newInternalExternalAuth, true)
-	if err != nil {
-		return utils.TrackError(err)
-	}
-
-	logger.Info(fmt.Sprintf("updating resource %s", oldInternalExternalAuth.ID))
-	_, err = f.clusterServiceClient.UpdateExternalAuth(ctx, oldInternalExternalAuth.ServiceProviderProperties.ClusterServiceID, csExternalAuthBuilder)
-	if err != nil {
-		return utils.TrackError(err)
+	// TODO is this check correct? what happens if it's nil at this point?
```
miguelsorianod: See this. I am not sure if this is the appropriate way to deal with it in this path

frontend/pkg/frontend/external_auth.go
```diff
@@ -470,23 +471,25 @@ func (f *Frontend) updateExternalAuthInCosmos(ctx context.Context, writer http.R
...
+		}
 	}
 
 	transaction := f.dbClient.NewTransaction(oldInternalExternalAuth.ID.SubscriptionID)
 
 	externalAuthUpdateOperation := database.NewOperation(
 		database.OperationRequestUpdate,
 		newInternalExternalAuth.ID,
-		newInternalExternalAuth.ServiceProviderProperties.ClusterServiceID,
+		ptr.Deref(newInternalExternalAuth.ServiceProviderProperties.ClusterServiceID, api.InternalID{}),
```
miguelsorianod, line 495: See this. I am not sure if this is the appropriate way to deal with it in this path

frontend/pkg/frontend/external_auth.go
```diff
@@ -571,18 +574,21 @@ func (f *Frontend) DeleteExternalAuth(writer http.ResponseWriter, request *http.
...
-		// StatusNotFound means we have stale data in Cosmos DB.
-		// This can happen in test environments if a user bypasses
-		// the RP to delete a resource (e.g. "ocm delete"). It can
-		// also happen if an asynchronous deletion operation fails.
-		// we will fall through and cancel all operations and go through as normal a deletion flow as we can to avoid
-		// leaking data related to the resource, like controller status.
-		logger.Info("clusterService externalauth missing, trying to clean up", "err", err)
-	} else if err != nil {
-		return utils.TrackError(err)
+	// TODO is this check correct? what happens if it's nil at this point?
```
miguelsorianod: See this. I am not sure if this is the appropriate way to deal with it in this path

frontend/pkg/frontend/external_auth.go
```diff
@@ -615,10 +621,15 @@ func (f *Frontend) addDeleteExternalAuthToTransaction(ctx context.Context, write
 		return utils.TrackError(err)
 	}
 
+	// TODO is this check correct? what happens if it's nil at this point?
+	externalAuthClusterServiceID := api.InternalID{}
```
miguelsorianod: See this. I am not sure if this is the appropriate way to deal with it in this path

internal/database/convert_externalauth.go
```diff
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"strings"
```
miguelsorianod, line 20: One of the important parts of the PR

frontend/pkg/frontend/external_auth.go
deads2k: as noted in the nodepool one.  Fail for now.

frontend/pkg/frontend/external_auth.go
deads2k: this is ok.

frontend/pkg/frontend/external_auth.go
deads2k: as above, fail for now.  Safer for the transition.

PR #5174, opened 2026-05-07
stop serializing old cluster fields

deads2k:
This is part three of of the plan in https://github.com/Azure/ARO-HCP/pull/5095 and it stops serializing unnecessary fields.  Cannot merge until https://github.com/Azure/ARO-HCP/pull/5171 goes to prod


mbarnes:
I should be back before this merges, but cursory read of the last commit LGTM.

deads2k:

PR #5175, opened 2026-05-07
operator logging and diagnostic improvements

mgahagan73:
https://redhat.atlassian.net/browse/ARO-26756

### What

Gather more diagnostic info when operator installation fails, add timing metrics. Move EventuallyVerify() into framework.

### Why

Currently if an operator fails to install due to timeout we get no useful debugging information beyond the timeout. Also we currently do not have an accurate picture of how long the operator installation takes.

### DIFF
--- a/test/util/verifiers/basic.go
+++ b/test/util/verifiers/basic.go
@@ -31,6 +31,13 @@ type HostedClusterVerifier interface {
 	Verify(ctx context.Context, restConfig *rest.Config) error
 }
 
+// DiagnosticVerifier is an optional interface that verifiers can implement to
+// provide additional context when EventuallyVerify times out. The returned
+// string is logged alongside the timeout error.
+type DiagnosticVerifier interface {
+	DiagnoseFailure(ctx context.Context, restConfig *rest.Config) string
+}
+
 // ExpectForbidden returns a verifier that runs inner and succeeds only when inner.Verify
 // returns a forbidden (403) error. Use it to assert that an identity must not be able
 // to perform the action implemented by inner, without defining separate "Cannot" verifiers.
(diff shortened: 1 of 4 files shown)


mgahagan73:
INT run shows elapsed time logging of operator installation now working (took around 1 minute)

mvacula02 (approved):
/lgtm

The only nit I have is that a file named _eventually.go_ might be confusing since Eventually is a Ginkgo keyword

test/util/verifiers/basic.go
```diff
@@ -31,6 +31,13 @@ type HostedClusterVerifier interface {
 	Verify(ctx context.Context, restConfig *rest.Config) error
 }
 
+// DiagnosticVerifier is an optional interface that verifiers can implement to
+// provide additional context when EventuallyVerify times out. The returned
+// string is logged alongside the timeout error.
+type DiagnosticVerifier interface {
+	DiagnoseFailure(ctx context.Context, restConfig *rest.Config) string
+}
```
mbukatov, line 39: I understand you do this because you moved out the EventuallyVerify so that you can run a verifier in a loop, so that after it fails, you need to run something else which will fetch the additional evidence. 

That said, this design a bad idea, because it makes the verifier interface more complex.

As a user of a verifier, running it in a test, I just want to create a verifier, and then add it into a list of other verifiers to be checked in parallel. I don't want to think whether this particular one needs bot be run in a loop via EventuallyVerify.

Moreover, each single verifier should diagnose and report a failure directly when an error is noticed. This should not happen elsewhere, but directly in a verifier. If a verifier fail and we don't get good evidence about what went wrong, that should be fixed in a particular verifier, because it's it's responsibility.

I would rather seen this implemented this way:

- move the eventually loop into a framework function
- use this loop within your verifier, so that one can just use this verifier via `err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyNodesReady(), verifiers.VerifyFoo(), verifiers.VerifyBar())`
- move the diagnostic functions into framework, so that these can be called from your verifiers, and other places

mbukatov (commented):
I would rather not increase complexity of our verifier interface.

Sorry for not providing this feedback over github sooner, I commented about it on slack when you asked for a review, but didn't get to look into this in more detail sooner, before it was merged.

Next time I will provide the comments here directly.

PR #5178, opened 2026-05-07
test: group node pool verifiers into a single one

bennerv:
### What

Groups the various node pool verifiers we have into a single one

### Why

We have duplicated node pool verification.  Additionally, the node pool verification should be consistent across any node pool created.  

Now this verifies: 
- node is ready
- node is scheduable
- can fetch logs on a pod running on the node - router
- node replica count matches nodes in cluster

### Testing

green e2e

This is fallout from the konnectivity pods failing during our swift rollout.  We failed for the kas to dial to kubelet due to konnectivity-agent having the wrong TLS cert bundle as we were setting domains when we shouldn't have been. 

This will test we're always able to get node logs when a node is running.


bennerv:
/close 

Stale.  No longer needed

PR #5180, opened 2026-05-07
use generic storage for clusters

deads2k:
part 4 of https://github.com/Azure/ARO-HCP/pull/5095

Can merge at the same time as https://github.com/Azure/ARO-HCP/pull/5174, separated for ease of review.


internal/database/convert_generic.go
```diff
@@ -93,3 +97,7 @@ func CosmosGenericToInternal[InternalAPIType any](cosmosObj *GenericDocument[Int
 
 	return &cosmosObj.Content, nil
 }
+
+type Defaulter interface {
+	EnsureDefaults()
+}
```
mbarnes, line 103: I like this, but I think I'd prefer it in the `api` package and add static checks to the resource types, like:

```
var _ Defaulter = &HCPOpenShiftCluster{}
```

deads2k: > I like this, but I think I'd prefer it in the `api` package and add static checks to the resource types, like:

Currently some do and some don't.  As a future improvement, ok.

deads2k:

PR #4566, opened 2026-03-21
feat: add node pool vs control plane version skew validations during nodepool creation

machi1990:
### What

feat: add node pool vs control plane version skew validation at nodepool creation time

Implements node pool version skew validation at CREATE time so version compatibility
between node pools and control planes is enforced before the pool exists, using the
same rules that already govern UPDATE. Extends the admission path so `ServiceProviderCluster`
context (needed to read active control-plane versions) is available on CREATE, not just UPDATE.

**Changes:**
- Unify CREATE and UPDATE version validation into the existing `validateNodePoolVersionChange`
  (`internal/admission/admit_nodepool.go`), removing the `op.Type == operation.Update` gate that
  previously skipped this check entirely on CREATE
- Distinguish CREATE vs UPDATE-only logic (desired-version skip, active-versions lookup) by
  checking the nil-ness of `oldObj` / `ServiceProviderNodePool` rather than `op.Type`, since most
  restrictions are universal to both operations
- Pass `ServiceProviderCluster` into the node pool admission context for CREATE (not only UPDATE),
  fetching/creating it via `GetOrCreateServiceProviderCluster` (`frontend/pkg/frontend/node_pool.go`)
- `internal/validation/validators.go` is unchanged — `main`'s existing `ValidateNodePoolVersionChange`
  already handles CREATE correctly (empty active versions naturally skip the UPDATE-only branches)
- Update integration test artifacts for CREATE-time skew and UPDATE flows that now require a valid
  initial version

### Why

While working on #4505 and further reviewing #4554, I realized the frontend was missing version skew
validation during nodepool creation. Without this validation, users could create nodepools with
versions incompatible with the control plane, leading to runtime failures.

**Why ServiceProviderCluster is required on CREATE:**
The validation needs access to active control plane versions (not just the cluster's desired version)
because:
- During control plane upgrades, multiple versions may be active simultaneously
- Nodepools must remain compatible with the **lowest** active CP version (cannot exceed it)
- Nodepools can be created during CP upgrades and must respect the active version range

**Validation rules (CREATE, same as UPDATE):**
- Node pool version ≤ lowest active control plane version
- Node pool version within N-2 minor skew of the highest active control plane version
- Cross-major compatibility via allowlist (gated by `FeatureExperimentalReleaseFeatures`)

UPDATE continues to enforce the same skew / upgrade-path rules; this PR makes CREATE subject to the
same compatibility policy instead of skipping validation entirely.

### Testing

[...]

PR #5184, opened 2026-05-07
Allow nodepool version downgrades for HCP clusters

adalrsjr1:
Fixes: [ARO-26792](https://redhat.atlassian.net/browse/ARO-26792)
(updated)

 ### What

 Allow nodepool version downgrades for HCP clusters. Constraints:
 - Same-major downgrades allowed within N-2 minor version skew of control plane
 - Cross-major downgrades gated under `FeatureExperimentalReleaseFeatures` AFEC
 - Upgrades skip minor versions (on [ARO-26793](https://redhat.atlassian.net/browse/ARO-26793)); downgrades can (N-2 provides the bound)

 An E2E test exercises a z-stream downgrade as the simplest no-edge scenario.

 Per @deads2k suggestion, this work is split into two PRs:
 1. Remove edge requirement, keep version existence check ([PR #5153](https://github.com/Azure/ARO-HCP/pull/5153))
 2. **(this PR)** Allow nodepool version downgrades

 ### Why

 HCP nodepools use the Replace upgrade strategy — nodes are destroyed and recreated
 at the target version, not upgraded in-place. A "downgrade" is operationally identical
 to an "upgrade" (both replace nodes). The previous restriction blocked valid rollback
 scenarios.

 Cincinnati graphs have no backward edges, so a downgrade is guaranteed to have no
 Cincinnati upgrade path. This makes downgrade the simplest proof that the no-edge
 scenario works end-to-end.

 ### Testing

 - **Unit tests** (`validators_test.go`, `admit_nodepool_test.go`,
   `nodepool_version_controller_test.go`): z-stream downgrade, y-stream downgrade,
   multi-minor downgrade within N-2, downgrade at N-2 boundary, downgrade beyond N-2
   fails, cross-major downgrade ±AFEC, y-stream upgrade without Cincinnati edge (AC2).
 - **Integration tests**: downgrade beyond N-2 skew returns "must be within 2 minor
   versions" error.
 - **E2E**: New test creates a nodepool at the latest 4.21.z patch and downgrades to
   the previous patch. Verifies nodes are recreated at the target version and the ARM
   GET reflects the downgraded version.

 **Dependencies:**
 - [DONE] [CS MR !326](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/326) 
 - [DONE] [PR #5153](https://github.com/Azure/ARO-HCP/pull/5153) (1/2) — removes Cincinnati
   edge requirement, keeps existence check 
 - [DONE] [CS MR !326](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/326) 
   — skips downgrade and cross-major validation in CS for ARO HCP. Without this, the
   E2E downgrade test will fail.


internal/validation/validators.go
```diff
@@ -877,12 +877,16 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
 
 	lowest, highest := apihelpers.FindLowestAndHighestNodePoolVersion(activeVersions)
 
-	// No partial downgrades: desiredVersion >= highest active version
+	// Downgrade handling: HCP nodepools use Replace strategy (nodes are recreated),
+	// so downgrades are operationally identical to upgrades. Only cross-major
```
deads2k: why would we block this case?

internal/admission/admit_nodepool.go
```diff
@@ -124,11 +124,11 @@ func AdmitNodePoolUpdate(newNodePool, oldNodePool *api.HCPOpenShiftClusterNodePo
 	return errs
 }
 
-// validateNodePoolVersionUpgrade validates that a node pool version change is a valid upgrade.
+// validateNodePoolVersionUpgrade validates that a node pool version change is valid.
 // It checks:
-//   - No downgrade: new version >= old version
-//   - No major version change: new major == old major (unless FeatureExperimentalReleaseFeatures is registered)
-//   - No skipping minor versions: new minor <= old minor + 1
+//   - Downgrades allowed (no cross-major downgrades)
```
deads2k: why stop cross-major downgrade?

test/e2e/nodepool_version_upgrade.go
```diff
@@ -237,4 +237,131 @@ var _ = Describe("Customer", func() {
 			labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible,
 			"4.20", "4.20"),
 	)
+
+	// Nodepool z-stream downgrade: proves the no-edge scenario works. Cincinnati has no
+	// backward edges, so a downgrade exercises a version change without a Cincinnati upgrade
+	// path. HCP nodepools use Replace strategy — nodes are recreated, not upgraded in-place.
+	DescribeTable("should downgrade a nodepool version",
```
deads2k, line 512: We only need to test one to prove the mechanics work. We check all the upgrades to be sure the transitions between versions work, but if the both versions work at the same time in the upgrade, we don't need to retest them all in the downgrade.  Just one will do.

internal/admission/admit_nodepool.go
ahitacat: Maybe it can be under FeatureFlag? I see there are potential unknowns,  i.e. can we have a node pool in 4.23 and a cp in 5.0?

internal/validation/validators.go
```diff
@@ -877,12 +877,16 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
 
 	lowest, highest := apihelpers.FindLowestAndHighestNodePoolVersion(activeVersions)
 
-	// No partial downgrades: desiredVersion >= highest active version
+	// Downgrade handling: HCP nodepools use Replace strategy (nodes are recreated),
```
ahitacat: Now that this allows downgrades, we need to check that the new version is not lower than cp minor-2, this is something hypershift doesn't allow.

Add also a unit test for this.

ahitacat (commented):
This will also required a change in CS as now it also validates against downgrades. 
Additionally there is an integration test related to downgrades https://github.com/Azure/ARO-HCP/tree/main/test-integration/frontend/artifacts/FrontendCRUD/NodePool/version-downgrade

internal/admission/admit_nodepool.go
deads2k: yes, we can.  It's required for migration.

adalrsjr1:
> This will also required a change in CS as now it also validates against downgrades. Additionally there is an integration test related to downgrades https://github.com/Azure/ARO-HCP/tree/main/test-integration/frontend/artifacts/FrontendCRUD/NodePool/version-downgrade

@ahitacat The integration test at t`est-integration/.../version-downgrade/` was updated to test the new N-2 skew constraint (downgrade target changed to 4.17.0, expected error updated). 

**Regarding CS**: edge validation is tracked in CS MR !322. The downgrade check in `ValidateNodePoolUpgradeVersion (line 280)` that fires before the `IsAroHcp()` bypass will be addressed in a follow-up CS I had to prepare to handle it.

internal/validation/validators.go
adalrsjr1: Added N-2 minor version skew validation in `ValidateNodePoolUpgrade` — same-major only. Unit tests added in both `admit_nodepool_test.go` and `nodepool_version_controller_test.go`: boundary succeeds at exactly 2 minors, fails at 3.

internal/admission/admit_nodepool.go
adalrsjr1: Addressed — cross-major downgrades are now gated under `FeatureExperimentalReleaseFeatures`. With the flag registered, a nodepool at 4.x can coexist with CP at 5.x (and vice versa for downgrades). The N-2 skew check is scoped to same-major only — cross-major skew is governed by `AllowControlPlaneNodePoolMajorVersionSkew`.

internal/validation/validators.go
adalrsjr1: Cross-major downgrades are now gated under FeatureExperimentalReleaseFeatures, same as cross-major upgrades. With the flag registered, cross-major downgrades are allowed. Also added N-2 minor version skew validation per @ahitacat 's feedback.

test-integration/frontend/artifacts/FrontendCRUD/NodePool/version-downgrade/07-httpReplace-nodepool-downgrade/expected-error.txt
```diff
@@ -1 +1 @@
-cannot downgrade from version 4.20.8 to 4.20.0
\ No newline at end of file
+must be within 2 minor versions of control plane version
```
machi1990, line 1: Hmm something is off here; isn't the intent to allow downgrade?

internal/validation/validators.go
```diff
@@ -877,33 +879,46 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
 		}
 		return ValidateMajorUpgrade(*lowest, desiredVersion)
 	}
 
-	// Minor skip validation
-	if lowest != nil && desiredVersion.Minor > lowest.Minor+1 {
+	// Minor skip validation (same-major upgrades only). Downgrades don't need minor-skip
+	// blocking — HCP nodepools use Replace strategy (destroy + recreate), so there's no
+	// in-place rollout that requires stepping through intermediate minors. The N-2 lower
+	// bound above provides the downgrade constraint.
```
machi1990: Why don't we allow skipping minor if we are doing a replace?
e.g we should allow someone moving from 4.19.z to 4.21.z without any issue, non?

internal/validation/validators.go
```diff
@@ -862,11 +862,13 @@ func ValidateMajorUpgrade(fromVersion, toVersion semver.Version) error {
...
-// - Cannot exceed lowest control plane version
-// - No major version changes without AFEC (uses existing ValidateMajorUpgrade)
-// - No minor version skipping
+// ValidateNodePoolUpgrade validates a node pool version change (upgrade or downgrade).
+// HCP nodepools use Replace strategy (nodes are destroyed and recreated), so downgrades
+// are operationally identical to upgrades. Constraints:
+//   - Cannot exceed lowest control plane version
+//   - Must be within 2 minor versions of control plane (N-2 skew, same major)
+//   - Cross-major changes require AFEC FeatureExperimentalReleaseFeatures
+//   - Upgrades cannot skip minor versions (downgrades can — N-2 provides the bound)
```
machi1990: Why can 't we skip minors if we are doing a replace strategy?

internal/validation/validators.go
```diff
@@ -862,11 +862,13 @@ func ValidateMajorUpgrade(fromVersion, toVersion semver.Version) error {
...
 func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api.HCPNodePoolActiveVersion, lowestCPVersion *semver.Version, allowMajorUpgrade bool) error {
```
machi1990: Is there a way this could be simplified to:
- remove the case where we were preventing downgrades and leave only the skew checks between control plane + cross major jumps gated under a feature flag?
The way that it is done with conditional IFs (downgrade vs upgrade) made it hard to read;

internal/validation/validators.go
JameelB: is this outdated perhaps? This PR indicates that we will allow to skip minor versions (to +2): https://github.com/Azure/ARO-HCP/pull/5239

internal/validation/validators.go
```diff
@@ -877,33 +879,46 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
+	// Lower bound (N-2 skew): node pool must be within 2 minor versions of control plane.
+	// Analogous to the upper-bound check above; together they define the allowed version window.
+	// Same-major only — cross-major skew is governed by AllowControlPlaneNodePoolMajorVersionSkew.
+	if lowestCPVersion != nil && desiredVersion.Major == lowestCPVersion.Major &&
+		desiredVersion.Minor < lowestCPVersion.Minor &&
+		lowestCPVersion.Minor-desiredVersion.Minor > 2 {
+		return fmt.Errorf(
+			"invalid node pool version %s: must be within 2 minor versions of control plane version %s",
+			desiredVersion.String(), lowestCPVersion.String(),
+		)
```
JameelB: this seems to overlap with the work that's being done in https://github.com/Azure/ARO-HCP/pull/4566 cc @cgiradkar @machi1990

internal/validation/validators.go
```diff
@@ -862,11 +862,13 @@ func ValidateMajorUpgrade(fromVersion, toVersion semver.Version) error {
...
-// - No downgrades from highest active version
-// - Cannot exceed lowest control plane version
-// - No major version changes without AFEC (uses existing ValidateMajorUpgrade)
-// - No minor version skipping
+// ValidateNodePoolUpgrade validates a node pool version change (upgrade or downgrade).
+// HCP nodepools use Replace strategy (nodes are destroyed and recreated), so downgrades
+// are operationally identical to upgrades. Constraints:
+//   - Cannot exceed lowest control plane version
+//   - Must be within 2 minor versions of control plane (N-2 skew, same major)
+//   - Cross-major changes require AFEC FeatureExperimentalReleaseFeatures
```
JameelB: There's specific upgrade paths for mapping cross major upgrades (see [AllowControlPlaneNodePoolMajorVersionSkew](https://github.com/Azure/ARO-HCP/blob/71ec3a14e8e85051ec2e143996b6029b9da85168/internal/api/utils.go#L55-L59)), do these apply for downgrade too?

internal/validation/validators.go
adalrsjr1: This integration test step validates a downgrade that exceeds the N-2 skew limit, e.g., NP at 4.20 trying to downgrade to 4.17 when CP is at 4.20 (that's 3 minors below CP, beyond the allowed 2). The expected error "must be within 2 minor versions of control plane" is the lower bound check, not a downgrade block.

internal/validation/validators.go
adalrsjr1: resolved by rebase PR #5239 relaxed to +2.

internal/validation/validators.go
adalrsjr1: Added ValidateCrossMajorNodePoolSkew + unit tests.

adalrsjr1:
test failing as CS wasn't bumped yet -- waiting #5271

internal/validation/validators.go
```diff
@@ -905,35 +924,48 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
+			"invalid node pool version %s: cannot exceed control plane version %s",
+			desiredVersion.String(), lowestCPVersion.String(),
 		)
 	}
 
-	// Check if the desiredVersion <= control plane versions
-	// TODO: We may relax this constraint in the future
-	if lowestCPVersion != nil && desiredVersion.GT(*lowestCPVersion) {
+	exceedsN2Skew := lowestCPVersion != nil && desiredVersion.Major == lowestCPVersion.Major &&
+		desiredVersion.Minor < lowestCPVersion.Minor &&
```
machi1990: Is this needed?
if we arrive here; it is becase, same major and minor <= cp minor.

internal/validation/validators.go
```diff
@@ -905,35 +924,48 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
+		}
+		return nil
 	}
 
-	// Minor skip validation
-	if lowest != nil && desiredVersion.Minor > lowest.Minor+1 {
+	// Same-major upgrade: minor skip limit (+2). Downgrades are bounded by the
+	// N-2 lower bound above; HCP Replace strategy (destroy + recreate) means no
+	// step-through requirement for either direction.
+	exceedsMinorSkip := lowest != nil && desiredVersion.Minor > lowest.Minor+2
```
machi1990: Is there a reason we don't allow this?

internal/validation/validators.go
```diff
@@ -905,35 +924,48 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
+	isCrossMajorDowngrade := highest != nil && desiredVersion.Major < highest.Major
+	if isCrossMajorUpgrade || isCrossMajorDowngrade {
 		if !allowMajorUpgrade {
 			return fmt.Errorf("major version changes are not supported")
 		}
-		return ValidateMajorUpgrade(*lowest, desiredVersion)
+		if isCrossMajorUpgrade {
+			return ValidateMajorUpgrade(*lowest, desiredVersion)
+		}
+		if lowestCPVersion != nil {
```
machi1990: This can never happen; cp version will always be there

internal/validation/validators.go
```diff
@@ -905,35 +924,48 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
-			desiredVersion.String(), highest.String(),
+			"invalid node pool version %s: cannot exceed control plane version %s",
+			desiredVersion.String(), lowestCPVersion.String(),
 		)
 	}
 
-	// Check if the desiredVersion <= control plane versions
-	// TODO: We may relax this constraint in the future
-	if lowestCPVersion != nil && desiredVersion.GT(*lowestCPVersion) {
+	exceedsN2Skew := lowestCPVersion != nil && desiredVersion.Major == lowestCPVersion.Major &&
```
machi1990: cp version will never be nil

internal/validation/validators.go
```diff
@@ -905,35 +924,48 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
 
 	lowest, highest := apihelpers.FindLowestAndHighestNodePoolVersion(activeVersions)
 
-	// No partial downgrades: desiredVersion >= highest active version
-	if highest != nil && desiredVersion.LT(*highest) {
+	// Version bounds relative to control plane
+	exceedsCPVersion := lowestCPVersion != nil && desiredVersion.GT(*lowestCPVersion)
```
machi1990: cp version will never be nil

internal/validation/validators.go
```diff
@@ -905,35 +924,48 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
+	isCrossMajorUpgrade := lowest != nil && desiredVersion.Major > lowest.Major
+	isCrossMajorDowngrade := highest != nil && desiredVersion.Major < highest.Major
+	if isCrossMajorUpgrade || isCrossMajorDowngrade {
 		if !allowMajorUpgrade {
 			return fmt.Errorf("major version changes are not supported")
 		}
-		return ValidateMajorUpgrade(*lowest, desiredVersion)
+		if isCrossMajorUpgrade {
+			return ValidateMajorUpgrade(*lowest, desiredVersion)
+		}
```
machi1990, line 965: Why do we care about this for a replace strategy?

adalrsjr1:
/retest

failing due to container identity exhaustion

internal/validation/validators.go
adalrsjr1: Removed desiredVersion.Minor < lowestCPVersion.Minor && — redundant since line 928 guarantees desired.Minor <= CP.Minor when same major

internal/validation/validators.go
adalrsjr1: will be addressed on #5239

internal/validation/validators.go
adalrsjr1: Removed if lowestCPVersion != nil wrapper, now always calls ValidateCrossMajorNodePoolSkew

internal/validation/validators.go
adalrsjr1: Removed lowestCPVersion != nil && from exceedsCPVersion

internal/validation/validators.go
adalrsjr1: Removed lowestCPVersion != nil && from exceedsN2Skew

internal/validation/validators.go
adalrsjr1: will be addressed on #5239

internal/validation/validators.go
```diff
@@ -905,29 +924,35 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
 
-	// No partial downgrades: desiredVersion >= highest active version
-	if highest != nil && desiredVersion.LT(*highest) {
+	// Version bounds relative to control plane
+	exceedsCPVersion := desiredVersion.GT(*lowestCPVersion)
+	if exceedsCPVersion {
 		return fmt.Errorf(
-			"invalid node pool version %s: cannot downgrade from current version %s",
-			desiredVersion.String(), highest.String(),
+			"invalid node pool version %s: cannot exceed control plane version %s",
```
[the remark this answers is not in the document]
adalrsjr1: satisfied as in https://github.com/Azure/ARO-HCP/pull/5184#discussion_r3272400591

internal/validation/validators.go
```diff
@@ -905,29 +925,45 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
 
 	lowest, highest := apihelpers.FindLowestAndHighestNodePoolVersion(activeVersions)
 
-	// No partial downgrades: desiredVersion >= highest active version
-	if highest != nil && desiredVersion.LT(*highest) {
+	// Version bounds relative to control plane
+	exceedsCPVersion := desiredVersion.GT(*lowestCPVersion)
+	if exceedsCPVersion {
```
ahitacat: This can be inline

internal/validation/validators.go
```diff
@@ -905,29 +925,45 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
+			"invalid node pool version %s: cannot exceed control plane version %s",
+			desiredVersion.String(), lowestCPVersion.String(),
 		)
 	}
 
-	// Check if the desiredVersion <= control plane versions
-	// TODO: We may relax this constraint in the future
-	if lowestCPVersion != nil && desiredVersion.GT(*lowestCPVersion) {
+	exceedsN2Skew := desiredVersion.Major == lowestCPVersion.Major &&
+		lowestCPVersion.Minor-desiredVersion.Minor > 2
```
ahitacat: This can be inline, also the most restrictive way is to check against the highest version instead of the lowest version.
cpActiveVersions: 4.20, 4.21
The  min nodepool should be 4.19 and not 4.18

internal/validation/validators.go
```diff
@@ -905,29 +925,45 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
 
-	// No major version change unless FeatureExperimentalReleaseFeatures is registered
-	if lowest != nil && desiredVersion.Major > lowest.Major {
+	// When CP and NP are already on different majors (e.g., CP 5.0 with NP 4.x),
+	// validate that the desired NP version can coexist with the CP version.
+	// This applies to both upgrades (4.21→4.22) and downgrades (4.22→4.21) within
+	// the same major. NP changes that cross majors (e.g., 5.0→4.22) skip this and
+	// go through the AFEC gate below.
+	isSameMajorNPChange := highest == nil || desiredVersion.Major == highest.Major
+	if desiredVersion.Major != lowestCPVersion.Major && isSameMajorNPChange {
```
ahitacat: This check against the CP should be also consider the highestCPversion.

internal/validation/validators.go
```diff
@@ -905,29 +925,45 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
 		)
 	}
 
-	// No major version change unless FeatureExperimentalReleaseFeatures is registered
-	if lowest != nil && desiredVersion.Major > lowest.Major {
+	// When CP and NP are already on different majors (e.g., CP 5.0 with NP 4.x),
+	// validate that the desired NP version can coexist with the CP version.
+	// This applies to both upgrades (4.21→4.22) and downgrades (4.22→4.21) within
+	// the same major. NP changes that cross majors (e.g., 5.0→4.22) skip this and
+	// go through the AFEC gate below.
```
ahitacat, line 947: Shouldn't this change also be under AFEC. CP 5.0 work should be under AFEC flag currently.

internal/validation/validators.go
```diff
@@ -905,29 +925,45 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
...
+	// Cross-major change (upgrade or downgrade): gated behind FeatureExperimentalReleaseFeatures.
+	isCrossMajorUpgrade := lowest != nil && desiredVersion.Major > lowest.Major
+	isCrossMajorDowngrade := highest != nil && desiredVersion.Major < highest.Major
+	if isCrossMajorUpgrade || isCrossMajorDowngrade {
 		if !allowMajorUpgrade {
 			return fmt.Errorf("major version changes are not supported")
 		}
-		return ValidateMajorUpgrade(*lowest, desiredVersion)
+		if isCrossMajorUpgrade {
+			return ValidateMajorUpgrade(*lowest, desiredVersion)
```
ahitacat, line 964: Is this required? This is kind of similar as checking the upgrade path from Cincinnati.

test-integration/frontend/artifacts/FrontendCRUD/NodePool/version-downgrade/07-httpReplace-nodepool-downgrade/nodepool.json
```diff
@@ -11,7 +11,7 @@
 		},
 		"version": {
 			"channelGroup": "stable",
-			"id": "4.20.0"
+			"id": "4.17.0"
```
ahitacat: min version we are supporting is 4.20.8 isn't this test failing? because of this as well?

ahitacat (commented):
Can you squash the commits, it could be easy to other people pick your changes into its current work 😄

internal/validation/validators.go
adalrsjr1: Added the !allowMajorUpgrade gate and a test case that verifies 4.22 → 4.21 with CP at 5.0 is rejected without AFEC.

internal/validation/validators.go
adalrsjr1: both ValidateCrossMajorNodePoolSkew calls now use highestCPVersion.

internal/validation/validators.go
adalrsjr1: CS bypasses all version validation for ARO HCP `(IsAroHcp() → return nil)`, and HyperShift's own validation is disabled (`SkipReleaseImageValidation: true`). Cincinnati only validates that the version exists in the channel graph, not that a specific NP version can coexist with a specific CP version. 

`ValidateCrossMajorNodePoolSkew` is currently the only enforcement for cross-major CP-NP compatibility in the stack.

test-integration/frontend/artifacts/FrontendCRUD/NodePool/version-downgrade/07-httpReplace-nodepool-downgrade/nodepool.json
adalrsjr1: The test wasn't failing — 4.20.8 is exactly at the minimum floor. But both N-2 skew and minimum version would reject the old downgrade target (4.17.0), making it unclear which constraint was being tested. Bumped to 4.23.0 → 4.20.0 so only N-2 skew applies.

adalrsjr1:
> Can you squash the commits, it could be easy to other people pick your changes into its current work 😄

@ahitacat Squashed! :) 
and comments addressed.

PR #5185, opened 2026-05-07
test,tooling: add a snapshot gatherer from Kusto

stevekuznetsov:
We would like to enable Agent-driven end-to-end test failure analysis and debugging. The first and most important part of this process is taking a failed job and, for each test, extracting some concise log outputs for the Agent to review. Conciseness here is absolutely key - our current log volumes are absolutely too high to feed into context windows and hope for a good outcome. This PR adds a new snapshot gathering tool that tries to reconstruct the flow of data through our system from first principles: start with the test client, determine what resources end up in which state stores, trace through all of our component handoffs, etc.

This PR also adds a new sub-command for tests, modeled after the observability gathering bits added recently for metrics. We've seen subtle (or clearly broken) regressions in log formats over time, and without any guarantee that our carefully-written Kusto queries remain valid, any ability to do post-hoc analysis is shot. To that end, the new test sub-command tries to capture the ability of our tooling to gather a snapshot, providing a custom link tools HTML page overview as well as jUnit for the queries we ran. We start in a best-effort mode but will graduate to blocking as we flush out the inconsistencies.

Agentic flows will *not* use the content created inside of a Prow job, as in production environments we won't have connectivity between the job runner and the Kusto databases we need to query. Instead, an asynchronous process will vendor in the gather-snapshot code and run it on demand.


tooling/hcpctl/pkg/snapshot/gatherer.go
```diff
@@ -0,0 +1,795 @@
...
+			return fmt.Errorf("failed to write markdown: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// executeKQL runs a KQL query string against the given database and returns
+// the result rows with column names and string values.
+func (g *Gatherer) executeKQL(ctx context.Context, kqlStr, database string, timeout time.Duration) ([]resultRow, error) {
```
janboll, line 630: Why not work with the existing KustoClient interface? Please change it so it works for you. It's a bit odd to create multiple clients in the same code base

stevekuznetsov:
The only other PR with approved & LGTM not needing rebase or with hold has failing tests, and Tide seems to have not picked this up for days. Going to try re-testing, but may just green-button it otherwise.

/retest

PR #5188, opened 2026-05-08
E2E tests: wait for node pool to be fully ready

miquelsi:
https://redhat.atlassian.net/browse/ARO-26861

### What

Wait for the node pool to be fully available before deploying a simple web app for verifying the cluster.

### Why

This change was needed because a node pool can report as created in ARM before the cluster’s ingress path is actually usable, which caused the test to move on too early and become flaky. Adding framework.WaitForNodePoolReady closes that gap by waiting for the full set of readiness signals that matter for workload traffic—router availability, canary route admission, and healthy route.openshift.io discovery—so subsequent validation steps run only after the new node pool is genuinely operational from an OpenShift perspective.

### DIFF
--- a/test/e2e/complete_cluster_create_multiversion.go
+++ b/test/e2e/complete_cluster_create_multiversion.go
@@ -165,6 +165,17 @@ var _ = Describe("ARO-HCP", func() {
 			)
 			Expect(err).NotTo(HaveOccurred())
 
+			By("waiting for node pool to be ready")
+			err = framework.WaitForNodePoolReady(ctx,
+				tc.Get20240610ClientFactoryOrDie(ctx).NewNodePoolsClient(),
+				adminRESTConfig,
+				*resourceGroup.Name,
+				clusterName,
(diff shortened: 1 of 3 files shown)


test/e2e/complete_cluster_create_multiversion.go
```diff
@@ -165,6 +165,17 @@ var _ = Describe("ARO-HCP", func() {
...
+			By("waiting for node pool to be ready")
+			err = framework.WaitForNodePoolReady(ctx,
+				tc.Get20240610ClientFactoryOrDie(ctx).NewNodePoolsClient(),
+				adminRESTConfig,
+				*resourceGroup.Name,
+				clusterName,
+				nodePoolName,
+				10*time.Minute,
+			)
+			Expect(err).NotTo(HaveOccurred())
```
mbukatov, line 177: Why isn't the ARO HCP service reporting that a node pool is ready when all these conditions you are checking here are met?

PR #5200, opened 2026-05-11
chore: bump cluster install versions

JakobGray:
### What

Update cluster install versions

### Why

Use latest available z version


machi1990 (approved):
/lgtm

PR #5201, opened 2026-05-11
test: verify config/ write permissions

raelga:
Empty file to test write permissions on config/. Safe to close after verification.


raelga (commented):
/close

PR #5203, opened 2026-05-11
Remove dynamic pod names from alert summaries to reduce incident noise

janboll:
### What/Why

Alert summaries are used as incident titles in the triaging system. Dynamic parts like pod names cause separate incidents for the same underlying problem. This removes pod names and other ephemeral identifiers from summaries while keeping them in descriptions.
Remove alert `LowOutputBytesProcessed`, cause we now have exporter providing info if kusto logs are flowing.


raelga (approved):
/lgtm
/approve

PR #5206, opened 2026-05-11
add genericdocument compatible serialization of nodepools

deads2k:
This stores all the data necessary to switch from custom cluster storage to the generic document storage for nodepools.

Part of a multi-step plan
1. https://github.com/Azure/ARO-HCP/pull/5206
2. https://github.com/Azure/ARO-HCP/pull/5207
3. https://github.com/Azure/ARO-HCP/pull/5208
4. https://github.com/Azure/ARO-HCP/pull/5209

PR #5207, opened 2026-05-11
read new fields for nodepool storage

deads2k:
part of https://github.com/Azure/ARO-HCP/pull/5206, must merge after https://github.com/Azure/ARO-HCP/pull/5206 makes it prod


internal/database/convert_nodepool.go
```diff
@@ -56,43 +59,26 @@ func InternalToCosmosNodePool(internalObj *api.HCPOpenShiftClusterNodePool) (*No
 		},
 	}
 
+	cosmosObj.InternalState.InternalAPI.CosmosMetadata = api.CosmosMetadata{}
```
miguelsorianod, line 64: Why are we now unsetting the cosmos obj CosmosMetadata top-level attribute as well as the one in InternalState.InternalAPI?

miguelsorianod: Is it because at some point we will stop having both the CosmosMetadata top-level attribute as well as  InternalState.InternalAPI?

internal/database/convert_nodepool.go
```diff
@@ -56,43 +59,26 @@ func InternalToCosmosNodePool(internalObj *api.HCPOpenShiftClusterNodePool) (*No
...
-	internalObj.TrackedResource = arm.TrackedResource{
-		Resource: arm.Resource{
-			ID:         resourceDoc.ResourceID,
-			Name:       resourceDoc.ResourceID.Name,
-			Type:       resourceDoc.ResourceID.ResourceType.String(),
-			SystemData: resourceDoc.SystemData,
-		},
-		Location: cosmosObj.InternalState.InternalAPI.Location,
-		Tags:     resourceDoc.Tags,
+	internalObj := cosmosObj.DeepCopy()
```
miguelsorianod, line 74: Is it correct/desired that calling DeepCopy of `cosmosObj` returns the internal api type? I understand this occurs because we now inline the internal api type on the cosmos type and because of that the DeepCopy method is promoted to the cosmos type, but it surprised me as usually when using the DeepCopy method on a type you expect the type itself being deepcopied and not just an inlined attribute of it.

internal/database/convert_nodepool.go
```diff
@@ -56,43 +59,26 @@ func InternalToCosmosNodePool(internalObj *api.HCPOpenShiftClusterNodePool) (*No
...
+	internalObj.ExistingCosmosUID = cosmosObj.ID
```
miguelsorianod, line 75: Do we do this because the one the inlined object that we retrieve from cosmos doesn't have that information? the same question applies to ETag and ResourceID

deads2k:
query confirms all data is ready for this

PR #5209, opened 2026-05-11
switch to using generic storage for nodepools

deads2k:
part of https://github.com/Azure/ARO-HCP/pull/5206, must merge after https://github.com/Azure/ARO-HCP/pull/5207 makes it to prod


deads2k:

PR #5215, opened 2026-05-11
add genericdocument compatible serialization of externalauths

deads2k:
This stores all the data necessary to switch from custom cluster storage to the generic document storage for externalauth.

Part of a multi-step plan


stevekuznetsov (approved):
/lgtm

PR #5219, opened 2026-05-11
[wip] add controllers that will delete clusterservice nodepools and clear c…

deads2k:
…luster-service IDs when they are gone

Demonstrate how to do this transition in a way that will mostly work.  Perfect, no. Orphan cleanup in `ClusterServiceMatchingClusters` will get stragglers.


frontend/pkg/frontend/node_pool.go
```diff
@@ -749,6 +750,9 @@ func (f *Frontend) addDeleteNodePoolToTransaction(ctx context.Context, writer ht
 
 	nodePool.ServiceProviderProperties.ActiveOperationID = operationDoc.ResourceID.Name
 	nodePool.Properties.ProvisioningState = operationDoc.Status
+	if nodePool.ServiceProviderProperties.DeletionTimestamp == nil {
+		nodePool.ServiceProviderProperties.DeletionTimestamp = &metav1.Time{Time: time.Now()}
```
miguelsorianod: Should this be UTC?

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go
```diff
@@ -0,0 +1,168 @@
...
+	if len(csID.String()) == 0 {
+		if elapsed := c.clock.Since(deletedAt); elapsed < missingClusterServiceIDTimeout {
+			// The frontend may still be in the middle of creating the cluster-service
+			// NodePool, or the controller that does so hasn't run yet. Re-check on the
+			// next sync; the resync interval and informer change events drive retries.
+			return nil
+		}
+		logger.Info("giving up on cluster-service NodePool delete — ClusterServiceID never appeared",
+			"deletionTimestamp", deletedAt)
+	} else if err := c.clusterServiceClient.DeleteNodePool(ctx, csID); err != nil {
```
miguelsorianod: At the moment of writing this, Clusters Service has a validation that prevents deleting the last existing nodepool of a Cluster. This has the following consequences on this controller:

* This means that an error will be returned in this controller if the nodepool being deleted is the last one
* With the changes of this MR,  a consequence of performing a delete cluster request is that deletion of all nodepools is triggered which will at some point error on the last existing one. For this specific case additionally the controllers will error on nodepool deletion because in CS there's a validation that prevents performing a nodepool delete when its parent cluster is being deleted.

frontend/pkg/frontend/node_pool.go
miguelsorianod, line 763: An aspect that is will be missing here are changes in the delete nodepool operation:

As soon as we move the cluster/nodepool creation to backend we won't have any guarantee that  `nodePool.ServiceProviderProperties.ClusterServiceID` will be there anymore at the point of nodepool deletion. I think that means we will need to change this and the operation nodepool delete to change their preconditions and what they do. The following approach might work:
1. Set `nil` InternalID when creating the delete nodepool operation in Cosmos
2. On the node pool delete operation controller:
  2.0. We now need to do a Cosmos Get of the NodePool because `InternalID` on the operation is never set anymore. This is an extra Cosmos call aside from the cosmos call to get the Operation
  2.1. Add new checks that prevent the controller from continuing:
    * Do not proceed until ClusterServiceID is nil in the NodePool internal api type
    * Do not proceed until the DeletionTimestamp is not nil in the NodePool internal api type
    * Do not proceed until the ClusterServiceDeletionTimestamp is not nil in the NodePool internal api type. This is needed because only checking DeletionTimestamp is not enough, as we only want to proceed when we know that the deletion controller has at least attempted to peform the nodepool delete. Maybe we don't need to check DeletionTimestamp as ClusterServiceDeletionTimestamp implies it? or maybe for additional robustness we want to check both
  2.2 Remove the `c.clusterServiceClient.GetNodePoolStatus` call, as that now is managed by NodePoolClusterServiceIDClearer. That controller is the one that unsets ClusterServiceID.

However, that means that the operation delete nodepool controller could potentially end up marking the operation as completed and remove the nodepool from cosmos when it's not actually removed from CS, in the case where the controllers weren't able to delete it from CS because they gave up waiting for the cs nodepoolid.

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go
```diff
@@ -0,0 +1,168 @@
...
+		if !errors.As(err, &ocmError) || ocmError.Status() != http.StatusNotFound {
+			return utils.TrackError(fmt.Errorf("failed to delete cluster-service NodePool: %w", err))
+		}
+		// 404 — could be a stale CSID or a race against an in-flight CS create;
+		// require the same dwell time as the missing-CSID branch before treating
+		// the NodePool as definitively gone.
+		if elapsed := c.clock.Since(deletedAt); elapsed < missingClusterServiceIDTimeout {
+			return nil
+		}
+		logger.Info("cluster-service NodePool already deleted", "clusterServiceID", csID.String())
```
miguelsorianod: Probably good idea to make this log more accurate: we might get a 404 in two cases:
1. The nodepool has really been deleted from CS side
2. The nodepool delete has returned a 404 because the controllers are racing and the create one for some reason has not created the nodepol yet, or it has failed to persist the fact in Cosmos

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// missingClusterServiceIDTimeout is how long we wait after DeletionTimestamp
+// for the ClusterServiceID to appear before concluding that the corresponding
+// Cluster Service NodePool was never created and we have no work to do.
+const missingClusterServiceIDTimeout = 120 * time.Second
```
miguelsorianod: This means we could end up in a situation where the cluster ends up orphan on CS side which means we would delete the NodePool Cosmos resource before the cluster is actually fully removed from CS side. Could that cause an issue when doing quick recreation with the same payload it would fail because the old resources on CS side are still there, or even worse, assume that the old ones are associated to the new one?

miguelsorianod: At a technical level, is there no way to prevent this kind of racing?

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go
deads2k: Change this in cluster-service.

deads2k: it would be completely fine and probably good, to send the customer a notification when they have a kube-apiserver and no nodepools for a duration of over an hour.

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go
deads2k: Accepted. We have an orphan controller cleans this up.

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go
deads2k: > * With the changes of this MR,  a consequence of performing a delete cluster request is that deletion of all nodepools is triggered which will at some point error on the last existing one. For this specific case additionally the controllers will error on nodepool deletion because in CS there's a validation that prevents performing a nodepool delete when its parent cluster is being deleted.

Update with however cluster-service communicates this back to the caller.

frontend/pkg/frontend/node_pool.go
deads2k: What if we added a contorller watching nodepools that waits until
1. deletionTimestamp is nonnil
2. deleteclsuterservicetimestamp is nonnil
3. clusterserviceId is nil
4. hostedclustercontent has all been deleted
5. maestrobundles have been deleted
6. whatever other conditions we want to wait for
7. then this controller deletes the nodepool in cosmos
8. then the nodepool-operation-delete waits until the nodepool doesn't exist and marks it completed.

miguelsorianod:
The PR has been updated to move deletion of nodepools to be asynchronous, including ordered cleanup of dependents. Aside from that, the cluster delete operation controller now waits until the Cosmos nodepool resources are deleted to continue its deletion process. This is because the nodepool asynchronous deletion process depends on parent cluster cosmos resources information (ServiceProviderCluster for example).

After the changes, we have the following controllers:
* NodePoolDeletionClusterServiceDeleter: takes care of triggering the CS nodepool deletion
* NodePoolDeletionClusterServiceIDClearer: takes care of checking whether nodepool deletion that has been triggered by NodePoolDeletionClusterServiceDeleter has been fully completed from CS side
* NodePoolScopedMaestroReadonlyBundlesDelete: takes care of deleting nodepool-scoped maestro readonly bundles
* NodePoolChildResourceCleanupController: takes care of deleting nodepool direct children cosmos resources (serviceprovidernodepools, ...)
* NodePoolDeletionController: takes care of deleting the nodepool cosmos resource itself
* NodePoolDeletionOperationStatus: takes care of reconciling the nodepool delete operation status and the nodepool provisioningStatus depending on what CS reports. This excludes CS reporting 404, which is reconciled in NodePoolDeletionClusterServiceIDClearer. This controller can end up marking the operation in a terminal state which unsets ActiveOperationID

Finally, the nodepool operation delete controller has been updated to now wait until the nodepool cosmos resource associated to the delete operation has been deleted. The nodepool operation controller also doesn't act if there's no activeoperation

backend/pkg/app/backend.go
```diff
@@ -591,6 +591,46 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
...
+		b.options.ResourcesDBClient,
+		activeOperationLister,
+		backendInformers,
+	)
+	nodePoolDeletionController := nodepooldeletion.NewNodePoolDeletionController(
+		b.options.ResourcesDBClient,
+		activeOperationLister,
+		backendInformers,
+	)
+	nodePoolScopedMaestroReadonlyBundlesDeleteController := nodepooldeletion.NewNodePoolScopedMaestroReadonlyBundlesDeleteController(
```
deads2k: one potential benefit of having this in the same controller as the one that creates them: they would never race.  thoughts?

backend/pkg/controllers/nodepooldeletion/node_pool_child_resource_cleanup_controller.go
```diff
@@ -0,0 +1,147 @@
...
+		if strings.EqualFold(childResource.ResourceType, api.NodePoolControllerResourceType.String()) {
+			continue
+		}
+
+		if childResource.ResourceID == nil {
+			return utils.TrackError(fmt.Errorf("child resource at cosmosID %q has no resourceID; refusing to delete", childResource.ID))
+		}
+
+		// We only delete the ServiceProviderNodePool if all the Maestro readonly bundles have been unset
+		if childResource.ResourceType == api.ServiceProviderNodePoolResourceType.String() {
```
deads2k: If this going to be per type, factor it as a map to function that can return a "shouldDelete" bool.

backend/pkg/controllers/nodepooldeletion/node_pool_child_resource_cleanup_controller.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to create untyped CRUD for node pool children: %w", err))
+	}
+
+	childIterator, err := untypedCRUD.List(ctx, nil)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to list node pool child resources: %w", err))
+	}
+	for _, childResource := range childIterator.Items(ctx) {
+		if strings.EqualFold(childResource.ResourceType, api.NodePoolControllerResourceType.String()) {
```
deads2k: no. Refactor as the map below and and if the map doesn't understand the type, do nothing.

backend/pkg/controllers/nodepooldeletion/node_pool_child_resource_cleanup_controller.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil
+	}
+
+	nodePoolResourceID := key.GetResourceID()
+	untypedCRUD, err := c.resourcesDBClient.UntypedCRUD(*nodePoolResourceID)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to create untyped CRUD for node pool children: %w", err))
+	}
+
+	childIterator, err := untypedCRUD.List(ctx, nil)
```
deads2k: make it listrecursive.

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_cluster_service_deleter_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get node pool: %w", err))
+	}
+	if !c.NeedsWork(nodePool) {
+		return nil
+	}
+
+	csID := nodePool.ServiceProviderProperties.ClusterServiceID
+	deletedAt := nodePool.ServiceProviderProperties.DeletionTimestamp.Time
```
deads2k: I could see making this `observedDeleteAt := c.clock.Now()` and storing it in a map for later retrieval to see if this controller has been around for at least two minutes.  that avoids the "backend was off and restarted after a day" problem.  Would need some commenting to explain it to folks.  Rest of the algorithm seems ok.

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go
```diff
@@ -0,0 +1,163 @@
...
+	spnpCRUD := c.resourcesDBClient.ServiceProviderNodePools(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, key.HCPNodePoolName)
+	spnp, spnpErr := spnpCRUD.Get(ctx, api.ServiceProviderNodePoolResourceName)
+	if spnpErr != nil && !database.IsNotFoundError(spnpErr) {
+		return utils.TrackError(fmt.Errorf("failed to get ServiceProviderNodePool: %w", spnpErr))
+	}
+	if spnp != nil && len(spnp.Status.MaestroReadonlyBundles) > 0 {
+		logger.Info("waiting for nodepool-scoped Maestro readonly bundles to be deleted before removing Cosmos entry",
+			"remainingBundles", len(spnp.Status.MaestroReadonlyBundles))
+		return nil
+	}
```
deads2k, line 162: factor into a separate function with a name that makes it obvious that other deletePreconditions should create similar functions.

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_operation_status_controller.go
```diff
@@ -0,0 +1,167 @@
...
+	return nodePool.ServiceProviderProperties.DeletionTimestamp != nil &&
+		nodePool.ServiceProviderProperties.ClusterServiceDeletionTimestamp != nil &&
+		nodePool.ServiceProviderProperties.ClusterServiceID != nil && len(nodePool.ServiceProviderProperties.ClusterServiceID.String()) > 0 &&
+		nodePool.ServiceProviderProperties.ActiveOperationID != ""
+}
+
+func (c *nodePoolDeletionOperationStatusController) SyncOnce(ctx context.Context, key controllerutils.HCPNodePoolKey) error {
+	logger := utils.LoggerFromContext(ctx)
+
+	cachedNodePool, err := c.nodePoolLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, key.HCPNodePoolName)
```
deads2k: if we delete in backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go, then crash, then start back up, when does this get called?

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_operation_status_controller.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// nodePoolDeletionOperationStatusController polls Cluster Service for the
+// NodePool status during deletion and updates the active operation's
+// provisioning state accordingly. It activates after the CS delete has been
+// issued (ClusterServiceDeletionTimestamp set) and stops once the
+// ClusterServiceID is cleared (404 handled by nodePoolClusterServiceIDClearer).
+type nodePoolDeletionOperationStatusController struct {
```
deads2k: I don't get this one.  Seems like the existing oepration_nodepool_delete ought to do this.

miguelsorianod: operation_nodepool_delete only waits until the cosmos resource is deleted.

This one reconciles the state of the delete operation and the resource itself over time **during** the progress of the operation.

The operation controller also is triggered based on the operation and not the nodepool. This controller is triggered on nodepool.

backend/pkg/controllers/nodepooldeletion/node_pool_child_resource_cleanup_controller.go
miguelsorianod: A downside of that is that then we will need to maintain the list over time and people can forget about it which means that it won't be deleted.

Or do you mean the extra check only? as in, the map represents "extra precondition checks for deletion" and not whether deletion occurs or not. Deletion would occur if the entry is not in the "extra precondition checks" map

backend/pkg/controllers/nodepooldeletion/node_pool_child_resource_cleanup_controller.go
miguelsorianod: I can do that, but this changes the current behavior (before this MR) where only direct child are deleted.

backend/pkg/app/backend.go
miguelsorianod: Aside from the semantical incoherence (it's called create but we would add the responsibility of deletion), how would we organize that controller? (use of cache vs direct cosmos retrieval, different preconditions, ...)

something like, using NP cached lister:
1. If NP deletiontimestamp == nil ---> path of creation
2. If NP deletiontimestamp != nil
  2.1 Get of SPNP. If it does not exist we are done
  2.2 If it exists, if NP CSDeletionTimestamp != nil && ClusterServiceID == nil --> path of deletion. Otherwise we are done

If we have determined it's path of creation or path of deletion we repeat the same with direct cosmos retrieval instead of cached lister.

Then:
* On the path of creation we would pass the retrieved cosmos NP, and then within it we would do a GetOrCreateServiceProviderNodePool (as the create controller has now) and the logic that we already have
* On path of deletion we would pass the retrieved cosmos NP, and do a GetServiceProviderNodepool. If it doesn't exist we are done, if it exists then we do the logic of checking if there are MaestroReadonlyBundles in the ServiceProviderNodePool and ensure we delete them

Did you have this in mind?

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_cluster_service_deleter_controller.go
miguelsorianod: How would we organize such a map? depending on the key, what occurs if a resource is recreated with the same name? How would we avoid potentially ending up with a large map (eventual deletion is not guaranteed if there are issues)

miguelsorianod: If we just want to see "if the controller has been around for at least two minutes" wouldn't an easier way would be creating the controller with time.Now.UTC when instantiated (or in the beginning of the Run method??? although in that case it would need to be added in genericWatchingController as that's the method that offers the Run) and then do a time.since and if it's been less than two minutes then we do nothing?¿

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_operation_status_controller.go
miguelsorianod: If the NodePool cosmos entry is deleted in the backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go and then there's a crash, then the operation_node_pool_delete.go controller will kick in (as the operation is still active), it will see that the cosmos entry is not there, it will then run `SetDeleteOperationAsCompleted` which will mark the operation as completed.

backend/pkg/controllers/nodepooldeletion/node_pool_child_resource_cleanup_controller.go
miguelsorianod: It is unclear as of now how this will evolve. See https://github.com/Azure/ARO-HCP/pull/5219#discussion_r3272843311 on the map approach

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_cluster_service_deleter_controller.go
deads2k: LRU with 50000 entries

deads2k: delete the LRU entry when we actaully delete the thing.

backend/pkg/controllers/nodepooldeletion/node_pool_child_resource_cleanup_controller.go
deads2k: map[resorucetype]shouldDelete

if map does not have an entry, delete it
if we don't want to delete (like controllers), then entry in the map and always return false
if we sometimes delete (like ServiceProviderNodePool), then use it.

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_operation_status_controller.go
miguelsorianod: As discussed the other day, we want to move the resource and operation reconciling fully to the delete operation controller. I made the change which means nodePoolDeletionOperationStatusController is fully removed.

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_controller.go
miguelsorianod: done

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_cluster_service_deleter_controller.go
miguelsorianod: changed

backend/pkg/controllers/nodepooldeletion/node_pool_child_resource_cleanup_controller.go
miguelsorianod: changed

backend/pkg/controllers/nodepooldeletion/node_pool_child_resource_cleanup_controller.go
miguelsorianod: added

backend/pkg/app/backend.go
miguelsorianod: Changed as I described in the comment above

backend/pkg/controllers/operationcontrollers/operation_node_pool_delete.go
```diff
@@ -104,21 +113,53 @@ func (c *operationNodePoolDelete) SynchronizeOperation(ctx context.Context, key
...
 		return nil
 	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get node pool: %w", err))
+	}
+
+	if !c.shouldReconcileOperationAndResourceStatus(nodePool) {
+		return nil
+	}
+	err = c.reconcileOperationAndResourceStatus(ctx, operation, nodePool)
```
deads2k, line 137: Will accept as a starting point, but we need to extend this with multiple checks like we did for create and update, so that we can relate status based on what the cleanup controllers are doing.  How many children resource of each type remain for instance.

backend/pkg/controllers/create_nodepool_scoped_maestro_readonly_bundles_controller.go
```diff
@@ -99,19 +120,105 @@ func NewCreateNodePoolScopedMaestroReadonlyBundlesController(
...
+	// First we do a quick check to see if there's some work needed. If not, we return early. If the cache is outdated
+	// a next reconcile cycle should deal with it.
+	cachedNodePool, err := c.nodePoolLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, key.HCPNodePoolName)
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get node pool from cache: %w", err))
+	}
+	cachedWorkNeeded, err := c.cachedShouldReconcileCreateOrDelete(ctx, cachedNodePool)
```
deads2k, line 171: Could add 
```

	cachedServiceProviderNodePool, err := c.serviceProviderNodePoolLister.Get(ctx, cachedNodePool.ID.SubscriptionID, cachedNodePool.ID.ResourceGroupName, cachedNodePool.ID.Parent.Name, cachedNodePool.ID.Name)
	if database.IsNotFoundError(err) {
		return false, nil
	}
	if err != nil {
		return false, utils.TrackError(fmt.Errorf("failed to get ServiceProviderNodePool from cache: %w", err))
	}

```

here and then instead of `cachedShouldReconcileCreateOrDelete` it's just `NeedsWork` as normal, right?

deads2k:
Some comments.   I'll accept all as followups

/lgtm

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_cluster_service_deleter_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	cacheKey := strings.ToLower(nodePool.ID.String())
+	firstSeenEntry, ok := c.firstSeenDeletionTimestampCache.Get(cacheKey)
+	if !ok {
+		c.firstSeenDeletionTimestampCache.Add(cacheKey, c.clock.Now().UTC())
+		return nil
+	}
+	firstSeenNodePoolDeletionTimestamp := firstSeenEntry.(time.Time)
+	if c.clock.Since(firstSeenNodePoolDeletionTimestamp) < firstSeenDeletionGracePeriod {
+		return nil
+	}
```
deads2k: remove

backend/pkg/controllers/nodepooldeletion/node_pool_deletion_cluster_service_deleter_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	// triggering deletion in scenarios where the nodepool was marked for deletion
+	// but the controllers were not available for some reason until some time afterwards
+	// for some reason. For example, both create and delete controllers not being available
+	// and restarting at the same time.
+	nodePoolDeletionTimestamp := nodePool.ServiceProviderProperties.DeletionTimestamp.Time
+	cacheKey := strings.ToLower(nodePool.ID.String())
+	firstSeenEntry, ok := c.firstSeenDeletionTimestampCache.Get(cacheKey)
+	if !ok {
+		c.firstSeenDeletionTimestampCache.Add(cacheKey, c.clock.Now().UTC())
+		return nil
```
deads2k: remove this

backend/pkg/controllers/nodepooldeletion/node_pool_cluster_service_delete_dispatch_controller.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+			logger.Info("cluster-service NodePool already deleted or race against in-flight CS create", "clusterServiceID", csID.String())
+		default:
+			return utils.TrackError(fmt.Errorf("failed to delete cluster-service NodePool: %w", err))
+		}
+	} else {
+		logger.Info("requested cluster-service NodePool delete", "clusterServiceID", csID.String())
+	}
+
+	nodePool.ServiceProviderProperties.ClusterServiceDeletionTimestamp = &metav1.Time{Time: c.clock.Now().UTC()}
```
deads2k, line 204: need to clear the clusterserviceID field here.

backend/pkg/controllers/nodepooldeletion/node_pool_cluster_service_delete_dispatch_controller.go
```diff
@@ -0,0 +1,215 @@
...
+	nodePoolDeletionTimestamp := nodePool.ServiceProviderProperties.DeletionTimestamp.Time
+	cacheKey := strings.ToLower(nodePool.ID.String())
+	var firstSeenNodePoolDeletionTimestamp time.Time
+	firstSeenEntry, ok := c.firstSeenDeletionTimestampCache.Get(cacheKey)
+	if ok {
+		firstSeenNodePoolDeletionTimestamp = firstSeenEntry.(time.Time)
+	} else {
+		firstSeenNodePoolDeletionTimestamp = c.clock.Now().UTC()
+		c.firstSeenDeletionTimestampCache.Add(cacheKey, firstSeenNodePoolDeletionTimestamp)
+	}
```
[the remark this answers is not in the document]
miguelsorianod: Do we want to change the key to be ResourceID + NodePoolDeletionTimestamp attribute? that would potentially increase the number of cache entries (if resources are recreated with same resource id ) but would ensure more safety around reuse of the cache entry leading to an incorrect evaluation.

PR #5223, opened 2026-05-11
mgmt-agent: new component to run on management clusters

stevekuznetsov:
Bootstrap a new component, the management agent, which runs on management clusters to take care of ARO-HCP specific functionality we don't/can't offload onto HyperShift.

Our first order of business - figure out how many accelerated (SWIFT) NICs each node has available (non-primary ones, as the primary one is in use by the node itself), and expose this as a resource that pods can schedule against.


mgmt-agent/cmd/options.go
```diff
@@ -0,0 +1,231 @@
...
+		LeaderElectionRenewDeadline: 10 * time.Second,
+		LeaderElectionRetryPeriod:   2 * time.Second,
+		Workers:                     2,
+	}
+}
+
+func (o *RawControllerOptions) BindFlags(cmd *cobra.Command) error {
+	klog.InitFlags(nil)
+
+	cmd.Flags().StringVar(&o.HealthAddress, "health-address", o.HealthAddress, "The bind address for the health check server (e.g., ':8080')")
```
deads2k, line 63: need metrics too.  I wonder if I remembered that for kube-applier.  kube clients can get good metrics almost for free.

mgmt-agent/deploy/templates/clusterrole.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
deads2k, line 2: matches expectations.

mgmt-agent/pkg/controller/controller.go
```diff
@@ -0,0 +1,282 @@
...
+	"k8s.io/klog/v2"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v6"
+)
+
+const (
+	// SwiftNICResourceName is the extended resource name for SWIFT NICs.
+	SwiftNICResourceName corev1.ResourceName = "aro-hcp.azure.com/swift-nic"
```
deads2k: you prefer azure.com?  I prefer aro.openshift.io

mgmt-agent/pkg/controller/controller.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get VMSS VM for node %s: %w", name, err)
+	}
+
+	nicCount := countSwiftNICs(vm)
+	logger.Info("Setting SWIFT NIC capacity on node", "node", name, "count", nicCount)
+
+	return applycorev1.Node(name).
+		WithStatus(applycorev1.NodeStatus().
+			WithCapacity(corev1.ResourceList{
```
deads2k, line 219: check in kube to be sure this is a list type map.  If it isn't and it's replaced atomically, you need to co-exist with other values.  It gets complicated and a patch will work better for us.

deads2k: Sweet, it's a map.  We're golden.

mgmt-agent/pkg/controller/controller.go
```diff
@@ -0,0 +1,282 @@
...
+		if !isPrimary && isAccelerated {
+			count++
+		}
+	}
+	return count
+}
+
+// defaultGetVM returns a GetVMFunc that uses the Azure Compute SDK.
+func defaultGetVM(credential azcore.TokenCredential) GetVMFunc {
+	return func(ctx context.Context, subscriptionID, resourceGroup, vmssName, instanceID string) (armcompute.VirtualMachineScaleSetVMsClientGetResponse, error) {
```
deads2k, line 252: personal preference.  I really dislike passing functions around.  I prefer interfaces and structs that fulfill them.

mgmt-agent/cmd/options.go
```diff
@@ -0,0 +1,231 @@
...
+	LeaderElectionRenewDeadline time.Duration
+	LeaderElectionRetryPeriod   time.Duration
+}
+
+func DefaultControllerOptions() *RawControllerOptions {
+	return &RawControllerOptions{
+		HealthAddress:               ":8080",
+		LeaderElectionLeaseDuration: 15 * time.Second,
+		LeaderElectionRenewDeadline: 10 * time.Second,
+		LeaderElectionRetryPeriod:   2 * time.Second,
```
deads2k: pretty short, check the library-go defaults.

mgmt-agent/cmd/options.go
```diff
@@ -0,0 +1,231 @@
...
+)
+
+type RawControllerOptions struct {
+	HealthAddress               string
+	Kubeconfig                  string
+	Namespace                   string
+	Workers                     int
+	LeaderElectionLeaseDuration time.Duration
+	LeaderElectionRenewDeadline time.Duration
+	LeaderElectionRetryPeriod   time.Duration
```
deads2k: these are basically never configured right via the CLI. Why allow it?

mgmt-agent/cmd/options.go
```diff
@@ -0,0 +1,231 @@
...
+
+const (
+	LeaderElectionLockName = "mgmt-agent-controller-leader"
+)
+
+type RawControllerOptions struct {
+	HealthAddress               string
+	Kubeconfig                  string
+	Namespace                   string
+	Workers                     int
```
deads2k: dont' allow configuration.

deads2k (commented):
minor comments.  Looks like it stands a good shot at working.

mgmt-agent/cmd/cmd.go
```diff
@@ -0,0 +1,67 @@
...
+		cmd.AddCommand(subCmd)
+	}
+	return cmd, nil
+}
+
+func NewControllerCommand() (*cobra.Command, error) {
+	opts := DefaultControllerOptions()
+	cmd := &cobra.Command{
+		Use:   "controller",
+		Short: "Start the mgmt-agent controller",
```
deads2k, line 57: add a long explanation for what this is for.  In my view, this is to run on the management cluster to configure cluster-scoped resources to suit the needs of running ARO-HCP and provide controller level analysis for custom metrics gathering about the state of the management cluster kube-API.

mgmt-agent/pkg/controller/controller.go
stevekuznetsov: People are too keen to do naughty things with their interior mutable state when you end up with a struct implementing that interface, even though in theory they can achieve the same thing in a closure, which is why I land on the other side - just run me a func that does one stupid simple thing and don't make me think about your test mock impl.

mgmt-agent/pkg/controller/controller.go
```diff
@@ -0,0 +1,305 @@
...
+
+	nicCount := countSwiftNICs(vm)
+	logger.Info("Setting SWIFT NIC capacity on node", "node", name, "count", nicCount)
+
+	return applycorev1.Node(name).
+		WithStatus(applycorev1.NodeStatus().
+			WithCapacity(corev1.ResourceList{
+				SwiftNICResourceName: *resource.NewQuantity(int64(nicCount), resource.DecimalSI),
+			}),
+		), nil
```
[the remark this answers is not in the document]
stevekuznetsov: Allocatable is handled by the kubelet.

stevekuznetsov:
/label lgtm

re-tagging after rebase

stevekuznetsov:
/label lgtm

re-tagging after rebase

PR #5224, opened 2026-05-11
Bump aro-tools: add repo and baseRef fields to ProwJobStep

Mustafa-Ali-code:
### What

Add optional `repo` and `baseRef` fields to `ProwJobStep` in the pipeline types and JSON schema. When set, sdp-pipelines will pass them as `--repo` and `--base-ref` to prowjobexecutor, overriding the defaults

### Why

When using the `commit` field to pin a ProwJob to a specific source commit, the prowjobexecutor defaults to `ARO-HCP` repo on `main` branch. Classic E2E jobs target `ARO-RP` on `master`, so the repo and base ref need to be overridable per step

The diff link in ARO-Tools: https://github.com/Azure/ARO-Tools/compare/621e0b33dfa1...ecd1d9a0477a


Mustafa-Ali-code:
The diff for the ARO-Tools in this bump is targeting includes 3 commits, all of which are adding optional fields (one to the LogsStep and the other to the ProwJobStep), and a Fixing schema change to skip datasource sync. None of those changes is doing anything that would affect the existing behavior. The test failure seen in prow seem to be complaining about a nodepool version upgrade, which is not related to the changes mentioned in the diff.

PR #5227, opened 2026-05-12
Enable cleanup-sweeper policy-driven rg-ordered workflow for ci01

roivaz:
* Add the rule to match ci01 resource groups
* Document the requirement for a policy that sets the createdAt tag on resource groups.
* Add a script to deploy the "ARO-CreatedAt Tag" policy in a subscription.

https://redhat.atlassian.net/browse/ARO-26089


raelga (approved):
/lgtm
/approve

PR #5236, opened 2026-05-12
Skip datasources sync for prow

janboll:
### What/Why
Add skipDatasourceSync to skip sync for certain environments.
Also add the skip for prow environments.
We commonly see 409 caused by concurrent updates

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -343,6 +343,7 @@ defaults:
     maxActiveTimeSeries: 2000000
     maxEventsPerMinute: 2000000
     alertRuleOwningTeamTag: ""
+    skipDatasourceSync: false
     alertsEnabled: false
     icm:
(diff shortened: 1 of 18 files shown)


config/config.yaml
```diff
@@ -1477,6 +1478,8 @@ clouds:
               alertsEnabled: true
       prow:
         defaults:
+          monitoring:
+            skipDatasourceSync: true
```
roivaz, line 1507: We need to add the same skip for the `ci01` env (it's the prow env that targets the new dedicated subscription)

raelga (approved):
/lgtm
/approve

config/config.yaml
janboll: added

raelga (approved):
/lgtm
/approve

PR #5238, opened 2026-05-12
use kube-applier to get data from management clusters

deads2k:
builds on other PRs and uses kube-applier instead of maestro for reads.


backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -0,0 +1,234 @@
...
+		"CreateClusterScopedReadDesires",
+		resourcesDBClient,
+		informers,
+		nil,
+		1*time.Minute,
+		syncer,
+	)
+}
+
+func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
```
deads2k, line 92: refactor to use the NeedsWork function on inputs, first cached, then live. This is crazy expensive. Check the cosmos resources first before ever touching cluster-service.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	if mcResourceID == nil {
+		return nil
+	}
+
+	csCluster, err := c.clusterServiceClient.GetCluster(ctx, *existingCluster.ServiceProviderProperties.ClusterServiceID)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get Cluster from Cluster Service: %w", err))
+	}
+	csClusterDomainPrefix := csCluster.DomainPrefix()
+	csClusterID := existingCluster.ServiceProviderProperties.ClusterServiceID.ID()
```
deads2k: all this for these two fields?  Surely we can store these in a more sensible place.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -0,0 +1,234 @@
...
+		// Registry doesn't have an entry yet for this MC (e.g. the fleet
+		// lister hasn't caught up). Skip and rely on retrigger.
+		return nil
+	}
+	parent := database.ResourceParent{
+		SubscriptionID:    key.SubscriptionID,
+		ResourceGroupName: key.ResourceGroupName,
+		ClusterName:       key.HCPClusterName,
+	}
+	crud, err := kaClient.ReadDesires(parent)
```
deads2k: smells weird.  I think other database storages we access with individual field names.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -0,0 +1,234 @@
...
+		if _, err := crud.Create(ctx, desired, nil); err != nil {
+			return utils.TrackError(fmt.Errorf("create ReadDesire: %w", err))
+		}
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("get ReadDesire: %w", err))
+	}
+	if managementClusterEqual(existing.Spec.ManagementCluster, desired.Spec.ManagementCluster) &&
+		existing.Spec.TargetItem == desired.Spec.TargetItem {
```
deads2k: this check should definitely be in "NeedsWork".  We know the name of the resource, so we can get it exactly.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	desired.Status = existing.Status
+	if _, err := crud.Replace(ctx, desired, nil); err != nil {
+		return utils.TrackError(fmt.Errorf("replace ReadDesire: %w", err))
+	}
+	return nil
+}
+
+// managementClusterEqual compares two ReadDesire.Spec.ManagementCluster pointers
+// for equality. Both may be nil; non-nil values compare by string form.
+func managementClusterEqual(a, b *azcorearm.ResourceID) bool {
```
deads2k: check for a helper in a utils package near where we compare ResourceType and create a similar function there.

backend/pkg/controllers/create_nodepool_scoped_read_desires_controller.go
```diff
@@ -0,0 +1,175 @@
...
+	return controllerutils.NewNodePoolWatchingController(
+		"CreateNodePoolScopedReadDesires",
+		resourcesDBClient,
+		informers,
+		1*time.Minute,
+		syncer,
+	)
+}
+
+func (c *createNodePoolScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPNodePoolKey) error {
```
deads2k, line 79: similar comments to the ClusterScoped one.

backend/pkg/controllers/read_and_persist_nodepool_scoped_kube_content_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	return controllerutils.NewNodePoolWatchingController(
+		"ReadAndPersistNodePoolScopedKubeContent",
+		resourcesDBClient,
+		informers,
+		1*time.Minute,
+		syncer,
+	)
+}
+
+func (c *readAndPersistNodePoolScopedKubeContentSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPNodePoolKey) error {
```
deads2k: Ok, we're only doing this until we can make all code directly aware of ReadDesire.  We'll do that in the next step.

deads2k: Actually, do we really need it? We'd get eventually consistent during a downgrade when maestro bundles are recreated. Let's eliminate these.

deads2k:
`microsoft.redhatopenshift/hcpopenshiftclusters/nodepools: [aro-hcp-snapshot] [microsoft.redhatopenshift/hcpopenshiftclusters/nodepools] backend/maestroBundleAssociations returns results expand_more`

I bet this actually worked and this is something from steveK

backend/pkg/controllers/cleanup_legacy_maestro_readonly_bundles_controller.go
```diff
@@ -0,0 +1,367 @@
...
+
+// buildMaestroClientsByProvisionShard mirrors the orphan-cleanup
+// controller's helper of the same name. Defined here so the cleanup
+// controller is self-contained; the orphan controller has its own copy
+// because the two will be removed on different timelines (this one
+// retires when MaestroReadonlyBundles is empty everywhere; the orphan
+// controller stays as long as any bundles can drift orphaned).
+func (c *cleanupLegacyMaestroReadonlyBundles) buildMaestroClientsByProvisionShard(ctx context.Context) (map[string]*shardMaestroClient, error) {
+	out := map[string]*shardMaestroClient{}
+	iter := c.clusterServiceClient.ListProvisionShards()
```
geoberle: we have all the shards as ManagementCluster documents in cosmosdb. they contain all the data you need to build a maestro client

also in `c.maestroClientForCluster` consider leveraging the SPC.status.ManagementClusterResourceID to match a maestro client build from a ManagementCluster cosmosdb doc.

doing both avoids CS API calls

deads2k: Great news!

backend/pkg/controllers/create_nodepool_scoped_read_desires_controller.go
```diff
@@ -0,0 +1,175 @@
...
+// (namespace = "ocm-<env>-<csClusterID>", name =
+// "<csClusterDomainPrefix>-<csNodePoolID>") match what CS itself uses;
+// see createNodePoolScopedMaestroReadonlyBundlesSyncer for the original
+// derivation.
+func nodePoolTarget(envIdentifier, csClusterID, csClusterDomainPrefix, csNodePoolID string) kubeapplier.ResourceReference {
+	return kubeapplier.ResourceReference{
+		Group:     hsv1beta1.SchemeGroupVersion.Group,
+		Version:   hsv1beta1.SchemeGroupVersion.Version,
+		Resource:  "nodepools",
+		Namespace: fmt.Sprintf("ocm-%s-%s", envIdentifier, csClusterID),
```
geoberle: do we have a helper func for HCP NS creation somewhere?

backend/pkg/controllers/create_nodepool_scoped_read_desires_controller.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// createNodePoolScopedReadDesiresSyncer ensures a ReadDesire exists per
+// HCPNodePool pointing at the nodepool's Hypershift NodePool object in the
+// management cluster. The kube-applier sidecar on the management cluster
+// observes the NodePool via that ReadDesire and writes the observed state
+// into ReadDesire.Status.KubeContent. The "read+persist" controller mirrors
+// that into ManagementClusterContent.
```
geoberle: stale comment? i was not able to see anything that still writes ManagementClusterContent

backend/pkg/controllers/cleanup_legacy_maestro_readonly_bundles_controller.go
geoberle: looks good now

backend/pkg/controllers/create_nodepool_scoped_read_desires_controller.go
geoberle: thx

PR #5239, opened 2026-05-12
ARO-26793: Relax nodepool minor version skip from +1 to +2 to align with the Kubernetes version skew policy

adalrsjr1:
[ARO-26793](https://redhat.atlassian.net/browse/ARO-26793)

  ### What

  Relax nodepool minor version skip validation from +1 to +2, allowing nodepool upgrades to span up to 2 minor versions (e.g., 4.20→4.22).

  ### Why

  1. **Kubernetes version skew policy** allows kubelet to be N-2 minor versions behind kube-apiserver — our nodepool validation should align with this
  2. **HCP Replace strategy** destroys and recreates nodes at the target version (not in-place upgrade), so there is no step-through requirement

  The CP version upper bound (`desiredVersion <= lowestCPVersion`) already prevents nodepools from exceeding the CP. The CP upgrade restriction (`OpenshiftVersionAtMostOneMinorSkew`) is unchanged.

  No Clusters Service changes are needed — CS already delegates all ARO HCP version policy to the RP [(CS MR !322)](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/322).

### Testing
  - **Unit tests** (`internal/validation`, `internal/admission`): +2 case flipped from negative to positive; +3 case added as new boundary
  - **Controller unit tests** (`backend/pkg/controllers/upgradecontrollers`): same treatment for `ValidateDesiredNodePoolVersion` table tests, plus `SyncOnce_SkipMinorVersionFails` updated to +3 scenario
  - **Integration tests** (`test-integration/frontend`): CP bumped from 4.22→4.24, skip-minor-version fixture updated from +2 to +3 target, added step 11 for +2 success case (4.20.8→4.22.0)
  - **E2E tests**: added `"should upgrade a nodepool skipping one minor version (+2)"` (4.20→4.22); removed Skip from no-edge upgrade test; filtered pre-release versions from `GetVersionPairWithoutUpgradeEdge` to avoid unstable EC builds

  - **Merge order**: This PR must be merged after #5153 (Cincinnati edge removal) and before #5184 (downgrade support). It is branched from #5153.
  - **One-line production change**: `validators.go:906` — `lowest.Minor+1` → `lowest.Minor+2`
  - The integration test CP version was bumped from 4.22.0 to 4.24.0 so that the +3 skip target (4.23.0) passes the CP upper bound and correctly hits the minor skip check
  - Doc comments aligned across `validators.go`, `admit_nodepool.go`, and `nodepool_version_controller.go`
  - `GetVersionPairWithoutUpgradeEdge` now filters pre-release versions (EC, RC, nightly) to avoid E2E failures from unstable builds lacking proper node images


internal/admission/admit_nodepool.go
```diff
@@ -128,7 +128,7 @@ func AdmitNodePoolUpdate(newNodePool, oldNodePool *api.HCPOpenShiftClusterNodePo
 // It checks:
 //   - No downgrade: new version >= old version
 //   - No major version change: new major == old major (unless FeatureExperimentalReleaseFeatures is registered)
-//   - No skipping minor versions: new minor <= old minor + 1
+//   - Minor version upgrades limited to +2: new minor <= old minor + 2
```
JameelB: would it be worth adding an integration test to test a +2 minor update succeeds as well on top of the failure scenarios?

JameelB (commented):
overall lgtm, just one minor comment

adalrsjr1:
/hold
pending on https://github.com/Azure/ARO-HCP/pull/5153 be merged

internal/validation/validators.go
```diff
@@ -931,9 +931,9 @@ func ValidateNodePoolUpgrade(desiredVersion semver.Version, activeVersions []api
 	}
 
 	// Minor skip validation
-	if lowest != nil && desiredVersion.Minor > lowest.Minor+1 {
+	if lowest != nil && desiredVersion.Minor > lowest.Minor+2 {
```
ahitacat, line 934: Why limit this?

This scenario could only happens if we have (hypothetically) a node pool which version is < 3 the control plane minor version. Otherwise the desiredVersion will be greater than the lowestCPVersion (check done in line 918).

But that scenario could not happen as there is also a skew control and nodepool should be at least < 2 the control plane minor version.

adalrsjr1: It's a safeguard that the skew won't never be larger than +2. However it is implicitly ensured before:

- CP can't upgrade past NP+2 (enforced by `clusterVersionSkewVersusNodePool` in utils.go:85)
- NP desired can't exceed CP (line 918)

@ahitacat If you agree, I will remove this check (and consequently all tests related), which will make this PR much leaner.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
```diff
@@ -99,8 +99,8 @@ func NewNodePoolVersionController(
 //   - The desired version must satisfy:
 //   - Not less than the highest active version (no downgrades)
 //   - Not greater than the lowest control plane version (node pools cannot exceed CP)
-//   - Not skip minor versions (only z-stream or +1 minor allowed)
-//   - Have a valid upgrade path in Cincinnati from the current version
+//   - Minor version upgrades limited to +2
```
ahitacat: I think this can be rephrase as it is covered by the CP-NP skew (noted that if that change this will change)

ahitacat (commented):
LGTM minor nitpick.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
adalrsjr1: done

adalrsjr1:
Failed due to nodepool creation timeout in the Private KeyVault test — the hosted control plane's kube-apiserver was unhealthy during bootstrap (azure-kms-provider-active crash-looping), preventing node registration. All alerts self-resolved.
/retest-required

PR #5241, opened 2026-05-12
fix(observability): rebuild etcd-availability as fleet summary

swiencki:
### What

Rebuild `etcd-availability` as a fleet summary and move it to `observability/grafana-dashboards/sre/user-journey/`.

### Why

Each stat panel rendered as N unlabeled tiles in prod (one per regional HCP Prom workspace) because Mixed datasource + multi-select All fanned queries out with no reduction step.

- Stat panels: `[merge, reduce reduceFields <sum|max>]` collapses N regional frames into one tile.
- Tables: `merge` unions per-region rows; sort fields fixed.
- Timeseries: `{{cluster}}` legend; p99 panels aggregate `by (le, cluster)`.
- Stale per-region thresholds and legendFormats recalibrated.
- Worst Regional P99 documented honestly (not a true fleet p99; PromQL cannot federate histograms across independent workspaces).
- `observability.yaml` retargets the User Journey folder to the new path.

### Testing

`make -C observability verify` passes. JSON valid at schemaVersion 41. Confirmed in dev Grafana that each stat tile renders as a single number; warning triangles in dev are the documented stale-datasource hazard, not a dashboard issue.

The Worst Regional P99 tiles use Reduce(max) across per-region p99s. This is the only mathematically defensible single-number representation without federation infra (Thanos/Mimir). Called out in the Dashboard Info panel.

### DIFF
--- a/observability/grafana-dashboards/sre/user-journey/etcd-availability.json
+++ b/observability/grafana-dashboards/sre/user-journey/etcd-availability.json
@@ -316,7 +316,7 @@
         "type": "datasource",
         "uid": "-- Mixed --"
       },
-      "description": "Highest regional p99 WAL fsync duration across the fleet (max of per-region p99s). NOTE: this is NOT a true cross-region p99; PromQL cannot aggregate histograms across independent Prometheus workspaces. Use this as a \"any region degraded?\" signal. SLO target per region: < 10ms.",
+      "description": "Worst per-cluster p99 WAL fsync duration across the fleet. Computes p99 per cluster, then takes the max across all clusters and regions. Use this as a \"any cluster degraded?\" signal. SLO target: < 10ms.",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -383,7 +383,7 @@
           "refId": "A"
         }
       ],
-      "title": "Worst Regional P99 WAL Fsync (5m)",
+      "title": "Worst Cluster P99 WAL Fsync (5m)",
       "transformations": [
         {
           "id": "merge",
@@ -406,7 +406,7 @@
         "type": "datasource",
         "uid": "-- Mixed --"
       },
-      "description": "Highest regional p99 backend commit duration across the fleet. NOT a true cross-region p99 (see WAL Fsync panel description). SLO target per region: < 25ms.",
+      "description": "Worst per-cluster p99 backend commit duration across the fleet. Computes p99 per cluster, then takes the max across all clusters and regions. SLO target: < 25ms.",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -473,7 +473,7 @@
           "refId": "A"
         }
       ],
-      "title": "Worst Regional P99 Backend Commit (5m)",
+      "title": "Worst Cluster P99 Backend Commit (5m)",
       "transformations": [
         {
           "id": "merge",
@@ -496,7 +496,7 @@
         "type": "datasource",
         "uid": "-- Mixed --"
       },
-      "description": "Highest regional p99 etcd peer network RTT across the fleet. NOT a true cross-region p99 (see WAL Fsync panel description). SLO target per region: < 50ms.",
+      "description": "Worst per-cluster p99 etcd peer network RTT across the fleet. Computes p99 per cluster, then takes the max across all clusters and regions. SLO target: < 50ms.",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -563,7 +563,7 @@
           "refId": "A"
         }
       ],
-      "title": "Worst Regional P99 Peer RTT (5m)",
+      "title": "Worst Cluster P99 Peer RTT (5m)",
       "transformations": [
         {
           "id": "merge",
--- a/observability/grafana-dashboards/sre/user-journey/etcd-availability.json
+++ b/observability/grafana-dashboards/sre/user-journey/etcd-availability.json
@@ -754,6 +754,20 @@
         {
           "id": "merge",
           "options": {}
+        },
+        {
+          "id": "organize",
+          "options": {
+            "excludeByName": {
+              "Time": true
+            },
+            "includeByName": {},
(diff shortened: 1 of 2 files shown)


swiencki:
Addressed in 9092b548f: set `decimals=0` on the integer-count panels (Fleet Total Stable Clusters, Fleet Total Leader Changes, Fleet Clusters Monitored, Leader Instability hourly table). Counts now render as whole numbers.

observability/grafana-dashboards/user-journey/etcd-availability.json
```diff
@@ -377,7 +383,7 @@
           "refId": "A"
         }
       ],
-      "title": "Fleet P99 WAL Fsync Duration",
```
SudoBrendan, line 380: I don't think that the expression matches the title. When we `sum by ... (le)` across all clusters - that gives us a regional aggregate. This means that a single "bad" cluster can get diluted by the fleet. We probably want a max per cluster or a max in the maxes of individual clusters or something?

observability/grafana-dashboards/user-journey/etcd-availability.json
```diff
@@ -466,7 +473,7 @@
           "refId": "A"
         }
       ],
-      "title": "Fleet P99 Backend Commit Duration",
```
SudoBrendan, line 469: same as above; `sum ... by (le)` should probably instead be a `max` of some kind if we're calling it "Worst".

observability/grafana-dashboards/user-journey/etcd-availability.json
```diff
@@ -555,7 +563,7 @@
           "refId": "A"
         }
       ],
-      "title": "Fleet P99 Peer Network RTT",
```
SudoBrendan, line 558: Same as above - `sum ... by (le)` -> `max` if we're calling it "Worst"?

observability/grafana-dashboards/sre/user-journey/etcd-availability.json
```diff
@@ -112,14 +113,14 @@
             "uid": "$datasource"
           },
           "editorMode": "code",
-          "expr": "((count(max by (cluster) (changes(etcd_server_leader_changes_seen_total[1d])) <= 5) or vector(0)) / count(count by (cluster) (etcd_server_leader_changes_seen_total))) * 100",
+          "expr": "count(max by (cluster) (changes(etcd_server_leader_changes_seen_total[1d])) <= 5) or vector(0)",
```
SudoBrendan: changes() just counts +1 for any change (e.g. 3 leader elections in a scrape interval returns 1). We likely want increase() instead. A few more instances of this throughout.

SudoBrendan (commented):
Overall this is really good, might consider a few tweaks to align on titles and queries - wdyt?

swiencki:
Addressed in 9d554f0ea:

- **`changes()` -> `increase()`**: All 9 occurrences of `changes(etcd_server_leader_changes_seen_total[...])` replaced with `increase()`. `changes()` only counts +1 per scrape interval regardless of how many elections happened; `increase()` gives the actual count.

- **`sum by (le)` -> `max(histogram_quantile(... sum by (cluster, le) ...))`**: The three "Worst Regional" panels (WAL Fsync, Backend Commit, Peer RTT) now compute p99 per cluster, then take the max. Previously `sum(rate(...)) by (le)` aggregated across all clusters, diluting any single bad cluster in the fleet average.

observability/grafana-dashboards/user-journey/etcd-availability.json
swiencki: Same fix applied here: query uses per-cluster max, title updated to "Worst Cluster P99 Peer RTT" in e8897e3.

observability/grafana-dashboards/user-journey/etcd-availability.json
swiencki: Same fix applied here: query uses per-cluster max, title updated to "Worst Cluster P99 Backend Commit" in e8897e3.

observability/grafana-dashboards/sre/user-journey/etcd-availability.json
swiencki: Fixed in 9d554f0: all 9 occurrences of `changes()` replaced with `increase()` across the dashboard.

observability/grafana-dashboards/user-journey/etcd-availability.json
swiencki: Addressed across two commits:
- 9d554f0: changed query to `max(histogram_quantile(0.99, sum by (cluster, le) ...))` so a single bad cluster is not diluted
- e8897e3: renamed panels from "Worst Regional P99" to "Worst Cluster P99" and updated descriptions to match the per-cluster max semantics

raelga (approved):
All 6 violator tables dropped their `organize` transforms, which means they now show raw field names (`cluster`, `Value`, `Time`) instead of the previous human-friendly labels (`Cluster Name`, `P99 Fsync (ms)`, `Changes/Hour`, etc.). The `Time` column is also now visible, which isn't useful for instant queries.

Was this intentional because `organize` doesn't work well with Mixed datasource fan-out? If so, worth a comment in the info panel. If not, restoring the column renaming would improve readability.

Overall `/lgtm`, you are already Tide ready, but may worth a follow-up PR.

/lgtm

swiencki (commented):
Thanks for the LGTM. You're right, the organize transforms were lost during the fleet summary rebuild. The original tables had column renames (cluster -> Cluster Name, Value -> P99 Fsync (ms), etc.) and hid the Time column.

Follow-up PR: #5430

raelga (approved):
/lgtm
/approve

Looks good to me. This is an observability-only dashboard change, CI is green, and the Mixed datasource merge/reduce handling addresses the fleet-summary rendering issue.

PR #5246, opened 2026-05-12
delete red hat openshift SALs during RG clean up

tyrenyabe:
https://redhat.atlassian.net/browse/ARO-26088

### What

Adds optional FPA authenticated cleanup for dangling Red Hat OpenShift service association links before deleting resource groups.

When the new FPA flags are provided, resource group cleanup discovers Red Hat OpenShift SALs under VNets/subnets and deletes them using an FPA authenticated ARM client before continuing with the existing resource group deletion flow.

When the FPA flags are not provided, cleanup behavior remains unchanged.

### Why

Expired e2e/dev resource group cleanup can fail when a customer integration subnet still contains a dangling `RedHatOpenShift` service association link. Azure then blocks subnet/VNet/RG deletion with `InUseSubnetCannotBeDeleted`.

### Testing

```bash
go test -run '^$' -v \
  ./test/util/framework/... \
  ./test/util/cleanup/resourcegroups/... \
  ./test/cmd/aro-hcp-tests/cleanup/...

go run ./test/cmd/aro-hcp-tests cleanup resource-groups --help | grep fpa
```

After this ARO-HCP PR is up, the next PR is the `openshift/release` wiring:

```text
download firstPartyCert2 to a temp file
pass --fpa-client-id
pass --fpa-tenant-id
pass --fpa-cert-path
```

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/test/cmd/aro-hcp-tests/cleanup/cmd.go
+++ b/test/cmd/aro-hcp-tests/cleanup/cmd.go
@@ -128,6 +128,9 @@ func newCleanupResourceGroupsCommand() *cobra.Command {
 	cmd.Flags().StringArrayVar(&rawOpt.ExcludeLocations, "exclude-location", rawOpt.ExcludeLocations, "Do not delete resource groups in these Azure locations (repeatable)")
 	cmd.Flags().BoolVar(&rawOpt.Tracked, "tracked", rawOpt.Tracked, "Use tracked resource groups")
 	cmd.Flags().StringVar(&rawOpt.SharedDir, "shared-dir", rawOpt.SharedDir, "Shared directory to use for tracked resource groups")
+	cmd.Flags().StringVar(&rawOpt.FPAClientID, "fpa-client-id", rawOpt.FPAClientID, "Client ID of the FPA used to delete Red Hat OpenShift service association links")
+	cmd.Flags().StringVar(&rawOpt.FPATenantID, "fpa-tenant-id", rawOpt.FPATenantID, "Tenant ID of the FPA used to delete Red Hat OpenShift service association links")
+	cmd.Flags().StringVar(&rawOpt.FPACertPath, "fpa-cert-path", rawOpt.FPACertPath, "Certificate path of the FPA used to delete Red Hat OpenShift service association links")
 
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
 		ctx, cancel := signal.NotifyContext(cmd.Context(), os.Interrupt)
--- a/test/cmd/aro-hcp-tests/cleanup/cmd.go
+++ b/test/cmd/aro-hcp-tests/cleanup/cmd.go
@@ -129,7 +129,6 @@ func newCleanupResourceGroupsCommand() *cobra.Command {
 	cmd.Flags().BoolVar(&rawOpt.Tracked, "tracked", rawOpt.Tracked, "Use tracked resource groups")
 	cmd.Flags().StringVar(&rawOpt.SharedDir, "shared-dir", rawOpt.SharedDir, "Shared directory to use for tracked resource groups")
 	cmd.Flags().StringVar(&rawOpt.FPAClientID, "fpa-client-id", rawOpt.FPAClientID, "Client ID of the FPA used to delete Red Hat OpenShift service association links")
-	cmd.Flags().StringVar(&rawOpt.FPATenantID, "fpa-tenant-id", rawOpt.FPATenantID, "Tenant ID of the FPA used to delete Red Hat OpenShift service association links")
 	cmd.Flags().StringVar(&rawOpt.FPACertPath, "fpa-cert-path", rawOpt.FPACertPath, "Certificate path of the FPA used to delete Red Hat OpenShift service association links")
 
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
--- a/test/util/framework/per_test_framework.go
+++ b/test/util/framework/per_test_framework.go
@@ -318,12 +318,11 @@ func isIgnorableResourceGroupCleanupError(err error) bool {
 
 type FPACredentials struct {
 	ClientID string
-	TenantID string
 	CertPath string
 }
 
 func (c FPACredentials) IsConfigured() bool {
-	return c.ClientID != "" && c.TenantID != "" && c.CertPath != ""
+	return c.ClientID != "" && c.CertPath != ""
 }
 
 type CleanupWorkflow string
--- a/test/util/framework/per_test_framework.go
+++ b/test/util/framework/per_test_framework.go
@@ -554,13 +554,6 @@ func (tc *perItOrDescribeTestContext) cleanupResourceGroup(ctx context.Context, 
 		ginkgo.GinkgoLogr.Info("no left behind managed resource groups found", "resourceGroup", resourceGroupName)
 	}
 
-	if fpaCredentials.IsConfigured() {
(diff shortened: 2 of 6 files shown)


test/cmd/aro-hcp-tests/cleanup/cmd.go
```diff
@@ -128,6 +128,9 @@ func newCleanupResourceGroupsCommand() *cobra.Command {
 	cmd.Flags().StringArrayVar(&rawOpt.ExcludeLocations, "exclude-location", rawOpt.ExcludeLocations, "Do not delete resource groups in these Azure locations (repeatable)")
 	cmd.Flags().BoolVar(&rawOpt.Tracked, "tracked", rawOpt.Tracked, "Use tracked resource groups")
 	cmd.Flags().StringVar(&rawOpt.SharedDir, "shared-dir", rawOpt.SharedDir, "Shared directory to use for tracked resource groups")
+	cmd.Flags().StringVar(&rawOpt.FPAClientID, "fpa-client-id", rawOpt.FPAClientID, "Client ID of the FPA used to delete Red Hat OpenShift service association links")
+	cmd.Flags().StringVar(&rawOpt.FPATenantID, "fpa-tenant-id", rawOpt.FPATenantID, "Tenant ID of the FPA used to delete Red Hat OpenShift service association links")
```
bennerv: The tenant ID is always going to be the same in which the resource cleaner is running in.  Therefore tenant should be redundant here.

test/util/framework/per_test_framework.go
```diff
@@ -287,6 +287,16 @@ func isIgnorableResourceGroupCleanupError(err error) bool {
 	return isResourceGroupNotFoundError(err)
 }
 
+type FPACredentials struct {
+	ClientID string
+	TenantID string
+	CertPath string
+}
+
+func (c FPACredentials) IsConfigured() bool {
```
bennerv, line 324: we shouldn't delete SALs unless we're in development as the production / integration / staging resource provider will clean them up for us (otherwise there's a bug in our RP).  

In dev, we have a resource cleaner which can/will clean up individual's dev environments which will leave behind stale resources which makes sense to clean up, but only in dev environments.

tuxerrante (commented):
Nice work on the SAL cleanup integration. Architecture, security, type safety, and error handling all look clean. Two things to address:

---

### 1. Race condition: missing `isResourceGroupNotFoundError` guard on subnet listing

[`serviceassociationlink_helper.go:110`](https://github.com/Azure/ARO-HCP/blob/153f569322f34d28ca1849e07f3c74b6bb226823/test/util/framework/serviceassociationlink_helper.go#L110)

VNet listing (line 100) gracefully returns `nil` when the RG is already gone, but subnet listing does not. If the RG is deleted between listing VNets and listing subnets (possible during concurrent cleanup), the subnet call propagates an error instead of handling it gracefully.

Suggest adding the same guard:
```go
if isResourceGroupNotFoundError(err) {
    return nil
}
```

This matches the pattern used in [`per_test_framework.go:366`](https://github.com/Azure/ARO-HCP/blob/153f569322f34d28ca1849e07f3c74b6bb226823/test/util/framework/per_test_framework.go#L366).

---

### 2. CI failures

`ci/prow/e2e-parallel` and `ci/prow/verify` are failing — worth investigating before merge. The `verify` failure may be a missing license header on the new `serviceassociationlink_helper.go` file or a `go generate` artifact mismatch.

---

### Informational: API version

`serviceAssociationLinkAPIVersion = "2021-08-01"` — the `armnetwork/v6` SDK's own `ServiceAssociationLinksClient` uses `2024-05-01`. Since the SDK only exposes `List` (no `Delete` for SALs), using `armresources.Client.BeginDeleteByID` with an explicit version is the right approach. `2021-08-01` may be deliberately chosen as a known-working version — consider updating to `2024-05-01` for consistency but not blocking.

test/util/framework/per_test_framework.go
tyrenyabe: Thanks for the info! I've added this to ensure that this will only run when the optional params are passed in. [I have a change on the prow side](https://github.com/openshift/release/pull/79144) to obtain FPA creds and pass those in. I will ensure that I check vault secret profile == dev  first before obtaining creds.

test/cmd/aro-hcp-tests/cleanup/cmd.go
tyrenyabe: removed. using test context for tenant id

tyrenyabe:
@microsoft-github-policy-service agree

test/util/framework/per_test_framework.go
```diff
@@ -515,6 +525,13 @@ func (tc *perItOrDescribeTestContext) cleanupResourceGroup(ctx context.Context,
 		ginkgo.GinkgoLogr.Info("no left behind managed resource groups found", "resourceGroup", resourceGroupName)
 	}
 
+	if fpaCredentials.IsConfigured() {
+		ginkgo.GinkgoLogr.Info("deleting any remaining RedHatOpenShift service association links in resource group", "resourceGroup", resourceGroupName)
+		if err := tc.deleteRedHatOpenShiftServiceAssociationLinks(ctx, resourceGroupName, fpaCredentials); err != nil {
+			return fmt.Errorf("failed to delete RedHatOpenShift service association links in %q: %w", resourceGroupName, err)
+		}
+	}
+
```
roivaz: `cleanupResourceGroup(...)` should not delete SALs -- it should only detect and report them. The reasoning mirrors the existing managed-resource-group pattern already documented in `docs/ci/cleanup.md` ("Why dev cleanup actively deletes managed resource groups but public-cloud cleanup does not"):

- **`cleanupResourceGroup(...)` (`--mode=standard`)** is the path used for INT/STG/PROD. In those environments there is always a live RP that owns SAL lifecycle, just as it owns managed resource groups. The logic here should match how we handle managed RGs: if leftover SALs are found, surface an error  but don't attempt to delete them -- the RP is the owner.

- **`cleanupResourceGroupNoRP(...)` (`--mode=no-rp`)** is the path used in DEV. The RP may already be gone (a fresh RP is created per e2e job run and then torn down at the end), so there is no owner left to clean up SALs. This is where active SAL deletion belongs.

See `docs/ci/cleanup.md` for the full rationale on why these two paths behave differently.

roivaz: An update to `docs/ci/cleanup.md` to include the SALs deletion workflow for DEV should also be included with the PR. Thanks!

tyrenyabe: @roivaz Updated! SAL deletion now only happens in `cleanupResourceGroupNoRP` (no-rp mode for dev). Also added documentation to docs/cleanup.md explaining the behavior difference.

test/util/framework/per_test_framework.go
```diff
@@ -490,7 +500,7 @@ func (tc *perItOrDescribeTestContext) waitForManagedResourceGroupsDeletion(ctx c
 // 1. delete all HCP clusters via the RP and wait for success
 // 2. wait for any managed resource groups to be deleted
 // 3. delete the resource group
-func (tc *perItOrDescribeTestContext) cleanupResourceGroup(ctx context.Context, resourceGroupName string, timeout time.Duration) error {
+func (tc *perItOrDescribeTestContext) cleanupResourceGroup(ctx context.Context, resourceGroupName string, timeout time.Duration, fpaCredentials FPACredentials) error {
```
roivaz: Better drop these 2 params from the function signature if they are unused in the `--mode=standard` workflow.

tyrenyabe: good call, removed these 2 params from this function signature

PR #5251, opened 2026-05-13
delete orphaned operation documents from cosmos

deads2k:
Adds a per-subscription mismatch controller that walks the resources container via the untyped CRUD client and deletes operation rows that:

  * are missing the top-level resourceID — unreachable by ARM path, the case the data dumper had to be made nil-safe for (see internal/serverutils/dump_data.go), or
  * carry a legacy pipe-delimited cosmosID — pre-migration cruft whose derived resourceID looks fine on read but whose row is stale; deleting lets the frontend re-write under the new UUID-cosmosID scheme on next access.

To make those rows actually reachable through the iterator, CosmosToInternal for *TypedDocument no longer errors when an old cosmosID can't be back-parsed; it surfaces the row with nil ResourceID and lets cleanup controllers handle it. Three iter consumers had to be hardened against the now-possible nil ResourceID:

  * controllerutils.DeleteRecursively: falls back to DeleteByCosmosID instead of panicking on Delete(nil)
  * operationcontrollers.SetDeleteOperationAsCompleted: same fallback rather than refusing-to-delete an unreachable child
  * serverutils.DumpDataToLogger: routes ResourceID through a nil-safe stringifier (mirrors the fix already on main in commit 6eb9565be)

The controller is wired into backend.go alongside deleteOrphanedCosmosResources at the same threadiness.

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/internal/database/convert_any.go
+++ b/internal/database/convert_any.go
@@ -47,13 +47,13 @@ func CosmosToInternal[InternalAPIType, CosmosAPIType any](obj *CosmosAPIType) (*
 				return any(cosmosObj).(*InternalAPIType), nil
 			}
 
-			// fill in the new ResourceID field for old data that is missing it. This means we didn't migrate something.
-			// this will happen frequently when the new backend is running against an old frontend.
-			resourceIDFromOldCosmosID, err := oldCosmosIDToResourceID(cosmosObj.ID)
-			if err != nil {
-				return nil, utils.TrackError(fmt.Errorf("expected old cosmosID and got %q", castObj.ID))
+			// Best-effort: derive ResourceID from a legacy pipe-delimited cosmosID. When that fails
+			// (e.g. the cosmosID is a bare UUID, as it is for operation documents), surface the doc
+			// with nil ResourceID rather than halting the iterator — cleanup controllers handle the
+			// nil case and need to see these rows to delete them.
+			if resourceIDFromOldCosmosID, err := oldCosmosIDToResourceID(cosmosObj.ID); err == nil {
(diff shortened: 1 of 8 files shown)


internal/database/convert_any.go
```diff
@@ -47,13 +47,13 @@ func CosmosToInternal[InternalAPIType, CosmosAPIType any](obj *CosmosAPIType) (*
 				return any(cosmosObj).(*InternalAPIType), nil
 			}
 
-			// fill in the new ResourceID field for old data that is missing it. This means we didn't migrate something.
-			// this will happen frequently when the new backend is running against an old frontend.
-			resourceIDFromOldCosmosID, err := oldCosmosIDToResourceID(cosmosObj.ID)
-			if err != nil {
-				return nil, utils.TrackError(fmt.Errorf("expected old cosmosID and got %q", castObj.ID))
+			// Best-effort: derive ResourceID from a legacy pipe-delimited cosmosID. When that fails
```
miguelsorianod, line 50: Is this intended to be temporary until the cleaners do their job?

Doesn't this have wide implications on all the controllers not only the cleaner ones, where now they could receive a resource that has a nil resourceID instead of receiving an error, and forces them to need to consider that everywhere or just panic?

deads2k:
/hold

we'll deal with this PR later.

internal/database/convert_any.go
deads2k: if we resolve the immediate problem, we may build a custom select/sweep/delete flow

PR #538, opened 2024-08-27
Towards updating node pool resources

mbarnes:
### What this PR does

This is similar to [#519 Towards updating hosted cluster resources](https://github.com/Azure/ARO-HCP/pull/519) but for node pools.

I did not do extensive testing of Cluster Service validation for node pools like I did for clusters.  That will come later.  For now, this just brings node pool updates on par with clusters in terms of PUT/PATCH support, static validation, etc.

Jira: Partially addresses [ARO-7594 - Frontend can Update/Patch a NodePool CR in CS](https://issues.redhat.com/browse/ARO-7594)
Link to demo recording:

PR #5257, opened 2026-05-13
fix: self-heal stale readonly Maestro bundles with no status feedback (AROSLSRE-833)

avollmer-redhat:
[AROSLSRE-833](https://issues.redhat.com/browse/AROSLSRE-833)

### What

Add self-healing logic to the `ReadAndPersistClusterScopedMaestroReadonlyBundlesContent` controller that detects and deletes stale readonly Maestro bundles, then clears their `MaestroAPIMaestroBundleID` in the ServiceProviderCluster so the create controller can recreate them.

### Why

In production (UK South), Maestro occasionally persists a readonly bundle to Postgres but never publishes the corresponding event to the MQTT broker. This causes the bundle to exist in Maestro without ever receiving status feedback from the management cluster, leaving cluster creation stuck indefinitely in `Provisioning` state. The proper upstream fix is [Maestro PR #538](https://github.com/openshift-online/maestro/pull/538) (UndeliveredDetector), but until that ships and we update our image pin, this backend-side workaround provides automatic recovery.

### How it works

After `readAndPersistMaestroReadonlyBundleContent` runs for each bundle reference, the new `isStaleMaestroReadonlyBundle` helper checks:

1. Is the ManagementClusterContent in Cosmos `Degraded=True` with reason `MaestroBundleStatusFeedbackNotAvailable`?
2. Does the Maestro bundle have a non-zero `CreationTimestamp` older than 5 minutes?

If both conditions are met, the controller:
1. Deletes the bundle from Maestro
2. Clears `MaestroAPIMaestroBundleID` in the ServiceProviderCluster (Cosmos)
3. Logs a structured message referencing [AROSLSRE-833](https://redhat.atlassian.net/browse/AROSLSRE-833)

The create controller then recreates the bundle on its next sync, giving Maestro another chance to publish the event.

### Loop safety

The self-heal **cannot** create an infinite delete-recreate loop because:
- It only triggers on the specific `MaestroBundleStatusFeedbackNotAvailable` degraded reason (not `MaestroBundleNotFound` or others)
- Zero-value `CreationTimestamp` is skipped as a guard
- The 5-minute threshold prevents premature deletion of newly (re)created bundles
- If the recreated bundle successfully publishes to MQTT, status feedback arrives, the degraded condition clears, and the self-heal stops firing

### Testing

- 7 new unit tests for `isStaleMaestroReadonlyBundle` covering: MCC not found, not degraded, wrong degraded reason, zero timestamp, fresh bundle, stale bundle, Maestro not-found
- All 22 controller helper tests pass
- Full `backend/pkg/controllers/` test suite passes
- Full `backend/...` build passes

- The delete logic was intentionally placed in `SyncOnce` rather than `calculateManagementClusterContentFromMaestroBundle` because the latter has no access to `ServiceProviderCluster` or `resourcesDBClient` — clearing `MaestroAPIMaestroBundleID` is mandatory for the create controller to recreate the bundle (it skips bundles with non-empty IDs at `create_cluster_scoped_maestro_readonly_bundles_controller.go:144`).
- The nodepool-scoped controller (`read_and_persist_nodepool_scoped_...`) has the same pattern and may need the same fix — tracked separately.
- This is a workaround; the proper fix is upstream Maestro PR #538.

### PR Checklist
- [ ] Screenshots included (if graph/UI/metrics changes) — N/A
- [x] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide) — draft PR, will verify
- [ ] Specific reviewers tagged — requesting @deads2k @msoriano

### DIFF
--- a/backend/pkg/controllers/maestro_readonly_bundle_helpers.go
+++ b/backend/pkg/controllers/maestro_readonly_bundle_helpers.go
@@ -81,6 +81,12 @@ func resourceIdentifiersFromManifestWork(mw *workv1.ManifestWork) []api.MaestroB
 // broker (AROSLSRE-833).
 const staleReadonlyBundleThreshold = 5 * time.Minute
 
+// maxStaleBundleDeleteRetries is the maximum number of times the self-heal
+// logic will delete and recreate a stale readonly bundle before giving up.
+// This acts as a circuit breaker against infinite delete/recreate loops when
+// the failure is deterministic rather than the transient event-delivery race.
+const maxStaleBundleDeleteRetries = 3
+
 // buildInitialReadonlyMaestroBundle builds an initial readonly Maestro Bundle for a given resource specified in obj.
 // objResourceIdentifier is the resource identifier of the resource specified in obj.
 // maestroBundleNamespacedName is the namespaced name of the Maestro Bundle.
@@ -308,8 +314,18 @@ func calculateManagementClusterContentFromMaestroBundle(
 func isStaleMaestroReadonlyBundle(
 	mcc *api.ManagementClusterContent,
 	bundle *workv1.ManifestWork,
+	staleBundleDeleteCount int,
+	clusterDeleting bool,
 ) bool {
 	if mcc == nil || bundle == nil {
+		return false
+	}
+
+	if clusterDeleting {
+		return false
+	}
+
+	if staleBundleDeleteCount >= maxStaleBundleDeleteRetries {
 		return false
 	}
 
--- a/backend/pkg/controllers/maestro_readonly_bundle_helpers.go
+++ b/backend/pkg/controllers/maestro_readonly_bundle_helpers.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	workv1 "open-cluster-management.io/api/work/v1"
 
@@ -72,6 +73,13 @@ func resourceIdentifiersFromManifestWork(mw *workv1.ManifestWork) []api.MaestroB
 	}
 	return result
 }
+
+// staleReadonlyBundleThreshold is how long a readonly bundle can have no status
+// feedback before it is considered stale and deleted so the create controller
+// can recreate it. This works around a Maestro race condition where the server
+// persists the bundle to Postgres but the event is never published to the
+// broker (AROSLSRE-833).
+const staleReadonlyBundleThreshold = 5 * time.Minute
 
 // buildInitialReadonlyMaestroBundle builds an initial readonly Maestro Bundle for a given resource specified in obj.
 // objResourceIdentifier is the resource identifier of the resource specified in obj.
@@ -227,31 +235,32 @@ func getSingleResourceStatusFeedbackRawJSONFromMaestroBundle(maestroBundle *work
 }
 
 // calculateManagementClusterContentFromMaestroBundle builds the desired ManagementClusterContent from a Maestro
-// bundle reference. parentResourceID is the ARM ID of the document parent
+// bundle reference. parentResourceID is the ARM ID of the document parent.
+// It also returns the ManifestWork that was fetched from Maestro (nil when the bundle was not found).
 func calculateManagementClusterContentFromMaestroBundle(
 	ctx context.Context,
 	parentResourceID *azcorearm.ResourceID,
 	maestroBundleReference *api.MaestroBundleReference,
 	maestroClient maestro.Client,
-) (*api.ManagementClusterContent, error) {
+) (*api.ManagementClusterContent, *workv1.ManifestWork, error) {
 	managementClusterContentResourceID := controllerutils.ManagementClusterContentResourceIDFromParentResourceID(parentResourceID, maestroBundleReference.Name)
 	desired := controllerutils.NewInitialManagementClusterContent(managementClusterContentResourceID)
 
 	existingMaestroBundle, err := maestroClient.Get(ctx, maestroBundleReference.MaestroAPIMaestroBundleName, metav1.GetOptions{})
 	if err != nil && !k8serrors.IsNotFound(err) {
-		return nil, utils.TrackError(fmt.Errorf("failed to get Maestro Bundle: %w", err))
+		return nil, nil, utils.TrackError(fmt.Errorf("failed to get Maestro Bundle: %w", err))
 	}
 	if k8serrors.IsNotFound(err) {
 		degradedCondition := buildDegradedCondition(metav1.ConditionTrue, "MaestroBundleNotFound", err.Error())
 		meta.SetStatusCondition(&desired.Status.Conditions, degradedCondition)
-		return desired, nil
+		return desired, nil, nil
 	}
 
 	rawBytes, err := getSingleResourceStatusFeedbackRawJSONFromMaestroBundle(existingMaestroBundle)
 	if err != nil {
 		degradedCondition := buildDegradedCondition(metav1.ConditionTrue, "MaestroBundleStatusFeedbackNotAvailable", err.Error())
 		meta.SetStatusCondition(&desired.Status.Conditions, degradedCondition)
-		return desired, nil
+		return desired, existingMaestroBundle, nil
 	}
 
 	kubeContentMaxSizeExceeded := len(rawBytes) > kubeContentMaxSizeBytes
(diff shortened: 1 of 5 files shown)


backend/pkg/controllers/read_and_persist_cluster_scoped_maestro_readonly_bundles_content_controller.go
```diff
@@ -125,14 +128,48 @@ func (c *readAndPersistClusterScopedMaestroReadonlyBundlesContentSyncer) SyncOnc
...
+			continue
+		}
+
+		logger.Info("self-healing stale Maestro readonly bundle (AROSLSRE-833): deleting bundle and clearing ID so the create controller can recreate it",
+			"bundleName", string(maestroBundleReference.Name),
+			"maestroAPIMaestroBundleName", maestroBundleReference.MaestroAPIMaestroBundleName,
+			"maestroAPIMaestroBundleID", maestroBundleReference.MaestroAPIMaestroBundleID,
+		)
+
+		err = maestroClient.Delete(ctx, maestroBundleReference.MaestroAPIMaestroBundleName, metav1.DeleteOptions{})
```
swiencki: What happens if the delete succeeds but the SPC Replace below fails? The ID is still set, next reconcile sees NotFound but doesn't clear it, create controller skips because ID is non-empty. Feels like the same stuck state we're trying to fix.

backend/pkg/controllers/maestro_readonly_bundle_helpers.go
```diff
@@ -199,31 +207,32 @@ func getSingleResourceStatusFeedbackRawJSONFromMaestroBundle(maestroBundle *work
...
 	parentResourceID *azcorearm.ResourceID,
 	maestroBundleReference *api.MaestroBundleReference,
 	maestroClient maestro.Client,
-) (*api.ManagementClusterContent, error) {
+) (*api.ManagementClusterContent, *workv1.ManifestWork, error) {
 	managementClusterContentResourceID := controllerutils.ManagementClusterContentResourceIDFromParentResourceID(parentResourceID, maestroBundleReference.Name)
 	desired := controllerutils.NewInitialManagementClusterContent(managementClusterContentResourceID)
 
 	existingMaestroBundle, err := maestroClient.Get(ctx, maestroBundleReference.MaestroAPIMaestroBundleName, metav1.GetOptions{})
 	if err != nil && !k8serrors.IsNotFound(err) {
```
swiencki, line 256: This reason covers a bunch of failure shapes beyond the "never published" race. If any of those are deterministic we'd loop delete/recreate every reconcile.

miguelsorianod:
Some questions with a quick look taken:

Is it really needed given that there's a maestro fix underway?
* Is it resilient to failures?
* Is the stale detection and deletion responsibility really on the "read and persist maestro readonly bundle" controllers?
* The delete is being done fire and forget. When you delete a maestro bundle it doesn't get deleted instantly. Can this race with the create one
* Why is the MR unsetting the maestro api maestro bundle id and not the whole maestro bundle reference in the SPC?
* Can we have false positives? I know that when the maestro bundle is initially created it takes a while until the feedback is there, which is expected as it's just been created and it takes a while for maestro to report back (server-agent, ...)
* If the resource on the other end is deleted (for example the hostedcluster is deleted as part of the deletion process) this will make the readonly bundle be created over and over
* The stale calculation considers it stale if there's no statusfeedback available for more than 5 minutes since creation. However, could there be a case where maestro could temporarily stop reporting feedback and then back again in a short period of time?. So is this calculation really correct?
* Is modifying calculateManagementClusterContentFromMaestroBundle to return 3 parameters really appropriate

avollmer-redhat:
@miguelsorianod thanks for the review, responding inline:

> Is it really needed given that there's a maestro fix underway?

Potentially this could land faster due to maestro team capacity and is it worth having some redundancy here post maestro team fix? If we're confident the Maestro fix ships quickly, this PR may not be worth the complexity. Happy to defer to the team's judgement here.

> Is it resilient to failures?

Partially — swiencki raised the same point. If Delete succeeds but the Cosmos Replace fails (412 ETag conflict), the bundle is gone but the ID is still set. Next cycle the Get returns NotFound which sets `MaestroBundleNotFound` (not `MaestroBundleStatusFeedbackNotAvailable`), so the stale check won't fire again but the create controller also won't recreate because the ID is non-empty. I need to handle this — either clear the ID first, or handle the NotFound+non-empty-ID case in the create controller.

> Is the stale detection and deletion responsibility really on the "read and persist" controllers?

It's here because this controller already has the Maestro client, the MCC from Cosmos, and access to the SPC. The create controller doesn't read MCC status so it can't detect staleness. A separate controller could work but felt like overkill for a temporary workaround.

> The delete is fire and forget. Can this race with the create controller?

The delete clears `MaestroAPIMaestroBundleID` and calls `Replace` before returning. The create controller checks for non-empty ID at line 144 — if Replace lands first, create sees empty ID and recreates. If create runs between Delete and Replace, the Replace gets a 412 and we retry next cycle. Not a tight race but the 412 case ties back to the resilience gap above.

> Why unset only the ID and not the whole bundle reference?

The bundle reference (`Name`, `MaestroAPIMaestroBundleName`) is set by the create controller when it first registers the bundle. Clearing just the ID tells the create controller "this bundle needs to be recreated" without losing the reference metadata. Clearing the whole reference would require the create controller to re-derive it.

> Can we have false positives?

Yes, and a time-based threshold doesn't actually guard against them. If a bundle is stuck because of some other issue (e.g. the agent is in a crash loop, the MC has connectivity problems, or there's a different bug holding up status feedback), it would look identical — `Manifests == []` for >5 minutes — and we'd delete a bundle that wasn't actually hit by the [AROSLSRE-833](https://redhat.atlassian.net/browse/AROSLSRE-833) race. Increasing the threshold doesn't help because those cases are also indefinite.

To avoid false positives we'd need a signal that distinguishes "event never published to broker" from "event published but agent hasn't reported back yet." We don't currently have that signal on the client side.

> If the resource on the other end is deleted (e.g. HostedCluster during deletion) this will make the readonly bundle be created over and over

Good catch. If the HC is gone, the agent can't feed back status, so `MaestroBundleStatusFeedbackNotAvailable` persists and we'd loop every 5 minutes. The mitigation is that `MaestroBundleStatusFeedbackNotAvailable` specifically means `len(Manifests) == 0` — the agent never saw the bundle at all. If the agent sees the bundle but the HC is gone, the agent reports back an error status (non-empty Manifests), which would set a different degraded reason. But I should verify this assumption.

> Could maestro temporarily stop reporting feedback and then resume?

The stale check only fires when `Manifests` has been empty since creation — it's not checking for a gap in an existing feedback stream. Once the agent reports back even once, `Manifests` is non-empty and the condition clears. Maestro losing connection temporarily would mean the existing status gets stale, but the Manifests array wouldn't go back to `[]`.

> Is modifying calculateManagementClusterContentFromMaestroBundle to return 3 params really appropriate?

The alternative was a separate Maestro Get call in SyncOnce, which was the original implementation and doubled the API calls. Threading the already-fetched bundle through the return avoids that. Open to suggestions if there's a cleaner way — could use an out-param struct instead of a third return value.

backend/pkg/controllers/read_and_persist_cluster_scoped_maestro_readonly_bundles_content_controller.go
avollmer-redhat: You're right — if Delete succeeds but Replace gets a 412 ETag conflict, the bundle is gone but the ID is still set. Next cycle Get returns NotFound → sets `MaestroBundleNotFound` (not `MaestroBundleStatusFeedbackNotAvailable`) so the stale check won't fire, but the create controller also skips because the ID is non-empty. Same stuck state.

Options: (a) clear the ID and Replace *before* the Delete so worst case we recreate a bundle that still exists (idempotent), or (b) teach the create controller to handle NotFound+non-empty-ID by clearing and retrying.

backend/pkg/controllers/maestro_readonly_bundle_helpers.go
avollmer-redhat, line 256: Agreed. `MaestroBundleStatusFeedbackNotAvailable` fires whenever `len(Manifests) == 0` — that covers the AROSLSRE-833 race but also agent crash loops, MC connectivity issues, or any other reason the agent never reports back. A time-based threshold doesn't help because those cases are also indefinite.

To avoid false positives we'd need a signal that distinguishes "event never published to broker" from "event published but agent hasn't reported back yet." We don't have that on the client side today.

miguelsorianod:
I understand Maestro fixed the issue and it was rolled out right @avollmer-redhat ?

In that case I think we can close the PR.

miguelsorianod:
Closing because we've removed the usage of maestro readonly bundles, replacing it by ReadDesires

PR #5260, opened 2026-05-13
allow zero UUID for cincinnati instead of delaying

deads2k:
UUID isn't strictly necessary and this will improve latency


raelga (approved):
/lgtm

deads2k:
/hold cancel

Removed stale tests

/label lgtm

PR #5268, opened 2026-05-14
feat: add more logging in frontend MigrateCosmosOrDie

miguelsorianod:
To indicate progress

### DIFF
--- a/frontend/pkg/frontend/migrate_cosmos.go
+++ b/frontend/pkg/frontend/migrate_cosmos.go
@@ -16,49 +16,72 @@ package frontend
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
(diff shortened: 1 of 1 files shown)


frontend/pkg/frontend/migrate_cosmos.go
```diff
@@ -16,167 +16,236 @@ package frontend
...
 	// This is a temporary change. Once deployed to production, we will remove this content and leave it empty
 	// for the next small migration we need to do.  Once datasets are large, we will start doing this inside of the backend.
 
 	subscriptionIterator, err := resourcesDBClient.Subscriptions().List(ctx, nil)
 	if err != nil {
+		logger.Error(err, "failed to list subscriptions")
 		panic(err)
 	}
+
+	logger.Info("migrating subscriptions")
```
deads2k, line 38: do we have log levels?

miguelsorianod: No. But I was thinking we could add a --log-level CLI attribute into frontend that takes a logr log level so we can start supporting it. I can take a follow-up action on that to add it. I think it could be nice.

janboll (commented):
/lgtm

PR #5272, opened 2026-05-14
Stop migrating operation records

deads2k:
All the ones we care about have been rekeyed. We never rewrite storage since old operations are removed after a week.  Currently have a large enough number that a 3ms latency per is causing a problem.  Remove the stanza.

Operations are already on generic storage, so no more migration expected.


deads2k:
This is blocking a region. Plan to button merge on green lint, verify, unit, integration testing.

PR #5276, opened 2026-05-14
kubeapplier: track the *Desire ManagementCluster string→*ResourceID 

deads2k:
Spec.ManagementCluster on ApplyDesire/DeleteDesire/ReadDesire is now an *azcorearm.ResourceID; regenerated zz_generated.deepcopy.go to match. The accessor on ManagementClusterAccessor now returns *ResourceID, and the database/informer/lister code that derives the Cosmos partition key now does so via lowercased(rid.String()) with nil-safe handling. Every test fixture that previously set ManagementCluster to a string literal now constructs an MC resourceID at package scope and stamps it into the spec; the kube-applier integration JSON fixtures and framework constant carry the full MC resourceID path so the JSON unmarshal into *ResourceID succeeds.


internal/api/kubeapplier/partition.go
```diff
@@ -14,13 +14,17 @@
...
 package kubeapplier
 
+import (
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+)
+
 // ManagementClusterAccessor is implemented by every *Desire so that the database
-// layer can read the partition-key value (the management cluster name) without
-// importing this package's concrete types.
+// layer can read the partition-key value (the management cluster resourceID)
```
stevekuznetsov, line 22: super-nit: the management cluster is an AKS cluster, so it has it's own Azure/ARM ResourceID, which I imagine is not what this will return me, unclear if others will have this confusion

internal/api/kubeapplier/types_apply_desire.go
```diff
@@ -46,8 +48,9 @@ type ApplyDesireSpec struct {
 	// reconcile this desire. It is the cosmos partition key for the
 	// kube-applier container; entries from one management cluster never see
 	// entries from another.
-	// TODO this may end up changing to be a resourceID
-	ManagementCluster string `json:"managementCluster"`
+	// This is the same resourceID as the fleetapi.ManagementCluster.CosmosMetadata.ResourceID
```
stevekuznetsov, line 51: Ah, there ya go

stevekuznetsov (approved):
/lgtm
/approve

PR #5287, opened 2026-05-15
Add Prometheus metrics instrumentation to Maestro GRPC client

garrickdabbs:
# Add Maestro GRPC Metrics for Backend Operations

Implements Maestro GRPC client observability to track throughput, errors, and latency as requested in [ARO-26754](https://redhat.atlassian.net/browse/ARO-26754).

### Metrics Implementation
- **3 new Prometheus metric families** tracking Maestro GRPC client operations:
  - `maestro_grpc_operations_total` - Total operations by type (create, get, delete, patch, list)
  - `maestro_grpc_errors_total` - Failed operations by type (excludes expected errors like NotFound/AlreadyExists)
  - `maestro_grpc_operation_duration_seconds` - Histogram of operation latency by type

- **Instrumentation wrapper** for Maestro client using decorator pattern (`backend/pkg/maestro/maestro_client_metrics.go`)
- **Expected error filtering** - NotFound on Get and AlreadyExists on Create are not counted as failures (part of normal GetOrCreateMaestroBundle pattern)
- **Metric pre-initialization** to ensure all operation labels appear in Prometheus from startup
- **Integration** into backend initialization (`backend/pkg/app/backend.go`)

### Grafana Dashboard
Added 3 new panels to `observability/grafana-dashboards/backend/backend.json`:
- **Maestro GRPC Operations Rate** - ops/sec by operation type
- **Maestro GRPC Error Rate** - error percentage with volume guard to prevent NaN/Inf
- **Maestro GRPC P95 Latency** - 95th percentile latency by operation

All panels follow existing dashboard patterns with `cluster` variable filtering.

image

### Alerts
**Alerts will be added in a follow-up PR** after 1-2 weeks of baseline data collection to establish appropriate thresholds, as requested by @mmazur.

## Files Changed
```
backend/pkg/maestro/maestro_client_metrics.go      | +170 (new file)
backend/pkg/maestro/maestro_client.go              | +25, -4
backend/pkg/app/backend.go                         | +2, -1
observability/grafana-dashboards/backend/backend.json | +158
```

## Testing & Validation
- ✅ Metrics validated in personal dev environment
- ✅ Prometheus scraping backend successfully
- ✅ Grafana dashboard panels displaying data
- ✅ All 3 metrics collecting operations with expected error filtering

## Related Issues
- Addresses [ARO-26754](https://redhat.atlassian.net/browse/ARO-26754) - Cluster provisioning latency diagnostics
- Related to [ARO-26043](https://redhat.atlassian.net/browse/ARO-26043) - MQTT metrics instrumentation

### DIFF
--- a/backend/pkg/maestro/maestro_client_metrics.go
+++ b/backend/pkg/maestro/maestro_client_metrics.go
@@ -44,7 +44,7 @@ func NewMaestroMetrics(r prometheus.Registerer) *MaestroMetrics {
 		errorsTotal: prometheus.NewCounterVec(
 			prometheus.CounterOpts{
 				Name: "maestro_grpc_errors_total",
-				Help: "Total number of failed Maestro GRPC operations",
+				Help: "Total number of unexpected Maestro GRPC operation failures (excludes expected errors like NotFound/AlreadyExists)",
 			},
 			[]string{"operation"},
 		),
--- a/backend/pkg/maestro/maestro_client_metrics.go
+++ b/backend/pkg/maestro/maestro_client_metrics.go
@@ -1,0 +1,170 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package maestro
+
+import (
+	"context"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	workv1 "open-cluster-management.io/api/work/v1"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type MaestroMetrics struct {
+	operationsTotal   *prometheus.CounterVec
+	errorsTotal       *prometheus.CounterVec
+	operationDuration *prometheus.HistogramVec
+}
+
+func NewMaestroMetrics(r prometheus.Registerer) *MaestroMetrics {
+	m := &MaestroMetrics{
+		operationsTotal: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "maestro_grpc_operations_total",
+				Help: "Total number of Maestro GRPC operations by operation type",
+			},
+			[]string{"operation"},
+		),
+		errorsTotal: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "maestro_grpc_errors_total",
+				Help: "Total number of failed Maestro GRPC operations",
+			},
+			[]string{"operation"},
+		),
+		operationDuration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name:    "maestro_grpc_operation_duration_seconds",
+				Help:    "Duration of Maestro GRPC operations in seconds",
+				Buckets: prometheus.DefBuckets,
+			},
+			[]string{"operation"},
+		),
+	}
+
+	r.MustRegister(
+		m.operationsTotal,
+		m.errorsTotal,
+		m.operationDuration,
+	)
+
+	// Pre-initialize all operation label values so metrics appear even with zero values
+	operations := []string{"create", "get", "delete", "patch", "list"}
+	for _, op := range operations {
+		m.operationsTotal.WithLabelValues(op)
+		m.errorsTotal.WithLabelValues(op)
+		m.operationDuration.WithLabelValues(op)
+	}
+
+	return m
+}
+
+type instrumentedMaestroClient struct {
+	client  Client
+	metrics *MaestroMetrics
+}
+
+func NewInstrumentedMaestroClient(client Client, metrics *MaestroMetrics) Client {
+	return &instrumentedMaestroClient{
+		client:  client,
+		metrics: metrics,
+	}
+}
+
+// isExpectedError returns true if the error is expected in normal operation
+// and should not be counted as a failure in error rate metrics.
+func isExpectedError(operation string, err error) bool {
+	// NotFound on Get is expected in GetOrCreateMaestroBundle pattern
+	// where Get is used to check existence before Create.
+	if operation == "get" && k8serrors.IsNotFound(err) {
+		return true
+	}
+
+	// AlreadyExists on Create is expected during concurrent creates
+	// in GetOrCreateMaestroBundle pattern (handled by retry with Get).
+	if operation == "create" && k8serrors.IsAlreadyExists(err) {
+		return true
+	}
+
+	return false
+}
+
+// observe records metrics for a completed operation.
+func (c *instrumentedMaestroClient) observe(operation string, start time.Time, err error) {
+	c.metrics.operationDuration.WithLabelValues(operation).Observe(time.Since(start).Seconds())
+
+	if err != nil && !isExpectedError(operation, err) {
+		c.metrics.errorsTotal.WithLabelValues(operation).Inc()
+	}
+}
+
+func (c *instrumentedMaestroClient) Create(ctx context.Context, manifestWork *workv1.ManifestWork, opts metav1.CreateOptions) (*workv1.ManifestWork, error) {
+	operation := "create"
+	start := time.Now()
+	c.metrics.operationsTotal.WithLabelValues(operation).Inc()
+
+	result, err := c.client.Create(ctx, manifestWork, opts)
+	c.observe(operation, start, err)
+
+	return result, err
+}
+
+func (c *instrumentedMaestroClient) Get(ctx context.Context, name string, opts metav1.GetOptions) (*workv1.ManifestWork, error) {
+	operation := "get"
+	start := time.Now()
+	c.metrics.operationsTotal.WithLabelValues(operation).Inc()
+
+	result, err := c.client.Get(ctx, name, opts)
+	c.observe(operation, start, err)
+
+	return result, err
+}
+
+func (c *instrumentedMaestroClient) Delete(ctx context.Context, name string, opts metav1.DeleteOptions) error {
+	operation := "delete"
+	start := time.Now()
+	c.metrics.operationsTotal.WithLabelValues(operation).Inc()
+
+	err := c.client.Delete(ctx, name, opts)
+	c.observe(operation, start, err)
+
+	return err
+}
+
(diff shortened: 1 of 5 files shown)


garrickdabbs:
/cc @geoberle @janboll @mbarnes @raelga 

  This PR adds Prometheus metrics instrumentation to the Maestro GRPC client (backend) and includes Grafana dashboard updates (observability). Ready for review!

garrickdabbs (commented):
Thanks for the detailed review! I've addressed the feedback in commit 50e145084:

## ✅ Fixed: Code Duplication (Comment on maestro_client_metrics.go:105-197)
Extracted the instrumentation pattern into an `observe()` helper function. This:
- Reduces the 5 duplicated methods to a single reusable helper
- Ensures consistent metrics recording across all operations
- Eliminates drift risk when adding new operations

## ✅ Fixed: Grafana Label Mismatch (Comment on backend.json:103)
Added `namespace="aro-hcp"` filter to all three Grafana dashboard panels to match the PrometheusRule alert queries. This prevents metric collision if multiple services emit the same metric name.

## ✅ Documented: Breaking API Change (Comment on maestro_client.go:91-112)
Added comprehensive godoc to `NewMaestroClientBuilder` explaining:
- The `metrics` parameter is optional (nil-safe by design)
- When nil, clients are not instrumented
- The signature change is intentional to enable optional instrumentation

The breaking change is acceptable because:
- Only 1 call site exists in the codebase (backend/pkg/app/backend.go:389), which is updated in this PR
- Other builders in the codebase follow this pattern (e.g., `NewFPAMIDataplaneClientBuilder`, `NewServiceManagedIdentityClientBuilder`)
- The nil-safety makes instrumentation truly optional

## ⏭️ No Change: MustRegister Pattern (Comment on maestro_client_metrics.go:36-80)
`r.MustRegister(...)` follows the existing codebase pattern:
- `operation_phase_metrics_controller.go:50`: `r.MustRegister(h.phaseInfo, ...)`
- `resource_metrics_controller.go:57`: `r.MustRegister(h.provisionState, ...)`

The panic-on-double-registration behavior is appropriate here:
- `NewMaestroMetrics` is called once during backend initialization
- Multiple registrations indicate a bug (double initialization) 
- Panicking early during startup is the established pattern for this type of misconfiguration

Keeping consistency with existing metrics code.

garrickdabbs:
## Copilot Security Review - Response

I've reviewed all 8 unresolved Copilot comments and addressed them in commit 03e8cb0. Here's the summary:

### ✅ Fixed (3 legitimate issues)

1. **Misleading metric help text** (maestro_client_metrics.go:48)
   - Removed "(dropped events)" - metric tracks all errors, not just dropped events
   
2. **Inaccurate function comment** (maestro_client_metrics.go:107)
   - Fixed `observe()` comment that claimed "defer-friendly" when it's not called via defer
   
3. **Grafana dashboard NaN/Inf** (backend.json:156)
   - Added `and operations > 0` guard to prevent division by zero

### 🚫 Dismissed (5 false positives)

4. **Breaking API change** (maestro_client.go:93, 120) - 2 duplicate comments
   - Only 1 caller exists in entire codebase, already updated in this PR
   - Internal package, not public API
   
5. **MustRegister panic risk** (maestro_client_metrics.go:80) - 2 duplicate comments
   - Standard pattern used in 6 other files in ARO-HCP
   - Metrics created once per backend instance
   - Registerer never nil (required field)

---

**Changes:** 2 files, 4 lines (4 insertions, 4 deletions)
- Fixed help text accuracy
- Improved dashboard robustness at zero traffic
- No functional changes

All legitimate security/quality concerns have been addressed.

backend/pkg/maestro/maestro_client_metrics.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		c.metrics.errorsTotal.WithLabelValues(operation).Inc()
+	} else {
+		switch eventType {
+		case "published":
+			c.metrics.eventsPublished.Inc()
+		case "subscribed":
+			c.metrics.eventsSubscribed.Inc()
+		}
+	}
```
swiencki, line 114: `observe` counts every non-nil error in `maestro_grpc_errors_total`. `GetOrCreateMaestroBundle` does a `Get` first; `NotFound` is the normal create path. During first-time bundle creation, `get` has 100% "error" rate, which trips `BackendMaestroGRPCClientErrorRate`. Either classify expected errors before counting (e.g., skip `NotFound` / `AlreadyExists`), or add a `status` label and exclude expected codes in the alert expression.

swiencki (commented):
These metrics are in-process counters/histograms that reset on pod restart. In #4393, @stevekuznetsov pushed for level-driven metrics (a la kube-state-metrics) specifically because in-memory state is lost on crash/eviction/OOM, which is when you need the signal most.

The RPC-layer observability is useful and complementary to #4393's operation-phase gauges, but the instrumentation pattern should match the established codebase convention. Worth considering whether these can be expressed as level-driven state (e.g., current operation status with last-RPC-result labels derived from the informer cache) rather than accumulated counters.

Also: the `BackendControllerPanic` summary annotation changed to interpolate `{{ $labels.controller }}` but the promtool test fixture wasn't updated. `promtool test rules` will fail on the annotation mismatch.

garrickdabbs:
(Edited for clarity)

@swiencki - Fixed the promtool test. Before I tackle the expected errors issue or redesign the metrics, I need to understand the right pattern here.

**ARO-26754's goal**: Diagnose why Maestro GRPC delays cause >20min cluster provisioning times ([ARO-26043](https://redhat.atlassian.net/browse/ARO-26043)). We need: "Is Maestro responding? How fast? What's failing?"

**Your #4393 pattern**: Level-driven gauges from Cosmos state tracking resource lifecycle phases.

These feel like different observability layers:
- **RPC health** (is infrastructure working): Rate/Errors/Duration counters at the call layer
- **Resource state** (where are my clusters): Gauges from the informer cache

**Should Maestro GRPC metrics be:**

**A) Standard RPC counters** (operations_total, errors_total, duration_seconds)
- Complements existing operation-phase gauges from #4393
- Trade-off: lose RPC stats on crash, but resource state survives in Cosmos

**B) Level-driven somehow**
- Not sure how to track "P95 latency >1s" or "5% error rate" without counters
- What would that look like?

I get that counters reset on crash/OOM. But we'd still have operation-phase metrics showing stuck clusters, pod events, and logs. Is the concern about losing historical RPC data, or is there a level-driven pattern for infrastructure monitoring I'm missing?

Happy to redesign—just want to understand what pattern you're recommending.

swiencki:
Apologies, my earlier suggestion conflated two layers. After re-reading [ARO-26754](https://redhat.atlassian.net/browse/ARO-26754) my earlier comment was off-base. The level-driven pattern applies to cluster lifecycle state persisted in Cosmos, not to RPC health. In-process counters are the correct pattern for gRPC error rates and latency.

backend/alerts/backend-prometheusRule.yaml
```diff
@@ -63,4 +63,60 @@ spec:
       annotations:
         description: 'Backend controller {{ $labels.controller }} has panicked {{ printf "%.0f" $value }} time(s) in the last 5 minutes.'
         runbook_url: 'https://eng.ms/docs/cloud-ai-platform/azure-core/azure-cloud-native-and-management-platform/control-plane-bburns/azure-red-hat-openshift/azure-redhat-openshift-team-doc/hcp/troubleshooting/backend-tsg.html'
-        summary: 'Backend controller is panicking'
```
mmazur, line 40: Please split out the alerts into a separate PR and let's merge the dashboard+metrics emission asap. I don't want to add alerts blind, I'd like at least a week or two of data to look at first.

mmazur (changes requested):

garrickdabbs:
@swiencki - Perfect, thanks for clarifying! 

All three concerns addressed in commits:
1. ✅ **Promtool test** (`c021f3ad5`) - Fixed BackendControllerPanic annotation mismatch
2. ✅ **Expected errors** (`904b244cb`) - Added context-aware error classification:
   - `NotFound` on Get: expected (GetOrCreateMaestroBundle checks existence first)
   - `AlreadyExists` on Create: expected (concurrent create race, handled by retry)
   - Real errors (network, auth, etc.) still counted

Ready for re-review!

garrickdabbs:
@mmazur - Done! Removed the Maestro GRPC alerts in commit `6fc699ac9`.

This PR now contains only:
- ✅ Metrics emission (maestro_grpc_operations_total, maestro_grpc_errors_total, maestro_grpc_operation_duration_seconds, etc.)
- ✅ Grafana dashboard panels (Operations Rate, Error Rate, P95 Latency)
- ✅ BackendControllerPanic summary fix (unrelated promtool test fix)

I'll create a follow-up PR for the alerts after we have 1-2 weeks of production data to establish proper thresholds.

backend/alerts/backend-prometheusRule.yaml
```diff
@@ -63,4 +63,4 @@ spec:
...
+        summary: 'Backend controller {{ $labels.controller }} is panicking'
```
mmazur:
```suggestion
        summary: 'Backend controller is panicking'
```

backend/alerts/backend-prometheusRule_test.yaml
```diff
@@ -71,7 +71,7 @@ tests:
       exp_annotations:
         description: 'Backend controller DoNothingExample has panicked 5 time(s) in the last 5 minutes.'
         runbook_url: 'https://eng.ms/docs/cloud-ai-platform/azure-core/azure-cloud-native-and-management-platform/control-plane-bburns/azure-red-hat-openshift/azure-redhat-openshift-team-doc/hcp/troubleshooting/backend-tsg.html'
-        summary: 'Backend controller is panicking'
+        summary: 'Backend controller DoNothingExample is panicking'
```
mmazur:
```suggestion
        summary: 'Backend controller is panicking'
```

mmazur (changes requested):

backend/pkg/maestro/maestro_client_metrics.go
```diff
@@ -0,0 +1,199 @@
...
+	return result, err
+}
+
+func (c *instrumentedMaestroClient) List(ctx context.Context, opts metav1.ListOptions) (*workv1.ManifestWorkList, error) {
+	operation := "list"
+	start := time.Now()
+	c.metrics.operationsTotal.WithLabelValues(operation).Inc()
+
+	result, err := c.client.List(ctx, opts)
+	c.observe(operation, start, err, eventTypeSubscribed)
```
swiencki: `maestro_grpc_events_subscribed_total` increments +1 per successful `List()` or `Get()` call regardless of result size, so with `ForEachMaestroBundle` paginating at Limit 400 it undercounts actual items received by up to 400x; consider either incrementing by `len(result.Items)` or dropping it since `operations_total` already covers call volume.

garrickdabbs: Ah , I see. Dropping the redundant counter.  Thank you!

mmazur:
@garrickdabbs you added some large file by accident?

garrickdabbs:
> @garrickdabbs you added some large file by accident?

@mmazur Yes I did, an artifact of local testing. Thank you for calling that out, removing it now.

mmazur:
That file is still there and being added to the repo history. Please squash all the 12 commits into a single commit, then push --force.

garrickdabbs:
@swiencki @mmazur , I squash merged, re-based, and resolved some new comments from copilot. Please review when you get a moment. Thanks!

PR #5290, opened 2026-05-15
feat: track frontend panics during http request handling via prometheus metric

miguelsorianod:
Add a Prometheus counter (frontend_http_request_panics_total) to the panic recovery middleware. The counter is incremented each time a panic is recovered during HTTP request handling. The metric has two labels:
- route: the matched route pattern with the method prefix stripped
- method: the HTTP method

The labels are expected to never be empty.

With this design, the total number of panics across all routes and methods can be obtained using sum(frontend_http_request_panics_total) when querying prometheus.

Note: since MiddlewarePanic is registered twice in the middleware chain, a single request could increment the counter twice if the inner panic middleware recovers and then when continuing the execution another panic occurs which is then caught by the outer panic middleware

### DIFF
--- a/admin/server/middleware/middleware.go
+++ b/admin/server/middleware/middleware.go
@@ -104,8 +104,12 @@ func (mux *MiddlewareMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	r = r.WithContext(ContextWithPattern(r.Context(), patt))
 
 	mainHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Capture the matched pattern when ServeMux returns or when a matched
+		// handler panics, so pre-mux middleware can read it from context.
+		defer func() {
+			*patt = r.Pattern
+		}()
 		mux.ServeMux.ServeHTTP(w, r)
-		*patt = r.Pattern
 	})
 
 	mux.middleware.Handler(mainHandler).ServeHTTP(w, r)
--- a/frontend/pkg/frontend/middleware_panic.go
+++ b/frontend/pkg/frontend/middleware_panic.go
@@ -20,18 +20,68 @@ import (
 	"runtime/debug"
 	"testing"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"k8s.io/component-base/metrics/legacyregistry"
+
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
+var (
+	// frontendHTTPRequestPanicsTotalCounterVec counts the total number of panics that occurred
+	// during request handling.
+	// To get the total panics use sum() in PromQL for the aggregate.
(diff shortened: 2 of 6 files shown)


miguelsorianod:
a follow-up will be defining an alert based on this

frontend/pkg/frontend/middleware_panic.go
```diff
@@ -18,12 +18,31 @@ import (
...
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
+var (
+	// httpRequestsPanicTotalCounter counts the total number of panics that occurred
+	// during request handling.
+	// To get the total panics use sum() in PromQL for the aggregate.
+	// The label values must never be empty.
+	httpRequestPanicsTotalCounter = promauto.With(legacyregistry.Registerer()).NewCounterVec(
```
miguelsorianod: There's also a internal/utils.PanicTotal used in `frontend/main.go` but that one seems to be focused around controllers (because it has a `controller` label), and it's used only set by the K8s runtime panic handlers (which are executed on controllers that use k8s utilruntime HandleCrash). On the http handling path panics that occur in the individual handler are captured and recovered by the middleware panic handler

admin/server/middleware/middleware.go
```diff
@@ -104,8 +104,12 @@ func (mux *MiddlewareMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	r = r.WithContext(ContextWithPattern(r.Context(), patt))
 
 	mainHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Capture the matched pattern when ServeMux returns or when a matched
+		// handler panics, so pre-mux middleware can read it from context.
+		defer func() {
```
miguelsorianod, line 109: I also added a new commit that sets *patt through a defer instead of after the call to `mux.ServeMux.ServeHTTP(w, r)` only. This covers both the scenario when `mux.ServeMux.ServeHTTP(w, r)` succeeds and also the one where it panics, which wasn't being covered.

frontend/pkg/frontend/middleware_panic.go
```diff
@@ -20,18 +20,68 @@ import (
...
 func MiddlewarePanic(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
 	// Do not catch panics when running "go test".
-	if !testing.Testing() {
+	middlewarePanicRecover(w, r, next, !testing.Testing())
+}
+
+// middlewarePanicRecover runs the next http.HandlerFunc. Additionally, when
+// recoverPanics is true it recovers panics, increments the
+// frontend_http_request_panics_total metric, and writes an HTTP 500 response.
+func middlewarePanicRecover(w http.ResponseWriter, r *http.Request, next http.HandlerFunc, recoverPanics bool) {
```
miguelsorianod, line 54: Refactored to make it unit testable

deads2k:
/lgtm
next step, ask Gerd how to wire up an alert and then do that.

PR #5295, opened 2026-05-16
hcpctl: snapshot: improve queries

stevekuznetsov:
Fixed some cardinality issues with fetching internal IDs and dependent state, removed the backend operation status log query since that seems to no longer exist, short-circuited query fetching on 4xx responses.


stevekuznetsov:
/label lgtm

Going to pre-tag this stuff as it's fixing up otherwise best-effort log gathering and test render code. If we land this and the jUnit looks good, we may be ready to actually make snapshot gathering required in e2e.

PR #5296, opened 2026-05-17
backend: expose node pool ID in CS State Dump

stevekuznetsov:
Previously, it was not possible to query for *only* the CS dumps for one node pool.


stevekuznetsov:
/label lgtm

Informative PR, zero downside.

PR #5310, opened 2026-05-18
Test framework API versioning changes

miquelsi:
https://redhat.atlassian.net/browse/ARO-27211

### What

- Added API version abstraction in framework
- The default API version is the oldest one (20240610). It can be configured using `ARO_HCP_E2E_API_VERSION` environment variable, or specifically for one test with `tc.SetHCPAPIVersionForTest(framework.HCPAPIVersion*)`
- Refactored cluster and node pool creation routing
- Added facade for cluster and node pool clients
- Added version-agnostic helpers for admin credential/revoke  helpers, nodepool disk-type validations and KeyVault visibility checks

### Why

- Decouple tests from SDK version churn: tests no longer need to hardcode which API client/helper to use for common operations.
- Centralize version selection: one place (env or test context) controls API version behavior.
- Safer migration path: wrappers preserve compatibility while enabling gradual cleanup.
- Cleaner test code: fewer low-level SDK calls in tests; framework owns version routing.
- Future-ready: adding a new API version mostly becomes framework wiring, not mass test rewrites.

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/test/e2e/cluster_create_complex_cilium_kv.go
+++ b/test/e2e/cluster_create_complex_cilium_kv.go
@@ -16,7 +16,6 @@ package e2e
 
 import (
 	"context"
-	"errors"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -25,12 +24,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
-	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5"
-
 	operatorclient "github.com/openshift/client-go/operator/clientset/versioned"
 
-	hcpsdk20251223preview "github.com/Azure/ARO-HCP/test/sdk/v20251223preview/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
 	"github.com/Azure/ARO-HCP/test/util/framework"
 	"github.com/Azure/ARO-HCP/test/util/labels"
 	"github.com/Azure/ARO-HCP/test/util/verifiers"
@@ -50,6 +45,7 @@ var _ = Describe("Customer", func() {
 			)
 
 			tc := framework.NewTestContext()
+			tc.SetHCPAPIVersionForTest(framework.HCPAPIVersion20251223)
 
 			if tc.UsePooledIdentities() {
 				err := tc.AssignIdentityContainers(ctx, 1, 60*time.Second)
@@ -81,25 +77,12 @@ var _ = Describe("Customer", func() {
 			)
 			Expect(err).NotTo(HaveOccurred())
 
-			By("creating the HCP cluster with no CNI and private etcd via v20251223preview")
-			clusterResource, err := framework.BuildHCPCluster20251223FromParams(clusterParams, tc.Location(), nil)
-			Expect(err).NotTo(HaveOccurred())
-
-			// Set KeyVault visibility to Private
-			if clusterResource.Properties != nil && clusterResource.Properties.Etcd != nil &&
-				clusterResource.Properties.Etcd.DataEncryption != nil &&
-				clusterResource.Properties.Etcd.DataEncryption.CustomerManaged != nil &&
-				clusterResource.Properties.Etcd.DataEncryption.CustomerManaged.Kms != nil {
-				clusterResource.Properties.Etcd.DataEncryption.CustomerManaged.Kms.Visibility = to.Ptr(hcpsdk20251223preview.KeyVaultVisibilityPrivate)
-			}
-
-			_, err = framework.CreateHCPCluster20251223AndWait(
+			By("creating the HCP cluster with no CNI and private etcd")
+			err = tc.CreateHCPClusterFromParam(
 				ctx,
 				GinkgoLogr,
-				tc.Get20251223ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient(),
 				*resourceGroup.Name,
-				customerClusterName,
-				clusterResource,
+				clusterParams,
 				45*time.Minute,
 			)
 			Expect(err).NotTo(HaveOccurred())
@@ -107,7 +90,6 @@ var _ = Describe("Customer", func() {
 			By("getting admin credentials for the cluster")
 			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster(
 				ctx,
-				tc.Get20240610ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient(),
 				*resourceGroup.Name,
 				customerClusterName,
 				10*time.Minute,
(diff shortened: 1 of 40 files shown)


test/util/framework/per_test_framework.go
```diff
@@ -1090,6 +1125,27 @@ func (tc *perItOrDescribeTestContext) TenantID() string {
 	return tc.perBinaryInvocationTestContext.tenantID
 }
 
+func (tc *perItOrDescribeTestContext) SetHCPAPIVersionForTest(version HCPAPIVersion) {
+	tc.contextLock.Lock()
+	defer tc.contextLock.Unlock()
+	tc.hcpAPIVersionOverride = &version
+}
```
mbukatov, line 1132: Could we log what version was selected? That will allow us to filter test results using particular API version if needed.

test/util/framework/per_test_framework.go
```diff
@@ -1090,6 +1125,27 @@ func (tc *perItOrDescribeTestContext) TenantID() string {
 	return tc.perBinaryInvocationTestContext.tenantID
 }
 
+func (tc *perItOrDescribeTestContext) SetHCPAPIVersionForTest(version HCPAPIVersion) {
+	tc.contextLock.Lock()
+	defer tc.contextLock.Unlock()
```
mbukatov, line 1130: I have one concern with this approach: it will make a test which deploys multiple hosted clusters using different APIs impossible. That said, I'm not sure how bit a problem this is.

test/e2e/cluster_create_complex_cilium_kv.go
```diff
@@ -50,6 +45,7 @@ var _ = Describe("Customer", func() {
 			)
 
 			tc := framework.NewTestContext()
+			tc.SetHCPAPIVersionForTest(framework.HCPAPIVersion20251223)
```
mbukatov, line 48: So basically a test case will use default API version, unless a different API version is selected using `tc.SetHCPAPIVersionForTest()`, right?

Personally I would prefer an explicit selection (so that every test would need make the decision explicitly), but this will work as well.

test/e2e/cluster_create_complex_cilium_kv.go
```diff
@@ -81,33 +77,19 @@ var _ = Describe("Customer", func() {
...
-			clusterResource, err := framework.BuildHCPCluster20251223FromParams(clusterParams, tc.Location(), nil)
-			Expect(err).NotTo(HaveOccurred())
-
-			// Set KeyVault visibility to Private
-			if clusterResource.Properties != nil && clusterResource.Properties.Etcd != nil &&
-				clusterResource.Properties.Etcd.DataEncryption != nil &&
-				clusterResource.Properties.Etcd.DataEncryption.CustomerManaged != nil &&
-				clusterResource.Properties.Etcd.DataEncryption.CustomerManaged.Kms != nil {
-				clusterResource.Properties.Etcd.DataEncryption.CustomerManaged.Kms.Visibility = to.Ptr(hcpsdk20251223preview.KeyVaultVisibilityPrivate)
-			}
```
mbukatov, line 94: It seems to me that these lines to reconfigure KeyVault visibility should be preserved.

test/e2e/cluster_create_complex_cilium_kv.go
```diff
@@ -156,55 +138,24 @@ var _ = Describe("Customer", func() {
...
-					Platform: &hcpsdk20251223preview.NodePoolPlatformProfile{
-						VMSize: to.Ptr(nodePoolParams.VMSize),
-						OSDisk: &hcpsdk20251223preview.OsDiskProfile{
-							SizeGiB:                to.Ptr(nodePoolParams.OSDiskSizeGiB),
-							DiskStorageAccountType: to.Ptr(hcpsdk20251223preview.DiskStorageAccountType(nodePoolParams.DiskStorageAccountType)),
-						},
-					},
-					AutoRepair: to.Ptr(true),
-				},
-			}
```
mbukatov, line 178: Here I assume these are all defaults we don't need specify when using NewDefaultNodePoolParams(). But will recheck.

test/e2e/cluster_create_complex_cilium_kv.go
```diff
@@ -156,55 +138,24 @@ var _ = Describe("Customer", func() {
...
 				*resourceGroup.Name,
+				clusterParams.ManagedResourceGroupName,
 				customerClusterName,
-				customerNodePoolName,
-				nodePool,
+				nodePoolParams,
 				45*time.Minute,
 			)
-			// We delay checking the error on purpose to get more details
-			// about the issue by running the verifiers.
```
mbukatov, line 190: I would keep this comment, explaining that the omission of immediate nodePoolErr checking is done on purpose.

test/e2e/cluster_create_complex_cilium_kv.go
```diff
@@ -156,55 +138,24 @@ var _ = Describe("Customer", func() {
...
-						clusterParams.ManagedResourceGroupName,
-						tc.LogDirPath)
-				}
-			}
+			Expect(nodePoolErr).NotTo(HaveOccurred())
 
 			By("verifying nodes become Ready with Cilium CNI")
 			err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyNodesReady(), verifiers.VerifyCiliumOperational("kube-system", "k8s-app=cilium"))
-			Expect(errors.Join(err, nodePoolErr, consoleLogErr)).NotTo(HaveOccurred())
+			Expect(err).NotTo(HaveOccurred())
```
mbukatov, line 158: This should be: `Expect(errors.Join(err, nodePoolErr)).NotTo(HaveOccurred())`

test/e2e/cluster_create_private_kv.go
```diff
@@ -79,24 +77,11 @@ var _ = Describe("Create HCPOpenShiftCluster with Private KeyVault", func() {
...
-			clusterResource, err := framework.BuildHCPCluster20251223FromParams(clusterParams, tc.Location(), nil)
-			Expect(err).NotTo(HaveOccurred())
-
-			// Set KeyVault visibility
-			if clusterResource.Properties != nil && clusterResource.Properties.Etcd != nil &&
-				clusterResource.Properties.Etcd.DataEncryption != nil &&
-				clusterResource.Properties.Etcd.DataEncryption.CustomerManaged != nil &&
-				clusterResource.Properties.Etcd.DataEncryption.CustomerManaged.Kms != nil {
-				clusterResource.Properties.Etcd.DataEncryption.CustomerManaged.Kms.Visibility = to.Ptr(hcpsdk20251223preview.KeyVaultVisibilityPrivate)
-			}
```
mbukatov, line 91: Here again, don't we need to make sure that KV visibility is set to private?

test/e2e/cluster_create_private_kv.go
```diff
@@ -108,33 +93,22 @@ var _ = Describe("Create HCPOpenShiftCluster with Private KeyVault", func() {
...
-				customerClusterName,
-				nil,
-			)
+			visibility, err := tc.GetClusterKeyVaultVisibility(ctx, *resourceGroup.Name, customerClusterName)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(cluster.Properties).ToNot(BeNil())
-			Expect(cluster.Properties.Etcd).ToNot(BeNil())
-			Expect(cluster.Properties.Etcd.DataEncryption).ToNot(BeNil())
-			Expect(cluster.Properties.Etcd.DataEncryption.CustomerManaged).ToNot(BeNil())
-			Expect(cluster.Properties.Etcd.DataEncryption.CustomerManaged.Kms).ToNot(BeNil())
```
mbukatov, line 123: Don't we want to keep this checks?

test/util/framework/deployment_helper.go
```diff
@@ -328,79 +349,110 @@ func (tc *perItOrDescribeTestContext) CreateHCPClusterFromParam(
...
-		ctx, cancel = context.WithTimeoutCause(ctx, timeout, fmt.Errorf("timeout '%f' minutes exceeded during CreateHCPClusterFromParam for cluster %s in resource group %s", timeout.Minutes(), parameters.ClusterName, resourceGroupName))
-		defer cancel()
-	}
-	clusterName := parameters.ClusterName
-
-	startTime := time.Now()
-	defer func() {
-		finishTime := time.Now()
-		tc.RecordTestStep(fmt.Sprintf("Deploy HCP cluster %s/%s", resourceGroupName, clusterName), startTime, finishTime)
-	}()
```
mbukatov, line 342: Did we drop these test steps on purpose?

test/util/framework/hcp_client_facade.go
```diff
@@ -0,0 +1,467 @@
...
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+
+	hcpsdk20240610preview "github.com/Azure/ARO-HCP/test/sdk/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
```
mbukatov, line 26: maybe we can use `/test/sdk/v20240610preview/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp` here directly

mbukatov (commented):
Thank you for looking into this quickly, this is an initial quick review.

Personally, I would rather avoid a per test global state of selected API version. But this approach would work as well.

Also, I need to think about the suggested interface over the API versions. I would prefer more explicit approach with version visible in name of a function or it's parameter in each API call. That said, I don't have a strong opinion about it right now, and the proposed approach will work as well.

mbarnes (commented):
The override mechanism that you've defined (`SetHCPAPIVersionForTest`) is fine for corner cases but I think a label-based approach would be more sustainable, so tests can declare the _minimum_ API version they need instead of locking themselves to one version with the override.  The override defeats the purpose of the abstraction.

Thankfully the way Azure API versions are named, they sort lexicographically over time.  So if you introduced a label for each API version in the pattern of, say, `MinimumVersionYYYYMMDD[preview]`, then you could derive which API versions a test is compatible with by sorting those labels and using whichever one the test declares as the starting point.

Just a suggestion.

miquelsi:
Closing as we won't want to consider API-agnostic tests.

PR #5311, opened 2026-05-18
Create cross-management cluster container informers for *Desire resources

deads2k:
Add union informers for our types so we can have controllers fire on changes read desire content.


deads2k:
simple rebase
/label lgtm

PR #5312, opened 2026-05-18
fix(e2e): retry subscription lookup with exponential backoff (AROSLSRE-884)

raelga:
[AROSLSRE-884](https://redhat.atlassian.net/browse/AROSLSRE-884)

### What

Add exponential backoff retry to `GetSubscriptionID` in the E2E test framework. Retries up to 4 times (15s, 30s, 60s, 120s) on both API errors and subscription-not-found, with structured logging on each attempt.

### Why

Starting ~15:00 UTC on 2026-05-18, all Prow E2E tests that create resource groups are panicking with `subscription with name '...' not found`. This affects both dev (PR e2e-parallel) and INT (branch e2e-integration). The root cause is external (Azure credential/RBAC change), but `GetSubscriptionID` fails immediately on a single attempt, making the framework fragile to transient subscription visibility issues.

### Testing

No new tests. The retry uses the same `wait.ExponentialBackoffWithContext` pattern established in [`hcp_helper.go:262`](https://github.com/Azure/ARO-HCP/blob/main/test/util/framework/hcp_helper.go#L262) for `UpdateHCPCluster`. Verified with `go build` and `make lint`.

This is a mitigation, not a fix for the underlying credential/RBAC issue. The Prow service principal needs to be checked separately.

### PR Checklist
- [x] Self-reviewed the diff

### DIFF
--- a/test/util/framework/resourcegroup_helper.go
+++ b/test/util/framework/resourcegroup_helper.go
@@ -24,6 +24,7 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/onsi/ginkgo/v2"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/ptr"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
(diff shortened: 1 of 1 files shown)


test/util/framework/resourcegroup_helper.go
```diff
@@ -34,20 +35,52 @@ import (
...
 		}
+
+		ginkgo.GinkgoLogr.Info("Subscription not found in list, will retry",
+			"subscription", subscriptionName,
+			"attempt", attempt)
+		return false, nil
+	})
+
+	if err != nil {
+		return "", fmt.Errorf("subscription with name '%s' not found after %d attempts: %w", subscriptionName, attempt, err)
```
gmfrasca, line 86: nit: the `err` here will always be a generic timeout from `wait.ExponentialBackoffWithContext` regardless of underlying cause.  

Mentioning this because the wording of this catchall error suggests we simply couldn't find the subscription, and could be slightly misleading if real API errors are causing the lookup failure (albeit, they _are_ logged earlier by GinkgoLogr, but aren't bubbled up themselves as part of the final error).

raelga: Good catch. Fixed in bb6081b4c. Now captures the last API error and includes it in the timeout message when available, so the final error distinguishes between "subscription not found in list" vs "API errors prevented the lookup".

tuxerrante (commented):
Nice work — clean, well-scoped mitigation for a blocker. The retry pattern follows the established [`wait.ExponentialBackoffWithContext`](https://github.com/Azure/ARO-HCP/blob/f2a67a4fc836fa41b2bb9c9f141d3e1fb35f1d75/test/util/framework/hcp_helper.go#L262) convention from [`hcp_helper.go`](https://github.com/Azure/ARO-HCP/blob/f2a67a4fc836fa41b2bb9c9f141d3e1fb35f1d75/test/util/framework/hcp_helper.go), and the PR description is thorough.

One optional nit:

[`resourcegroup_helper.go:84`](https://github.com/Azure/ARO-HCP/blob/f2a67a4fc836fa41b2bb9c9f141d3e1fb35f1d75/test/util/framework/resourcegroup_helper.go#L84) — `lastErr` is formatted with `%v`, which drops the error chain. Since the project uses Go 1.20+, you could use a second `%w` to preserve it:

```go
return "", fmt.Errorf("subscription lookup for '%s' failed after %d attempts (last API error: %w): %w", subscriptionName, attempt, lastErr, err)
```

Low severity since callers panic on error today, but it's a cheap improvement if you're doing another push.

stevekuznetsov:
/hold

This was an ARM outage - we have no evidence that we need any sort of retry or backoff for this, and it would not have helped.

raelga:
/close 

As per @stevekuznetsov comment.

PR #5324, opened 2026-05-19
feat: add count methods to database crud types

miguelsorianod:
Useful to count number of occurrences of a given cosmos
resource type without receiving the resource contents at
both Cosmos DB and Go levels.


miguelsorianod:
After discussion around this, we decided we are not going to add this. Even though it would be more efficient, we would need to maintain additional code and it was decided we prefer to keep it simple at the expense of efficiency.

Closing.

PR #3795, opened 2026-01-15
Update hook weight clustersizingconfiguration

pepedocs:
### What

This PR updates the hook weight to delay helm updates to clustersizingconfiguration cr AND use server side apply to fix ownership conflicts. Both the hypershift operator and helm manages this cr and thus they cause conflicts - This is consistent in my personal dev env and that it breaks creation of infrastructure. The following is an example log of this error;

```
[12:37:46.206] INFO: Deployment complete. {
  "resourceGroup": "management",
  "routine": "consumer-7",
  "serviceGroup": "Microsoft.Azure.ARO.HCP.RP.HypershiftOperator",
  "step": "deploy"
}
[12:37:46.206] ERROR: Step errored. {
  "err": "error running Helm release deployment Step, failed to deploy helm release: failed post-install: warning: Hook post-install aro-hcp-hypershift-operator/templates/cluster.clustersizingconfiguration.yaml failed: conflict occurred while applying object /cluster scheduling.hypershift.openshift.io/v1alpha1, Kind=ClusterSizingConfiguration: Apply failed with 5 conflicts: conflicts with \"hypershift-operator-manager\" using scheduling.hypershift.openshift.io/v1alpha1:\n- .spec.sizes[name=\"large\"].criteria.from\n- .spec.sizes[name=\"medium\"].criteria.from\n- .spec.sizes[name=\"medium\"].criteria.to\n- .spec.sizes[name=\"small\"].criteria.to\n- .spec.transitionDelay.decrease",
  "resourceGroup": "management",
  "routine": "consumer-7",
  "serviceGroup": "Microsoft.Azure.ARO.HCP.RP.HypershiftOperator",
  "step": "deploy"
}
[12:37:46.206] DEBUG: Consumer thread shutting down. {
  "routine": "consumer-7"
}
[12:37:46.206] DEBUG: Waiting for execution to finish. {}
[12:37:46.206] DEBUG: Execution finished. {}
[12:37:46.206] ERROR: command failed {
[...]

PR #5328, opened 2026-05-19
Fix: Race condition resource ClusterSizingConfiguration//cluster still exists

patriksuba:
[ARO-27214](https://redhat.atlassian.net/browse/ARO-27214)

### What

This PR fixes ownership conflicts in the clustersizingconfiguration custom resource for HyperShift by making two changes:
1. Increases the Helm hook weight from 2 to 10 for the clustersizingconfiguration resource
2. Enables server-side apply by adding the ArgoCD annotation "argocd.argoproj.io/sync-options": 
  "ServerSideApply=true"

### Why

Problem: There were ownership conflicts when multiple controllers or processes tried to manage the clustersizingconfiguration resource simultaneously during Helm deployments.

Solution:
- Higher hook weight (10) - Delays the Helm hook execution, ensuring it runs after other resources are created/updated, reducing race conditions
- Server-side apply - Kubernetes server-side apply allows multiple controllers to manage different fields of the same resource without ownership conflicts, as opposed to client-side apply which requires a single owner
  
This ensures the clustersizingconfiguration resource can be properly managed by both Helm and other controllers (like ArgoCD or the HyperShift operator) without conflicts.

### Testing

In progress

Base PR #3795 

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/hypershiftoperator/deploy/templates/cluster.clustersizingconfiguration.yaml
+++ b/hypershiftoperator/deploy/templates/cluster.clustersizingconfiguration.yaml
@@ -6,7 +6,8 @@ metadata:
   name: cluster
   annotations:
     "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-weight": "2"
+    "helm.sh/hook-weight": "10"
+    "argocd.argoproj.io/sync-options": "ServerSideApply=true"
 spec:
   concurrency:
     limit: 9999
@@ -54,7 +55,8 @@ metadata:
   name: cluster
   annotations:
     "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-weight": "2"
+    "helm.sh/hook-weight": "10"
+    "argocd.argoproj.io/sync-options": "ServerSideApply=true"
 spec:
   concurrency:
     limit: 5
--- a/hypershiftoperator/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_hypershift.yaml
+++ b/hypershiftoperator/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_hypershift.yaml
@@ -738,7 +738,8 @@ metadata:
   name: cluster
   annotations:
     "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-weight": "2"
+    "helm.sh/hook-weight": "10"
+    "argocd.argoproj.io/sync-options": "ServerSideApply=true"
 spec:
   concurrency:
     limit: 9999


hypershiftoperator/deploy/templates/cluster.clustersizingconfiguration.yaml
```diff
@@ -6,7 +6,8 @@ metadata:
   name: cluster
   annotations:
     "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-weight": "2"
+    "helm.sh/hook-weight": "10"
+    "argocd.argoproj.io/sync-options": "ServerSideApply=true"
```
roivaz, line 10: We deploy using Helm, not ArgoCD - this annotation will just be ignored. Furthermore, we already use server-side-apply with force-conflicts in Helm, see https://github.com/Azure/ARO-Tools/blob/main/tools/helm/options.go#L386-L387 and https://github.com/Azure/ARO-Tools/blob/main/tools/helm/options.go#L413-L414
I'm not entirely sure why we are getting this bad interaction between the two field managers if we are already using `ForceConflicts` in Helm.

roivaz: The error `"resource ClusterSizingConfiguration//cluster still exists. status: Current, message: Resource is current"` is **not** a field manager ownership conflict. It's a race condition in Helm's hook lifecycle.

### The sequence

1. Helm installs the HyperShift operator via the weight-1 Job hook (`installer.job.yaml`)
2. The operator starts up and may create its own default `ClusterSizingConfiguration/cluster` CR
3. When Helm reaches the weight-2 hook (`cluster.clustersizingconfiguration.yaml`), since no `helm.sh/hook-delete-policy` is explicitly set, Helm [defaults to `before-hook-creation`](https://github.com/helm/helm/blob/340b06d8/pkg/action/hooks.go#L85-L92)
4. Helm [deletes the existing resource and calls `WaitForDelete`](https://github.com/helm/helm/blob/340b06d8/pkg/action/hooks.go#L82-L83)
5. The HyperShift operator's reconciliation loop **immediately recreates** the CR
6. The [`waitForDelete` in the StatusWatcher](https://github.com/helm/helm/blob/340b06d8/pkg/kube/statuswait.go#L130-L153) never sees `NotFoundStatus` because the operator keeps bringing the resource back
7. Helm times out and produces the ["still exists" error](https://github.com/helm/helm/blob/340b06d8/pkg/kube/statuswait.go#L147-L148)

### Why it's intermittent (5%)

It depends on whether the HyperShift operator pod (started by the weight-1 Job) is fully running and reconciling fast enough to recreate the CR before Helm's `WaitForDelete` watch confirms deletion.

## Proposed Fix

Set `"helm.sh/hook-delete-policy": "hook-failed"` on the `ClusterSizingConfiguration` resource:

```yaml
annotations:
  "helm.sh/hook": post-install,post-upgrade
  "helm.sh/hook-weight": "2"
  "helm.sh/hook-delete-policy": "hook-failed"
```

This works because:

1. **Skips the `before-hook-creation` delete** — Helm won't try to delete the resource before applying, so there's no fight with the operator's reconciler
2. **SSA Create still works** — Helm's `Create` for hooks uses [`patchResourceServerSide`](https://github.com/helm/helm/blob/340b06d8/pkg/kube/client.go#L310-L333) (an SSA PATCH), which is an upsert — it succeeds whether the resource exists or not
3. **`WatchUntilReady` succeeds immediately** — for non-Job/non-Pod resources, the StatusWatcher uses [`alwaysReady`](https://github.com/helm/helm/blob/340b06d8/pkg/kube/statuswait.go#L69-L73) which returns `CurrentStatus` right away
4. **No resource disappears during rollouts** — unlike `before-hook-creation`, the CR is never deleted, so there's no window where the config is missing in production

roivaz: I opened https://github.com/Azure/ARO-HCP/pull/5335 with this approach.

patriksuba:
Closing in favor of #5335

PR #5330, opened 2026-05-19
test/e2e: be explicit in our failures

stevekuznetsov:
When teaching an Agentic flow to debug our end-to-end tests, I often found that it needed to read the source code of our tests to understand what on earth was failing, when presented with opaque and fairly useless error messages from Ginkgo. This is a good indication that humans would have just as much of an issue and that we should improve our tests. It should be clearly obvious what we were attempting to do when we failed, in addition to the proximal failure reason.


mgahagan73:
Can you also add instructions to the ## Assertion Messages section of test/AGENTS.md ? There is already a section for BeNil() for similar reasons.

Also how feasible is it add a static check for these types of annotations? I didn't know if there was prior art somewhere we could draw on.

deads2k:
Not a personal priority, but doesn't offend me.  Keep your tool working

/lgtm

test/e2e/admin_api.go
```diff
@@ -319,43 +319,43 @@ var _ = Describe("SRE", func() {
...
-			Expect(err).NotTo(HaveOccurred())
-			Expect(logs).NotTo(BeEmpty())
+			Expect(err).NotTo(HaveOccurred(), "failed to retrieve serial console logs for VM %q", vmName)
+			Expect(logs).NotTo(BeEmpty(), "serial console logs for VM %q should not be empty", vmName)
 
 			By("verifying serial console logs contain boot information")
 			// Serial console logs typically contain boot messages, kernel output, or systemd logs
 			// We just verify that we got some content back
-			Expect(len(logs)).To(BeNumerically(">", 0))
+			Expect(len(logs)).To(BeNumerically(">", 0), "serial console logs length should be greater than 0")
```
mbukatov, line 343: Why do we have both `NotTo(BeEmpty()` and `To(BeNumerically(">", 0)` with vm name present in just one:

```
Expect(logs).NotTo(BeEmpty(), "serial console logs for VM %q should not be empty", vmName)
Expect(len(logs)).To(BeNumerically(">", 0), "serial console logs length should be greater than 0")
```

I would suggest to keep just one line, and to make sure the vm name is in the message.

mbukatov (commented):
This is a good effort. Noticed one issue, but otherwise /lgtm

test/e2e/admin_api.go
stevekuznetsov: This PR changes no implementations, just adds commentary. Feel free to follow-up

stevekuznetsov:
/label lgtm

tagging after rebase

stevekuznetsov:
resource group cleanup timed out

/retest

stevekuznetsov:
AdminAPI deployment failure

stevekuznetsov:
/label lgtm

rebase

PR #5303, opened 2026-05-18
More fine grained incident correlations

mmazur:
Please ignore, this doesn't work, just need it for work tracking.

### PR Checklist
- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
- [ ] Screenshots included (if graph/UI/metrics changes)
- [ ] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [ ] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

PR #5332, opened 2026-05-19
feat: allow per-alert correlationId overrides; use for hostedcluster KAS alerts

mmazur:
## Summary
- The prometheus rules generator now respects `correlationId` annotations from source YAML instead of always overwriting them with the default `<AlertName>/{{ $labels.cluster }}`
- Adds `correlationId` override to `hostedcluster-KubeAPIServer-ErrorBudgetBurn` so each hosted cluster (`_id`) gets its own IcM incident, instead of collapsing all hosted clusters on a management cluster into one
- Updates KSM test expectations to match

Supersedes #5303


janboll (commented):
/lgtm

PR #5339, opened 2026-05-20
fix(region-pipeline): emit skipSync as unambiguous boolean

raelga:
## Summary

Fixes the JSON-schema validation failure that's blocking every `bump-aro-hcp` auto-PR (Azure DevOps definition `440895`) from passing the `incremental-test-pipeline` gate in `sdp-pipelines`.

## Root cause

Commit [`131c934f1`](https://github.com/Azure/ARO-HCP/commit/131c934f1) ("Skip datasources sync for prow", PR #5236) added this line to `dev-infrastructure/region-pipeline.yaml`:

```yaml
    skipSync: {{ .monitoring.skipDatasourceSync }}
```

When `sdp-pipelines` invokes `aro ev2 manifests generate`, it loads `region-pipeline.yaml` against a **placeholder configuration** (`dunderCfg`) for schema validation — see `tooling/pkg/ev2/manifests/generate/generate.go:116` and `mapping.go:NewDunderPlaceholders`. Every leaf value in `dunderCfg` is replaced by a string of the form `__<dot.path>__`. The rendered line becomes:

```yaml
    skipSync: __monitoring.skipDatasourceSync__
```

This is a YAML **string**, so the schema validator (which requires `skipSync: boolean` per `pipeline.schema.v1`) rejects it:

```
failed to generate Ev2 manifests for public/int: failed to load pipelines:
failed to precompile pipeline: failed to validate pipeline schema:
pipeline is not compliant with schema pipeline.schema.v1:
  - at '/resourceGroups/0/steps/3/skipSync': got string, want boolean
```

The dunder mechanism is designed to substitute **strings**; there's no analogous mechanism for non-string types in pipeline YAML. Other templated fields in this file (e.g. `grafanaName: "{{ .monitoring.grafanaName }}"`) work because they're quoted and the schema declares them `string`.

## Fix

Switch from a bare templated boolean to a conditional emit:

```yaml
    {{- if .monitoring.skipDatasourceSync }}
    skipSync: true
    {{- end }}
```

- **Validation pass** (`dunderCfg`): `.monitoring.skipDatasourceSync` is a non-empty string (truthy) → emits `skipSync: true` (literal boolean) → schema validates ✅
- **Runtime** (real config):
  - `skipDatasourceSync=true` (prow, ci01) → emits `skipSync: true` → sync skipped (unchanged behavior)
  - `skipDatasourceSync=false` (int/stg/prod) → field omitted → `step.SkipSync` defaults to `false` in Go → sync proceeds (unchanged behavior)

## Verification

Reproduced the exact CI failure locally with a small Go program that:
1. Reads `dev-infrastructure/region-pipeline.yaml`
2. Builds a dunder-placeholder config (same as sdp-pipelines does)
3. Calls `config.PreprocessContent` and `types.ValidatePipelineSchema`

Before the fix:
```
SCHEMA FAILED: ... '/resourceGroups/0/steps/3/skipSync': got string, want boolean
```
After the fix:
```
SCHEMA OK
```

Also verified with real config values that the conditional behaves correctly:
- `skipDatasourceSync=true`  → `skipSync present=true value=true type=bool`
- `skipDatasourceSync=false` → `skipSync present=false value=<nil> type=<nil>` (field omitted)

## Failing build for reference

- Pipeline: `incremental-test-pipeline` (definition 446903)
- Build: 164831058 (PR `15799428` → bump ARO-HCP to `517156cee690`)
- Failing step: `⚙️ Resolve Step Manifests for Microsoft.Azure.ARO.HCP.Global` (deployment-target `public/int`)

### DIFF
--- a/dev-infrastructure/region-pipeline.yaml
+++ b/dev-infrastructure/region-pipeline.yaml
@@ -60,7 +60,9 @@ resourceGroups:
       maximumRetryCount: 3
       durationBetweenRetries: 1m
     grafanaName: "{{ .monitoring.grafanaName }}"
-    skipSync: {{ .monitoring.skipDatasourceSync }}
+    {{- if .monitoring.skipDatasourceSync }}
+    skipSync: true
+    {{- end }}
     identityFrom:
       resourceGroup: global
       step: output
--- a/Makefile
+++ b/Makefile
@@ -201,7 +201,7 @@ CONTAINER_RUNTIME ?= docker
 
 mega-lint:
 	$(CONTAINER_RUNTIME) run --rm \
-		-e FILTER_REGEX_EXCLUDE='hypershiftoperator/deploy/crds/|acm/deploy/helm/multicluster-engine-config/charts/policy/charts|dev-infrastructure/global-pipeline.yaml|tooling/templatize/testdata/pipeline.yaml|hypershiftoperator/deploy/templates/cluster.clustersizingconfiguration.yaml' \
+		-e FILTER_REGEX_EXCLUDE='hypershiftoperator/deploy/crds/|acm/deploy/helm/multicluster-engine-config/charts/policy/charts|dev-infrastructure/global-pipeline.yaml|dev-infrastructure/region-pipeline.yaml|tooling/templatize/testdata/pipeline.yaml|hypershiftoperator/deploy/templates/cluster.clustersizingconfiguration.yaml' \
 		-e REPORT_OUTPUT_FOLDER=/tmp/report \
 		-v $${PWD}:/tmp/lint:Z \
 		docker.io/oxsecurity/megalinter-ci_light:v9


dev-infrastructure/region-pipeline.yaml
```diff
@@ -60,7 +60,9 @@ resourceGroups:
       maximumRetryCount: 3
       durationBetweenRetries: 1m
     grafanaName: "{{ .monitoring.grafanaName }}"
-    skipSync: {{ .monitoring.skipDatasourceSync }}
+    {{- if .monitoring.skipDatasourceSync }}
```
janboll, line 63: /hold
We should be consistent with typing.
let's use string for this I will create the required PRs

raelga: Thanks @janboll!

/close

PR #5341, opened 2026-05-20
feat: add prometheus alerting on frontend http request handling panics

miguelsorianod:
We add a new prometheus rule FrontendHttpRequestPanics that triggers when there's been a panic caught by the frontend http request handling panic middleware in the last 5 minutes.


miguelsorianod:
> /hold until https://github.com/Azure/ARO-HCP/pull/5290 is merged

After discussing about this, there's no need to hold until the metric is there. If the alert is defined and the metric does not exist the alert will not have anything to act on and will therefore not fire

/hold cancel

PR #5342, opened 2026-05-20
Specify API version suffixes in test framework helper functions

miquelsi:
https://redhat.atlassian.net/browse/ARO-27211

### What
This change is a versioned helper refactor across E2E tests and framework utilities.
It renames generic HCP helper methods to explicit API-versioned names and updates all test callsites accordingly.

- Standardized 20240610 helper names in framework utilities (cluster, nodepool, external auth, admin-credential helpers).
- Standardized 20251223 helper names for preview-specific cluster build/create flows.
- Updated E2E tests to call explicit versioned helpers instead of ambiguous generic helpers.
- Kept behavior intact: this is a naming/clarity refactor with no intended functional changes.

### Why

- Removes ambiguity now that multiple API versions are used in the same codebase.
- Prevents accidental use of the wrong SDK/API version in tests.
- Makes test intent and reviewability much clearer at each callsite.
- Creates a scalable naming pattern for future API-version additions without hidden coupling.

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/test/e2e/admin_api.go
+++ b/test/e2e/admin_api.go
@@ -61,13 +61,13 @@ var _ = Describe("SRE", func() {
 			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for admin-api-breakglass")
 
 			By("creating cluster parameters")
-			clusterParams := framework.NewDefaultClusterParams()
+			clusterParams := framework.NewDefaultClusterParams20240610()
 			clusterParams.ClusterName = engineeringClusterName
 			managedResourceGroupName := framework.SuffixName(*resourceGroup.Name, "-managed", 64)
 			clusterParams.ManagedResourceGroupName = managedResourceGroupName
 
 			By("creating customer resources")
-			clusterParams, err = tc.CreateClusterCustomerResources(ctx,
+			clusterParams, err = tc.CreateClusterCustomerResources20240610(ctx,
 				resourceGroup,
 				clusterParams,
 				map[string]interface{}{
@@ -81,7 +81,7 @@ var _ = Describe("SRE", func() {
 			Expect(err).NotTo(HaveOccurred(), "failed to create customer resources for breakglass cluster")
 
 			By("creating the HCP cluster")
-			err = tc.CreateHCPClusterFromParam(
+			err = tc.CreateHCPClusterFromParam20240610(
 				ctx,
 				GinkgoLogr,
 				*resourceGroup.Name,
@@ -276,13 +276,13 @@ var _ = Describe("SRE", func() {
 			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for serial console test")
 
 			By("creating cluster parameters")
-			clusterParams := framework.NewDefaultClusterParams()
+			clusterParams := framework.NewDefaultClusterParams20240610()
 			clusterParams.ClusterName = engineeringClusterName
 			managedResourceGroupName := framework.SuffixName(*resourceGroup.Name, "-managed", 64)
 			clusterParams.ManagedResourceGroupName = managedResourceGroupName
 
 			By("creating customer resources")
-			clusterParams, err = tc.CreateClusterCustomerResources(ctx,
+			clusterParams, err = tc.CreateClusterCustomerResources20240610(ctx,
 				resourceGroup,
 				clusterParams,
 				map[string]interface{}{
@@ -296,7 +296,7 @@ var _ = Describe("SRE", func() {
 			Expect(err).NotTo(HaveOccurred(), "failed to create customer resources for serial console cluster")
 
 			By("creating the HCP cluster")
-			err = tc.CreateHCPClusterFromParam(
+			err = tc.CreateHCPClusterFromParam20240610(
 				ctx,
 				GinkgoLogr,
 				*resourceGroup.Name,
@@ -306,12 +306,12 @@ var _ = Describe("SRE", func() {
 			Expect(err).NotTo(HaveOccurred(), "failed to create HCP cluster %q for serial console test", engineeringClusterName)
 
 			By("creating a nodepool to provision worker VMs")
-			nodePoolParams := framework.NewDefaultNodePoolParams()
+			nodePoolParams := framework.NewDefaultNodePoolParams20240610()
 			nodePoolParams.ClusterName = engineeringClusterName
 			nodePoolParams.NodePoolName = "worker"
 			nodePoolParams.Replicas = int32(1)
 
-			err = tc.CreateNodePoolFromParam(ctx,
+			err = tc.CreateNodePoolFromParam20240610(ctx,
 				GinkgoLogr,
 				*resourceGroup.Name,
 				managedResourceGroupName,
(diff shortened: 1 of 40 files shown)


test/e2e/admin_api.go
mbukatov: Shouldn't we version `NewDefaultClusterParams()` as well? When I look at the implementation, I see that it's based on current API level. And new API version could require different set of defaults.

test/util/framework/deployment_helper.go
```diff
@@ -360,7 +360,7 @@ func (tc *perItOrDescribeTestContext) CreateHCPClusterFromParam(
 	return nil
 }
 
-func (tc *perItOrDescribeTestContext) CreateHCPCluster20251223FromParam(
+func (tc *perItOrDescribeTestContext) CreateHCPClusterFromParam20251223(
```
mbukatov, line 363: Good catch, I have missed that we have this function, since it was not used in few cases when this would make sense.

test/util/framework/deployment_helper.go
```diff
@@ -426,9 +426,9 @@ func (tc *perItOrDescribeTestContext) CreateNodePoolFromParam(
 		return fmt.Errorf("nodePoolName parameter not found or empty")
 	}
 
-	nodePool := BuildNodePoolFromParams(parameters, tc.Location())
+	nodePool := BuildNodePoolFromParams20240610(parameters, tc.Location())
```
mbukatov, line 429: This is good, but there is one problem: we don't have `CreateNodePoolFromParam20251223` variant.

mbukatov (commented):
I like this approach more, and I believe it's moving into a good direction.

That said, I noticed that we don't version `NewDefaultClusterParams()`, wouldn't that make sense to version (and multiple versions of) as well?

Moreover I noticed that we have one important gap in the functions: we have `tc.CreateNodePoolFromParam20240610` but there is no `tc.CreateNodePoolFromParam20251223` (unlike for reateHCPClusterFromParam variants):

```
$ rg 'func.*CreateNodePoolFromParam.*'
util/framework/deployment_helper.go
406:func (tc *perItOrDescribeTestContext) CreateNodePoolFromParam20240610(
$ rg 'func.*CreateHCPClusterFromParam.*'
util/framework/deployment_helper.go
324:func (tc *perItOrDescribeTestContext) CreateHCPClusterFromParam20240610(
363:func (tc *perItOrDescribeTestContext) CreateHCPClusterFromParam20251223(
```

While fixing that, we will also need to define `BuildNodePoolFromParams20251223`.

Optional followup for this is to move `cluster_create_complex_cilium_kv.go` and `nodepool_ephemeral_osdisk.go` to use this missing function.

Minor easy to fix issues I noticed we need to fix as well:

- old references to previous naming scheme in comments or go doc comments every now and then (here I would suggest to task claude code to draft a fix for this issue specifically, review the result, and include it in a separate commit)
- updating AGENTS.md to cover the new naming scheme, to make sure the approach is followed consistently by people as well as bots updating or reviewing the code

test/util/framework/deployment_helper.go
miquelsi: I know, but I don't think it's necessary right now. It should be added when a test explicitly needs it (if ever).

test/e2e/admin_api.go
miquelsi: Done

test/util/framework/deployment_helper.go
mbukatov: Deployment of a cluster and a nodepool is a reasonable minimal set in my opinion. This is already needed in `cluster_create_complex_cilium_kv.go` and `nodepool_ephemeral_osdisk.go` tests.

test/e2e/admin_api.go
mbukatov: Don't we need NewDefaultClusterParams20251223 as well?

test/util/framework/deployment_params.go
```diff
@@ -162,8 +190,8 @@ func DefaultOpenshiftNodePoolChannelGroup() string {
 	return channelGroup
 }
 
-func NewDefaultClusterParams() ClusterParams {
-	return ClusterParams{
+func NewDefaultClusterParams20240610() ClusterParams20240610 {
+	return ClusterParams20240610{
```
mbukatov, line 194: Don't we need `NewDefaultClusterParams20251223` as well?

mbukatov (commented):
Thank you for adding `CreateNodePoolFromParam20251223()` function.

I see it defined and used:

```
$ rg -o 'func.*CreateNodePoolFromParam[0-9]+'
util/framework/deployment_helper.go
406:func (tc *perItOrDescribeTestContext) CreateNodePoolFromParam20240610
472:func (tc *perItOrDescribeTestContext) CreateNodePoolFromParam20251223
$ rg -I -o 'CreateNodePoolFromParam[0-9]+' | sort | uniq -c
     27 CreateNodePoolFromParam20240610
      3 CreateNodePoolFromParam20251223
```

That said, we may still need `NewDefaultClusterParams20251223()` and used it when this API level is used.

test/util/framework/deployment_params.go
miquelsi: I didn't introduce that because currently there is no test that needs a cluster created using that API version. But since I already added `ClusterParams20251223`, there is no harm in introducing `NewDefaultClusterParams20251223` too.

mbukatov: Thank you for introducing it.

I did a quick check, and I see that there are cases where `CreateHCPClusterFromParam20251223` is used:

```
$ rg -I -o 'CreateHCPClusterFromParam[0-9]+' | sort | uniq -c
     38 CreateHCPClusterFromParam20240610
      3 CreateHCPClusterFromParam20251223
```

```
$ rg  -o CreateHCPClusterFromParam20251223
util/framework/deployment_helper.go
363:CreateHCPClusterFromParam20251223
538:CreateHCPClusterFromParam20251223

e2e/idms_lifecycle.go
99:CreateHCPClusterFromParam20251223
```

So we can already update `e2e/idms_lifecycle.go` to use the new function you introduced there (as the params are created by `framework.NewDefaultClusterParams20240610` and cluster is created by `tc.CreateHCPClusterFromParam20251223`).

test/e2e/nodepool_ephemeral_osdisk.go
```diff
@@ -112,24 +111,21 @@ var _ = Describe("Nodepool Ephemeral OS Disk", func() {
...
+				GinkgoLogr,
 				*resourceGroup.Name,
+				clusterParams.ManagedResourceGroupName,
 				customerClusterName,
-				customerNodePoolName,
-				nodePool,
-				framework.NodePoolCreationTimeout,
-			)
+				nodePoolParams,
+				45*time.Minute)
```
mbukatov: I guess we want to keep `framework.NodePoolCreationTimeout` here.

mbukatov (commented):
The updates looks good, I noticed one discrepancy wrt timeout changes, but beyond that, I don't see any other reasons for updating this PR.

test/e2e/nodepool_ephemeral_osdisk.go
```diff
@@ -139,21 +135,14 @@ var _ = Describe("Nodepool Ephemeral OS Disk", func() {
...
-				if time.Now().Before(timeBombDeadline) {
-					Skip("v20251223preview deployed but DiskType field not present in nodepool response; skipping until rollout completes")
-				}
-				Fail(fmt.Sprintf("DiskType field still not present in v20251223preview nodepool response as of %s deadline", timeBombDeadline.Format(time.RFC3339)))
-			}
+			Expect(created.Properties.Platform.OSDisk.DiskType).ToNot(BeNil(), "created nodepool response Properties.Platform.OSDisk.DiskType was nil")
 			Expect(*created.Properties.Platform.OSDisk.DiskType).To(Equal(hcpsdk20251223preview.OsDiskTypeEphemeral), "expected created nodepool OSDisk.DiskType to be Ephemeral")
 			Expect(created.Properties.AutoRepair).ToNot(BeNil(), "created nodepool response Properties.AutoRepair was nil")
-			Expect(*created.Properties.AutoRepair).To(BeTrue(), "expected created nodepool AutoRepair to be true")
-
```
mbukatov, line 156: Did we drop this line by mistake?

mbukatov (commented):
Run a review via claude and noticed one more thing.

test/e2e/nodepool_ephemeral_osdisk.go
miquelsi: Yes, thanks for catching it!

PR #5344, opened 2026-05-20
test: skip minor upgrade test when no install version found

deads2k:
Narrowest possible nil pointer fix to easy to reason about merge.

The y-stream upgrade E2E test panics with nil pointer dereference at line 119 when no possibleInstallVersion has an upgrade path to the target minor. This happens when Cincinnati has no valid upgrade edges for the version combination being tested.

Add a nil check on installVersion before proceeding, skipping the test with a descriptive message instead of panicking.


deads2k:
discussed in slack. unbreaking CI. safe.  compiles. merging.

PR #5345, opened 2026-05-20
manage ingress on hostedCluster by cluster-admin

deads2k:
Adds proof that it is possible to use azure key vault certificates with ingress on the hostedcluster.


docs/ingress-certificates/design.md
```diff
@@ -0,0 +1,276 @@
...
+The closest first-party Azure-supported option is the [Azure Key Vault
+Provider for Secrets Store CSI
+Driver](https://github.com/Azure/secrets-store-csi-driver-provider-azure)
+(the same component used today for the service cluster's Istio gateway). It
+supports auto-rotation polling and can sync to a Kubernetes Secret via the
+`secretObjects` field on a `SecretProviderClass`.
+
+It is the wrong shape for **this** use case for two reasons:
+
+1. **The synced Kubernetes Secret only exists while a pod actively mounts the
```
miguelsorianod, line 43: Although this is correct for SecretProviderClass, what we do to solve that issue for some secrets we need to sync from keyvault that land in the management cluster is to use SecretProviderClass + SecretSync combination, which allows syncing to a K8s Secret without the need to have it mounted in a pod. SecretSync info: https://github.com/kubernetes-sigs/secrets-store-sync-controller.

This is what we use for example to sync some secrets from the management cluster's azure keyvault (for example the K8s APIServer certs) to the management cluster. SecretSync also supports automatically propagating changes on an existing secret in keyvault secret to the k8s secret. For example when having configured certificate in keyvault with rotation.

For the specific case of syncing from keyvault to the management cluster External Secrets Operator was also explored and it was discarded due to security concerns, although those concerns might not apply if it would be deployed in the data plane by the user (more info about the security concerns in https://redhat.atlassian.net/browse/ARO-15770?focusedCommentId=11161405)

docs/ingress-certificates/design.md
```diff
@@ -0,0 +1,276 @@
+# Design: Customer-Managed Ingress Serving Certificates from Azure Key Vault
+
+## Goal
+
+Let an ARO HCP customer:
```
miguelsorianod, line 5: We were recently made aware that Hypershift provides the ability to configure, as a creation-only configuration, the visibility of the ingress via a K8s annotation:
```
	// IngressControllerLoadBalancerScope is an annotation that allows the specification of the LoadBalancer scope for ingress controller.
	IngressControllerLoadBalancerScope = "hypershift.openshift.io/ingress-controller-load-balancer-scope"
```

although from the doc it's not clear if that's only for the default ingress or also for additional ingresses that are created.

In ARO-HCP (and OCM Commercial too) the visibility is configured via an ACM Policy, as well as the TLS certificate itself. In ARO-HCP the default ingress TLS certificate for now is generated by the service.

Did we consider Hypershift providing the ability to set the TLS certificate too via HostedCluster CR, given that it provides a pseudo-api to configure the visibility of it?

Trying to understand where do we want to go around configuration of those two aspects or what we offer or we don't offer (only user-provided, both service-provided and user-provided, for the user-provided case which mechanism directly in dataplane or via ARM API than then it's configured somehow etc...)

deads2k: We want to end up not in the way of cluster-admins configuring their ingress.

test/e2e/cluster_ingress_cert_from_kv.go
```diff
@@ -0,0 +1,592 @@
...
+					Audiences: []*string{to.Ptr("api://AzureADTokenExchange")},
+				},
+			}, nil)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("granting the UAMI Key Vault Certificate User + Key Vault Secrets User on the customer Key Vault")
+			roleAssignmentsClient, err := armauthorization.NewRoleAssignmentsClient(subscriptionID, cred, nil)
+			Expect(err).NotTo(HaveOccurred())
+			kvScope := fmt.Sprintf(
+				"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.KeyVault/vaults/%s",
```
deads2k, line 199: I think this means it's contained to the same resourcegroup so our clean up works.

PR #5352, opened 2026-05-21
Arobit: Bump timeout and add retries

janboll:
### What/Why

Increase the timeout of the helm deployment, cause we saw issues during node upgrades and other deployment issues resolving after some time. Also add a retry if deployment is still in progress, so it wil wait at max 33 minutes before failing in these cases now.


ashishmax31 (approved):

raelga (approved):
/lgtm

PR #5355, opened 2026-05-21
fix(arobit): recover mdsd pods from CSI mount race (AROSLSRE-879)

raelga:
## What

Mitigate the AKS secrets-store CSI driver mount race that leaves arobit-forwarder mdsd-clusterlogs pods stuck in CrashLoopBackOff with `No certificate was found that matched SAN` — see [AROSLSRE-879](https://redhat.atlassian.net/browse/AROSLSRE-879).

## Why the wrapper alone wasn't enough

The original version of this PR wrapped mdsd's `command:` with a polling loop that waits for `/geneva/geneva_auth/gcscert.pem` to appear before exec'ing the real entrypoint. That cleans up the log timeline during the normal startup race, but it **cannot recover** the pathological case where the CSI driver loses the initial NodePublishVolume race and the tmpfs comes up empty for the lifetime of the pod sandbox:

- Pod-level CSI volumes are mounted by `NodePublishVolume`, which is only called on **pod sandbox creation**, NOT on container restart.
- Container restarts inside the same pod (whether kubelet-driven crashloops or our wrapper exiting nonzero) keep the same empty tmpfs.
- AKS's rotation reconciler walks `SecretProviderClassPodStatus` objects, but SPCPS is only created at the END of a successful `NodePublishVolume`, so broken pods are invisible to the reconciler.
- **The only recovery is pod recreation.**

## What this PR ships

Two layered pieces, both gated on `mdsd.enabled && clusterType == "mgmt"`:

### 1. mdsd-clusterlogs polling wrapper (kept from the original revision)

Wraps the container `command:` with a POSIX-sh poll that waits up to 300s for a non-empty cert at `/geneva/geneva_auth/gcscert.pem` before exec'ing `/start_mdsd.sh`. Uses only POSIX builtins (the mdsd image is distroless, no openssl/jq). On healthy nodes the cert is present within seconds and mdsd starts cleanly; on stuck nodes mdsd no longer panic-spams the log timeline while it waits.

### 2. `mdsd-cert-watchdog` sidecar (NEW — this is the actual recovery mechanism)

A second container in the same pod that polls the same cert from the same CSI mount. If the cert never appears within 300s, it asks the K8s API to delete its own pod:

```sh
kubectl delete pod "$POD_NAME" -n "$POD_NAMESPACE" --grace-period=30 --wait=false
```

The DaemonSet then schedules a replacement, which gets a fresh `NodePublishVolume` call and a new shot at the race. Once the cert is present (or after a successful delete request) the watchdog `exec sleep infinity` and stays idle until kubelet terminates it.

### Image

The watchdog uses `mcr.microsoft.com/aks/command/runtime`, which is:
- Already pinned + image-updater tracked under `platform-utils` in `config/config.yaml` (`defaults.aksCommandRuntime.image`)
- Already in our supply chain (used by `cluster-service/cspr/orphaned-namespace-cleaner.yaml`)
- Available in prod ACR mirrors (mcr.microsoft.com)
- Contains a recent kubectl (v1.35.3) and a POSIX `sh`

No new image dependency.

### RBAC

`forwarder-watchdog-rbac.yaml` adds a namespace-scoped `Role` named `arobit-cert-watchdog` granting only `pods:delete`, bound to the existing `arobit-forwarder` ServiceAccount. Strictly less privileged than the cluster-wide `arobit-forwarder` ClusterRole; blast radius limited to the `arobit` namespace. The Role can't be `resourceNames`-scoped because DaemonSet pod names are randomly generated.

## Why both pieces

- Wrapper covers the common case (cert arrives a few seconds late) — quiet startup, no panic-loop noise.
- Watchdog covers the pathological case (cert never arrives because CSI lost the race) — bounded auto-recovery via pod deletion.

## Verification

- `cd tooling/helmtest && go test -run TestHelmTemplate -count=1` passes.
- Helmtest fixtures regenerated; only `zz_fixture_TestHelmTemplate_helmtest_mdsd_and_kusto_enabled_mgmt.yaml` is affected. All SVC fixtures are unchanged because the `clusterType == "mgmt"` gate filters everything out for SVC clusters.
- Smoke-tested the wrapper script inside the pinned mdsd image (`docker pull` + `docker run --entrypoint sh`) — `/bin/sh`, `/usr/bin/test`, `/bin/sleep` all present.
- Smoke-tested kubectl + the watchdog script inside the pinned aksCommandRuntime image — kubectl v1.35.3, `/bin/sh` (dash) present.

## Out of scope

Filed separately:
- File an AKS / secrets-store CSI driver upstream issue once we have a clean repro signal from the watchdog logs.
- Add a Prometheus alert that fires when watchdog deletes a pod (so we can quantify how often the race fires).
# Upstream context

Direct upstream tracking for our race:

* [**kubernetes-sigs/secrets-store-csi-driver#1051**](https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/1051) — `"Apparent race condition causes pods to mount secret volumes with no object content"` — open since 2022 (`kind/bug`, `lifecycle/frozen`). Reporter's logs show the same `"already mounted to target"` short-circuit returning success against an empty mount.
* [**kubernetes-sigs/secrets-store-csi-driver#2025**](https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/2025) — `"fix: treat empty existing mount as stale in ensureMountPoint"` — open PR from April 2026. Detects an empty target directory in `ensureMountPoint` (`pkg/secrets-store/utils.go`) and forces a re-mount instead of taking the `!rotationEnabled && mounted` early-return in `nodeserver.go:157`. This is exactly the fix our race needs.

The watchdog in [Azure/ARO-HCP#5355](https://github.com/Azure/ARO-HCP/pull/5355) is the recovery mechanism until #2025 lands and AKS ships the fixed driver.

Related kubelet-side hardening (different code path, same class of bug):

* [kubernetes/kubernetes#121271](https://github.com/kubernetes/kubernetes/issues/121271) / [#139045](https://github.com/kubernetes/kubernetes/pull/139045) — fix for the rotation/republish path where kubelet's cleanup-on-error deleted CSI mount dirs after a `requiresRepublish=true` refresh failure. Merged 2026-05-21. Different trigger from our initial-mount race but useful precedent for tightening CSI mount lifecycle invariants.

### DIFF
--- a/observability/arobit/deploy/templates/forwarder-daemonset.yaml
+++ b/observability/arobit/deploy/templates/forwarder-daemonset.yaml
@@ -109,9 +109,45 @@ spec:
         - name: mdsd-clusterlogs
           image: {{ .Values.forwarder.mdsd.image.registry }}/{{ .Values.forwarder.mdsd.image.repository }}@{{ .Values.forwarder.mdsd.image.digest }}
           imagePullPolicy: '{{ .Values.forwarder.mdsd.image.pullPolicy }}'
+          # AROSLSRE-879: defensively wait for the Geneva auth cert before exec'ing
+          # mdsd. The AKS secrets-store CSI driver can lose the initial
+          # NodePublishVolume race, leaving /geneva/geneva_auth/ empty; mdsd then
+          # panics with "No certificate was found that matched SAN" on every
+          # restart. The mdsd-cert-watchdog sidecar below recovers stuck pods by
+          # deleting them; this wrapper keeps the normal startup race window quiet.
+          # Upstream bug: kubernetes-sigs/secrets-store-csi-driver#1051
+          # Upstream fix: kubernetes-sigs/secrets-store-csi-driver#2025 (in flight)
+          # Remove this wrapper once #2025 lands and AKS ships the fixed driver.
+          # Polls for a non-empty cert file (mdsd does its own X.509 + SAN
+          # validation), then exec's the real entrypoint. If the cert never
+          # appears we exit non-zero so the kubelet restart counter increments and
+          # existing container-restart alerts still fire. Uses /bin/sh and
+          # /bin/sleep, both present in the distroless mdsd image (the watchdog
+          # sidecar already uses /bin/sleep directly). Timeout and poll interval
+          # are sourced from the watchdog's chart values so the two recovery
+          # windows stay coordinated when tuned.
           command:
-              - /start_mdsd.sh
+            - /bin/sh
+            - -c
+            - |
+              CERT=/geneva/geneva_auth/gcscert.pem
+              WAITED=0
+              while [ "$WAITED" -lt "$TIMEOUT_SECONDS" ]; do
+                if [ -s "$CERT" ]; then
+                  echo "[mdsd-init] Geneva auth cert present after ${WAITED}s; exec'ing mdsd"
+                  exec /start_mdsd.sh
+                fi
+                echo "[mdsd-init] Geneva auth cert not yet mounted at ${CERT} (waited ${WAITED}s of ${TIMEOUT_SECONDS}s)"
+                sleep "$INTERVAL_SECONDS"
+                WAITED=$((WAITED + INTERVAL_SECONDS))
+              done
+              echo "[mdsd-init] Giving up after ${TIMEOUT_SECONDS}s; Geneva auth cert never appeared at ${CERT}. This usually means the CSI driver lost the initial mount race; see AROSLSRE-879. Exiting to trigger container restart."
+              exit 1
           env:
+            - name: TIMEOUT_SECONDS
+              value: '{{ .Values.forwarder.watchdog.timeoutSeconds }}'
+            - name: INTERVAL_SECONDS
+              value: '{{ .Values.forwarder.watchdog.intervalSeconds }}'
             - name: MONITORING_GCS_AUTH_ID
               value: '{{ .Values.forwarder.mdsd.geneva.clusterLogsSan }}'
             - name: MONITORING_GCS_AUTH_ID_TYPE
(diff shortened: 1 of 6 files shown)


observability/arobit/deploy/templates/forwarder-daemonset.yaml
```diff
@@ -109,8 +109,35 @@ spec:
         - name: mdsd-clusterlogs
           image: {{ .Values.forwarder.mdsd.image.registry }}/{{ .Values.forwarder.mdsd.image.repository }}@{{ .Values.forwarder.mdsd.image.digest }}
           imagePullPolicy: '{{ .Values.forwarder.mdsd.image.pullPolicy }}'
+          # AROSLSRE-879: defensively wait for the Geneva auth cert to appear
```
stevekuznetsov: Please update this with a link to the upstream PR that fixed the issue and a comment about roughly what version of K8s we expect to be able to remove this.

observability/arobit/deploy/templates/forwarder-daemonset.yaml
```diff
@@ -109,8 +109,35 @@ spec:
...
+          # exit non-zero so the kubelet restart counter increments and the
+          # existing container-restart alerts still fire. We only use POSIX
+          # builtins because the mdsd image is distroless (no openssl).
           command:
-              - /start_mdsd.sh
+            - /bin/sh
+            - -c
+            - |
+              CERT=/geneva/geneva_auth/gcscert.pem
+              TIMEOUT_SECONDS=300
```
stevekuznetsov: Couldn't you achieve this also by an exec readiness probe on startup that just checks the file?

raelga: A probe doesn't gate the entrypoint. mdsd still execs immediately and SIGABRTs in milliseconds when the cert is missing, so every lost race bumps `restart_count` and trips our restart alerts. The wrapper gates the exec itself, so on healthy nodes the cert-arrival window is just silent instead of a crash burst.

(An initContainer would see the same empty mount and hang too, still needs the watchdog.)

observability/arobit/deploy/templates/forwarder-daemonset.yaml
raelga: Pushed. Upstream tracking already exists: `kubernetes-sigs/secrets-store-csi-driver#1051` (open since 2022, frozen) with fix PR `#2025` from April targeting the exact early-return path in `ensureMountPoint`/`nodeserver.go:157`. The YAML comment now links both with a "remove when #2025 lands and AKS ships the fixed driver" note instead of a K8s version.

observability/arobit/deploy/templates/forwarder-daemonset.yaml
```diff
@@ -109,8 +109,35 @@ spec:
...
+          # its own X.509 + SAN validation) and only then exec the real
+          # entrypoint. If the cert never appears within the bounded wait we
+          # exit non-zero so the kubelet restart counter increments and the
+          # existing container-restart alerts still fire. We only use POSIX
+          # builtins because the mdsd image is distroless (no openssl).
           command:
-              - /start_mdsd.sh
+            - /bin/sh
+            - -c
+            - |
```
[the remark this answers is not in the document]
raelga: Addressed in `796c7778c` as part of the response to Steve. `sleep` is present in the distroless mdsd image (the watchdog sidecar already uses `/bin/sleep` directly); the new comment block drops the "POSIX builtins" framing and calls out the `/bin/sh` + `/bin/sleep` dependency explicitly. No further action needed.

observability/arobit/testdata/zz_fixture_TestHelmTemplate_helmtest_mdsd_and_kusto_enabled_mgmt.yaml
```diff
@@ -678,8 +678,35 @@ spec:
...
+          # AROSLSRE-879: defensively wait for the Geneva auth cert to appear
+          # before exec'ing mdsd. The AKS secrets-store CSI driver can lose the
+          # initial mount race on a node, leaving /geneva/geneva_auth/ empty and
+          # causing mdsd to panic with "No certificate was found that matched
+          # SAN" on every restart. We poll for a non-empty cert file (mdsd does
+          # its own X.509 + SAN validation) and only then exec the real
+          # entrypoint. If the cert never appears within the bounded wait we
+          # exit non-zero so the kubelet restart counter increments and the
+          # existing container-restart alerts still fire. We only use POSIX
+          # builtins because the mdsd image is distroless (no openssl).
```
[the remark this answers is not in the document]
raelga: Fixture regenerated in `796c7778c` to match the updated template wording. No further action needed.

observability/arobit/deploy/templates/forwarder-daemonset.yaml
```diff
@@ -109,8 +109,40 @@ spec:
...
+          # /bin/sleep, both present in the distroless mdsd image (the watchdog
+          # sidecar already uses /bin/sleep directly).
           command:
-              - /start_mdsd.sh
+            - /bin/sh
+            - -c
+            - |
+              CERT=/geneva/geneva_auth/gcscert.pem
+              TIMEOUT_SECONDS=300
+              INTERVAL=15
```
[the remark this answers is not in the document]
raelga: Applied: the wrapper now sources `TIMEOUT_SECONDS` and `INTERVAL_SECONDS` as env vars from `.Values.forwarder.watchdog.timeoutSeconds/intervalSeconds`, matching the watchdog's naming and keeping the two recovery windows in lockstep. Comment block updated to call this out. Fixture regenerated. New SHA `7725d5176`. No further action or commit needed from you — this is resolved.

raelga:
## Watchdog self-delete mechanism verified on personal dev

Tested on `pers-usw3rael-mgmt-1` with a standalone pod using the same RBAC and kubectl self-delete logic as the watchdog sidecar.

### Test setup

Deployed a pod with the watchdog script (45s timeout, 15s interval) and a namespace-scoped Role granting `pods:delete` to the pod's ServiceAccount.

### Results

```
=== EVENTS ===
67s  Normal  Scheduled  pod/watchdog-test  Successfully assigned arobit/watchdog-test to aks-userswft2-10093028-vmss000000
67s  Normal  Pulled     pod/watchdog-test  Container image already present on machine
67s  Normal  Created    pod/watchdog-test  Container created
67s  Normal  Started    pod/watchdog-test  Container started
21s  Normal  Killing    pod/watchdog-test  Stopping container watchdog

=== POD STATUS ===
Error from server (NotFound): pods "watchdog-test" not found
```

Pod started, waited 3 rounds (0s/15s/30s), reached 45s timeout, called `kubectl delete pod` on itself, and was terminated. Pod is `NotFound` after deletion.

### What was validated

1. Namespace-scoped `pods:delete` Role + RoleBinding work (no cluster-wide permissions needed)
2. `kubectl delete pod "$POD_NAME"` from within the pod successfully triggers deletion
3. The DaemonSet controller would recreate the pod after deletion (not tested here since standalone pod, but DaemonSet behavior is standard)

### What couldn't be tested in personal dev

The full DaemonSet delete-loop with mdsd + CSI cert mount couldn't be tested because:
- `mdsd.enabled` is `false` in all dev environments (Geneva certs only exist in INT+)
- When forced to `true`, the CSI SecretProviderClass fails on `NodePublishVolume` because the Geneva cert secret doesn't exist in the personal dev keyvault, blocking pod startup entirely (pods stuck in `ContainerCreating`)

This limitation actually demonstrates the exact failure mode the watchdog addresses: when CSI mount fails, no container in the pod can start, so only pod recreation (via the watchdog's `kubectl delete`) can recover.

### Full DaemonSet validation path

To validate the complete cycle (cert mount race -> watchdog delete -> DaemonSet recreate), the PR needs to be tested in INT where Geneva certs exist. The watchdog will be deployed in Audit mode first (300s timeout, no actual deletion initially) to observe the race frequency before enabling deletion.

PR #5361, opened 2026-05-21
mgmt-agent: add informer-driven data dumping

stevekuznetsov:
We have a large number of GVRs on the management cluster which we do not directly create or control from the service cluster, but are nevertheless crucial for deep understanding of errors in HostedCluster state. We can start a set of informers for API groups of interest and dump state as we see it.

This approach is naive, but should be functional. Areas for improvement:
- backoff/rate-limiting: high churn rate on status may cause large log volumes; we can try to alleviate this with a logging backoff policy that could give us guarantees within one pod restart period
- resource consumption: in order to not put a high load on the management cluster's API server, we use LIST+WATCH and keep an in-memory store of all the resource we care about; if we find the data set is larger than we'd like, we can revisit this


mgmt-agent/deploy/templates/clusterrole.yaml
```diff
@@ -20,3 +20,10 @@ rules:
   verbs:
   - patch
   - update
+- apiGroups:
+  - "*"
+  resources:
+  - "*"
+  verbs:
+  - list
+  - watch
```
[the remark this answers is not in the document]
stevekuznetsov: Do we care? Maybe we exclude core/v1?

mgmt-agent/pkg/controller/resourcewatcher.go
```diff
@@ -0,0 +1,195 @@
...
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+// watchedGroupSuffixes is the hardcoded list of API group domain suffixes
+// whose resources will be discovered and watched.
+var watchedGroupSuffixes = []string{
+	"open-cluster-management.io",
```
miguelsorianod, line 38: Maybe tracking SecretProviderClass and SecretSync CRs would be useful too?

We end up managing several of them but I wonder if there might be some managed by Hypershift?

stevekuznetsov: If we manage them, then we should have read desires - for the HyperShift ones, I am not so sure. I don't think they create any.

stevekuznetsov:
```
+ podman pull --storage-opt ignore_chown_errors=true mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:vnext-preview
Trying to pull mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:vnext-preview...
Error: initializing source docker://mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:vnext-preview: too many requests to registry
```

/retest

raelga (approved):
/lgtm
/approve

PR #5363, opened 2026-05-21
feat: add backend hcp backup controller

tony-schndr:
https://redhat.atlassian.net/browse/ARO-25416

### What

Adds a backend controller to create and manage velero backup schedules via KubeApplier per hosted cluster.

This new enhancement uses configuration driven schedules to create / reconcile the velero backup schedule on mgmt clusters per hosted cluster.  The configuration includes a global "pause" state in order to pause the backups incase backups are impacting normal cluster operations.

Admin API endpoints are added to perform List/Get to view the available backups and create on demand backups of hosted clusters.  The ability to pause a backup schedule is also exposed through the admin api.

### Why

To implement the backup capability required for Disaster Recovery of Hosted Cluster

### Testing

Unit / E2E testing is included

### PR Checklist
- [ ] Self-reviewed the diff


backend/pkg/controllers/backupcontroller/backup_controller.go
```diff
@@ -0,0 +1,397 @@
...
+	maestroClient, err := createMaestroClientFromManagementCluster(ctx, mc, c.maestroSourceEnvironmentIdentifier, c.maestroClientBuilder)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to create Maestro client: %w", err))
+	}
+
+	clusterID := existingCluster.ServiceProviderProperties.ClusterServiceID.ID()
+	clusterName := existingCluster.CustomerProperties.DNS.BaseDomainPrefix
+	hcNamespace := c.getHostedClusterNamespace(c.maestroSourceEnvironmentIdentifier, clusterID)
+	hcpNamespace := fmt.Sprintf("%s-%s", hcNamespace, clusterName)
+	manifestWorkName := fmt.Sprintf("%s-dr", clusterID)
```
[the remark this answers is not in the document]
geoberle: legit feedback. most controllers interacting with these fields check upfront, e.g.
backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go

admin/server/handlers/hcp/backup_profile.go
```diff
@@ -0,0 +1,126 @@
...
+			return
+		}
+
+		spc, err := database.GetOrCreateServiceProviderCluster(request.Context(), resourcesDBClient, resourceID)
+		if err != nil {
+			http.Error(writer, fmt.Sprintf("failed to get service provider cluster: %v", err), http.StatusInternalServerError)
+			return
+		}
+
+		spc.Status.BackupState = state
```
geoberle: from an api perspective, how we handle spc.Status.BackupState feels more like a spec than a status.
i will check the rest of the code to verify

admin/server/handlers/hcp/backup_profile.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	if spc.Status.LastBackupTime != nil {
+		resp.LastBackupTime = spc.Status.LastBackupTime.String()
+	}
+	return resp
+}
+
+// GetBackupProfile returns the backup profile for a cluster.
+func GetBackupProfile(resourceDBClient database.ResourcesDBClient) http.Handler {
+	return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
```
geoberle: nit: for new handlers we switched from HandlerFunc to pure http.Handler

admin/server/handlers/hcp/backup_profile.go
```diff
@@ -0,0 +1,126 @@
...
+			http.Error(writer, fmt.Sprintf("failed to encode output: %v", err), http.StatusInternalServerError)
+			return
+		}
+	})
+}
+
+// PatchBackupProfile updates the backup schedule state for a cluster.
+// Only the "state" field can be updated; other fields are read-only.
+func PatchBackupProfile(resourcesDBClient database.ResourcesDBClient) http.Handler {
+	return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
```
geoberle: nit: for new handlers we switched from HandlerFunc to pure http.Handler

backend/pkg/controllers/backupcontroller/backup_controller.go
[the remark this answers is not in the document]
adalrsjr1: resolved

backend/pkg/controllers/backupcontroller/backup_controller.go
```diff
@@ -0,0 +1,397 @@
...
+	ctx context.Context,
+	mc *fleet.ManagementCluster,
+	maestroSourceEnvironmentIdentifier string,
+	maestroClientBuilder maestro.MaestroClientBuilder,
+) (maestro.Client, error) {
+	provisionShardID := mc.Status.ClusterServiceProvisionShardID.ID()
+	maestroSourceID := maestro.GenerateMaestroSourceID(maestroSourceEnvironmentIdentifier, provisionShardID)
+
+	return maestroClientBuilder.NewClient(ctx, mc.Status.MaestroRESTAPIURL, mc.Status.MaestroGRPCTarget, mc.Status.MaestroConsumerName, maestroSourceID)
+}
```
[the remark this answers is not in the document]
adalrsjr1: resolved - moved to kube-applier

docs/backups.md
```diff
@@ -0,0 +1,256 @@
...
+
+ARO-HCP uses [Velero](https://velero.io/) to perform automated backups of Hosted Control Plane (HCP) resources. The backup system is composed of:
+
+- A **backup schedule controller** in the backend service that creates and manages Velero Schedule resources on management clusters via Maestro.
+- An **admin API** that exposes endpoints for on-demand backup creation, listing, and pause/resume of scheduled backups.
+- **Velero** deployed on each management cluster with the Azure and HyperShift plugins.
+- **Azure Blob Storage** as the backup storage backend.
+
+Backups capture the Kubernetes resources that define a hosted control plane — not volume data. This allows disaster recovery by recreating the control plane from backed-up manifests.
+
```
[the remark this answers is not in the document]
adalrsjr1: resolved

backend/pkg/controllers/backupcontroller/backup_controller.go
```diff
@@ -0,0 +1,397 @@
...
+
+	maestroClient, err := createMaestroClientFromManagementCluster(ctx, mc, c.maestroSourceEnvironmentIdentifier, c.maestroClientBuilder)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to create Maestro client: %w", err))
+	}
+
+	clusterID := existingCluster.ServiceProviderProperties.ClusterServiceID.ID()
+	clusterName := existingCluster.CustomerProperties.DNS.BaseDomainPrefix
+	hcNamespace := c.getHostedClusterNamespace(c.maestroSourceEnvironmentIdentifier, clusterID)
+	hcpNamespace := fmt.Sprintf("%s-%s", hcNamespace, clusterName)
```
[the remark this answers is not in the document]
adalrsjr1: resolved

backend/pkg/controllers/backupcontroller/delete_orphaned_backup_manifestworks.go
```diff
@@ -0,0 +1,310 @@
...
+	index map[string]*backupShardServiceProviderClusters,
+) error {
+	logger := utils.LoggerFromContext(ctx)
+	var syncErrors []error
+
+	for csShardID, shardToSPCs := range index {
+		logger = logger.WithValues("csProvisionShardID", csShardID)
+		ctx = utils.ContextWithLogger(ctx, logger)
+		logger.Info("processing cluster service provision shard", "csProvisionShardID", csShardID, "spcCount", len(shardToSPCs.serviceProviderClusters))
+		maestroClient := shardToSPCs.maestroClient
```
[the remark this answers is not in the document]
adalrsjr1: resolved

admin/server/handlers/hcp/backups.go
```diff
@@ -0,0 +1,314 @@
...
+	if len(hostedClusters.Items) > 1 {
+		return nil, fmt.Errorf("multiple hosted clusters found for cluster %s", clusterID)
+	}
+	return &hostedClusters.Items[0], nil
+}
+
+type GetBackupResponse struct {
+	ResourceID string
+	Backup     BackupResponse
+}
```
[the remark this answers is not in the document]
adalrsjr1: resolved

backend/deploy/templates/backend.backup-config.configmap.yaml
geoberle, line 1: i don't see the need to have backup schedules massively customizable per environment.
proposal: get rid of the configmap and configmap mounting. instead define the schedules to be created in code with all the details like names, ttl, ...
a change in config or a change in code both require a rollout and i don't think we need different values in different envs.

also: keep in mind that the number of config.yaml sourced parameters you can add to a helm deployment are limited in EV2 to 64 (59 usable by us). the schedules here consume 9 of them we can safe a lot of parameters by not making this customizable. and we will need the parameter quota for the CS -> backend migration.

backend/cmd/root.go
```diff
@@ -64,6 +65,7 @@ type BackendRootCmdFlags struct {
 	InsecureIgnoreUserAzureManagedIdentitiesThatNeedManagedIdentitiesDataplaneAvailableAndUseMock bool
 	ExitOnPanic                                                                                   bool
 	AzureClusterScopedIdentitiesRoleSetName                                                       string
+	BackupConfigPath                                                                              string
```
geoberle: in relation to my comment about not having the need for configmap templates for schedules: all changes in here related to backup config would go away

backend/pkg/controllers/backupcontroller/backup_controller.go
```diff
@@ -0,0 +1,437 @@
...
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("get ReadDesire CRUD: %w", err))
+	}
+
+	clusterID := existingCluster.ServiceProviderProperties.ClusterServiceID.ID()
+	clusterName := existingCluster.CustomerProperties.DNS.BaseDomainPrefix
+	hcNamespace := c.getHostedClusterNamespace(c.hostedClusterNamespaceEnvIdentifier, clusterID)
+	hcpNamespace := fmt.Sprintf("%s-%s", hcNamespace, clusterName)
+
+	clusterPaused := spc.Status.BackupState == api.BackupScheduleStatePaused
```
geoberle: mentioned in admin api again. spc.Status.BackupState should be in spec, not status

backend/pkg/controllers/backupcontroller/backup_controller.go
```diff
@@ -0,0 +1,437 @@
...
+
+func (c *backupScheduleSyncer) CooldownChecker() controllerutil.CooldownChecker {
+	return c.cooldownChecker
+}
+
+// getHostedClusterNamespace returns the namespace for the hosted cluster.
+// Format: ocm-<envName>-<csClusterID>
+func (c *backupScheduleSyncer) getHostedClusterNamespace(envName string, csClusterID string) string {
+	return fmt.Sprintf("ocm-%s-%s", envName, csClusterID)
+}
```
geoberle: there is already a func like this in the codebase
backend/pkg/controllers/cs_maestro_utils.go
hostedClusterNamespace

tooling/templatize/go.mod
```diff
@@ -312,4 +312,6 @@ require (
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.2 // indirect
 )
 
+replace helm.sh/helm/v4 => github.com/geoberle/helm/v4 v4.0.0-20251102095138-e64345e0f7ed
```
geoberle: i guess this is a rebase error. we removed this a while ago

admin/server/handlers/hcp/backups.go
geoberle, line 1: lets remove this handler and tackle it in another PR. we should not use directl MC kube-api access but leverage kube-applier.
BUT for that we need to implement ReadDesireMany first, as described in the design docs of kube-applier.

backend/pkg/controllers/backupcontroller/backup_controller.go
```diff
@@ -0,0 +1,437 @@
...
+		}
+
+	case action.updateSPC != nil:
+		spcCRUD := c.cosmosClient.ServiceProviderClusters(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+		_, err = spcCRUD.Replace(ctx, action.updateSPC, nil)
+		if err != nil {
+			return utils.TrackError(fmt.Errorf("failed to update ServiceProviderCluster status: %w", err))
+		}
+
+	}
```
geoberle: i think for just 3 actions, the action based approach is too heavy
consider just going for simple level based
if (needsDoingA) { do it}
if (needsDoingB) { do it }
...

consider handling all three things in our syncOnce call, level driven and idempotent

backend/pkg/controllers/backupcontroller/delete_orphaned_backup_desires.go
```diff
@@ -0,0 +1,318 @@
...
+	cosmosClient database.ResourcesDBClient
+
+	queue workqueue.TypedRateLimitingInterface[string]
+
+	kubeApplierDBClients database.KubeApplierDBClients
+}
+
+// NewDeleteOrphanedBackupDesiresController periodically finds backup ApplyDesires
+// in kube-applier containers that are not referenced by any ServiceProviderCluster
+// and deletes them.
```
geoberle: do we have a race condition here? the backup controller creates XDesire for schedules in a first step and writes the names down into the SPC in a second. if this controller runs in between, it cleans up the XDesires

geoberle: backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go does cleanup for orphaned XDesire documents. the semantics are a bit different from this one though.

this: cleans up everything not mentioned in the SPC
the other: cleans up everything where the cluster is gone

it is not exactly the same but i want you to think about it to figure out if we could go without storing desire names in SPC and relying completely on cluster_child_resources_cleanup_controller.go for cleanup

backend/pkg/controllers/backupcontroller/backup_controller.go
```diff
@@ -0,0 +1,437 @@
...
+// recordDesireNamesInStatus checks if the SPC status has the ApplyDesire names recorded.
+// If not, returns an action to update the SPC. If already recorded, passes through.
+func (c *backupScheduleSyncer) recordDesireNamesInStatus(_ context.Context, state *backupSyncState) (bool, *backupAction, error) {
+	if slices.Equal(state.spc.Status.BackupScheduleApplyDesireNames, state.applyDesireNames) {
+		return false, nil, nil
+	}
+
+	state.spc.Status.BackupScheduleApplyDesireNames = state.applyDesireNames
+	return true, &backupAction{updateSPC: state.spc}, nil
+}
```
geoberle: i wonder if we need this. i know maestro read only bundles did this but the replacement ReadDesire don't. they rely on well known names for their desire objects. we could do the same and save us the trouble of updating SPC.

backend/pkg/controllers/backupcontroller/backup_controller.go
```diff
@@ -0,0 +1,437 @@
...
+			needsUpdate = true
+		}
+	}
+
+	if !needsUpdate {
+		return true, nil, nil
+	}
+
+	state.spc.Status.LastBackupTime = latestTime
+	state.spc.Status.LastBackupStatus = latestPhase
```
geoberle: do we need this here for anything at this point? is there any consumer of these two fields?

dev-infrastructure/templates/svc-mgmt-aks-permissions.bicep
```diff
@@ -36,3 +39,23 @@ resource sessiongateAksAccess 'Microsoft.Authorization/roleAssignments@2022-04-0
...
+
+resource adminApiAksAccess 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
+  scope: aksCluster
+  name: guid(resourceGroup().id, aksClusterName, adminApiMIResourceId, aksClusterRBACAdminRoleId)
+  properties: {
+    roleDefinitionId: aksClusterRBACAdminRoleId
+    principalId: adminApiMSI.properties.principalId
+    principalType: 'ServicePrincipal'
+  }
+}
```
geoberle: we need to find a way to do it without giving admin api access to the mgmt clusters.
followup PR

tony-schndr: Removed access

backend/pkg/controllers/backupcontroller/delete_orphaned_backup_desires.go
tony-schndr: I was able to move this to cluster_child_resources_cleanup_controller.go , Create DeleteDesire, wait for deletion, then delete all related Desires.

tony-schndr: Should be fixed with new deletion flow using cluster_child_resources_cleanup_controller.go

backend/cmd/root.go
tony-schndr: Still using configmap, down to two parameters `pause` and `backupCadence`

backend/deploy/templates/backend.backup-config.configmap.yaml
tony-schndr: Down to two parameters `pause` and `backupCadence`

admin/server/handlers/hcp/backup_profile.go
tony-schndr: Agreed, my goof.  Moved it to spec

backend/pkg/controllers/backupcontroller/backup_controller.go
tony-schndr: No, its removed now

backend/pkg/controllers/backupcontroller/backup_controller.go
tony-schndr: Removed the update to spc.

backend/pkg/controllers/backupcontroller/backup_controller.go
tony-schndr: fixed

backend/pkg/controllers/backupcontroller/backup_controller.go
tony-schndr: fixed

tooling/templatize/go.mod
tony-schndr: fixed

admin/server/handlers/hcp/backups.go
tony-schndr: Removed

config/config.schema.json
```diff
@@ -1294,6 +1294,15 @@
         "maestro": {
           "$ref": "#/definitions/backendMaestroConfig"
         },
+        "backupCadence": {
+          "type": "string",
+          "description": "Backup cadence preset: 'normal' creates hourly/daily/weekly schedules, 'testing' creates a 5-minute cycle schedule",
+          "enum": ["normal", "testing"]
```
adalrsjr1: nit: I would suggest `cadence: production` rather `cadence: normal`? production/testing are more meaningful than normal/testing

backend/pkg/controllers/backupcontroller/schedule_controller.go
```diff
@@ -0,0 +1,367 @@
+// Copyright 2026 Microsoft Corporation
+//
```
miguelsorianod, line 2: Have we addressed what happens on cluster/nodepool/externalauth deletion?

There should be an ordered deletion process, including k8s resources, cosmos resources, ...

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -324,12 +332,19 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
...
+		}
+		return true, nil
+	}
 	// extraDeleteGates uses lowercased kubeapplier.*DesireResourceTypeName keys. Types not
 	// in the map are deleted unconditionally.
 	extraDeleteGates := map[string]func(ctx context.Context, resourceID *azcorearm.ResourceID) (bool, error){
-		// strings.ToLower(kubeapplier.ClusterScopedReadDesireResourceType.String()): c.extraDeleteGateShouldDeleteReadDesire,
-		// strings.ToLower(kubeapplier.ClusterScopedApplyDesireResourceType.String()): c.extraDeleteGateShouldDeleteApplyDesire,
-		// strings.ToLower(kubeapplier.ClusterScopedDeleteDesireResourceType.String()): c.extraDeleteGateShouldDeleteDeleteDesire,
+		strings.ToLower(kubeapplier.ClusterScopedApplyDesireResourceType.String()):  skipBackupDesires,
```
miguelsorianod: In the future there will probably be more handling than just backupdesires. Can we prepare it for that so we can expand appropriately?

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -160,6 +164,10 @@ func (c *clusterChildResourcesCleanupController) SyncOnce(ctx context.Context, k
 		strings.ToLower(api.ClusterControllerResourceType.String()): func(ctx context.Context, resourceID *azcorearm.ResourceID) (bool, error) { return false, nil },
 	}
 
+	if err := c.ensureBackupScheduleKubeObjectsDeleted(ctx, clusterResourceID); err != nil {
```
miguelsorianod: This doesn't belong at this level but at the ensureClusterScopedKubeApplierResourcesDeleted level. At this level the only ones are the two containers: resources and kubeappliers. Specifics on how to delete cosmos resources from within those containers belongs inside the corresponding methods

tony-schndr: fixed

backend/pkg/controllers/backupcontroller/schedule_controller.go
tony-schndr: The desires are deleted when the cluster is deleted.  The actual backups will be deleted when the backup TTL expires by velero

test/e2e/hcp_backups.go
```diff
@@ -0,0 +1,337 @@
...
+		labels.CoreInfraService,
+		labels.DevelopmentOnly,
+		labels.AroRpApiCompatible,
+		labels.Slow,
+		func(ctx context.Context) {
+			cluster := createBackupTestCluster(ctx, backupTestClusterConfig{
+				resourceGroupPrefix: "manual-bkp-e2e",
+				clusterName:         "manual-bkp-cluster",
+				nsgName:             "manual-bkp-nsg-name",
+				vnetName:            "manual-bkp-vnet-name",
```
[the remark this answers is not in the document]
tony-schndr: Shouldn't matter since resource group is unique

test/e2e/hcp_backups.go
```diff
@@ -0,0 +1,337 @@
...
+		labels.DevelopmentOnly,
+		labels.AroRpApiCompatible,
+		labels.Slow,
+		func(ctx context.Context) {
+			cluster := createBackupTestCluster(ctx, backupTestClusterConfig{
+				resourceGroupPrefix: "pause-bkp-e2e",
+				clusterName:         "pause-bkp-cluster",
+				nsgName:             "pause-bkp-nsg-name",
+				vnetName:            "pause-bkp-vnet-name",
+				subnetName:          "pause-bkp-vnet-subnet1",
```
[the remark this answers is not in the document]
tony-schndr: shouldn't matter since the resource group is unique

test/e2e/hcp_backups.go
```diff
@@ -0,0 +1,336 @@
...
+				}
+				for _, s := range resp.Schedules {
+					if s.LastBackupTime != "" {
+						return true, nil
+					}
+				}
+				return false, nil
+			}, scheduleWaitTimeout, schedulePollInterval).Should(BeTrue(), "at least one schedule should have a LastBackupTime")
+
+			By("pausing the backup schedule")
```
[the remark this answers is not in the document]
tony-schndr: Tests cannot run in production since PROW has no access to geneva actions.

backend/pkg/controllers/backupcontroller/backup_controller.go
```diff
@@ -0,0 +1,394 @@
...
+		return false
+	}
+	if existingCluster.ServiceProviderProperties.BillingDocumentCosmosID == "" {
+		return false
+	}
+	return true
+}
+
+func (c *backupScheduleSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	existingCluster, err := c.cosmosClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
```
ahitacat: We could use a cachedCluster with the clusterListers to check if it needs work before calling cosmosClient. This patterns is repeated across most of the controllers.

test/e2e/hcp_backups.go
```diff
@@ -0,0 +1,319 @@
...
+		err := tc.AssignIdentityContainers(ctx, 1, framework.IdentityContainerAssignmentRetryInterval)
+		Expect(err).NotTo(HaveOccurred(), "failed to assign identity containers")
+	}
+
+	By("creating a resource group")
+	resourceGroup, err := tc.NewResourceGroup(ctx, cfg.resourceGroupPrefix, tc.Location())
+	Expect(err).NotTo(HaveOccurred(), "failed to create resource group")
+
+	By("creating cluster parameters")
+	clusterParams := framework.NewDefaultClusterParams20240610()
```
ahitacat: Why still using 20240610 api and not a newer one?

test/e2e/hcp_backups.go
```diff
@@ -0,0 +1,319 @@
...
+						return false, fmt.Errorf("schedule %s LastBackupTime changed from %s to %s while paused", name, lastTime, currentTime)
+					}
+				}
+				return true, nil
+			}, framework.BackupWaitTimeout, framework.BackupWaitInterval).Should(BeTrue(), "no schedule should execute new backups while paused")
+		})
+
+})
+
+func doAdminAPIRequest[T any](ctx context.Context, httpClient *http.Client, method, url string, expectedStatus int, body io.Reader) (T, error) {
```
ahitacat: This seems to be general enough to be in admin_api_helpers.go of the framework, no?

test/e2e/hcp_backups.go
tony-schndr: I was following the guidance of testing with the oldest API.  I guess it makes more sense to bump it to the 2025 API since 2024 is being retired.

adalrsjr1 (commented):
mistake comment

admin/server/handlers/hcp/backups.go
```diff
@@ -0,0 +1,444 @@
...
+
+		ad, rd, err := buildOnDemandBackupDesires(
+			bCtx.resourceID.SubscriptionID, bCtx.resourceID.ResourceGroupName, bCtx.resourceID.Name,
+			backupName, bCtx.managementClusterResourceID, hcpBackup,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to build desires: %w", err)
+		}
+
+		if _, err := adCrud.Create(request.Context(), ad, nil); err != nil {
```
adalrsjr1: non-indempotent due to timetamp (L280). Two sequential requests with same payload won't be considered the same because the timestamp has changed. 

The same happens in L301.

backend/pkg/controllers/backupcontroller/schedule_controller.go
```diff
@@ -0,0 +1,350 @@
...
+		}
+		if database.IsNotFoundError(adErr) {
+			if _, err := adCrud.Create(ctx, desired, nil); err != nil {
+				return false, utils.TrackError(fmt.Errorf("failed to create ApplyDesire %s: %w", desired.ResourceID.Name, err))
+			}
+		}
+		if database.IsNotFoundError(rdErr) {
+			if _, err := rdCrud.Create(ctx, desiredRDs[i], nil); err != nil {
+				return false, utils.TrackError(fmt.Errorf("failed to create ReadDesire %s: %w", desiredRDs[i].ResourceID.Name, err))
+			}
```
adalrsjr1: when both AD and RD are missing, the AD create falls through into the RD create; two writes with no exit between them. It should return after creating AD and the next reconcile create the RD

backend/pkg/controllers/backupcontroller/schedule_controller.go
```diff
@@ -0,0 +1,350 @@
...
+		if ad.Spec.Type == kubeapplier.ApplyDesireTypeDelete {
+			if !isDesireSuccessful(ad.Status.Conditions) {
+				continue
+			}
+			if err := rdCrud.Delete(ctx, name); err != nil && !database.IsNotFoundError(err) {
+				return false, utils.TrackError(fmt.Errorf("failed to delete stale ReadDesire %s: %w", name, err))
+			}
+			if err := adCrud.Delete(ctx, name); err != nil && !database.IsNotFoundError(err) {
+				return false, utils.TrackError(fmt.Errorf("failed to delete successful stale ApplyDesire %s: %w", name, err))
+			}
```
adalrsjr1, line 305: same comment as in L219

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -316,11 +320,18 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
 		return nil
 	}
 
+	// Backup desires are managed by ensureBackupScheduleKubeObjectsDeleted
+	// and must not be swept here while that cleanup is in progress.
+	skipBackupDesires := func(_ context.Context, resourceID *azcorearm.ResourceID) (bool, error) {
+		if strings.HasPrefix(resourceID.Name, backup.BackupScheduleDesireNamePrefix) {
```
adalrsjr1, line 328: only check `backupschedule-` prefix, `ondemandbackup-` is swept without Delete-type conversion.

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -316,11 +320,18 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
...
+			return false, nil
+		}
+		return true, nil
+	}
 	// extraDeleteGates uses lowercased kubeapplier.*DesireResourceTypeName keys. Types not
 	// in the map are deleted unconditionally.
 	extraDeleteGates := map[string]func(ctx context.Context, resourceID *azcorearm.ResourceID) (bool, error){
-		// strings.ToLower(kubeapplier.ClusterScopedReadDesireResourceType.String()): c.extraDeleteGateShouldDeleteReadDesire,
-		// strings.ToLower(kubeapplier.ClusterScopedApplyDesireResourceType.String()): c.extraDeleteGateShouldDeleteApplyDesire,
+		strings.ToLower(kubeapplier.ClusterScopedApplyDesireResourceType.String()): skipBackupDesires,
```
adalrsjr1, line 336: this looks like a race condition; is there a change of ReadDesires being swept before `ensureBackupScheduleKubeObjectsDeleted` can reference them?

backend/pkg/controllers/backupcontroller/ondemand_cleanup_controller.go
```diff
@@ -0,0 +1,178 @@
...
+		if !strings.HasPrefix(name, backup.OndemandBackupDesireNamePrefix) {
+			continue
+		}
+		// KubeContent non-nil means the Velero Backup still exists on the management cluster.
+		if rd.Status.KubeContent != nil && rd.Status.KubeContent.Raw != nil {
+			continue
+		}
+		// Guard against newly-created ReadDesires (also nil KubeContent): if the
+		// ApplyDesire still exists the backup has not yet been applied.
+		if _, err := adCrud.Get(ctx, name); err == nil || !database.IsNotFoundError(err) {
```
adalrsjr1: the condition treats transient errorrs the same as 'AD Exists"; for example

err == NetworkTimeout; !IsNotFoundError --> true; Will continue without logging the transient error;

admin/server/handlers/hcp/backups.go
tony-schndr: This endpoint creates backups, by nature it will not be idempotent.

backend/pkg/controllers/backupcontroller/schedule_controller.go
```diff
@@ -0,0 +1,350 @@
...
+		return utils.TrackError(fmt.Errorf("failed to get cached ServiceProviderCluster: %w", err))
+	}
+
+	mcResourceID := cachedSPC.Status.ManagementClusterResourceID
+	if mcResourceID == nil {
+		return nil
+	}
+
+	kaClient := c.kubeApplierDBClients.For(ctx, mcResourceID)
+	if kaClient == nil {
```
stevekuznetsov, line 143: Is this a programmer bug? When could this ever happen?

backend/pkg/controllers/backupcontroller/schedule_controller.go
```diff
@@ -0,0 +1,350 @@
...
+	rdCrud, err := kaClient.ReadDesiresForCluster(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("get ReadDesire CRUD: %w", err))
+	}
+
+	if cachedCluster.ServiceProviderProperties.ClusterServiceID == nil {
+		return nil
+	}
+	clusterName := cachedCluster.CustomerProperties.DNS.BaseDomainPrefix
+	if clusterName == "" {
```
stevekuznetsov, line 160: The sheer number of these early-exits make me a bit suspicious - why? Do they relate to a true state we expect to sit in, or are they just defensive, or?

backend/pkg/controllers/backupcontroller/schedule_controller.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	return false, nil
+}
+
+func applyDesireNeedsUpdate(existing, desired *kubeapplier.ApplyDesire) bool {
+	if existing == nil {
+		return true
+	}
+
+	var existingRaw, desiredRaw []byte
```
stevekuznetsov: I think you want `semantic.Equal` or `cmp.Diff` or trust that the `.Raw` is deterministically serialized and do a strict comparison

stevekuznetsov: also - do other controllers need this kind of logic? why or why not? what are they doing?

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -316,11 +320,18 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
 		return nil
 	}
 
+	// Backup desires are managed by ensureBackupScheduleKubeObjectsDeleted
```
stevekuznetsov: I'm not sure, but this seems fragile. Do we have any other options?

backend/pkg/controllers/backupcontroller/schedule_controller.go
tony-schndr:
```
	// For returns the client for the given management cluster, constructing it
	// on demand. It walks the configured ManagementClusterLister to find the
	// management cluster whose ResourceID matches; the container name comes
	// from the management cluster's Status.KubeApplierCosmosContainerName.
	// Returns nil if no management cluster matches, if the lister errors, or
	// if the matched management cluster has no container name configured.
```

It seems more correct to return an error instead of nil, I will change this.

backend/pkg/controllers/backupcontroller/schedule_controller.go
tony-schndr: Defensive and not a state I would expect since `NeedsWork` should be enough of a gate.  Makes more sense to return errors here instead.

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
tony-schndr: What seems fragile?  The fact that I need to set the ApplyDesires to Delete, wait for delete to be successful, then remove from cosmos?

backend/pkg/controllers/backupcontroller/schedule_controller.go
tony-schndr: Updated to return an error

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
tony-schndr: Updated the comment to point out the reasoning why ondemandbackup is not considered.

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
tony-schndr: Yes, I removed the readDesire deletion from `ensureBackupScheduleKubeObjectsDeleted` to avoid any races.  I'm keying off of the status of the ApplyDesire in Delete mode so the ReadDesire can go away whenever, nothing is dependent on the ReadDesire.

internal/backup/backup.go
```diff
@@ -0,0 +1,76 @@
...
+	"statefulsets",
+	"hostedcontrolplane",
+	"cluster",
+	"azurecluster",
+	"azuremachinetemplate",
+	"azuremachine",
+	"machinedeployment",
+	"machineset",
+	"machine",
+	"route",
```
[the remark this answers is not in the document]
tony-schndr: This does not matter.

test/util/framework/constants.go
```diff
@@ -45,3 +45,10 @@ const (
 const (
 	IdentityContainerAssignmentRetryInterval = 60 * time.Second
 )
+
+// Backup timeouts
+const (
+	BackupTimeout      = 10 * time.Minute
+	BackupWaitTimeout  = 6 * time.Minute
+	BackupWaitInterval = 30 * time.Second
+)
```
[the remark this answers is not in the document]
tony-schndr: The backup controller is tuned so watches and reacts to changes, if it doesn't resolve within 6 minutes then there is something wrong with the controller.  It should take < 1 minute to propagate changes to the management cluster and at most 5 minutes for a backup to appear.

backend/pkg/controllers/backupcontroller/schedule_controller.go
tony-schndr: Changed to return an error here.

backend/pkg/controllers/backupcontroller/schedule.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/vmware-tanzu/velero/pkg/builder"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-HCP/internal/backup"
+)
+
+func NewScheduledBackup(clusterID, clusterName, hcNamespace, hcpNamespace, backupName, cronSchedule string, ttl time.Duration, paused bool) *velerov1api.Schedule {
+	scheduleName := fmt.Sprintf("%s-%s", clusterID, backupName)
+	backup := backup.NewBackup(scheduleName, clusterID, ttl, hcNamespace, hcpNamespace)
```
adalrsjr1: nit: variable `backup` shadowing import `backup`

admin/server/handlers/hcp/backups.go
```diff
@@ -0,0 +1,474 @@
...
+		}
+
+		hcNamespace := hcReadDesire.Spec.TargetItem.Namespace
+		if hcNamespace == "" {
+			return fmt.Errorf("HostedCluster ReadDesire has empty namespace")
+		}
+		hcpNamespace := fmt.Sprintf("%s-%s", hcNamespace, clusterName)
+
+		timestamp := bCtx.time.Now().Format("20060102150405")
+		backupName := fmt.Sprintf("%s-%s", clusterServiceID, timestamp)
```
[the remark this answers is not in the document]
tony-schndr: One second should be enough, this is a manually invoked endpoint by SRE, who shouldn't be spamming this.

velero/deploy/templates/kustomize-patch.yaml
```diff
@@ -39,8 +39,6 @@ data:
       template:
         spec:
           priorityClassName: service-lifecycle-critical
-          nodeSelector:
-            aro-hcp.azure.com/role: infra
           tolerations:
           - key: "infra"
             operator: "Equal"
```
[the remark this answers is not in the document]
tony-schndr: required for snapshot upload on nodes hosting hcp

tony-schndr:
Closed, superseded by https://github.com/Azure/ARO-HCP/pull/6261

PR #5365, opened 2026-05-21
refactor verifiers to improve/simplify polling

mgahagan73:
### What

followup/refactor of  https://github.com/Azure/ARO-HCP/pull/5175
### Why

Use of EventuallyVerify() in previous form broke parallel verifier workflow.


mbukatov:
I don't know why I missed this, will review it tomorrow.

test/util/verifiers/daemonset.go
```diff
@@ -0,0 +1,66 @@
...
+	name      string
+}
+
+func (v verifyDaemonSetReady) Name() string {
+	return daemonSetVerifierName(v.namespace, v.name)
+}
+
+func daemonSetVerifierName(namespace, name string) string {
+	return fmt.Sprintf("VerifyDaemonSetReady(%s/%s)", namespace, name)
+}
```
mbukatov, line 35: Why don't include this function in Name() directly? It's the only place this function is used.

test/util/verifiers/poll.go
```diff
@@ -0,0 +1,162 @@
...
+		diagnose: cfg.diagnose,
+	}
+}
+
+type pollVerifier struct {
+	inner    HostedClusterVerifier
+	timeout  time.Duration
+	interval time.Duration
+	diagnose diagnoseFunc
+}
```
mbukatov: This looks like a special verifier, which wraps another verifier instance, and reruns it based on the given parameters.

I'm not convinced that such an interface is desirable. This functionality should not be provided as a verifier imho. Each verifier can have different polling needs, some may not need it at all. What if some verifier needs an exponential back off and use [ExponentialBackoffWithContext](https://github.com/kubernetes/apimachinery/blob/v0.36.1/pkg/util/wait/backoff.go#L493)? Then it can't use this functionality.

That said the fact that you control it via a particular verifier (rg. via poll options of VerifyCatalogSourceReady) is good.

test/AGENTS.md
```diff
@@ -30,6 +30,9 @@
 
 * **K8S client-go:** Use this client to communicate with created HCP clusters. Client requires rest Config which is provided by method `GetAdminRESTConfigForHCPCluster` with 10 minutes timeout.
 * **HostedClusterVerifier:** This interface is designed for all verifiers and provides the essential `Name` and `Verify` methods for extension.
+* **Parallel checks:** Use `VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyFoo(), verifiers.VerifyBar(), ...)` to run independent verifiers in parallel. Each verifier is responsible for its own polling, diagnostics, and delta-only logging when it needs to wait.
+* **Polling:** Pass `verifiers.WithPolling(timeout, interval)` to verifier constructors when a condition may take time; omit poll options for a single-shot `Verify`. This is the only poll option for tests — do not call internal `maybePoll`. Use `verifiers.DefaultPollInterval` when the test does not need a custom interval. Reuse `verifiers.VerifyDaemonSetReady(namespace, name, ...)` for any DaemonSet readiness check; `VerifyGlobalPullSecretSyncer` is a thin alias for the syncer in kube-system.
```
mbukatov: I don't think it's a good idea to make the polling explicit by making it visible in the verifier interface. That creates unreasonable complexity imho.

test/AGENTS.md
```diff
@@ -75,9 +78,9 @@ When writing tests that poll until a condition is met (e.g. `Eventually(...)`, `
...
+5. **Polling inside verifiers:** Use `verifiers.WithPolling(timeout, interval)` on verifier constructors so delta-only logging and elapsed-time reporting live in `test/util/verifiers/poll.go`. When polling completes, verifiers must log actual wall-clock duration (success and timeout) via `GinkgoLogr` (`elapsed` field) and `GinkgoWriter` (`[VerifierName] succeeded after 1m23s`). Tests should call `Expect(verifier.Verify(ctx, cfg)).NotTo(HaveOccurred(), "intent message")` — the annotation describes what the test was trying to do; the returned error carries the proximal failure (see **Descriptive error checks** above). Do not add test-level `EventuallyVerify`-style wrappers that bypass `VerifyHCPCluster` parallelism.
```
mbukatov: That sounds as if verifiers which does polling already, but not using this verifier are doing it wrong. That is not the case.

mbukatov (changes requested):
I don't like the idea of rerunning a whole verifier via a verifier interface. This is something which should be done insider of a verifier. Moreover such a simple approach won't work with more complex verifiers, which may eg. do multiple pooling loops with some extra processing, rerunning the whole thing is not desirable.

That said, I see that you:

- removed EventuallyVerify
- remove DiagnosticVerifier interface 

Which is a good step forward.

test/util/verifiers/daemonset.go
mbukatov: This is still valid comment. Why not just merge `daemonSetVerifierName` into `Name()`?

test/util/verifiers/poll.go
```diff
@@ -0,0 +1,131 @@
...
+	ctx context.Context,
+	name string,
+	restConfig *rest.Config,
+	timeout time.Duration,
+	diagnose diagnoseFunc,
+	check func(context.Context, *rest.Config) error,
+) error {
+	if timeout == 0 {
+		return check(ctx, restConfig)
+	}
```
mbukatov: I would rather avoid API where timeout set to zero means no polling. This is not obvious, one can easily miss that when writing E2E test or during a review.

The decision whether to poll should be on the author of the verifier, not the user who just runs in a test. E2E test author won't be checking implementation details of a verifier to decide whether to poll or not, or how to set a timeout. E2E test author will rather just select a verifier for a given functionality and pass required information (such as kubeadmin restConfig).

Another discrepancy I see here is that it won't run diagnose function even if the run once case fail.

So I propose to use pollUntilReady in verifiers directly, and drop verifyOnceOrPoll.

test/util/verifiers/poll.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	// DefaultPollInterval is the standard interval between verifier poll attempts in E2E tests.
+	DefaultPollInterval = 15 * time.Second
+	// DefaultDiagnoseTimeout is the timeout for collecting failure diagnostics after a poll times out.
+	DefaultDiagnoseTimeout = 30 * time.Second
```
mbukatov, line 32: Could we have this as a mandatory parameters of the verifyOnceOrPoll function? This will highly depend on what action a verifier is doing, and I don't see any way to redefine it if needed.

mbukatov (commented):
Thank you for the changes, I like that you moved the functionality into a function one can use from a verifier.

That said, I would simplify this further: use pollUntilReady in verifiers directly, and drop verifyOnceOrPoll. See review comments for reasoning and context.

PR #5368, opened 2026-05-22
hcpctl: snapshot: add a local analysis CLI

stevekuznetsov:
hcpctl: snapshot: separate queries by phase

When we know that a test failed during the normal phase, or during
cleanup, it's best to give agents a set of logs that start and stop
during the timestamps that are pertinent, so as to not lead them astray
with deletion-related content for normal analysis or vice versa.

---

hcpctl: snapshot: add a local analysis CLI

---

hcpctl: snapshot: provide cluster filter in dev

---

### DIFF
--- a/docs/snapshot-analysis.md
+++ b/docs/snapshot-analysis.md
@@ -1,0 +1,101 @@
+# Snapshot Gathering and Analysis
+
+The `hcpctl snapshot` commands gather structured diagnostic data for failed
+E2E tests and run LLM-driven root cause analysis on the results.
+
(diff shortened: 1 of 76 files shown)


stevekuznetsov:
I don't want this to bit-rot, and I won't be around to respond to reviews. This is entirely an opt-in tool, so there's little to no harm in just stamping this so it merges in as long as it doesn't break the snapshot gathering jUnit.

/label lgtm

raelga (approved):
/lgtm

PR #5378, opened 2026-05-25
feat: persist IntentFailed for node pool version validation errors

ahitacat:
[ARO-26304](https://redhat.atlassian.net/browse/ARO-26304)

### What

Persist node pool version validation errors as `IntentFailed` conditions on the controller document instead of returning them as transient errors, and teach the operation controller to read those conditions to determine operation outcomes.

Similar work was done for control plane upgrade controller https://github.com/Azure/ARO-HCP/pull/5042

### Why

Previously, version validation errors (downgrades, Cincinnati `VersionNotFound`, missing upgrade paths, exceeding control plane version, skipping minor versions) were returned as errors from `SyncOnce`, causing the controller to retry them indefinitely. These are user errors that won't resolve on retry. By persisting them as `IntentFailed` conditions, the operation controller can detect them and fail the operation with a clear error message, while still retrying transient Cincinnati errors (e.g. 503).

### Testing

- Added new `TestNodePoolVersionSyncer_SyncOnce_IntentFailed` table with 6 test cases covering: successful resolution (IntentFailed=False), downgrade, Cincinnati VersionNotFound, Cincinnati upstream error (transient), exceeds control plane, and skipping minor versions
- Added operation controller tests covering: exact version match succeeds, IntentFailed marks operation failed, stale IntentFailed from previous version is ignored, 29s timeout for version resolution, and version mismatch without IntentFailed stays Accepted
- Updated `TestNodePoolVersionSyncer_SyncOnce_DesiredVersionUnchangedOnFailure_ChangedOnSuccess` to reflect that VersionNotFound now returns nil (persists IntentFailed) instead of returning an error

- `service_provider_cluster_conditions.go` was renamed to `service_provider_conditions.go` since conditions now apply to both clusters and node pools
- The operation controller uses a `[version X.Y.Z]` prefix in the IntentFailed condition message to scope conditions to a specific customer version request, preventing stale errors from blocking new requests. There could be another approach to consider a dedicated field for cleaner version scoping. https://github.com/Azure/ARO-HCP/pull/5378#discussion_r3297706893
- The operation controller waits twice (59s timout) the informers relist period used by the nodepool version controller. This would cover most of the calculation edge cases in which the new desiredVersion has not yet calculated.
- `subscriptionLister` was added to the version controller for feature flag resolution (experimental release features)

### PR Checklist
- [x] Self-reviewed the diff

### DIFF
--- a/backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
@@ -16,26 +16,36 @@ package operationcontrollers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 	"time"
 
+	"github.com/blang/semver/v4"
+
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
 	utilsclock "k8s.io/utils/clock"
+	"k8s.io/utils/lru"
 
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/upgradecontrollers"
 	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/database"
 	"github.com/Azure/ARO-HCP/internal/ocm"
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
 type operationNodePoolUpdate struct {
-	clock                utilsclock.PassiveClock
-	resourcesDBClient    database.ResourcesDBClient
-	clusterServiceClient ocm.ClusterServiceClientSpec
-	notificationClient   *http.Client
+	resourcesDBClient               database.ResourcesDBClient
+	clusterServiceClient            ocm.ClusterServiceClientSpec
+	notificationClient              *http.Client
+	clock                           utilsclock.PassiveClock
+	desiredVersionMismatchFirstSeen *lru.Cache
 }
 
 // NewOperationNodePoolUpdateController returns a new Controller instance that
@@ -60,10 +70,11 @@ func NewOperationNodePoolUpdateController(
 	activeOperationInformer cache.SharedIndexInformer,
 ) controllerutils.Controller {
 	syncer := &operationNodePoolUpdate{
-		clock:                clock,
-		resourcesDBClient:    resourcesDBClient,
-		clusterServiceClient: clusterServiceClient,
-		notificationClient:   notificationClient,
+		resourcesDBClient:               resourcesDBClient,
+		clusterServiceClient:            clusterServiceClient,
+		notificationClient:              notificationClient,
+		clock:                           clock,
+		desiredVersionMismatchFirstSeen: lru.New(100000),
 	}
 
 	controller := NewGenericOperationController(
@@ -104,6 +115,156 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key 
 	if !c.ShouldProcess(ctx, operation) {
 		return nil // no work to do
 	}
+	if len(operation.InternalID.String()) == 0 {
+		// Cannot proceed yet
+		// TODO when we update to clusterservice node pool creation async, we need to handle this correctly.
+		return nil
+	}
 
-	return pollNodePoolStatus(ctx, c.clock, c.resourcesDBClient, c.clusterServiceClient, operation, c.notificationClient)
+	operationalState, err := c.determineOperationState(ctx, operation)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	var persistErr *arm.CloudErrorBody
+	if operationalState.provisioningState == arm.ProvisioningStateFailed {
+		persistErr = &arm.CloudErrorBody{
+			Code:    arm.CloudErrorCodeInvalidRequestContent,
+			Message: operationalState.message,
+		}
+	}
+
+	logger.Info("updating status")
+	if err := UpdateOperationStatus(ctx, c.clock, c.resourcesDBClient, operation, operationalState.provisioningState, persistErr, postAsyncNotificationFn(c.notificationClient)); err != nil {
+		return utils.TrackError(err)
+	}
+	return nil
+}
+
+func (c *operationNodePoolUpdate) determineOperationState(ctx context.Context, operation *api.Operation) (*operationState, error) {
+	logger := utils.LoggerFromContext(ctx)
+	errs := []error{}
+	operationStates := []*operationState{}
+
+	if operationState, err := c.desiredVersionResolutionOperationState(ctx, operation); err != nil {
+		errs = append(errs, utils.TrackError(err))
+	} else {
+		operationStates = append(operationStates, operationState)
+	}
+	if operationState, csErr := c.nodePoolServiceUpdateOperationState(ctx, operation); csErr != nil {
+		errs = append(errs, utils.TrackError(csErr))
+	} else {
+		operationStates = append(operationStates, operationState)
+	}
+
+	if err := errors.Join(errs...); err != nil {
+		return nil, err
+	}
+	if len(operationStates) == 0 {
+		return nil, errors.New("no operation states")
+	}
+	slices.SortStableFunc(operationStates, compareOperationState)
+	if operationStates[0] == nil {
+		return nil, errors.New("nil operation state")
+	}
+	logger.Info("determined node pool update operation status", "operationStates", operationStates)
+	picked, err := pickWorstOperationState(operationStates)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+	logger.Info("picked node pool update operation status", "provisioningState", picked.provisioningState, "message", picked.message)
+	return picked, nil
+}
+
+func (c *operationNodePoolUpdate) desiredVersionResolutionOperationState(ctx context.Context, operation *api.Operation) (*operationState, error) {
+	existingNodePool, err := c.resourcesDBClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).
+		NodePools(operation.ExternalID.Parent.Name).Get(ctx, operation.ExternalID.Name)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+
+	existingServiceProviderNodePool, err := c.resourcesDBClient.ServiceProviderNodePools(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName, operation.ExternalID.Parent.Name, operation.ExternalID.Name).Get(ctx, api.ServiceProviderNodePoolResourceName)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+
+	resultingDesiredVersion := existingServiceProviderNodePool.Spec.NodePoolVersion.DesiredVersion
+	if resultingDesiredVersion == nil {
+		return nil, utils.TrackError(fmt.Errorf("service provider node pool has no desired version"))
+	}
+
+	customerDesiredVersion := semver.MustParse(existingNodePool.Properties.Version.ID)
+
+	operationID := strings.ToLower(operation.ResourceID.String())
+	// If the operation is cancelled, its desiredVersionMismatchFirstSeen entry is never
+	// explicitly removed. This is safe because operation.ResourceID is unique per operation,
+	// so stale entries won't cause false matches for newer operations and will eventually
+	// be evicted by the LRU.
+	if customerDesiredVersion.EQ(*resultingDesiredVersion) {
+		c.desiredVersionMismatchFirstSeen.Remove(operationID)
+		return newOperationState(arm.ProvisioningStateSucceeded, ""), nil
+	}
+
+	nodePoolKey := controllerutils.HCPNodePoolKey{
+		SubscriptionID:    operation.ExternalID.SubscriptionID,
+		ResourceGroupName: operation.ExternalID.ResourceGroupName,
+		HCPClusterName:    operation.ExternalID.Parent.Name,
+		HCPNodePoolName:   operation.ExternalID.Name,
+	}
+
+	controllerCRUD := c.resourcesDBClient.HCPClusters(nodePoolKey.SubscriptionID, nodePoolKey.ResourceGroupName).NodePools(nodePoolKey.HCPClusterName).Controllers(nodePoolKey.HCPNodePoolName)
+	controllerDoc, getControllerErr := controllerCRUD.Get(ctx, upgradecontrollers.NodepoolVersionControllerName)
+	if getControllerErr != nil {
+		return nil, utils.TrackError(getControllerErr)
+	}
+
+	intentFailedCondition := apimeta.FindStatusCondition(controllerDoc.Status.Conditions, api.ControllerConditionTypeIntentFailed)
+
+	if intentFailedCondition == nil {
+		return newOperationState(arm.ProvisioningStateAccepted, "customer desired version not yet calculated"), nil
+	}
+	// Customer desired version differs from the service provider resolved version, and the
+	// NodePoolVersion controller has not yet set IntentFailed (VersionUpgradeNotAccepted)
+	// for this version. Stay Accepted while resolution runs; fail once elapsed exceeds
+	// 59s from the first time this process observed the mismatch for this operation.
+	// This avoids immediately failing long-running operations after controller restarts
+	// and is double the relistDuration of the nodepool and serviceProviderNodePool informers.
+	// This will not solve all the edge cases, but it will give enough time to the other controllers to act.
+	if intentFailedCondition.Status != metav1.ConditionTrue || intentFailedCondition.Reason != api.VersionUpgradeNotAcceptedReason {
+		pending := newOperationState(arm.ProvisioningStateAccepted, "customer desired version does not match resolved desired version")
+		firstSeen, ok := c.desiredVersionMismatchFirstSeen.Get(operationID)
+		if !ok {
+			c.desiredVersionMismatchFirstSeen.Add(operationID, c.clock.Now())
+			return pending, nil
+		}
+		if c.clock.Since(firstSeen.(time.Time)) <= 59*time.Second {
+			return pending, nil
+		}
+		msg := fmt.Sprintf(
+			"timed out after 59s waiting for resolution of desired version from '%s' node pool version",
+			existingNodePool.Properties.Version.ID,
+		)
+		c.desiredVersionMismatchFirstSeen.Remove(operationID)
+		return newOperationState(arm.ProvisioningStateFailed, msg), nil
+	}
+	c.desiredVersionMismatchFirstSeen.Remove(operationID)
+	return newOperationState(arm.ProvisioningStateFailed, intentFailedCondition.Message), nil
+}
+
+func (c *operationNodePoolUpdate) nodePoolServiceUpdateOperationState(ctx context.Context, operation *api.Operation) (*operationState, error) {
+	logger := utils.LoggerFromContext(ctx)
+	nodePoolStatus, err := c.clusterServiceClient.GetNodePoolStatus(ctx, operation.InternalID)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+	newOperationStatus, opError, err := convertNodePoolStatus(operation, nodePoolStatus)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+	logger.Info("new status via cluster-service", "newStatus", newOperationStatus, "newOperationError", opError)
+	msg := ""
+	if opError != nil {
+		msg = opError.Message
+	}
+	return newOperationState(newOperationStatus, msg), nil
 }
--- a/backend/pkg/controllers/operationcontrollers/operation_node_pool_update_test.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_node_pool_update_test.go
@@ -17,16 +17,22 @@ package operationcontrollers
 import (
 	"context"
 	"testing"
+	"time"
 
+	"github.com/blang/semver/v4"
 	"github.com/go-logr/logr/testr"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 
-	utilsclock "k8s.io/utils/clock"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clocktesting "k8s.io/utils/clock/testing"
+	"k8s.io/utils/lru"
+	"k8s.io/utils/ptr"
 
 	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
 
+	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/database"
 	"github.com/Azure/ARO-HCP/internal/databasetesting"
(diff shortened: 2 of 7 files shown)


backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -100,6 +113,158 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+		return nil, utils.TrackError(getControllerErr)
+	}
+
+	intentFailedCondition := apimeta.FindStatusCondition(controllerDoc.Status.Conditions, api.ControllerConditionTypeIntentFailed)
+	// Only respect IntentFailed if it's for the current customer desired version.
+	// If the customer requested a different version since the IntentFailed was set, ignore it.
+	// The condition message is formatted as: "[version X.Y.Z] <error details>"
+	// This allows version-specific condition scoping without modifying the API type.
+	// TODO(PR review): Consider adding a new field to ControllerStatus (LastObservedCustomerDesiredVersion) for cleaner version tracking.
+	isRelevantIntentFailed := intentFailedCondition != nil &&
```
ahitacat: Instead of adding this prefix and calculate we could have add it in the controllerStatus. However, as this controllerStatus is shared among cluster and nodepool I didn't want to add conditions that weren't shared.

To the reviewers: Is this approach correct? or should I opt for adding a new field?

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -16,24 +16,35 @@ package operationcontrollers
...
 
 type operationNodePoolUpdate struct {
-	resourcesDBClient    database.ResourcesDBClient
-	clusterServiceClient ocm.ClusterServiceClientSpec
-	notificationClient   *http.Client
+	resourcesDBClient               database.ResourcesDBClient
+	clusterServiceClient            ocm.ClusterServiceClientSpec
+	notificationClient              *http.Client
+	clock                           clock.PassiveClock
+	desiredVersionMismatchFirstSeen *lru.Cache
```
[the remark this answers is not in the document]
ahitacat: Similar comment in the controller PR https://github.com/Azure/ARO-HCP/pull/5042#discussion_r3182851260

This lru.cache is thread safe. https://github.com/kubernetes/utils/blob/master/lru/lru.go#L28

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -100,6 +113,158 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+		ResourceGroupName: operation.ExternalID.ResourceGroupName,
+		HCPClusterName:    operation.ExternalID.Parent.Name,
+		HCPNodePoolName:   operation.ExternalID.Name,
+	}
+
+	controllerDoc, getControllerErr := controllerutils.GetOrCreateController(
+		ctx,
+		c.resourcesDBClient,
+		operation.ExternalID,
+		"NodePoolVersion",
```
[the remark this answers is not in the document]
ahitacat: I will keep it as this as is aligned with the rest of operation controllers. But the change can be a followup.

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
JameelB: > should I opt for adding a new field?

do you mean adding a `LastObservedCustomerDesiredVersion` to Controller.Status that you mention in the comment above? Is the Controller.Status shared with anything else apart from cluster/node pool? 

My personal preference would be using `LastObservedCustomerDesiredVersion` since it could be useful for the cluster too? If we can't change the api though, i'm happy with what's proposed here since the prefix format is predictable.

ahitacat: > do you mean adding a LastObservedCustomerDesiredVersion to Controller.Status that you mention in the comment above?

Yes, I mean that. The controller.Status is not shared yet, but there are no many controllers and I thought it can be potentially shared.
I opted here to no add that into an object I don't know how it will be used. Although I didn't like to parse a message to get this information.

@miguelsorianod do you have any opinion about this?

ahitacat: After reviewing this we opted to not based this in the message prefix wait at least 2 runs of the informers used by the controller in case of the condition is nil or intentFailed. This should be enough for the case I was trying to solve.
I'm testing this manually.

miguelsorianod: As discussed yesterday, let's do the following approach:

When IntentFailed is set to false, we ignore it until at least there's been double the nodepool/serviceprovidernodepool informer relist period time.

This would cover most of the cases. If it's not enough we could consider increasing the time.
The time increase should only impact when there's an upgrade, and not during all update operations, as well as the specific case where the intent failed within the upgrade.
The increased experience time occurs when there's a failure so experiencing an increased time there is less impactful.

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -100,6 +113,149 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+	}
+
+	intentFailedCondition := apimeta.FindStatusCondition(controllerDoc.Status.Conditions, api.ControllerConditionTypeIntentFailed)
+
+	// Customer desired minor differs from the service provider resolved version, and the
+	// NodePoolVersion controller has not yet set IntentFailed (VersionUpgradeNotAccepted).
+	// Stay Accepted while resolution runs; fail once elapsed exceeds 59s from the first
+	// time this process observed the mismatch for this operation, so a
+	// controller restart does not immediately fail long-running operations.
+	if intentFailedCondition == nil || intentFailedCondition.Status != metav1.ConditionTrue || intentFailedCondition.Reason != api.VersionUpgradeNotAcceptedReason {
```
miguelsorianod: As we discussed, when there's no condition (nil) this would mark the operation as failed which shouldn't be the case.

ahitacat: Indeed, I forgot about this. I have added a check when it is nil to keep the operation Accepted until it is resolved.

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -100,6 +113,153 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+		// Customer desired version differs from resolved version, and NodePoolVersion controller
+		// has not yet set IntentFailed for this version. Stay Accepted while resolution runs; fail once
+		// elapsed exceeds 59s from first observation. Double of the releaseDuration of the nodepool and serviceProviderNodePool informers
+		pending := newOperationState(arm.ProvisioningStateAccepted, "customer desired version does not match resolved desired version")
+		firstSeen, ok := c.desiredVersionMismatchFirstSeen.Get(operation.ResourceID.String())
+		if !ok {
+			c.desiredVersionMismatchFirstSeen.Add(operation.ResourceID.String(), c.clock.Now())
+			return pending, nil
+		}
+		if c.clock.Since(firstSeen.(time.Time)) <= 59*time.Second {
```
JameelB, line 240: out of curiosity, why the change from 29 to 59s?

ahitacat: so we have 2 runs of the informers. Each informer relist every 30 secs.

https://github.com/Azure/ARO-HCP/pull/5378#discussion_r3309727488

JameelB: ah i see this is double the time of the informer relist period, ignore this.

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: I manual tested this. Created a node pool with version 4.20.8, upgrade it to 4.20.9 that it doesn't exists in Cincinnati. The controller failed, the intentFailed condition was set and the operation was marked as failed.

```
{
    "id": "/providers/Microsoft.RedHat
OpenShift/locations/westus3/hcpOperationStatuses/dff9145a-4559-4306-a64c-cf3727cf254b",
    "name": "dff9145a-4559-4306-a64c-cf3727cf254b",
    "status": "Failed",
    "startTime": "2026-05-27T11:03:16.203534595Z",
    "endTime": "2026-05-27T11:03:29.915339813Z",
    "error": {
        "message": "version 4.20.9 not found in Cincinnati channel stable-4.20"
    }
}
```

Then upgrade it to 4.20.15 and the operation succeeded.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
```diff
@@ -200,22 +210,53 @@ func (c *nodePoolVersionSyncer) SyncOnce(ctx context.Context, key controllerutil
 	if nodePool.Properties.Version.ID == "" {
 		return nil
 	}
-	customerDesiredVersion := semver.MustParse(nodePool.Properties.Version.ID)
+	customerDesiredVersion, err := semver.Parse(nodePool.Properties.Version.ID)
```
miguelsorianod: Is there some reason this has been changed from MustParse to Parse?

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -104,6 +114,150 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
 	if !c.ShouldProcess(ctx, operation) {
 		return nil // no work to do
 	}
+	if len(operation.InternalID.String()) == 0 {
```
miguelsorianod, line 118: As a relevant note, when we move update triggering from frontend to backend the operation won't have an InternalID anymore, so this will need to be refactored

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -104,6 +114,150 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+
+	resultingDesiredVersion := existingServiceProviderNodePool.Spec.NodePoolVersion.DesiredVersion
+	if resultingDesiredVersion == nil {
+		return nil, utils.TrackError(fmt.Errorf("service provider node pool has no desired version"))
+	}
+
+	customerDesiredVersion := semver.MustParse(existingNodePool.Properties.Version.ID)
+
+	if customerDesiredVersion.EQ(*resultingDesiredVersion) {
+		c.desiredVersionMismatchFirstSeen.Remove(operation.ResourceID.String())
```
miguelsorianod: What happens when the entry in the cache if a nodepool delete occurs?

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -104,6 +114,150 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+
+	if intentFailedCondition == nil {
+		return newOperationState(arm.ProvisioningStateAccepted, "customer desired version not yet calculated"), nil
+	}
+	// Customer desired version differs from the service provider resolved version, and the
+	// NodePoolVersion controller has not yet set IntentFailed (VersionUpgradeNotAccepted)
+	// for this version. Stay Accepted while resolution runs; fail once elapsed exceeds
+	// 59s from the first time this process observed the mismatch for this operation.
+	// This avoids immediately failing long-running operations after controller restarts
+	// and is double the releaseDuration of the nodepool and serviceProviderNodePool informers.
```
miguelsorianod: what is "releaseDuration" ?

miguelsorianod: We should document a bit better that it doesn't fully cover all cases but must of them by having double the value to give time to the other controller to act

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -104,6 +114,150 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+	if intentFailedCondition.Status != metav1.ConditionTrue || intentFailedCondition.Reason != api.VersionUpgradeNotAcceptedReason {
```
miguelsorianod, line 233: Does this cover "Unknown" condition status

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -104,6 +114,150 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+		pending := newOperationState(arm.ProvisioningStateAccepted, "customer desired version does not match resolved desired version")
```
miguelsorianod, line 234: How does this match with other update changes non related to the version changes?

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -104,6 +114,150 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+		}
+		if c.clock.Since(firstSeen.(time.Time)) <= 59*time.Second {
+			return pending, nil
+		}
+		msg := fmt.Sprintf(
+			"timed out after 59s waiting for resolution of desired version from '%s' node pool version",
+			existingNodePool.Properties.Version.ID,
+		)
+		c.desiredVersionMismatchFirstSeen.Remove(operation.ResourceID.String())
+		return newOperationState(arm.ProvisioningStateFailed, msg), nil
```
miguelsorianod, line 248: What happens if we fail to set the new operation state and we've removed from the cache the value?

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -104,6 +114,150 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+	}
+	c.desiredVersionMismatchFirstSeen.Remove(operation.ResourceID.String())
+	return newOperationState(arm.ProvisioningStateFailed, intentFailedCondition.Message), nil
```
miguelsorianod, line 251: What happens if we fail to set the new operation state and we've removed from the cache the value?

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: Yes, that is what the comment below describes. It is similar to the operation_cluster_update.

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: Do you mean if a nodepool is being delete at the same time as we are calculating this?
This cache will continue to have the key, the new `desiredVersion` will be calculating here and eventually it will succeed (most of the times as delete a nodepool take longer than 59s)

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: a typo, should be relistDuration

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: Yes

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: It depends on the status of the other operation.
```
// provisioningStatePriority is a logical merge order that decides what the most important state to return is.
// For instance, if one check is succeeded, one is failed, and one is accepted, then failed is the most
// reasonable state for the operation.
var provisioningStatePriority = map[arm.ProvisioningState]int{
	"":                                  -1, // causes an error
	arm.ProvisioningStateFailed:         0,
	arm.ProvisioningStateCanceled:       10,
	arm.ProvisioningStateDeleting:       20,
	arm.ProvisioningStateProvisioning:   30,
	arm.ProvisioningStateAwaitingSecret: 35,
	arm.ProvisioningStateUpdating:       40,
	arm.ProvisioningStateAccepted:       50,
	arm.ProvisioningStateSucceeded:      100,
}
```
In the worst case (nodepool has a mismatch in the calculation of the desiredVersion and the intentFailed is unkown or false) and the update operation has succeed. It will take at least 59s (since this was first run) to suceed. Meaning, that operation will be in Accepted state.

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: It will run again this code, desiredVersionMismatchFirstSeen[key] will be nil, and it will add it and check again.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
ahitacat: I was considering what copilot proposed. https://github.com/Azure/ARO-HCP/pull/5378/changes#r3333670941

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
miguelsorianod: I don't think this will succeed because nodepool delete will cancel active operations

ahitacat: True, if the current operation gets delete. This operation will be marked as canceled, and we will not have second run of this controller for the operation.

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: done

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: In this case as the status of Failed has more priority than the cancel state, it will mark this operation as failed.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller_test.go
```diff
@@ -779,438 +978,6 @@ func TestNodePoolVersionSyncer_ValidateDesiredNodePoolVersion(t *testing.T) {
...
-	}
-
-	ctx = utils.ContextWithLogger(ctx, logr.Discard())
-	err := syncer.SyncOnce(ctx, testKey)
-
-	require.Error(t, err)
-	assert.Contains(t, err.Error(), "not found in Cincinnati")
-}
-
-func TestNodePoolVersionSyncer_SyncOnce_DowngradeVersionNotInCincinnatiFails(t *testing.T) {
```
ahitacat, line 1023: I have not added this one in the table driven tests `TestNodePoolVersionSyncer_SyncOnce_IntentFailed` as it is covered with the VersionNotInCincinnati test. I have added a failure when the downgrade is outsied the skew < cp.MinorVersion -2

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
```diff
@@ -104,6 +114,156 @@ func (c *operationNodePoolUpdate) SynchronizeOperation(ctx context.Context, key
...
+
+	nodePoolKey := controllerutils.HCPNodePoolKey{
+		SubscriptionID:    operation.ExternalID.SubscriptionID,
+		ResourceGroupName: operation.ExternalID.ResourceGroupName,
+		HCPClusterName:    operation.ExternalID.Parent.Name,
+		HCPNodePoolName:   operation.ExternalID.Name,
+	}
+
+	controllerCRUD := c.resourcesDBClient.HCPClusters(nodePoolKey.SubscriptionID, nodePoolKey.ResourceGroupName).NodePools(nodePoolKey.HCPClusterName).Controllers(nodePoolKey.HCPNodePoolName)
+	controllerDoc, getControllerErr := controllerCRUD.Get(ctx, "NodePoolVersion")
```
miguelsorianod: It would be safer if we define it in a constant because if the name changes it will stop matching

backend/pkg/controllers/operationcontrollers/operation_node_pool_update_test.go
```diff
@@ -35,12 +44,18 @@ import (
...
-		verify        func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient, fixture *nodePoolTestFixture)
+		name                                  string
+		nodePoolState                         string
+		nodePoolMsg                           string
+		customerVersionID                     string
+		serviceProviderNodePoolDesiredVersion *semver.Version
+		nodePoolVersionControllerConditions   []metav1.Condition
+		seedMismatchFirstSeenAt               time.Time
+		expectError                           bool
+		verify                                func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient, fixture *nodePoolTestFixture)
```
miguelsorianod: nitpick: name verifyDB, as the checks should be restricted to DB related checks

backend/pkg/controllers/operationcontrollers/operation_node_pool_update_test.go
```diff
@@ -107,23 +122,184 @@ func TestOperationNodePoolUpdate_SynchronizeOperation(t *testing.T) {
...
 			ctx = utils.ContextWithLogger(ctx, testr.New(t))
 			ctrl := gomock.NewController(t)
-			defer ctrl.Finish()
 
 			fixture := newNodePoolTestFixture()
 			cluster := fixture.newCluster()
 			nodePool := fixture.newNodePool()
+
+			// Set node pool version
+			var desiredVersion *semver.Version
```
miguelsorianod: The test design can be improved, there seems to be several workarounds which will make the tests hard to scale when new needs arise.

Have the test case struct have:
* serviceProviderNodePool attribute
* nodePool attribute
* nodePoolVersionController attribute

Then the test body will setup the DB mock adding those as resources.

This avoid having the workarounds of customerVersionID, serviceProviderNodePoolDesiredVersion,  nodePoolVersionControllerConditions with the if-else and hardcoded values.

If you want avoid duplication with the new approach you can also rely on helpers, fixtures, ...

backend/pkg/controllers/operationcontrollers/operation_node_pool_update_test.go
```diff
@@ -35,12 +44,18 @@ import (
...
-		nodePoolMsg   string
-		expectError   bool
-		verify        func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient, fixture *nodePoolTestFixture)
+		name                                  string
+		nodePoolState                         string
+		nodePoolMsg                           string
+		customerVersionID                     string
+		serviceProviderNodePoolDesiredVersion *semver.Version
+		nodePoolVersionControllerConditions   []metav1.Condition
+		seedMismatchFirstSeenAt               time.Time
```
miguelsorianod: nitpick: be a bit more specific about the name so the reader understands what is it about

backend/pkg/controllers/operationcontrollers/operation_node_pool_update_test.go
```diff
@@ -35,12 +44,18 @@ import (
...
+	t.Parallel()
+	testClockNow := mustParseTime("2024-06-01T12:00:00Z")
 	tests := []struct {
-		name          string
-		nodePoolState string
-		nodePoolMsg   string
-		expectError   bool
-		verify        func(t *testing.T, ctx context.Context, db *databasetesting.MockResourcesDBClient, fixture *nodePoolTestFixture)
+		name                                  string
+		nodePoolState                         string
```
miguelsorianod: To avoid having nodePoolState, nodePoolMsg attributes, which are related to CS (not evident to the reader with the current names), we can have instead a setupCSMock that sets the desired mock state.

backend/pkg/controllers/upgradecontrollers/nodepool_version_controller_test.go
```diff
@@ -345,14 +349,221 @@ func TestNodePoolVersionSyncer_SyncOnce(t *testing.T) {
...
+			if tt.wantSyncErr {
+				require.Error(t, err)
+				assert.ErrorContains(t, err, tt.wantErrContains)
+			} else {
+				require.NoError(t, err)
+			}
+
+			serviceProviderNodePool, getServiceProviderNodePoolErr := mockResourcesDBClient.ServiceProviderNodePools(
+				testSubscriptionID, testResourceGroupName, testClusterName, testNodePoolName,
+			).Get(ctx, api.ServiceProviderNodePoolResourceName)
```
miguelsorianod: Instead of all these checks and conditions, would having a verifyDB function encapsulate it better? there could be function helpers within the test function too if the same code is repeated

backend/pkg/controllers/operationcontrollers/operation_node_pool_update.go
ahitacat: I will do the same for the cluster similar controller.

PR #5390, opened 2026-05-26
Mitigate Prometheus Agent WAL replay OOM crash loop (AROSLSRE-948)

janboll:
https://redhat.atlassian.net/browse/AROSLSRE-948

### What/Why

- Tune remote_write queue: maxSamplesPerSend 2000→10000, capacity 2500→50000, maxShards 500→50

- Remove CPU limits for Prometheus to allow burst during replay

### DIFF
--- a/observability/prometheus/deploy/templates/prometheus.yaml
+++ b/observability/prometheus/deploy/templates/prometheus.yaml
@@ -1,6 +1,6 @@
 {{- $none := "NONE" -}}
 {{- $setRequests := or (ne (.Values.prometheusSpec.resources.requests.cpu | toString) $none) (ne (.Values.prometheusSpec.resources.requests.memory | toString) $none) -}}
-{{- $setLimits := or (ne (.Values.prometheusSpec.resources.limits.cpu | toString) $none) (ne (.Values.prometheusSpec.resources.limits.memory | toString) $none) -}}
+{{- $setLimits := ne (.Values.prometheusSpec.resources.limits.memory | toString) $none -}}
 apiVersion: monitoring.coreos.com/v1alpha1
 kind: PrometheusAgent
 metadata:
@@ -22,9 +22,6 @@ spec:
     {{- end }}
     {{- if $setLimits }}
     limits:
-      {{- if ne (.Values.prometheusSpec.resources.limits.cpu | toString) $none }}
-      cpu: '{{ .Values.prometheusSpec.resources.limits.cpu }}'
-      {{- end }}
       {{- if ne (.Values.prometheusSpec.resources.limits.memory | toString) $none }}
       memory: '{{ .Values.prometheusSpec.resources.limits.memory }}'
       {{- end }}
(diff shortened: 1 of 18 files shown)


observability/prometheus/deploy/templates/prometheus.yaml
```diff
@@ -100,6 +97,19 @@ spec:
     maxSkew: 1
     topologyKey: kubernetes.io/hostname
     whenUnsatisfiable: ScheduleAnyway
+  {{- if .Values.prometheusSpec.purgeWAL }}
+  initContainers:
+  - name: purge-wal
+    image: '{{ .Values.prometheusSpec.image.registry }}/{{ .Values.prometheusSpec.image.repository }}@sha256:{{ .Values.prometheusSpec.image.sha }}'
+    command: ["sh", "-c", "rm -rf /prometheus/wal/* && echo 'WAL purged'"]
```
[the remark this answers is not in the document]
janboll: fixed

observability/prometheus/deploy/templates/prometheus.yaml
```diff
@@ -100,6 +97,19 @@ spec:
     maxSkew: 1
     topologyKey: kubernetes.io/hostname
     whenUnsatisfiable: ScheduleAnyway
+  {{- if .Values.prometheusSpec.purgeWAL }}
+  initContainers:
+  - name: purge-wal
+    image: '{{ .Values.prometheusSpec.image.registry }}/{{ .Values.prometheusSpec.image.repository }}@sha256:{{ .Values.prometheusSpec.image.sha }}'
+    command: ["sh", "-c", "rm -rf /prometheus/wal && mkdir -p /prometheus/wal && echo 'WAL purged'"]
```
hbhushan3: When this toggle is true, any node drain, image bump, or rolling restart silently drops all buffered samples. Was a one-shot Job or manual procedure considered instead of a permanent StatefulSet init container?

observability/prometheus/deploy/templates/prometheus.yaml
```diff
@@ -1,6 +1,6 @@
 {{- $none := "NONE" -}}
 {{- $setRequests := or (ne (.Values.prometheusSpec.resources.requests.cpu | toString) $none) (ne (.Values.prometheusSpec.resources.requests.memory | toString) $none) -}}
-{{- $setLimits := or (ne (.Values.prometheusSpec.resources.limits.cpu | toString) $none) (ne (.Values.prometheusSpec.resources.limits.memory | toString) $none) -}}
+{{- $setLimits := ne (.Values.prometheusSpec.resources.limits.memory | toString) $none -}}
```
hbhushan3, line 3: Why are we removing cpu limit? Mgmt-plane Prometheus now shares nodes with HCP/CAPI controllers with no CPU ceiling. Is there a PriorityClass or dedicated nodepool bounding the blast radius during replay, or is the assumption that memory pressure will hit first?

janboll: No, it's running a single shard on the infra nodepools, where no customer pods are running there

observability/prometheus/deploy/templates/prometheus.yaml
janboll: The intended use for this agent setup is, we never need the WAL, cause it's supposed to ship metrics right away. The toggle should only be true when we see the issue to mitigate without jit, and then set it to false again after 'fixing'

janboll: removed the container

PR #5398, opened 2026-05-26
Add InstanceVersion for our persisted types

deads2k:
This version is incremented every time we mutate the value. This allows us to compare two instances from storage and know which one is more recent.  This will allow us to use and a changefeed and a relist to determine whether an item is newer and should be placed into the cache.


deads2k:
this looks pretty good.

mbarnes (approved):
For the last commit:
/lgtm

I don't know of a compelling reason to go this route at this time, but I'll mention it FYI.

This looks like it could be a good fit for [Cosmos DB pre-triggers](https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-write-stored-procedures-triggers-udfs?tabs=javascript#triggers), which is a snippet of JavaScript code that you can register through the Cosmos SDK to execute before specific operations like create or replace.  (There's also post-triggers which run after.)

I know ARO Classic uses Cosmos DB triggers for things like [lease management and setting bililng timestamps](https://github.com/Azure/ARO-RP/blob/d2d3e2b6abdaa4ea6f32456c904c513c0624580d/pkg/deploy/assets/rp-production.json#L1032-L1107).  Since all Cosmos DB access in ARO-HCP is going through the `database` Go module, I'm not sure the advantage would be significant enough to bother with... other than saving a bit of logic in present and future CRUD helpers.

Just a Cosmos DB feature to be aware of.

deads2k:
/hold
re labelling after rebase.
/label lgtm



cannot go out until we're into prod at least once more.

deads2k:
/hold cancel

simple rebase since @mbarnes lgtmed before

/label lgtm

PR #5406, opened 2026-05-27
feat: remove clusters service cluster create from frontend

JakobGray:
https://redhat.atlassian.net/browse/ARO-24824

### What

Removes CS cluster creation from frontend so it can be handled asynchronously in the backend. Also remove flags that are not used.

### Why

Removing the call to CS will improve the response time of the create request.

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/frontend/cmd/cmd.go
+++ b/frontend/cmd/cmd.go
@@ -53,11 +53,8 @@ type FrontendOpts struct {
 	auditLogQueueSize  int
 	auditConnectSocket bool
 
-	clustersServiceURL            string
-	clusterServiceProvisionShard  string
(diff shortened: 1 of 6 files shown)


machi1990:
Suggestion:

This PR can include the changes in https://github.com/Azure/ARO-HCP/pull/4821 and be used as the means to test and validate that the changes in https://github.com/Azure/ARO-HCP/pull/4821 are working as intended when we reach steady state.

The order of merge and production reachness remains the same:
- https://github.com/Azure/ARO-HCP/pull/4821 goes to prod first
- This PR is then rebased on top of it and merged as well

PR #5408, opened 2026-05-27
Revert " Cilium Connectivity Check Verifier into E2E tests using Cilium CNI plugin"

janboll:
https://redhat.atlassian.net/browse/ARO-27342

Reverts Azure/ARO-HCP#4812

PR #5407, opened 2026-05-27
create echo-a echo-b first to sove Race condition

janboll:
https://redhat.atlassian.net/browse/ARO-27342

### PR Checklist
- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
- [ ] Screenshots included (if graph/UI/metrics changes)
- [ ] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [ ] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

PR #5422, opened 2026-05-28
Revert fix cilium connect

janboll:
https://redhat.atlassian.net/browse/ARO-27342

### What

Bring back changes revert in #5408 and add fix to it from #5407.

The fix updates the Cilium connectivity check verifier to reduce a scheduler race by deploying the echo-a/echo-b Deployments first, waiting for their pods to be scheduled, and then creating the remaining connectivity-check resources.

### DIFF
--- a/test/util/verifiers/cilium.go
+++ b/test/util/verifiers/cilium.go
@@ -276,6 +276,23 @@ func (v verifyCiliumConnectivityChecks) Verify(ctx context.Context, adminRESTCon
 			return scheduled >= 2, nil
 		})
 		if waitErr != nil {
+			events, eventsErr := kubeClient.CoreV1().Events(namespaceName).List(ctx, metav1.ListOptions{})
+			if eventsErr != nil {
+				logger.Error(eventsErr, "failed to list events for debugging", "namespace", namespaceName)
+			} else {
+				logger.Info("listing events for debugging echo scheduling failure", "namespace", namespaceName, "eventCount", len(events.Items))
+				for _, event := range events.Items {
+					logger.Info("event",
+						"type", event.Type,
+						"reason", event.Reason,
+						"message", event.Message,
+						"object", fmt.Sprintf("%s/%s", event.InvolvedObject.Kind, event.InvolvedObject.Name),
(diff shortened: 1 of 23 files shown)


test/util/verifiers/cilium.go
```diff
@@ -120,3 +130,304 @@ func (v verifyCiliumOperational) Verify(ctx context.Context, adminRESTConfig *re
...
+			for _, pod := range pods.Items {
+				if pod.Spec.NodeName != "" {
+					scheduled++
+				}
+			}
+			return scheduled >= 2, nil
+		})
+		if waitErr != nil {
+			return fmt.Errorf("echo-a/echo-b pods were not scheduled in time: %w", waitErr)
+		}
```
mbukatov, line 297: We need to at least dump all the events from the test namespace in case of any scheduling problem. Otherwise it will be impossible to diagnose failures.

mbukatov (commented):
/hold

Asking to improve error reporting.

test/util/verifiers/cilium.go
```diff
@@ -120,3 +130,304 @@ func (v verifyCiliumOperational) Verify(ctx context.Context, adminRESTConfig *re
...
+					"type", event.Type,
+					"reason", event.Reason,
+					"message", event.Message,
+					"object", fmt.Sprintf("%s/%s", event.InvolvedObject.Kind, event.InvolvedObject.Name),
+					"count", event.Count,
+					"firstTimestamp", event.FirstTimestamp,
+					"lastTimestamp", event.LastTimestamp,
+				)
+			}
+		}
```
mbukatov, line 408: Here is where we log the events in the end of the test. This is what we need in case of "echo-a/echo-b pods were not scheduled in time" failure as well. Either just copy the code there (it's a simple loop over events) or move it into a function and call it on both places.

test/util/verifiers/cilium.go
mbukatov: Addressed in ba315ffe34076724570ecfdb989b91d8e5c63637

PR #5426, opened 2026-05-28
Fix kusto exporter log freshness

janboll:
https://redhat.atlassian.net/browse/AROSLSRE-962

### What/Why

Fix kusto exporter log freshness: sort before limit, make order configurable

  The exporter checks log freshness by querying Kusto and inspecting the
  first returned row's timestamp. Previously, queries applied `limit`
  before `order by`, returning an arbitrary subset that was then sorted —
  so the exporter was measuring the age of a random log, not the most
  recent one.

  - Add OrderBy type (string enum: asc/desc) to QueryOptions with a NewQueryOptions() constructor that defaults to asc
  - Templatize `| order by timestamp` direction in all 10 KQL templates
  - Move `| order by` before `| limit` so sorting happens before truncation
  - Set OrderBy to desc in the exporter so it sees the newest log first
  - Replace prettylog with slog.TextHandler for plain text CLI logging
  - Migrate all QueryOptions call sites to use NewQueryOptions()

### DIFF
--- a/test/cmd/aro-hcp-tests/custom-link-tools/options.go
+++ b/test/cmd/aro-hcp-tests/custom-link-tools/options.go
@@ -307,12 +307,11 @@ func (o Options) Run(ctx context.Context) error {
 			if err != nil {
 				return utils.TrackError(fmt.Errorf("failed to create query factory: %w", err))
 			}
-			queryOpts := kusto.QueryOptions{
-				ResourceGroupName: rg,
-				TimestampMin:      ti.StartTime,
-				TimestampMax:      ti.EndTime,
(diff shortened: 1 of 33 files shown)


raelga:
/hold

Let's try to clean the queue for https://github.com/Azure/ARO-HCP/pull/5457

raelga:
https://github.com/Azure/ARO-HCP/pull/5457 merged.

/unhold
/retest

raelga (approved):
/lgtm
/approve

raelga (approved):
/lgtm
/approve

PR #5429, opened 2026-05-28
prepare internal api attributes related to async nodepool deletion trigger in backend

miguelsorianod:
This PR adds the DeletionTimestamp and ClusterServiceDeletionTimestamp in the NodePool internal API type:

DeletionTimestamp is the the timestamp at which the NodePool deletion
was requested. A nil value indicates that the NodePool deletion has not been requested.

ClusterServiceDeletionTimestamp is written when a dispatch of a Cluster
Service Delete NodePool request against Cluster Service for this node
pool has been handled. It is set after a successful DeleteNodePool call
to Cluster Service, but also when it's determined that no delete call is
needed but we consider we should behave as if the delete call was successfully
issued (for example, if the parent cluster of the nodepool is already being
uninstalled, because cluster-service will already take care of deleting the
nodepool as part of the cluster teardown).
A nil value indicates that the Cluster Service Deletion has not been requested.

These two attributes are not used on the backend side.

This PR also adds the UsesNewNodePoolDeletionApproach attribute for both NodePool and Operation internal api types. This will be used as a gate when we decide to switch nodepool deletion to be with async deletion trigger in backend.


internal/api/types_nodepool.go
```diff
@@ -56,6 +57,29 @@ type HCPOpenShiftClusterNodePoolProperties struct {
...
+	// The timestamp is in UTC.
+	// TODO this attribute is not in use yet. Do not rely on it.
+	ClusterServiceDeletionTimestamp *metav1.Time `json:"clusterServiceDeletionTimestamp,omitempty"`
+
+	// TODO Temporary field to track whether the node pool operation is using the new deletion approach.
+	// We are migrating from the node pool cs deletion synchronous in frontend to the backend, to be fully asynchronous
+	// This boolean is true for NodePool delete operations that are created with new deletion approach.
+	// This will be removed once all nodepools whose deletion was triggered before the new approach is fully rolled out have been
+	// fully deleted in all ARO-HCP permanent environments, for all regions.
+	UsesNewNodePoolDeletionApproach bool `json:"usesNewNodePoolDeletionApproach"`
```
miguelsorianod, line 82: This one does not have `omitempty`, intentionally. In that way we can see it explicitly too (using Kusto etc.)

internal/api/types_operation.go
```diff
@@ -70,6 +70,13 @@ type Operation struct {
 	Status arm.ProvisioningState `json:"status,omitempty"`
 	// Error is an OData error, present when Status is "Failed" or "Canceled"
 	Error *arm.CloudErrorBody `json:"error,omitempty"`
+
+	// Temporary field to track whether the node pool operation is using the new deletion approach.
+	// We are migrating from the node pool cs deletion synchronous in frontend to the backend, to be fully asynchronous
+	// This boolean is true for NodePool delete operations that are created with new deletion approach.
+	// This will be removed once all nodepools whose deletion was triggered before the new approach is fully rolled out have been
+	// fully deleted in all ARO-HCP permanent environments, for all regions.
+	UsesNewNodePoolDeletionApproach bool `json:"usesNewNodePoolDeletionApproach"`
```
miguelsorianod, line 79: Similar to the nodepool one, it does not have omitempty, intentionally. In that way we can see it explicitly too (using Kusto etc.)

miguelsorianod, line 79: As another note here, this means it will be serialized for all operations, including non nodepool ones, as the operation type is shared for all resource types

frontend/pkg/frontend/node_pool.go
```diff
@@ -785,6 +786,9 @@ func (f *Frontend) addDeleteNodePoolToTransaction(ctx context.Context, writer ht
 		return utils.TrackError(err)
 	}
 
+	if nodePool.ServiceProviderProperties.DeletionTimestamp == nil {
```
miguelsorianod, line 789: This PR also starts setting the deletion timestamp in UTC to the nodepool internal api type when addDeleteNodePoolToTransaction is called, which occurs not only during nodepool deletion but also during cluster deletion

internal/api/types_nodepool.go
```diff
@@ -56,6 +57,29 @@ type HCPOpenShiftClusterNodePoolProperties struct {
...
+	// Service Delete NodePool request against Cluster Service for this node
+	// pool has been handled. It is set after a successful DeleteNodePool call
+	// to Cluster Service, but also when it's determined that no delete call is
+	// needed but we consider we should behave as if the delete call was successfully
+	// issued (for example, if the parent cluster of the nodepool is already being
+	// uninstalled, because cluster-service will already take care of deleting the
+	// nodepool as part of the cluster teardown).
+	// A nil value indicates that the Cluster Service Deletion has not been requested.
+	// The timestamp is in UTC.
+	// TODO this attribute is not in use yet. Do not rely on it.
```
miguelsorianod, line 74: Although we add ClusterServiceDeletionTimestamp in this PR, new code should not use it until the work to move nodepool deletion to async is completed. I added a TODO note too to try to make this clear. This also means another MR will be needed at some point to remove the TODO.

internal/api/types_nodepool.go
deads2k: good

internal/api/types_operation.go
deads2k: yeah, stuff happens.

miguelsorianod:
I verified I see an occurrence of a nodepool being deleted that has both the deletion timestamp in UTC as well as the usesNewNodePoolDeletionApproach boolean set to false:
```
...
			"serviceProviderProperties": {
				"clusterServiceID": "/api/aro_hcp/v1alpha1/clusters/2qj599h1t9j1bfdjegivpsv460miqku3/node_pools/np-1",
				"activeOperationId": "64af687f-ed54-4224-bd4b-683241055ea2",
				"deletionTimestamp": "2026-05-28T20:32:11.0000000Z",
				"usesNewNodePoolDeletionApproach": false
			},
...
```

I also made the following verifications in Kusto:
* There are no datadumps of nodepools in Deleting state that do not have deletiontimestamp.
* There are no datadumps of nodepools that do not contain usesNewNodePoolDeletionApproach. All of them are set to false
* There are no hcpOperationStatuses that do not contain usesNewNodePoolDeletionApproach. All of them are set to false
* There are no nodepools nor operations that currently contain ClusterServiceDeletionTimestamp

PR #5433, opened 2026-05-29
feat: temp best-effort preventing deletion of last nodepool in frontend

miguelsorianod:
We are removing the validation that prevents the last nodepool from being deleted on the Cluster Service side. This has been done because we want to start allowing the deletion of the last nodepool. Unfortunately, we have detected that it's currently not possible to delete the last NodePool (OCPBUGS-86702) but at the same time we need the removal on Cluster Service side because even in that case we need to allow the triggering to allow moving the deletion of nodepools to be triggered asynchronously an in an ordered delete process on backend which includes a cluster deletion that will trigger the deletion of nodepools too on backend. At the same time and until OCPBUGS-86702 is resolved we want to prevent the last nodepool from being deleted in most cases. This commit adds a best-effort check in the form of a nodepool admission validation on delete.

### DIFF
--- a/frontend/pkg/frontend/node_pool.go
+++ b/frontend/pkg/frontend/node_pool.go
@@ -707,6 +707,30 @@ func (f *Frontend) DeleteNodePool(writer http.ResponseWriter, request *http.Requ
 		return utils.TrackError(err)
 	}
 
+	// We retrieve all node pools for the cluster, including the one we are attempting to
+	// delete, and we pass them to the delete admission validation.
+	// TODO once OCPBUGS-86702 is resolved, we should remove this retrieval and the check of last nodepool being
+	// deleted in the delete admission validation when we decide we want to allow the deletion of the last node pool.
+	nodePoolIterator, err := f.resourcesDBClient.HCPClusters(nodePool.ID.SubscriptionID, nodePool.ID.ResourceGroupName).NodePools(nodePool.ID.Parent.Name).List(ctx, nil)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	clusterNodePools := make([]*api.HCPOpenShiftClusterNodePool, 0)
+	for _, clusterNodePool := range nodePoolIterator.Items(ctx) {
+		clusterNodePools = append(clusterNodePools, clusterNodePool)
+	}
+	if err := nodePoolIterator.GetError(); err != nil {
+		return utils.TrackError(err)
+	}
+
+	nodePoolDeleteAdmissionContext := &admission.NodePoolDeleteAdmissionContext{
+		ClusterNodePools: clusterNodePools,
+	}
+	err = arm.CloudErrorFromFieldErrors(admission.AdmitNodePoolOnDelete(ctx, nodePoolDeleteAdmissionContext, nodePool))
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
 	// Temporary check until creation and deletion interaction with CS is moved to the backend: if a delete arrives and the node
 	// pool has not been created in CS or the ClusterServiceID reference has not been persisted in Cosmos, return an error.
 	if nodePool.ServiceProviderProperties.ClusterServiceID == nil || len(nodePool.ServiceProviderProperties.ClusterServiceID.String()) == 0 {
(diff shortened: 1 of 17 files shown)


internal/admission/utils.go
```diff
@@ -70,6 +70,22 @@ func ValidateClusterNodePoolsMinorVersionSkew(ctx context.Context, resourcesDBCl
 	return errors.Join(errs...)
 }
 
+// countClusterNodePools returns the number of node pools stored for the given cluster.
+func countClusterNodePools(ctx context.Context, resourcesDBClient database.ResourcesDBClient, clusterResourceID *azcorearm.ResourceID) (int, error) {
```
miguelsorianod: We count the nodepools of the cluster independently on the state. This means that if all the other nodepools aside from the nodepool being deleted are being in deleting state, we will also allow deletion.

frontend/pkg/frontend/node_pool.go
```diff
@@ -707,6 +707,11 @@ func (f *Frontend) DeleteNodePool(writer http.ResponseWriter, request *http.Requ
 		return utils.TrackError(err)
 	}
 
+	err = arm.CloudErrorFromFieldErrors(admission.AdmitNodePoolOnDelete(ctx, f.resourcesDBClient, nodePool))
```
deads2k: no error wrapping like this.  Just the plain `utils.TrackError` wrap.  If something special is needed, do it in the general error handling.

frontend/pkg/frontend/node_pool.go
```diff
@@ -707,6 +707,11 @@ func (f *Frontend) DeleteNodePool(writer http.ResponseWriter, request *http.Requ
...
+	if err != nil {
+		return utils.TrackError(err)
+	}
```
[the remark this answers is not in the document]
deads2k: comment look wrong to me.  it's still in cosmos to be deleted when teh cluster is removed as expected.

internal/admission/admit_nodepool.go
```diff
@@ -71,6 +72,27 @@ func mutateNodePoolPlatform(ctx context.Context, admissionContext *NodePoolAdmis
 	return errs
 }
 
+// AdmitNodePoolOnDelete performs non-static checks before deleting a node pool.
+func AdmitNodePoolOnDelete(ctx context.Context, resourcesDBClient database.ResourcesDBClient, nodePool *api.HCPOpenShiftClusterNodePool) field.ErrorList {
```
deads2k: we've done the rest with AdmissionContext.  Create a NodePoolDeleteAdmissionContext and place all the nodepools in there.  Then test the list in here.

test/e2e/simple_negative_cases.go
```diff
@@ -227,7 +227,12 @@ var _ = Describe("Customer", func() {
 					}
 				}
 			}
-			// TEST CASE: https://issues.redhat.com/browse/ARO-22240 to be implemented here
+
+			// TEST CASE: https://issues.redhat.com/browse/ARO-22240 more remaining tests to be implemented here
+			// TODO once OCPBUGS-86702 is fixed, we should remove this test case as we intend to end up allowing deletion of the last node pool for an existing cluster
+			By("attempting to delete the last node pool from a cluster")
+			_, err = nodePoolClient.BeginDelete(ctx, *resourceGroup.Name, clusterParams.ClusterName, nodePoolParams.NodePoolName, nil)
```
deads2k: this will fail on prod, right?  needs to delayed until this gets to prod.  Let's say a two week time bomb.

deads2k: can't this be an integration test?

frontend/pkg/frontend/node_pool.go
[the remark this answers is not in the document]
miguelsorianod: +1, I thought the same

internal/admission/admit_nodepool.go
miguelsorianod: Done

test/e2e/simple_negative_cases.go
miguelsorianod: As discussed, we don't need a full e2e for this. I moved it to an integration test in frontend

miguelsorianod: Moved to a frontendcrud integration test

frontend/pkg/frontend/node_pool.go
miguelsorianod: As discussed through other channels, CloudErrorFromFieldErrors converts from FieldErrors to CloudErrors, which we think it's ok

PR #5440, opened 2026-05-29
feat: introduce asynchronous node pool deletion in backend in a disabled state

miguelsorianod:
This PR builds on top of https://github.com/Azure/ARO-HCP/pull/5429.

The backend logic to perform asynchronous node pool deletion is disabled
for now: all new controllers and the updated node pool delete operation
controller branch on UsesNewNodePoolDeletionApproach, which is not
set by the frontend yet. Legacy synchronous deletion behavior is preserved
via legacySynchronizeOperation in operation_node_pool_delete.

Changes:
- Add NodePoolClusterServiceDeleteDispatchController to call Cluster Service
  DeleteNodePool after DeletionTimestamp is set. It also stamps
  ClusterServiceDeletionTimestamp
- Add NodePoolClusterServiceIDClearerController to clear ClusterServiceID
  once from the NodePool's serviceProviderProperties when Cluster Service
  reports the node pool is gone
- Add NodePoolChildResourcesCleanupController to recursively delete Cosmos
  child resources under the node pool (with gates for Maestro bundles cleanup
  as they depend on ServiceProviderNodepool, and ignoring controllers
  documents
- Add NodePoolDeletionController to delete the NodePool Cosmos document once
  CS cleanup, Maestro bundles cleanup and child resources cleanup are completed
- Update OperationNodePoolDeleteController to complete operations when the
  NodePool Cosmos document is removed (new path), as well as keep CS polling
  to update the node pool delete operation and the nodepool delete
  status from CS state. The controller also keeps the legacy node pool delete
  operation controller behavior by branching depending on UsesNewNodePoolDeletionApproach
- Block cluster Cosmos deletion in OperationClusterDeleteController until all
  node pool documents are removed
- Add MaestroBundleReferenceList.Remove() helper and lister testing utilities

This PR can only be rolled out once https://github.com/Azure/ARO-HCP/pull/5429 has been rolled out to all prod regions

### DIFF
--- a/backend/pkg/controllers/operationcontrollers/operation_cluster_delete.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_cluster_delete.go
@@ -116,10 +116,31 @@ func (c *operationClusterDelete) SynchronizeOperation(ctx context.Context, key c
 		// TODO when we update to make clusterserice creation async, we need to handle this correctly.
 		return nil
 	}
+
 	clusterStatus, err := c.clusterServiceClient.GetClusterStatus(ctx, operation.InternalID)
 	var ocmGetClusterError *ocmerrors.Error
 	if err != nil && errors.As(err, &ocmGetClusterError) && ocmGetClusterError.Status() == http.StatusNotFound {
 		logger.Info("cluster was deleted")
+
+		// Some nodepool controllers require of the Cosmos document to do their cleanup work so we block
+		// the cluster cosmos deletion until the nodepools are gone.
+		nodePoolIterator, err := c.resourcesDBClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).NodePools(operation.ExternalID.Name).List(ctx, nil)
+		if err != nil {
+			return utils.TrackError(err)
+		}
+		foundAtLeastOneNodePool := false
+		for range nodePoolIterator.Items(ctx) {
+			foundAtLeastOneNodePool = true
+			break
+		}
+		err = nodePoolIterator.GetError()
+		if err != nil {
+			return utils.TrackError(err)
+		}
+		if foundAtLeastOneNodePool {
+			logger.Info("cluster still has cosmos nodepools")
+			return nil
+		}
 
 		// Update the Cosmos DB billing document with a deletion timestamp.
 		// Do this before calling setDeleteOperationAsCompleted so that in
--- a/backend/pkg/controllers/operationcontrollers/operation_node_pool_delete.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_node_pool_delete.go
@@ -45,16 +45,25 @@ type operationNodePoolDelete struct {
 // follows an asynchronous node pool deletion operation to completion and updates
 // the corresponding operation document in Cosmos DB.
 //
-// Operation documents relevant to this controller will have the following values:
-//
-//	ResourceType: Microsoft.RedHatOpenShift/hcpOpenShiftClusters/nodePools
-//	     Request: Delete
-//	      Status: any non-terminal value
+// The controller has the following responsibilities:
+//   - While the NodePool Cosmos document is present, it reconciles the
+//     operation and the node pool status.
+//   - When the NodePool Cosmos document is deleted (by the nodePoolDeletionController),
+//     it marks the operation as Succeeded. It also cleans up child
+//     resources. TODO Note: This last part is handled by other controllers too but
+//     because the SetDeleteOperationAsCompleted is still reused by other operations
+//     that have not been migrated to asynchronous flow yet this remains.
 //
 // Note that "to completion" does not imply success. An operation is considered
 // complete when its status field reaches what Azure defines as a terminal value;
 // any of "Succeeded", "Failed", or "Canceled". Once the operation status reaches
 // a terminal value, there will be no further updates to the operation document.
+//
+// Operation documents relevant to this controller will have the following values:
+//
+//	ResourceType: Microsoft.RedHatOpenShift/hcpOpenShiftClusters/nodePools
+//	     Request: Delete
+//	      Status: any non-terminal value
 func NewOperationNodePoolDeleteController(
 	clock utilsclock.PassiveClock,
 	resourcesDBClient database.ResourcesDBClient,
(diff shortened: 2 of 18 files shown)


miguelsorianod:
/hold This PR can only be rolled out when:
* https://github.com/Azure/ARO-HCP/pull/5429 has been rolled out to all prod regions
* CS change that allows deleting last nodepool has been rolled out to all prod regions
* We have verified e2e work as expected with the new approach

backend/pkg/controllers/operationcontrollers/operation_node_pool_delete.go
```diff
@@ -104,25 +161,72 @@ func (c *operationNodePoolDelete) SynchronizeOperation(ctx context.Context, key
...
+			return utils.TrackError(fmt.Errorf("failed to get cluster-service NodePool status: %w", err))
+		}
+		// 404 - CS has finished deleting. nodePoolClusterServiceIDClearer will clear the ID.
+		logger.Info("cluster-service NodePool gone - skipping operation update", "clusterServiceID", nodePoolCSID.String())
+		return nil
+	}
+
+	// If the node pool is in the Ready state from CS side, we wait until the Cosmos NodePool document is deleted, which
+	// will be picked up by a next reconciliation of this controller and we will update the operation to Succeeded.
+	if nodePoolStatus.State().NodePoolStateValue() == string(NodePoolStateReady) {
```
miguelsorianod, line 225: This is the check I mentioned where I added that if the nodepool on CS side is ready we do a no-op from this controller. This avoids marking the operation in a succeeded terminal state here and then in a next reconcile cycle of this controller once the Cosmos nodepool entry is gone it will mark both the resource and the operation as succeeded, stopping the reconciling completely.

backend/pkg/controllers/operationcontrollers/operation_node_pool_delete.go
```diff
@@ -93,6 +102,54 @@ func (c *operationNodePoolDelete) ShouldProcess(ctx context.Context, operation *
...
+
+	err = UpdateOperationStatus(ctx, c.clock, c.resourcesDBClient, operation, newOperationStatus, newOperationError, postAsyncNotificationFn(c.notificationClient))
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	return nil
+}
+
 func (c *operationNodePoolDelete) SynchronizeOperation(ctx context.Context, key controllerutils.OperationKey) error {
```
miguelsorianod, line 153: This file has been changed to have two different paths depending on the operation.UsesNewNodePoolDeletionApproach boolean. When that boolean is false we have the logic that we had for this controller before the introduction of this MR (this is the "legacy" path). When it's true it executes the new logic being introduced for async nodepool deletion here.

backend/pkg/controllers/create_nodepool_scoped_maestro_readonly_bundles_controller.go
```diff
@@ -99,19 +120,101 @@ func NewCreateNodePoolScopedMaestroReadonlyBundlesController(
...
+}
+
+// shouldReconcileDelete returns whether the delete reconciliation path may run: the node pool is deleting, The node pool
+// is considered as fully deleted from Cluster's Service perspective (ClusterServiceDeletionTimestamp set, ClusterServiceID cleared), and
+// and the ServiceProviderNodePool still lists Maestro readonly bundle references to remove.
+func (c *createNodePoolScopedMaestroReadonlyBundlesSyncer) shouldReconcileDelete(nodePool *api.HCPOpenShiftClusterNodePool, serviceProviderNodePool *api.ServiceProviderNodePool) bool {
+	// TODO temporary check to skip the new deletion approach for NodePools that were created before the new approach was implemented.
+	// This will be removed once all nodepools whose deletion was triggered before the new approach is fully rolled out have been
+	// fully deleted in all ARO-HCP permanent environments, for all regions.
+	if !nodePool.ServiceProviderProperties.UsesNewNodePoolDeletionApproach {
```
miguelsorianod: We now gate the new nodepool delete async codepaths based on the UsesNewNodePoolDeletionApproach. The PR should contain it in all the places and controllers we introduced the new async delete nodepool functionality.

deads2k:
> This PR can only be rolled out once https://github.com/Azure/ARO-HCP/pull/5429 has been rolled out to all prod regions

I think this PR can actually rollout regardless since it is properly gated in each controller

/lgtm

backend/pkg/controllers/operationcontrollers/operation_cluster_delete.go
```diff
@@ -116,11 +116,32 @@ func (c *operationClusterDelete) SynchronizeOperation(ctx context.Context, key c
...
 	}
+
 	clusterStatus, err := c.clusterServiceClient.GetClusterStatus(ctx, operation.InternalID)
 	var ocmGetClusterError *ocmerrors.Error
 	if err != nil && errors.As(err, &ocmGetClusterError) && ocmGetClusterError.Status() == http.StatusNotFound {
 		logger.Info("cluster was deleted")
 
+		// Some nodepool controllers require of the Cosmos document to do their cleanup work so we block
+		// the cluster cosmos deletion until the nodepools are gone.
+		nodePoolIterator, err := c.resourcesDBClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).NodePools(operation.ExternalID.Name).List(ctx, nil)
```
miguelsorianod, line 127: This change on **cluster** delete is not guarded so it would be effective after merging the PR. Even with old deletion nodepool approach this should work as when a cluster is deleted frontend triggers cascade nodepool deletions

backend/pkg/controllers/operationcontrollers/operation_cluster_delete.go
```diff
@@ -116,11 +116,32 @@ func (c *operationClusterDelete) SynchronizeOperation(ctx context.Context, key c
...
+			break
+		}
+		err = nodePoolIterator.GetError()
+		if err != nil {
+			return utils.TrackError(err)
+		}
+		if foundAtLeastOneNodePool {
+			logger.Info("cluster still has cosmos nodepools")
+			return nil
+		}
```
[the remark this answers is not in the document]
miguelsorianod: This is not accurate.
With nodepools with the old deletion:
* When the cluster is deleted the RP will create a corresponding operation for the nodepools of the cluster. The nodepool delete operation will reconcile the state from CS and once the nodepool is gone from CS side the cosmos entry will be removed and the cosmos resource removed, which will allow this to continue

* With nodepools with new deletion method: the code path is not enable yet.

PR #5444, opened 2026-05-29
feat: enable nodepool delete from frontend to backend

miguelsorianod:
This builds on top of https://github.com/Azure/ARO-HCP/pull/5440.

Move Cluster Service interaction out of the frontend delete path and
hand it off to the backend's async node pool deletion controllers.

The frontend now only:
- Creates a delete operation with UsesNewNodePoolDeletionApproach
  set to true
- Sets DeletionTimestamp and sets UsesNewNodePoolDeletionApproach to
  true in the NodePool

Removed from the frontend:
- Direct clusterServiceClient.DeleteNodePool() call

Node Pool Delete operations no longer set an InternalID as it is not
needed anymore when using the new node pool deletion approach in
backend.

Legacy deletion (operations without UsesNewNodePoolDeletionApproach) is
unchanged and still handled by OperationNodePoolDeleteController.

### DIFF
--- a/frontend/pkg/frontend/node_pool.go
+++ b/frontend/pkg/frontend/node_pool.go
@@ -17,7 +17,6 @@ package frontend
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"maps"
 	"net/http"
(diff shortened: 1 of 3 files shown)


miguelsorianod:
/hold we cannot merge this PR for now. We need to:
* Wait until https://github.com/Azure/ARO-HCP/pull/5440 is merged
* Wait until we have verified e2e with the approach here
* Wait until the change in CS that relaxes last nodepool validation has been rolled out all the way to aro-hcp prod in all regions
* Wait until https://github.com/Azure/ARO-HCP/pull/5433 has been rolled out all the way to aro-hcp prod in all regions
* Extra checks have been performed

miguelsorianod:

PR #5445, opened 2026-05-29
api: Add Conditions to ARM Resource Types

bennerv:
### What

Adds a read-only conditions API field to our resource types to expose status information and the addition of a new api version 2026-06-30-preview.  

### Why

We can leverage this to expose read-only information similar to k8s conditions which tell the customer if the cluster is degraded.  Because of the controller implementation of the backend, and the fact that customers can make changes to permissions, infra, etc. which can result in degradation of functionality on the cluster, this allows us to expose that information without requiring an ongoing customer operation to do so to move the resource type into a failed state. 

### Testing

The e2e test, integration test, and unit test additions should cover this use case.  

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
+++ b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
@@ -129,6 +129,11 @@ model HcpOpenShiftClusterProperties {
   /** OpenShift internal image registry */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   clusterImageRegistry?: ClusterImageRegistryProfile;
+
+  /** Status of the cluster resource */
+  @visibility(Lifecycle.Read)
+  @added(Versions.v2026_06_30_preview)
+  status?: ResourceStatus;
 }
 
 /** The resource provisioning state. */
@@ -716,6 +721,11 @@ model NodePoolProperties {
    */
   @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   nodeDrainTimeoutMinutes?: int32;
+
+  /** Status of the node pool resource */
+  @visibility(Lifecycle.Read)
+  @added(Versions.v2026_06_30_preview)
+  status?: ResourceStatus;
 }
 
 /** The taint effect the same as in Kubernetes */
@@ -963,6 +973,11 @@ model ExternalAuthProperties {
   @removed(Versions.v2026_06_30_preview)
   condition?: ExternalAuthCondition;
 
+  /** Status of the external auth resource */
+  @visibility(Lifecycle.Read)
+  @added(Versions.v2026_06_30_preview)
+  status?: ResourceStatus;
+
   /** Token Issuer profile */
   issuer: TokenIssuerProfile;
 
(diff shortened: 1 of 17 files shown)


api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2024-06-10-preview/openapi.json
```diff
@@ -3294,7 +3294,7 @@
     },
     "StatusType": {
       "type": "string",
-      "description": "Representation of the possible values of a external auths condition status.",
+      "description": "Representation of the possible values of a condition status.",
```
bennerv, line 3297: technically a breaking change due to a description change, but it's just a description so i'm sure that the review team would be okay with it.

internal/api/enums.go
```diff
@@ -223,6 +223,22 @@ var (
 	)
 )
 
+type ConditionType string
+
+const (
+	ConditionTypeAvailable   ConditionType = "Available"
+	ConditionTypeDegraded    ConditionType = "Degraded"
+	ConditionTypeProgressing ConditionType = "Progressing"
```
bennerv: Our API states that we have a union of these, plus a string value so we can introduce new ones anywhere over time.  

I'm fine if we want to remove this and just accept a string type, or define them as consts with their own type.  Dealer's choice.

internal/database/crud_helpers.go
```diff
@@ -182,6 +184,12 @@ func serializeItem[InternalAPIType, CosmosAPIType any](newObj *InternalAPIType)
 		return nil, nil, fmt.Errorf("invalid partitionKey found in object")
 	}
 
+	if conditionsHolder, ok := any(newObj).(api.ConditionsHolder); ok {
+		if errs := validation.ValidateConditionsForPersist(conditionsHolder); len(errs) > 0 {
```
bennerv: this validates on persist to cosmosdb using the validation functions we have.  Since this is a read-only property from a customer's perspective, validation doesn't make sense externally, and the likelihood that this changes more frequently is higher due to the service owning it and _eventually_ controllers writing this status.

internal/databasetesting/mock_conditions_validation_test.go
```diff
@@ -0,0 +1,210 @@
...
+			ProvisioningState: arm.ProvisioningStateSucceeded,
+			ClusterServiceID:  &internalID,
+		},
+	}
+	_, err = mock.HCPClusters(subscriptionID, resourceGroupName).Create(ctx, cluster, nil)
+	if err != nil {
+		t.Fatalf("Failed to create cluster: %v", err)
+	}
+
+	validExternalAuth := func() *api.HCPOpenShiftClusterExternalAuth {
```
bennerv: Don't need tests for other types since the test is validating the conditions, not the types themselves.

internal/validation/validate_externalauth.go
```diff
@@ -104,6 +104,11 @@ func validateExternalAuthProperties(ctx context.Context, op operation.Operation,
 	//Condition         ExternalAuthCondition       `json:"condition,omitzero"`
```
bennerv: We're not actually setting `Condition` anywhere for external auth.  I'm thinking of removing it completely even if our older APIs expose it, and use the newer `Conditions` field rather.  

I don't think it makes sense to write `ToExternal` code that converts a set `Conditions[]` field to a singular `Condition`.  Which is the most important one?  Do we combine them?  I think it's not worth our time and we should remove the older one.

test-integration/frontend/cross_version_roundtrip_test.go
```diff
@@ -208,8 +209,9 @@ func clusterCreatePayload(clusterName, apiVersion string) []byte {
   "type": "Microsoft.RedHatOpenShift/hcpOpenShiftClusters"
 }`, clusterName, subscriptionID, subscriptionID))
 
-	case v2025:
+	case v2025, v2026:
 		// v2025 payload — includes all optional fields (autoscaling, nodeDrainTimeoutMinutes)
+		// v2026 payload - current the same as 2025, includes conditions as read only field
```
bennerv: when we add more fields, this payload should change, for now it's the same.

bennerv:
/hold

#5503 and https://github.com/Azure/azure-rest-api-specs-pr/pull/28781 (review is green)

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
```diff
@@ -246,6 +247,23 @@ func newKmsKey(from *api.KmsKey) generated.KmsKey {
 	}
 }
 
+func newConditions(from []metav1.Condition) []*generated.Condition {
+	// TODO: Decide which internal condition fields to expose to customers.
+	// Internal conditions carry controller-level detail that should be
+	// filtered or transformed before being returned in the ARM API.
+	return nil
+}
```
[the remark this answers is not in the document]
bennerv: for now we return an empty struct, we can figure out how to plumb it later.

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
```diff
@@ -261,6 +262,16 @@ func newKmsKey(from *api.KmsKey) generated.KmsKey {
...
 
+func newConditions(from []metav1.Condition) []*generated.Condition {
+	return []*generated.Condition{}
+}
+
+func newResourceStatus(from []metav1.Condition) *generated.ResourceStatus {
+	return &generated.ResourceStatus{
+		Conditions: newConditions(from),
+	}
+}
```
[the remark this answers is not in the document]
bennerv: this is intentional

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -979,6 +994,73 @@ model ExternalAuthProperties {
...
+model Condition {
+  /** Type of the condition.
+   * This is a PascalCase identifier representing the type of the condition.
+   */
+  @visibility(Lifecycle.Read)
+  @maxLength(316)
+  @pattern(
+    "^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?([A-Za-z0-9]([-A-Za-z0-9_.]*[A-Za-z0-9])?)$",
+    "Condition type format following Kubernetes condition type conventions."
+  )
```
mbarnes, line 1018: Do we need pattern declarations for read-only fields?  Isn't this typically for input validation?
Same question I guess for min/max length limits, etc. elsewhere.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -1004,7 +1086,7 @@ model ExternalAuthCondition {
   message: string;
 }
 
-/** Representation of the possible values of a external auths condition status. */
+/** Representation of the possible values of a condition status. */
 union StatusType {
```
mbarnes, line 1090: nit: Non-blocking, but maybe move this in closer proximity to the `Condition` model?

mbarnes (commented):
Looks good to me overall.  I was just curious if we really need all those TypeSpec decorators on read-only fields.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
bennerv: Asking David about this to see if we can relax it.  For now, we can leave it as is.

PR #5465, opened 2026-06-01
fix(observability): increase KubeVersionMismatch alert threshold to 90m (AROSLSRE-961)

mmazur:
## Summary
- Increase KubeVersionMismatch alert `for:` duration from 15 minutes to 90 minutes
- KubeVersionMismatch fires on HCP updates, which take much longer than 15 minutes. At this point 90 minutes is a good threshold as the vast majority of upgrades, even delayed ones, should fit within this period.
- Regenerated bicep alert rules (HCP + Msft)
- Split `make sync-upstream` from `make alerts` in `observability/Makefile` so that `make alerts` no longer re-downloads the upstream YAML, allowing local `for:` overrides to persist. The `observability-prometheus` CI step runs `make alerts` and previously would always overwrite local changes. The upstream pin (`prometheusOperatorVersion`) has not been bumped since it was introduced (Apr 2025) — when it is, run `make sync-upstream`, review the diff, restore any local overrides, then `make alerts`.

Jira: https://redhat.atlassian.net/browse/AROSLSRE-961


janboll (commented):
/lgtm

PR #5466, opened 2026-06-01
Add Kusto ingestion latency alerts

janboll:
https://redhat.atlassian.net/browse/AROSLSRE-956

### What
Add two metric alerts for Kusto clusters to detect ingestion pipeline slowdowns:

1. Dynamic threshold alert (Severity 3 - Warning): Uses Azure's machine-learning baseline to detect when ingestion latency deviates above its learned average. Evaluated every 5 minutes over a 30-minute window. Fires when 3 of the last 4 evaluation periods breach the dynamic baseline (i.e. sustained for ~15 minutes). A single transient spike won't trigger it, but a flicker that recovers within one period is tolerated.

2. Static threshold alert (Severity 2 - Error): Hard ceiling at 900 seconds (15 minutes). Catches slow latency creep that the dynamic alert would adapt to over time as its baseline shifts upward.

Alerts are gated on Kusto being enabled and the Kusto cluster region matching the monitoring stack region, since Kusto is deployed once per geography.

### DIFF
--- a/dev-infrastructure/configurations/kusto-monitoring.tmpl.bicepparam
+++ b/dev-infrastructure/configurations/kusto-monitoring.tmpl.bicepparam
@@ -3,5 +3,5 @@ using '../templates/kusto-monitoring.bicep'
 param kustoClusterId = '__kustoClusterId__'
 param kustoRegion = '__kustoRegion__'
 param regionLocation = '{{ .region }}'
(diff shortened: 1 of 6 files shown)


raelga (approved):
/lgtm
/approve

raelga (approved):
/lgtm
/approve

PR #5468, opened 2026-06-01
ARO-27438: log "deleting resource" on delete handlers

swiencki:
[ARO-27438](https://redhat.atlassian.net/browse/ARO-27438)

### What

Emit a structured `deleting resource <id>` log line when a cluster, node pool, or external auth delete request is accepted, in `DeleteCluster`, `DeleteNodePool`, and `DeleteExternalAuth`. This mirrors the existing `creating resource <id>` and `updating resource <id>` lines (see `cluster.go`, `node_pool.go`, `external_auth.go`).

The line is placed right after the provisioning-state conflict check passes, so it fires once per accepted delete and carries the full resource path. The ARM 404 fast-path (resource already gone) intentionally does not log.

### Why

Today the frontend logs create and update with the resource path, but there is no equivalent for delete. Fleet/observability tooling can therefore count delete outcomes (succeeded/failed from backend events) but has no delete-attempt signal. Without a delete denominator we cannot compute a delete success rate, and we cannot detect a delete that hangs without ever reaching a terminal event ("stuck in deleting").

This adds the missing signal with no behavior, schema, or API change. It is purely observability.

### Testing

No new tests. This is a single structured log statement per handler with no control-flow change; existing `frontend/pkg/frontend` unit tests pass and the frontend builds/vets clean. The new lines reuse the same `logger` and resource `.ID` already in scope, identical to the create/update log pattern.

- Placement is symmetric across all three handlers: immediately after `checkForProvisioningStateConflict(... OperationRequestDelete ...)`.
- Consumers should treat this as a delete-attempt marker (one per accepted request), not a success indicator.

### PR Checklist
- [x] Self-reviewed the diff

### DIFF
--- a/frontend/pkg/frontend/cluster.go
+++ b/frontend/pkg/frontend/cluster.go
@@ -705,6 +705,8 @@ func (f *Frontend) DeleteCluster(writer http.ResponseWriter, request *http.Reque
 		return utils.TrackError(err)
 	}
 
+	logger.Info(fmt.Sprintf("deleting resource %s", cluster.ID))
+
 	transaction := f.resourcesDBClient.NewTransaction(cluster.ID.SubscriptionID)
 	if err := f.addDeleteClusterToTransaction(ctx, writer, request, transaction, cluster); err != nil {
 		return utils.TrackError(err)
--- a/frontend/pkg/frontend/external_auth.go
+++ b/frontend/pkg/frontend/external_auth.go
@@ -578,6 +578,8 @@ func (f *Frontend) DeleteExternalAuth(writer http.ResponseWriter, request *http.
 		return utils.TrackError(err)
 	}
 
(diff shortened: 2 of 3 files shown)


raelga (approved):
Code-wise this looks safe and low-risk, but I would not treat the new log as a precise `“accepted delete”` signal.

The log is added after the provisioning-state conflict check, but before later failure points: missing `ClusterServiceID` for node pools/external auth, Cluster Service delete failures, and transaction failures. So it means `“delete attempt reached the handler and passed the conflict check”`, not necessarily `"frontend accepted the delete with 202”`.

I’d be comfortable merging if observability consumers use it as a delete-attempt denominator, matching the existing create/update log style. If they need `“accepted delete”`, **the log should move after the transaction succeeds / just before `StatusAccepted`**.

/lgtm
/approve

PR #5470, opened 2026-06-01
fix(recreate-system-pool): skip forced-evidence on large NRP-KVS window; align AKS version to 1.35.4

hbhushan3:
## What

Two related changes addressing a false-positive forced-evidence trigger and the cosmetic Kubernetes version drift behind it.

### 1. `recreate-system-pool`: guard the forced-evidence path
- Added `shouldTriggerForcedEvidence(windowMin)` which returns `false` when the initial NRP-KVS lookback window is already `>= triggerEvidenceWindowMin` (60m).
- Gated the `forcedEvidencePath` call in `runWith` so the tool no longer triggers a reconcile + 60m poll when the initial window already covered enough time to surface an active wedge (the AKS RP retries a wedged VMSS write roughly every 3 minutes, so a long zero-hit window already proves there is no active storm).
- This prevents an unnecessary `CreateOrUpdate` against an already-mitigated system pool whose ARM state is still stale `Failed`.

### 2. Config: align AKS `kubernetesVersion` to `1.35.4`
- Bumped both svc and mgmt `aks.kubernetesVersion` defaults from `1.35.1` to `1.35.4` to match the version AKS has already patched control planes to out-of-band, removing the desired-vs-current drift. AKS never downgrades, so the config default is the source of the gap.

## Tests
- `TestShouldTriggerForcedEvidence` — table-driven coverage including the large-window skip case.
- New `runWith` case `guard1_fail_large_window_skips_forced_evidence` asserting execution stops after detection with no reconcile/poll calls.
- Verified locally: `ok ... recreate-system-pool`.

## Notes
- The window guard is a targeted stopgap. A follow-up should add a dataplane-health guard (skip forced-evidence when system nodes are Ready/schedulable) so the behavior is correct independent of the configured lookback window.
- Do not hand-edit rendered region config for the version bump; the default in `config/config.yaml` is the source of truth and propagates to all regions.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -468,7 +468,7 @@ defaults:
       vnetAddressPrefix: "10.128.0.0/14"
       subnetPrefix: "10.128.8.0/21"
       podSubnetPrefix: "10.128.64.0/18"
-      kubernetesVersion: 1.35.1
+      kubernetesVersion: 1.35.4
       networkDataplane: "cilium"
       networkPolicy: "cilium"
       systemAgentPool:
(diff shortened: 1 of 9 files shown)


hbhushan3:
/hold

Needs testing. And we are discussing solutions.

PR #5472, opened 2026-06-02
feat: expand nodepool recreate script to include user and infra nodepools

sclarkso:
### What

Extends recreate-system-pool from handling only the AKS system nodepool to detecting and remediating all management cluster nodepools — system, user (swft), and infra. The NRP-KVS corruption is bound to a VMSS ARM resource ID and can affect any SwiftV2 pool, not just the system pool. The tool is also now generalizable for other issues requiring nodepool recreation.

  Detection — detect() now iterates all pools and returns a []wedgedPool list instead of a single bool. Each pool is checked independently for wedge-compatible provisioningState and NRP-KVS evidence on its own VMSS prefix (aks-<poolName>-). Activity log is queried once and filtered per-pool. Pools are classified as
  system/user/infra using AKS Mode field + aro-hcp.azure.com/role node label.

  Remediation — Pools are processed sequentially, system first:
  - System pool: existing systmp dance (AKS requires at least one System-mode pool)
  - User/infra pools: direct drain → delete → recreate (no temporary pool needed)

  Forced evidence — Now per-pool: triggers a reconcile on the specific wedged pool, watches its VMSS prefix for NRP-KVS errors, aborts the triggered LRO.  If there are 0 confirmed nodepools, SKIPS reconcile on infra nodepools since they cannot have this failure state, and we only have 1 hour for this step to run.

  Safety — evalClusterSafety replaces evalNonSystemPools (checks at least one non-system pool has count > 0). Per-pool time budget check (5 min minimum) prevents timeout overrun when multiple pools are wedged. 

  SKIP_GUARDS — Returns all pools as confirmed wedged for testing against healthy clusters.

  - dev-infrastructure/scripts/recreate-system-pool/main.go — multi-pool detection, remediation, forced evidence
  - dev-infrastructure/scripts/recreate-system-pool/main_test.go — updated mock orchestrator, new test cases
  - dev-infrastructure/mgmt-pipeline.yaml — updated step comment

### Why

The NRP-KVS corruption (ICM 798003653) that wedged the system pool in INT and Stage can hit any SwiftV2 pool's VMSS. User pools (userswft1-3) are SwiftV2-enabled in all production environments. Having the self-healing tool only cover the system pool leaves user pool wedges as a manual firefight.

### Testing
 
 - 342 unit test runs pass, including new cases for classifyPool, evalClusterSafety, evalPoolWedged, user-pool-only remediation, and multi-pool ordering
  - Full multi-pool recreation validated on personal dev env (pers-usw3scla-mgmt-1, 6 pools):
  All 6 pools Succeeded after completion. PDB eviction retries during system pool drain resolved within the 10 min drain window.

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/dev-infrastructure/mgmt-pipeline.yaml
+++ b/dev-infrastructure/mgmt-pipeline.yaml
@@ -205,8 +205,10 @@ resourceGroups:
   # KVS corruption (AROSLSRE-924 / ICM 798003653). Checks all pools
   # (system, user/swft, infra). Runs BEFORE the cluster ARM step so
   # the cluster PUT does not get stuck behind a wedged VMSS update.
-  # The Go binary exits 0 (no-op) if the cluster does not exist yet
-  # (greenfield rollout), or if no pools are in a wedge-compatible state.
+  # Exits 0 (no-op) when: cluster does not exist (greenfield), cluster
+  # state is not recoverable, no pools are wedge-compatible, or NRP-KVS
+  # evidence does not reach threshold (including after forced-evidence
+  # probing). Exits non-zero when pools are skipped due to time budget.
   - name: pipeline-msi-reader-permissions
     action: ARM
     template: templates/pipeline-msi-reader-permissions.bicep
@@ -238,9 +240,9 @@ resourceGroups:
         step: subscription-output
         name: subscriptionId
     - name: NRP_FAIL_THRESHOLD
-      value: '10'
+      value: '5'
     - name: NRP_FAIL_WINDOW_MIN
-      value: '360'
+      value: '15'
     dependsOn:
     - resourceGroup: management
       step: cx-oncert-public-kv-issuer
--- a/dev-infrastructure/mgmt-pipeline.yaml
+++ b/dev-infrastructure/mgmt-pipeline.yaml
@@ -224,6 +224,7 @@ resourceGroups:
     action: Shell
     command: ./scripts/recreate-system-pool/recreate-system-pool
     workingDir: .
+    timeout: 120m
     shellIdentity:
       input:
         resourceGroup: global
--- a/dev-infrastructure/mgmt-pipeline.yaml
+++ b/dev-infrastructure/mgmt-pipeline.yaml
@@ -201,12 +201,12 @@ resourceGroups:
         name: mgmtKeyVaultUrl
     issuer:
       value: OneCertV2-PublicCA
-  # Detect and self-heal the AKS system pool when NRP KVS corruption
-  # has wedged it (AROSLSRE-924 / ICM 798003653). Runs BEFORE the cluster
-  # ARM step so the cluster PUT does not get stuck behind a wedged
-  # VMSS update. The Go binary checks cluster existence first and exits
-  # 0 (no-op) if the cluster has not been created yet (greenfield
-  # rollout), or if any of the four detection guards fail.
+  # Detect and self-heal any AKS nodepool whose VMSS is wedged by NRP
+  # KVS corruption (AROSLSRE-924 / ICM 798003653). Checks all pools
+  # (system, user/swft, infra). Runs BEFORE the cluster ARM step so
+  # the cluster PUT does not get stuck behind a wedged VMSS update.
+  # The Go binary exits 0 (no-op) if the cluster does not exist yet
+  # (greenfield rollout), or if no pools are in a wedge-compatible state.
   - name: pipeline-msi-reader-permissions
     action: ARM
     template: templates/pipeline-msi-reader-permissions.bicep
--- a/dev-infrastructure/mgmt-pipeline.yaml
+++ b/dev-infrastructure/mgmt-pipeline.yaml
@@ -224,7 +224,6 @@ resourceGroups:
     action: Shell
     command: ./scripts/recreate-system-pool/recreate-system-pool
     workingDir: .
-    timeout: 120m
     shellIdentity:
       input:
         resourceGroup: global
--- a/dev-infrastructure/scripts/recreate-system-pool/main.go
+++ b/dev-infrastructure/scripts/recreate-system-pool/main.go
@@ -256,7 +256,7 @@ type orchestrator interface {
 	deletePool(ctx context.Context, pool string) error
 	recreatePool(ctx context.Context, poolName string, live *armcs.AgentPool) error
 	reconcileTagPut(ctx context.Context) error
-	triggerPoolReconcile(ctx context.Context, poolName string, live *armcs.AgentPool) error
+	triggerPoolScaleUp(ctx context.Context, poolName string, live *armcs.AgentPool) error
 	pollForNRPEvidence(ctx context.Context, poolName string, vmssPrefix string, timeout time.Duration, pollInterval time.Duration, windowMin int, threshold int) (int, error)
 	abortPoolReconcile(ctx context.Context, poolName string) error
 }
@@ -527,10 +527,12 @@ func remediateNonSystemPool(ctx context.Context, orch orchestrator, wp wedgedPoo
 	return nil
 }
 
-// forcedEvidenceForPool triggers a no-op reconcile on the given pool
-// so the AKS RP attempts a fresh VMSS write. It then polls the
-// activity log for NRP-KVS Failed events and aborts the LRO it started
-// once threshold-many hits are observed or the timeout elapses.
+// forcedEvidenceForPool triggers a scale-up (+1 node) on the given pool
+// to force a VMSS write. A no-op reconcile may not trigger a VMSS write
+// for user pools, but a scale-up always does because it creates a new VM
+// instance. It then polls the activity log for NRP-KVS Failed events and
+// aborts the LRO once threshold-many hits are observed or the timeout
+// elapses.
 //
 // Returns a *wedgedPool if evidence reached the threshold and the caller
 // should proceed with the remediate flow. Returns nil with no error if
@@ -544,12 +546,12 @@ func forcedEvidenceForPool(ctx context.Context, cfg *config, orch orchestrator, 
 		return nil, fmt.Errorf("forced evidence snapshot %s: %w", sp.name, err)
 	}
 
-	if err := orch.triggerPoolReconcile(ctx, sp.name, live); err != nil {
-		logf("WARN: triggerPoolReconcile for %s failed: %v; treating as no-op", sp.name, err)
+	if err := orch.triggerPoolScaleUp(ctx, sp.name, live); err != nil {
+		logf("WARN: triggerPoolScaleUp for %s failed: %v; treating as no-op", sp.name, err)
 		return nil, nil
 	}
 
-	logf("triggered pool %s reconcile; polling activity log every %ds for up to %dm",
+	logf("triggered pool %s scale-up; polling activity log every %ds for up to %dm",
 		sp.name, triggerEvidencePollIntervalSec, triggerEvidenceTimeoutMin)
 	hits, pollErr := orch.pollForNRPEvidence(
 		ctx,
@@ -561,6 +563,7 @@ func forcedEvidenceForPool(ctx context.Context, cfg *config, orch orchestrator, 
 		cfg.threshold,
 	)
 
+	// Abort the scale-up LRO and restore the original pool spec.
 	abortCtx, abortCancel := context.WithTimeout(context.Background(), abortTriggerTimeoutMin*time.Minute)
 	defer abortCancel()
 	if abortErr := orch.abortPoolReconcile(abortCtx, sp.name); abortErr != nil {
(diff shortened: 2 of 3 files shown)


dev-infrastructure/scripts/recreate-system-pool/main.go
```diff
@@ -251,32 +297,59 @@ func runWith(ctx context.Context, cfg *config, orch orchestrator) error {
...
+				return err
+			}
+			if wp != nil {
+				confirmed = append(confirmed, *wp)
+			}
 		}
+		sort.Slice(confirmed, func(i, j int) bool {
+			return confirmed[i].category < confirmed[j].category
+		})
 	}
```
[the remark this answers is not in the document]
sclarkso: This is intended, forced reconciliation on all 6 nodepools (as is the case in stage right now) would lead to this step timing out.

dev-infrastructure/scripts/recreate-system-pool/main.go
```diff
@@ -808,48 +943,36 @@ func evalClusterState(provisioningState string) (bool, string) {
...
-			systemFound = true
-			if p.Properties.MinCount != nil {
-				systemMin = *p.Properties.MinCount
-			}
-			if p.Properties.ProvisioningState != nil {
-				systemProvState = *p.Properties.ProvisioningState
-			}
+		if *p.Name == systemPoolName || *p.Name == systmpPoolName {
 			continue
 		}
```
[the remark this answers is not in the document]
sclarkso: fixed in code

sclarkso:
1. Post-LRO re-detect carry-forward: Pools confirmed in phase 1 that come back as suspected (still wedged but no NRP evidence in the new window) are carried forward for remediation. Pools that disappeared entirely from re-detect (recovered to Succeeded) are correctly dropped.
  2. Tag reconcile before skip error: reconcileTagPut now runs before returning the time-budget error, so pools that were successfully remediated get the cluster-level tag touch even when later pools are skipped.

sclarkso:
Changed reconcile to scale up as Rael said only reconcile didn't work before.

dev-infrastructure/scripts/recreate-system-pool/main.go
```diff
@@ -132,35 +137,28 @@ const (
...
-	// the cluster-state / non-system-pools / system-wedge checks PASS
+	// triggerEvidence forces an AKS RP reconcile of the wedged pool when
+	// the cluster-state / cluster-safety / pool-wedge checks PASS
 	// but the NRP-KVS-storm check has no recent NRP-KVS events in the
 	// configured lookback window. The trigger gives the wedge a chance
 	// to produce fresh evidence (or to prove the wedge is not NRP-KVS).
 	// Times are short relative to the AKS RP retry cadence (~3 min) so
 	// threshold-many retries can accumulate during the wait window.
-	triggerEvidenceTimeoutMin      = 30 // wait at most this long for evidence
+	triggerEvidenceTimeoutMin      = 10 // wait at most this long for evidence
```
raelga: This is too short for the current pipeline threshold. Final constants show `triggerEvidenceTimeoutMin = 10`, while the pipeline still sets `NRP_FAIL_THRESHOLD = 10`. With the documented AKS retry cadence around ~3 minutes, a 10-minute window is unlikely to accumulate 10 failures, making forced evidence frequently inconclusive.

dev-infrastructure/scripts/recreate-system-pool/main.go
```diff
@@ -132,35 +137,28 @@ const (
 	lroAbortAgeMin    = 30
 	lroLookupWindow   = "14d"
 	systmpReadyTOMin  = 10
-	systemReadyTOMin  = 10
+	poolReadyTOMin    = 10
 	pollIntervalSec   = 30
-	overallTimeoutMin = 60
+	overallTimeoutMin = 55
+	minPoolBudgetMin  = 5
```
raelga: Remediation can take drain up to 10m + delete + recreate + wait-ready up to 10m. Starting remediation with ~5 minutes left risks timing out mid-sequence after a pool is deleted. The minimum should reflect the full per-pool worst-case budget, or each pool remediation should run under its own deadline.

```suggestion
minPoolBudgetMin  = 20
```

dev-infrastructure/scripts/recreate-system-pool/main.go
raelga:
```suggestion
	triggerEvidenceTimeoutMin      = 20 // wait at most this long for evidence
```

dev-infrastructure/mgmt-pipeline.yaml
raelga: With a 20-minute forced-evidence timeout, a threshold of 5 should be reasonable.

```suggestion
      value: '5'
```

raelga, line 244: The pipeline sets `NRP_FAIL_THRESHOLD=10` and `NRP_FAIL_WINDOW_MIN=360`, but forced evidence uses `triggerEvidenceTimeoutMin=10` and `triggerEvidenceWindowMin=60`. That means the same threshold is evaluated over a different effective collection period than initial detection. This is easy to misread and may be miscalibrated.

Either expose forced-evidence threshold/window separately, or add comments/logging that explicitly say: initial detection uses `NRP_FAIL_WINDOW_MIN`; forced evidence uses its own timeout/lookback.

dev-infrastructure/scripts/recreate-system-pool/main.go
```diff
@@ -1287,23 +1509,31 @@ func (c *clients) reconcileTagPut(ctx context.Context) error {
...
+	}
+	if body.Properties.Count != nil {
+		*body.Properties.Count++
+	} else {
+		body.Properties.Count = ptr(int32(2))
 	}
-	logf("triggering AgentPool CreateOrUpdate on %q (no-op reconcile with live spec)", systemPoolName)
-	if _, err := c.pools.BeginCreateOrUpdate(ctx, c.cfg.resourceGroup, c.cfg.clusterName, systemPoolName, *body, nil); err != nil {
-		return fmt.Errorf("begin trigger system reconcile: %w", err)
+	if body.Properties.EnableAutoScaling != nil && *body.Properties.EnableAutoScaling && body.Properties.MinCount != nil {
```
raelga, line 1522: If autoscaling is enabled and MaxCount == MinCount, increasing `MinCount` without also increasing MaxCount can produce an invalid request (`minCount` > [maxCount]). For Stage/Prod pools we’ve seen min/max like 1/3, 3/6, so likely okay today, but the helper should be defensive.

dev-infrastructure/scripts/recreate-system-pool/main.go
```diff
@@ -1287,23 +1509,31 @@ func (c *clients) reconcileTagPut(ctx context.Context) error {
...
+		*body.Properties.MinCount++
+	}
+	logf("triggering scale-up on %q (count=%d) to force VMSS write", poolName, *body.Properties.Count)
+	if _, err := c.pools.BeginCreateOrUpdate(ctx, c.cfg.resourceGroup, c.cfg.clusterName, poolName, *body, nil); err != nil {
+		return fmt.Errorf("begin trigger pool scale-up %s: %w", poolName, err)
 	}
 	return nil
```
raelga, line 1532: The cleanup is only `abortPoolReconcile`; if abort fails or the scale-up completes before abort, the pool may stay scaled up. The code should explicitly restore the original pool spec from the snapshot after the probe.

dev-infrastructure/scripts/recreate-system-pool/main.go
```diff
@@ -1260,12 +1482,12 @@ func (c *clients) recreateSystem(ctx context.Context, live *armcs.AgentPool) err
...
+	logf("pool %s ARM-Succeeded; waiting for %d Ready k8s node(s)", poolName, expected)
+	return c.waitForReadyNodes(ctx, poolName, int(expected), poolReadyTOMin*time.Minute)
 }
 
 // ---------------------------------------------------------------------------
-// step 8 :: no-op tag reconcile via SDK tag PATCH
+// no-op tag reconcile via SDK tag PATCH
 // ---------------------------------------------------------------------------
 
 func (c *clients) reconcileTagPut(ctx context.Context) error {
```
raelga, line 1484: `reconcileTagPut` intentionally kicks an ARM reconciliation. If the script already knows there are still confirmed wedged pools, triggering a cluster reconcile may immediately run into the remaining broken VMSS state again. It’s safer to return the skipped-pools error before tag reconcile, so we don’t deliberately kick ARM when we know remediation is incomplete.

raelga (changes requested):

dev-infrastructure/scripts/recreate-system-pool/main.go
sclarkso: Updated Timeout

dev-infrastructure/scripts/recreate-system-pool/main.go
sclarkso: Updated Min

dev-infrastructure/mgmt-pipeline.yaml
sclarkso: Updated NRP_FAIL_THRESHOLD

dev-infrastructure/scripts/recreate-system-pool/main.go
sclarkso: Scale-up cleanup: added restorePoolSpec to revert original spec after probe

dev-infrastructure/scripts/recreate-system-pool/main.go
sclarkso: Scale-up: bump MaxCount when MinCount would exceed it

dev-infrastructure/scripts/recreate-system-pool/main.go
sclarkso: Tag reconcile moved after skip error (don't kick ARM with incomplete remediation)

dev-infrastructure/mgmt-pipeline.yaml
sclarkso: added comment clarifying that forced evidence uses it's own timeout.

dev-infrastructure/scripts/recreate-system-pool/main.go
```diff
@@ -174,6 +173,65 @@ const (
...
+func poolVMSSPrefix(poolName string) string {
+	return "aks-" + poolName + "-"
+}
+
+func tmpPoolName(poolName string) string {
+	if poolName == systemPoolName {
+		return systmpPoolName
+	}
+	return poolName + tmpPoolSuffix
+}
```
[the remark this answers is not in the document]
sclarkso: updated, even though none of our current nodepool names are 12 chars.

raelga (approved):
Awesome work @sclarkso!

/lgtm
/approve

dev-infrastructure/scripts/recreate-system-pool/main.go
```diff
@@ -127,40 +132,40 @@ import (
...
 	defaultThreshold  = 5
 	defaultWindowMin  = 15
 	lroAbortAgeMin    = 30
 	lroLookupWindow   = "14d"
 	systmpReadyTOMin  = 10
-	systemReadyTOMin  = 10
+	poolReadyTOMin    = 10
 	pollIntervalSec   = 30
-	overallTimeoutMin = 60
+	overallTimeoutMin = 55
```
raelga: Let's bump this to 110 minutes to be safe.

dev-infrastructure/mgmt-pipeline.yaml
```diff
@@ -238,9 +240,9 @@ resourceGroups:
...
         name: subscriptionId
     - name: NRP_FAIL_THRESHOLD
-      value: '10'
+      value: '5'
     - name: NRP_FAIL_WINDOW_MIN
-      value: '360'
+      value: '15'
     dependsOn:
     - resourceGroup: management
       step: cx-oncert-public-kv-issuer
```
raelga, line 248:
```suggestion
    workingDir: .
    timeout: 120m
```

That gives enough room for:

 - forced evidence: up to 20m
 - system pool remediation: ~25–35m
 - one or two non-system pools: ~20m each
 - final tag reconcile/postflight
 - Ev2 wrapper overhead + 2m failure sleep

Keep the step timeout ~10m higher than the script timeout so the Go binary exits cleanly before Ev2 kills the shell.

sclarkso: added this to mgmt-pipeline.

dev-infrastructure/scripts/recreate-system-pool/main.go
sclarkso: updated overallTimeoutMin to 110

raelga:
/close 

Pivoting our mitigation strategy to reduce risk.
After reviewing the proposed fix to self-heal all nodepools (PR #[5472](https://github.com/Azure/ARO-HCP/pull/5472)), we've decided to close it. The script had grown to ~2,000 LOC, which introduces too large of a blast radius for our current fragile state.

To be more risk-averse, we are splitting the remediation into two distinct, smaller steps:
:one: System Nodepools: We will continue using the existing, validated system nodepool script as our stable baseline.
:two: User Nodepools: We are creating a separate, stripped-down script exclusively for the user nodepools. We will reuse the proven system logic to keep the net-new code changes as minimal as possible.

PR #5476, opened 2026-06-02
feat: introduce asynchronous external auth deletion in backend in a disabled state

miguelsorianod:
The backend logic to perform asynchronous external auth deletion is disabled for now: all new controllers and the updated external auth delete operation controller branch on UsesNewExternalAuthDeletionApproach, which is not set by the frontend yet. Legacy synchronous deletion behavior is preserved via legacySynchronizeOperation in operation_externa_auth_delete.

Changes:
- Add ExternalAuthClusterServiceDeleteDispatchController to call Cluster Service DeleteExternalAuth after DeletionTimestamp is set. It also stamps ClusterServiceDeletionTimestamp
- Add ExternalAuthClusterServiceIDClearerController to clear ClusterServiceID once from the ExternalAuth's serviceProviderProperties when Cluster Service reports the external auth is gone
- Add ExternalAuthChildResourcesCleanupController to recursively delete Cosmos child resources under the external auth, and ignoring controllers documents
- Add ExternalAuthDeletionController to delete the External Auth Cosmos document once CS cleanup and child resources cleanup are completed
- Update OperationExternalAuthDeleteController to complete operations when the ExternalAuth Cosmos document is removed (new path), as well as keep CS polling to update the externalauth delete operation and the externalauth delete status from CS state. The controller also keeps the legacy external auth delete operation controller behavior by branching depending on UsesNewExternalAuthDeletionApproach
- Block cluster Cosmos deletion in OperationClusterDeleteController until all external auth documents are removed

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/backend/pkg/controllers/operationcontrollers/operation_cluster_delete.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_cluster_delete.go
@@ -142,6 +142,26 @@ func (c *operationClusterDelete) SynchronizeOperation(ctx context.Context, key c
 			return nil
 		}
 
+		// Some external auth controllers require the Cosmos document to do their cleanup work, so we block
+		// the cluster cosmos deletion until the external auths are gone.
+		externalAuthIterator, err := c.resourcesDBClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).ExternalAuth(operation.ExternalID.Name).List(ctx, nil)
+		if err != nil {
+			return utils.TrackError(err)
+		}
+		foundAtLeastOneExternalAuth := false
+		for range externalAuthIterator.Items(ctx) {
+			foundAtLeastOneExternalAuth = true
+			break
+		}
+		err = externalAuthIterator.GetError()
+		if err != nil {
+			return utils.TrackError(err)
+		}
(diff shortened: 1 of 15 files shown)


backend/pkg/controllers/operationcontrollers/operation_cluster_delete.go
```diff
@@ -142,6 +142,26 @@ func (c *operationClusterDelete) SynchronizeOperation(ctx context.Context, key c
 			return nil
 		}
 
+		// Some externalauths controllers require of the Cosmos document to do their cleanup work so we block
+		// the cluster cosmos deletion until the externalauths are gone.
+		externalAuthIterator, err := c.resourcesDBClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).ExternalAuth(operation.ExternalID.Name).List(ctx, nil)
```
miguelsorianod, line 147: Similar to what was done in https://github.com/Azure/ARO-HCP/pull/5440, here we are also introducing changing the operation **cluster** delete behavior to wait until all the externalauth cosmos resources are fully gone to proceed with cluster deletion.

Even with old deletion externalauth approach this should work as when a cluster is deleted frontend triggers cascade externalauth deletions

PR #5486, opened 2026-06-03
fix: resolve cross-subscription ARM deployment diagnostics (ARO-27236)

gmfrasca:
Addresses [ARO-27236](https://redhat.atlassian.net/browse/ARO-27236)

### What

Introduce a subscription-aware `OperationsClientGetter` factory that lazily creates and caches `DeploymentOperationsClient` instances per Azure subscription. Populate `Resource.SubscriptionID` from parsed ARM resource IDs to carry subscription context through the operation tree.

Additionally, consolidate `fetchSubscriptionScopedOperationsFor` into `fetchOperationsFor` by branching on `resourceGroup == ""`. This fixes a pre-existing bug where subscription-scoped parent deployments would recurse into child deployments using `NewListAtSubscriptionScopePager` instead of `NewListPager`, which failed for resource-group-scoped children.

### Why

The `DeploymentOperationsClient` was bound to a single subscription, causing `ResourceGroupNotFound` errors when traversing child deployments that reside in a different subscription. This broke deployment operation visualization and timing metadata collection for cross-subscription bicep deployments.

### Testing

- Unit tests for `operationFor` validating `SubscriptionID` extraction from ARM resource IDs (including cross-subscription nested deployments)
- Unit tests for `NewCachedOperationsClientGetter` validating default client reuse and lazy creation/caching for new subscriptions
- Existing E2E deployment timing collection exercises the updated `fetchOperationsFor` path
- Unit tests for cross-subscription and same-subscription routing in `fetchOperationsFor`

- The `OperationsClientGetter` closure uses a `sync.Mutex` to protect the internal client cache, even though current usage is sequential. This is cheap insurance against future concurrent callers.
- The pipeline and test framework each have their own copy of `NewCachedOperationsClientGetter` with slightly different signatures (the test framework version accepts `*azcorearm.ClientOptions`). This is intentional — they have different dependency chains and the test framework needs to pass custom client options for its Azure environment.

### PR Checklist
- [x] Self-reviewed the diff

### DIFF
--- a/tooling/templatize/pkg/pipeline/arm_operations.go
+++ b/tooling/templatize/pkg/pipeline/arm_operations.go
@@ -18,8 +18,10 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 )
@@ -44,15 +46,47 @@ type Operation struct {
 type Resource struct {
 	// ResourceType is the resource provider and resource name, like "Microsoft.KeyVault/vaults".
 	ResourceType string `json:"resourceType"`
+	// SubscriptionID is the Azure subscription in which the resource exists.
+	// Not serialized to avoid leaking sensitive identifiers into published artifacts.
+	SubscriptionID string `json:"-"`
 	// ResourceGroup is the Azure resource group in which the resource exists.
 	ResourceGroup string `json:"resourceGroup"`
 	// Name is the name of the resource.
 	Name string `json:"name"`
 }
 
-func DetermineOperationsForResourceGroupDeployment(client *armresources.DeploymentOperationsClient, resourceGroup, deploymentName string) DetailsProducer {
+// OperationsClientGetter returns a DeploymentOperationsClient for the given subscription ID.
+// Implementations should cache clients to avoid redundant authentication.
+type OperationsClientGetter func(subscriptionID string) (*armresources.DeploymentOperationsClient, error)
+
+// NewCachedOperationsClientGetter creates an OperationsClientGetter that caches clients per subscription.
+// The provided defaultClient is used for the given subscriptionID; clients for other subscriptions
+// are created lazily using the provided credential.
+func NewCachedOperationsClientGetter(subscriptionID string, defaultClient *armresources.DeploymentOperationsClient, cred azcore.TokenCredential, opts *azcorearm.ClientOptions) OperationsClientGetter {
+	// mu guards the clients map, which may be accessed concurrently if
+	// multiple deployment operations are fetched in parallel.
+	var mu sync.Mutex
+	clients := map[string]*armresources.DeploymentOperationsClient{
+		subscriptionID: defaultClient,
+	}
+	return func(subID string) (*armresources.DeploymentOperationsClient, error) {
+		mu.Lock()
+		defer mu.Unlock()
+		if c, ok := clients[subID]; ok {
+			return c, nil
+		}
+		c, err := armresources.NewDeploymentOperationsClient(subID, cred, opts)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create operations client for subscription: %w", err)
+		}
+		clients[subID] = c
+		return c, nil
+	}
+}
+
+func DetermineOperationsForResourceGroupDeployment(getClient OperationsClientGetter, subscriptionID, resourceGroup, deploymentName string) DetailsProducer {
 	return func(ctx context.Context) (*ExecutionDetails, error) {
-		ops, err := fetchOperationsFor(ctx, client, resourceGroup, deploymentName)
+		ops, err := fetchOperationsFor(ctx, getClient, subscriptionID, resourceGroup, deploymentName)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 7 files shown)


tooling/templatize/pkg/pipeline/arm_operations_test.go
```diff
@@ -0,0 +1,145 @@
...
+	})
+
+	t.Run("creates and caches client for new subscription", func(t *testing.T) {
+		defaultClient := &armresources.DeploymentOperationsClient{}
+		getter := NewCachedOperationsClientGetter("sub-1", defaultClient, nil, nil)
+
+		client1, err := getter("sub-2")
+		require.NoError(t, err)
+		require.NotNil(t, client1)
+		assert.NotSame(t, defaultClient, client1)
```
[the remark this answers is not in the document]
gmfrasca: False positive - `armresources.NewDeploymentOperationsClient` accepts nil credentials without error (credential validation is deferred to the first API call). The test passes as-is, both locally and in CI.

tooling/templatize/pkg/pipeline/arm_operations.go
```diff
@@ -44,25 +46,57 @@ type Operation struct {
...
+	return func(subID string) (*armresources.DeploymentOperationsClient, error) {
+		mu.Lock()
+		defer mu.Unlock()
+		if c, ok := clients[subID]; ok {
+			return c, nil
+		}
+		c, err := armresources.NewDeploymentOperationsClient(subID, cred, opts)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create operations client for subscription: %w", err)
+		}
```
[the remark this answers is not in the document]
gmfrasca: Intentional - subscription IDs are omitted from our error format strings to avoid leaking sensitive identifiers into public Prow CI artifacts. Cross-subscription child deployment IDs are discovered at runtime from ARM resource IDs and may not be covered by ci-operator's secret censoring (which only replaces values from mounted secret files).

tooling/templatize/pkg/pipeline/arm_operations.go
```diff
@@ -72,31 +106,64 @@ func DetermineOperationsForSubscriptionDeployment(client *armresources.Deploymen
...
 
-func fetchOperationsFor(ctx context.Context, client *armresources.DeploymentOperationsClient, resourceGroup, deploymentName string) ([]Operation, error) {
+// fetchOperationsFor retrieves ARM deployment operations, handling both resource-group-scoped
+// and subscription-scoped deployments. Recursively fetches child deployment operations,
+// resolving the correct subscription client when a nested deployment crosses subscriptions.
+func fetchOperationsFor(ctx context.Context, getClient OperationsClientGetter, subscriptionID, resourceGroup, deploymentName string) ([]Operation, error) {
+	client, err := getClient(subscriptionID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get operations client for subscription: %w", err)
+	}
```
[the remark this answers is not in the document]
gmfrasca: Intentional - subscription IDs are omitted from our error format strings to avoid leaking sensitive identifiers into public Prow CI artifacts. Cross-subscription child deployment IDs are discovered at runtime from ARM resource IDs and may not be covered by ci-operator's secret censoring (which only replaces values from mounted secret files).

tooling/templatize/pkg/pipeline/arm_operations.go
```diff
@@ -72,31 +106,64 @@ func DetermineOperationsForSubscriptionDeployment(client *armresources.Deploymen
...
 
+	// Recurse into nested deployments, which may target a different subscription
 	for i, op := range operations {
 		if op.Resource == nil {
 			continue
 		}
 		if strings.EqualFold(op.Resource.ResourceType, "Microsoft.Resources/deployments") {
-			children, err := fetchOperationsFor(ctx, client, op.Resource.ResourceGroup, op.Resource.Name)
+			// Use the child's subscription if available, otherwise inherit the parent's
+			childSub := subscriptionID
```
swiencki, line 162: The cross-sub routing decision (child sub vs parent sub) doesn't seem to have a direct unit test covering it.

tooling/templatize/cmd/entrypoint/visualize/options.go
```diff
@@ -203,9 +203,10 @@ func operationFor(op pipeline.Operation) (ARMOperation, error) {
 	var resource *ARMResource
 	if op.Resource != nil {
 		resource = &ARMResource{
-			ResourceType:  op.Resource.ResourceType,
-			ResourceGroup: op.Resource.ResourceGroup,
-			Name:          op.Resource.Name,
+			ResourceType:   op.Resource.ResourceType,
+			SubscriptionID: op.Resource.SubscriptionID,
```
swiencki: This field will always be empty since the source structs tag it `json:"-"`.

swiencki (commented):
The consolidation also fixes sub-scoped parent to RG-scoped child recursion, which seems worth calling out in the description.

tooling/templatize/pkg/pipeline/arm_operations.go
gmfrasca: Added `TestFetchOperationsForCrossSubscription` and `TestFetchOperationsForSameSubscription`, and updated PR description to reflect this and overarching review comment

tooling/templatize/cmd/entrypoint/visualize/options.go
gmfrasca: Good catch - removed `SubscriptionID` from `ARMResource`. 

It was originally added for consistency with the source `pipeline.Resource` / `timing.Resource` structs. Those structs were later updated to use `json:"-"` to prevent leaking child subscription IDs in timing artifacts, since child IDs are discovered at runtime and might not be caught by ci-operator censoring. Since it's never read in the visualization path, there's no reason to carry it here.

raelga (approved):
/lgtm
/approve

Consolidating `fetchSubscriptionScopedOperationsFor` into `fetchOperationsFor` cleanly fixes the RG-scoped child recursion bug, and the per-subscription client cache with parent→child sub fallback is the right approach for cross-sub nested deployments.

Good test coverage on the routing (`TestFetchOperationsFor{Cross,Same}Subscription` asserting the exact client call order) and on sub-ID extraction. Agree with the json:"-" choice to keep child sub IDs out of public artifacts.

@gmfrasca The PR description links the deprecated [issues.redhat.com](issues.redhat.com) host, it should be [redhat.atlassian.net](redhat.atlassian.net).

PR #5488, opened 2026-06-03
 Cache role definition fetch errors

shitaljante:
https://redhat.atlassian.net/browse/ARO-27501

### What

Extends RoleDefinitionsCachedReader so failed Azure GetByID calls are cached in memory, not only successful responses.

- Success entries: unchanged 6 hour TTL (roleDefinitionResourceIDCacheKeySuccessTTL)
- Error entries: new 5 minute TTL (roleDefinitionResourceIDCacheKeyErrorTTL)

### Why

Enhancement to https://github.com/Azure/ARO-HCP/pull/4403. 
It will reduce repeated Azure API calls when the same invalid or missing role definition is requested multiple times, helping reduce load on the service.

### DIFF
--- a/backend/pkg/azure/cachedreader/role_definitions.go
+++ b/backend/pkg/azure/cachedreader/role_definitions.go
@@ -18,7 +18,6 @@ package cachedreader
 
 import (
 	"context"
-	"fmt"
 	"sync"
 	"time"
 
@@ -32,9 +31,13 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
-// roleDefinitionResourceIDCacheKeyTTL defines how long a cached role definition is considered valid.
+// roleDefinitionResourceIDCacheKeySuccessTTL defines how long a cached successful GetByID is valid.
 // After this TTL (6 hours), the cached entry is stale and the next GetCachedByID refreshes it from Azure.
-const roleDefinitionResourceIDCacheKeyTTL = 6 * time.Hour
+const roleDefinitionResourceIDCacheKeySuccessTTL = 6 * time.Hour
+
+// roleDefinitionResourceIDCacheKeyErrorTTL defines how long a cached GetByID that returned an error is valid.
+// After this TTL (5 minutes), the cached entry is stale and the next GetCachedByID retries Azure.
+const roleDefinitionResourceIDCacheKeyErrorTTL = 5 * time.Minute
 
 // RoleDefinitionsCachedReader exposes cached reads of role definitions by ARM resource ID.
 type RoleDefinitionsCachedReader interface {
@@ -44,6 +47,9 @@ type RoleDefinitionsCachedReader interface {
 // roleDefinitionsCachedReader wraps an uncached azureclient.RoleDefinitionsClient. GetCachedByID
 // responses are cached in memory with a TTL and singleflight deduplication; other RoleDefinitionsClient
 // methods are not exposed on this type.
+// TTLs used for cached role definition entries.
+// Successful lookups are cached for 6 hours (roleDefinitionResourceIDCacheKeySuccessTTL).
+// Error responses are cached for 5 minutes (roleDefinitionResourceIDCacheKeyErrorTTL).
 type roleDefinitionsCachedReader struct {
 	// inner is the uncached Azure Role Definitions client used for live GetByID calls.
 	inner azureclient.RoleDefinitionsClient
@@ -61,6 +67,8 @@ type roleDefinitionsCachedReader struct {
 
 type cachedGetByIDResponse struct {
 	response armauthorization.RoleDefinitionsClientGetByIDResponse
+	// err contains the error returned by the GetByID call made to Azure. nil if no error in the GetByID call.
+	err error
 	// lastUpdate is the wall-clock instant (UTC) when this cache entry was written, from time.Now().UTC().
 	// It is compared with time.Since for TTL / staleness checks (same interpretation as time.Now).
 	lastUpdate time.Time
@@ -83,7 +91,7 @@ func (c *roleDefinitionsCachedReader) GetCachedByID(ctx context.Context, roleDef
 	c.roleDefinitionsCacheLock.RLock()
 	entry := c.roleDefinitionsCache[roleDefinitionResourceID]
 	c.roleDefinitionsCacheLock.RUnlock()
-	return entry.response, nil
+	return entry.response, entry.err
 }
 
 // ensureCachedGetByID loads a fresh role definition when missing or stale. Concurrent callers for the
@@ -98,13 +106,11 @@ func (c *roleDefinitionsCachedReader) ensureCachedGetByID(ctx context.Context, r
 	}
 	_, err, _ := c.sfGroup.Do(roleDefinitionResourceID, func() (interface{}, error) {
 		resp, err := c.inner.GetByID(ctx, roleDefinitionResourceID, options)
-		if err != nil {
-			return nil, utils.TrackError(fmt.Errorf("failed to get role definition for '%s': %w", roleDefinitionResourceID, err))
-		}
 		c.roleDefinitionsCacheLock.Lock()
 		defer c.roleDefinitionsCacheLock.Unlock()
 		c.roleDefinitionsCache[roleDefinitionResourceID] = cachedGetByIDResponse{
 			response:   resp,
+			err:        err,
 			lastUpdate: c.clock.Now().UTC(),
 		}
 		return nil, nil
@@ -116,7 +122,11 @@ func (c *roleDefinitionsCachedReader) ensureCachedGetByID(ctx context.Context, r
 }
 
 func (c *roleDefinitionsCachedReader) isStale(entry cachedGetByIDResponse) bool {
-	return c.clock.Since(entry.lastUpdate) > roleDefinitionResourceIDCacheKeyTTL
+	ttl := roleDefinitionResourceIDCacheKeySuccessTTL
+	if entry.err != nil {
+		ttl = roleDefinitionResourceIDCacheKeyErrorTTL
+	}
+	return c.clock.Since(entry.lastUpdate) > ttl
 }
 
 var _ RoleDefinitionsCachedReader = (*roleDefinitionsCachedReader)(nil)
--- a/backend/pkg/azure/cachedreader/role_definitions_test.go
+++ b/backend/pkg/azure/cachedreader/role_definitions_test.go
@@ -51,7 +51,6 @@ func TestRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
 	}
 
 	ctx := context.Background()
-	innerErr := errors.New("azure unavailable")
 	cachedResponse := roleDefByIDResponse(testRoleDefinitionRID)
 	initialCached := roleDefByIDResponse(testRoleDefinitionRID)
 	refreshedCached := roleDefByIDResponse(testRoleDefinitionRID)
@@ -99,13 +98,15 @@ func TestRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
 			},
 		},
 		{
-			name: "propagates azure api interaction error when it occurs",
+			name: "caches error within error TTL",
 			setupClient: func(ctrl *gomock.Controller) azureclient.RoleDefinitionsClient {
 				mockClient := azureclient.NewMockRoleDefinitionsClient(ctrl)
-				mockClient.EXPECT().GetByID(gomock.Any(), testRoleDefinitionRID, nil).Return(armauthorization.RoleDefinitionsClientGetByIDResponse{}, innerErr).Times(1)
+				apiErr := errors.New("service unavailable")
+				// Times(1): all GetCachedByID calls within error TTL must not call Azure again.
+				mockClient.EXPECT().GetByID(gomock.Any(), testRoleDefinitionRID, nil).Return(armauthorization.RoleDefinitionsClientGetByIDResponse{}, apiErr).Times(1)
 				return mockClient
 			},
-			clock: utilsclock.RealClock{},
+			clock: clocktesting.NewFakePassiveClock(time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)),
 			calls: []struct {
 				advanceClockBy   time.Duration
 				roleDefinitionID string
@@ -116,12 +117,25 @@ func TestRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
 				{
 					roleDefinitionID: testRoleDefinitionRID,
 					wantError:        true,
-					wantErrContains:  "failed to get role definition",
+					wantErrContains:  "service unavailable",
+				},
+				{
+					advanceClockBy:   2 * time.Minute,
+					roleDefinitionID: testRoleDefinitionRID,
+					wantError:        true,
+					wantErrContains:  "service unavailable",
+				},
+				{
+					// At exactly error TTL (isStale uses >), entry must still be fresh.
+					advanceClockBy:   roleDefinitionResourceIDCacheKeyErrorTTL - 2*time.Minute,
+					roleDefinitionID: testRoleDefinitionRID,
+					wantError:        true,
+					wantErrContains:  "service unavailable",
 				},
 			},
 		},
 		{
-			name: "error is not cached; next call retries",
+			name: "recovers after error TTL and caches success for success TTL",
 			setupClient: func(ctrl *gomock.Controller) azureclient.RoleDefinitionsClient {
 				mockClient := azureclient.NewMockRoleDefinitionsClient(ctrl)
 				gomock.InOrder(
@@ -130,7 +144,7 @@ func TestRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
 				)
 				return mockClient
 			},
-			clock: utilsclock.RealClock{},
+			clock: clocktesting.NewFakePassiveClock(time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)),
 			calls: []struct {
 				advanceClockBy   time.Duration
 				roleDefinitionID string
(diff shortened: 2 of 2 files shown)


backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -98,11 +107,15 @@ func (c *roleDefinitionsCachedReader) ensureCachedGetByID(ctx context.Context, r
...
 	_, err, _ := c.sfGroup.Do(roleDefinitionResourceID, func() (interface{}, error) {
 		resp, err := c.inner.GetByID(ctx, roleDefinitionResourceID, options)
-		if err != nil {
-			return nil, utils.TrackError(fmt.Errorf("failed to get role definition for '%s': %w", roleDefinitionResourceID, err))
-		}
 		c.roleDefinitionsCacheLock.Lock()
 		defer c.roleDefinitionsCacheLock.Unlock()
+		if err != nil {
+			c.roleDefinitionsCache[roleDefinitionResourceID] = cachedGetByIDResponse{
+				err:        utils.TrackError(fmt.Errorf("failed to get role definition for '%s': %w", roleDefinitionResourceID, err)),
```
shitaljante: @miguelsorianod Should we differentiate between the transient 5xx errors and not found errors like 404 ?

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -83,6 +89,9 @@ func (c *roleDefinitionsCachedReader) GetCachedByID(ctx context.Context, roleDef
 	c.roleDefinitionsCacheLock.RLock()
 	entry := c.roleDefinitionsCache[roleDefinitionResourceID]
 	c.roleDefinitionsCacheLock.RUnlock()
+	if entry.err != nil {
```
miguelsorianod: Is there some reason this is checked here? doesn't just `return entry.response, entry.error` work the same?

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -98,11 +107,15 @@ func (c *roleDefinitionsCachedReader) ensureCachedGetByID(ctx context.Context, r
 	}
 	_, err, _ := c.sfGroup.Do(roleDefinitionResourceID, func() (interface{}, error) {
 		resp, err := c.inner.GetByID(ctx, roleDefinitionResourceID, options)
-		if err != nil {
-			return nil, utils.TrackError(fmt.Errorf("failed to get role definition for '%s': %w", roleDefinitionResourceID, err))
-		}
 		c.roleDefinitionsCacheLock.Lock()
 		defer c.roleDefinitionsCacheLock.Unlock()
+		if err != nil {
```
miguelsorianod: Is there some reason this is checked here instead of just storing the values?

backend/pkg/azure/cachedreader/role_definitions.go
miguelsorianod: No. All errors treated the same

miguelsorianod: This is not correct. We shouldn't decorate the error returned by the API. This type is a transparent cache

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -61,6 +65,8 @@ type roleDefinitionsCachedReader struct {
 
 type cachedGetByIDResponse struct {
 	response armauthorization.RoleDefinitionsClientGetByIDResponse
+	// err is set for negative cache entries (failed GetByID). When non-nil, response is ignored.
```
miguelsorianod: error just contains the error returned by the GetByID call made to Azure. nil if no error in the GetByID call

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -32,9 +32,13 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
-// roleDefinitionResourceIDCacheKeyTTL defines how long a cached role definition is considered valid.
+// roleDefinitionResourceIDCacheKeySuccessTTL defines how long a cached successful GetByID is valid.
 // After this TTL (6 hours), the cached entry is stale and the next GetCachedByID refreshes it from Azure.
-const roleDefinitionResourceIDCacheKeyTTL = 6 * time.Hour
+const roleDefinitionResourceIDCacheKeySuccessTTL = 6 * time.Hour
+
+// roleDefinitionResourceIDCacheKeyErrorTTL defines how long a cached GetByID error is valid.
```
miguelsorianod: a cached getByID error -> a cached GetByID that returned an error

backend/pkg/azure/cachedreader/role_definitions.go
miguelsorianod, line 49: need to update the description of the type to explain the new TTL behavior

miguelsorianod, line 48: Once comments addressed update unit tests to reflect new behavior

backend/pkg/azure/cachedreader/role_definitions.go
shitaljante: I was explicitly checking whether the cache entry contained an error or a successful response, but returning entry.response, entry.err directly should work the same way.

backend/pkg/azure/cachedreader/role_definitions.go
shitaljante: same as https://github.com/Azure/ARO-HCP/pull/5488#discussion_r3353983642

backend/pkg/azure/cachedreader/role_definitions.go
shitaljante: Updated to return the error as-is.

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -141,20 +146,22 @@ func TestRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
...
 				{
+					advanceClockBy:   roleDefinitionResourceIDCacheKeyErrorTTL + time.Second,
 					roleDefinitionID: testRoleDefinitionRID,
 					wantResponse:     cachedResponse,
 				},
 			},
 		},
 		{
-			name: "refreshes expired cache entry from inner client",
+			name: "successfully fetches role definition after error TTL expired",
```
miguelsorianod: What is the difference now between `successfully fetches role definition after error TTL expired` and `refreshes expired error cache entry from inner client`. It doesn't seem correct as they are both equal now.

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -102,7 +100,9 @@ func TestRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
 			name: "propagates azure api interaction error when it occurs",
```
miguelsorianod: propagates -> propagates and caches

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -168,12 +175,13 @@ func TestRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
...
+					wantErrContains:  "temporary",
 				},
 				{
-					advanceClockBy:   roleDefinitionResourceIDCacheKeyTTL + time.Second,
+					advanceClockBy:   roleDefinitionResourceIDCacheKeyErrorTTL + time.Second,
 					roleDefinitionID: testRoleDefinitionRID,
-					wantResponse:     refreshedCached,
+					wantResponse:     cachedResponse,
 				},
 			},
```
miguelsorianod, line 210: I am missing tests that correctly verify that the TTL differences between success and failure

shitaljante: added a test case for error and then success TTL

backend/pkg/azure/cachedreader/role_definitions_test.go
```diff
@@ -116,12 +117,25 @@ func TestRoleDefinitionsCachedReader_GetCachedByID(t *testing.T) {
...
+					advanceClockBy:   roleDefinitionResourceIDCacheKeyErrorTTL - 2*time.Minute,
+					roleDefinitionID: testRoleDefinitionRID,
+					wantError:        true,
+					wantErrContains:  "service unavailable",
 				},
 			},
 		},
 		{
-			name: "error is not cached; next call retries",
+			name: "recovers after error TTL and caches success for success TTL",
```
shitaljante, line 138: @miguelsorianod updated changes. 
Now this case should cover the behaviour between the timeline of error TTL and success TTL

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -44,6 +47,9 @@ type RoleDefinitionsCachedReader interface {
 // roleDefinitionsCachedReader wraps an uncached azureclient.RoleDefinitionsClient. GetCachedByID
 // responses are cached in memory with a TTL and singleflight deduplication; other RoleDefinitionsClient
 // methods are not exposed on this type.
+// TTLs used for cached role definition entries.
+// Successful lookups are cached for 6 hours (roleDefinitionResourceIDCacheKeySuccessTTL).
+// Error responses are cached for 5 minutes (roleDefinitionResourceIDCacheKeyErrorTTL).
```
tuxerrante, line 52: Nit: the type-level doc block restates the literal TTL values ("6 hours", "5 minutes") that are already in the const comments above. If either TTL changes in the future, these lines go stale silently. Consider referencing the const names without repeating the values:

```go
// Successful lookups use roleDefinitionResourceIDCacheKeySuccessTTL.
// Error responses use roleDefinitionResourceIDCacheKeyErrorTTL.
```

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -32,9 +31,13 @@ import (
...
 
-// roleDefinitionResourceIDCacheKeyTTL defines how long a cached role definition is considered valid.
+// roleDefinitionResourceIDCacheKeySuccessTTL defines how long a cached successful GetByID is valid.
 // After this TTL (6 hours), the cached entry is stale and the next GetCachedByID refreshes it from Azure.
-const roleDefinitionResourceIDCacheKeyTTL = 6 * time.Hour
+const roleDefinitionResourceIDCacheKeySuccessTTL = 6 * time.Hour
+
+// roleDefinitionResourceIDCacheKeyErrorTTL defines how long a cached GetByID that returned an error is valid.
+// After this TTL (5 minutes), the cached entry is stale and the next GetCachedByID retries Azure.
+const roleDefinitionResourceIDCacheKeyErrorTTL = 5 * time.Minute
```
tuxerrante, line 40: What drove the choice of 5 minutes for the error TTL?

Some Azure ARM context that may be relevant:
- The Azure SDK for Go [retry policy](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/policy) already retries transient errors (408, 500, 502, 503, 504) with exponential backoff before surfacing them. By the time your code sees the error, the SDK may have retried 3 times over ~30 seconds — so the error likely indicates a sustained issue, not a blip.
- ARM throttling (429) includes a `Retry-After` header that can be as short as a few seconds ([docs](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling)). Caching a 429 for 5 minutes may impose a longer penalty than Azure intended.
- Azure RBAC role definition propagation can take [10-30 minutes](https://learn.microsoft.com/en-us/azure/role-based-access-control/troubleshooting) — a 5-minute cached 404 during propagation is negligible vs. 30, but worth knowing about.

Was 5 minutes chosen empirically from observing error patterns, or is it a reasonable first guess that could be tuned later?

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -83,7 +91,7 @@ func (c *roleDefinitionsCachedReader) GetCachedByID(ctx context.Context, roleDef
 	c.roleDefinitionsCacheLock.RLock()
 	entry := c.roleDefinitionsCache[roleDefinitionResourceID]
 	c.roleDefinitionsCacheLock.RUnlock()
-	return entry.response, nil
+	return entry.response, entry.err
```
tuxerrante, line 94: The previous implementation wrapped errors with the role definition resource ID:
```go
fmt.Errorf("failed to get role definition for '%s': %w", roleDefinitionResourceID, err)
```

Now the raw Azure error is returned as-is — consistent with the "transparent cache" contract. But if Azure returns a generic error (e.g. "InternalServerError"), the resource ID context is gone from the error chain.

Does the caller of `GetCachedByID` add the resource ID back into its own error handling / structured logging? If not, during an Azure regional issue affecting role definitions, oncall won't be able to correlate which role definitions are failing with which clusters.

(This may be a non-issue if the caller already logs the ID — just flagging it since the context was explicitly removed here.)

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -98,13 +106,11 @@ func (c *roleDefinitionsCachedReader) ensureCachedGetByID(ctx context.Context, r
 	}
 	_, err, _ := c.sfGroup.Do(roleDefinitionResourceID, func() (interface{}, error) {
 		resp, err := c.inner.GetByID(ctx, roleDefinitionResourceID, options)
-		if err != nil {
-			return nil, utils.TrackError(fmt.Errorf("failed to get role definition for '%s': %w", roleDefinitionResourceID, err))
-		}
 		c.roleDefinitionsCacheLock.Lock()
 		defer c.roleDefinitionsCacheLock.Unlock()
 		c.roleDefinitionsCache[roleDefinitionResourceID] = cachedGetByIDResponse{
```
tuxerrante, line 111: The cache map has no size limit or eviction beyond TTL-based staleness (and stale entries are only cleaned up on access, not proactively).

Currently the role definition IDs are hardcoded built-in Azure role UUIDs (~12 entries from `internal/azure/cluster_scoped_identities_config.go`), so unbounded growth is unlikely today. But if future callers pass dynamically constructed IDs (e.g. custom roles per cluster), the map could grow without bound — especially error entries, which are small (~100 bytes each) but never evicted until re-accessed after TTL.

In a pathological scenario where a bug in resource ID construction generates thousands of unique malformed IDs, each error entry lives until accessed again — effectively forever if no one re-requests that exact ID.

Worth considering a soft cap or periodic GC sweep — even a simple `if len(cache) > maxEntries { clear oldest }` in the write path would prevent silent memory growth.

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -98,13 +106,11 @@ func (c *roleDefinitionsCachedReader) ensureCachedGetByID(ctx context.Context, r
...
 			response:   resp,
+			err:        err,
```
tuxerrante, line 113: The decision to treat all errors the same (per earlier discussion) trades simplicity for precision. For awareness, here's how different Azure error types interact with a flat 5-minute cache:

| Error type | Nature | 5-min cache effect |
|---|---|---|
| 404 NotFound | Permanent (unless creating) | Correct — avoids hammering |
| 429 Throttled | Transient, has `Retry-After` | May overshoot Azure's ask |
| 401/403 Auth | Changes on credential rotation | Delays recovery after refresh |
| 500/502/503 | Usually resolves in seconds | Long after SDK already retried |
| `context.Canceled` | Not an Azure error | Should arguably never be cached |

Not a request to change — the simplicity is valuable, especially with only ~12 hardcoded role IDs today. Just making sure the tradeoff is consciously accepted.

backend/pkg/azure/cachedreader/role_definitions.go
```diff
@@ -98,11 +107,15 @@ func (c *roleDefinitionsCachedReader) ensureCachedGetByID(ctx context.Context, r
...
-		}
 		c.roleDefinitionsCacheLock.Lock()
 		defer c.roleDefinitionsCacheLock.Unlock()
+		if err != nil {
+			c.roleDefinitionsCache[roleDefinitionResourceID] = cachedGetByIDResponse{
+				err:        utils.TrackError(fmt.Errorf("failed to get role definition for '%s': %w", roleDefinitionResourceID, err)),
+				lastUpdate: c.clock.Now().UTC(),
+			}
+			return nil, nil
+		}
```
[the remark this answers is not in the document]
tuxerrante: +1 on this concern. When multiple goroutines coalesce via singleflight, only the first goroutine's context is used for the actual `GetByID` call. If that context is cancelled (e.g. webhook timeout), the cancellation error is cached for 5 minutes, and all subsequent callers — with healthy contexts — get served `context.Canceled` from cache.

Since this component doesn't have active call sites yet (PR #4403 introduced it), the practical impact depends on the concurrency profile of the future callers. But a one-line guard would make this safe regardless:

```go
resp, err := c.inner.GetByID(ctx, roleDefinitionResourceID, options)
if ctx.Err() != nil {
    return nil, nil // don't cache context lifecycle errors
}
```

Has there been observation of context cancellations in existing (non-cached) role definition lookups in production?

backend/pkg/azure/cachedreader/role_definitions.go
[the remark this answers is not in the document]
tuxerrante: To expand on this: since the callback always returns `nil, nil`, the error path on lines 118-120 (`utils.TrackError(err)`) is effectively dead code — `err` from `sfGroup.Do` is always nil.

The design is intentional: results (including errors) flow through the cache map, not through singleflight's return channel. But this contract is non-obvious to future maintainers. A short comment on line 116 would make it explicit:

```go
// Results flow through the cache map; singleflight is used only for deduplication.
return nil, nil
```

backend/pkg/azure/cachedreader/role_definitions.go
miguelsorianod: Yes, we are aware of it and we traded to simplicity at least for now :+1:

backend/pkg/azure/cachedreader/role_definitions.go
miguelsorianod: Yes, we are also aware of this. Right now the role definition retrieval is limited to the azure cluster scoped identities configuration which has a few roles. Those are the roles that we limit interaction to.
If in the future there's a new need this should be revisited. Probably changing to a lru at that point would be the approach

backend/pkg/azure/cachedreader/role_definitions.go
miguelsorianod: @shitaljante could you address this in a new MR?

backend/pkg/azure/cachedreader/role_definitions.go
miguelsorianod: It's a reasonable first guess that could be tuned later

backend/pkg/azure/cachedreader/role_definitions.go
miguelsorianod: The caller can add the detail needed for clarity. We can't return the error decorated as it's a transparent cache

backend/pkg/azure/cachedreader/role_definitions.go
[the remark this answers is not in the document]
shitaljante: good point, the callback always returns err as nil, results including Azure errors flow through the cache map, not singleflight's return channel. I'll add the suggested comment in the follow-up PR

backend/pkg/azure/cachedreader/role_definitions.go
shitaljante: I'll update the changes in new PR.

PR #5436, opened 2026-05-29
feat(fleet): add pipeline-based management cluster registration

geoberle:
Adds the fleet register CLI command and its pipeline integration, and removes the ManagementClusterMigrationController

The register command writes Stamp and ManagementCluster documents directly to CosmosDB during management cluster provisioning — the same data the migration controller previously synced from ClustersService, but driven from the pipeline instead.

### What this PR does:

- fleet register subcommand that creates/updates Stamp and ManagementCluster documents in CosmosDB
- Registration pipeline (fleet/registration/) that runs the command as a Kubernetes Job during management cluster provisioning
- Deterministic provision shard ID generation via UUID v5 same namespace UUID and AKS cluster name as the existing cluster-service Makefile, so both paths always produce the same ID
- Removes the ManagementClusterMigrationController from the backend -— the register command now owns Stamp and ManagementCluster document creation
- AdminAPI endpoint for fleet data access + E2E test validating registration succeeded

### What this PR does NOT touch:

- ClustersService provision shard registration (still helm-based)
- Maestro consumer registration (still via maestro agent job)
- These will move to fleet controllers in follow-up PRs

https://redhat.atlassian.net/browse/ARO-26925

### Testing

The `Fleet should have registered stamps with ready management clusters` E2E test verifies that registration succeeded and an approved stamp and ready mgmt cluster are available in cosmos DB.

### PR Checklist
- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
- [ ] Screenshots included (if graph/UI/metrics changes)
- [ ] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [ ] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

PR #5495, opened 2026-06-03
Revert "feat(fleet): add pipeline-based management cluster registration"

geoberle:
Reverts Azure/ARO-HCP#5436

we need the backfill functionality for the status.KubeApplierCosmosContainerName field in mgmt cluster docs running within the backend so even a partial backend rollout ensures the field is set and kube-applier works.


raelga (approved):
/lgtm
/approve

raelga:
Rebased onto latest `main` to resolve the conflict in the generated `docs/pipelines.md` (regenerated from `topology.yaml`). Force push (`--force-with-lease`) was needed because the rebase rewrote the commit history, so the old remote commit is no longer an ancestor.

raelga:
Force-merge justification: this revert is required before we can safely continue the prod promotion. It restores the backend-side backfill for `status.KubeApplierCosmosContainerName` in management cluster documents, which is needed to recover full kube-applier functionality in STAGE/PROD even if a rollout only partially succeeds.

The current STAGE/PROD state is missing this field because the last successful full rollout happened before the field existed, later rollouts failed at least partially, and the backfill was later moved out of the backend into pipeline-based registration. Keeping the backfill in the backend is the safer recovery path because a successful backend rollout is enough to repair the existing management cluster documents.

This PR has already passed an E2E run. However, Tide/CI has been unreliable and slow today, with repeated multi-hour batch failures and a low probability of getting two consecutive green batches quickly. We are also at the end of the week and need to proceed with the prod rollout to restore full functionality rather than continue waiting indefinitely on flaky CI.

We are aware this may interrupt the currently running Tide batch and accept that trade-off.

davidffrench:
Approved force-merge.

Restoring production is of the highest priority and saving hours this close to the weekend to increase the chance of a successful prod rollout justifies the force merge.

PR #5502, opened 2026-06-03
Switch to changefeed instead of repeated list

deads2k:
Using changefeeds to drive a watch mechanism for resources and kubeapplier.  This will improve our efficiency.  We will resolve our deletion behavior in the near future.


deads2k:
This had interesting impacts on timing. Seems we aren't getting deletes via the changefeed, so relisting remains necessary, with at least two cycles to see content is actually removed. We could choose what latency that will be, but it will be.

Alternatively, we could choose to leave a "not really here" marker that our lister could interpret as "not here".

deads2k:
Neat test we could do. If we combine a replace and a delete in a single transaction, do we get a changefeed notification for the replace.  If so, we could make that particular field into a synthetic delete on the watch channel.

mbarnes:
> Neat test we could do. If we combine a replace and a delete in a single transaction, do we get a changefeed notification for the replace. If so, we could make that particular field into a synthetic delete on the watch channel.

With the Go SDK currently only supporting "[latest version](https://learn.microsoft.com/en-us/azure/cosmos-db/change-feed-modes?tabs=latest-version#latest-version-change-feed-mode)" mode, I'd expect no notification.  (There's no "latest version" of a deleted item.)  If the SDK ever supports "[all versions and deletes](https://learn.microsoft.com/en-us/azure/cosmos-db/change-feed-modes?tabs=latest-version#all-versions-and-deletes-change-feed-mode)" mode, then we should get separate notifications for the replace and the delete.

It's frustrating to be stuck with the lesser capable mode.  I'm going to approach upstream about it.

backend/pkg/informers/cosmos_db_list_watch.go
```diff
@@ -0,0 +1,1208 @@
+// Copyright 2025 Microsoft Corporation
```
machi1990: 2026

backend/pkg/informers/cosmos_db_list_watch.go
```diff
@@ -0,0 +1,1208 @@
...
+		}
+
+		c.continuationTokens.Store(feedRange, continuationToken)
+	}
+
+	return nil
+}
+
+//
+//// toKey converts an Azure resource ID to a lowercase
```
machi1990: Do we plan to clean these comments?

machi1990 (commented):
Took the initial look.

Another approach we could take on the delete part is what's described here https://learn.microsoft.com/en-us/azure/cosmos-db/change-feed-modes?tabs=latest-version#features-of-each-mode for the latest version mode

> This mode of change feed doesn't log deletes. You can capture deletes by setting a "soft-delete" flag within your items instead of deleting them directly. For example, you can add an attribute in the item called deleted with the value true, and then set a Time to Live (TTL) on the item. The change feed captures it as an update and the item is automatically deleted when the TTL expires. Alternatively, you can set a finite expiration period for your items by using the [TTL capability](https://learn.microsoft.com/en-us/azure/cosmos-db/time-to-live). With this solution, you have to process the changes within a shorter time interval than the TTL expiration period.

backend/pkg/informers/informers.go
```diff
@@ -50,57 +51,38 @@ func (listWatchWithoutWatchListSemantics) IsWatchListSemanticsUnSupported() bool
...
-	ControllerRelistDuration               = 30 * time.Second
+	SubscriptionRelistDuration             = 30 * time.Minute
+	ClusterRelistDuration                  = 30 * time.Minute
+	NodePoolRelistDuration                 = 30 * time.Minute
+	ExternalAuthRelistDuration             = 30 * time.Minute
+	ServiceProviderClusterRelistDuration   = 30 * time.Minute
+	ServiceProviderNodePoolRelistDuration  = 30 * time.Minute
+	ControllerRelistDuration               = 30 * time.Minute
+	AllOperationsRelistDuration            = 30 * time.Minute
+	ActiveOperationsRelistDuration         = 30 * time.Minute
```
deads2k, line 63: this may do funky things to deletion now that deletion is checking.  If that happens we'll tune these back down.

internal/database/informers/changefeed_list_watch.go
```diff
@@ -0,0 +1,450 @@
...
+
+	select {
+	case <-c.done:
+		cancel(fmt.Errorf("watch closed"))
+	case <-ctx.Done():
+	}
+}
+
+// TODO this breaks on the delete and recreate scenario. We need to add a true UUID.
+func (c *ChangeFeedWatcher[InternalAPIType, InternalAPITypePointer, CosmosAPIType]) beginDeliveryToWatcher(resourceIDToInitialInstanceVersion *sync.Map) {
```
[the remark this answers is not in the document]
machi1990: How likely are we to be bitten by this? 
e.g if I create a cluster; its gets ready, and I delete it. Then recreate it; what will it look then in terms of the change field; What's the actual problem of not having this?

backend/pkg/informers/cosmos_db_list_watch.go
```diff
@@ -0,0 +1,450 @@
...
+
+	logger := utils.LoggerFromContext(ctx)
+	logger.Info("starting change feed watcher")
+	defer logger.Info("finished change feed watcher")
+
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
+	ctx, cancel := context.WithCancelCause(ctx)
+	defer cancel(fmt.Errorf("finished"))
```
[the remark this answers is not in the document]
machi1990: This makes sense

backend/pkg/informers/cosmos_db_list_watch.go
```diff
@@ -0,0 +1,450 @@
...
+		resourceIDToInstanceVersion.Store(strings.ToLower(currObj.GetResourceID().String()), currObj.GetInstanceVersion())
+
+		list.Items = append(list.Items,
+			runtime.RawExtension{
+				Object: any(currObj).(runtime.Object),
+			})
+	}
+	if err := iter.GetError(); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
machi1990: I believe this handled via 
```
prevFeedWatcher.Stop()
```

In https://github.com/deads2k/ARO-HCP/blob/1093118eca56c6148de31ff7b9ab62e1dbdddca6/backend/pkg/informers/cosmos_db_list_watch.go#L86

backend/pkg/informers/cosmos_db_list_watch.go
```diff
@@ -0,0 +1,450 @@
...
+
+	c.currentWatcher = newChangeFeedWatcher[InternalAPIType, InternalAPITypePointer, CosmosAPIType](c.desiredResourceTypes, c.clock, c.cosmosClient, c.clock.Now(), c.relistDuration)
+	go c.currentWatcher.Run(ctx)
+
+	resourceIDToInstanceVersion := &sync.Map{}
+
+	iter, err := c.globalLister.List(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
machi1990: Same as https://github.com/Azure/ARO-HCP/pull/5502/changes#r3442763926

machi1990:
Ready to tag once:
- the result channel is closed 
- the integration test failure which looks related and e2e are addressed.
The e2e are all constantly failing with
```
  "error": {
        "message": "timed out after 59s waiting for resolution of desired version from 'x.y.z' node pool version"
      }
```

internal/database/informers/changefeed_list_watch.go
```diff
@@ -0,0 +1,450 @@
...
+		if changeFeedStatus == http.StatusOK {
+			for _, doc := range response.Documents {
+				err = c.processDocument(ctx, doc)
+				if err != nil {
+					return err
+				}
+			}
+		}
+
+		// Do not record the new continuation token until we have successfully
```
stevekuznetsov, line 461: Do we care about double-processing?

deads2k: the processDocument tracks last instanceVersion, so we don't double process.  We double consume here, but don't double process

backend/pkg/informers/informers.go
```diff
@@ -50,57 +51,38 @@ func (listWatchWithoutWatchListSemantics) IsWatchListSemanticsUnSupported() bool
...
-		},
-	}
+func NewSubscriptionInformerWithRelistDuration(lister database.GlobalLister[arm.Subscription], cosmosClient database.ResourcesDBClient, relistDuration time.Duration) cache.SharedIndexInformer {
+	lw := NewChangeFeedListWatcher[arm.Subscription, *arm.Subscription, database.GenericDocument[arm.Subscription]](
+		[]azcorearm.ResourceType{azcorearm.NewResourceType("Microsoft.Resources", "subscriptions")},
+		utilsclock.RealClock{},
+		lister,
+		cosmosClient,
+		relistDuration,
+	)
```
mbarnes, line 83: Minor, but since we're passing a `ResourcesDBClient` directly now, you could eliminate the lister parameter.  Even in test scenarios, the lister is derived from the DB client.  Similarly for other informer constructors.

```
func NewSubscriptionInformerWithRelistDuration(cosmosClient database.ResourcesDBClient, relistDuration time.Duration) cache.SharedIndexInformer {
	lw := NewChangeFeedListWatcher[arm.Subscription, *arm.Subscription, database.GenericDocument[arm.Subscription]](
		[]azcorearm.ResourceType{azcorearm.NewResourceType("Microsoft.Resources", "subscriptions")},
		utilsclock.RealClock{},
		cosmosClient.ResourcesGlobalListers().Subscriptions(),
		cosmosClient,
		relistDuration,
	)
```

backend/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Azure/ARO-HCP/internal v0.0.0-00010101000000-000000000000
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.21.1
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.13.1
-	github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos v1.4.2
+	github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos v1.5.0-beta.4
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2 v2.2.0
```
mbarnes, line 10: Should bump the version to latest, which as of today is [v1.5.0-beta.7](https://github.com/Azure/azure-sdk-for-go/releases/tag/sdk%2Fdata%2Fazcosmos%2Fv1.5.0-beta.7).  Contains a number of bug fixes and efficiency improvements since beta-4.

internal/database/informers/changefeed_list_watch.go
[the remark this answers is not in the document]
mbarnes: I assume it's that we won't get a deletion notification from the change feed; we notice deletion from periodic relists.  If the cluster is recreated with the same name then we may get a "latest version" notification for the new cluster from the change feed and never notice the deletion by relist.  I don't know exactly what would happen next but we'd be treating it as the same cluster; that seems potentially bad.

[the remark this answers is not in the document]
mbarnes: Given how long cluster or node pool deletion then recreation takes, I would think we can tune relists to be frequent enough to notice deletions in time.  External auth may be another story.

backend/go.mod
deads2k: looks like it fixed the panic

deads2k:
this PR breaks
```
    RESPONSE 200: 200 OK
    {
      "id": "/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationStatuses/fa7508ca-2857-4b50-ab50-1841044fef24",
      "name": "fa7508ca-2857-4b50-ab50-1841044fef24",
      "status": "Failed",
      "startTime": "2026-06-24T14:17:07.868801073Z",
      "endTime": "2026-06-24T14:18:17.995990692Z",
      "error": {
        "message": "timed out after 59s waiting for resolution of desired version from '4.21.21' node pool version"
      }
```

somehow.  I bet it's expecting a resync that never comes

stevekuznetsov:
Deletion via tombstone update + some garbage collection controller that would allow all of our controllers to respond to changefeed events as opposed to a hybrid with relists for deletion likely leaves us in a less tenuous state

deads2k:
Reminder to self, we're looking for ERRORs from data_dump.go for mismatches between cache and live nodepools

deads2k:
huh, looks like the clusters really didn't complete.

/retest

deads2k:
replaced by https://github.com/Azure/ARO-HCP/pull/6020

PR #5505, opened 2026-06-03
defer utilruntime.HandleCrash() in production goroutines

deads2k:
Codify the convention in CLAUDE.md and add the defer to every `go func` / `go fn(...)` launched in non-test code, so a panic inside a goroutine honors utilruntime.ReallyCrash instead of silently bringing down the process. Wait helpers (wait.Until etc.) already wrap in HandleCrash and are left alone.


bennerv (approved):
/lgtm

PR #5506, opened 2026-06-03
Align cluster admission with node-pool per-struct pattern

deads2k:
Cleans up some cruft so future authors can follow the shared pattern.

### DIFF
--- a/internal/admission/utils.go
+++ b/internal/admission/utils.go
@@ -29,24 +29,48 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils/apihelpers"
 )
 
-// ValidateClusterNodePoolsMinorVersionSkew lists all node pools for the cluster in Cosmos and checks clusterVersion against each pool
-// using the same skew rules (n-2 minor, cross-major allowlist) for:
-//   - the customer node pool properties.version.id (when non-empty),
-//   - the service provider node pool lowest and highest active versions when they exist.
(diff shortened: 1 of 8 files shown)


internal/admission/utils.go
```diff
@@ -29,24 +29,48 @@ import (
...
-// ValidateClusterNodePoolsMinorVersionSkew lists all node pools for the cluster in Cosmos and checks clusterVersion against each pool
-// using the same skew rules (n-2 minor, cross-major allowlist) for:
-//   - the customer node pool properties.version.id (when non-empty),
-//   - the service provider node pool lowest and highest active versions when they exist.
+// ValidateClusterNodePoolsMinorVersionSkew lists all node pools for the cluster
+// in Cosmos (along with their service provider records) and delegates to
+// admitClusterNodePoolsMinorVersionSkew. The frontend cluster admission path
+// prefetches the same list into ClusterAdmissionContext.ClusterNodePools and
+// calls the inner helper directly; this entry point exists for callers that
+// don't have an admission context handy (currently the backend upgrade
```
machi1990: @deads2k let's also refactor https://github.com/deads2k/ARO-HCP/blob/b86bbf35e85247552f200dff91efec59849495bd/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go#L293 to align with what's done here and remove `ValidateClusterNodePoolsMinorVersionSkew` that takes the DB handle.

deads2k: Pushed.

PR #5514, opened 2026-06-04
feat: introduce asynchronous cluster deletion in backend in a disabled state

miguelsorianod:
The backend logic to perform asynchronous cluster deletion is disabled for now: all new controllers and the updated cluster delete operation controller branch on UsesNewClusterDeletionApproach, which is not set by the frontend yet. Legacy synchronous deletion behavior is preserved via legacySynchronizeOperation in operation_cluster_delete.

Changes:
- Add ClusterClusterServiceDeleteDispatchController to call Cluster Service DeleteCluster after DeletionTimestamp is set. It also stamps ClusterServiceDeletionTimestamp
- Add ClusterClusterServiceIDClearerController to clear ClusterServiceID once from the Cluster's serviceProviderProperties when Cluster Service reports the cluster is gone
- Add ClusterChildResourcesCleanupController to recursively delete Cosmos child resources under the cluster with gates for Maestro bundles cleanup as they depend on ServiceProviderCluster, and ignoring controllers documents as well as the entire subtree of NodePools and ExternalAuths as those are deleted by their corresponding deletion processes
- The ClusterChildResourcesCleanupController does not proceed until the Cluster's NodePools and ExternalAuths are fully gone from Cosmos. This is intentional to prevent deleting cluster child resources that the NodePool and ExternalAuth deletion processes might depend on
- Add ClusterDeletionController to delete the Cluster Cosmos document once CS cleanup, Maestro bundles cleanup and child resources cleanup are completed. The controller does not proceed until ExternalAuth and NodePool Cosmos documents are gone. The controller also ignores the child resources of ExternalAuth and NodePool resources. The controller also marks the cluster billing document as deleted before deleting the Cluster Cosmos document itself
- Both ClusterChildResourcesCleanupController and ClusterDeletionController skip checking the child resources of NodePool and ExternalAuth resources, as those should be cleant by their corresponding deletion processes
- Update OperationClusterDeleteController to complete operations when the Cluster Cosmos document is removed (new path), as well as keep CS polling to update the cluster delete operation and the cluster delete status from CS state. The controller also keeps the legacy cluster delete operation controller behavior by branching depending on UsesNewClusterDeletionApproach

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/backend/pkg/controllers/billingcontrollers/create_billing_doc.go
+++ b/backend/pkg/controllers/billingcontrollers/create_billing_doc.go
@@ -63,6 +63,8 @@ func (c *createBillingDoc) NeedsWork(ctx context.Context, existingCluster *api.H
 		return false
 	}
 
+	// TODO should we add a cluster.DeletionTimestamp == nil check here?
+
 	// Skip if the billing document has already been created (BillingDocumentCosmosID is set)
 	if len(existingCluster.ServiceProviderProperties.BillingDocumentCosmosID) != 0 {
 		return false
--- a/backend/pkg/controllers/billingcontrollers/create_billing_doc.go
+++ b/backend/pkg/controllers/billingcontrollers/create_billing_doc.go
@@ -63,8 +63,6 @@ func (c *createBillingDoc) NeedsWork(ctx context.Context, existingCluster *api.H
 		return false
 	}
 
-	// TODO should we add a cluster.DeletionTimestamp == nil check here?
-
 	// Skip if the billing document has already been created (BillingDocumentCosmosID is set)
 	if len(existingCluster.ServiceProviderProperties.BillingDocumentCosmosID) != 0 {
 		return false
--- a/backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
+++ b/backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
@@ -39,23 +39,26 @@ import (
 // deletion pipelines. The orphan scraper handles controller status after the
 // Cluster document itself is removed.
 type clusterChildResourcesCleanupController struct {
-	cooldownChecker   controllerutil.CooldownChecker
-	clusterLister     listers.ClusterLister
-	resourcesDBClient database.ResourcesDBClient
+	cooldownChecker      controllerutil.CooldownChecker
+	clusterLister        listers.ClusterLister
+	resourcesDBClient    database.ResourcesDBClient
+	kubeApplierDBClients database.KubeApplierDBClients
 }
 
 var _ controllerutils.ClusterSyncer = (*clusterChildResourcesCleanupController)(nil)
 
 func NewClusterChildResourcesCleanupController(
 	resourcesDBClient database.ResourcesDBClient,
+	kubeApplierDBClients database.KubeApplierDBClients,
 	activeOperationLister listers.ActiveOperationLister,
 	informers informers.BackendInformers,
 ) controllerutils.Controller {
 	_, clusterLister := informers.Clusters()
 	syncer := &clusterChildResourcesCleanupController{
-		cooldownChecker:   controllerutils.DefaultActiveOperationPrioritizingCooldown(activeOperationLister),
-		clusterLister:     clusterLister,
-		resourcesDBClient: resourcesDBClient,
+		cooldownChecker:      controllerutils.DefaultActiveOperationPrioritizingCooldown(activeOperationLister),
+		clusterLister:        clusterLister,
+		resourcesDBClient:    resourcesDBClient,
+		kubeApplierDBClients: kubeApplierDBClients,
 	}
 
 	return controllerutils.NewClusterWatchingController(
@@ -114,7 +117,11 @@ func (c *clusterChildResourcesCleanupController) SyncOnce(ctx context.Context, k
 	// We must not delete cluster-scoped resources (like ServiceProviderCluster,
 	// ManagementClusterContent) until all nodepools and externalauths are fully
 	// deleted, because their deletion controllers may depend on cluster-scoped
-	// resources.
+	// resources. ServiceProviderCluster also carries ManagementClusterResourceID,
+	// which we need to reach the per-MC kube-applier container, so cluster-scoped
+	// kube-applier *Desires must be removed before the ServiceProviderCluster
+	// document. Nodepool-scoped *Desires are cleaned up by the nodepool deletion
+	// pipeline.
 	allNodePoolsGone, err := deletePreconditionAllNodePoolsDeleted(ctx, c.resourcesDBClient, key)
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("failed to check nodepool precondition: %w", err))
@@ -134,10 +141,6 @@ func (c *clusterChildResourcesCleanupController) SyncOnce(ctx context.Context, k
 	}
 
 	clusterResourceID := cluster.ID
-	untypedCRUD, err := c.resourcesDBClient.UntypedCRUD(*clusterResourceID)
-	if err != nil {
-		return utils.TrackError(fmt.Errorf("failed to create untyped CRUD for cluster children: %w", err))
-	}
 
 	// skipSubtreeTypes lists resource types whose entire subtrees are skipped.
 	// A child resource is left alone if its type path starts with any of these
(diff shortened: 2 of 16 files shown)


backend/pkg/controllers/billingcontrollers/create_billing_doc.go
```diff
@@ -63,6 +63,8 @@ func (c *createBillingDoc) NeedsWork(ctx context.Context, existingCluster *api.H
 		return false
 	}
 
+	// TODO should we add a cluster.DeletionTimestamp == nil check here?
```
miguelsorianod: See this comment

backend/pkg/controllers/clusterdeletion/cluster_deletion_controller.go
```diff
@@ -0,0 +1,266 @@
...
+	if !preconditionMet {
+		return nil
+	}
+
+	// Mark billing document as deleted before deleting the cluster document.
+	// This is idempotent: if the billing document was already marked or never
+	// existed, MarkBillingDocumentDeleted returns nil.
+	// ErrAmbiguousResult (multiple billing docs for one cluster) is a data
+	// integrity issue that retrying won't fix, so we log and continue.
+	err = controllerutils.MarkBillingDocumentDeleted(ctx, c.billingDBClient, cluster.ID, c.passiveClock.Now())
```
miguelsorianod, line 184: This controller for the case of cluster deletion has a different aspect compared to the nodepool and external auth ones:

For nodepools and external auths this kind of controller checks delete preconditions and when those occur the cosmos resource is deleted. This controller does the same but it is also responsible of marking the billing document as deleted. The task seems small enough that I put it in this same controller.

backend/pkg/controllers/clusterdeletion/cluster_deletion_controller.go
```diff
@@ -0,0 +1,266 @@
...
+	preconditionMet, err := c.deletePreconditionAllMaestroClusterScopedReadonlyBundlesCleared(ctx, key)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to check precondition: %w", err))
+	}
+	if !preconditionMet {
+		return nil
+	}
+
+	// Preconditio: all node pool Cosmos docs must be deleted
+	preconditionMet, err = deletePreconditionAllNodePoolsDeleted(ctx, c.resourcesDBClient, key)
```
miguelsorianod, line 153: This precondition is also checked in cluster_child_resources_cleanup_controller.go but I decided to check it here too to be double-safe

backend/pkg/controllers/clusterdeletion/cluster_deletion_controller.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to check precondition: %w", err))
+	}
+	if !preconditionMet {
+		return nil
+	}
+
+	// Precondition: all external auth Cosmos docs must be deleted
+	preconditionMet, err = deletePreconditionAllExternalAuthsDeleted(ctx, c.resourcesDBClient, key)
```
miguelsorianod, line 162: Same comment as https://github.com/Azure/ARO-HCP/pull/5514/changes#r3355372716

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -0,0 +1,271 @@
...
+	clusterResourceID := cluster.ID
+	untypedCRUD, err := c.resourcesDBClient.UntypedCRUD(*clusterResourceID)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to create untyped CRUD for cluster children: %w", err))
+	}
+
+	// skipSubtreeTypes lists resource types whose entire subtrees are skipped.
+	// A child resource is left alone if its type path starts with any of these
+	// types, because those subtrees have their own deletion pipelines.
+	skipSubtreeTypes := []azcorearm.ResourceType{
```
miguelsorianod, line 148: This is a relevant and new difference compared to the nodepool and externalauth corresponding controllers:

nodepools and externalauth resources are child resources of cluster resources. This means that if we don't prevent it, this controller woud recursively delete the nodepool resources and externalauth resources (which also includes their controllers, managementclustercontents, serviceprovidernodepool, ...)

For the case of controllers, aside from excluding specific cosmos resource types (defined in extraDeleteGates), we also define the ability to skip entire subtrees, shown here.

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+	if !c.NeedsWork(cluster) {
+		return nil
+	}
+
+	// We must not delete cluster-scoped resources (like ServiceProviderCluster,
+	// ManagementClusterContent) until all nodepools and externalauths are fully
+	// deleted, because their deletion controllers may depend on cluster-scoped
+	// resources.
+	allNodePoolsGone, err := deletePreconditionAllNodePoolsDeleted(ctx, c.resourcesDBClient, key)
```
miguelsorianod, line 125: Another relevant difference here compared to externalauth and nodepool corresponding controllers is that we gate the cluster child resources cleanup controller to not act until all the nodepools and externalauths cosmos resources are fully deleted.

This is because those resources can depend on cluster-related child resources, so we don't want to delete them until the deletion process of the externalauths and nodepools has completed. For example, we had a nodepool deletion process stage that took care of cleaning up maestro readonly bundles which required having access to the ServiceProviderCluster to see the associated management cluster resource id. Because we've removed the maestro readonly bundles that specific example dependency is not there anymore but it showcases the potential need hence the gating here.

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to check nodepool precondition: %w", err))
+	}
+	if !allNodePoolsGone {
+		logger.Info("waiting for all nodepools to be deleted before cleaning up cluster child resources")
+		return nil
+	}
+
+	allExternalAuthsGone, err := deletePreconditionAllExternalAuthsDeleted(ctx, c.resourcesDBClient, key)
```
miguelsorianod, line 134: Same comment as https://github.com/Azure/ARO-HCP/pull/5514/changes#r3355409075

backend/pkg/controllers/clusterdeletion/cluster_deletion_controller.go
deads2k: Let's keep billing distinct and not couple it here

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
deads2k: That sounds reasonable

deads2k: We should find a way to report some kind status back to be aggregated like, "waiting on X resources of type Y"

deads2k: Oh. And we need to check for nested resources in the kube-applier containers

deads2k: Oh. Very neat thing. We should not delete the serviceprovider cluster until the kube-applier content is gone, since finding that cosmos container requires the management cluster

backend/pkg/controllers/clusterdeletion/cluster_deletion_controller.go
```diff
@@ -0,0 +1,267 @@
...
+		return false, utils.TrackError(fmt.Errorf("failed to get cluster: %w", err))
+	}
+
+	// Orphaned controller status documents may linger under nodepools and
+	// externalauths after their parent documents have been deleted. We skip
+	// those here because the orphan scraper handles them.
+	skipSubtreeTypes := []azcorearm.ResourceType{
+		api.NodePoolResourceType,
+		api.ExternalAuthResourceType,
+	}
```
deads2k, line 241: couldn't we skip just those controller types instead of the entire subtree?

deads2k:
My comment can be a followup

/lgtm
/hold in case there are ordering concerns.  you can release the hold at your discretion

backend/pkg/controllers/clusterdeletion/cluster_deletion_controller.go
miguelsorianod: This was added because if for some reason there are other lingering child resources on that tree other than controllers we don't want to be blocked permanently at this stage. The cleanup process should belong to the corresponding resource. Maybe the comment could be more accurate I can update the comment if we find it relevant.

PR #5515, opened 2026-06-04
Use mode: runtime-selected for e2e-parallel

roivaz:
https://redhat.atlassian.net/browse/ARO-27243

The dev slot catalog was reshaped from several fixed regional pools per subscription to one runtime-selected pool per subscription because the previous fixed-mode direction does not fit within Azure’s resource group limits for the E2E identity pool.

With the current E2E identity-pool model, each fixed regional pool needs its own distinct identity_container_prefix, which means its own dedicated set of MSI container resource groups. That footprint scales quickly: a pool with 15 slots and 20 identity containers consumes about 300 resource groups, so modeling three fixed regions for the same subscription would consume about 900 RGs in that subscription by itself, leaving almost no headroom under Azure’s practical ~980 RG limit.

Reusing the same identity container prefix across fixed regional pools is not a safe alternative, because different Boskos leases would then point at the same underlying E2E identity-pool RG set.

For that reason, dev now uses one runtime-selected pool per customer subscription, with the runtime region selected by the job and the shared E2E identity pool provisioned once in westus3. This keeps the current rollout aligned with the immediate goal of multi-subscription scaling in a single region, while avoiding an unsustainable RG explosion.


raelga (approved):
/lgtm
/approve

PR #5522, opened 2026-06-04
read the domain base prefix from CS and infer the rest of the properties from the hosted cluster cr

machi1990:
### What

read the domain base prefix from CS and infer the rest of the properties from the hosted cluster cr

### Why

We need to continue reading the base domain prefix from CS as it is required by the hostedcluster CreateClusterScopedReadDesires controller.

The start reading the rest of the properties when the hostedcluster cr has been synced from the management cluster.
- the api url is infered from KubeAPIServerDNSName and the ControlPlaneEndpoint.Port
- the console url is inferred from the BaseDomain; this will contain a value even though there are cases where it won't be available up until when the customer creates the nodepool + configure external auth. Attempt to access it will show that the app isn't available.
- the parent base domain is inferred from KubeAPIServerDNSName by stripping api.basedomainprefix from it
- the issuer url is taken from hostedCluster.Spec.IssuerURL

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/backend/pkg/controllers/clusterpropertiescontroller/cluster_base_domain_prefix_sync.go
+++ b/backend/pkg/controllers/clusterpropertiescontroller/cluster_base_domain_prefix_sync.go
@@ -24,6 +24,7 @@ import (
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
 	"github.com/Azure/ARO-HCP/backend/pkg/informers"
 	"github.com/Azure/ARO-HCP/backend/pkg/listers"
+	"github.com/Azure/ARO-HCP/internal/api"
 	controllerutil "github.com/Azure/ARO-HCP/internal/controllerutils"
 	"github.com/Azure/ARO-HCP/internal/database"
 	unionkubeapplierinformers "github.com/Azure/ARO-HCP/internal/database/unioninformers/kubeapplier"
@@ -35,6 +36,7 @@ import (
 // Cluster Service to Cosmos DB when the field is unset.
 type clusterBaseDomainPrefixSyncer struct {
 	cooldownChecker      controllerutil.CooldownChecker
+	clusterLister        listers.ClusterLister
 	resourcesDBClient    database.ResourcesDBClient
 	clusterServiceClient ocm.ClusterServiceClientSpec
 }
@@ -50,8 +52,11 @@ func NewClusterBaseDomainPrefixSyncController(
 	informers informers.BackendInformers,
 	kubeApplierInformers *unionkubeapplierinformers.UnionKubeApplierInformers,
 ) controllerutils.Controller {
+	_, clusterLister := informers.Clusters()
+
 	syncer := &clusterBaseDomainPrefixSyncer{
 		cooldownChecker:      controllerutils.DefaultActiveOperationPrioritizingCooldown(activeOperationLister),
+		clusterLister:        clusterLister,
 		resourcesDBClient:    resourcesDBClient,
 		clusterServiceClient: clusterServiceClient,
 	}
@@ -70,8 +75,28 @@ func (c *clusterBaseDomainPrefixSyncer) CooldownChecker() controllerutil.Cooldow
 	return c.cooldownChecker
 }
 
+func (c *clusterBaseDomainPrefixSyncer) needsWork(existingCluster *api.HCPOpenShiftCluster) bool {
+	if existingCluster.ServiceProviderProperties.ClusterServiceID == nil ||
+		len(existingCluster.ServiceProviderProperties.ClusterServiceID.String()) == 0 {
+		return false
+	}
+
+	return len(existingCluster.CustomerProperties.DNS.BaseDomainPrefix) == 0
+}
+
 func (c *clusterBaseDomainPrefixSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
 	logger := utils.LoggerFromContext(ctx)
+
+	cachedCluster, err := c.clusterLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get cluster from cache: %w", err))
+	}
+	if !c.needsWork(cachedCluster) {
+		return nil
+	}
 
 	clusterCRUD := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName)
 	existingCluster, err := clusterCRUD.Get(ctx, key.HCPClusterName)
@@ -81,30 +106,23 @@ func (c *clusterBaseDomainPrefixSyncer) SyncOnce(ctx context.Context, key contro
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
 	}
-
-	if len(existingCluster.CustomerProperties.DNS.BaseDomainPrefix) > 0 {
+	if !c.needsWork(existingCluster) {
 		return nil
 	}
-
-	if existingCluster.ServiceProviderProperties.ClusterServiceID == nil ||
-		len(existingCluster.ServiceProviderProperties.ClusterServiceID.String()) == 0 {
-		return nil
-	}
-
-	originalCluster := existingCluster.DeepCopy()
 
 	csCluster, err := c.clusterServiceClient.GetCluster(ctx, *existingCluster.ServiceProviderProperties.ClusterServiceID)
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("failed to get cluster from Cluster Service: %w", err))
 	}
 
-	existingCluster.CustomerProperties.DNS.BaseDomainPrefix = csCluster.DomainPrefix()
+	replacement := existingCluster.DeepCopy()
+	replacement.CustomerProperties.DNS.BaseDomainPrefix = csCluster.DomainPrefix()
 
-	if equality.Semantic.DeepEqual(originalCluster, existingCluster) {
+	if equality.Semantic.DeepEqual(existingCluster, replacement) {
 		return nil
 	}
 
-	if _, err := clusterCRUD.Replace(ctx, existingCluster, nil); err != nil {
+	if _, err := clusterCRUD.Replace(ctx, replacement, nil); err != nil {
 		return utils.TrackError(fmt.Errorf("failed to replace Cluster: %w", err))
 	}
 
--- a/backend/pkg/controllers/clusterpropertiescontroller/cluster_base_domain_prefix_sync.go
+++ b/backend/pkg/controllers/clusterpropertiescontroller/cluster_base_domain_prefix_sync.go
@@ -1,0 +1,113 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clusterpropertiescontroller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/backend/pkg/informers"
+	"github.com/Azure/ARO-HCP/backend/pkg/listers"
+	controllerutil "github.com/Azure/ARO-HCP/internal/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/database"
+	unionkubeapplierinformers "github.com/Azure/ARO-HCP/internal/database/unioninformers/kubeapplier"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// clusterBaseDomainPrefixSyncer synchronizes CustomerProperties.DNS.BaseDomainPrefix from
+// Cluster Service to Cosmos DB when the field is unset.
+type clusterBaseDomainPrefixSyncer struct {
+	cooldownChecker      controllerutil.CooldownChecker
+	resourcesDBClient    database.ResourcesDBClient
+	clusterServiceClient ocm.ClusterServiceClientSpec
+}
+
+var _ controllerutils.ClusterSyncer = (*clusterBaseDomainPrefixSyncer)(nil)
+
+// NewClusterBaseDomainPrefixSyncController creates a controller that synchronizes
+// CustomerProperties.DNS.BaseDomainPrefix from Cluster Service to Cosmos DB.
+func NewClusterBaseDomainPrefixSyncController(
+	resourcesDBClient database.ResourcesDBClient,
+	clusterServiceClient ocm.ClusterServiceClientSpec,
+	activeOperationLister listers.ActiveOperationLister,
+	informers informers.BackendInformers,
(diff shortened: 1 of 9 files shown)


machi1990:
/assign @deads2k 

as discussed via DM

machi1990:
@deads2k synced and we need to split the one that reads from hosted cluster onto a separate controller.

Additionally, we always update cosmos if the values differ

machi1990:
/hold 

Until https://github.com/Azure/ARO-HCP/pull/5522#issuecomment-4624560136 is fixed

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -17,146 +17,108 @@ package clusterpropertiescontroller
...
-	if needsIssuerURL {
-		existingCluster.ServiceProviderProperties.Platform.IssuerURL = csCluster.Azure().OidcIssuerUrl()
-	}
-	if needsBaseDomainPrefix {
-		existingCluster.CustomerProperties.DNS.BaseDomainPrefix = csCluster.DomainPrefix()
+	existingCluster.ServiceProviderProperties.Console.URL = fmt.Sprintf("https://console-openshift-console.apps.%s", hostedCluster.Spec.DNS.BaseDomain)
+	existingCluster.ServiceProviderProperties.DNS.BaseDomain = strings.TrimPrefix(
+		hostedCluster.Spec.KubeAPIServerDNSName,
+		fmt.Sprintf("api.%s.", domainPrefix),
+	)
```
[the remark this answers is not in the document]
machi1990: This can't occur, if `BaseDomainPrefix ` is empty, there won't be a hosted cluster read via ReadDesired

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -17,146 +17,108 @@ package clusterpropertiescontroller
...
+	if hostedCluster.Status.ControlPlaneEndpoint.Port != 0 {
+		existingCluster.ServiceProviderProperties.API.URL = fmt.Sprintf("https://%s", net.JoinHostPort(hostedCluster.Spec.KubeAPIServerDNSName, strconv.Itoa(int(hostedCluster.Status.ControlPlaneEndpoint.Port))))
 	}
+	existingCluster.ServiceProviderProperties.Platform.IssuerURL = hostedCluster.Spec.IssuerURL
```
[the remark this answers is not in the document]
machi1990: This is intended. We want to pick up the changes of any of these values when they do change.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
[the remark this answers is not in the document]
machi1990: This can't happen. HostedCluster ReadDesire will only exist if the base domain prefix exists

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync_test.go
```diff
@@ -352,42 +307,34 @@ func newTestCluster(opts ...func(*api.HCPOpenShiftCluster)) *api.HCPOpenShiftClu
...
+	raw, err := json.Marshal(hostedCluster)
+	require.NoError(t, err)
+
+	return &kubeapplier.ReadDesire{
+		CosmosMetadata: api.CosmosMetadata{
+			ResourceID: resourceID,
+		},
+		Spec: kubeapplier.ReadDesireSpec{
+			ManagementCluster: api.Must(azcorearm.ParseResourceID(
+				"/providers/microsoft.redhatopenshift/stamps/1/managementclusters/mgmt-a")),
```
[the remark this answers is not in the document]
machi1990: No need

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync_test.go
```diff
@@ -281,52 +210,78 @@ func TestClusterPropertiesSyncer_SyncOnce(t *testing.T) {
...
-func newTestCluster(opts ...func(*api.HCPOpenShiftCluster)) *api.HCPOpenShiftCluster {
+func newSeededReadDesireLister(ctx context.Context, readDesire *kubeapplier.ReadDesire) (dblisters.ReadDesireLister, error) {
+	mockKubeApplierDB, err := databasetesting.NewMockKubeApplierDBClientWithResources(ctx, []any{readDesire})
+	if err != nil {
+		return nil, err
+	}
+
+	kubeApplierClients := databasetesting.NewMockKubeApplierDBClients()
+	managementClusterID := api.Must(azcorearm.ParseResourceID(
+		"/providers/microsoft.redhatopenshift/stamps/1/managementclusters/mgmt-a"))
```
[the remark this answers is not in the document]
machi1990: no need

backend/pkg/controllers/clusterpropertiescontroller/cluster_base_domain_prefix_sync.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (c *clusterBaseDomainPrefixSyncer) CooldownChecker() controllerutil.CooldownChecker {
+	return c.cooldownChecker
+}
+
+func (c *clusterBaseDomainPrefixSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	logger := utils.LoggerFromContext(ctx)
+
+	clusterCRUD := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName)
+	existingCluster, err := clusterCRUD.Get(ctx, key.HCPClusterName)
```
deads2k: followup to use the needswork pattern with cache first.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -17,146 +17,108 @@ package clusterpropertiescontroller
...
-// SyncOnce performs a single reconciliation of cluster properties.
-// It checks if the Console.URL, DNS.BaseDomain, ManagedIdentitiesDataPlaneIdentityURL,
-// Platform.IssuerURL, or DNS.BaseDomainPrefix fields are unset, and if so, fetches the
-// values from Cluster Service and updates Cosmos.
 func (c *clusterPropertiesSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
 	logger := utils.LoggerFromContext(ctx)
 
-	// Get the cluster from Cosmos
 	clusterCRUD := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName)
 	existingCluster, err := clusterCRUD.Get(ctx, key.HCPClusterName)
```
deads2k: close followup: use only the cache.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -17,146 +17,108 @@ package clusterpropertiescontroller
...
+		return nil
 	}
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
 	}
 
-	// Check if we have a cluster service ID to query
-	if existingCluster.ServiceProviderProperties.ClusterServiceID == nil || len(existingCluster.ServiceProviderProperties.ClusterServiceID.String()) == 0 {
-		return nil
+	hostedCluster, err := maestrohelpers.GetCachedHostedClusterForCluster(ctx, c.readDesireLister, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
```
deads2k, line 99: on not found, return nil. We'll be retriggered when a value is set.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -17,146 +17,108 @@ package clusterpropertiescontroller
...
 
-	// Fetch the cluster from Cluster Service
-	csCluster, err := c.clusterServiceClient.GetCluster(ctx, *existingCluster.ServiceProviderProperties.ClusterServiceID)
-	if err != nil {
-		return utils.TrackError(fmt.Errorf("failed to get cluster from Cluster Service: %w", err))
-	}
-
-	// Take a copy before making changes for comparison
 	originalCluster := existingCluster.DeepCopy()
+	domainPrefix := existingCluster.CustomerProperties.DNS.BaseDomainPrefix
```
deads2k: Empty here needs to be handled in some special way I bet.  Check it early in this sync and probably return?

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -17,146 +17,108 @@ package clusterpropertiescontroller
...
 	}
 
-	// Fetch the cluster from Cluster Service
-	csCluster, err := c.clusterServiceClient.GetCluster(ctx, *existingCluster.ServiceProviderProperties.ClusterServiceID)
-	if err != nil {
-		return utils.TrackError(fmt.Errorf("failed to get cluster from Cluster Service: %w", err))
-	}
-
-	// Take a copy before making changes for comparison
 	originalCluster := existingCluster.DeepCopy()
```
deads2k: I have a PR that is standardizing names as `existing`.  then `replacement := existing.DeepCopy()` and we modify replacement as we go.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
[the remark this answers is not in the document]
deads2k: > This can't occur, if BaseDomainPrefix is empty, there won't be a hosted cluster read via ReadDesired

It can if the other controller doesn't run for some reason.

deads2k (changes requested):
definitely like this. needs some updates.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
machi1990: the helper `maestrohelpers.GetCachedHostedClusterForCluster` does it for us; it returns `nil,nil`

And we return `nil` here https://github.com/machi1990/ARO-HCP/blob/2a93884da203791d5fe8b7fd184a9a15c6ea77e2/backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go#L98

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
machi1990: I didn't need to check for this because when we are here it is because the HostedCluster isn't `nil` https://github.com/machi1990/ARO-HCP/blob/2a93884da203791d5fe8b7fd184a9a15c6ea77e2/backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go#L97 and it'll only be nil if its synced meaning that the domain base prefix used as the read desire target is there.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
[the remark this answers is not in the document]
machi1990: Same reason as https://github.com/Azure/ARO-HCP/pull/5522#discussion_r3358516690 when we reach here it is because the HC cr exists and has been synced; the condition for it is the base domain prefix existing

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
machi1990: When we do https://github.com/Azure/ARO-HCP/pull/5522#discussion_r3358435413 returning early will matter because we might be operating on a stale cache

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -17,146 +17,113 @@ package clusterpropertiescontroller
...
-	if equality.Semantic.DeepEqual(originalCluster, existingCluster) {
+	if equality.Semantic.DeepEqual(existingCluster, replacement) {
 		return nil
 	}
 
-	// Write the updated cluster back to Cosmos
-	if _, err := clusterCRUD.Replace(ctx, existingCluster, nil); err != nil {
+	if _, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Replace(ctx, replacement, nil); err != nil {
 		return utils.TrackError(fmt.Errorf("failed to replace Cluster: %w", err))
 	}
```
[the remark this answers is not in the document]
machi1990: I believe we are fine to use the cache value; if stale, cosmo will reject the update with 409 Conflict because the e-tag is stale and we'll retry and not override values. Are you of the same opinion @deads2k or we follow the bot advise?

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
machi1990: done

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
machi1990: modified and since we are using the cache, I check it early and return nil

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
machi1990: Did it here; we are only using the cache and not performing a read from live DB. 
But the bot seem unhappy about it and I think it is confused. I'd like your input on my reasoning to reject it's comment https://github.com/Azure/ARO-HCP/pull/5522#discussion_r3358697516

backend/pkg/controllers/clusterpropertiescontroller/cluster_base_domain_prefix_sync.go
machi1990: done

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -17,146 +17,122 @@ package clusterpropertiescontroller
...
-	if needsBaseDomainPrefix {
-		existingCluster.CustomerProperties.DNS.BaseDomainPrefix = csCluster.DomainPrefix()
+	replacement.ServiceProviderProperties.DNS.BaseDomain = strings.TrimPrefix(
+		hostedCluster.Spec.KubeAPIServerDNSName,
+		kubeAPIServerDNSNamePrefix,
+	)
+	if hostedCluster.Status.ControlPlaneEndpoint.Port != 0 {
+		replacement.ServiceProviderProperties.API.URL = fmt.Sprintf("https://%s", net.JoinHostPort(hostedCluster.Spec.KubeAPIServerDNSName, strconv.Itoa(int(hostedCluster.Status.ControlPlaneEndpoint.Port))))
 	}
+	replacement.ServiceProviderProperties.Platform.IssuerURL = hostedCluster.Spec.IssuerURL
```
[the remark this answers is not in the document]
machi1990: These fields won't be missing.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -17,146 +17,122 @@ package clusterpropertiescontroller
...
-	if equality.Semantic.DeepEqual(originalCluster, existingCluster) {
+	if equality.Semantic.DeepEqual(existingCluster, replacement) {
 		return nil
 	}
 
-	// Write the updated cluster back to Cosmos
-	if _, err := clusterCRUD.Replace(ctx, existingCluster, nil); err != nil {
+	if _, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Replace(ctx, replacement, nil); err != nil {
 		return utils.TrackError(fmt.Errorf("failed to replace Cluster: %w", err))
 	}
```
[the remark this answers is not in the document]
machi1990: Discussion about this are happening here https://github.com/Azure/ARO-HCP/pull/5522#discussion_r3358697516

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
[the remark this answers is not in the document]
deads2k: We can use the cached value.

PR #5524, opened 2026-06-04
add conditions for clusters, nodepools, and externalauths for us to s…

deads2k:
We'll start by aggregating controller status degraded into a degraded field with some inertia.  But let's start with just the types.


bennerv (approved):
/lgtm

internal/api/types_cluster.go
```diff
@@ -32,6 +32,19 @@ type HCPOpenShiftCluster struct {
 	CustomerProperties        HCPOpenShiftClusterCustomerProperties        `json:"customerProperties,omitempty"`
 	ServiceProviderProperties HCPOpenShiftClusterServiceProviderProperties `json:"serviceProviderProperties,omitempty"`
 	Identity                  *arm.ManagedServiceIdentity                  `json:"identity,omitempty"`
+	Status                    HCPOpenShiftClusterStatus                    `json:"status"`
```
swiencki, line 35: Is dropping the backend-owned `Status` on the next customer PUT/PATCH the intended behavior here, or should it be preserved through the update path once conditions are populated?

PR #5525, opened 2026-06-04
Add degraded summarization controllers for clusters, nodepools, and externalauths

deads2k:
These are for internal use. We could use similar techniques for a condition we'll expose to customers to determine reason and message.

### DIFF
--- a/backend/pkg/controllers/statuscontrollers/nodepool_degraded_aggregator_test.go
+++ b/backend/pkg/controllers/statuscontrollers/nodepool_degraded_aggregator_test.go
@@ -1,0 +1,240 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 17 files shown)


backend/pkg/controllers/statuscontrollers/nodepool_degraded_aggregator_test.go
```diff
@@ -0,0 +1,240 @@
...
+	).Inertia
+
+	// Parent cluster — also needed so the resourcesDBClient is happy when the
+	// node-pool CRUD looks up the parent path internally during Replace.
+	parentClusterID := api.Must(azcorearm.ParseResourceID(
+		"/subscriptions/" + testSubscriptionID +
+			"/resourceGroups/" + testResourceGroupName +
+			"/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/" + testClusterName,
+	))
+	_ = parentClusterID
```
bennerv: what's the point of this?  `_ = parentClusterID`

deads2k: cruft of refactors.

PR #5532, opened 2026-06-05
use 45m timeout for multiversion cluster/nodepool creation

trevorwilliams2025:
[AROSLSRE-1026](https://redhat.atlassian.net/browse/AROSLSRE-1026)

 ## Summary
  - 8 consecutive failures of `e2e-integration-e2e-parallel` (Jun 3–4), all in `complete_cluster_create_multiversion.go` with `context deadline exceeded` against the 20-min
  `ClusterCreationTimeout`/`NodePoolCreationTimeout`.
  - Kusto evidence (INT, hcp-int-uk): clusters reach CS-ready at ~T+7–9 min but ARM `provisioningState` remains `Provisioning` when the test cancels at T+20 min. Total
  time-to-Succeeded for nightly images (4.23/5.0) is ~22–25 min.
  - Introduces dedicated `MultiversionClusterCreationTimeout` / `MultiversionNodePoolCreationTimeout` constants at 45 min so multiversion tests get the headroom they need
  without changing the budget for unrelated suites.

  ## Changes
  - `test/util/framework/constants.go`: new `Multiversion*` constants (45m)
  - `test/e2e/complete_cluster_create_multiversion.go`: 3 call sites switched to the new constants
  - `test/e2e/README.md`: timeout table updated

  ## Test plan
  - [x ] Verify `go build ./test/...` passes
  - [ x] Confirm no other test files reference the new constants (scoped to multiversion only)
  - [ ] Watch next `e2e-integration-e2e-parallel` run for green multiversion results

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -213,6 +213,8 @@ Framework helpers include an API version suffix. See [`test/AGENTS.md`](../AGENT
 |----------|----------|-------------|
 | Provisioning | `ClusterCreationTimeout` | `CreateHCPClusterFromParam20240610`, `CreateHCPClusterAndWait20240610`; preview: `CreateHCPClusterFromParam20251223`, `CreateHCPClusterAndWait20251223` |
 | Provisioning | `NodePoolCreationTimeout` | `CreateNodePoolFromParam20240610`, `CreateNodePoolAndWait20240610`; preview: `CreateNodePoolFromParam20251223`, `CreateNodePoolAndWait20251223` |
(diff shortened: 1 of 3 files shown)


raelga:
cc @mbukatov @mvacula02 

We need to investigate why this test is now taking longer before bumping the timeout over 2x.

/hold

mbukatov:
> We need to investigate why this test is now taking longer before bumping the timeout over 2x.

Agreed, this indeed looks like a workaround. We should never default to this approach.

PR #5536, opened 2026-06-05
Fix systemd logs

janboll:
https://redhat.atlassian.net/browse/ARO-27532

### What/Why

Systemd output ended up in the wrong config file. This moves it to the correct output.conf and hence enables writing to the systemd table again


janboll:
proof it's working: https://dataexplorer.azure.com/clusters/hcp-dev-us-2.eastus2/databases/ServiceLogs?query=H4sIAAAAAAAAAyuuLC5JzU3xyU8v5qpRKM9ILUpVUCjJzE0tLknMLVCwU0hMz9cwNErRBMqmZBaXZOYllygk55QCdRUBABmgc088AAAA
systemdLogs
| where  timestamp > ago(12d)
| distinct cluster

PR #5539, opened 2026-06-05
Gate update completion on customer-properties matching observed status

deads2k:
Before reporting a cluster update operation Succeeded, require that the customer-settable HCPOpenShiftClusterCustomerProperties fields match the observed values mirrored into ServiceProviderCluster.Status. While any desired field still differs from its observed counterpart, the update stays ProvisioningStateAccepted with a message listing the diverging fields, so callers can see exactly which intent has not yet landed.

Today this covers the version profile (id, channelGroup). The diff walker is structured per-struct (mirroring internal/validation) so we can extend the same pattern to the remaining user-settable fields as their observed counterparts are added to ServiceProviderCluster.Status.

/hold no tests


deads2k:
we went in a different direction.

PR #5542, opened 2026-06-07
feat: stagger e2e starts

sclarkso:
### What

test/util/framework/constants.go — Added ClusterCreateStaggerInterval() function that reads E2E_CLUSTER_CREATE_STAGGER_SECONDS env var and returns a time.Duration. Returns 0 (disabled) when unset.

test/util/framework/identities_helper.go — Two changes:
  1. assignNTo() now returns (int, error) instead of error. The int is the count of already-assigned containers before this assignment — the caller's ordinal position in the pool.
  2. AssignIdentityContainers() uses the ordinal to compute a stagger delay (ordinal * staggerInterval) and sleeps before returning, with context cancellation support and logging.

### Why

With E2E_CLUSTER_CREATE_STAGGER_SECONDS=15 and 20 MSI containers, the first test starts immediately, the second waits 15s, the third waits 30s, ..., the 20th waits ~5 minutes. This spreads the CosmosDB write load across the provisioning window instead of a single burst.

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/test/util/framework/constants.go
+++ b/test/util/framework/constants.go
@@ -14,7 +14,11 @@
 
 package framework
 
-import "time"
+import (
+	"os"
+	"strconv"
+	"time"
+)
 
 // When updating timeouts, see test/e2e/README.md#updating-e2e-timeouts.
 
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
/hold

I don't think we want to do this. We should be running as many tests at the same time as our service can handle, and no more. Issues in the backend due to scalability should be found and fixed, instead of trying to reduce load from tests, as long as we are within correct boundaries.

The CosmosDB write load during a cluster lifetime is not specific to the first few seconds, so if we find the scalability limit the correct way to meter out test load is to change the concurrency instead of staggering starts.

sclarkso:
@stevekuznetsov I don't plan to do this.  I am trying to investigate an mtpnc bug, and I want to eliminate concurrency as the cause.

PR #5553, opened 2026-06-08
feat: enable externalauth delete from frontend to backend

miguelsorianod:
Move Cluster Service interaction out of the frontend delete path and hand it off to the backend's async external auth deletion controllers.

The frontend now only:
- Creates a delete operation with UsesNewExternalAuthDeletionApproach set to true
- Sets DeletionTimestamp and sets UsesNewExternalAuthDeletionApproach to true in the ExternalAuth

Removed from the frontend:
- Direct clusterServiceClient.DeleteExternalAuth() call

External Auth Delete operations no longer set an InternalID as it is not needed anymore when using the new external auth deletion approach in backend.

Legacy deletion (operations without UsesNewExternalAuthDeletionApproach) is unchanged and still handled by OperationExternalAuthDeleteController.

### DIFF
--- a/test/e2e/external_auth_list_and_verify.go
+++ b/test/e2e/external_auth_list_and_verify.go
@@ -17,13 +17,16 @@ package e2e
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	"k8s.io/apimachinery/pkg/util/rand"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 
 	hcpsdk "github.com/Azure/ARO-HCP/test/sdk/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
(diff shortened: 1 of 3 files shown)


miguelsorianod:
Checked in the logs occurrences of the different externalauth controllers. For example:

Occurrence of externalauth dispatch controller acting when the externalauth is being deleted because of parent cluster being deleted:
	"msg": "ExternalAuth already being deleted by cluster-service via parent cluster deletion",

Occurrence of externalauth csid clearer controller:
	"msg": "cluster-service ExternalAuth gone. Clearing ClusterServiceID",

Occurrence of externalauth delete controller:
	"msg": "deleting external auth from Cosmos",

Occurrence indicating externalauth operation delete proceeding when corresponding cosmos entry deletion occurs:
	"msg": "external auth document deleted - completing operation",

Ocurren indicating cluster delete operation now waits until externalauths are deleted:
	"msg": "cluster still has cosmos externalauths",

test/e2e/external_auth_list_and_verify.go
```diff
@@ -199,5 +202,39 @@ var _ = Describe("Customer", func() {
 			Expect(*updatedResult.Properties.ProvisioningState).To(Equal(hcpsdk.ExternalAuthProvisioningStateSucceeded), "updated external auth provisioning state should be Succeeded")
 			Expect(*updatedResult.Properties.Claim.Mappings.Username.Prefix).To(Equal(*expectedExternalAuth.Properties.Claim.Mappings.Username.Prefix), "updated external auth Username.Prefix should match expected value")
 
+			By("deleting the external auth")
```
miguelsorianod, line 205: Updated this externalauth test to also do a final intentional deletion of the externalauth and verifications around it.

The reason is because there was currently no e2e test that directly deleted an externalauth. The e2e tests exercised deletion of externalauth indirectly being triggered by its parent cluster deletion, but we should also exercise the direct deletion without cluster deletion.

test/util/framework/constants.go
```diff
@@ -28,7 +28,8 @@ const (
 
 // Deletion timeouts
 const (
-	HCPClusterDeletionTimeout = 45 * time.Minute
+	HCPClusterDeletionTimeout   = 45 * time.Minute
+	ExternalAuthDeletionTimeout = 25 * time.Minute
```
miguelsorianod, line 32: I checked p99 percentiles in different envs over external auth delete operations before this PR over 14 days and the highest I found was 22:53 minutes in prod uksouth. Because of that I set 25 minutes here as the deletion timeout.

miguelsorianod:

PR #5556, opened 2026-06-08
PoC nodepool updates based on CS sha calculation and management cluster state

miguelsorianod:
Move Cluster Service node pool PATCH out of the frontend. ARM updates now only persist desired state in Cosmos and create an update operation. Backend controllers reconcile the rest.

Add a node pool updatable config hash (labels, scaling, taints, nodeDrainTimeoutMinutes) and store it on ServiceProviderNodePoolStatus.

A new NodePoolClusterServiceUpdateDispatch controller compares the hash against Cosmos desired state and calls UpdateNodePool when it changes, retrying transient CS 400 rejections when the cluster or node pool is not yet updatable.

Extend the node pool update operation controller to follow the cluster create pattern: poll Cluster Service status, gate on Cosmos/MC readiness (ReadDesire NodePool mirror, spec/status match, rollout conditions), and only mark the operation Succeeded when all layers agree.

Also add helpers for decoding the per-node-pool NodePool ReadDesire

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -32,6 +32,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.43.0
 	go.uber.org/mock v0.6.0
 	golang.org/x/sync v0.20.0
+	k8s.io/api v0.35.3
 	k8s.io/apimachinery v0.35.3
 	k8s.io/client-go v0.35.3
 	k8s.io/component-base v0.35.3
@@ -142,7 +143,6 @@ require (
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	k8s.io/api v0.35.3 // indirect
 	k8s.io/kube-openapi v0.0.0-20260317180543-43fb72c5454a // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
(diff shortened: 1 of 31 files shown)


miguelsorianod:
I think the rollout of this could be:
1. Merge frontend changes that introduce the `UsesNewNodePoolUpdateApproach` boolean in the Cosmos internal NodePool type. Set to `false` for all NodePools. Also the change that introduces the `ClusterServiceUpdatableConfigHashForUpdateDispatch` string in ServiceProviderNodePool status. TODO do we want it in status or in spec of ServiceProviderNodePool?
1. Merge backend code that calculates the updatable config sha in the nodepool create dispatch controller
1. Merge backend update dispatch controller changes gated on UsesNewNodePoolUpdateApproach. Even with `UsesNewNodePoolUpdateApproach` set to false this controller will act and calculate the updatable config sha for pre-existing NodePools. This ensures all nodepools not being deleted contain the calculated sha
1. Merge backend code that changes the nodepool update operation to rely on a combination of CS state + Hypershift NodePool state, picking the worst. This code should not need to depend on the UsesNewNodePoolUpdateApproach boolean
1. Merge frontend code that sets the UsesNewNodePoolUpdateApproach to true in the internal NodePool cosmos objectwhen a nodepool update is requested. Here the update operation document does not contain this bool as it's not needed. This makes new nodepool updates via ARM to start exercising the backend dispatch controller
1. Cleanup of boolean gating

miguelsorianod:
At the end we implemented the approach by getting state from both cosmos desired and CS desired on each reconcile and compare them.

Replaced by https://github.com/Azure/ARO-HCP/pull/5876.

Closing

PR #5570, opened 2026-06-09
feat: enable cluster delete from frontend to backend

miguelsorianod:
Move Cluster Service interaction out of the frontend delete
path and hand it off to the backend's async cluster deletion controllers.

The frontend now only:

- Creates a delete operation with UsesNewClusterDeletionApproach set to true
- Sets DeletionTimestamp and sets UsesNewClusterDeletionApproach to true in the Cluster

Removed from the frontend:

- Direct clusterServiceClient.DeleteCluster() call

Cluster Delete operations no longer set an InternalID as it is not needed
anymore when using the cluster deletion approach in backend.

Legacy deletion (operations without UsesNewClusterDeletionApproach) is unchanged and still handled by OperationClusterDeleteController.

### DIFF
--- a/frontend/pkg/frontend/cluster.go
+++ b/frontend/pkg/frontend/cluster.go
@@ -17,7 +17,6 @@ package frontend
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"maps"
 	"net/http"
@@ -31,8 +30,6 @@ import (
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
-
-	ocmerrors "github.com/openshift-online/ocm-sdk-go/errors"
 
 	"github.com/Azure/ARO-HCP/internal/admission"
 	"github.com/Azure/ARO-HCP/internal/api"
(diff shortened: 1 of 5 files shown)


backend/pkg/controllers/operationcontrollers/operation_cluster_delete.go
```diff
@@ -107,85 +117,73 @@ func (c *operationClusterDelete) SynchronizeOperation(ctx context.Context, key c
...
+	clusterCRUD := c.resourcesDBClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName)
+	cluster, err := clusterCRUD.Get(ctx, operation.ExternalID.Name)
+	if database.IsNotFoundError(err) {
+		logger.Info("cluster document deleted - completing operation")
+		err = SetDeleteOperationAsCompleted(ctx, c.clock, c.resourcesDBClient, operation, postAsyncNotificationFn(c.notificationClient))
+		if err != nil {
+			return utils.TrackError(err)
+		}
 		return nil
 	}
```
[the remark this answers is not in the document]
miguelsorianod: The feedback is not correct. The billing doc for the new deletion approach is now marked in the deletion cluster controller

backend/pkg/controllers/operationcontrollers/operation_cluster_delete.go
```diff
@@ -107,85 +117,73 @@ func (c *operationClusterDelete) SynchronizeOperation(ctx context.Context, key c
...
-		}
-		err = externalAuthIterator.GetError()
-		if err != nil {
-			return utils.TrackError(err)
-		}
-		if foundAtLeastOneExternalAuth {
-			logger.Info("cluster still has cosmos externalauths")
-			return nil
-		}
+func (c *operationClusterDelete) shouldReconcileOperationAndResourceStatus(cluster *api.HCPOpenShiftCluster) bool {
```
swiencki, line 157: The nil-ClusterServiceID path where `shouldReconcileOperationAndResourceStatus` returns false and the operation stays `Accepted` indefinitely could use an end-to-end test to confirm that's the intended behavior.

miguelsorianod: That is already covered by a unit test of the SynchronizeOperation. No need to have an e2e test for that.

miguelsorianod:

PR #5572, opened 2026-06-09
ARO-27344: Use GetLatestInstallVersion for nightly E2E version selection

adalrsjr1:
On the nightly channel, two test groups failed because `GetLatestVersionInMinor` queries Cincinnati, which returns bare GA versions (e.g. `4.22.0`) that don't exist in CS as nightly builds. Use `GetLatestInstallVersion` instead, which dispatches to the CI release stream API for nightly and returns proper tags that CS recognizes.

Also fix `control_plane_minor_upgrade.go` which truncated the discovered version to Major.Minor, causing the RP's hardcoded patch map to produce `openshift-v4.20.22-nightly` — a version CS doesn't have.

PR #5409 stopped the tests from using RC/EC anchors on nightly (partial fix), but the GA anchor 4.22.0 still leaked through Cincinnati into the version selection — and 4.22.0 doesn't exist in CS as a nightly build either.

[ARO-27344](https://issues.redhat.com/browse/ARO-27344)

### What

- Nodepool +2 skip test and y-stream downgrade test: replace `GetLatestVersionInMinor` with `GetLatestInstallVersion` at 4 call sites. `GetLatestInstallVersion` dispatches to the CI release stream API for nightly, returning proper tags like `4.22.0-0.nightly-multi-2026-06-09-013153`.
- Control plane minor upgrade test: pass the full nightly version tag instead of truncating to Major.Minor. The RP's `NewOpenShiftVersionXYZ` cannot resolve Major.Minor to a specific nightly build.
- New `IsVersionNotFoundError` in `version_helper.go`: covers both Cincinnati and nightly sentinel errors, used at the call sites that switched to `GetLatestInstallVersion`.

### Why

Cincinnati's nightly graph contains GA nodes (e.g. `4.22.0`) with zero outgoing edges. `GetLatestVersionInMinor` returned `4.22.0` as the latest version. The RP then constructed `openshift-v4.22.0-nightly` — a version ID that doesn't exist in CS, because CS discovers nightly versions from the release stream API (full timestamp tags). This caused `InvalidRequestContent: Version 'openshift-v4.22.0-nightly' doesn't exist` on nightly E2E runs.

For the control plane minor upgrade test, the same issue occurred via a different path: the test truncated the version to Major.Minor (`4.20`), and the RP's hardcoded patch map resolved it to `openshift-v4.20.22-nightly`.

### Testing

- nodepool upgrade/downgrade tests need to pass on `stage-e2e-parallel-ocp-nightly`
- Live probes confirmed the release stream API returns proper nightly tags that match CS version IDs
- Other tests (z-stream upgrades, y-stream upgrades) are unchanged

- Only test infrastructure code is changed (`test/util/framework/`, `test/e2e/`), no production code
- The fix is targeted: only the 2 failing test groups + the CP minor upgrade test are modified
- Tests that were previously skipping on nightly continue to skip — no behavior change for those
- `GetLatestInstallVersion` already existed and correctly dispatches to the release stream API for nightly
- Non-nightly channels are completely unchanged

### PR Checklist
- [x] Self-reviewed the diff

### DIFF
--- a/test/e2e/control_plane_minor_upgrade.go
+++ b/test/e2e/control_plane_minor_upgrade.go
@@ -111,6 +111,18 @@ var _ = Describe("Customer", func() {
 					previousMinor.String(), targetVer.String()))
 			}
 
+			// Resolve the upgrade target version early so a missing nightly tag skips
+			// before burning resources on cluster creation.
+			upgradeVersionId := targetMinor
+			if channelGroup == "nightly" {
+				resolved, err := framework.GetLatestInstallVersion(ctx, channelGroup, targetMinor)
+				if framework.IsVersionNotFoundError(err) {
+					Skip(fmt.Sprintf("no nightly version for %s: %v", targetMinor, err))
+				}
+				Expect(err).NotTo(HaveOccurred(), "failed to resolve nightly upgrade version")
+				upgradeVersionId = resolved
+			}
+
 			tc := framework.NewTestContext()
 			if tc.UsePooledIdentities() {
 				err := tc.AssignIdentityContainers(ctx, 1, framework.IdentityContainerAssignmentRetryInterval)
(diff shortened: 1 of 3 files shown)


test/e2e/control_plane_minor_upgrade.go
```diff
@@ -178,18 +189,29 @@ var _ = Describe("Customer", func() {
...
-			By(fmt.Sprintf("triggering control plane y-stream upgrade to %s (target minor %s)", targetMinor,
+			// Nightly needs the full version tag; the RP cannot resolve Major.Minor to a nightly build.
+			upgradeVersionId := targetMinor
+			if channelGroup == "nightly" {
+				resolved, err := framework.GetLatestInstallVersion(ctx, channelGroup, targetMinor)
+				if framework.IsVersionNotFoundError(err) {
+					Skip(fmt.Sprintf("no nightly version for %s: %v", targetMinor, err))
+				}
+				Expect(err).NotTo(HaveOccurred(), "failed to resolve nightly upgrade version")
+				upgradeVersionId = resolved
```
swiencki: The nightly upgrade-version resolve should move before cluster creation so a missing target-minor tag skips cheaply instead of burning ~30 min of stage capacity.

miquelsi (approved):
/approve

machi1990:
/lgtm

This is good to fix the issue, I believe we need a follow to unify the version helpers for our e2e to avoid future problems like these. A single entry method that caters for the different channels needs and the e2e just relies on those single entry methods.

@miquelsi @adalrsjr1 thoughts on that followup?

PR #5584, opened 2026-06-10
feat: expose cluster control plane version state in backend Prometheus metrics

machi1990:
### What

feat: expose cluster control plane version state in backend Prometheus metrics

### Why

Add backend_cluster_version_info so fleet and upgrade dashboards (to be added as followup) can track each cluster's target and active OpenShift versions from ServiceProviderCluster data.

The cluster metrics handler now reads desired and active control plane versions on sync, emits one time series per version/state pair, and cleans up series when a cluster is deleted. Active versions take precedence over desired for the same version string.

Example metric

1. 4.19.20 is desired
```
backend_cluster_version_info{
  resource_id="/subscriptions/aabbccdd-1111-2222-3333-444455556666/resourcegroups/my-rg/providers/microsoft.redhatopenshift/hcpopenshiftclusters/my-cluster",
  subscription_id="aabbccdd-1111-2222-3333-444455556666",
  version="4.19.20",
  state="desired"
} 1
```
2. it becomes partial: active version in the cluster but partially upgraded to

```
backend_cluster_version_info{
  resource_id="/subscriptions/aabbccdd-1111-2222-3333-444455556666/resourcegroups/my-rg/providers/microsoft.redhatopenshift/hcpopenshiftclusters/my-cluster",
  subscription_id="aabbccdd-1111-2222-3333-444455556666",
  version="4.19.20",
  state="partial"
} 1
```

3. once completed: the upgrade is complete

```
backend_cluster_version_info{
  resource_id="/subscriptions/aabbccdd-1111-2222-3333-444455556666/resourcegroups/my-rg/providers/microsoft.redhatopenshift/hcpopenshiftclusters/my-cluster",
  subscription_id="aabbccdd-1111-2222-3333-444455556666",
  version="4.19.20",
  state="`completed"
} 1
```

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/backend/pkg/app/backend.go
+++ b/backend/pkg/app/backend.go
@@ -400,7 +400,7 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context, 
 		"ClusterMetrics", clusterInformer, clusterHandler)
 
 	serviceProviderClusterInformer, _ := backendInformers.ServiceProviderClusters()
-	clusterVersionMetricsHandler := metricscontrollers.NewClusterVersionMetricsHandler(b.options.MetricsRegisterer)
+	clusterVersionMetricsHandler := metricscontrollers.NewClusterVersionMetricsHandler(b.options.MetricsRegisterer, unionReadDesireLister)
 	clusterVersionMetricsController := metricscontrollers.NewController(
 		"ClusterVersionMetrics", serviceProviderClusterInformer, clusterVersionMetricsHandler)
 
--- a/backend/pkg/app/backend.go
+++ b/backend/pkg/app/backend.go
@@ -395,9 +395,14 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context, 
 	_, unionReadDesireLister := unionKubeApplierInformers.ReadDesires()
 
 	clusterInformer, clusterLister := backendInformers.Clusters()
-	clusterHandler := metricscontrollers.NewClusterMetricsHandler(b.options.MetricsRegisterer, b.options.ResourcesDBClient)
+	clusterHandler := metricscontrollers.NewClusterMetricsHandler(b.options.MetricsRegisterer)
 	clusterMetricsController := metricscontrollers.NewController(
 		"ClusterMetrics", clusterInformer, clusterHandler)
+
+	serviceProviderClusterInformer, _ := backendInformers.ServiceProviderClusters()
+	clusterVersionMetricsHandler := metricscontrollers.NewClusterVersionMetricsHandler(b.options.MetricsRegisterer)
+	clusterVersionMetricsController := metricscontrollers.NewController(
+		"ClusterVersionMetrics", serviceProviderClusterInformer, clusterVersionMetricsHandler)
 
 	_, billingLister := backendInformers.BillingDocs()
 
@@ -801,6 +806,7 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context, 
 				go clusterDeletionController.Run(ctx, 20)
 				go operationPhaseMetricsController.Run(ctx, 1)
 				go clusterMetricsController.Run(ctx, 1)
+				go clusterVersionMetricsController.Run(ctx, 1)
 				go nodePoolMetricsController.Run(ctx, 1)
 				go externalAuthMetricsController.Run(ctx, 1)
 				go placementSyncController.Run(ctx, 20)
--- a/backend/pkg/app/backend.go
+++ b/backend/pkg/app/backend.go
@@ -395,7 +395,7 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context, 
 	_, unionReadDesireLister := unionKubeApplierInformers.ReadDesires()
 
 	clusterInformer, clusterLister := backendInformers.Clusters()
-	clusterHandler := metricscontrollers.NewClusterMetricsHandler(b.options.MetricsRegisterer)
+	clusterHandler := metricscontrollers.NewClusterMetricsHandler(b.options.MetricsRegisterer, b.options.ResourcesDBClient)
 	clusterMetricsController := metricscontrollers.NewController(
 		"ClusterMetrics", clusterInformer, clusterHandler)
 
--- a/backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
+++ b/backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
@@ -109,17 +109,21 @@ func (metricsHandler *clusterVersionMetricsHandler) clusterUUIDMetricLabel(
 func (metricsHandler *clusterVersionMetricsHandler) versionStatesFromServiceProviderCluster(serviceProviderCluster *api.ServiceProviderCluster) map[string]string {
 	versionStates := make(map[string]string)
 
-	// Desired is emitted only when the target z-stream is not yet active. Active versions
-	// (partial or completed) override desired for the same version string.
-	if serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion != nil {
-		versionStates[serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion.String()] = "desired"
-	}
-
 	for _, activeVersion := range serviceProviderCluster.Status.ControlPlaneVersion.ActiveVersions {
-		if activeVersion.Version == nil {
+		version := activeVersion.Version.String()
+		// ActiveVersions is newest-first; skip duplicates that can appear after a rollback to the same z-stream.
+		if _, ok := versionStates[version]; ok {
 			continue
 		}
-		versionStates[activeVersion.Version.String()] = strings.ToLower(string(activeVersion.State))
+		versionStates[version] = strings.ToLower(string(activeVersion.State))
+	}
+
+	// Desired is emitted only when the target z-stream is not yet active.
+	if serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion != nil {
+		desiredVersion := serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion.String()
+		if _, ok := versionStates[desiredVersion]; !ok {
+			versionStates[desiredVersion] = "desired"
+		}
 	}
 
 	return versionStates
--- a/backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
+++ b/backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
@@ -42,7 +42,7 @@ func NewClusterVersionMetricsHandler(
 		readDesireLister: readDesireLister,
 		clusterVersionInfo: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: "backend_cluster_version_info",
-			Help: "Information about cluster versions. Value is 1 when version is in the specified state. States: desired (target selected, upgrade not started), partial (upgrade in progress), completed (upgrade finished). Use partial and completed for fleet and upgrade-progress metrics; desired is pre-upgrade only.",
+			Help: "Information about cluster versions. Value is 1 when version is in the specified state. States: desired (target selected, but version hasn't reflected in the cluster active versions), partial (upgrade in progress), completed (upgrade finished). Use partial and completed for fleet and upgrade-progress metrics.",
 		}, []string{"resource_id", "subscription_id", "cluster_uuid", "version", "state"}),
 	}
 	prometheusRegisterer.MustRegister(metricsHandler.clusterVersionInfo)
@@ -94,11 +94,11 @@ func (metricsHandler *clusterVersionMetricsHandler) clusterUUIDMetricLabel(
 	)
 	if err != nil {
 		logger := utils.LoggerFromContext(ctx)
-		logger.Info("error getting cluster UUID, continuing with empty", "err", err.Error())
+		logger.Error(utils.TrackError(err), "error getting cluster UUID, continuing with empty")
 	}
 	if !found {
 		logger := utils.LoggerFromContext(ctx)
-		logger.Info("missing cluster UUID, continuing with empty")
+		logger.V(1).Info("missing cluster UUID, continuing with empty")
 	}
 	if err != nil || !found {
 		return ""
--- a/backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
+++ b/backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
@@ -1,0 +1,118 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metricscontrollers
+
+import (
+	"context"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+type clusterVersionMetricsHandler struct {
+	clusterVersionInfo *prometheus.GaugeVec
+	resourcesDBClient  database.ResourcesDBClient
+}
+
+func newClusterVersionMetricsHandler(
+	prometheusRegisterer prometheus.Registerer,
+	resourcesDBClient database.ResourcesDBClient,
+) *clusterVersionMetricsHandler {
+	metricsHandler := &clusterVersionMetricsHandler{
+		clusterVersionInfo: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "backend_cluster_version_info",
+			Help: "Information about cluster versions. Value is 1 when version is in the specified state. States: desired (target selected, upgrade not started), partial (upgrade in progress), completed (upgrade finished). Use partial and completed for fleet and upgrade-progress metrics; desired is pre-upgrade only.",
+		}, []string{"resource_id", "subscription_id", "version", "state"}),
+		resourcesDBClient: resourcesDBClient,
+	}
+	prometheusRegisterer.MustRegister(metricsHandler.clusterVersionInfo)
+	return metricsHandler
+}
+
+func (metricsHandler *clusterVersionMetricsHandler) Sync(ctx context.Context, clusterVersionObject clusterMetricsObject) {
+	clusterResourceID := clusterVersionObject.ResourceID()
+	resourceID := resourceIDMetricLabel(clusterResourceID)
+	if len(resourceID) == 0 {
+		return
+	}
+	subscriptionID := subscriptionIDMetricLabel(clusterResourceID)
+
+	serviceProviderCluster, err := metricsHandler.resourcesDBClient.ServiceProviderClusters(
+		clusterResourceID.SubscriptionID,
+		clusterResourceID.ResourceGroupName,
+		clusterResourceID.Name,
+	).Get(ctx, api.ServiceProviderClusterResourceName)
+	if err != nil {
+		if database.IsNotFoundError(err) {
+			metricsHandler.clusterVersionInfo.DeletePartialMatch(prometheus.Labels{"resource_id": resourceID})
+		} else {
+			utils.LoggerFromContext(ctx).Error(err, "failed to get ServiceProviderCluster for cluster version metrics; retaining previous metrics",
+				"resource_id", resourceID)
+		}
+		return
+	}
+
+	metricsHandler.clusterVersionInfo.DeletePartialMatch(prometheus.Labels{"resource_id": resourceID})
+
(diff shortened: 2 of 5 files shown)


backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	for version, state := range versionStatesFromServiceProviderCluster(serviceProviderCluster) {
+		metricsHandler.clusterVersionInfo.With(prometheus.Labels{
+			"resource_id":     resourceID,
+			"subscription_id": subscriptionID,
+			"version":         version,
+			"state":           state,
+		}).Set(1.0)
+	}
```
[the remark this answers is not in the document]
machi1990: While this is true, it's okay to start with this approach and we'll refine over time. For now, I followed the approach done in https://github.com/machi1990/ARO-HCP/blob/14c1ecd9e4f79eb385f2fcbc04e159448114b5ae/backend/pkg/controllers/metricscontrollers/resource_metrics_controller.go#L71 
If the churn becomes too expensive we can improve

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,113 @@
...
+	clusterResourceID := clusterVersionObject.ResourceID()
+	resourceID := resourceIDMetricLabel(clusterResourceID)
+	if len(resourceID) == 0 {
+		return
+	}
+	subscriptionID := subscriptionIDMetricLabel(clusterResourceID)
+
+	deleteSelector := prometheus.Labels{"resource_id": resourceID}
+	metricsHandler.clusterVersionInfo.DeletePartialMatch(deleteSelector)
+
```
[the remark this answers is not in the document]
machi1990: Same comment as https://github.com/Azure/ARO-HCP/pull/5584#discussion_r3387827911

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,113 @@
...
+	serviceProviderCluster, err := metricsHandler.resourcesDBClient.ServiceProviderClusters(
+		clusterResourceID.SubscriptionID,
+		clusterResourceID.ResourceGroupName,
+		clusterResourceID.Name,
+	).Get(ctx, api.ServiceProviderClusterResourceName)
+	if err != nil {
+		// ServiceProviderCluster may not exist yet, that's okay
+		return
+	}
```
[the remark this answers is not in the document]
machi1990: addressed

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,118 @@
...
+			Name: "backend_cluster_version_info",
+			Help: "Information about cluster versions. Value is 1 when version is in the specified state. States: desired (target selected, upgrade not started), partial (upgrade in progress), completed (upgrade finished). Use partial and completed for fleet and upgrade-progress metrics; desired is pre-upgrade only.",
+		}, []string{"resource_id", "subscription_id", "version", "state"}),
+		resourcesDBClient: resourcesDBClient,
+	}
+	prometheusRegisterer.MustRegister(metricsHandler.clusterVersionInfo)
+	return metricsHandler
+}
+
+func (metricsHandler *clusterVersionMetricsHandler) Sync(ctx context.Context, clusterVersionObject clusterMetricsObject) {
```
swiencki: This gauge reads SPC data but is driven by the cluster informer (1h resync), so upgrade transitions won't surface until the next cluster doc write or hourly relist; consider driving it from the SPC informer (30s relist) instead.

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,118 @@
...
+	clusterResourceID := clusterVersionObject.ResourceID()
+	resourceID := resourceIDMetricLabel(clusterResourceID)
+	if len(resourceID) == 0 {
+		return
+	}
+	subscriptionID := subscriptionIDMetricLabel(clusterResourceID)
+
+	serviceProviderCluster, err := metricsHandler.resourcesDBClient.ServiceProviderClusters(
+		clusterResourceID.SubscriptionID,
+		clusterResourceID.ResourceGroupName,
```
swiencki: An in-memory SPC lister with 30s relist already exists, so this Cosmos point-read per cluster event is unnecessary and will amplify on update bursts.

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
machi1990: Isn't the relist 30s for all cluster related objs?

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
machi1990: Good catch

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
machi1990: Regardless of my question above, you made me realise that perhaps I want a separate controller for this; I am exploring that

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
machi1990: fixed

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
machi1990: fixed, thanks for pointing this out

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (metricsHandler *clusterVersionMetricsHandler) Delete(serviceProviderClusterKey string) {
+	if len(serviceProviderClusterKey) == 0 {
+		return
+	}
+	serviceProviderClusterResourceID, err := azcorearm.ParseResourceID(serviceProviderClusterKey)
+	if err != nil {
+		return
+	}
```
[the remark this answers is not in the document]
machi1990: This isn't going to occur

backend/pkg/app/backend.go
```diff
@@ -399,6 +399,11 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
 	clusterMetricsController := metricscontrollers.NewController(
 		"ClusterMetrics", clusterInformer, clusterHandler)
 
+	serviceProviderClusterInformer, _ := backendInformers.ServiceProviderClusters()
```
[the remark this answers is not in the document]
machi1990:
```
ServiceProviderClusters() (cache.SharedIndexInformer, listers.ServiceProviderClusterLister)
```

it doesn't return an error, I don't think this comment is valid, unless otherwise I misunderstood it.

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,98 @@
...
+		if activeVersion.Version == nil {
+			continue
+		}
+
+		state := "completed"
+		if activeVersion.State == configv1.PartialUpdate {
+			state = "partial"
+		}
+
+		versionStates[activeVersion.Version.String()] = state
```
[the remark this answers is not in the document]
machi1990: There are no any other states;

Only `partial` and `completed`.
But I can likely do better than this. I'll lowercase it and avoid the need for constant.

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler_test.go
```diff
@@ -0,0 +1,236 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metricscontrollers
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
```
swiencki, line 20: Consider adding a Sync-twice test (like `TestClusterMetricsHandler_PhaseTransitionDeletesOldSeries` in the sibling handler) to verify old version/state tuples are cleaned up when the cluster's version state changes.

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,126 @@
...
+	})
+}
+
+func (metricsHandler *clusterVersionMetricsHandler) clusterUUIDMetricLabel(
+	ctx context.Context,
+	clusterResourceID *azcorearm.ResourceID,
+) string {
+	clusterUUID, found, err := maestrohelpers.GetCachedHostedClusterUUIDForCluster(
+		ctx,
+		metricsHandler.readDesireLister,
```
swiencki, line 90: If `activeVersion.State` is ever empty, this emits `state=""` which the help text doesn't document; could skip entries with empty state to match the documented vocabulary.

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+type clusterVersionMetricsHandler struct {
+	clusterVersionInfo *prometheus.GaugeVec
+	readDesireLister   dblisters.ReadDesireLister
+}
+
+// NewClusterVersionMetricsHandler creates a metrics handler for cluster version metrics.
+func NewClusterVersionMetricsHandler(
+	prometheusRegisterer prometheus.Registerer,
```
swiencki, line 38: "desired is pre-upgrade only" is slightly overstated; during a multi-step upgrade `desired` coexists with `partial`/`completed` for different versions. Maybe "desired (target version not yet observed in control plane history)" is more precise.

swiencki (commented):
A few suggestions from reviewing the metric design and test coverage.

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
machi1990: it'll never be empty

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
machi1990: A given version will be marked as desired only if the upgrade hasn't reflected back in active versions.
This is what is pre-upgrade only for that version.

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,126 @@
...
+	prometheusRegisterer.MustRegister(metricsHandler.clusterVersionInfo)
+	return metricsHandler
+}
+
+func (metricsHandler *clusterVersionMetricsHandler) Sync(ctx context.Context, serviceProviderCluster *api.ServiceProviderCluster) {
+	resourceID := resourceIDMetricLabel(serviceProviderCluster.ResourceID.Parent)
+	subscriptionID := subscriptionIDMetricLabel(serviceProviderCluster.ResourceID.Parent)
+	clusterUUID := metricsHandler.clusterUUIDMetricLabel(ctx, serviceProviderCluster.ResourceID.Parent)
+
+	metricsHandler.clusterVersionInfo.DeletePartialMatch(prometheus.Labels{"resource_id": resourceID})
```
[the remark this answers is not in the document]
machi1990: This is fine for now; as we scale, we can refine the solution

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,126 @@
...
+	if serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion != nil {
+		versionStates[serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion.String()] = "desired"
+	}
+
+	for _, activeVersion := range serviceProviderCluster.Status.ControlPlaneVersion.ActiveVersions {
+		if activeVersion.Version == nil {
+			continue
+		}
+		versionStates[activeVersion.Version.String()] = strings.ToLower(string(activeVersion.State))
+	}
```
JameelB: How will this deal with downgrades? If it's represented like so: `activeVersions: [x.y.zOld, x.y.z+1, x.y.zOld]` , this may return an incorrect state. 

Since the latest version is the downgrade, `x.y.zOld` will be set in the versionStates with a 'state=partial' originally. But then, when it gets to the original version, it overrides it to completed, so the versionStates returned will be as follows:`[version=x.y.zOld, state=completed], [version=x.y.z+1, state=completed]`

After syncing with @machi1990 offline, since the order matters, we could take most recent state of that version only.

JameelB: Since rollback isn't supported for control plane yet, this will be handled as part of the feature implementation. Resolving thread as won't do for now.

machi1990: Thanks. For now, I am skipping duplicates version.
We don't have downgrades now for control plane so this is likely not occur but skipping makes sense.
For when we've downgrades, I see the active version controller being modified to cater for this. 
That can be dealt with when we've rollback feature

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// NewClusterVersionMetricsHandler creates a metrics handler for cluster version metrics.
+func NewClusterVersionMetricsHandler(
+	prometheusRegisterer prometheus.Registerer,
+	readDesireLister dblisters.ReadDesireLister,
+) Handler[*api.ServiceProviderCluster] {
+	metricsHandler := &clusterVersionMetricsHandler{
+		readDesireLister: readDesireLister,
+		clusterVersionInfo: prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
swiencki, line 43: `Sync` should early-return on empty `resourceID` and skip active versions with nil `Version` or empty `State` to avoid a panic and emitting undocumented label values.

machi1990: Those cases;
- resourceID being empty, 
- version being nil
- state being empty 

Won't ever occur in the system. I don't feel the need to add defensive code for this.
I am ignoring this comment unless otherwise you work me a real scenario where this can occur.

geoberle:
/approved
/hold
lets limit the number of concurrent versions we report on, just we have an upper bound of timeseries we expose

machi1990:
> lets limit the number of concurrent versions we report on, just we have an upper bound of timeseries we expose

Thanks @geoberle I capped it to 3. 

I'll unhold and ask @JameelB for another review

/unhold

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,137 @@
...
+		activeVersions = activeVersions[:maxReportedActiveVersions]
+	}
+	for _, activeVersion := range activeVersions {
+		version := activeVersion.Version.String()
+		// ActiveVersions is newest-first; skip duplicates that can appear after a rollback to the same z-stream.
+		if _, ok := versionStates[version]; ok {
+			continue
+		}
+		versionStates[version] = strings.ToLower(string(activeVersion.State))
+	}
```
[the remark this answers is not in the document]
machi1990: It'll never be nil

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for version, state := range metricsHandler.versionStatesFromServiceProviderCluster(serviceProviderCluster) {
+		metricsHandler.clusterVersionInfo.With(prometheus.Labels{
+			"resource_id":     resourceID,
+			"subscription_id": subscriptionID,
+			"cluster_uuid":    clusterUUID,
+			"version":         version,
+			"state":           state,
+		}).Set(1.0)
+	}
```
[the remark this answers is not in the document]
machi1990: It doesn't matter

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler_test.go
```diff
@@ -0,0 +1,324 @@
...
+func TestClusterVersionMetricsHandler_SetsDesiredAndActiveVersions(t *testing.T) {
+	ctx := context.Background()
+	prometheusRegistry := prometheus.NewRegistry()
+	handler := newTestClusterVersionMetricsHandler(t, prometheusRegistry, newTestHostedClusterReadDesireLister(t, "11111111-1111-1111-1111-111111111111"))
+
+	cluster := newTestCluster(t, "cluster-1", arm.ProvisioningStateSucceeded, nil)
+	serviceProviderCluster := newTestServiceProviderCluster(t, cluster, "4.19.20", []api.HCPClusterActiveVersion{
+		{Version: ptr.To(semver.MustParse("4.19.19")), State: configv1.CompletedUpdate},
+		{Version: ptr.To(semver.MustParse("4.19.20")), State: configv1.PartialUpdate},
+	})
```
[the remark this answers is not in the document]
machi1990: version will never be nil

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,135 @@
...
+		versionStates[version] = strings.ToLower(string(activeVersion.State))
+	}
+
+	// Desired is emitted only when the target z-stream is not yet active.
+	if serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion != nil {
+		desiredVersion := serviceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion.String()
+		if _, ok := versionStates[desiredVersion]; !ok {
+			versionStates[desiredVersion] = "desired"
+		}
+	}
```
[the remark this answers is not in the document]
machi1990: The cap is on 3 active versions

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (metricsHandler *clusterVersionMetricsHandler) versionStatesFromServiceProviderCluster(serviceProviderCluster *api.ServiceProviderCluster) map[string]string {
+	versionStates := make(map[string]string)
+
+	activeVersions := serviceProviderCluster.Status.ControlPlaneVersion.ActiveVersions
+	// Bound concurrent version time series per cluster: the version we are on or moving to and the two previous versions.
+	const maxReportedActiveVersions = 3
+	if len(activeVersions) > maxReportedActiveVersions {
+		activeVersions = activeVersions[:maxReportedActiveVersions]
+	}
```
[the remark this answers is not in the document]
machi1990: the cap is on 3 active version

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,135 @@
...
+	for _, activeVersion := range activeVersions {
+		version := activeVersion.Version.String()
+		// ActiveVersions is newest-first; skip duplicates that can appear after a rollback to the same z-stream.
+		if _, ok := versionStates[version]; ok {
+			continue
+		}
+		versionStates[version] = strings.ToLower(string(activeVersion.State))
+	}
```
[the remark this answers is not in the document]
machi1990: Noo; version will never be nil

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (metricsHandler *clusterVersionMetricsHandler) versionStatesFromServiceProviderCluster(serviceProviderCluster *api.ServiceProviderCluster) map[string]string {
+	versionStates := make(map[string]string)
+
+	activeVersions := serviceProviderCluster.Status.ControlPlaneVersion.ActiveVersions
+	// Bound concurrent version time series per cluster: the version we are on or moving to and the two previous versions.
+	const maxReportedActiveVersions = 3
+	if len(activeVersions) > maxReportedActiveVersions {
+		activeVersions = activeVersions[:maxReportedActiveVersions]
```
swiencki, line 115: The cap trims active versions to 3 but desired is appended after, so 3 active plus a distinct desired emits 4 series (not the 3 the comment claims); fold desired into the same budget and add a test for that case.

backend/pkg/controllers/metricscontrollers/cluster_version_metrics_handler.go
```diff
@@ -0,0 +1,135 @@
...
+			"resource_id":     resourceID,
+			"subscription_id": subscriptionID,
+			"cluster_uuid":    clusterUUID,
+			"version":         version,
+			"state":           state,
+		}).Set(1.0)
+	}
+}
+
+func (metricsHandler *clusterVersionMetricsHandler) Delete(serviceProviderClusterKey string) {
```
swiencki, line 70: This metric keys off the ServiceProviderCluster, which the cluster delete path never removes (the frontend deletes only NodePools and ExternalAuth, and there is no DeleteServiceProviderCluster), so because the controller is level-driven these series keep re-emitting for deleted clusters, unlike backend_cluster_provision_state which keys off the cluster informer and clears; we should reconcile against cluster existence or clean up on cluster deletion.

PR #5588, opened 2026-06-10
Alerting queues cleanups

mmazur:
https://redhat.atlassian.net/browse/AROSLSRE-843

## Summary

- Lots of bigger and smaller cleanups
- Emptying out the SRE alerting queue

### DIFF
--- a/backend/alerts/backend-prometheusRule.yaml
+++ b/backend/alerts/backend-prometheusRule.yaml
@@ -19,11 +19,11 @@ spec:
             workqueue_depth{namespace="aro-hcp"}
           )
         ) > 10
-      for: 5m
+      for: 15m
       labels:
         severity: warning
       annotations:
-        description: 'Backend controller workqueue {{ $labels.name }} has had a depth > 10 for more than 5 minutes, indicating work is accumulating faster than it can be processed.'
+        description: 'Backend controller workqueue {{ $labels.name }} has had a depth > 10 for more than 15 minutes, indicating work is accumulating faster than it can be processed.'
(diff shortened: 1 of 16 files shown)


backend/alerts/backend-prometheusRule.yaml
```diff
@@ -12,40 +12,14 @@ spec:
...
-        summary: 'Backend controller workqueue retry hot loop'
     - alert: BackendControllerQueueDepthHigh
       expr: |
         max by (name, cluster) (
           max without(prometheus_replica) (
             workqueue_depth{namespace="aro-hcp"}
           )
         ) > 10
-      for: 5m
+      for: 15m
```
swiencki, line 22: The description still says "more than 5 minutes" but `for:` is now 15m, so responders will see stale triage context.

backend/alerts/backend-retryhotloop-prometheusRule.yaml
```diff
@@ -0,0 +1,41 @@
...
+            max without(prometheus_replica) (
+              rate(workqueue_adds_total{namespace="aro-hcp"}[10m])
+            )
+          )
+        ) > 0.5
+      for: 10m
+      labels:
+        severity: warning
+      annotations:
+        correlationId: 'BackendControllerRetryHotLoop/{{ $labels.cluster }}/{{ $labels.name }}'
```
swiencki, line 38: `docs/alerts.md` line 102 uses this exact alert as the canonical example of default per-cluster coalescing, so the override here makes that doc example wrong and should be updated in the same PR.

PR #5589, opened 2026-06-10
Avoid unnecessary error/requeue on conflict.

deads2k:
When cluster properties sync gets a conflict, it doesn't need to requeue since the informer will eventually observe the update in state and trigger the controller.

Fixes the excessive retries in CI that look like

> 	"msg": "Error syncing; requeuing for later retry",
> 	"err": "(wrapped at cluster_properties_sync.go:133) failed to replace Cluster: PUT https://arohcpci01-rp-j0297856-westus3.documents.azure.com:443/dbs/arohcpci01-rp-j0297856/colls/Resources/docs/686daff4-1742-55d9-8c6e-926f6a728fbd\n--------------------------------------------------------------------------------\nRESPONSE 412: 412 Precondition Failed\nERROR CODE: 412 Precondition Failed\n--------------------------------------------------------------------------------\n{\n  \"code\": \"PreconditionFailed\",\n  \"message\": \"Operation cannot be performed because one of the specified precondition is not met., \\r\\nRequestStartTime: 2026-06-10T13:04:51.1507101Z, RequestEndTime: 2026-06-10T13:04:51.1519557Z,  Number of regions attempted:1\\r\\n{\\\"systemHistory\\\":

ref https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/5584/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2064672805890297856

### DIFF
--- a/backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
+++ b/backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
@@ -129,7 +129,11 @@ func (c *clusterPropertiesSyncer) SyncOnce(ctx context.Context, key controllerut
 		return nil
 	}
 
-	if _, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Replace(ctx, replacement, nil); err != nil {
+	if _, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Replace(ctx, replacement, nil); database.IsPreconditionFailedError(err) {
+		// if we have a conflict error, then we're guaranteed that our informer will eventually see an update and trigger us again.
+		// there is no need to report an error since the retry will happen when the reflector sees the update and puts an Update into the informer.
+		return nil
+	} else if err != nil {
 		return utils.TrackError(fmt.Errorf("failed to replace Cluster: %w", err))
 	}
 


machi1990 (approved):
/lgtm

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -129,7 +129,11 @@ func (c *clusterPropertiesSyncer) SyncOnce(ctx context.Context, key controllerut
 		return nil
 	}
 
-	if _, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Replace(ctx, replacement, nil); err != nil {
+	if _, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Replace(ctx, replacement, nil); database.IsPreconditionFailedError(err) {
```
miguelsorianod, line 132: Would it be cleaner, avoid ifelse/indentation:
```
err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Replace(ctx, replacement, nil)

if err != nil && database.IsPreconditionFailedError(err) {
  return nil
}
if err != nil {
  return err
}
```

Additionally, I understand this is not particular to this specific case only, but in general. Is the idea to in every replace of every type change it to this?

miguelsorianod: Does this also impact the time it takes to retry again because it's not an error anymore but a non error return?

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
```diff
@@ -129,7 +129,11 @@ func (c *clusterPropertiesSyncer) SyncOnce(ctx context.Context, key controllerut
...
+		// if we have a conflict error, then we're guaranteed that our informer will eventually see an update and trigger us again.
+		// there is no need to report an error since the retry will happen when the reflector sees the update and puts an Update into the informer.
+		return nil
```
swiencki, line 135: Possible for a follow up pr, consider adding a `V(1)` log line on the 412 branch (like the management cluster controllers do for cache-miss skips) so conflicts leave a breadcrumb when debugging.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
deads2k: > Additionally, I understand this is not particular to this specific case only, but in general. Is the idea to in every replace of every type change it to this?

the nil check in `if err != nil && database.IsPreconditionFailedError(err) {` is unnecessary.  Otherwise, yes, that's a good nesting cleanup.

backend/pkg/controllers/clusterpropertiescontroller/cluster_properties_sync.go
deads2k: > Possible for a follow up pr, consider adding a `V(1)` log line on the 412 branch (like the management cluster controllers do for cache-miss skips) so conflicts leave a breadcrumb when debugging.

We run the service.  Do you think we'd turn on V(1) for debugging? Do we  have a history of doing so?  At least one person I worked with didn't even know how to do it.

deads2k: Basically I'm saying that if we don't change log level (we basically never do in ROSA near as I can tell), we choose above or below the line.  Since we can redeploy at wil, we can add a log line when we need it if this is below the line (my strong suspicion)

PR #5593, opened 2026-06-10
 PoC cluster updates based on comparing CS state and cosmos desired state

miguelsorianod:
### PR Checklist
- [ ] Self-reviewed the diff


backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get cluster: %w", err))
+	}
+	if !clusterShouldProceed(cluster) {
+		return nil
+	}
+
```
deads2k, line 173: need to have a cooldown tha tprevents hitting this more than once per minute.

miguelsorianod:
the rollout process of this should be:
1. Introduction of backend changes. When introducing it, a noop should occur because we shouldn't get differences compared to the actual states. In this state, when a new update arrives, frontend will still send the request to CS and only afterwards will the new desired state be stored in Cosmos. At that point, the controller can kick-in but no difference should be detected in the dispatch, making it a no-op. The operation controller will perform the added state checks but that should not have an impact other than being more thorough state checks
2. Introduction of the frontend changes. With that, interaction with CS from frontend will be removed and when an update is arrived the desired state will be stored in Cosmos which will eventually be reconciled by the controllers

During cluster creation, the update dispatch controller will kick-in already but the desired content should match actual state so no changes should be triggered. If the update dispatch controller kicks in before the create one, the patch call to CS will fail and no change should occur. Once the cluster is created on CS side during the creation process, when the update controller kicks in no difference should be observed and no patch call should be sent to CS. Even if somehow it's sent, because of the CS cluster not being in ready state, CS should reject that request.

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
miguelsorianod: added

frontend/pkg/frontend/cluster.go
```diff
@@ -657,37 +655,11 @@ func (f *Frontend) updateHCPClusterInCosmos(ctx context.Context, writer http.Res
...
-			return utils.TrackError(err)
-		}
-	}
-
 	transaction := f.resourcesDBClient.NewTransaction(oldInternalCluster.ID.SubscriptionID)
 	clusterUpdateOperation := database.NewOperation(
 		database.OperationRequestUpdate,
 		oldInternalCluster.ID,
-		ptr.Deref(oldInternalCluster.ServiceProviderProperties.ClusterServiceID, api.InternalID{}),
+		api.InternalID{},
```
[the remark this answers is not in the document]
miguelsorianod: Known issue. It will be fixed once github.com/Azure/ARO-HCP/pull/4821 is merged

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -114,46 +138,87 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
+	clusterCSID := existingCluster.ServiceProviderProperties.ClusterServiceID
+	existingCSCluster, err := c.clusterServiceClient.GetCluster(ctx, *clusterCSID)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to get cluster from cluster service: %w", err))
+	}
+
+	existingServiceProviderCluster, err := c.serviceProviderClusterLister.Get(ctx, operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName, operation.ExternalID.Name)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to get service provider cluster from cache: %w", err))
+	}
```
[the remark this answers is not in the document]
miguelsorianod: not an issue in this context. When an update is issue the cluster should have been created and therefore the SPC too, which should have been cached a long time ago. If it doesn't for some reason it's ok to return an error

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
```diff
@@ -0,0 +1,436 @@
...
+			)
+		}
+		if len(csObservedAuthorizedCIDRs) > 0 {
+			return false, fmt.Sprintf(
+				"Cluster Service k8sAPIServerAuthorizedCIDRs is %s, want %s",
+				csObservedMessage, ocm.CSCIDRBlockAllowAccessModeAllowAll,
+			)
+		}
+		return true, ""
+	}
```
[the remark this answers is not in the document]
miguelsorianod: Not an issue. CS should explicitly set allow_all iif there are no blocks defined

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	serviceProviderCluster, err := c.resourcesDBClient.ServiceProviderClusters(cluster.ID.SubscriptionID, cluster.ID.ResourceGroupName, cluster.ID.Name).Get(ctx, api.ServiceProviderClusterResourceName)
+	if database.IsNotFoundError(err) {
+		// We expect the service provider cluster to be created when the cluster exists. If it doesn't exist, we wait for the next sync.
+		logger.Info("service provider cluster not found, waiting")
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get service provider cluster: %w", err))
+	}
```
deads2k, line 183: do this before calling cluster-serivce.

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	// We marshal the desired config coming from cosmos and the actual config coming from cluster service into JSON for
+	// logging purposes
+	desiredConfigJSON, err := ocm.ClusterUpdateDispatchConfigJSONFromRP(cluster, serviceProviderCluster)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	actualConfigJSON, err := ocm.ClusterUpdateDispatchConfigJSONFromCS(clusterServiceCluster)
+	if err != nil {
+		return utils.TrackError(err)
+	}
```
deads2k: could we do this and simply diff the json instead of `ClusterUpdateDispatchConfigDiffers`?

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	logger.Info("update dispatch config differs between RP and CS",
```
deads2k, line 208: add a `cmp.Diff` as a logging key.

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	actualConfigJSON, err := ocm.ClusterUpdateDispatchConfigJSONFromCS(clusterServiceCluster)
+	if err != nil {
+		return err
+	}
+
+	logger.Info("update dispatch config differs between RP and CS",
+		"clusterServiceID", clusterCSID.String(),
+		"desiredConfig", desiredConfigJSON,
+		"actualConfig", actualConfigJSON,
+	)
```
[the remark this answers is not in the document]
deads2k: bad bot comment.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -114,46 +138,87 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
 	operationStates := []*operationState{}
 
-	if operationState, err := c.desiredVersionResolutionOperationState(ctx, operation); err != nil {
+	if operationState, err := c.desiredVersionResolutionOperationState(ctx, operation, existingCluster, existingServiceProviderCluster); err != nil {
 		errs = append(errs, utils.TrackError(err))
 	} else {
-		operationStates = append(operationStates, operationState)
+		operationStates = append(operationStates, operationState.withSource("controlPlaneDesiredVersionResolution"))
+	}
+	if operationState, csErr := c.clusterServiceClusterStatusOperationState(ctx, operation, existingCSCluster.Status()); csErr != nil {
```
deads2k: if `ClusterClusterServiceUpdateDispatch` never runs, what keeps this from returning "succeeded".  Simple option
1. create a field indicating the last operationID which `ClusterUpdateDispatchConfigJSONFromRP` was logically  used to write to clusterservice
2. read clusterservice and make sure that `ClusterUpdateDispatchConfigJSONFromCS` matches before making this call.
3. if they do not match, indicate something not-finished.

miguelsorianod: What I had in mind is that because CS status check is not the only state calculation source, even though that specific state partial check might not be accurate, the overall operation won't succeed because the other checks won't be passing. For example, imagine that someone changes the CIDR block list and adds a new one:
If the operation controller runs before that's been dispatched, the CS check will be marked as passed, but the check that verifies that the HostedCluster **spec** has the defined CIDRs won't pass, having the overall operation as updating.

The downside is that the partial check might be inaccurate temporarily

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
miguelsorianod: I can move it, but the position here was intentional, where we avoid hitting the DB if the get call to CS fails for some reason

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
miguelsorianod: It is technically doable but I preferred hiding that detail into ClusterUpdateDispatchConfigDiffers.
Calculating the diff of the json also makes us needing to calculate it, which if we don't leverage a library might be non trivial. If we leverage a library we could use https://github.com/google/go-cmp but it's described by themselves as:
```
It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal. 
```

Because the hash calculation works, and I didn't want the controllers needing to be aware of the details I abstracted it in the `ClusterUpdateDispatchConfigDiffers` function.

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
miguelsorianod: I thought logging the diff would be useful too, but see https://github.com/Azure/ARO-HCP/pull/5593#discussion_r3542380282

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
miguelsorianod: An alternative, if we really care about the specific partial check not being accurate temporarily could be:

In the operation controller, because we need to do the full CS get call anyway, as we also do state calculations using parts of the returned CS information too, is to convert desired cosmos to the update dispatch config, as well as what's returned by the CS get call. Then compare them and if they differ we don't call this check at all or we mark it as updating.

This means the update operation controllers will do a full CS Get call, which they are currently doing in all the proposed update PRs.

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
miguelsorianod: hmm because the hash is equal a direct equality comparison might be enough.

However, for the comment in https://github.com/Azure/ARO-HCP/pull/5593#discussion_r3538653173 we would still need to take a decision.

If we are fine with using go-cmp even with that disclaimer we could leverage it.

backend/pkg/controllers/clusterupdate/cluster_cluster_service_update_dispatch_controller.go
```diff
@@ -206,6 +197,10 @@ func (c *clusterClusterServiceUpdateDispatchSyncer) SyncOnce(ctx context.Context
 	if err != nil {
 		return utils.TrackError(err)
 	}
+	if desiredConfigJSON == actualConfigJSON {
```
deads2k, line 200: bytes.Equal?

miguelsorianod:
Superseded by https://github.com/Azure/ARO-HCP/pull/6047

PR #5597, opened 2026-06-10
ARO-27686 | feat(observability): expose subscription_id on KSM HostedCluster metrics

swiencki:
https://redhat.atlassian.net/browse/ARO-27686

### What

Add `subscription_id` to the KSM custom-resource-state config for HostedCluster CRDs, mapping `HostedCluster.Spec.Platform.Azure.SubscriptionID` as a label on all `hostedClusterAPI_*` metrics.

### Why

Downstream alert filtering and dashboards need `subscription_id` to distinguish internal/dev subscriptions from customer subscriptions. This label is the foundation for excluding internal subs from RP lane alerting (generator changes in a follow-up PR).

### Testing

Helm test fixtures regenerated via `UPDATE=true go test ./tooling/helmtest/...`, all existing tests pass.

### DIFF
--- a/observability/prometheus/values-mgmt.yaml
+++ b/observability/prometheus/values-mgmt.yaml
@@ -101,6 +101,7 @@ kube-prometheus-stack:
               namespace: [metadata, namespace]
               api_url: [spec, services, "[service=APIServer]", servicePublishingStrategy, route, hostname]
               _id: [spec, clusterID]
+              subscription_id: [spec, platform, azure, subscriptionID]
             metrics:
             - name: "kubeapiserver_available"
               help: "KubeAPIServer availability condition"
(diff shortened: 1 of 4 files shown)


observability/prometheus/values-mgmt.yaml
```diff
@@ -101,6 +101,7 @@ kube-prometheus-stack:
               namespace: [metadata, namespace]
               api_url: [spec, services, "[service=APIServer]", servicePublishingStrategy, route, hostname]
               _id: [spec, clusterID]
```
machi1990, line 103: Do you think it'll be useful to also also expose the resource id ?

it'll be in the annotation with key `azure.microsoft.com/hcp-cluster-azure-resource-id`

swiencki: Good idea.

One option would be to expose it as a separate `hostedClusterAPI_info` metric with `type: Info` in the KSM config rather than adding it as a label on every metric. That keeps the per-sample payload lean while still making the resource ID available via a `group_left` join in dashboards and alerts.

machi1990: That could work as well. I'll leave that to you then if you want to do it here or as a followup

swiencki: Opened #5636 as a follow-up for this. Went with adding `resource_id` as a label (same pattern as `_id` and `subscription_id`) rather than a separate info metric, since it's consistent with the existing config.

PR #5600, opened 2026-06-10
Refine node pool version controllers

deads2k:
This separates the responsibility of setting active versions from setting desired version.

This removes cluster-service from the active version and desired version flow.

This will resolve failures caused by cluster-service nodepools missing during deletion.

### DIFF
--- a/backend/pkg/controllers/upgradecontrollers/nodepool_active_version_controller.go
+++ b/backend/pkg/controllers/upgradecontrollers/nodepool_active_version_controller.go
@@ -1,0 +1,149 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package upgradecontrollers
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/blang/semver/v4"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/backend/pkg/informers"
+	"github.com/Azure/ARO-HCP/backend/pkg/listers"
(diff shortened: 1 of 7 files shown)


backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go
```diff
@@ -84,72 +91,96 @@ func NewNodePoolVersionController(
...
+		return true
+	}
+	customerDesiredVersion, err := semver.ParseTolerant(nodePool.Properties.Version.ID)
+	if err != nil {
+		// Unparseable version; let SyncOnce surface the panic/error path —
+		// we don't suppress work just because we can't parse it here.
+		return true
+	}
+	return !customerDesiredVersion.EQ(*serivceProviderNodePool.Spec.NodePoolVersion.DesiredVersion)
+}
```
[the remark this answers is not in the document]
deads2k: naw, dont' accept nil. panic there is fine

backend/pkg/controllers/upgradecontrollers/nodepool_active_version_controller.go
```diff
@@ -0,0 +1,149 @@
...
+		// once the kube-applier writes status.
+		return nil
+	}
+	if len(hsNodePool.Status.Version) == 0 {
+		// Mirror exists but the Hypershift NodePool hasn't reported a
+		// version yet; nothing to record.
+		return nil
+	}
+
+	actualVersion, err := semver.ParseTolerant(hsNodePool.Status.Version)
```
ahitacat, line 126: This version should be with X.Y.Z, right? Why not using MustParse instead?

backend/pkg/controllers/upgradecontrollers/nodepool_active_version_controller.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to parse NodePool Status.Version %q: %w", hsNodePool.Status.Version, err))
+	}
+
+	oldActiveVersions := cachedServiceProviderNodePool.Status.NodePoolVersion.ActiveVersions
```
ahitacat, line 131: I'm inclined to use nodesInfo for the current activeVersions. Currently we store the previous Version watched and the new one. But this is not 100% correct, nodesInfo gives us the current version applied to the nodes.

With the current calculation we can end up saying that a nodepool has 4.19 and 4.20 activeVersion when in reality it has been completely upgrade to 4.20 and no nodes are in 4.19 version. 

Addressed in followed up PR https://github.com/Azure/ARO-HCP/pull/5603

machi1990 (commented):
is this superseded by https://github.com/Azure/ARO-HCP/pull/5603 ?

PR #5614, opened 2026-06-11
refactor: read kusto region from config instead of deriving it from name

mmazur:
## Why

`kusto.location` in the rendered config already encodes the region where the Kusto cluster lives. There's no need to re-derive it from the kusto cluster name via a hardcoded geo short-ID map.

Drop `resolveKustoRegion` and `kustoGeoToRegion` and read `kusto.location` directly, consistent with how all other kusto config fields are already read.


stevekuznetsov (approved):
/lgtm
/approve

PR #5617, opened 2026-06-11
feat: add DesiredHostedClusterControlPlaneSize knob

deads2k:
Adds an SRE-selectable control-plane sizing tier (Small/Medium/Large) on ServiceProviderClusterSpec, an admin POST handler that sets only that field, and a backend controller that observes it. The controller just logs the value for now; once cluster-service exposes the field, the controller will forward it.


internal/api/types_serviceprovider_cluster.go
```diff
@@ -36,6 +36,26 @@ const (
...
+// HostedClusterControlPlaneSize enumerates the SRE-selectable control-plane
+// sizing tiers we expose for a HostedCluster. Stored as a *string on
+// ServiceProviderClusterSpec so unset is distinguishable from "explicitly chosen."
+type HostedClusterControlPlaneSize string
+
+const (
+	HostedClusterControlPlaneSizeSmall  HostedClusterControlPlaneSize = "Small"
+	HostedClusterControlPlaneSizeMedium HostedClusterControlPlaneSize = "Medium"
+	HostedClusterControlPlaneSizeLarge  HostedClusterControlPlaneSize = "Large"
+)
```
machi1990, line 50: There are also xlarge and xxlarge size;

See https://github.com/deads2k/ARO-HCP/blob/cb1d0d1846947db91f661d541e32eead972863f6/hypershiftoperator/deploy/templates/cluster.clustersizingconfiguration.yaml#L64 

I was thinking we could do it dynamically by reading the size configuration off the management cluster or but we are better off hardcoding all the sizes here; so let's add the missing ones

internal/api/types_serviceprovider_cluster.go
```diff
@@ -67,6 +87,13 @@ type ServiceProviderClusterSpec struct {
 	// Once this contains the critical values, we will create it on management clusters.
 	// We may or may not choose to store the actual state in status.  We may choose to store the actual state independently.
 	DesiredHostedCluster *v1beta1.HostedCluster `json:"desiredHostedCluster,omitempty"`
+
+	// DesiredHostedClusterControlPlaneSize is the SRE-selected control plane
+	// sizing tier (one of "Small", "Medium", "Large") to apply to the
```
machi1990: There is also xlarge, xxlarge as sizes

backend/pkg/controllers/clusterpropertiescontroller/desired_control_plane_size_sync.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get ServiceProviderCluster from cache: %w", err))
+	}
+	if !c.NeedsWork(cachedSPC) {
+		return nil
+	}
+
+	// TODO: once cluster-service exposes a control-plane-size field, replace
+	// this log line with the CS write that propagates the value.
```
machi1990: The small challenge with this is that; updating the cluster.properties in CS will move the cluster to `updating` state in CS and potentially customer update calls done by the frontend will get an error from CS until the cluster moves back to ready state. I guess that's acceptable since for this to happen the customer has to ask for it

machi1990: If we agree that we use the .properties for this, I believe the CS call can be wired even right now. Assuming that we agree on the expected consts to pass;

CS uses the size naming case as they are used in the clustersizingconfig e.g`e2e_minimal` - https://github.com/openshift-online/aro-hcp-clusters-service/blob/f2727191ba0f0e78d25419b7fa69d5176d994a09/pkg/clusterprovisioner/acm/aro/hosted_cluster_size_override.go#L14 
as long as we agree to the values, cs can incorporate those it its change.

admin/server/handlers/hcp/desiredcontrolplanesize.go
```diff
@@ -0,0 +1,89 @@
...
+// use a pointer-string so a missing field is distinguishable from an
+// explicit empty string (which we reject).
+type desiredControlPlaneSizeRequest struct {
+	Size *string `json:"size"`
+}
+
+func (h *HCPDesiredControlPlaneSizeHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) error {
+	resourceID, err := utils.ResourceIDFromContext(request.Context())
+	if err != nil {
+		return fmt.Errorf("failed to get resource ID: %w", err)
```
rogbas: Shouldn't it be returning an `arm.CloudError` here instead?

machi1990 (changes requested):
When we branch in the CS code, let's also take care of this code path https://github.com/deads2k/ARO-HCP/blob/cb1d0d1846947db91f661d541e32eead972863f6/internal/ocm/convert.go#L417 which is exercised on update so that: 

1. we don't remove the property when the cluster is updated 
2. when the value is unset via the admin API (I bet we want to support for this), we remove it 

@deads2k

backend/pkg/controllers/clusterpropertiescontroller/desired_control_plane_size_sync.go
deads2k: > The small challenge with this is that; updating the cluster.properties in CS will move the cluster to updating state in CS and potentially customer update calls done by the frontend will get an error from CS until the cluster moves back to ready state. I guess that's acceptable since for this to happen the customer has to ask for it


I'm wiring this up now and realizing that it won't do that because updates haven't moved.  I'm thinking through options now

internal/ocm/convert.go
```diff
@@ -424,6 +424,31 @@ func BuildCSCluster(resourceID *azcorearm.ResourceID, tenantID string, hcpCluste
...
+// path) so the two cannot disagree.
+//
+// Precedence:
+//  1. ServiceProviderCluster.Spec.DesiredHostedClusterControlPlaneSize, when
+//     set, wins.
+//  2. Otherwise, hcpCluster.ServiceProviderProperties.ExperimentalFeatures.
+//     ControlPlanePodSizing == MinimalControlPlanePodSizing sets the legacy
+//     CSPropertyEnabled sentinel.
+//  3. Otherwise the property is absent.
+func DesiredHostedClusterSizeOverride(serviceProviderCluster *api.ServiceProviderCluster, hcpCluster *api.HCPOpenShiftCluster) (string, bool) {
```
machi1990, line 450: As discussed via chat, CS expects the values to use lower case as in https://github.com/Azure/ARO-HCP/blob/main/hypershiftoperator/deploy/templates/cluster.clustersizingconfiguration.yaml#L64

We need to lowercase our the `Spec.DesiredHostedClusterControlPlaneSize` before feeding it onto CS

internal/ocm/convert.go
```diff
@@ -424,6 +424,31 @@ func BuildCSCluster(resourceID *azcorearm.ResourceID, tenantID string, hcpCluste
...
+	if serviceProviderCluster != nil && serviceProviderCluster.Spec.DesiredHostedClusterControlPlaneSize != nil {
+		return *serviceProviderCluster.Spec.DesiredHostedClusterControlPlaneSize, true
+	}
+	if hcpCluster != nil && hcpCluster.ServiceProviderProperties.ExperimentalFeatures.ControlPlanePodSizing == api.MinimalControlPlanePodSizing {
+		return CSPropertyEnabled, true
```
machi1990: Here we could return `e2e_minimal` and get rid of the old behaviour that was relying on bool.

backend/pkg/controllers/clusterpropertiescontroller/desired_control_plane_size_sync.go
machi1990: Once updates are moved to the backend this solves the CS' state management; 
Context on the CS behaviour; CS only allows a cluster to be updatable when it's state is `ready`: see https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/cmd/clusters-service/service/azure/aro_hcp_cluster_service.go#L553 . When a cluster is updated, CS moves the state of it to `pending_updating`. update of this field. 
If the admin API is called to change the size; CS state will be `pending_update` until that change lands in the management cluster and the state will move back to `ready`. If at the same time (before the cluster moves back to `ready`) the customer wants to change another field within their control, the request in CS will fail with 400: https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/cmd/clusters-service/service/azure/aro_hcp_cluster_service.go#L554 and the backend will propagate it back to the customer: https://github.com/Azure/ARO-HCP/blob/main/frontend/pkg/frontend/cluster.go#L684

This behaviour happens due to the sync call made from the frontend to CS; if we move it to the background, it eventually reconciles and the customer won't get the 400.

But as mentioned above, it's fine since in these circumstances, the customer via support ticket or something, asked for a bump of the size, when they get a 400, it's an easy sell to explain to them what happened. That's can get us moving until the Update call to CS from the frontend is moved to the backend.

internal/api/types_serviceprovider_cluster.go
```diff
@@ -36,6 +36,32 @@ const (
 	ServiceProviderClusterResourceName = "default"
 )
 
+// HostedClusterControlPlaneSize enumerates the SRE-selectable control-plane
+// sizing tiers we expose for a HostedCluster. Stored as a *string on
+// ServiceProviderClusterSpec so unset is distinguishable from "explicitly chosen."
+type HostedClusterControlPlaneSize string
+
+const (
```
machi1990, line 44: I was thinking if we need to include the `E2E_Minimal` size and after self debating I don't think so;
This size isn't meant to be used by the customer.
It's an internal only size, we use the AFEC flag - experimental one to set this.
No need to expose them here and only leave the available values to size matching prod workload that can be used by customers

machi1990 (commented):
The code change looks good, do you think it's possible to add an e2e? @deads2k

PR #5622, opened 2026-06-12
fix: unify install version prediction

sclarkso:
### What

Replace the control_plane_minor_upgrade E2E pre-check with one that predicts the install version using the same helper the backend uses (ocm.NewOpenShiftVersionXYZ), and filter pre-release versions out of the existing upgrade-path loop.

### Why

The test sends OpenshiftVersionId: "4.21" and the backend resolves bare minors to a specific z-stream via a hardcoded map (currently pinning 4.21 → 4.21.15). The existing pre-check loop validated upgrade paths from whichever 4.21.x Cincinnati happened to have an unconditional edge for — not from the z-stream the backend actually installs. When those two diverged, the test created a cluster that was guaranteed to fail its upgrade (no upgrade path found from 4.21.15 to 4.22.0). It also accepted pre-release upgrade targets (EC/RC) that aren't valid GA upgrade endpoints.

By predicting through NewOpenShiftVersionXYZ and rejecting pre-releases, the test now skips deterministically when no GA upgrade path exists from the version the backend will actually install — eliminating the wasted ~15-minute cluster create + spurious failure. The fix is generic and works for any minor in the test table, present or future.

### Testing

   - Static checks: go vet ./test/e2e/... and go build ./test/e2e/... pass cleanly.
   - Live Cincinnati simulation: Drove the new prediction + GA-filter code directly against api.openshift.com for all three production channels (stable, fast, candidate) and all currently-mapped minors (4.19→4.20,
    4.20→4.21, 4.21→4.22). Confirmed NewOpenShiftVersionXYZ predicts the exact z-streams the backend pins (4.19.30, 4.20.22, 4.21.15) and that the upgrade-path check correctly produces PROCEED for 4.19→4.20 and 4.20→4.21 and SKIP for all three 4.21→4.22 channels in today's graph state.
   - Cross-check against real failure artifacts: Confirmed that the failing
    4.21→4.22 INT run at 2026-06-11 20:13 UTC installed 4.21.15 and failed with no upgrade path found from 4.21.15 to 4.22.0 — the exact version and failure the new pre-check now predicts and skips on.
   - Cincinnati edge audit: Verified that 4.21.15 currently has no unconditional GA path to any
    4.22.x in candidate-4.22 / stable-4.22 / fast-4.22, and that the conditional-only paths (e.g. to 4.22.0 and 4.22.0-rc.*) would correctly be filtered out by the new GA check.

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/test/e2e/control_plane_minor_upgrade.go
+++ b/test/e2e/control_plane_minor_upgrade.go
@@ -37,6 +37,7 @@ import (
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/upgradecontrollers"
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/cincinnati"
+	"github.com/Azure/ARO-HCP/internal/ocm"
 	hcpsdk20240610preview "github.com/Azure/ARO-HCP/test/sdk/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
 	"github.com/Azure/ARO-HCP/test/util/framework"
 	"github.com/Azure/ARO-HCP/test/util/labels"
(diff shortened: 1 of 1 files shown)


machi1990:
> By predicting through NewOpenShiftVersionXYZ and rejecting pre-releases, the test now skips deterministically when no GA upgrade path exists from the version the backend will actually install — eliminating the wasted ~15-minute cluster create + spurious failure. The fix is generic and works for any minor in the test table, present or future.

If we reject GA versions, how are we going to be testing upgrades from X.Y that's released to an X.Y+1 that's not GA-ed yet ? @sclarkso

PR #5629, opened 2026-06-12
Support kmsKey.version rotation

ahitacat:
https://redhat.atlassian.net/browse/ARO-27470
https://redhat.atlassian.net/browse/ARO-27471

### What

Makes `KmsKey.Version` updatable via PATCH in the `v20260630preview` API so that customers can rotate their etcd encryption key version without recreating the cluster.

### Why

Customers using customer-managed encryption for etcd need to rotate their KMS key version periodically (security policy, key expiry, compliance). Today, all KMS fields — including `version` — are immutable after cluster creation. This forces a destructive cluster recreate just to update a key version, which is operationally unacceptable.

This PR is the **RP-side half** of the feature. Cluster Service is being updated separately to accept KMS key version changes on its PATCH endpoint. Both sides must be in place for end-to-end rotation to work, but neither breaks independently — the RP will send the version, and CS will either accept it (if updated) or ignore it (if not yet updated).

This PR has the changes for the ARM API spec and the frontend changes and adds E2E tests.

### Testing

**Unit tests**:
- Validation: "mutable kms key version" test verifies version-only change passes; "immutable kms visibility" expected errors updated to reflect removed parent-level guards
- Admission: 4 new test cases covering version-gated rotation at 4.21, 4.22, during upgrade, and no-change
- OCM conversion: "UPDATE - sends KMS key version" test verifies etcd encryption appears in the CS builder on update; `getBaseCSClusterBuilder(true)` updated to include default PlatformManaged etcd for update expectations

**Manual verification**: PATCH with key version change tested against dev environment confirming CS accepts the update and other fields remain unchanged.

**E2E tests**:
Added a simple E2E test that rotates the key version once and then disables the old key.
The test has a timebomb on the APIVersion being used. So it should be OK to merge, environments that doesn't have this API version will not use this test.
Example can be found in latest prod jobs https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-Azure-ARO-HCP-main-periodic-prod-e2e-parallel/2075037316002353152
Where the test `Customer should create a cluster using the 2026-06-30-preview API version` is skipped.

- The `@renamedFrom` suffix (`NoUpdate`) is a TypeSpec convention for this pattern — it doesn't appear in the generated API surface. The generated OpenAPI spec and SDK models are unchanged for older API versions. Verify this by checking that `internal/api/v20240610preview/generated/` and `internal/api/v20251223preview/generated/` are unmodified.

KMS key version rotation requires HyperShift changes that only land in OpenShift >= 4.22. The admission controller rejects key version changes when the cluster's **lowest active control plane version** is below 4.22. Why lowest and not requested? During a rolling upgrade, a cluster can have multiple active control plane versions (e.g. 4.22 + 4.23). Using the lowest ensures the rotation is only allowed when all active components support it.

### PR Checklist
- [x] Self-reviewed the diff

### Requirements to merge this change
- [x] CS rolled out to production
- [x] CPO override with the fix https://github.com/openshift/hypershift/pull/9010 merged
- [x] New HO image is deployed in ARO-HCP https://github.com/Azure/ARO-HCP/pull/6161

### DIFF
--- a/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
+++ b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
@@ -94,6 +94,13 @@ model HcpOpenShiftClusterProperties {
 
   /** Configure ETCD. */
   @visibility(Lifecycle.Read, Lifecycle.Create)
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "etcd")
+  etcdNoUpdate?: EtcdProfile;
+
+  /** Configure ETCD. */
+  @added(Versions.v2026_06_30_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   etcd?: EtcdProfile;
 
   /** imageDigestMirrors is a set of rules to allow pulling images from a
@@ -551,11 +558,21 @@ model KmsKey {
   /** name is the name of the keyvault key used for encryption/decryption. */
   @maxLength(255)
   @minLength(1)
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "name")
+  nameImplicitVisibility: string;
+
+  /** name is the name of the keyvault key used for encryption/decryption. */
+  @added(Versions.v2026_06_30_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  @maxLength(255)
+  @minLength(1)
   name: string;
 
   /** version contains the version of the key to use. */
   @maxLength(255)
   @minLength(1)
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   version: string;
 }
 
@@ -576,6 +593,15 @@ model EtcdProfile {
    * If not specified platform managed keys are used.
    */
   @visibility(Lifecycle.Read, Lifecycle.Create)
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "dataEncryption")
+  dataEncryptionNoUpdate?: EtcdDataEncryptionProfile;
+
+  /** ETCD Data Encryption settings.
+   * If not specified platform managed keys are used.
+   */
+  @added(Versions.v2026_06_30_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   dataEncryption?: EtcdDataEncryptionProfile;
 }
 
@@ -591,6 +617,15 @@ model EtcdDataEncryptionProfile {
    * Required when keyManagementMode is "CustomerManaged".
    */
   @visibility(Lifecycle.Read, Lifecycle.Create)
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "customerManaged")
+  customerManagedNoUpdate?: CustomerManagedEncryptionProfile;
+
+  /** Specify customer managed encryption key details.
+   * Required when keyManagementMode is "CustomerManaged".
+   */
+  @added(Versions.v2026_06_30_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   customerManaged?: CustomerManagedEncryptionProfile;
 }
 
@@ -607,6 +642,16 @@ model CustomerManagedEncryptionProfile {
    * Required when encryptionType is "KMS".
    */
   @visibility(Lifecycle.Read, Lifecycle.Create)
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "kms")
+  kmsNoUpdate?: KmsEncryptionProfile;
+
+  /** The Key Management Service (KMS) encryption key details.
+   *
+   * Required when encryptionType is "KMS".
+   */
+  @added(Versions.v2026_06_30_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   kms?: KmsEncryptionProfile;
 }
 
@@ -629,6 +674,13 @@ model KmsEncryptionProfile {
 
   /** The details of the active key. */
   @visibility(Lifecycle.Read, Lifecycle.Create)
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "activeKey")
+  activeKeyNoUpdate: KmsKey;
+
+  /** The details of the active key. */
+  @added(Versions.v2026_06_30_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   activeKey: KmsKey;
 }
 
--- a/backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
@@ -80,6 +80,9 @@ func (c *operationClusterUpdate) hypershiftHostedClusterSpecMatchesDesired(clust
 	if matches, message := c.hypershiftHostedClusterImageContentSourcesSpecMatchesDesired(cluster.CustomerProperties.ImageDigestMirrors, hostedCluster.Spec.ImageContentSources); !matches {
 		return false, message
 	}
+	if matches, message := c.hypershiftHostedClusterEtcdSecretEncryptionSpecMatchesDesired(cluster.CustomerProperties.Etcd.DataEncryption, hostedCluster.Spec.SecretEncryption); !matches {
+		return false, message
+	}
 	return true, ""
 }
 
@@ -331,6 +334,46 @@ func (c *operationClusterUpdate) hypershiftHostedClusterImageContentSourcesSpecM
 		return false, fmt.Sprintf("hypershift HostedCluster has unexpected imageContentSource %q", source)
 	}
 
+	return true, ""
+}
+
+func (c *operationClusterUpdate) hypershiftHostedClusterEtcdSecretEncryptionSpecMatchesDesired(desired api.EtcdDataEncryptionProfile, observed *v1beta1.SecretEncryptionSpec) (bool, string) {
+
+	if observed == nil {
+		return false, "unexpected hypershift HostedCluster secret encryption is not set"
+	}
+
+	if matches, message := c.hypershiftHostedClusterCustomerManagedSecretEncryptionSpecMatchesDesired(desired, observed); !matches {
+		return matches, message
+	}
+
+	return true, ""
+}
+
+func (c *operationClusterUpdate) hypershiftHostedClusterCustomerManagedSecretEncryptionSpecMatchesDesired(desired api.EtcdDataEncryptionProfile, observed *v1beta1.SecretEncryptionSpec) (bool, string) {
+	if desired.KeyManagementMode != api.EtcdDataEncryptionKeyManagementModeTypeCustomerManaged {
+		return false, fmt.Sprintf("support for desired key management mode %q for updates is not implemented", desired.KeyManagementMode)
+	}
+
+	if desired.CustomerManaged.EncryptionType != api.CustomerManagedEncryptionTypeKMS {
+		return false, fmt.Sprintf("support for desired customer managed key encryption type %s for updates is not implemented", desired.CustomerManaged.EncryptionType)
+	}
+
+	if observed.Type != v1beta1.KMS {
+		return false, fmt.Sprintf("hypershift HostedCluster secret encryption is: %q, want :%q", observed.Type, desired.CustomerManaged.EncryptionType)
+	}
+
+	if observed.KMS == nil {
+		return false, "unexpected hypershift HostedCluster kms secret encryption configuration unset"
+	}
+
+	if observed.KMS.Azure == nil {
+		return false, "unexpected hypershift HostedCluster kms secret encryption azure configuration unset"
+	}
+
+	if observed.KMS.Azure.ActiveKey.KeyVersion != desired.CustomerManaged.Kms.ActiveKey.Version {
+		return false, fmt.Sprintf("hypershift HostedCluster kms secret encryption active key version is: %q, want: %q", observed.KMS.Azure.ActiveKey.KeyVersion, desired.CustomerManaged.Kms.ActiveKey.Version)
+	}
 	return true, ""
 }
 
--- a/internal/admission/admit_cluster.go
+++ b/internal/admission/admit_cluster.go
@@ -277,6 +277,7 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
 	errs := field.ErrorList{}
 
 	errs = append(errs, admitClusterVersionProfile(ctx, admissionContext, op, fldPath.Child("version"), &newObj.Version, safe.Field(oldObj, validation.ToClusterCustomerPropertiesVersion))...)
+	errs = append(errs, admitClusterEtcdKmsKeyVersionChange(ctx, admissionContext, op, fldPath.Child("etcd", "dataEncryption", "customerManaged", "kms", "activeKey", "version"), newObj, oldObj)...)
 	errs = append(errs, admitClusterPlatform(ctx, admissionContext, op, fldPath.Child("platform"), &newObj.Platform)...)
 
 	return errs
@@ -462,4 +463,43 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
 	}
 
 	return errs
+}
+
+// minKmsKeyVersionRotationVersion is the minimum OCP version whose CPO
+// supports etcd KMS key version rotation.
+var minKmsKeyVersionRotationVersion = semver.Version{Major: 4, Minor: 22}
+
+// admitClusterEtcdKmsKeyVersionChange rejects KMS key version changes when the
+// cluster's active control plane version predates the CPO support (< 4.22).
+// Only runs for API versions >= v20260630Preview where version is mutable;
+// older API versions block version changes via the immutability check in validation.
+func admitClusterEtcdKmsKeyVersionChange(_ context.Context, admissionContext *ClusterAdmissionContext, op operation.Operation, fldPath *field.Path, newObj, oldObj *api.HCPOpenShiftClusterCustomerProperties) field.ErrorList {
+	if op.Type != operation.Update || oldObj == nil {
+		return nil
+	}
+
+	if newObj.Etcd.DataEncryption.KeyManagementMode != api.EtcdDataEncryptionKeyManagementModeTypeCustomerManaged && newObj.Etcd.DataEncryption.KeyManagementMode != oldObj.Etcd.DataEncryption.KeyManagementMode {
+		return field.ErrorList{field.Forbidden(fldPath, "KMS key version rotation is only supported for customer-managed encryption")}
+	}
+
+	if newObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version == oldObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version {
+		return nil
+	}
+
+	apiVersion := api.APIVersionFromOptions(op.Options)
+	if apiVersion.LT(api.APIVersionV20260630Preview) {
+		return field.ErrorList{field.Forbidden(fldPath, "KMS key version is immutable for this API version")}
+	}
+
+	if admissionContext.ServiceProviderCluster == nil {
+		return field.ErrorList{field.InternalError(fldPath, errors.New("cannot validate KMS key version rotation"))}
+	}
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(admissionContext.ServiceProviderCluster.Status.ControlPlaneVersion.ActiveVersions)
+	clusterVersion := semver.Version{Major: lowest.Major, Minor: lowest.Minor}
+	if clusterVersion.LT(minKmsKeyVersionRotationVersion) {
+		return field.ErrorList{field.Invalid(fldPath, newObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version, fmt.Sprintf("KMS key version rotation requires cluster version %s or above", minKmsKeyVersionRotationVersion))}
+	}
+
+	return nil
 }
--- a/internal/admission/admit_cluster_test.go
+++ b/internal/admission/admit_cluster_test.go
@@ -591,27 +591,48 @@ func TestAdmitCluster_Update(t *testing.T) {
 		}
 	}
 
+	kmsEtcdProfile := func(keyVersion string) api.EtcdProfile {
+		return api.EtcdProfile{
+			DataEncryption: api.EtcdDataEncryptionProfile{
+				KeyManagementMode: api.EtcdDataEncryptionKeyManagementModeTypeCustomerManaged,
+				CustomerManaged: &api.CustomerManagedEncryptionProfile{
+					Kms: &api.KmsEncryptionProfile{
+						ActiveKey: api.KmsKey{
+							Name:      "test-key",
+							VaultName: "test-vault",
+							Version:   keyVersion,
+						},
+					},
+				},
+			},
+		}
+	}
+
 	tests := []struct {
 		name                         string
 		oldClusterVersionID          string
-		clusterVersionID             string
+		channelGroup                 string
+		etcd                         api.EtcdProfile
+		options                      []string
 		serviceProviderClusterStatus api.ServiceProviderClusterStatus
 		nodePools                    []*api.HCPOpenShiftClusterNodePool
 		serviceProviderNodePools     []*api.ServiceProviderNodePool
+		newClusterFromOld            func(*api.HCPOpenShiftCluster) //This method uses a copy of the oldCluster, changes are applied to that copy.
 		expectErrors                 []utils.ExpectedError
 	}{
 		{
 			name:                         "empty desired version skips admission",
 			oldClusterVersionID:          "4.10",
-			clusterVersionID:             "",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.22"),
 			nodePools:                    []*api.HCPOpenShiftClusterNodePool{makeTestNodePool("np1", "4.10.0")},
-			expectErrors:                 []utils.ExpectedError{},
+			newClusterFromOld: func(oldCopy *api.HCPOpenShiftCluster) {
+				oldCopy.CustomerProperties.Version.ID = ""
+			},
+			expectErrors: []utils.ExpectedError{},
 		},
 		{
 			name:                         "unchanged version skips admission",
 			oldClusterVersionID:          "5.0",
-			clusterVersionID:             "5.0",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.22"),
 			nodePools:                    []*api.HCPOpenShiftClusterNodePool{makeTestNodePool("workers", "4.20.0")},
 			expectErrors:                 []utils.ExpectedError{},
@@ -619,9 +640,10 @@ func TestAdmitCluster_Update(t *testing.T) {
 		{
 			name:                         "unparsable old version id",
 			oldClusterVersionID:          "4.x",
-			clusterVersionID:             "4.22",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.22"),
-			nodePools:                    nil,
+			newClusterFromOld: func(c *api.HCPOpenShiftCluster) {
+				c.CustomerProperties.Version.ID = "4.22"
+			},
 			expectErrors: []utils.ExpectedError{
 				{FieldPath: "properties.version.id", Message: "Invalid character(s) found in minor number"},
 			},
@@ -629,25 +651,28 @@ func TestAdmitCluster_Update(t *testing.T) {
 		{
 			name:                         "skips skew vs lowest when old minor matches lowest active cluster version",
 			oldClusterVersionID:          "4.21",
-			clusterVersionID:             "4.23",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.21"),
-			nodePools:                    nil,
-			expectErrors:                 []utils.ExpectedError{},
+			newClusterFromOld: func(c *api.HCPOpenShiftCluster) {
+				c.CustomerProperties.Version.ID = "4.23"
+			},
+			expectErrors: []utils.ExpectedError{},
 		},
 		{
 			name:                         "allows 4.22 to 5.0 with active cluster version 4.22",
 			oldClusterVersionID:          "4.22",
-			clusterVersionID:             "5.0",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.22"),
-			nodePools:                    nil,
-			expectErrors:                 []utils.ExpectedError{},
+			newClusterFromOld: func(c *api.HCPOpenShiftCluster) {
+				c.CustomerProperties.Version.ID = "5.0"
+			},
+			expectErrors: []utils.ExpectedError{},
 		},
 		{
 			name:                         "rejects 5.1 when old minor below lowest active cluster version",
 			oldClusterVersionID:          "4.21",
-			clusterVersionID:             "5.1",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.22"),
-			nodePools:                    nil,
+			newClusterFromOld: func(c *api.HCPOpenShiftCluster) {
+				c.CustomerProperties.Version.ID = "5.1"
+			},
 			expectErrors: []utils.ExpectedError{
 				{FieldPath: "properties.version.id", Message: "invalid upgrade path"},
 			},
@@ -655,9 +680,10 @@ func TestAdmitCluster_Update(t *testing.T) {
 		{
 			name:                         "rejects 4.24 when old minor below lowest active cluster version",
 			oldClusterVersionID:          "4.21",
-			clusterVersionID:             "4.24",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.22"),
-			nodePools:                    nil,
+			newClusterFromOld: func(c *api.HCPOpenShiftCluster) {
+				c.CustomerProperties.Version.ID = "4.24"
+			},
 			expectErrors: []utils.ExpectedError{
 				{FieldPath: "properties.version.id", Message: "only upgrade to the next minor is allowed"},
 			},
@@ -665,9 +691,10 @@ func TestAdmitCluster_Update(t *testing.T) {
 		{
 			name:                         "rejects version below highest active cluster version",
 			oldClusterVersionID:          "4.22",
-			clusterVersionID:             "4.21",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.22"),
-			nodePools:                    nil,
+			newClusterFromOld: func(c *api.HCPOpenShiftCluster) {
+				c.CustomerProperties.Version.ID = "4.21"
+			},
 			expectErrors: []utils.ExpectedError{
 				{FieldPath: "properties.version.id", Message: "must be at least"},
 			},
@@ -675,17 +702,19 @@ func TestAdmitCluster_Update(t *testing.T) {
 		{
 			name:                         "allows upgrade across adjacent active cluster minors",
 			oldClusterVersionID:          "4.21",
-			clusterVersionID:             "4.22",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersions("4.22", "4.21"),
-			nodePools:                    nil,
-			expectErrors:                 []utils.ExpectedError{},
+			newClusterFromOld: func(c *api.HCPOpenShiftCluster) {
+				c.CustomerProperties.Version.ID = "4.22"
+			},
+			expectErrors: []utils.ExpectedError{},
 		},
 		{
 			name:                         "rejects skip minor vs lowest when fleet spans minors",
 			oldClusterVersionID:          "4.21",
-			clusterVersionID:             "4.22",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersions("4.20", "4.22"),
-			nodePools:                    nil,
+			newClusterFromOld: func(c *api.HCPOpenShiftCluster) {
+				c.CustomerProperties.Version.ID = "4.22"
+			},
 			expectErrors: []utils.ExpectedError{
 				{FieldPath: "properties.version.id", Message: "only upgrade to the next minor is allowed"},
 			},
@@ -693,9 +722,11 @@ func TestAdmitCluster_Update(t *testing.T) {
 		{
 			name:                         "rejects when node pool over two minors behind",
 			oldClusterVersionID:          "4.20",
-			clusterVersionID:             "4.21",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.20"),
 			nodePools:                    []*api.HCPOpenShiftClusterNodePool{makeTestNodePool("workers", "4.17.0")},
+			newClusterFromOld: func(c *api.HCPOpenShiftCluster) {
+				c.CustomerProperties.Version.ID = "4.21"
+			},
 			expectErrors: []utils.ExpectedError{
 				{FieldPath: "properties.version.id", Message: "must not be more than two minor versions ahead"},
 			},
@@ -703,7 +734,6 @@ func TestAdmitCluster_Update(t *testing.T) {
 		{
 			name:                         "allows no-op version with node pools in skew",
 			oldClusterVersionID:          "4.20",
-			clusterVersionID:             "4.20",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.20"),
 			nodePools: []*api.HCPOpenShiftClusterNodePool{
 				makeTestNodePool("workers", "4.18.0"),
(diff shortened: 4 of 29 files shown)


api/redhatopenshift/HcpCluster.Management/main.tsp
```diff
@@ -25,4 +25,8 @@ enum Versions {
   /** 2025-12-23-preview version */
   @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v6)
   v2025_12_23_preview: "2025-12-23-preview",
+
+  /** 2026-06-30-preview version */
+  @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v6)
+  v2026_06_30_preview: "2026-06-30-preview",
```
ahitacat, line 31: Added this to make the change only related to the new version. This should be equal to other API changes.

api/redhatopenshift/HcpCluster.Management/hcpCluster-models.tsp
ahitacat, line 570: I wasn't sure if I need to add the` @visibility(Lifecycle.Read, Lifecycle.Create)` to these as well. If so, I do this need to be also in the vaultName even there is no part of the latest api?

mbarnes: Omitting it implies the default `@visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)`.
But for clarity's sake, explicit is better than implicit, imo.

mbarnes: If you're not sure, generate the `openapi.json` with your change and check the diff.

api/redhatopenshift/HcpCluster.Management/main.tsp
mbarnes: :+1: I'd like to see a baseline of the new version merged so all the PRs that add features for it don't have to do this.  I'll check with @bennerv about it.

internal/admission/admit_cluster.go
```diff
@@ -260,3 +248,32 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+	if newObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version == oldObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version {
+		return nil
+	}
+
+	if admissionContext.ServiceProviderCluster == nil {
+		return field.ErrorList{field.InternalError(fldPath, errors.New("cannot validate KMS key version rotation"))}
+	}
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(admissionContext.ServiceProviderCluster.Status.ControlPlaneVersion.ActiveVersions)
+	if lowest != nil && lowest.LT(minKmsKeyVersionRotationVersion) {
```
machi1990: You want to be comparing major.minor only here.
e.g a nightly version will be false here and wrongly return an error.

See https://go.dev/play/p/ZJcZu3Uy_Nx as an example.

internal/admission/admit_cluster_test.go
```diff
@@ -586,14 +538,52 @@ func TestAdmitCluster_Update(t *testing.T) {
 			serviceProviderNodePools:     []*api.ServiceProviderNodePool{makeServiceProviderNodePool("workers", "4.22.1", "4.22.0")},
 			expectErrors:                 []utils.ExpectedError{},
 		},
+		{
```
machi1990, line 896: Also add a test case for nightly version and watch it fail before fixing https://github.com/Azure/ARO-HCP/pull/5629#discussion_r3481569015

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -546,11 +524,29 @@ model KmsKey {
   /** name is the name of the keyvault key used for encryption/decryption. */
   @maxLength(255)
   @minLength(1)
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "name")
+  nameNoUpdate: string;
```
mbarnes: What's the intended change here?  The absence of a `@visibility` decorator implies this field is updateable but it's named `nameNoUpdate`.  But `name` was already create-only, wasn't it?

Maybe the original `name` field just needed an explicit `@visibility(Lifecycle.Read, Lifecycle.Create)` decorator?  See if that introduces unwanted changes in older versions.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -546,11 +524,29 @@ model KmsKey {
...
+  @maxLength(255)
+  @minLength(1)
   name: string;
 
   /** version contains the version of the key to use. */
   @maxLength(255)
   @minLength(1)
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "version")
+  versionNoUpdate: string;
```
mbarnes: Here too, for `versionNoUpdate` I think we want an explicit `@visibility(Lifecycle.Read, Lifecycle.Create)` because this looks like it's implicitly updateable as written.

mbarnes (changes requested):
Mainly focused on the TypeSpec for a first pass.  Spotted what looks like a couple visibility issues with `KmsKey` fields.  I don't see changes in the older API versions so they may not be issues in practice but I'd still request changes for clarity.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
ahitacat: Yup exactly is to solve that issue.
I first added the decorator `@visibility(Lifecycle.Read, Lifecycle.Create)`, as it was not explicit. That changed the openapi for older versions.

I have just followed the same pattern with the names, but they can be changed. Although, if we are OK with the change to older versions done by added the visibility explicit I can remove this new fields.

internal/admission/admit_cluster.go
ahitacat: good catch!
Thinking about this, this should also check for the exact .z Thus, not all the 4.22 clusters will have this feature.

machi1990: Well hypershift can provide the override of the CPO for all the previous GA-d .z in 4.22.

machi1990: i.e once the backport PR is merged, ask for the hypershift override of the CPO image to be provided so that 4.22.1 or 4.22.0 have this feature until when a new z is available with this change that the clusters will automatically upgrade too. An example of that done for previous features is https://github.com/openshift/hypershift/pull/8610

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mbarnes: I see now.  The field was meant to be create-only but was mistakenly left implicitly updatable even though the parent model was create-only.

Can you just add a comment there saying the same, because it looks really weird otherwise.  And maybe name the variable something like `nameImplicitVisibility` instead of `nameNoUpdate` which is misleading.

I think we're fine in terms of RP validation; I'll double-check but I think we would have captured the intent.  Possible bug in CAPI depending on how they interpret Swagger specs.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mbarnes: (round two)  I still think this change can be dropped.  Adding an explicit `@visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)` introduces an explicit `x-ms-mutability` attribute into the older Swagger versions but it's a backward-compatible change.

[Here's where the TypeSpec docs say no visibility decorator has the same semantics as read/create/update](https://azure.github.io/typespec-azure/docs/howtos/arm/resource-type/#property-visibility-and-other-constraints).

ahitacat: Ok I will make this change, drop the versionNoUpdate and just add the explicit visibility.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -517,11 +524,21 @@ model KmsKey {
...
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  @maxLength(255)
+  @minLength(1)
   name: string;
 
   /** version contains the version of the key to use. */
   @maxLength(255)
   @minLength(1)
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   version: string;
```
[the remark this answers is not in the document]
mbarnes: The change for older API versions is to simply tag the `version` field as _explicitly_ updateable instead of leaving it _implicitly_ updateable.  There's no actual semantic change so it's backward-compatible.

Notice there's no change to the Go models for older API versions.

mbarnes:
Informal LGTM on the TypeSpec changes.

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,43 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+	if newObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version == oldObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version {
+		return nil
+	}
+
+	if admissionContext.ServiceProviderCluster == nil {
+		return field.ErrorList{field.InternalError(fldPath, errors.New("cannot validate KMS key version rotation"))}
+	}
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(admissionContext.ServiceProviderCluster.Status.ControlPlaneVersion.ActiveVersions)
+	if lowest != nil {
```
machi1990: Consider this scenario:

`activeVersions: [{version: 4.21.19, state=Partial}, {4.22.4, state=Partial}]` this can happen because of lack of https://issues.redhat.com/browse/ARO-22686: CP upgrades being decoupled from DP state. 

In this scenario, lowest will be 4.21.19, do we want to allow the rotation or not? 

If we rely on the cluster X.Y version (the desired one) will we be okay here and achieve the same thing?

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,43 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
 
 	return errs
 }
+
+var minKmsKeyVersionRotationVersion = semver.Version{Major: 4, Minor: 22, Patch: 4}
```
machi1990: I wonder if for all the channels we can simply consider X.Y and do
X.Y < 4.22 then return error, otherwise return nil

See https://github.com/Azure/ARO-HCP/pull/5629/#discussion_r3491361106 for what's required.

This will simplify the logic in the function `admitClusterEtcdKmsKeyVersionChange(...)`

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,43 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+		return nil
+	}
+
+	if admissionContext.ServiceProviderCluster == nil {
+		return field.ErrorList{field.InternalError(fldPath, errors.New("cannot validate KMS key version rotation"))}
+	}
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(admissionContext.ServiceProviderCluster.Status.ControlPlaneVersion.ActiveVersions)
+	if lowest != nil {
+		if oldObj.Version.ChannelGroup == "nightly" {
```
machi1990: see the proposal here https://github.com/Azure/ARO-HCP/pull/5629/#discussion_r3499039949

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,43 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+			lowestMajorMinor := semver.Version{Major: lowest.Major, Minor: lowest.Minor}
+			minMajorMinor := semver.Version{Major: minKmsKeyVersionRotationVersion.Major, Minor: minKmsKeyVersionRotationVersion.Minor}
+			if lowestMajorMinor.LT(minMajorMinor) {
+				return field.ErrorList{field.Forbidden(fldPath, fmt.Sprintf("KMS key version rotation requires cluster version %d.%d or above", minKmsKeyVersionRotationVersion.Major, minKmsKeyVersionRotationVersion.Minor))}
+			}
+		} else if lowest.LT(minKmsKeyVersionRotationVersion) {
+			return field.ErrorList{field.Forbidden(fldPath, fmt.Sprintf("KMS key version rotation requires cluster version %s or above", minKmsKeyVersionRotationVersion))}
+		}
```
machi1990: See the proposal here https://github.com/Azure/ARO-HCP/pull/5629/#discussion_r3499039949 to only consider X.Y and drop the .Z part

internal/admission/admit_cluster.go
ahitacat: This is the case I had in mind to use the lowest. What if the 4.22.4 is not done yet because some operator is stuck. Does it mean the key rotation will work or not? I opted for the most restrictive version check.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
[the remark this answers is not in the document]
ahitacat: The change for older versions is from implicitly updateable to explicitly updateable, but there's no actual semantic change so it's backward-compatible.

internal/admission/admit_cluster.go
ahitacat: However, I see we are using the new controlPlaneVersion status for 4.22 clusters. So this should be calculated better and I don't expect to have that long history of version applied partially.

machi1990: > I see we are using the new controlPlaneVersion status for 4.22 clusters. So this should be calculated better and I don't expect to have that long history of version applied partially.

As soon as the cluster moves to 4.22 we should see those so we are good I guess; barring some really nasty bugs that prevents all the control plane from being rolled out: e.g the OVN cp vs data plane skew that prevents the CP side from being rolled out.

I guess using the restrictive version make sense. It'll fail with a clear error message rather than it being hidden from the user. The action item is then the user reports that I see X and Y failures while my cluster is in 4.22 and we tell them; it's because of this foo-bar stuck upgrade

internal/ocm/convert.go
```diff
@@ -397,6 +398,23 @@ func BuildCSCluster(resourceID *azcorearm.ResourceID, tenantID string, hcpCluste
 		))
 	}
 
+	// Etcd encryption is set outside withImmutableAttributes so that KMS
+	// key version updates reach CS on both create and update PATCHes.
+	if hcpCluster.CustomerProperties.Etcd.DataEncryption.KeyManagementMode != "" || hcpCluster.CustomerProperties.Etcd.DataEncryption.CustomerManaged != nil {
```
miguelsorianod: When can "hcpCluster.CustomerProperties.Etcd.DataEncryption.KeyManagementMode" be empty? are we not guaranteed to have a set value here?

miguelsorianod: If we check for keymanagementmode customermanaged, can we avoid checking nil-ness of hcpCluster.CustomerProperties.Etcd.DataEncryption.CustomerManaged as it should be mandatory?

internal/ocm/convert.go
```diff
@@ -397,6 +398,23 @@ func BuildCSCluster(resourceID *azcorearm.ResourceID, tenantID string, hcpCluste
...
+		etcdEncryption, err := convertEtcdRPToCS(hcpCluster.CustomerProperties.Etcd)
```
miguelsorianod: That means we are going to send the whole section during updates too. Does CS deal correctly with that?

ahitacat: Yes, it doesn't block that no updatable etcd field are nil. It just ignores them for when looking for changes in the `getUpdateChanges()`

internal/ocm/convert.go
ahitacat: > When can "hcpCluster.CustomerProperties.Etcd.DataEncryption.KeyManagementMode" be empty? are we not guaranteed to have a set value here?

You are right, this can be nil. 

> can we avoid checking nil-ness of hcpCluster.CustomerProperties.Etcd.DataEncryption.CustomerManaged as it should be mandatory?

This shouldn't be nil if the hcpCluster.CustomerProperties.Etcd.DataEncryption.KeyManagementMode is `CustomerManaged`. I can check for that instead. I wanted to avoid have problems in the future when we implement the PlatformManaged and the kms is not configured the same way.

miguelsorianod: > You are right, this can be nil.

Can it? my understanding is that we set it when converting to internal. Can we check?

> This shouldn't be nil if the hcpCluster.CustomerProperties.Etcd.DataEncryption.KeyManagementMode is CustomerManaged. I can check for that instead. I wanted to avoid have problems in the future when we implement the PlatformManaged and the kms is not configured the same way.

Because we currently don't support platform managed, don't we have a validation that enforces only customer managed is allowed as of now?

ahitacat: > Can it? 

No no, my bad, can't be nil. We have a validation that it is required.

> Because we currently don't support platform managed, don't we have a validation that enforces only customer managed is allowed as of now?

Yeah, but when we add platform managed we won't need to change a validation that is not for that type. So, I can just add the check that if the CustomerManaged is set, we can check if the keyVersion has change or not. If it is CustomerManaged this will not be nil.

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,37 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+		return nil
+	}
+
+	if newObj.Etcd.DataEncryption.KeyManagementMode != api.EtcdDataEncryptionKeyManagementModeTypeCustomerManaged {
+		return nil
+	}
+
+	if newObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version == oldObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version {
+		return nil
+	}
```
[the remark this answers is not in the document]
mbarnes: Agree with Copilot here.

internal/api/apiversions.go
```diff
@@ -0,0 +1,72 @@
...
+package api
+
+import "strings"
+
+const (
+	apiVersionOptionPrefix = "api-version="
+
+	APIVersionV20240610Preview = "2024-06-10-preview"
+	APIVersionV20251223Preview = "2025-12-23-preview"
+	APIVersionV20260630Preview = "2026-06-30-preview"
```
mbarnes: I like this solution a lot.  Do you think it would be worth defining a custom string type for these constants, just for better type safety?

```
type APIVersion string

const (
        APIVersionV20240610Preview APIVersion = "2024-06-10-preview"
        ...
)

func APIVersionOption(version APIVersion) string {
        ...
}

func apiVersionFromOptions(options []string) APIVersion {
        ...
}

...
```

Perhaps also have `BuildValidationOptions` take an `api.Version` interface instead of a string, just to reduce the type casting between plain string and custom string.

internal/validation/validate_common.go
```diff
@@ -47,6 +48,16 @@ func AFECsToValidationOptions(features []arm.Feature) []string {
 	return ret
 }
```
mbarnes, line 49: Suggest replacing calls to `AFECsToValidationOptions` with `BuildValidationOptions` everywhere in the frontend.  (Looks like external auth validation doesn't take an `operation.Operation` yet, but node pools do.)

internal/validation/validate_cluster.go
```diff
@@ -885,7 +880,9 @@ func validateKmsKey(ctx context.Context, op operation.Operation, fldPath *field.
 	errs = append(errs, MaxLen(ctx, op, fldPath.Child("vaultName"), &newObj.VaultName, nil, 255)...)
 
 	//Version   string `json:"version"`
-	errs = append(errs, immutableByCompare(ctx, op, fldPath.Child("version"), &newObj.Version, safe.Field(oldObj, toKmsKeyVersion))...)
+	if !api.APIVersionGE(op.Options, api.APIVersionV20260630Preview) {
```
mbarnes: `api.APIVersionLT` would read a little clearer than `!api.APIVersionGE`, no?

In any case, please add a comment here explaining the version check.
```
// The version field was made mutable in version 2026-06-30-preview.
```

mbarnes (dismissed):
Had a few suggestions, but I like how the API version options turned out.  Nice job.

internal/admission/admit_cluster.go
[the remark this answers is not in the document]
ahitacat: The logic here is a bit complex. I wanted to avoid check for nilness. Motivated by @miguelsorianod's comment https://github.com/Azure/ARO-HCP/pull/5629#discussion_r3504722463. 
Currently when creating a cluster it requires the CustomerManaged configuration, so this shouldn't be nil once we update the cluster.

In the future, when we introduce other types of DataEncryption. We will need to decide if it will be possible to change, in day2, from one type to another and how that will look.

I think the change to no check nilness is correct for now. And we will need to add more logic once other types are implemented.

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,34 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+
+	if admissionContext.ServiceProviderCluster == nil {
+		return field.ErrorList{field.InternalError(fldPath, errors.New("cannot validate KMS key version rotation"))}
+	}
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(admissionContext.ServiceProviderCluster.Status.ControlPlaneVersion.ActiveVersions)
+	if lowest != nil {
+		if lowest.Major == minKmsKeyVersionRotationVersion.Major && lowest.Minor < minKmsKeyVersionRotationVersion.Minor {
+			return field.ErrorList{field.Forbidden(fldPath, fmt.Sprintf("KMS key version rotation requires cluster version %s or above", minKmsKeyVersionRotationVersion))}
```
machi1990: I think this is more a badrequest non?

internal/admission/admit_cluster.go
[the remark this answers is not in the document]
mbarnes: Ok, if you're sure it's safe that's fine.

internal/validation/validate_common.go
```diff
@@ -47,6 +48,16 @@ func AFECsToValidationOptions(features []arm.Feature) []string {
...
 
+// BuildValidationOptions combines AFEC feature flags and the ARM API version
+// into a single options slice for operation.Operation.Options.
+func BuildValidationOptions(features []arm.Feature, apiVersion api.APIVersion) []string {
```
mbarnes, line 53: This is fine as is, but what I meant by my suggestion was to take an `api.Version` and do the `.String()` call here.  It would just make calling this a little prettier.

```
Options: validation.BuildValidationOptions(subscription.GetRegisteredFeatures(), versionedInterface)
```

You could even go further and take an `arm.Subscription` object instead of the feature slice.
```
Options: validation.BuildValidationOptions(subscription, versionedInterface)
```

It's just cosmetic niceties so I'll leave it up to you.

config/config.yaml
```diff
@@ -700,7 +699,7 @@ defaults:
     image:
       registry: arohcpsvcdev.azurecr.io
       repository: arohcpbackend
-      digest: sha256:1403f55f7616aac0d2c676f6669cce518f1e72d0e315358fb0d4d1e3b0804f6b # 55611cd (2026-05-29 22:32)
+      digest: sha256:1aecde2e0ab3c3d9df7c9e2bc20e354329f5699b86a1c0b88768dae9124863db # 55611cd (2026-05-29 22:32)
```
miguelsorianod: I think this probably doesn't belong here

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,34 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
 
 	return errs
 }
+
+var minKmsKeyVersionRotationVersion = semver.Version{Major: 4, Minor: 22}
```
miguelsorianod, line 470: document

internal/validation/validate_cluster.go
```diff
@@ -254,7 +254,6 @@ func validateClusterCustomerProperties(ctx context.Context, op operation.Operati
 	errs = append(errs, Maximum(ctx, op, fldPath.Child("nodeDrainTimeoutMinutes"), &newObj.NodeDrainTimeoutMinutes, safe.Field(oldObj, toNodeDrainTimeoutMinutes), 10080)...)
 
 	//Etcd                    EtcdProfile                 `json:"etcd,omitempty"`
-	errs = append(errs, immutableByReflect(ctx, op, fldPath.Child("etcd"), &newObj.Etcd, safe.Field(oldObj, toEtcd))...)
```
miguelsorianod, line 263: Is removing this needed because the immutable cascades?

internal/api/apiversions.go
```diff
@@ -0,0 +1,76 @@
...
+// Returns an empty string if no API version option is present.
+func apiVersionFromOptions(options []string) APIVersion {
+	for _, opt := range options {
+		if version, ok := strings.CutPrefix(opt, apiVersionOptionPrefix); ok {
+			return APIVersion(version)
+		}
+	}
+	return APIVersion("")
+}
+
```
miguelsorianod, line 63: Will the comparisons work correctly when we have preview and non preview apis? consider the number part might be the same

internal/validation/validate_common.go
```diff
@@ -47,6 +48,16 @@ func AFECsToValidationOptions(features []arm.Feature) []string {
...
+	options := AFECsToValidationOptions(features)
+	if apiVersion != "" {
```
miguelsorianod, line 57: Is it possible that we get an empty apiVersion?

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,34 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+		return nil
+	}
+
+	if admissionContext.ServiceProviderCluster == nil {
+		return field.ErrorList{field.InternalError(fldPath, errors.New("cannot validate KMS key version rotation"))}
+	}
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(admissionContext.ServiceProviderCluster.Status.ControlPlaneVersion.ActiveVersions)
+	if lowest != nil {
+		if lowest.Major == minKmsKeyVersionRotationVersion.Major && lowest.Minor < minKmsKeyVersionRotationVersion.Minor {
```
miguelsorianod: If lowest is lower than minKmsKeyVersionRotationVersion.Major this would not be triggered.

Shouldn't it be:

if lowest.Major < minKmsKeyVersionRotationVersion || (lowest.Major == minKmsKeyVersionRotationVersion && lowest.Minor < minKmsKeyVersionRotationVersion.Minor) ?

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,34 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+	if newObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version == oldObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version {
+		return nil
+	}
+
+	if admissionContext.ServiceProviderCluster == nil {
+		return field.ErrorList{field.InternalError(fldPath, errors.New("cannot validate KMS key version rotation"))}
+	}
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(admissionContext.ServiceProviderCluster.Status.ControlPlaneVersion.ActiveVersions)
+	if lowest != nil {
```
miguelsorianod: So if lowest is nil, this means no check would be performed, which means it would let the admission pass, no matter the value of the new version. Is that correct?

internal/api/apiversions.go
```diff
@@ -0,0 +1,76 @@
...
+func APIVersionEQ(options []string, version APIVersion) bool {
+	v := apiVersionFromOptions(options)
+	return v != "" && v == version
+}
+
+func APIVersionNE(options []string, version APIVersion) bool {
+	v := apiVersionFromOptions(options)
+	return v != "" && v != version
```
miguelsorianod: The design here seems to for one part trust that the input will be correct (apiVersionFromOptions) and for another part not (we check if it's empty string or not). Is there some reason for that?

internal/admission/admit_cluster.go
machi1990: I think this stems from the fact that we'll never support a major version below OCP v4.

internal/validation/validate_cluster.go
ahitacat: Yes, I needed to removed all the immutableByReflect for all the keyVersion parents.

internal/admission/admit_cluster.go
machi1990: Precisely this https://github.com/Azure/ARO-HCP/blob/main/internal/validation/validate_cluster.go#L373

internal/admission/admit_cluster_test.go
```diff
@@ -340,119 +340,148 @@ func TestAdmitCluster_Update(t *testing.T) {
...
 		expectErrors                 []utils.ExpectedError
 	}{
 		{
 			name:                         "empty desired version skips admission",
 			oldClusterVersionID:          "4.10",
-			clusterVersionID:             "",
 			serviceProviderClusterStatus: serviceProviderClusterStatusWithActiveControlPlaneVersion("4.22"),
 			nodePools:                    []*api.HCPOpenShiftClusterNodePool{makeTestNodePool("np1", "4.10.0")},
-			expectErrors:                 []utils.ExpectedError{},
+			mutateNew: func(c *api.HCPOpenShiftCluster) {
```
miguelsorianod: This scales better 👍 

nitpick: describe what occurs.

mutateCluster -> newClusterFromOld, and the parameter instead of c is "old" (or oldCopy ?), to make it from explicit. I am not strict with the name if you think a better description

describe that it doesn't modify the original cluster, rather old is a copy already and it applies modifications to the copy

internal/admission/admit_cluster.go
machi1990: If we want to solidify this further abstracted from the the minimum major we support; 
I'll suggest we do this


```suggestion
		if minKmsKeyVersionRotationVersion.GT(semver.Versions{Major: lowest.Major, Minor: lowest.Minor}) {
```
and error

internal/admission/admit_cluster.go
```diff
@@ -260,3 +261,34 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+var minKmsKeyVersionRotationVersion = semver.Version{Major: 4, Minor: 22}
+
+// admitClusterEtcdKmsKeyVersionChange rejects KMS key version changes when the
+// cluster's active control plane version is older than 4.22 version
+func admitClusterEtcdKmsKeyVersionChange(_ context.Context, admissionContext *ClusterAdmissionContext, op operation.Operation, fldPath *field.Path, newObj, oldObj *api.HCPOpenShiftClusterCustomerProperties) field.ErrorList {
+	if op.Type != operation.Update || oldObj == nil {
+		return nil
+	}
+
+	if newObj.Etcd.DataEncryption.KeyManagementMode != api.EtcdDataEncryptionKeyManagementModeTypeCustomerManaged {
```
miguelsorianod: I understand here we assume the key management mode doesn't change right? and then we let that be covereed in the validation. Is this understanding correct?

internal/api/apiversions.go
ahitacat: It should work for:
```
<date>-preview
vX
```
If we want to number it in a different (i.e. vXalphaY) we need to change this. And compare preview with something that semver can parse. This can done in a follow up PR, I think it will require a bit of discussion.

internal/validation/validate_common.go
ahitacat: It is highly improbable, unless this mechanism is used in the controllers where the api-version is not part of the context.

The operation is used in the controllers to re-use validations.
https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/upgradecontrollers/nodepool_version_controller.go#L205

Just added in case it is used in the controllers.

internal/admission/admit_cluster.go
ahitacat: Yes, but lowest can't be nil, once the Cluster is created the activeVersions are set. I will change this so it doesn't check for lowest being nil.

internal/api/apiversions.go
ahitacat: apiVersionFromOptions can return nil, if the option doesn't have that version. I expect at least that the version you are trying to compare is not empty.

internal/admission/admit_cluster.go
ahitacat: Yes

internal/api/apiversions.go
miguelsorianod: > It should work for:
> <date>-preview
> vX

are we sure of that? Go compares strings using lexicographic order. If I understand correctly, it will consider v20250610 as smaller than v20250610, which is not accurate.

> If we want to number it in a different (i.e. vXalphaY) we need to change this. And compare preview with something that semver can parse. This can done in a follow up PR, I think it will require a bit of discussion.

Even in that case, we should at least document the current behavior and limitations.

If we are not going to change anything for now, we should at least document the behavior and the current limitations. However, based on my comment above it seems we don't have correct behavior for preview vs non-preview

miguelsorianod: According to https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/uniform-versioning.md#api-versions, it doesn't seem like we will have two versions with same date one with stable and the other without it, so it seems unlikely.
I would still update our logic to handle it for semantic consistency if it's not a big effort.

And as mentioned above let's document behavior and limitations

internal/api/apiversions.go
miguelsorianod: Why not trusting that it's not going to be empty?

internal/admission/admit_cluster.go
miguelsorianod: :+1:

internal/validation/validate_common.go
```diff
@@ -47,6 +48,18 @@ func AFECsToValidationOptions(features []arm.Feature) []string {
 	return ret
 }
 
+// BuildValidationOptions combines AFEC feature flags and the ARM API version
+// into a single options slice for operation.Operation.Options.
+func BuildValidationOptions(features []arm.Feature, apiVersion api.APIVersion) []string {
+	options := AFECsToValidationOptions(features)
+	//apiVersion can be empty if this is call from within the backend
+	// as backend doesn't have this context
```
miguelsorianod, line 56: Here there seems to be a mismatch when reusing the validate in backend:
We have code that calls validations in:
* Frontend, where there's information that is specific to the API request itself
* Backend, where the information is not specific to the API request itself

That reuse with that mismatch causes issues like this one. It would be good to raise this to the wider group to see if we want to tackle it differently.

miguelsorianod:
/lgtm

I don't approve yet because it depends on https://github.com/Azure/ARO-HCP/pull/5937 being merged, as well as CS changes being fully rolled out

ahitacat:
@miguelsorianod rebased with the https://github.com/Azure/ARO-HCP/pull/5937 merged in main. We are still waiting on CS to be fully rolled out.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,219 @@
...
+			Expect(fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged).NotTo(BeNil(), "fetched cluster CustomerManaged was nil")
+			Expect(fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged.Kms).NotTo(BeNil(), "fetched cluster Kms was nil")
+			Expect(fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey).NotTo(BeNil(), "fetched cluster ActiveKey was nil")
+			Expect(fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version).NotTo(BeNil(), "fetched cluster key Version was nil")
+			Expect(*fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version).To(Equal(newKeyVersion),
+				"cluster should reference the new KMS key version after round-trip GET")
+
+			By("verifying StorageVersionMigration succeeded for re-encryption")
+			err = verifiers.VerifyStorageVersionMigrationSucceeded().Verify(ctx, adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "all StorageVersionMigration resources should reach Succeeded state after KMS key rotation")
```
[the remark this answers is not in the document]
ahitacat: This is ok, by this time the migration is completed and it doesn't need to wait.

internal/admission/admit_cluster.go
```diff
@@ -271,3 +272,42 @@ func admitClusterVersionProfile(ctx context.Context, admissionContext *ClusterAd
...
+		return field.ErrorList{field.InternalError(fldPath, errors.New("cannot validate KMS key version rotation"))}
+	}
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(admissionContext.ServiceProviderCluster.Status.ControlPlaneVersion.ActiveVersions)
+	clusterVersion := semver.Version{Major: lowest.Major, Minor: lowest.Minor}
+	if clusterVersion.LT(minKmsKeyVersionRotationVersion) {
+		return field.ErrorList{field.Invalid(fldPath, newObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version, fmt.Sprintf("KMS key version rotation requires cluster version %s or above", minKmsKeyVersionRotationVersion))}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
ahitacat: This cannot happen during an update

test/e2e-setup/bicep/modules/customer-infra.bicep
```diff
@@ -115,6 +128,24 @@ resource etcdEncryptionKey 'Microsoft.KeyVault/vaults/keys@2024-12-01-preview' =
...
+resource testRunnerKeyVaultCryptoOfficerRoleAssignment 'Microsoft.Authorization/roleAssignments@2022-04-01' = if (assignKeyVaultCryptoOfficer && testRunnerPrincipalId != '') {
+  name: guid(resourceGroup().id, testRunnerPrincipalId, keyVaultCryptoOfficerRoleId, customerKeyVault.id)
+  scope: customerKeyVault
+  properties: {
+    principalId: testRunnerPrincipalId
+    principalType: testRunnerPrincipalType
+    roleDefinitionId: keyVaultCryptoOfficerRoleId
+  }
+}
+
```
mbukatov, line 138: This looks good. I have one request though: could we rename "testRunner" to "customer"? This will make it aligned with the rest of the naming scheme used in this file, and make it clear that it's a customer's account, not some special privileged CI account, which we are referencing here.

test/e2e-setup/bicep/modules/customer-infra.bicep
```diff
@@ -19,6 +19,19 @@ param clusterName string = ''
 @description('If set to true, creates a private KeyVault with publicNetworkAccess disabled')
 param privateKeyVault bool = false
 
+@description('Principal ID of the test runner for role assignments on test resources.')
+param testRunnerPrincipalId string = ''
```
mbukatov: It's worth considering whether it wouldn't be easier to use [`deployer`](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-functions-deployment) bicep function to get this information directly without passing it to bicep. Then we don't need the code to fetch this information in the test code.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,225 @@
...
+				*resourceGroup.Name,
+				clusterName,
+				framework.GetAdminRESTConfigTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to get admin REST config for cluster")
+
+			By("ensuring the cluster is viable")
+			err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "failed to verify HCP cluster viability for patch-tags test")
+
```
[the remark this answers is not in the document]
mbukatov: Here the copilot has a good point, please have it fixed.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,225 @@
...
+			clusterParams.ManagedResourceGroupName = managedResourceGroupName
+
+			By("getting test runner principal ID for Key Vault role assignment")
+			identityDetails, err := tc.GetCurrentAzureIdentityDetails(ctx)
+			Expect(err).NotTo(HaveOccurred(), "failed to get current Azure identity details")
+
+			principalType := "ServicePrincipal"
+			if identityDetails.PrincipalType == framework.PrincipalTypeDSTSUser {
+				principalType = "User"
+			}
```
mbukatov: As I note in the comment for the bicep file, if we fetch this in the bicep file directly, we don't need to handle it in the test case code.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,225 @@
...
+			updateKeyResp, err := keyClient.UpdateKey(ctx, clusterParams.EtcdEncryptionKeyName, clusterParams.EtcdEncryptionKeyVersion, keyParams, nil)
+			Expect(err).NotTo(HaveOccurred(), "failed to disable old KMS key version: %v", clusterParams.EtcdEncryptionKeyVersion)
+			Expect(updateKeyResp.Attributes.Enabled).ToNot(BeNil(), "update key response did not include enabled attribute")
+			Expect(*updateKeyResp.Attributes.Enabled).To(BeFalse(), "old key version should be disabled")
+
+			GinkgoLogr.Info("Old key version was disabled successfully",
+				"keyVersion", clusterParams.EtcdEncryptionKeyVersion,
+				"enabled", updateKeyResp.Attributes.Enabled,
+			)
```
mbukatov, line 213: We should run a full cluster verifier again, including VerifyStorageVersionMigrationSucceeded, something like:

```
err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyStorageVersionMigrationSucceeded())
```

To make sure the cluster is still fully functional after the original key was disabled.

mbukatov (commented):
I checked the E2E test only and provide few suggestions.

test/e2e-setup/bicep/modules/customer-infra.bicep
ahitacat: This is no longer required as it has change to use the deployer()

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
```diff
@@ -80,6 +80,11 @@ func (c *operationClusterUpdate) hypershiftHostedClusterSpecMatchesDesired(clust
 	if matches, message := c.hypershiftHostedClusterImageContentSourcesSpecMatchesDesired(cluster.CustomerProperties.ImageDigestMirrors, hostedCluster.Spec.ImageContentSources); !matches {
 		return false, message
 	}
+	if cluster.CustomerProperties.Etcd.DataEncryption.KeyManagementMode == api.EtcdDataEncryptionKeyManagementModeTypeCustomerManaged {
```
ahitacat: Added this check so it will not conflict with platform managed implementation in the future

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
```diff
@@ -334,6 +339,35 @@ func (c *operationClusterUpdate) hypershiftHostedClusterImageContentSourcesSpecM
 	return true, ""
 }
 
+func (c *operationClusterUpdate) hypershiftHostedClusterEtcdDataEncryptionCustomerManagedSpecMatchesDesired(desired api.EtcdDataEncryptionProfile, observed *v1beta1.SecretEncryptionSpec) (bool, string) {
+
+	if observed == nil {
+		return false, "unexpected hypershift HostedCluster secret encryption is not set"
+	}
+
+	if desired.CustomerManaged.EncryptionType != api.CustomerManagedEncryptionTypeKMS {
```
ahitacat, line 358: If line 83 is OK I can simply remove this one

mbukatov (commented):
the e2e test /lgtm

ahitacat:
/hold cancel
CS has been rolled out
latest HO is in Int, we can safely merge this

miguelsorianod:
Regarding the non e2e changes, they lgtm with the exception of what we discussed about some refreshing of the openapi related changes

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,321 @@
...
+			By("verify the Cluster is still fully functional")
+			err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyStorageVersionMigrationSucceeded())
+			Expect(err).NotTo(HaveOccurred(), "all StorageVersionMigration resources should reach Succeeded state after KMS key rotation")
+
+			By("rotating the KMS key a second time")
+
+			GinkgoLogr.Info("Creating new key version (second rotation)",
+				"keyVaultName", clusterParams.KeyVaultName,
+				"keyName", clusterParams.EtcdEncryptionKeyName,
+				"keyVersion", newKeyVersion)
```
mbukatov: Could we update the "keyVersion" string to something like "currentKeyVersion"? I could be easily confused to assume that this is the new version of the key. In the previous rotation, we avoided this confusion by using "originalKeyVersion" name.

test/e2e-setup/bicep/modules/customer-infra.bicep
```diff
@@ -115,6 +118,24 @@ resource etcdEncryptionKey 'Microsoft.KeyVault/vaults/keys@2024-12-01-preview' =
   }
 }
 
+// Key Vault Crypto Officer: allows key management operations (create, rotate, disable).
+// Assigned to the test runner principal so E2E tests can rotate etcd encryption keys.
+// https://www.azadvertizer.net/azrolesadvertizer/14b46e9e-c2b7-41b4-b07b-48a6ebf60603.html
```
miguelsorianod: nitpick: better to refer to the official documentation: https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles/security#key-vault-crypto-officer

test/e2e-setup/bicep/modules/customer-infra.bicep
```diff
@@ -19,6 +19,9 @@ param clusterName string = ''
 @description('If set to true, creates a private KeyVault with publicNetworkAccess disabled')
 param privateKeyVault bool = false
 
+@description('Assign Key Vault Crypto Officer to the deployer on the Key Vault.')
```
miguelsorianod: "on the key vault". What keyvault is it? it would be good to clarify this at both doc and naming level

mbukatov (commented):
minor comment about logging, otherwise extension of E2E test looks good

test/e2e-setup/bicep/modules/customer-infra.bicep
ahitacat: There is a mix usage of azadvertizer and the official documentation. I will move to the official one but we should consider change the other ones.

miguelsorianod:
/hold

so I can add tags without triggering merge. Unhold when you think it's ready to merge and when @mbukatov has given his ok on the tests side.

miguelsorianod:
> /approve

I do not have permissions because it touches api files too

mbarnes:
For the API changes:
/lgtm

mbarnes:
/approve

But there's still a **do-not-merge** tag on this.  I don't know if it's ok to remove.

PR #5632, opened 2026-06-12
api: scaffold new API version 2026-06-30-preview

bennerv:
## Summary
- Adds the `2026-06-30-preview` API version as a copy of `2025-12-23-preview` with no feature changes
- Establishes the version entry point (TypeSpec, OpenAPI spec, autorest config, Go models, conversion methods, frontend registration, integration test expected files) so new features can be added incrementally in follow-up PRs
- No conditions (`Condition[]`) are included — that will land separately

## Test plan
- [x] `cd api && make generate` — TypeSpec compiles, OpenAPI spec + Go models generate cleanly
- [x] `go test ./internal/api/v20260630preview/...` — all conversion, fuzz, roundtrip, and patch validation tests pass
- [x] `go test ./internal/api/...` — all API version tests pass (v20240610preview, v20251223preview, v20260630preview)
- [x] `go test ./frontend/...` — frontend tests pass
- [x] `make lint` — 0 issues
- [x] Verified new OpenAPI spec contains no `conditions` (plural) field — only the pre-existing `ExternalAuthCondition` (singular)


internal/api/v20260630preview/zero_value_roundtrip_test.go
```diff
@@ -0,0 +1,347 @@
...
+// round-trip safety for cluster-level fields. All tests pass after the
+// PtrOrNil → Ptr fix.
+func TestClusterZeroValueRoundTripThroughJSON(t *testing.T) {
+	tests := []struct {
+		name  string
+		setup func() *api.HCPOpenShiftCluster
+		check func(t *testing.T, result *api.HCPOpenShiftCluster)
+	}{
+		{
+			name: "HostPrefix zero must survive round-trip",
```
bennerv, line 125: nit: these should be new api fields introduced by the API version, not existing, but keeping them as this is a noop introduction of a new API version.

geoberle:
@SudoBrendan wrote a very comprehensive document about what kinds of tests we need to introduce when we add a new API version that comes with their own defaults: docs/api-version-defaults-and-storage.md

we can prepare for the laters PRs that will introduce new fields by adding conversion consistency tests to convert_defaults_consistency_test.go, or you add that as part of the followup PR, your choice.

bennerv:
> @SudoBrendan wrote a very comprehensive document about what kinds of tests we need to introduce when we add a new API version that comes with their own defaults: docs/api-version-defaults-and-storage.md
> 
> we can prepare for the laters PRs that will introduce new fields by adding conversion consistency tests to convert_defaults_consistency_test.go, or you add that as part of the followup PR, your choice.

I missed a few files that had 2024/2025 api references but not the new one.  I added those. 

I added `convert_defaults_consistency_test.go` but didn't update the `cross_version_roundtrip_test.go` to handle any newer api version specifics.  I'm tempted to move away from this and instead use the existing `frontendCRUD` integration tests with json to handle this.  

I think it's in a good state now, a skill to generate new API versions to follow after I test it not included in this PR.

test-integration/frontend/cross_version_roundtrip_test.go
twolff-gh, line 94: Do we want to add a 2024 to 2026 case at all?
2025 and 2026 are basically the same now right?

Do we care about only forward, or backward too?

bennerv: https://github.com/Azure/ARO-HCP/pull/5632#issuecomment-4711227323

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/main.tsp
```diff
@@ -25,4 +25,8 @@ enum Versions {
   /** 2025-12-23-preview version */
   @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v6)
   v2025_12_23_preview: "2025-12-23-preview",
+
+  /** 2026-06-30-preview version */
+  @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v6)
+  v2026_06_30_preview: "2026-06-30-preview",
```
mbarnes, line 31: Note if we're writing up a skill for Claude:
New API versions should use the latest `CommonTypes` version copied over from [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/common-types/resource-management).  That's still v6 as of now, but it's something to have Claude check.

internal/database/convert_defaults_consistency_test.go
```diff
@@ -171,6 +187,31 @@ func TestEnsureDefaultsConsistencyCluster(t *testing.T) {
...
+				} else if c.canonicalVal != *c.externalPtrVal {
+					t.Errorf("ensured default = %q, versioned default = %q", c.canonicalVal, *c.externalPtrVal)
+				}
+			})
+		}
+	})
+	t.Run("v20260630preview", func(t *testing.T) {
+		externalDefault := &v20260630preview.HcpOpenShiftCluster{}
+		v20260630preview.SetDefaultValuesCluster(externalDefault)
+
```
mbarnes, line 214: Non-blocking suggestion, but it looks like we could split the `t.Run` logic out so it doesn't have to be repeated for every new version.

mbarnes (approved):
/lgtm

A little concerned about all the repetition in the E2E framework, especially if we're going to be adding new API versions frequently.  But I guess I'll let QE decide if it needs refactoring.

PR #5634, opened 2026-06-12
fix: add serviceProviderCluster fixtures and fix test step ordering

cgiradkar:
## Summary

This PR fixes integration test failures caused by missing test data required for nodepool version validation. The core issue was that nodepool creation tests lacked serviceProviderCluster fixtures containing cluster version information, causing validation to attempt dereferencing nil pointers and crashing the test server.

## Context: Why These Changes Are Needed

When nodepool version validation was introduced (to prevent version skew between nodepools and control plane), the validation logic began requiring serviceProviderCluster data to check active cluster versions. Many existing integration tests predated this requirement and never had this fixture data, causing them to fail with cryptic EOF errors when the validation code panicked.

## Problem Investigation Summary

The EOF errors were masking the real issue - the test HTTP server was panicking and closing connections. The panic occurred at `validators.go:960` when dereferencing a nil `*semver.Version` pointer. This pointer was nil because:

1. Tests lacked `serviceProviderCluster` fixtures entirely, OR
2. Tests had the fixture but loaded it AFTER attempting nodepool creation (due to alphabetical step ordering)

### 1. Added Missing Test Fixtures

**What to review:** New `03-loadCosmos-serviceProviderCluster/serviceProviderCluster.json` files across multiple test scenarios.

**Why this approach:** Each test scenario needs cluster version data that matches its test cluster. The fixture structure mirrors what the backend creates when clusters are created via HTTP API - it contains:
- `active_versions` array with semantic version strings (e.g., "4.20.0")
- `desired_version` in the spec

### 2. Fixed Test Step Execution Order

**What to review:** Directory renames from `03-httpCreate-*` to `04-httpCreate-*` (or `05-` in some cases).

**Why this matters:** The test framework processes steps in **lexicographical order**. When both `03-httpCreate-nodepool` and `03-loadCosmos-serviceProviderCluster` existed, the former executed first because 'h' < 'l' in ASCII. This meant nodepool validation ran before the required cluster version data was loaded.

**Solution rationale:** Rather than changing the test framework's ordering logic (which would be risky), I renumbered directories to enforce correct execution order: load serviceProviderCluster at step 03, then create nodepool at step 04+.

**Tests with multiple nodepools:** Some tests (like `create-current`) have multiple `04-httpCreate-*` steps. This is intentional and safe - as long as they all run AFTER the `03-loadCosmos` step, the alphabetical ordering within step 04 doesn't matter.

### 3. Programmatic Fixture Creation in `version_compliance_test.go`

**What to review:** New `createServiceProviderClusterForVersionCompliance` helper function.

**Why this pattern:** The `version_compliance_test.go` creates clusters programmatically rather than from fixtures. To maintain consistency with its testing approach, it also creates serviceProviderCluster documents programmatically rather than loading from fixture files.

**Alternative considered:** I could have made this test load from fixture files like FrontendCRUD tests do, but that would break the test's design pattern of being fully self-contained and not relying on external JSON files.

## Testing & Verification

**Manual verification performed:**
1. Ran full `make test-integration` - all tests pass
2. Specifically verified previously failing tests now pass: `missing-required-fields`, `invalid-version`, `create-current`, `delete-not-last` (these were failing for the modified code for skew validation PR #4566 )

## Rollout Considerations

**No production impact:** These are test-only changes. No business logic modified, no API behavior changed.

## Review Focus Areas

1. **Step numbering consistency:** Confirm all nodepool creation steps now occur after serviceProviderCluster load
2. **No behavioral changes:** Confirm no changes to validation logic itself - only test infrastructure


test-integration/frontend/artifacts/FrontendCRUD/NodePool/cluster-creating/03-loadCosmos-serviceProviderCluster/serviceProviderCluster.json
```diff
@@ -0,0 +1,23 @@
+{
```
miguelsorianod, line 1: Do we really need to add this? is it something that can't be covered by unit tests?

cgiradkar: by this, do you mean all the service provider cluster fixtures? 
There is indeed a way of simulating SPC load in cosmos. If we do that, then all the SPC fixtrures can be safely deleted and this PR would no longer be needed. Is this what we want or do we want to keep fixtures?

ahitacat: It is covered in the unit test, but when we add the need of the serviceProviderCluster to create a node pool this integration test will also need the serviceProviderCluster. This validation will be added in https://github.com/Azure/ARO-HCP/pull/4566

miguelsorianod: There are ongoing discussions on changing the design of the integration tests to not need to do this. Let's wait until a decision has been taken there

miguelsorianod: After discussing about it, we want to redesign the integration tests but we don't want to block on changes that are time sensitive.

>It is covered in the unit test, but when we add the need of the serviceProviderCluster to create a node pool this integration test will also need the serviceProviderCluster. This validation will be added in #4566 

@ahitacat can you expand on this? I see this validation you mention is in main already, not particular to that PR, so not sure what it is changing here.

If this is something that is needed because of changes on 4566, why are those not introduced there? This PR mentions this fixes an issue but the integration tests in main are not failing and there doesn't seem to be flakiness on them either

cgiradkar: >  why are those not introduced there

Related changes are added in that PR. But since other unrelated tests also needed the SPC(service provider cluster) fixture and the number of files to be changed was getting quite large so I opened this PR for keeping it clean.

test-integration/frontend/artifacts/FrontendCRUD/NodePool/invalid-name-too-long-name/03-loadCosmos-serviceProviderCluster/serviceProviderCluster.json
```diff
@@ -0,0 +1,23 @@
+{
+	"id": "6b690bec-0c16-4ecb-8f67-781caf40bba7",
```
miguelsorianod: id and partitionkey are the same. They shouldn't. Review this and set appropriate values. Check for all resources being added in the PR

test-integration/frontend/version_compliance_test.go
```diff
@@ -246,3 +248,47 @@ func prettyJSON(t *testing.T, v any) string {
 	}
 	return strings.TrimSpace(string(b))
 }
+
+func createServiceProviderClusterForVersionCompliance(
```
miguelsorianod: Do we know we need to add this because of another MR? If not, don't add it. Only when needed it

test-integration/frontend/cross_version_roundtrip_test.go
```diff
@@ -285,6 +285,58 @@ func createClusterAndComplete(
 
 	parsedID := api.Must(azcorearm.ParseResourceID(resourceID))
 	require.NoError(t, integrationutils.MarkOperationsCompleteForName(ctx, testInfo.ResourcesDBClient(), subscriptionID, parsedID.Name))
+
+	createServiceProviderClusterForTesting(t, ctx, testInfo, clusterName)
+}
+
+// createServiceProviderClusterForTesting inserts a serviceProviderCluster document
```
miguelsorianod, line 363: https://github.com/Azure/ARO-HCP/pull/5634/changes#r3511444455

test-integration/frontend/version_compliance_test.go
cgiradkar: yes we need it

test-integration/frontend/artifacts/FrontendCRUD/NodePool/invalid-name-too-long-name/03-loadCosmos-serviceProviderCluster/serviceProviderCluster.json
cgiradkar: updated

test-integration/frontend/version_compliance_test.go
miguelsorianod: Can you point me why and for what specifically?

test-integration/frontend/cross_version_roundtrip_test.go
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/5634#discussion_r3545299876

test-integration/frontend/cross_version_roundtrip_test.go
```diff
@@ -371,20 +370,18 @@ func createServiceProviderClusterForTesting(
 ) {
 	t.Helper()
 
-	cosmosTestInfo, ok := testInfo.StorageIntegrationTestInfo.(*integrationutils.CosmosIntegrationTestInfo)
```
miguelsorianod: What is this change?

test-integration/frontend/version_compliance_test.go
cgiradkar: Clarification: This is needed by PR #4566 when SPC will be required to be loaded in Cosmos not just simulate the step while testing. Technically this helper function will be needed then but since we're separating that part in this PR for cleanliness

test-integration/frontend/cross_version_roundtrip_test.go
cgiradkar: The helper returns early on the mock path and never loads the SPC. Hence, removing this function call in favor of `arm.ResourceIDStringToCosmosID`

PR #5637, opened 2026-06-12
fix(frontend): return 409 on DELETE against Failed resources to prevent retry storms

swiencki:
[ARO-27726](https://redhat.atlassian.net/browse/ARO-27726)

### What

Return 409 Conflict on DELETE against resources in Failed provisioning state, and skip Failed clusters in the subscription sweep.

### Why

Without this, ARM and CAPZ controllers retry DELETE against stuck-Failed clusters indefinitely, generating 14k+ wasted operations across 5 clusters in 14 days.

### Testing

Updated `TestCheckForProvisioningStateConflict` to expect 409 for Failed state on delete operations (cluster and node pool).

### DIFF
--- a/frontend/pkg/frontend/helpers.go
+++ b/frontend/pkg/frontend/helpers.go
@@ -64,7 +64,8 @@ func checkForProvisioningStateConflict(
 	case database.OperationRequestCreate:
 		// Resource must already exist for there to be a conflict.
 	case database.OperationRequestDelete:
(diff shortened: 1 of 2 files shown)


swiencki:
Superseded by [CS !358](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/358) which fixes the root cause in Cluster Service's pending delete worker rather than blocking deletes at the RP frontend.

PR #5639, opened 2026-06-12
feat: add SDK pipeline policy to retry version-not-found on cluster create/update

bennerv:
## Summary

- Adds a reusable `retryVersionNotFoundPolicy` Azure SDK pipeline policy that retries transient `InvalidRequestContent` errors when an OpenShift version is not yet registered in Cluster Service (Cincinnati advertises z-stream builds before CS registers them)
- Policy scopes to PUT/PATCH requests on `Microsoft.RedHatOpenShift/hcpOpenShiftClusters` only, excluding child resources (node pools, external auth)
- Removes the inline retry loop from the z-stream upgrade E2E test since the policy now handles it transparently

Slack thread: https://redhat-external.slack.com/archives/C075PHEFZKQ/p1781300876636439

## Test plan

- [x] Unit tests pass: 12 new tests covering pass-through for wrong methods/resource types, retry+succeed on PUT and PATCH, retry exhaustion, non-matching errors, context cancellation, and backoff bounds
- [x] All existing framework tests still pass
- [x] Lint clean (`make lint` — 0 issues)
- [ ] E2E z-stream upgrade test passes in CI with the policy handling retries instead of the inline loop

### DIFF
--- a/test/e2e/control_plane_automated_z_stream_upgrade.go
+++ b/test/e2e/control_plane_automated_z_stream_upgrade.go
@@ -16,7 +16,6 @@ package e2e
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


raelga (approved):
/lgtm
/approve

PR #5643, opened 2026-06-14
api: add ingress type to ARM API and wire to Cluster Service

zgalor:
[ARO-27757](https://redhat.atlassian.net/browse/ARO-27757)

## Summary
- Adds `IngressProfile` with `type` field to the `v2026-06-30-preview` ARM API
- `IngressType` is a new union: `Public | Private | Disabled | string` (open enum)
  - `Public`: ingress visible from internet (maps to CS `listening=external`)
  - `Private`: ingress not visible from internet (maps to CS `listening=internal`)
  - `Disabled`: defined in API but rejected in frontend validation (future feature)
  - `string`: open enum for future extensibility
- Wires ingress type through to Cluster Service `ingresses[0].listening` during cluster creation
- Field is immutable after creation and defaults to Public

## Changes
- **TypeSpec**: Added `IngressProfile` model and `IngressType` union with `@added(Versions.v2026_06_30_preview)`. `IngressType` is separate from `Visibility` (used by `ApiProfile`)
- **Generated code**: OpenAPI spec for v2026-06-30-preview, Go models, serde (auto-generated)
- **Internal types**: Added `CustomerIngressProfile` struct with `IngressType`, defaults in `NewDefaultHCPOpenShiftCluster` and `EnsureDefaults`
- **Versioned API (v20260630preview)**: Added `newIngressProfile`/`normalizeIngress`, wired into `ConvertToExternal`/`ConvertToInternal`/`SetDefaults`
- **Cross-version preservation**: Added ingress to `preserveUnknownClusterFields` in both v20240610preview and v20251223preview
- **Validation**: Added `validateCustomerIngressProfile` with required, immutable, and enum checks. `Disabled` is explicitly rejected
- **CS conversion**: New `convertIngressTypeToListening` function, builds `Ingresses(NewIngressList().Items(NewIngress().Default(true).Listening(...)))` in `BuildCSCluster`
- **Tests**: Private ingress conversion test, ingress immutability test, Disabled rejection test, defaults consistency tests (canonical↔constructor↔versioned), pre-existing data test, fuzz roundtrip exclusions, cross-version roundtrip tests (v2026→v2024 and v2026→v2025 PUT/PATCH)

## Depends on
- CS MR !353 ([ARO-26911](https://redhat.atlassian.net/browse/ARO-26911)) — e2e-parallel will fail until CS accepts `ingresses` field

## Test plan
- [x] `make -C api generate` — TypeSpec compiles, OpenAPI + Go models generate
- [x] `make -C api validate-examples` — example validation passes (all versions)
- [x] `go test ./internal/...` — all internal tests pass
- [x] `go test ./frontend/...` — frontend tests pass
- [x] `go test ./backend/...` — backend tests pass
- [x] `make verify-deepcopy verify-json-format` — verify checks pass


zgalor:
**Note:** The `e2e-parallel` failure is expected — this PR sends the `ingresses` field to Cluster Service on cluster creation, but CS doesn't accept it yet. This depends on [CS MR !353](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/353) ([ARO-26911](https://redhat.atlassian.net/browse/ARO-26911)) being merged and deployed to the test environment.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/examples/2025-12-23-preview/HcpOpenShiftClusters_CreateOrUpdate_MaximumSet_Gen.json
```diff
@@ -122,6 +125,9 @@
             "visibility": "Public",
             "url": "https://api.test.shrd.usw3test.hcp.osadev.cloud:443"
           },
+          "ingress": {
```
miguelsorianod: Does this control the **default ingress** visibility, or the visibility of all ingresses created over time? Depending on the answer, the naming can be confusing.

Are we always going to offer a **default ingress**, or is it possible or being discussed that that will be on end-users responsibility?

zgalor: This controls the default ingress visibility - the single default OpenShift ingress that's created with the cluster. It maps to CS's `ingresses[0].default=true` with the corresponding listening method (external/internal).                                                                                        

ARO HCP clusters have a single ingress configuration exposed through the ARM API. The field is immutable after creation. The naming follows the existing ApiProfile pattern (singular) since it represents a single cluster-level configuration, not a collection.

zgalor:
@bennerv @mbarnes PTAL

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2025-12-23-preview/openapi.json
```diff
@@ -2256,6 +2256,14 @@
             "create"
           ]
         },
+        "ingress": {
+          "$ref": "#/definitions/IngressProfile",
+          "description": "Shows the cluster ingress profile",
```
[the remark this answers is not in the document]
mbarnes: Agree with the suggestion.

internal/api/v20251223preview/hcpopenshiftclusters_methods.go
```diff
@@ -73,6 +73,12 @@ func SetDefaultValuesCluster(obj *HcpOpenShiftCluster) {
 	if obj.Properties.API.Visibility == nil {
 		obj.Properties.API.Visibility = ptr.To(generated.VisibilityPublic)
 	}
+	if obj.Properties.Ingress == nil {
+		obj.Properties.Ingress = &generated.IngressProfile{}
+	}
+	if obj.Properties.Ingress.Visibility == nil {
+		obj.Properties.Ingress.Visibility = ptr.To(generated.VisibilityPublic)
+	}
```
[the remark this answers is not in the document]
mbarnes: It's consistent with `APIProfile`.  Disregard.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/examples/2025-12-23-preview/HcpOpenShiftClusters_CreateOrUpdate_MaximumSet_Gen.json
miguelsorianod: Should we call it defaultIngress then?

internal/api/v20251223preview/hcpopenshiftclusters_methods.go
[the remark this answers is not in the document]
mbarnes: Literally right above the lines being commented on.
```
	if obj.Properties.API == nil {
		obj.Properties.API = &generated.APIProfile{}
	}
```

internal/api/v20240610preview/conversion_fuzz_test.go
```diff
@@ -56,6 +56,9 @@ func TestRoundTripInternalExternalInternal(t *testing.T) {
 			// It cannot roundtrip through this version's external type.
 			// Cross-version preservation is handled by preserveUnknownClusterFields.
 			j.ImageDigestMirrors = nil
+			// Ingress was added in v2025_12_23_preview and does not exist in v20240610preview.
+			// Cross-version preservation is handled by preserveUnknownClusterFields.
+			j.Ingress = api.CustomerIngressProfile{}
```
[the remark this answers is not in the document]
mbarnes: > but this hook is for *api.HCPOpenShiftClusterStatus

wut?  Copilot is hallucinating here.

internal/api/v20251223preview/hcpopenshiftclusters_methods.go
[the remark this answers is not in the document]
mbarnes: I'm less familiar with patching behavior than I used to be, but this is following the same pattern as `APIProfile`.  I don't see a problem.

internal/ocm/convert.go
```diff
@@ -375,6 +375,14 @@ func BuildCSCluster(resourceID *azcorearm.ResourceID, tenantID string, hcpCluste
...
 		}
 		clusterAPIBuilder.Listening(apiListening)
+
+		ingressListening, err := convertVisibilityToListening(hcpCluster.CustomerProperties.Ingress.Visibility)
+		if err != nil {
+			return nil, nil, err
+		}
+		clusterBuilder.Ingresses(arohcpv1alpha1.NewIngressList().Items(
+			arohcpv1alpha1.NewIngress().Default(true).Listening(ingressListening),
+		))
```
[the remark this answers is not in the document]
mbarnes: It's fine.  Builder methods return the builder instance.

mbarnes (changes requested):
Two issues:
1.  This should be added to the `2026-06-30-preview` version, not `2025-12-23-preview`, although I realize that's not a fair ask at this moment since we haven't yet merged the [scaffolding for the new version](https://github.com/Azure/ARO-HCP/pull/5632) yet.
2. Either way, we'll need some new test cases for the new "public" ingress default in `/internal/database/convert_defaults_consistency_test.go` and I think a couple of the integration tests.  There's a DDR in-repo about default handling across API versions, with a [detailed checklist for new fields](https://github.com/Azure/ARO-HCP/blob/main/docs/api-version-defaults-and-storage.md#adding-a-new-field).

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/examples/2025-12-23-preview/HcpOpenShiftClusters_CreateOrUpdate_MaximumSet_Gen.json
zgalor: I don't think we should. I think ARO HCP doesn't have the concept of default ingress.
@bennerv @mbarnes keep me honest here

miguelsorianod: Isn't default ingress an OpenShift concept (https://github.com/openshift/cluster-ingress-operator#installing) ?

zgalor:
@mbarnes, I addressed your comments.

Moved the ned field to `2026-06-30-preview` version and added integration tests according to the DDR you shared. Note that some e2e might fail until the updated CS is rolled out.

Please take another look

mbarnes:
For the **ci/prow/api-validation** error, it looks like you might just need to run `make -C api generate` again.

Overall LGTM.  I'll give formal approval once api-validation is fixed.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/examples/2025-12-23-preview/HcpOpenShiftClusters_CreateOrUpdate_MaximumSet_Gen.json
miguelsorianod: @bennerv @mbarnes @zgalor this comment has not been addressed

zgalor: @miguelsorianod I don't have an answer. Not sure if it's not for the BU to decide

zgalor: But even in older PRs, it was always called ingress

https://github.com/Azure/ARO-HCP/pull/2894

https://github.com/Azure/ARO-HCP/pull/2281

davidffrench: I believe we should be consistent with HyperShift on this. What is the field called in the HostedCluster CR? Also what is it called in ARO Classic and ROSA HCP?

zgalor: There is no corresponding field in the HostedCluster CR. It is set using an annotation for the loadbalancer:
`hypershift.openshift.io/ingress-controller-load-balancer-scope`

zgalor: In ROSA HCP, we have CS API -> the field is called `ingresses`, because you can define multiple ingresses.
Each item in that list has a `default` bool stating if it's the default ingress. There has to be exactly one (Which is very bad UX and I wouldn't take as an example)

I don't know how this looks in ARO Classic

mbarnes: This is a BU question and I don't know what the plans are in this space, but if we wanted to future-proof it we could add an extra level of nesting like:

```
"ingress": {
    # required
    "default": {
        "visibility": "public"
    }
}
```
which leaves room for supporting more in the future.  Perhaps:
```
"ingress": {
    # required
    "default": {
        "visibility": "public"
    },
    # optional
    "custom": [
        {
            "name": ...
            ... 
        },
        ...
    ],
}
```

davidffrench: Based on the conversation in [slack](https://redhat-external.slack.com/archives/C075PHEFZKQ/p1781793140345969?thread_ts=1781526199.322059&cid=C075PHEFZKQ), the proposal is

```
ingress:
  type: Public|Private|Disabled|Future
```

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-06-30-preview/openapi.json
```diff
@@ -2256,6 +2256,14 @@
             "create"
           ]
         },
+        "ingress": {
+          "$ref": "#/definitions/IngressProfile",
+          "description": "The cluster ingress configuration",
```
miguelsorianod, line 2261: Shouldn't we mention **default** ingress in the description, if the decision is that we don't wanto to make it explicit in the name?

zgalor: Actually, no.
The idea is that if in the future we support additional ingresses, they will also be included in this block, as a custom type

zgalor:
@mbarnes @bennerv PTAL

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-06-30-preview/openapi.json
miguelsorianod: How are we going to support additional ingresses, including both the default and additional ones in the future if this is defined as a discriminated union type? by definition, you specify the type and its corresponding one and only section within it, as it is a union discriminated by a type

mbarnes: Yeah, I never got a satisfactory answer to that in the [Slack thread](https://redhat-external.slack.com/archives/C075PHEFZKQ/p1781526199322059).  Problem with just using the discriminated union is we'd be mixing the type of default ingress with whether or not there's other ingresses: two orthogonal concerns as I see it.

So I guess the enum would have to accommodate all combinations.  Something ugly like:

```
union IngressType {
  string,

  /** Ingress is disabled. */
  Disabled: "Disabled",

  /** The default ingress is visible from the internet. */
  Public: "Public",

  /** The default ingress is not visible from the internet. */
  Private: "Private",

  /** The default ingress is visible from the internet and there are other custom ingresses. */
  PublicAndMultiple: "PublicAndMultiple",

  /** The default ingress is not visible from the internet and there are other custom ingresses. */
  PrivateAndMultiple: "PrivateAndMultiple"
}
```

I don't like it.

mbarnes: @bennerv informs me he doesn't think multiple ingresses will be exposed in any future ARM API version.  Additional ingresses besides default would be a day-2 customer configuration on-cluster.

> Shouldn't we mention default ingress in the description, if the decision is that we don't wanto to make it explicit in the name?

I'm going to update the descriptions for `Public` and `Private` to mention default ingress, like I showed above.

mbarnes:
/lgtm

But since I contributed a commit, I assume will need another approval from @bennerv or someone else.

internal/validation/validate_cluster.go
```diff
@@ -545,6 +551,21 @@ func validateCustomerAPIProfile(ctx context.Context, op operation.Operation, fld
...
 
+var (
+	toIngressType = func(oldObj *api.CustomerIngressProfile) *api.IngressType { return &oldObj.Type }
+)
+
+func validateCustomerIngressProfile(ctx context.Context, op operation.Operation, fldPath *field.Path, newObj, oldObj *api.CustomerIngressProfile) field.ErrorList {
+	errs := field.ErrorList{}
+
+	// Type      IngressType `json:"type,omitempty"`
+	errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("type"), &newObj.Type, nil)...)
```
miguelsorianod, line 562: I think this might be incorrect too with the current state, because on older API versions this parameter doesn't exist and it shouldn't be required. This is related to https://github.com/Azure/ARO-HCP/pull/5643/changes#r3529725859

internal/api/types_cluster.go
```diff
@@ -311,6 +319,9 @@ func (cluster *HCPOpenShiftCluster) EnsureDefaults() {
 	if len(cluster.CustomerProperties.API.Visibility) == 0 {
 		cluster.CustomerProperties.API.Visibility = VisibilityPublic
 	}
+	if len(cluster.CustomerProperties.Ingress.Type) == 0 {
```
miguelsorianod, line 322: Upon further review on this, isn't defaulting also needed on the versioned API types?

cc @zgalor @SudoBrendan

internal/validation/validate_cluster.go
zgalor: Validation won't fail on older API versions because `EnsureDefaults()` populates `Ingress.Type = Public` before validation runs. This is the same pattern used for `API.Visibility`, `Network.NetworkType`, and `Platform.OutboundType` - all are struct fields (not pointers) that get backfilled by `EnsureDefaults()` so validation always sees a value.

internal/api/types_cluster.go
zgalor: Defaulting on the versioned API type is already done - see `SetDefaults()` in `v20260630preview/hcpopenshiftclusters_methods.go` lines 76-81, which defaults `Ingress.Type` to `Public` when the caller omits it. This follows the same pattern as `API.Visibility` and `Platform.OutboundType`.

For older API versions (v20240610, v20251223), the `ingress` field doesn't exist in the versioned type, so there's nothing to default there. The internal type's `EnsureDefaults()` handles backfilling for existing Cosmos documents that predate the field.

miguelsorianod: > Defaulting on the versioned API type is already done - see SetDefaults() in v20260630preview/hcpopenshiftclusters_methods.go lines 76-81, which defaults Ingress.Type to Public when the caller omits it. This follows the same pattern as API.Visibility and Platform.OutboundType.

Yes I was already aware of that part. That covers the cosmos side.

> For older API versions (v20240610, v20251223), the ingress field doesn't exist in the versioned type, so there's nothing to default there. The internal type's EnsureDefaults() handles backfilling for existing Cosmos documents that predate the field.

Don't we need to make sure that ConvertToInternal which takes care of converting the versioned to internal also has the default applied? Cosmos is not the only path / usage.

internal/validation/validate_cluster.go
miguelsorianod: Ignore my comment, I wasn't indenting to send that one but somehow I sent the comment

internal/api/types_cluster.go
zgalor: `EnsureDefaults()` is called immediately after `ConvertToInternal()` in the frontend request flow - see `frontend.go:703`. So, for creates via older API versions: `ConvertToInternal(nil)` produces an empty `Ingress.Type`, then `EnsureDefaults()` sets it to `Public` before validation runs.

For updates via older API versions: `ConvertToInternal(existing)` calls `preserveUnknownClusterFields`, which copies the existing `Ingress` value from the stored document, so the field is preserved.

Both paths are covered.

miguelsorianod: > For updates via older API versions: ConvertToInternal(existing) calls preserveUnknownClusterFields, which copies the existing Ingress value from the stored document, so the field is preserved.

That covers updates, yes.

But for creates, should we cover the case where `EnsureDefaults` is not necessarily called afterwards for some reason in some flow?

zgalor: The architecture intentionally keeps `ConvertToInternal` as a pure conversion function — it never sets defaults, for any field. This is documented in the [DDR: API Version Defaults and Storage Consistency](https://github.com/Azure/ARO-HCP/blob/main/docs/api-version-defaults-and-storage.md), which describes the two-layer defaulting design:

1. **`SetDefaults()` (versioned layer)** — applied *before* `ConvertToInternal`, populates defaults for fields known to the requesting API version
2. **`EnsureDefaults()` (internal layer)** — applied *after* `ConvertToInternal` on CREATE and PREFLIGHT paths, serves as a backstop for fields unknown to the requesting API version

From the DDR:

> `EnsureDefaults()` is also called on the CREATE and PREFLIGHT paths after `ConvertToInternal(nil)` as a backstop for fields unknown to the requesting API version's `SetDefaultValues*()`. This ensures that when a CREATE comes via an older API version, fields introduced in newer versions still get their canonical defaults.

No other field (`Visibility`, `OutboundType`, `NetworkType`, `EtcdKeyManagementMode`, `ClusterImageRegistryState`) sets defaults inside `ConvertToInternal` — `Ingress.Type` follows the same pattern.

If a future code path were to call `ConvertToInternal` without `EnsureDefaults`, that would be a bug in the caller, not a gap in the conversion function.

PR #5648, opened 2026-06-15
feat: move and enable asynchronous CS delete interaction for clusters, nodepools and externalauths

miguelsorianod:
This PR enables the deletion process of clusters, nodepools and externalauth to occur through backend controllers.

This includes removing CS delete interaction in frontend, occurring now in the backend.

### DIFF
--- a/backend/pkg/controllers/validationcontrollers/cluster_validation_controller.go
+++ b/backend/pkg/controllers/validationcontrollers/cluster_validation_controller.go
@@ -81,6 +81,9 @@ func (c *clusterValidationSyncer) SyncOnce(ctx context.Context, key controllerut
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
 	}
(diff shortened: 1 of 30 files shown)


backend/pkg/controllers/validationcontrollers/cluster_validation_controller.go
```diff
@@ -81,6 +81,9 @@ func (c *clusterValidationSyncer) SyncOnce(ctx context.Context, key controllerut
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
 	}
+	if existingCluster.ServiceProviderProperties.DeletionTimestamp != nil {
```
deads2k, line 84: followup: looks like part of "should process" to me.  This is where we need to start using listers for serviceprovidecluster first?

deads2k:
followup desired.

/lgtm

miguelsorianod:
/hold cancel

removing hold as the needed changes in CS and Frontend have been rolled out to production for all regions

PR #5650, opened 2026-06-15
feat: add cluster and nodepool controllers that ensure serviceprovider cosmos resources exist

miguelsorianod:
Also update calls in backend that used GetOrCreateServiceProviderCluster and
GetOrCreateServiceProviderNodePool to just Get variants where if the
resource is not there we return early to retry later.

### DIFF
--- a/backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
+++ b/backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
@@ -108,9 +108,12 @@ func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key
 	// ServiceProviderCluster.Status.ManagementClusterResourceID; until that
 	// lands we have nowhere to write the ReadDesire, so skip and wait for
 	// the next reconcile cycle.
-	spc, err := database.GetOrCreateServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID())
+	spc, err := database.GetServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID())
+	if database.IsNotFoundError(err) {
+		return nil
+	}
 	if err != nil {
-		return utils.TrackError(fmt.Errorf("failed to get or create ServiceProviderCluster: %w", err))
+		return utils.TrackError(fmt.Errorf("failed to get ServiceProviderCluster: %w", err))
 	}
 	mcResourceID := spc.Status.ManagementClusterResourceID
 	if mcResourceID == nil {
--- a/backend/pkg/controllers/upgradecontrollers/control_plane_active_version_controller.go
+++ b/backend/pkg/controllers/upgradecontrollers/control_plane_active_version_controller.go
@@ -104,9 +104,12 @@ func (c *controlPlaneActiveVersionSyncer) SyncOnce(ctx context.Context, key cont
 		return utils.TrackError(err)
 	}
 
-	existingServiceProviderCluster, err := database.GetOrCreateServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID())
+	existingServiceProviderCluster, err := database.GetServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID())
+	if database.IsNotFoundError(err) {
+		return nil
+	}
 	if err != nil {
-		return utils.TrackError(fmt.Errorf("failed to get or create ServiceProviderCluster: %w", err))
+		return utils.TrackError(fmt.Errorf("failed to get ServiceProviderCluster: %w", err))
 	}
 	// Use NeedsUpdate (semantic equality) instead of slices.Equal: HCPClusterActiveVersion holds
 	// *semver.Version, and Go's `==` (which slices.Equal relies on) compares those pointers, not
--- a/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
+++ b/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
@@ -127,9 +127,12 @@ func (c *controlPlaneDesiredVersionSyncer) SyncOnce(ctx context.Context, key con
 		return nil
 	}
 
-	existingServiceProviderCluster, err := database.GetOrCreateServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID())
+	existingServiceProviderCluster, err := database.GetServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID())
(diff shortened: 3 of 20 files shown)


backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -108,9 +108,12 @@ func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key
 	// ServiceProviderCluster.Status.ManagementClusterResourceID; until that
```
miguelsorianod, line 108: Maybe these controllers can fit in the createXXXcontrollers directories once we introduce cs create async creation MRs

internal/database/service_provider_cluster.go
```diff
@@ -84,3 +100,27 @@ func GetOrCreateServiceProviderCluster(
 
 	return existingServiceProviderCluster, nil
 }
+
+// CreateServiceProviderCluster creates the singleton ServiceProviderCluster document
+// for the given cluster resource ID. A nil return means the document was created or
+// already exists (HTTP 409 Conflict).
+func CreateServiceProviderCluster(
```
miguelsorianod, line 106: Maybe we shouldn't provide these helpers and just do it in the controllers where needed

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -290,10 +293,13 @@ func (c *controlPlaneDesiredVersionSyncer) desiredControlPlaneZVersion(ctx conte
 		if err := validation.OpenshiftVersionAtMostOneMinorSkew(actualLatestMinorVersion.String(), desiredMinorVersion.String()); err != nil {
 			return nil, utils.TrackError(err)
 		}
-		clusterNodePools, err := c.listClusterAdmissionNodePools(ctx, clusterResourceID)
+		clusterNodePools, ready, err := c.listClusterAdmissionNodePools(ctx, clusterResourceID)
```
miguelsorianod, line 296: The change in this controller here is non trivial:

The listClusterAdmissionNodePools attempts to list the nodepools and get their corresponding serviceprovidernodepool cosmos resources.

It could in theory happen that a nodepool resource has been created in cosmos but not its corresponding serviceprovidernodepool. Because of this, I updated the listClusterAdmissionNodePools to now also provide a boolean that indicates whether the reconcile logic of this controller can proceed further (the "ready" boolean). If we detect a nodepool that doesn't contain a corresponding serviceprovidernodepool then we stop the execution returning nil, nil. What this does is that the desiredControlPlaneZVersion returns with a result of nil, and because of that no update occurs and if there was some IntentFailed condition is set to false

backend/pkg/controllers/upgradecontrollers/control_plane_active_version_controller.go
```diff
@@ -104,9 +104,12 @@ func (c *controlPlaneActiveVersionSyncer) SyncOnce(ctx context.Context, key cont
 		return utils.TrackError(err)
 	}
 
-	existingServiceProviderCluster, err := database.GetOrCreateServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID())
+	existingServiceProviderCluster, err := database.GetServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID())
```
miguelsorianod, line 107: Although this PR updates backend to change references from GetOrCreateServiceProviderXXX to GetServiceProviderXXX, we still need to decide what to do on the frontend side:

In some frontend endpoints I noticed that we have GetOrCreateServiceProviderXXX calls. For example, in `updateNodePoolInCosmos`. We have the following options:
* Do a get and fail if it does not exist: This will be surfaced to the consumer of the api as internal error or some kind of error
* Do getorcreateserviceprovidercluster: this will ensure it exists. It might error if there's some update being done on it at the same time, which will be surfaced to the end-user
* Do a get and fail if it does not exist, but when we have the cs asynchronous creation process, use the serviceproviderXXX cosmos resource existence as a gate to consider the creation completed. This will improve, but not guarantee, the chance that the serviceprovider resource is there

backend/pkg/controllers/upgradecontrollers/control_plane_active_version_controller.go
```diff
@@ -104,9 +104,12 @@ func (c *controlPlaneActiveVersionSyncer) SyncOnce(ctx context.Context, key cont
...
+	if database.IsNotFoundError(err) {
```
miguelsorianod, line 108: Maybe we should add a log here to be explicit about these cases. It's possible it could be logged multiple times in that case, but might be worth it.

internal/database/service_provider_nodepool.go
```diff
@@ -38,6 +38,23 @@ func newInitialServiceProviderNodePool(npResourceID *azcorearm.ResourceID) *api.
 	}
 }
 
+// GetServiceProviderNodePool gets the singleton ServiceProviderNodePool instance
+// named `default` for the given node pool resource ID.
+func GetServiceProviderNodePool(
```
miguelsorianod, line 43: Maybe it's not worth providing this, and the controllers just do a get, as we do with other type of resources.

internal/database/service_provider_nodepool.go
```diff
@@ -85,3 +102,28 @@ func GetOrCreateServiceProviderNodePool(
 
 	return existingServiceProviderNodePool, nil
 }
+
+// CreateServiceProviderNodePool creates the singleton ServiceProviderNodePool document
+// for the given node pool resource ID. A nil return means the document was created or
+// already exists (HTTP 409 Conflict).
+func CreateServiceProviderNodePool(
```
miguelsorianod, line 108: Maybe it's not worth providing this and the controllers that need it should do it. However, that would force us to make newInitialServiceProviderNodePool public or repeat it in the N places in which case we would lose the guarantee of the creation content being consistent everywhere.

We also need to consider what are we going to do with GetOrCreateServiceProviderXXX calls in frontend.

internal/database/service_provider_cluster.go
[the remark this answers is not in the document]
miguelsorianod: fixed

internal/database/service_provider_nodepool.go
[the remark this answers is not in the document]
miguelsorianod: fixed

davidffrench:
Hi @miguelsorianod — this PR has had no activity for 78 days. Is it still valid and something you plan to continue, or should it be closed? Please let me know. Thanks!

PR #5652, opened 2026-06-15
ARO-27732 - Remove the orphaned resources

ziccardi:
### What

Implement a new periodic controller to identify and clean up orphaned Azure cluster managed resource groups.

The controller:

- Watches subscriptions using the shared `SubscriptionWatchingController` pattern.
- Lists Azure resource groups in the subscription.
- Filters only resource groups that:
  - have a `ManagedBy` value;
  - are in the backend Azure location;
  - are managed by an HCP OpenShift cluster resource ID.
- Lists HCP clusters from the database for the same subscription.
- Builds a set of existing HCP cluster resource IDs.
- Identifies managed resource groups whose `ManagedBy` resource ID points to an HCP cluster that no longer exists.
- Uses structured Azure resource ID parsing and case-insensitive resource ID comparison.
- Checks the current resource group state before attempting deletion.
- Skips resource groups that are already being deleted.
- Handles subscription processing independently so a failure in one subscription does not prevent processing of others.
- Logs detected orphaned managed resource groups and cleanup actions for observability.
- Emits metrics for detected orphaned managed resource groups and deletion failures.
- Adds Prometheus alerts for orphaned managed resource group detection and cleanup failures.

### Environment variables

The controller is guarded behind environment variables so it can be rolled out safely:

- `CLEAN_ORPHANED_MANAGED_RESOURCE_GROUPS=enabled`
  - Enables the orphaned managed resource group cleanup controller.
  - If this variable is not set to `enabled`, the controller is not started.

- `CLEAN_ORPHANED_MANAGED_RESOURCE_GROUPS_MODE=readwrite`
  - Allows the controller to initiate deletion of orphaned managed resource groups.
  - By default, the controller runs in read-only mode and only logs what would be deleted.

This allows the controller to be deployed first in detection/read-only mode, observed through logs, metrics, and alerts, and then switched to read-write mode once we are confident in the results.

### Alerts and metrics

The PR adds Prometheus metrics for:

- orphaned managed resource groups found;
- orphaned managed resource group deletion failures.

The PR also adds alerts for:

- orphaned managed resource groups being detected;
- orphaned managed resource group deletion failures.

### Why

A previously identified race condition could leave Azure managed resource groups orphaned when cluster resources were not cleaned up correctly.

Although the underlying race condition has been resolved, orphaned managed resource groups may still exist in customer subscriptions as a result of the historical bug.

This controller provides the foundation for detecting and eventually removing those orphaned resources, helping reduce resource sprawl and unnecessary Azure consumption.

The controller relies on the `ManagedBy` resource ID to determine whether a managed resource group is still associated with an existing HCP OpenShift cluster.

A managed resource group is considered orphaned only when:

- The resource group has a `ManagedBy` value.
- The `ManagedBy` value points to an HCP OpenShift cluster resource ID.
- That HCP OpenShift cluster no longer exists in the set of clusters listed from the database.

The commit history is intentionally granular to make the review easier. It will be squashed before merge.

### PR Checklist
- [ ] Self-reviewed the diff


backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,215 @@
...
+		if err := allHCPClusters.GetError(); err != nil {
+			return utils.TrackError(err)
+		}
+
+		logger.Info("Found HCP clusters in subscription", "subscriptionID", subscriptionID, "count", len(clusterResourceIDs))
+
+		resourceGroupsPager := rgClient.NewListPager(&armresources.ResourceGroupsClientListOptions{
+			Top: ptr.To(resourceGroupListPageSize),
+		})
+		for resourceGroupsPager.More() {
```
machi1990: consider listing MRGs first before listing clusters; this way you deal with situation where a new MRG appeared because a new cluster (after the existing clusters have been listed) was created. If that situation occurs, you might trigger deletion of that cluster's MRG

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,224 @@
...
+		// List all resource groups in this subscription
+		// We list resource groups in chunks to avoid loading everything at once
+		resourceGroupsPager := rgClient.NewListPager(&armresources.ResourceGroupsClientListOptions{
+			Top: ptr.To(resourceGroupListPageSize),
+		})
+		for resourceGroupsPager.More() {
+			resourceGroupPage, err := resourceGroupsPager.NextPage(ctx)
+			if err != nil {
+				return utils.TrackError(err)
+			}
```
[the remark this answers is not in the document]
machi1990: This makes sense

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,224 @@
...
+			return utils.TrackError(err)
+		}
+
+		// Get all HCP clusters in this subscription and build a map for quick lookup
+		// Key is the lowercase cluster resource ID string
+		clusterResourceIDs := make(map[string]struct{})
+		allHCPClusters, err := c.resourcesDBClient.HCPClusters(subscriptionID, "").List(ctx, nil)
+		if err != nil {
+			return utils.TrackError(err)
+		}
```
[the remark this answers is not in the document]
machi1990: This makes sense as well

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,215 @@
...
+					continue
+				}
+
+				if rg.ManagedBy == nil {
+					continue
+				}
+
+				managedByResourceID := strings.ToLower(*rg.ManagedBy)
+
+				if _, exists := clusterResourceIDs[managedByResourceID]; exists {
```
machi1990: Also check that the managed by is pointed to aro hcp as being the manager. 

i.e the resource type should lowercased should match "microsoft.redhatopenshift/hcpopenshiftclusters"

machi1990: i.e everything else managed by someone else should be ignored

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
machi1990, line 1: wire this controller in the backend/pkg/app.go

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/Azure/ARO-HCP/backend/pkg/listers"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+const (
+	// resourceGroupListPageSize is the number of resource groups to fetch per page
+	// when listing resource groups in a subscription
+	resourceGroupListPageSize int32 = 100
+)
```
machi1990: inline

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,215 @@
...
+		clusterResourceIDs := make(map[string]struct{})
+		allHCPClusters, err := c.resourcesDBClient.HCPClusters(subscriptionID, "").List(ctx, nil)
+		if err != nil {
+			return utils.TrackError(err)
+		}
+
+		for _, cluster := range allHCPClusters.Items(ctx) {
+			if cluster == nil || cluster.ID == nil {
+				continue
+			}
```
machi1990: Is this check needed?

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,215 @@
...
+		for resourceGroupsPager.More() {
+			resourceGroupPage, err := resourceGroupsPager.NextPage(ctx)
+			if err != nil {
+				return utils.TrackError(err)
+			}
+
+			for _, rg := range resourceGroupPage.Value {
+				if rg == nil || rg.Name == nil {
+					continue
+				}
```
machi1990: is this check needed?

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
machi1990: you could do that in the line above

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+const (
+	// resourceGroupListPageSize is the number of resource groups to fetch per page
+	// when listing resource groups in a subscription
+	resourceGroupListPageSize int32 = 100
+)
+
+type cleanOrphanedManagedResourceGroup struct {
```
machi1990: I am thinking in the long run once CS is no longer there, we'll want a controller to delete the MRG in the backend; so it makes sense to keep this controller as the one that will have ultimate that responsibility.

Perhaps, we might then want to rename it to  
```suggestion
type deleteManagedResourceGroup struct {
```

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,215 @@
...
+	subscriptionLister    listers.SubscriptionLister
+	resourcesDBClient     database.ResourcesDBClient
+	azureFPAClientBuilder azureclient.FirstPartyApplicationClientBuilder
+
+	// queue is where incoming work is placed to de-dup and to allow "easy"
+	// rate limited requeues on errors
+	queue workqueue.TypedRateLimitingInterface[string]
+}
+
+// NewCleanOrphanedManagedResourceGroupController periodically looks for managed resource groups
```
machi1990: same comment as https://github.com/Azure/ARO-HCP/pull/5652/changes#r3414212476

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,215 @@
...
+// that are not referenced by any HCPOpenShiftCluster in the database and deletes them.
+func NewCleanOrphanedManagedResourceGroupController(
+	subscriptionLister listers.SubscriptionLister,
+	resourcesDBClient database.ResourcesDBClient,
+	azureFPAClientBuilder azureclient.FirstPartyApplicationClientBuilder,
+) controllerutils.Controller {
+	c := &cleanOrphanedManagedResourceGroup{
+		name:                  "CleanOrphanedManagedResourceGroup",
```
machi1990: same comment as https://github.com/Azure/ARO-HCP/pull/5652/changes#r3414212476

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
machi1990: if you agree, do the change throughout including the file name

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		for resourceGroupsPager.More() {
+			resourceGroupPage, err := resourceGroupsPager.NextPage(ctx)
+			if err != nil {
+				return utils.TrackError(err)
+			}
+
+			for _, rg := range resourceGroupPage.Value {
+				// Only track RH-managed resource groups
+				managedByLowercase := strings.ToLower(*rg.ManagedBy)
+				if !strings.Contains(managedByLowercase, "microsoft.redhatopenshift/hcpopenshiftclusters/") {
```
machi1990: if you parse managedBy to Azure Resource ID then you can use .ResourceType which will indicate the resource type to be used for this. That's a more sound check that contenance check one

machi1990: let's not hardcode it; let's use https://github.com/ziccardi/ARO-HCP/blob/1db534f4efd5d016f81d71cc9aa1ba858536bbeb/internal/api/registry.go#L85

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		})
+		for resourceGroupsPager.More() {
+			resourceGroupPage, err := resourceGroupsPager.NextPage(ctx)
+			if err != nil {
+				return utils.TrackError(err)
+			}
+
+			for _, rg := range resourceGroupPage.Value {
+				// Only track RH-managed resource groups
+				managedByLowercase := strings.ToLower(*rg.ManagedBy)
```
miguelsorianod: This can panic if the resource group being processed is not a managed resourcegroup

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	for _, subscription := range subscriptions {
+		subscriptionID := subscription.ResourceID.SubscriptionID
+		tenantID := *subscription.Properties.TenantId
+
+		rgClient, err := c.azureFPAClientBuilder.ResourceGroupsClient(tenantID, subscriptionID)
+		if err != nil {
+			return utils.TrackError(err)
+		}
+
+		resourceGroupsPager := rgClient.NewListPager(&armresources.ResourceGroupsClientListOptions{
```
miguelsorianod: This will put quite some strain. Have we though potential other alternatives other than listing the MRGs directly and then filtering?

miguelsorianod: What happens with MRGs that might belong to other regions other than the region where we are running? we shouldn't be deleting resources that are not under our regional control

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,235 @@
...
+				Name: "CleanOrphanedManagedResourceGroup",
+			},
+		),
+	}
+
+	return c
+}
+
+// SyncOnce implements the main sync logic for the controller.
+func (c *cleanOrphanedManagedResourceGroup) SyncOnce(ctx context.Context, _ any) error {
```
miguelsorianod: the MRG is a associated to the cluster. Let's be explicit on the names of the types, controllers etc about this. ...ClusterManagedResourceGroup...

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	// Get all subscriptions
+	subscriptions, err := c.subscriptionLister.List(ctx)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	logger.Info("Retrieved subscriptions", "count", len(subscriptions))
+
+	// List all managed resource groups
+	managedResourceGroups := make(map[string]*managedResourceGroupInfo)
```
miguelsorianod: Describe what is the key

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,235 @@
...
+					managedBy:      *rg.ManagedBy,
+				}
+			}
+		}
+	}
+	logger.Info("Found managed resource groups", "count", len(managedResourceGroups))
+
+	// List all HCP clusters
+	logger.Info("Phase 2: Listing all HCP clusters")
+	clusterResourceIDs := make(map[string]struct{})
```
miguelsorianod: Describe what is the key

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
miguelsorianod: As an additional note: this will list all the resourcegroups in the customer's account, and keep all the responses in-memory

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// resourceGroupListPageSize is the number of resource groups to fetch per page
+// when listing resource groups in a subscription
+const resourceGroupListPageSize int32 = 100
+
+// managedResourceGroupInfo holds information about a managed resource group
+type managedResourceGroupInfo struct {
+	subscriptionID string
```
miguelsorianod: The only aspects that seem needed are the MRG resourceid and the managedBy resourceid

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
miguelsorianod: What will occur is that there will be two:
1. One that deletes as part of the deletion process
2. One that deletes orphans

I think based on that the original name is correct.

backend/pkg/azure/client/resource_groups_client.go
```diff
@@ -30,6 +30,7 @@ type ResourceGroupsClient interface {
 		*runtime.Poller[armresources.ResourceGroupsClientDeleteResponse], error)
 	Get(ctx context.Context, resourceGroupName string, options *armresources.ResourceGroupsClientGetOptions) (
 		armresources.ResourceGroupsClientGetResponse, error)
+	NewListPager(options *armresources.ResourceGroupsClientListOptions) *runtime.Pager[armresources.ResourceGroupsClientListResponse]
```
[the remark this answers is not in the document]
miguelsorianod: No need. Just update mocks and whatever needs to implement the interface

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
miguelsorianod: +1. That is a must, as otherwise accidental deletions can occur.

backend/pkg/controllers/clean_orphaned_managed_resource_group_controller.go
ziccardi: I think I agree with @miguelsorianod . I will revert to the old naming.

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
ziccardi: This should be addressed now. Mind a look?

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
ziccardi: > As an additional note: this will list all the resourcegroups in the customer's account, and keep all the responses in-memory

We discussed this and agreed that it shouldn't be an issue. Is there anything in particular that concerns you?

ziccardi: Yeah, good point. I now filter the resources by region.

miguelsorianod: The concern was the azure api rate limits and amount of data transferred (we don't have control over how many resources a user's account has)

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,393 @@
...
+	resourcesDBClient database.ResourcesDBClient,
+	azureFPAClientBuilder azureclient.FirstPartyApplicationClientBuilder,
+) controllerutils.Controller {
+	c := &cleanOrphanedClusterManagedResourceGroup{
+		name:                  "CleanOrphanedClusterManagedResourceGroup",
+		location:              location,
+		subscriptionLister:    subscriptionLister,
+		resourcesDBClient:     resourcesDBClient,
+		azureFPAClientBuilder: azureFPAClientBuilder,
+		queue: workqueue.NewTypedRateLimitingQueueWithConfig(
```
bennerv: why not just make this an extension of the `subscription_watching_controller`?

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,393 @@
...
+			return nil
+		}
+	}
+
+	logger.Info("Initiating deletion of orphaned cluster managed resource group",
+		"subscriptionID", subscriptionID,
+		"resourceGroup", resourceGroupName,
+		"managedBy", managedBy)
+
+	_, err = rgClient.BeginDelete(ctx, resourceGroupName, nil)
```
bennerv: What happens if the deletion fails?  We're triggering the deletion, but not watching its long-running operation / async result.  .  

As a result, we're throwing away any result it may have and losing the error message. 

The resource group will just continue to go through a delete -> fail -> delete cycle, not hot looping because of the cooldown that exists, but will never unstick a managed resource group (unless something changes).

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,393 @@
...
+
+		managedResourceGroups, err := c.listManagedResourceGroupsForSubscription(ctx, rgClient)
+		if err != nil {
+			logger.Error(err, "Failed to list managed resource groups for subscription",
+				"subscriptionID", subscriptionID)
+			errs = append(errs, err)
+			continue
+		}
+
+		clusterResourceIDs, err := c.listClusterResourceIDsForSubscription(ctx, subscription)
```
bennerv: `c.listClusterResourceIDsForSubscription` -> ensures the keys are lowercase?

bennerv: yup
```
// listClusterResourceIDsForSubscription lists all HCP cluster resource IDs for a single subscription
// and returns them as a set (map with empty struct values) with lowercase keys.
```

bennerv (commented):
Should we consider just logging / emitting metrics on resource groups to delete before enabling the deletion functionality? 

Also, we're missing disabling this functionality in dev environments.

Brain dump: 
- [ARO HCP Service Role ](https://dev.azure.com/msazure/One/_git/AD-RBAC-RoleDefinitions?path=/MICROSOFT.REDHATOPENSHIFT/PROD/Azure%20Red%20Hat%20OpenShift%20Hosted%20Control%20Planes%20RP%20Service%20Role.json)has resourceGroups/read permissions across entire subscription @ RP registration time, but **not** `resourceGroups/delete`
- Service Owner role defined as our ManagedBy role in the RP manifest gives us `resourceGroups/delete` when we construct a managed resource group with the ManagedBy flag. 
- As a result, the resource groups created by the 1p app should not be able to delete any resource group in dev unless we explicitly grant permissions over the first party service principal in our tenant.  Our dev clusters utilize the `managedBy` field on RG.
- Dev clusters can't have this controller enabled as we share the mock 1p application and it will have /delete access over all managed resource groups, regardless of which dev env created it.

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
ziccardi: I wasn't aware of that component. Refactoring!
Thanks!

ziccardi:
> Should we consider just logging / emitting metrics on resource groups to delete before enabling the deletion functionality?
> 

@bennerv Hi Ben! I added to env vars (only in the code, not in config in this PR) to enable the controller and to change the controller mode from READONLY (default) or READWRITE.

Mind another look?

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
ziccardi: Added polling to get the result of the deletion. Please give a look!

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	clusterResourceIDs, err := c.listClusterResourceIDsForSubscription(ctx, subscriptionID)
+	if err != nil {
+		logger.Error(err, "Failed to list cluster resource IDs for subscription",
+			"subscriptionID", subscriptionID)
+		return utils.TrackError(err)
+	}
+
+	// Identify and clean up orphaned managed resource groups for this subscription
+	eg, egCtx := errgroup.WithContext(ctx)
+	eg.SetLimit(maxConcurrentDeletions)
```
ziccardi, line 319: With this I limit the number of concurrent routines per execution. Do you think I should move this to the controller, so I can limit the total number of routine per controller instance? @machi1990 @bennerv

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,366 @@
...
+const resourceGroupListPageSize int32 = 100
+
+// deletionPollTimeout is the maximum time to wait for a resource group deletion to complete
+const deletionPollTimeout = 15 * time.Minute
+
+// maxConcurrentDeletions is the maximum number of resource group deletions to run concurrently
+const maxConcurrentDeletions = 20
+
+var (
+	orphanedMRGsFound = promauto.With(legacyregistry.Registerer()).NewCounterVec(
```
machi1990, line 50: Is the usage of legacy registry intentional?

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,366 @@
...
+
+// resourceGroupListPageSize is the number of resource groups to fetch per page
+// when listing resource groups in a subscription
+const resourceGroupListPageSize int32 = 100
+
+// deletionPollTimeout is the maximum time to wait for a resource group deletion to complete
+const deletionPollTimeout = 15 * time.Minute
+
+// maxConcurrentDeletions is the maximum number of resource group deletions to run concurrently
+const maxConcurrentDeletions = 20
```
machi1990: They all could be inlined

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	clusterResourceIDs := make(map[string]struct{})
+
+	allHCPClusters, err := c.resourcesDBClient.HCPClusters(subscriptionID, "").List(ctx, nil)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+
+	if err := allHCPClusters.GetError(); err != nil {
+		return nil, utils.TrackError(err)
+	}
```
machi1990: Should be checked after the iteration; move after .Items for loop

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,366 @@
...
+		clusterResourceIDs[strings.ToLower(cluster.ID.String())] = struct{}{}
+	}
+
+	return clusterResourceIDs, nil
+}
+
+// SyncOnce implements the main sync logic for the controller for a single subscription.
+func (c *cleanOrphanedClusterManagedResourceGroup) SyncOnce(ctx context.Context, key controllerutils.SubscriptionKey) error {
+	logger := utils.LoggerFromContext(ctx)
+	subscriptionID := key.SubscriptionID
```
machi1990: no need for this variable, could be inlined

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
machi1990: I think this is fine.

machi1990 (changes requested):
left a few comments

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
ziccardi: Yeah, I tried to do the same we do in many other places:

- https://github.com/Azure/ARO-HCP/blob/024f7f77fe96e8ae5ae78e490a34f6b014144dfc/backend/pkg/controllers/controllerutils/controller_metrics.go#L26
- https://github.com/Azure/ARO-HCP/blob/d8d7a86757fa736f70ccb5b57b4de309477dc2cb/frontend/pkg/frontend/middleware_panic.go#L37
- ...

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -0,0 +1,364 @@
...
+		return err
+	}
+
+	logger.Info("Successfully initiated deletion of orphaned cluster managed resource group",
+		"subscriptionID", subscriptionID,
+		"resourceGroup", resourceGroupName,
+		"managedBy", managedBy)
+
+	return c.pollResourceGroupDeletion(ctx, subscriptionID, resourceGroupName, managedBy, poller)
+}
```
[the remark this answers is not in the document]
ziccardi: @machi1990 @bennerv This is what we discussed and is the reason why this was a fire and forget in the beginning. Can I close this comment or we want to change something?

backend/pkg/app/backend.go
```diff
@@ -786,6 +795,9 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
 				go createNodePoolScopedReadDesiresController.Run(ctx, 20)
 				go maestroDeleteOrphanedReadonlyBundlesController.Run(ctx, 20)
 				go cleanupLegacyMaestroReadonlyBundlesController.Run(ctx, 1)
+				if os.Getenv("CLEAN_ORPHANED_MANAGED_RESOURCE_GROUPS") == "enabled" {
+					go cleanOrphanedClusterManagedResourceGroupController.Run(ctx, 1)
+				}
```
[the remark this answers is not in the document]
ziccardi: @machi1990 @bennerv Do we want to make this change?

backend/pkg/controllers/clean_orphaned_cluster_managed_resource_group_controller.go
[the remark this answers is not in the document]
machi1990: yes, this can de closed

backend/pkg/app/backend.go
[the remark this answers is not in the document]
machi1990: 20 is okay.

machi1990:
/lgtm

The rough rollout plan of this controller agreed with @bennerv @ziccardi is:
- We merge this PR and have it deployed in all envs as readonly
- @ziccardi will then grab the logs from all the envs and analyse
- @ziccardi will then work on properly detecting Stage vs Prod for the UK south region: We'll need a config to detect weather it is Stage and if the Stage AFEC flag is enabled on the subscription then only we can delete for Stage. This avoid deleting an MRG for a prod cluster. 
- @ziccardi that detection will be merged and deployed in readonly
- @ziccardi analyse the log; if all looks good, enable the controller starting from lower envs to higher one. Otherwise, fix and go back to step above

@bennerv I don't have approval permission, can you give the PR another look and approve it so that we get it deployed in readonly mode? Thanks

raelga (approved):
/lgtm
/approve

PR #5670, opened 2026-06-16
feat(backend): surface silent kube-applier gate in logs (AROSLSRE-1188)

raelga:
Jira: [AROSLSRE-1188](https://redhat.atlassian.net/browse/AROSLSRE-1188) — follow-up to [ARO-27510](https://redhat.atlassian.net/browse/ARO-27510)

### What

Add structured logging in `KubeApplierDBClients.For()` so its silent `return nil` gates become queryable:

- `mc == nil` at `V(1)` (fleet informer lag; transient).
- empty `Status.KubeApplierCosmosContainerName` at `Info` — not expected in steady state; this is the [ARO-27510](https://redhat.atlassian.net/browse/ARO-27510) signature.
- `NewContainer` construction failure at `Error`.

The caller `createClusterScopedReadDesiresSyncer.SyncOnce` builds a cluster-scoped logger and injects it into `ctx` (`utils.ContextWithLogger`) so `For()`'s logs carry the subscription / resource-group / cluster fields. No logging is added on the controller's transient early-return gates, per convention.

No behavioural change — observability-only.

### Why

In [ARO-27510](https://redhat.atlassian.net/browse/ARO-27510), the STG Fleet management-cluster document was missing `Status.KubeApplierCosmosContainerName`. `For()` returned nil, the per-cluster ReadDesire was never written, and cluster-create hung at `Provisioning("")` until the 20-minute e2e timeout — with **no signal anywhere in Kusto**. The empty-container-name gate now logs loudly so this class of regression surfaces immediately next time.

### Testing

- `go build ./backend/pkg/controllers/... ./internal/database/...`
- `go vet` clean on both packages.
- `go test ./backend/pkg/controllers/ ./internal/database/ -count=1` — pass.

Per review feedback, transient-state logging on the controller's early-exit gates was dropped; the only logging added is in the kube-applier client. The cluster-scoped logger is retained solely to propagate cluster-identifying fields into `For()`'s logs via `ctx`.

### PR Checklist
- [x] Self-reviewed the diff

### DIFF
--- a/backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
+++ b/backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
@@ -90,6 +90,15 @@ func NewCreateClusterScopedReadDesiresController(
 }
 
 func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	logger := utils.LoggerFromContext(ctx).WithValues(utils.LogValues{}.
+		AddSubscriptionID(key.SubscriptionID).
+		AddResourceGroup(key.ResourceGroupName).
+		AddHCPClusterName(key.HCPClusterName)...)
+	// Inject the cluster-scoped logger so downstream helpers (e.g.
+	// kubeApplierDBClients.For) that read utils.LoggerFromContext(ctx)
+	// also emit the cluster-identifying fields.
+	ctx = utils.ContextWithLogger(ctx, logger)
+
 	existingCluster, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
 	if database.IsNotFoundError(err) {
 		return nil
@@ -137,7 +146,9 @@ func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key
 	kaClient := c.kubeApplierDBClients.For(ctx, mcResourceID)
 	if kaClient == nil {
 		// Registry doesn't have an entry yet for this MC (e.g. the fleet
-		// lister hasn't caught up). Skip and rely on retrigger.
+		// lister hasn't caught up). Skip and rely on retrigger. When the MC
+		// document is registered but misconfigured (e.g. missing its
+		// kube-applier container name), For() surfaces that loudly.
 		return nil
 	}
 	crud, err := kaClient.ReadDesiresForCluster(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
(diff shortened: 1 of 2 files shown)


backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -90,15 +90,22 @@ func NewCreateClusterScopedReadDesiresController(
 }
 
 func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	logger := utils.LoggerFromContext(ctx).WithValues(utils.LogValues{}.
+		AddSubscriptionID(key.SubscriptionID).
+		AddResourceGroup(key.ResourceGroupName).
+		AddHCPClusterName(key.HCPClusterName)...)
+
```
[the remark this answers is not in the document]
raelga: Good catch — addressed in c60b79a9b: the cluster-scoped logger is now injected back into `ctx` via `utils.ContextWithLogger(ctx, logger)` right after it is built, so `kubeApplierDBClients.For` and any other downstream `utils.LoggerFromContext(ctx)` consumers inherit the subscription/resource-group/cluster fields.

raelga:
Per review feedback (thanks @geoberle): dropped the logging on the controller's transient early-exit gates — convention is not to log on transient state. The only logging added is now in `KubeApplierDBClients.For()` (empty container name at `Info`, the [ARO-27510](https://redhat.atlassian.net/browse/ARO-27510) signature; mc-not-found at `V(1)`; `NewContainer` failure at `Error`).

`SyncOnce` still builds a cluster-scoped logger and injects it into `ctx` so `For()`'s logs carry the subscription / resource-group / cluster fields. Title/description updated to match. Force-pushed `f549b82d3`.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -90,6 +90,15 @@ func NewCreateClusterScopedReadDesiresController(
 }
 
 func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	logger := utils.LoggerFromContext(ctx).WithValues(utils.LogValues{}.
```
miguelsorianod, line 93: @raelga this shouldn't be needed. This controller is a ClusterWatchingController, which is a genericWatchingController, and all of those already set those attributes. Can we undo this specific change?

internal/database/kube_applier_client.go
```diff
@@ -408,19 +408,30 @@ func (c *kubeApplierDBClients) For(ctx context.Context, managementClusterResourc
 		return existing
 	}
 
+	logger := utils.LoggerFromContext(ctx).WithValues(
```
miguelsorianod, line 411: This will add this extra information into the logger but it is not being stored in the context, which means that inner calls to other functions that receive the context won't log with the managementClusterResourceID key. Is that intended?

PR #5673, opened 2026-06-16
fix: prevent resource leaks in Maestro REST client and subscription metrics ticker

rhamitarora:
NO-JIRA

### What

Fixes two resource leaks in long-running loops:

- **Backend (Maestro REST client):** `newRESTClient` clones `http.DefaultTransport`,
  giving each client its own connection pool. Tearing the client down via context
  cancellation only stopped the gRPC client and never drained the REST transport's
  idle connections, leaking TCP connections/FDs across repeated client creations.
  `newRESTClient` now returns the underlying `*http.Client`, and `NewClient` registers
  `context.AfterFunc(ctx, restHTTPClient.CloseIdleConnections)` so the existing
  cancellation-based teardown also drains the pool. The `AfterFunc` stop function is
  captured and invoked (with an immediate drain) on the gRPC-init error path so a
  failed construction doesn't retain the `http.Client` until `ctx` is cancelled.
- **Frontend (metrics ticker):** `SubscriptionCollector.Run` created a
  `time.NewTicker(30s)` that was never stopped; added `defer t.Stop()` to release the
  timer on shutdown.

### Why

The orphan/cleanup controllers rebuild Maestro clients every sync cycle. Without
draining the REST client's idle connection pool on teardown, idle TCP connections and
file descriptors accumulate for the process lifetime, risking FD exhaustion. Similarly,
a `time.Ticker` is not garbage-collected until `Stop()` is called, so the unstopped
ticker leaks its runtime timer when the collector loop exits.

### Testing

`go build ./...` and `go vet ./...` pass for the affected packages.

`context.AfterFunc` requires Go 1.21+; this module targets Go 1.25.7 (`backend/go.mod`),
so it compiles cleanly. The PR bundles a small frontend ticker fix alongside the backend
teardown fix — both are self-contained "resource cleanup on teardown" changes. If you
prefer strict one-PR-per-task, the frontend change can be split into its own PR.

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/backend/pkg/maestro/maestro_client.go
+++ b/backend/pkg/maestro/maestro_client.go
@@ -76,10 +76,14 @@ func NewClient(
 	// connection pool. Callers tear the client down by cancelling ctx, so we
 	// drain the pool's idle connections on cancellation to avoid leaking TCP
 	// connections and file descriptors across repeated client creations.
-	context.AfterFunc(ctx, restHTTPClient.CloseIdleConnections)
+	stopDrainIdleConns := context.AfterFunc(ctx, restHTTPClient.CloseIdleConnections)
 
 	grpcClient, err := newGRPCSourceWorkClient(ctx, maestroGRPCAPIEndpoint, restClient, maestroSourceID)
 	if err != nil {
+		// Unregister the teardown callback so we don't retain references until
+		// ctx is cancelled when client creation fails here.
+		stopDrainIdleConns()
+		restHTTPClient.CloseIdleConnections()
 		return nil, utils.TrackError(fmt.Errorf("failed to create maestro grpc source work client: %w", err))
 	}
 
--- a/backend/pkg/maestro/maestro_client.go
+++ b/backend/pkg/maestro/maestro_client.go
@@ -71,7 +71,13 @@ func NewClient(
 	ctx context.Context,
 	maestroRESTAPIEndpoint string, maestroGRPCAPIEndpoint string, maestroConsumerName string, maestroSourceID string,
 ) (Client, error) {
-	restClient := newRESTClient(maestroRESTAPIEndpoint)
+	restClient, restHTTPClient := newRESTClient(maestroRESTAPIEndpoint)
+	// The REST client uses a cloned http.Transport, which owns its own
+	// connection pool. Callers tear the client down by cancelling ctx, so we
+	// drain the pool's idle connections on cancellation to avoid leaking TCP
+	// connections and file descriptors across repeated client creations.
+	context.AfterFunc(ctx, restHTTPClient.CloseIdleConnections)
+
 	grpcClient, err := newGRPCSourceWorkClient(ctx, maestroGRPCAPIEndpoint, restClient, maestroSourceID)
 	if err != nil {
 		return nil, utils.TrackError(fmt.Errorf("failed to create maestro grpc source work client: %w", err))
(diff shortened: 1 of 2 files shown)


backend/pkg/maestro/maestro_client.go
```diff
@@ -71,7 +71,13 @@ func NewClient(
 	ctx context.Context,
 	maestroRESTAPIEndpoint string, maestroGRPCAPIEndpoint string, maestroConsumerName string, maestroSourceID string,
 ) (Client, error) {
-	restClient := newRESTClient(maestroRESTAPIEndpoint)
+	restClient, restHTTPClient := newRESTClient(maestroRESTAPIEndpoint)
+	// The REST client uses a cloned http.Transport, which owns its own
+	// connection pool. Callers tear the client down by cancelling ctx, so we
+	// drain the pool's idle connections on cancellation to avoid leaking TCP
+	// connections and file descriptors across repeated client creations.
+	context.AfterFunc(ctx, restHTTPClient.CloseIdleConnections)
```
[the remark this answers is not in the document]
rhamitarora: Thanks for flagging this. [context.AfterFunc](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) was added in Go 1.21, and this module targets Go 1.25.7 (see go 1.25.7 in [backend/go.mod](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)), so it's well within the supported toolchain and compiles fine — no compile-time break. The repo doesn't support Go < 1.21, so the goroutine fallback isn't needed here. Keeping context.AfterFunc as the idiomatic approach.

backend/pkg/maestro/maestro_client.go
```diff
@@ -71,9 +71,19 @@ func NewClient(
 	ctx context.Context,
 	maestroRESTAPIEndpoint string, maestroGRPCAPIEndpoint string, maestroConsumerName string, maestroSourceID string,
 ) (Client, error) {
-	restClient := newRESTClient(maestroRESTAPIEndpoint)
+	restClient, restHTTPClient := newRESTClient(maestroRESTAPIEndpoint)
```
miguelsorianod, line 74: There is no need to change the signature to return two parameters. The own `restClient` allows you to access the underlying http client

backend/pkg/maestro/maestro_client.go
```diff
@@ -71,9 +71,19 @@ func NewClient(
...
+	// The REST client uses a cloned http.Transport, which owns its own
```
miguelsorianod, line 75: As a side note, we are not really using Maestro anymore in backend so there shouldn't be connections anymore. I guess we can still introduce it just in case for some reason we end up needing maestro temporarily again

backend/pkg/maestro/maestro_client.go
```diff
@@ -71,9 +71,19 @@ func NewClient(
...
+	// connection pool. Callers tear the client down by cancelling ctx, so we
+	// drain the pool's idle connections on cancellation to avoid leaking TCP
+	// connections and file descriptors across repeated client creations.
+	stopDrainIdleConns := context.AfterFunc(ctx, restHTTPClient.CloseIdleConnections)
```
miguelsorianod, line 79: Out of curiosity, what is that we detected? it's true that after the maestro client's context is cancelled the pool remains there with connections, but it should be temporary and not an observed permanent increase: the http connection pool that we have, which is a clone of the default one has an IdleConnTimeout of 90 seconds, so those should not be leaked permanently

PR #5682, opened 2026-06-16
feat: enable customer-managed key for nodePool OS disk encryption

cadenmarchese:
https://redhat.atlassian.net/browse/ARO-27738

### What

allows BYOK for nodePool OS disk encryption. Currently, since we don't pass the diskEncryptionSet ID to cluster service, all nodePool OS disks fall back to platform-managed keys.

This also unblocks mHSM support for nodePool OS Disk encryption (but not etcd, that requires HyperShift changes)

This requires service managed identity (or the mock) to have Reader role over the disk encryption set, otherwise e2e fails with the following:

```
: Nodepool OS Disk Encryption should create a nodepool with customer-managed disk encryption via DES expand_less23m40s{fail [github.com/Azure/ARO-HCP/test/e2e/nodepool_osdisk_encryption.go:134]: failed to create nodepool des-np with DES Unexpected error:     <*fmt.wrapError \| 0xc0015ba6c0>:      failed to create NodePool des-np: failed waiting for nodepool="des-np" for cluster "des-encrypt" in resourcegroup="des-encrypt-6jl8pq" to finish creating: GET https://rp.j6608000.hcpsvc.osadev.cloud/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationStatuses/c915d1eb-59bc-41f2-800e-47d6179fae28     --------------------------------------------------------------------------------     RESPONSE 200: 200 OK     ERROR CODE: InternalServerError     --------------------------------------------------------------------------------     {       "id": "/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationStatuses/c915d1eb-59bc-41f2-800e-47d6179fae28",       "name": "c915d1eb-59bc-41f2-800e-47d6179fae28",       "status": "Failed",       "startTime": "2026-07-16T22:00:30.305420763Z",       "endTime": "2026-07-16T22:00:45.224060261Z",       "error": {         "code": "InternalServerError",         "message": "[clusterServiceNodePoolStatus] GET https://management.azure.com/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/des-encrypt-6jl8pq/providers/Microsoft.Compute/diskEncryptionSets/des-encrypt-des-o3nbqnh7nwvje\n--------------------------------------------------------------------------------\nRESPONSE 403: 403 Forbidden\nERROR CODE: AuthorizationFailed\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"AuthorizationFailed\",\n    \"message\": \"The client '33e65307-a696-491e-831f-42341ce3a9b5' with object id 'b26c2343-0863-4a55-bb7e-dfb2544999c1' does not have authorization to perform action 'Microsoft.Compute/diskEncryptionSets/read' over scope '/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/des-encrypt-6jl8pq/providers/Microsoft.Compute/diskEncryptionSets/des-encrypt-des-o3nbqnh7nwvje' or the scope is invalid. If access was recently granted, please refresh your credentials.\"\n  }\n}\n--------------------------------------------------------------------------------\n"       }     }     -------------------------------------------------------------------------------- | : Nodepool OS Disk Encryption should create a nodepool with customer-managed disk encryption via DES expand_less | 23m40s | {fail [github.com/Azure/ARO-HCP/test/e2e/nodepool_osdisk_encryption.go:134]: failed to create nodepool des-np with DES Unexpected error:     <*fmt.wrapError \| 0xc0015ba6c0>:      failed to create NodePool des-np: failed waiting for nodepool="des-np" for cluster "des-encrypt" in resourcegroup="des-encrypt-6jl8pq" to finish creating: GET https://rp.j6608000.hcpsvc.osadev.cloud/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationStatuses/c915d1eb-59bc-41f2-800e-47d6179fae28     --------------------------------------------------------------------------------     RESPONSE 200: 200 OK     ERROR CODE: InternalServerError     --------------------------------------------------------------------------------     {       "id": "/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationStatuses/c915d1eb-59bc-41f2-800e-47d6179fae28",       "name": "c915d1eb-59bc-41f2-800e-47d6179fae28",       "status": "Failed",       "startTime": "2026-07-16T22:00:30.305420763Z",       "endTime": "2026-07-16T22:00:45.224060261Z",       "error": {         "code": "InternalServerError",         "message": "[clusterServiceNodePoolStatus] GET https://management.azure.com/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/des-encrypt-6jl8pq/providers/Microsoft.Compute/diskEncryptionSets/des-encrypt-des-o3nbqnh7nwvje\n--------------------------------------------------------------------------------\nRESPONSE 403: 403 Forbidden\nERROR CODE: AuthorizationFailed\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"AuthorizationFailed\",\n    \"message\": \"The client '33e65307-a696-491e-831f-42341ce3a9b5' with object id 'b26c2343-0863-4a55-bb7e-dfb2544999c1' does not have authorization to perform action 'Microsoft.Compute/diskEncryptionSets/read' over scope '/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/des-encrypt-6jl8pq/providers/Microsoft.Compute/diskEncryptionSets/des-encrypt-des-o3nbqnh7nwvje' or the scope is invalid. If access was recently granted, please refresh your credentials.\"\n  }\n}\n--------------------------------------------------------------------------------\n"       }     }     --------------------------------------------------------------------------------
: Nodepool OS Disk Encryption should create a nodepool with customer-managed disk encryption via DES expand_less | 23m40s
{fail [github.com/Azure/ARO-HCP/test/e2e/nodepool_osdisk_encryption.go:134]: failed to create nodepool des-np with DES Unexpected error:     <*fmt.wrapError \| 0xc0015ba6c0>:      failed to create NodePool des-np: failed waiting for nodepool="des-np" for cluster "des-encrypt" in resourcegroup="des-encrypt-6jl8pq" to finish creating: GET https://rp.j6608000.hcpsvc.osadev.cloud/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationStatuses/c915d1eb-59bc-41f2-800e-47d6179fae28     --------------------------------------------------------------------------------     RESPONSE 200: 200 OK     ERROR CODE: InternalServerError     --------------------------------------------------------------------------------     {       "id": "/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationStatuses/c915d1eb-59bc-41f2-800e-47d6179fae28",       "name": "c915d1eb-59bc-41f2-800e-47d6179fae28",       "status": "Failed",       "startTime": "2026-07-16T22:00:30.305420763Z",       "endTime": "2026-07-16T22:00:45.224060261Z",       "error": {         "code": "InternalServerError",         "message": "[clusterServiceNodePoolStatus] GET https://management.azure.com/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/des-encrypt-6jl8pq/providers/Microsoft.Compute/diskEncryptionSets/des-encrypt-des-o3nbqnh7nwvje\n--------------------------------------------------------------------------------\nRESPONSE 403: 403 Forbidden\nERROR CODE: AuthorizationFailed\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"AuthorizationFailed\",\n    \"message\": \"The client '33e65307-a696-491e-831f-42341ce3a9b5' with object id 'b26c2343-0863-4a55-bb7e-dfb2544999c1' does not have authorization to perform action 'Microsoft.Compute/diskEncryptionSets/read' over scope '/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/des-encrypt-6jl8pq/providers/Microsoft.Compute/diskEncryptionSets/des-encrypt-des-o3nbqnh7nwvje' or the scope is invalid. If access was recently granted, please refresh your credentials.\"\n  }\n}\n--------------------------------------------------------------------------------\n"       }     }     --------------------------------------------------------------------------------
```

### Why

This is required to get BYOK for nodePool OS Disk encryption working, both for regular keyvaults, and mHSM keyvaults. Today the HCP API will accept a disk encryption set but won't use it for the actual encryption.

### PR Checklist
- [x] Self-reviewed the diff

### DIFF
--- a/internal/ocm/convert.go
+++ b/internal/ocm/convert.go
@@ -224,6 +224,17 @@ func convertEnableEncryptionAtHostToCSBuilder(in coreapi.NodePoolPlatformProfile
 	return arohcpv1alpha1.NewAzureNodePoolEncryptionAtHost().State(state)
 }
 
+func buildCSOsDisk(osDisk coreapi.OSDiskProfile, storageAccountType, persistence string) *arohcpv1alpha1.AzureNodePoolOsDiskBuilder {
+	builder := arohcpv1alpha1.NewAzureNodePoolOsDisk().
+		SizeGibibytes(int(*osDisk.SizeGiB)).
+		StorageAccountType(storageAccountType).
+		Persistence(persistence)
+	if osDisk.EncryptionSetID != nil {
+		builder.SseEncryptionSetResourceId(osDisk.EncryptionSetID.String())
+	}
+	return builder
+}
+
 func convertClusterImageRegistryStateRPToCS(in coreapi.ClusterImageRegistryProfile) (string, error) {
 	switch in.State {
 	case metadataapi.ClusterImageRegistryStateDisabled:
@@ -629,10 +640,7 @@ func BuildCSNodePool(ctx context.Context, nodePool *coreapi.HCPOpenShiftClusterN
 				ResourceName(strings.ToLower(nodePool.Name)).
 				VMSize(nodePool.Properties.Platform.VMSize).
 				EncryptionAtHost(convertEnableEncryptionAtHostToCSBuilder(nodePool.Properties.Platform)).
-				OsDisk(arohcpv1alpha1.NewAzureNodePoolOsDisk().
-					SizeGibibytes(int(*nodePool.Properties.Platform.OSDisk.SizeGiB)).
-					StorageAccountType(csDiskStorageAccountType).
-					Persistence(csPersistence))).
+				OsDisk(buildCSOsDisk(nodePool.Properties.Platform.OSDisk, csDiskStorageAccountType, csPersistence))).
 			AvailabilityZone(nodePool.Properties.Platform.AvailabilityZone).
 			AutoRepair(nodePool.Properties.AutoRepair)
 	}
--- a/test/e2e-setup/bicep/modules/customer-infra.bicep
+++ b/test/e2e-setup/bicep/modules/customer-infra.bicep
@@ -22,6 +22,15 @@ param privateKeyVault bool = false
 @description('Assign Key Vault Crypto Officer role to the deployer on the customer KeyVault that contains etcd encryption key')
 param assignKeyVaultCryptoOfficer bool = false
 
+@description('Enable soft delete on the customer Key Vault')
+param enableKeyVaultSoftDelete bool = false
+
+@description('Enable purge protection on the customer Key Vault (requires soft delete)')
+param enableKeyVaultPurgeProtection bool = false
+
+@description('Soft delete retention in days')
+param keyVaultSoftDeleteRetentionInDays int = 7
+
 //
 // Variables
 //
@@ -99,7 +108,9 @@ resource customerKeyVault 'Microsoft.KeyVault/vaults@2024-12-01-preview' = {
   location: resourceGroup().location
   properties: {
     enableRbacAuthorization: true
-    enableSoftDelete: false
+    enableSoftDelete: enableKeyVaultSoftDelete ? true : null
+    enablePurgeProtection: enableKeyVaultPurgeProtection ? true : null
+    softDeleteRetentionInDays: enableKeyVaultSoftDelete ? keyVaultSoftDeleteRetentionInDays : null
     tenantId: subscription().tenantId
     publicNetworkAccess: privateKeyVault ? 'Disabled' : 'Enabled'
     sku: {
--- a/test/e2e/nodepool_osdisk_encryption.go
+++ b/test/e2e/nodepool_osdisk_encryption.go
@@ -1,0 +1,165 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/msi/armmsi"
+
+	hcpsdk20251223preview "github.com/Azure/ARO-HCP/test/sdk/v20251223preview/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("Nodepool OS Disk Encryption", func() {
+	It("should create a nodepool with customer-managed disk encryption via DES",
+		labels.RequireNothing,
+		labels.Critical,
+		labels.Positive,
+		labels.AroRpApiCompatible,
+		labels.MIContainers(1),
+		func(ctx context.Context) {
+			const (
+				customerClusterName  = "des-encrypt"
+				customerNodePoolName = "des-np"
+			)
+
+			tc := framework.NewTestContext()
+
+			if tc.UsePooledIdentities() {
+				err := tc.AssignIdentityContainers(ctx, 1, framework.IdentityContainerAssignmentRetryInterval)
+				Expect(err).NotTo(HaveOccurred(), "failed to assign pooled identity containers")
+			}
+
+			By("creating a resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "des-encrypt", tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create resource group")
+
+			By("creating cluster parameters")
+			clusterParams := framework.NewDefaultClusterParams20251223()
+			clusterParams.ClusterName = customerClusterName
+			managedResourceGroupName := framework.SuffixName(*resourceGroup.Name, "-managed", 64)
+			clusterParams.ManagedResourceGroupName = managedResourceGroupName
+
+			By("creating customer resources (infrastructure and managed identities)")
+			clusterParams, err = tc.CreateClusterCustomerResources20251223(ctx,
+				resourceGroup,
+				clusterParams,
+				map[string]interface{}{
+					"assignKeyVaultCryptoOfficer":       true,
+					"enableKeyVaultSoftDelete":          true,
+					"enableKeyVaultPurgeProtection":     true,
+					"keyVaultSoftDeleteRetentionInDays": 7,
+				},
+				TestArtifactsFS,
+				framework.RBACScopeResourceGroup,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create cluster customer resources")
+
+			By("resolving service managed identity principal ID")
+			Expect(clusterParams.UserAssignedIdentitiesProfile).NotTo(BeNil(), "cluster params missing UserAssignedIdentitiesProfile")
+			Expect(clusterParams.UserAssignedIdentitiesProfile.ServiceManagedIdentity).NotTo(BeNil(), "cluster params missing ServiceManagedIdentity resource ID")
+
+			serviceMIResourceID, err := azcorearm.ParseResourceID(*clusterParams.UserAssignedIdentitiesProfile.ServiceManagedIdentity)
+			Expect(err).NotTo(HaveOccurred(), "failed to parse service managed identity resource ID")
+
+			subscriptionID, err := tc.SubscriptionID(ctx)
+			Expect(err).NotTo(HaveOccurred(), "failed to get subscription ID")
+
+			creds, err := tc.AzureCredential()
+			Expect(err).NotTo(HaveOccurred(), "failed to get Azure credentials")
+
+			msiClientFactory, err := armmsi.NewClientFactory(subscriptionID, creds, nil)
+			Expect(err).NotTo(HaveOccurred(), "failed to create MSI client factory")
+
+			serviceMI, err := msiClientFactory.NewUserAssignedIdentitiesClient().Get(ctx, serviceMIResourceID.ResourceGroupName, serviceMIResourceID.Name, nil)
+			Expect(err).NotTo(HaveOccurred(), "failed to get service managed identity")
+			Expect(serviceMI.Properties.PrincipalID).NotTo(BeNil(), "service managed identity has no principal ID")
+
+			By("creating disk encryption set backed by KeyVault")
+			desResourceID, err := tc.CreateDiskEncryptionSet(ctx, *resourceGroup.Name, clusterParams.KeyVaultName, customerClusterName, *serviceMI.Properties.PrincipalID, tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create disk encryption set")
+
+			By("creating the HCP cluster")
+			err = tc.CreateHCPClusterFromParam20251223(ctx,
+				GinkgoLogr,
+				*resourceGroup.Name,
+				clusterParams,
+				nil,
+				framework.ClusterCreationTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create HCP cluster %s", customerClusterName)
+
+			By("creating the nodepool with disk encryption set")
+			nodePoolParams := framework.NewDefaultNodePoolParams20251223()
+			nodePoolParams.ClusterName = customerClusterName
+			nodePoolParams.NodePoolName = customerNodePoolName
+			nodePoolParams.EncryptionSetID = desResourceID
+
+			err = tc.CreateNodePoolFromParam20251223(ctx,
+				GinkgoLogr,
+				*resourceGroup.Name,
+				managedResourceGroupName,
+				customerClusterName,
+				nodePoolParams,
+				framework.NodePoolCreationTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create nodepool %s with DES", customerNodePoolName)
+
+			By("verifying nodepool ARM resource has encryptionSetId")
+			created, err := framework.GetNodePool20251223(ctx,
+				tc.Get20251223ClientFactoryOrDie(ctx).NewNodePoolsClient(),
+				*resourceGroup.Name,
+				customerClusterName,
+				customerNodePoolName,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to get nodepool %s", customerNodePoolName)
+			Expect(created.Properties).ToNot(BeNil(), "nodepool Properties was nil")
+			Expect(created.Properties.ProvisioningState).ToNot(BeNil(), "nodepool ProvisioningState was nil")
+			Expect(*created.Properties.ProvisioningState).To(Equal(hcpsdk20251223preview.ProvisioningStateSucceeded), "nodepool %s should be Succeeded", customerNodePoolName)
+			Expect(created.Properties.Platform).ToNot(BeNil(), "nodepool Platform was nil")
+			Expect(created.Properties.Platform.OSDisk).ToNot(BeNil(), "nodepool OSDisk was nil")
+			Expect(created.Properties.Platform.OSDisk.EncryptionSetID).ToNot(BeNil(),
+				"nodepool OSDisk.EncryptionSetID should be set")
+			Expect(strings.EqualFold(*created.Properties.Platform.OSDisk.EncryptionSetID, desResourceID)).To(BeTrue(),
+				"nodepool OSDisk.EncryptionSetID should match the DES resource ID")
+
+			By("getting credentials to verify cluster health")
+			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
+				ctx,
+				tc.Get20240610ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient(),
+				*resourceGroup.Name,
+				customerClusterName,
+				framework.GetAdminRESTConfigTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to get admin REST config")
+
+			By("verifying cluster health, node readiness, and VM OS disk encryption in parallel")
+			computeFactory := tc.GetARMComputeClientFactoryOrDie(ctx)
+			err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig,
+				verifiers.VerifyNodeCount(customerClusterName, int(nodePoolParams.Replicas)),
+				verifiers.VerifyNodesReady(),
+				verifiers.VerifyVMOSDiskCustomerEncryption(computeFactory, managedResourceGroupName, customerNodePoolName, desResourceID),
+			)
+			Expect(err).NotTo(HaveOccurred(), "cluster verification failed")
+		})
+})
--- a/test/util/framework/helpers_v20240610preview.go
+++ b/test/util/framework/helpers_v20240610preview.go
@@ -93,6 +93,7 @@ type NodePoolParams20240610 struct {
 	AutoScaling      *NodePoolAutoScalingParams
 	AvailabilityZone string
 	Tags             map[string]*string
+	EncryptionSetID  string
 }
 
 // ========================================================================
@@ -1259,6 +1260,10 @@ func BuildNodePoolFromParams20240610(
 				AvailabilityZone: to.Ptr(parameters.AvailabilityZone),
 			},
 		},
+	}
+
+	if parameters.EncryptionSetID != "" {
+		nodePool.Properties.Platform.OSDisk.EncryptionSetID = to.Ptr(parameters.EncryptionSetID)
 	}
 
 	if parameters.AutoScaling != nil {
--- a/test/util/framework/per_invocation_framework.go
+++ b/test/util/framework/per_invocation_framework.go
@@ -81,6 +81,11 @@ var (
 const (
 	StandardPollInterval            = 10 * time.Second
 	StandardResourceGroupExpiration = 4 * time.Hour
+	// keyVaultPurgeTimeout bounds the best-effort soft-deleted Key Vault purge
+	// on resource-group teardown so a stalled purge (or Azure control-plane
+	// stall) can never hang the cleanup goroutine and flake the overall E2E
+	// teardown.
+	keyVaultPurgeTimeout = 5 * time.Minute
 )
 
 // azureRetryOptions configures the Azure SDK retry policy for e2e tests.
--- a/test/util/framework/per_invocation_framework.go
+++ b/test/util/framework/per_invocation_framework.go
@@ -81,11 +81,6 @@ var (
 const (
 	StandardPollInterval            = 10 * time.Second
 	StandardResourceGroupExpiration = 4 * time.Hour
-	// keyVaultPurgeTimeout bounds the best-effort soft-deleted Key Vault purge
-	// on resource-group teardown so a stalled purge (or Azure control-plane
-	// stall) can never hang the cleanup goroutine and flake the overall E2E
-	// teardown.
-	keyVaultPurgeTimeout = 5 * time.Minute
 )
 
 // azureRetryOptions configures the Azure SDK retry policy for e2e tests.
--- a/test/util/framework/per_test_framework.go
+++ b/test/util/framework/per_test_framework.go
@@ -45,10 +45,8 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	armauthorization "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v3"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5"
-	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions"
@@ -685,11 +683,6 @@ func (tc *perItOrDescribeTestContext) cleanupResourceGroup(ctx context.Context, 
 		return fmt.Errorf("failed to cleanup resource group: %w", err)
 	}
 
-	// Deleting the resource group only soft-deletes any Key Vaults it contained;
-	// their globally-unique names stay reserved until purged. Purge them so a
-	// later run reusing a colliding vault name does not hit VaultAlreadyExists.
-	tc.purgeDeletedKeyVaultsInResourceGroup(ctx, resourceGroupName)
-
 	// we want non-conformant clusters to be visible at the end, without impeding our ability to clean up the resource group
 	return nonConformantErr
 }
@@ -745,96 +738,7 @@ func (tc *perItOrDescribeTestContext) cleanupResourceGroupNoRP(ctx context.Conte
 		return fmt.Errorf("failed to cleanup resource group: %w", err)
 	}
 
-	// Purge any Key Vaults left soft-deleted by the resource group deletion so
-	// their globally-unique names are immediately reusable by later runs.
-	tc.purgeDeletedKeyVaultsInResourceGroup(ctx, resourceGroupName)
-
 	return nil
-}
-
-// purgeDeletedKeyVaultsInResourceGroup purges any soft-deleted Key Vaults that
-// belonged to the given resource group. Deleting a resource group only places
-// its vaults into a recoverable (soft-deleted) state, and Azure keeps the
-// globally-unique vault name reserved for the soft-delete retention window (up
-// to 90 days). Because the e2e customer Key Vault name is derived
-// deterministically from the resource group id
-// (cust-kv-${uniqueString(resourceGroup().id, ...)}), a not-yet-purged vault
-// would cause a VaultAlreadyExists collision on a later run that reuses the
-// name. This is best-effort: failures (including a missing
-// Microsoft.KeyVault/locations/deletedVaults/purge/action permission) are
-// logged but never fail cleanup.
-func (tc *perItOrDescribeTestContext) purgeDeletedKeyVaultsInResourceGroup(ctx context.Context, resourceGroupName string) {
-	// Bound the whole best-effort purge so a stuck vault purge or Azure
-	// control-plane stall cannot hang teardown indefinitely.
-	ctx, cancel := context.WithTimeout(ctx, keyVaultPurgeTimeout)
-	defer cancel()
-
-	creds, err := tc.AzureCredential()
-	if err != nil {
-		ginkgo.GinkgoLogr.Error(err, "unable to purge soft-deleted key vaults: failed to get azure credentials", "resourceGroup", resourceGroupName)
-		return
-	}
-	subscriptionID, err := tc.SubscriptionID(ctx)
-	if err != nil {
-		ginkgo.GinkgoLogr.Error(err, "unable to purge soft-deleted key vaults: failed to get subscription id", "resourceGroup", resourceGroupName)
-		return
-	}
-	vaultsClient, err := armkeyvault.NewVaultsClient(subscriptionID, creds, tc.perBinaryInvocationTestContext.getClientFactoryOptions())
-	if err != nil {
-		ginkgo.GinkgoLogr.Error(err, "unable to purge soft-deleted key vaults: failed to build key vault client", "resourceGroup", resourceGroupName)
-		return
-	}
-
-	// Soft-deleted vaults expose their original resource id via VaultID; match
-	// the resource group segment case-insensitively.
-	rgMarker := strings.ToLower("/resourcegroups/" + resourceGroupName + "/")
-
-	pager := vaultsClient.NewListDeletedPager(nil)
-	for pager.More() {
-		page, err := pager.NextPage(ctx)
-		if err != nil {
-			ginkgo.GinkgoLogr.Error(err, "unable to list soft-deleted key vaults", "resourceGroup", resourceGroupName)
-			return
-		}
-		for _, deleted := range page.Value {
-			if deleted == nil || deleted.Name == nil || deleted.Properties == nil ||
-				deleted.Properties.VaultID == nil || deleted.Properties.Location == nil {
-				continue
-			}
-			if !strings.Contains(strings.ToLower(*deleted.Properties.VaultID), rgMarker) {
-				continue
-			}
-			ginkgo.GinkgoLogr.Info("purging soft-deleted key vault",
-				"keyVault", *deleted.Name, "location", *deleted.Properties.Location, "resourceGroup", resourceGroupName)
-			poller, err := vaultsClient.BeginPurgeDeleted(ctx, *deleted.Name, *deleted.Properties.Location, nil)
-			if err != nil {
-				// A 404 means the vault was already purged or its soft-delete
-				// window expired between the list and the purge; that is the
-				// desired end state, so treat it as a no-op rather than noise.
-				if isKeyVaultNotFound(err) {
-					continue
-				}
-				ginkgo.GinkgoLogr.Error(err, "failed to start purge of soft-deleted key vault; a colliding name may block a later run until it is purged or expires",
-					"keyVault", *deleted.Name, "resourceGroup", resourceGroupName)
-				continue
-			}
-			if _, err := poller.PollUntilDone(ctx, &runtime.PollUntilDoneOptions{Frequency: StandardPollInterval}); err != nil {
-				if isKeyVaultNotFound(err) {
-					continue
-				}
-				ginkgo.GinkgoLogr.Error(err, "failed to purge soft-deleted key vault; a colliding name may block a later run until it is purged or expires",
-					"keyVault", *deleted.Name, "resourceGroup", resourceGroupName)
-			}
-		}
-	}
-}
-
-// isKeyVaultNotFound reports whether err is an Azure 404 response, which for a
-// purge means the vault is already gone (already purged or soft-delete window
-// expired) and can be treated as a successful no-op.
-func isKeyVaultNotFound(err error) bool {
-	var respErr *azcore.ResponseError
-	return errors.As(err, &respErr) && respErr.StatusCode == http.StatusNotFound
 }
 
 func (tc *perItOrDescribeTestContext) collectDebugInfoForResourceGroup(ctx context.Context, resourceGroupName string) error {
--- a/test/util/framework/per_test_framework.go
+++ b/test/util/framework/per_test_framework.go
@@ -45,8 +45,10 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	armauthorization "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v3"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5"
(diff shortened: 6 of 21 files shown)


bennerv:
Can you add an e2e test that creates a nodepool with os disk encryption set, and uses an azure client to confirm they're actually set when the node pools are created?

internal/ocm/convert.go
```diff
@@ -188,6 +188,17 @@ func convertEnableEncryptionAtHostToCSBuilder(in api.NodePoolPlatformProfile) *a
...
+func buildCSOsDisk(osDisk api.OSDiskProfile, storageAccountType, persistence string) *arohcpv1alpha1.AzureNodePoolOsDiskBuilder {
+	builder := arohcpv1alpha1.NewAzureNodePoolOsDisk().
+		SizeGibibytes(int(*osDisk.SizeGiB)).
+		StorageAccountType(storageAccountType).
+		Persistence(persistence)
+	if osDisk.EncryptionSetID != nil {
+		builder.SseEncryptionSetResourceId(osDisk.EncryptionSetID.String())
+	}
+	return builder
+}
```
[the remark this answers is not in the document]
cadenmarchese: This is carry-over from current behavior. I am not familiar enough with the HCP API to know if this truly needs a nil check or not, but I guess it can't hurt. Deferring to others for input.

test/e2e/nodepool_osdisk_encryption.go
```diff
@@ -0,0 +1,190 @@
...
+			Expect(err).NotTo(HaveOccurred(), "failed to create cluster customer resources")
+
+			By("creating a KeyVault and DES for OS disk encryption")
+			subscriptionID, err := tc.SubscriptionID(ctx)
+			Expect(err).NotTo(HaveOccurred(), "failed to get subscription ID")
+
+			desResourceID := fmt.Sprintf(
+				"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/diskEncryptionSets/%s-des",
+				subscriptionID, *resourceGroup.Name, customerClusterName,
+			)
```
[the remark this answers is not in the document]
cadenmarchese: I'm going to let e2e run first to identify any other potential issues before proceeding with fixing this.

test/util/framework/deployment_params.go
```diff
@@ -280,6 +280,7 @@ type NodePoolParams20240610 struct {
 	// AutoScaling enables nodepool autoscaling. When set, Replicas is ignored.
 	AutoScaling      *NodePoolAutoScalingParams
 	AvailabilityZone string
+	EncryptionSetID  string
 }
```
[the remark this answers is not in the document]
cadenmarchese: Not super necessary. If other reviewers disagree I will revisit.

test/e2e/nodepool_osdisk_encryption.go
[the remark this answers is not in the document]
cadenmarchese: Fixed

internal/ocm/convert.go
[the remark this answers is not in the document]
miguelsorianod: I think current behavior around this should be fine:
In the per-version api go types of nodepools, when SizeGiB is nil they default to a default value. This means under normal circumstances (nodepool creation via RP frontend) that field should not be nil in the internal api type. If it's nil for some reason it would be a programming error and this panicking should be fine.

test/e2e/nodepool_osdisk_encryption.go
```diff
@@ -0,0 +1,197 @@
...
+
+			By("verifying each VM's OS disk has EncryptionAtRestWithCustomerKey")
+			disksClient := computeFactory.NewDisksClient()
+			for _, vm := range workerVMs {
+				verifyVMOSDiskCustomerEncryption(ctx, disksClient, managedResourceGroupName, vm, desResourceID)
+			}
+		})
+})
+
+func verifyVMOSDiskCustomerEncryption(
```
miguelsorianod: nitpick: no need to newline parameters. We accept long lines in the project

test/e2e/nodepool_osdisk_encryption.go
```diff
@@ -0,0 +1,197 @@
...
+	ctx context.Context,
+	disksClient *armcompute.DisksClient,
+	managedResourceGroup string,
+	vm *armcompute.VirtualMachine,
+	expectedDESResourceID string,
+) {
+	vmName := to.Ptr("<unknown>")
```
miguelsorianod: why defaulting to unknown? let's set an expect like it's done for other attributes here

miguelsorianod:
What are the code changes themselves look good to me.

About the e2e, without being familiar with the functionality on Azure side, the test conceptually seems correct. I will let others more familiar with it give their view.

I left some minor comments.

test/e2e/nodepool_osdisk_encryption.go
cadenmarchese: done, thanks!

test/e2e/nodepool_osdisk_encryption.go
cadenmarchese: done, thanks!

test/util/framework/deployment_params.go
```diff
@@ -336,6 +336,7 @@ type NodePoolParams20240610 struct {
 	AutoScaling      *NodePoolAutoScalingParams
 	AvailabilityZone string
 	Tags             map[string]*string
+	EncryptionSetID  string
 }
```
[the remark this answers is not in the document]
cadenmarchese: duplicate comment.

test/util/framework/hcp_helper.go
```diff
@@ -1076,6 +1076,10 @@ func BuildNodePoolFromParams20240610(
 		},
 	}
 
+	if parameters.EncryptionSetID != "" {
+		nodePool.Properties.Platform.OSDisk.EncryptionSetID = to.Ptr(parameters.EncryptionSetID)
+	}
```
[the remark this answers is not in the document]
cadenmarchese: duplicate comment.

cadenmarchese:
The CI custom role was updated to accommodate this e2e test. Context:
```
 Nodepool OS Disk Encryption should create a nodepool with customer-managed disk encryption via DES expand_less | 23m40s | {fail [github.com/Azure/ARO-HCP/test/e2e/nodepool_osdisk_encryption.go:134]: failed to create nodepool des-np with DES Unexpected error:     <*fmt.wrapError \| 0xc0015ba6c0>:      failed to create NodePool des-np: failed waiting for nodepool="des-np" for cluster "des-encrypt" in resourcegroup="des-encrypt-6jl8pq" to finish creating: GET https://rp.j6608000.hcpsvc.osadev.cloud/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationStatuses//XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX     --------------------------------------------------------------------------------     RESPONSE 200: 200 OK     ERROR CODE: InternalServerError     --------------------------------------------------------------------------------     {       "id": "/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/providers/Microsoft.RedHatOpenShift/locations/westus3/hcpOperationStatuses//XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",       "name": "/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",       "status": "Failed",       "startTime": "2026-07-16T22:00:30.305420763Z",       "endTime": "2026-07-16T22:00:45.224060261Z",       "error": {         "code": "InternalServerError",         "message": "[clusterServiceNodePoolStatus] GET https://management.azure.com/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/des-encrypt-6jl8pq/providers/Microsoft.Compute/diskEncryptionSets/des-encrypt-des-o3nbqnh7nwvje\n--------------------------------------------------------------------------------\nRESPONSE 403: 403 Forbidden\nERROR CODE: AuthorizationFailed\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"AuthorizationFailed\",\n    \"message\": \"The client '/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX' with object id '/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX does not have authorization to perform action 'Microsoft.Compute/diskEncryptionSets/read' over scope '/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/des-encrypt-6jl8pq/providers/Microsoft.Compute/diskEncryptionSets/des-encrypt-des-o3nbqnh7nwvje' or the scope is invalid. If access was recently granted, please refresh your credentials.\"\n  }\n}\n--------------------------------------------------------------------------------\n"       }     }     --------------------------------------------------------------------------------
```

test/e2e-setup/bicep/modules/disk-encryption-set.bicep
```diff
@@ -0,0 +1,86 @@
...
+var desKeyVaultName = 'des-kv-${randomSuffix}'
+
+resource desKeyVault 'Microsoft.KeyVault/vaults@2024-12-01-preview' = {
+  name: desKeyVaultName
+  location: resourceGroup().location
+  properties: {
+    enableRbacAuthorization: true
+    enableSoftDelete: true
+    enablePurgeProtection: true
+    tenantId: subscription().tenantId
```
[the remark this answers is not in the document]
cadenmarchese: Per my commit message, purge protection is required for this scenario.

mbarnes:
/lgtm

for the `/internal/ocm` bits

miguelsorianod:
non e2e changes lgtm

miguelsorianod:
/lgtm

for the non e2e changes part

test/e2e/nodepool_osdisk_encryption.go
```diff
@@ -0,0 +1,210 @@
...
+
+			By("verifying each VM's OS disk has EncryptionAtRestWithCustomerKey")
+			disksClient := computeFactory.NewDisksClient()
+			for _, vm := range workerVMs {
+				verifyVMOSDiskCustomerEncryption(ctx, disksClient, managedResourceGroupName, vm, desResourceID)
+			}
+		})
+})
+
+func verifyVMOSDiskCustomerEncryption(ctx context.Context, disksClient *armcompute.DisksClient, managedResourceGroup string, vm *armcompute.VirtualMachine, expectedDESResourceID string) {
```
stevekuznetsov: if you can, `cmp.Diff()` is a much easier and more straightforward way to say "the shape of this look like that"

test/e2e-setup/bicep/modules/disk-encryption-set.bicep
```diff
@@ -0,0 +1,86 @@
...
+    sku: {
+      family: 'A'
+      name: 'standard'
+    }
+  }
+}
```
mbukatov: This looks as if we are deploying another key vault instance here. Correct me if I'm wrong. The rest of the comment assumes that it is the case.

We should not create a new key vault, but use existing key vault deployed as part of customer infra in customer-infra.bicep. Why? I expect a customer will create a single key vault instance to manage their keys for ARO HCP hosted cluster.

The solution here is to integrate this into existing customer-infra.bicep file, so that one can enable this os disk encryption by passing few extra variables into `tc.CreateClusterCustomerResources2024061()` function. If you would like to keep the implementation in a separate bicep file, that is possible, just use it from the customer-infra.bicep.

Or is there any reason for a customer to be required to create two separate key vault instances so that some keys are managed in one and others in the 2nd instance?

test/e2e/nodepool_osdisk_encryption.go
```diff
@@ -0,0 +1,210 @@
...
+				framework.WithTemplateFromFS(TestArtifactsFS, "test-artifacts/generated-test-artifacts/modules/disk-encryption-set.json"),
+				framework.WithDeploymentName(fmt.Sprintf("des-%s", customerClusterName)),
+				framework.WithScope(framework.BicepDeploymentScopeResourceGroup),
+				framework.WithClusterResourceGroup(*resourceGroup.Name),
+				framework.WithParameters(map[string]interface{}{
+					"clusterName":          customerClusterName,
+					"serviceMiPrincipalId": *serviceMI.Properties.PrincipalID,
+				}),
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create disk encryption set")
```
mbukatov, line 100: My [note about bicep module organization](https://github.com/Azure/ARO-HCP/pull/5682/changes#r3645167831) has impact here as well.

test/e2e/nodepool_osdisk_encryption.go
```diff
@@ -0,0 +1,210 @@
...
+
+			workerVMs := filterNodePoolVMs(vms, customerNodePoolName)
+			By(fmt.Sprintf("found %d VMs for nodepool %s", len(workerVMs), customerNodePoolName))
+			Expect(workerVMs).ToNot(BeEmpty(), "expected at least one VM for nodepool %s", customerNodePoolName)
+
+			By("verifying each VM's OS disk has EncryptionAtRestWithCustomerKey")
+			disksClient := computeFactory.NewDisksClient()
+			for _, vm := range workerVMs {
+				verifyVMOSDiskCustomerEncryption(ctx, disksClient, managedResourceGroupName, vm, desResourceID)
+			}
```
mbukatov: Would it make sense to make this check into a separate verifier?

mbukatov (commented):
Providing feedback on E2E test.

test/e2e/nodepool_osdisk_encryption.go
mbukatov: When you have the verifier, it would be nice to do this test to check how error reporting works:

- in a separate commit, add the verifier into another test which doesn't use customer managed os disk encryption
- trigger a e2e test run
- see how it fails (and if the error looks clear)
- remove that commit

test/e2e-setup/bicep/modules/disk-encryption-set.bicep
cadenmarchese: @mbukatov thanks, a question before I proceed.

We're not reusing the existing cluster keyvault here because that one explicitly sets `enablePurgeProtection: false`, and this feature requires that to be `true`. Currently for the cluster keyvault, `purgeDeletedKeyVaultsInResourceGroup` seems to be cleaning / purging the cluster keyvaults as needed, but if we change cluster keyvault to `enablePurgeProtection: true` for all e2e runs, I assume it will break e2e cleanup.

From a customer's perspective, re-using a key from the cluster keyvault would work fine, but it wouldn't work in our e2e because of this limitation.

This means that in order to have e2e for this feature, we would need to be OK with having these keyvaults in a soft deleted state until such time as they are purged by Azure. It's probably OK for the purposes of this test, since this separate keyvault gets a unique name, so there shouldn't be name collisions.

Let me know what you think.

mbukatov: It would make sense to parametrize the `enableSoftDelete` property of the `customerKeyVault` resource in `customer-infra.bicep` file with `false` as a default, so that the other E2E test cases can use the same config as before, but the node pool OS disk encryption test can set it to `true` as needed. This way, we will be testing a scenario closer to what a customer would do.

As far as the cleanup is concerned, I assumed that the `purgeDeletedKeyVaultsInResourceGroup()` function can delete even a key vault with soft delete and purge protection. If that is not the case, it will make sense to update the cleanup to handle this:

- Check if we can actually get the keyvault with `enablePurgeProtection` deleted during cleanup, and if not, at least log it without failing the cleanup
- Limit the impact of accumulated soft deleted key vaults and the resulting conflict and set `softDeleteRetentionInDays: 7` in for the `customerKeyVault` resource in `customer-infra.bicep`

Moreover you won't avoid that cleanup work, since the separate key vault will also have to be eventually deleted.

cadenmarchese: Unfortunately, we can't freely enable/disable purge protection on a keyvault for the purposes of this test only. Once purge protection is enabled, it can't be disabled again. Once enabled, purgeDeletedKeyVaultsInResourceGroup() won't clean it up (that function is for soft deleted keyvaults without purge protection). Given that, I'll go ahead with your suggestion to maintain the separate keyvault but set softDeleteRetentionInDays: 7 on it so that it's automatically cleaned up in Azure. I think that this is the best approach to ensure that this test is accurate to the BYOK UX and doesn't impact other test assumptions.

test/e2e/nodepool_osdisk_encryption.go
cadenmarchese: I hope it's OK, but I stuck with Gomega since I moved this whole function into util/verifiers which seems to stick to that convention for more basic assertions.

test/e2e/nodepool_osdisk_encryption.go
cadenmarchese: Added - I'll ensure everything passes first, and then add your suggested test commit.

cadenmarchese: Here are the logs from that test: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/5682/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2081754856711786496

test/e2e/nodepool_ephemeral_osdisk.go
```diff
@@ -196,6 +196,12 @@ var _ = Describe("Nodepool Ephemeral OS Disk", func() {
 			for _, vm := range workerVMs {
 				verifyVMHasEphemeralOSDisk(vm)
 			}
+
+			By("verifying each VM's OS disk has EncryptionAtRestWithCustomerKey (expected to fail: no DES configured)")
+			disksClient := computeFactory.NewDisksClient()
+			for _, vm := range workerVMs {
+				verifiers.VerifyVMOSDiskCustomerEncryption(ctx, disksClient, managedResourceGroupName, vm, "fake-des-resource-id")
+			}
```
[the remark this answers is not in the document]
cadenmarchese: This was intentional, the ask was to simulate a failure to confirm behavior was as intended: https://github.com/Azure/ARO-HCP/pull/5682#discussion_r3645236358

test/e2e-setup/bicep/modules/disk-encryption-set.bicep
mbukatov: > Unfortunately, we can't freely enable/disable purge protection on a keyvault for the purposes of this test only.

Understood.

>  Given that, I'll go ahead with your suggestion to maintain the separate keyvault but set softDeleteRetentionInDays: 7 on it so that it's automatically cleaned up in Azure. I think that this is the best approach to ensure that this test is accurate to the BYOK UX and doesn't impact other test assumptions.

My suggestion was to use a single key vault (as already present in the customer infra bicep file), but set it's enablePurgeProtection to true and set softDeleteRetentionInDays to 7 days (the minimum if I recall right) so that it could be used in the node pool os disk use case as well as other test cases.

Creating a separate key vault is not a good approach. In both cases (single vs 2 key vautls per test), we will end up with one key vault which won't be fully deleted after the test, but with 2 key vaults per test, we will deviate from the customer's use case and expectations about the bicep setup files as currently used.

test/util/verifiers/osdisk_encryption.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+	"strings"
+
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5"
+)
+
+func VerifyVMOSDiskCustomerEncryption(ctx context.Context, disksClient *armcompute.DisksClient, managedResourceGroup string, vm *armcompute.VirtualMachine, expectedDESResourceID string) {
```
mbukatov: Please use [HostedClusterVerifier interface](https://github.com/Azure/ARO-HCP/blob/f27abd776540093384f07e94c39caf913fb90c69/test/util/verifiers/basic.go#L29) so that:

- this verifier can be executed in the same way as any other verifiers
- this verifier can be initialized and passed as an additional verifier into `verifiers.VerifyHCPCluster()` function call (so that it will execute all the verifiers at once in parallel, and report on the result)

An example of expected usage (I'm guessing a parameters you will pass to the VerifyVMOSDiskCustomerEncryption for the purpose of this demonstration here):

```
err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyNodesReady(), verifiers.VerifyVMOSDiskCustomerEncryption(managedResourceGroup, expectedDESResourceID))
```

That will then run a set of default verifiers as well as VerifyNodesReady and VerifyVMOSDiskCustomerEncryption in parallel, and report the combined failures in resulting `err` value.

This is useful because:

- we speed up a test runtime a bit by running these verifications in parallel
- we get more context when something fails (imagine an error which makes 2 verifiers to fail because something is fundamentally wrong with nodes in the node pool)

mbukatov:
Also note [my comment on the key vault](https://github.com/Azure/ARO-HCP/pull/5682#discussion_r3665462405) (it could be easily missed in a resolved conversation)

test/e2e-setup/bicep/modules/disk-encryption-set.bicep
cadenmarchese: We can do it that way, just noting that this also involves us removing all the cleanup logic around key vault cleanup (such as `purgeDeletedKeyVaultsInResourceGroup`)

test/util/verifiers/osdisk_encryption.go
cadenmarchese: Implemented in https://github.com/Azure/ARO-HCP/pull/5682/commits/657052dac79d83a1f97ff422addee64892c0df09

test/e2e-setup/bicep/modules/disk-encryption-set.bicep
cadenmarchese: Implemented in https://github.com/Azure/ARO-HCP/pull/5682/commits/3cbaa946f1d2a587f1ef848521d3696d58bff3fd

test/e2e/nodepool_osdisk_encryption.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("Nodepool OS Disk Encryption", func() {
+	It("should create a nodepool with customer-managed disk encryption via DES",
+		labels.RequireNothing,
+		labels.Critical,
+		labels.Positive,
+		labels.AroRpApiCompatible,
```
mgahagan73, line 38: Now that the MI Container aware scheduler has merged, you will need to add a label depicting how many Managed Identity containers this test will use. e.g. labels.MIContainers(n) where n is the number of MI Containers. Please also rebase to pick up the container checks that were added to make verify to avoid any last minute surprises.

mbukatov:
will recheck on Monday

test/e2e-setup/bicep/modules-4.19/customer-infra.bicep
```diff
@@ -75,7 +75,9 @@ resource customerKeyVault 'Microsoft.KeyVault/vaults@2024-12-01-preview' = {
   location: resourceGroup().location
   properties: {
     enableRbacAuthorization: true
-    enableSoftDelete: false
+    enableSoftDelete: true
+    enablePurgeProtection: true
+    softDeleteRetentionInDays: 7
```
mgahagan73: What is the quota for keyvaults in a given sub/region? Do we need to have something in place to monitor/purge soft-deleted keys (and other resources) so we don't end up in a situation where CI runs start failing because we exhaust quota? We don't necessarily have to address it in this PR, I want to make sure we document any additional work we need to do in the short term to prevent CI reliability issues.

cadenmarchese: @mgahagan73 thanks for your review. There is no quota for keyvaults in a subscription. It doesn't necessarily mean that it's unlimited, since there are still resource-wide limitations in a resource group, but since these are going into distinct resource groups and being deleted immediately after use, we should never hit those limits.

It's worth noting that 7 days is the shortest we can set that `softDeleteRetentionInDays` value, and enablePurgeProtection and enableSoftDelete are required for the feature to work.

mbukatov: Purpose of this 4.19 bicep module directory is to keep the cluster setup from the time we targeted OCP 4.19 frozen, and so we don't need (and should not) update it with new features.

Updating `test/e2e-setup/bicep/modules/customer-infra.bicep` is enough.

test/e2e-setup/bicep/modules/customer-infra.bicep
```diff
@@ -99,7 +99,9 @@ resource customerKeyVault 'Microsoft.KeyVault/vaults@2024-12-01-preview' = {
   location: resourceGroup().location
   properties: {
     enableRbacAuthorization: true
-    enableSoftDelete: false
+    enableSoftDelete: true
+    enablePurgeProtection: true
```
mbukatov: Could we parametrize these (`enableSoftDelete` as well as `enablePurgeProtection`) so that:

- the value is not hard coded here, but we use a new bicep module parameter(see for example `param persistTagValue bool = false`)
- the default value of the new bicep parameter is false (so that we keep the previous behaviour for all the other E2E tests)
- the disk encryption test can redefine the new parameter(s) from the outside (via `infraParameters` argument of `tc.CreateClusterCustomerResources*()` function)

This is what I meant in my previous review, sorry for not explaining this better.

test/e2e-setup/bicep/modules/disk-encryption-set.bicep
mbukatov: Thank you, this is the reuse I had in mind.

test/util/framework/helpers_v20240610preview.go
```diff
@@ -92,6 +92,7 @@ type NodePoolParams20240610 struct {
 	AutoScaling      *NodePoolAutoScalingParams
 	AvailabilityZone string
 	Tags             map[string]*string
+	EncryptionSetID  string
```
mbukatov, line 96: Are you sure we want to extend structure of 20240610 API? I assumed that older APIs are frozen, and that the new features are included in the new API level.

test/util/framework/per_invocation_framework.go
```diff
@@ -81,11 +81,6 @@ var (
 const (
 	StandardPollInterval            = 10 * time.Second
 	StandardResourceGroupExpiration = 4 * time.Hour
-	// keyVaultPurgeTimeout bounds the best-effort soft-deleted Key Vault purge
-	// on resource-group teardown so a stalled purge (or Azure control-plane
-	// stall) can never hang the cleanup goroutine and flake the overall E2E
-	// teardown.
-	keyVaultPurgeTimeout = 5 * time.Minute
```
mbukatov, line 88: Do we really want to drop this?

test/util/framework/per_test_framework.go
```diff
@@ -745,98 +738,9 @@ func (tc *perItOrDescribeTestContext) cleanupResourceGroupNoRP(ctx context.Conte
...
-}
-
-// isKeyVaultNotFound reports whether err is an Azure 404 response, which for a
-// purge means the vault is already gone (already purged or soft-delete window
-// expired) and can be treated as a successful no-op.
-func isKeyVaultNotFound(err error) bool {
-	var respErr *azcore.ResponseError
-	return errors.As(err, &respErr) && respErr.StatusCode == http.StatusNotFound
-}
-
```
mbukatov, line 839: I don't think we want to entirely drop this. I assumed that we will detect and skip deletion of purge protection key vaults (since as you noted, we can't delete them), but actually delete the keyvaults without this protection (basically all the others).

mbukatov (commented):
Checking updates in E2E test code.

test/util/framework/per_test_framework.go
cadenmarchese: Are there other keyvaults in-scope for cleanup, apart from this customer KV?

internal/ocm/convert.go
```diff
@@ -224,6 +224,17 @@ func convertEnableEncryptionAtHostToCSBuilder(in api.NodePoolPlatformProfile) *a
 	return arohcpv1alpha1.NewAzureNodePoolEncryptionAtHost().State(state)
 }
 
+func buildCSOsDisk(osDisk api.OSDiskProfile, storageAccountType, persistence string) *arohcpv1alpha1.AzureNodePoolOsDiskBuilder {
+	builder := arohcpv1alpha1.NewAzureNodePoolOsDisk().
+		SizeGibibytes(int(*osDisk.SizeGiB)).
+		StorageAccountType(storageAccountType).
+		Persistence(persistence)
```
bennerv, line 231: if the persistence is ephemeral, can you configure a disk encryption set on it?

test/e2e-setup/bicep/modules-4.19/customer-infra.bicep
bennerv: why are we adding these?

test/e2e-setup/bicep/modules/disk-encryption-set.bicep
```diff
@@ -0,0 +1,77 @@
+@description('Cluster name used to ensure unique resource names within the resource group')
```
bennerv: why not just do this in code rather than bicep?

test/util/framework/helpers_v20240610preview.go
bennerv, line 96: should be added to every API version, not just this 20240610 one.

test/util/framework/per_test_framework.go
```diff
@@ -745,98 +738,9 @@ func (tc *perItOrDescribeTestContext) cleanupResourceGroupNoRP(ctx context.Conte
...
-	}
-}
-
-// isKeyVaultNotFound reports whether err is an Azure 404 response, which for a
-// purge means the vault is already gone (already purged or soft-delete window
-// expired) and can be treated as a successful no-op.
-func isKeyVaultNotFound(err error) bool {
-	var respErr *azcore.ResponseError
-	return errors.As(err, &respErr) && respErr.StatusCode == http.StatusNotFound
-}
```
bennerv, line 838: Why are we modifying this?  The `cleanup-sweeper` is a separate job that runs occasionally and will free up these key vault names if the e2e test stops suddenly or falls over without the cleanup step executing.

raelga (changes requested):
/hold

Heads up: this PR is deleting a bunch of unrelated cleanup code that already lives on `main`. It looks like a bad rebase/merge, not an intentional change.

The culprit is commit a916e20ae ("feat: add e2e test for BYOK OS Disks"). It reintroduces a stale, pre-purge copy of `test/util/framework/per_test_framework.go` and wipes several cleanup-sweeper files that `main` already has:

- `test/util/framework/per_test_framework.go`: removes `purgeDeletedKeyVaultsInResourceGroup` + `isKeyVaultNotFound` and their two call sites (-96) from #6176
- `tooling/cleanup-sweeper/pkg/engine/steps/keyvault/purge_deleted.go` (-177) and `role_assignments_sweeper.go` (-27) from #4582
- `tooling/cleanup-sweeper/pkg/engine/steps/keyvault/purge_orphaned_deleted.go` (-206) from #6177
- plus the matching `_test.go` files and `resourcegroup_ordered_cleanup.go`

Evidence it's a rebase artifact, not intentional: the purge commit 54d605122 is an ancestor of both your branch head and the merge-base with `main` (1078d7868), so your branch history contained this code and a later commit on the branch reverted it. All of these files still exist on current `upstream/main`.

Why it matters: that purge logic prevents `VaultAlreadyExists` collisions on e2e reruns (deleting an RG only soft-deletes its Key Vaults, and the customer KV name is derived deterministically from the RG id). Dropping it reintroduces that flake class and removes live sweeper functionality.

Could you rebase again and keep those files from `main`? Your actual BYOK DES changes (`disk-encryption-set.bicep`, `nodepool_osdisk_encryption.go`, verifiers, etc.) look fine, it's just the framework/sweeper files that got clobbered. Holding until that's sorted.

internal/ocm/convert.go
cadenmarchese: Yes, [hypershift explicitly tests this](https://github.com/openshift/hypershift/blob/8c9a9b95edc69c9a37e64e8d9f4b903b2cdc22d2/hypershift-operator/controllers/nodepool/azure_test.go#L266-L296) and it's also doable in AKS.

test/util/framework/helpers_v20240610preview.go
cadenmarchese: It should be in this helper file since the v2024* API supports a disk encryption set on the nodepool just like the newer versions, but it looks like I should also add this to the helper files for the other newer API versions as well.

test/e2e-setup/bicep/modules-4.19/customer-infra.bicep
cadenmarchese: I can revert the modules-4.19 changes. In general though `enablePurgeProtection: true` and `enableSoftDelete: true` are required for a DES+BYOK to work, otherwise Azure will reject the DES.

test/util/framework/per_test_framework.go
cadenmarchese: Revisiting this, I had added this removal previously because I assumed that we used the same keyvault across all e2e tests (meaning that the cleanup wouldn't work anymore as part of my change). Now that I see this was not correct and we use 1 keyvault per test case, I have reverted this.

test/util/framework/per_test_framework.go
cadenmarchese: I had added this removal previously because I assumed that we used the same keyvault across all e2e tests (meaning that the cleanup wouldn't work anymore as part of my change). Now that I see this was not correct and we use 1 keyvault per test case, I have reverted this.

test/util/framework/per_invocation_framework.go
cadenmarchese: I had added this removal previously because I assumed that we used the same keyvault across all e2e tests (meaning that the cleanup wouldn't work anymore as part of my change). Now that I see this was not correct and we use 1 keyvault per test case, I have reverted this so that it can continue sweeping keyvaults that are not part of my test.

internal/ocm/convert.go
cadenmarchese: Resolving this, but I wouldn't be opposed to a separate e2e case for this at some point, if we feel the need to confirm that this works as intended across the whole stack.

cadenmarchese:
@raelga the change you mentioned was intentional but was due to a misunderstanding. I removed the sweeper because I assumed that we used the same keyvault across all e2e tests (meaning that the sweeper wouldn't function anymore as part of my purge protection changes). Now that I see this was not correct and we use 1 keyvault per test case, I have reverted this so that it can continue sweeping keyvaults that are not part of my test.

test/e2e-setup/bicep/modules/customer-infra.bicep
```diff
@@ -99,7 +108,9 @@ resource customerKeyVault 'Microsoft.KeyVault/vaults@2024-12-01-preview' = {
   location: resourceGroup().location
   properties: {
     enableRbacAuthorization: true
-    enableSoftDelete: false
+    enableSoftDelete: enableKeyVaultSoftDelete ? true : null
+    enablePurgeProtection: enableKeyVaultPurgeProtection ? true : null
+    softDeleteRetentionInDays: enableKeyVaultSoftDelete ? keyVaultSoftDeleteRetentionInDays : null
```
bennerv, line 113: why are we changing this?

test/e2e/nodepool_osdisk_encryption.go
```diff
@@ -0,0 +1,164 @@
...
+				Expect(err).NotTo(HaveOccurred(), "failed to assign pooled identity containers")
+			}
+
+			By("creating a resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "des-encrypt", tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create resource group")
+
+			By("creating cluster parameters")
+			clusterParams := framework.NewDefaultClusterParams20240610()
+			clusterParams.ClusterName = customerClusterName
```
bennerv, line 59: nit: use the 2025 api version.

test/e2e-setup/bicep/modules/customer-infra.bicep
cadenmarchese: We need to omit the values if they are false, otherwise Azure still complains:
```
    RESPONSE 200: 200 OK
    ERROR CODE: DeploymentFailed
    {
      "status": "Failed",
      "error": {
        "code": "DeploymentFailed",
        "message": "At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-deployment-operations for usage details.",
        "details": [
          {
            "code": "BadRequest",
            "message": "{\r\n  \"error\": {\r\n    \"code\": \"BadRequest\",\r\n    \"message\": \"The property \\\"enablePurgeProtection\\\" cannot be set to false. Enabling the purge protection for a vault is an irreversible action.\"\r\n  }\r\n}"
          }
        ]
      }
    }
```

test/e2e/nodepool_osdisk_encryption.go
cadenmarchese: Sure, but why not bump up to a 2026 if we're already bumping it up?

cadenmarchese: bumped in b2de3db98260b45a767206335f702dc9f124959b

internal/ocm/convert.go
bennerv: > Yes, hypershift explicitly tests this and it's also doable in AKS.

This is a unit test, not an actual e2e test.  The unit test isn't exercising the integration with Azure.  

https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks#unsupported-features
> Updating a CMK key version or key rotation of customer managed key isn't supported with Ephemeral OS disk. VMs using Ephemeral OS disks need to be deleted before updating or rotating the keys and can be re-created later.

Seems like a fragile area.

bennerv: Just tested it.  It still works 🤷 .  Resolving.

test/util/verifiers/osdisk_encryption.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	var errs []error
+	for _, vm := range workerVMs {
+		if err := v.verifyVM(ctx, disksClient, vm); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) > 0 {
+		return fmt.Errorf("OS disk encryption verification failed for %d/%d VMs: %v", len(errs), len(workerVMs), errs)
```
mbukatov: Instead of passing list of errors `errs` via `%v`, join the errors together as `errors.Join(errs...)` and pass it as `%w`, that would preserve more context about the errors.

test/util/framework/per_test_framework.go
```diff
@@ -745,98 +738,9 @@ func (tc *perItOrDescribeTestContext) cleanupResourceGroupNoRP(ctx context.Conte
...
-// its vaults into a recoverable (soft-deleted) state, and Azure keeps the
-// globally-unique vault name reserved for the soft-delete retention window (up
-// to 90 days). Because the e2e customer Key Vault name is derived
-// deterministically from the resource group id
-// (cust-kv-${uniqueString(resourceGroup().id, ...)}), a not-yet-purged vault
-// would cause a VaultAlreadyExists collision on a later run that reuses the
-// name. This is best-effort: failures (including a missing
-// Microsoft.KeyVault/locations/deletedVaults/purge/action permission) are
-// logged but never fail cleanup.
-func (tc *perItOrDescribeTestContext) purgeDeletedKeyVaultsInResourceGroup(ctx context.Context, resourceGroupName string) {
```
mbukatov, line 766: I still see removal of the key vault cleanup in the PR, is this intentional?

mbukatov (commented):
Thank you for updating the E2E test code, now the test code is aligned with the existing code.

I have 2 comments though:

- one minor note about improvements in error reporting in the new verifier
- I still see the removal of key vault cleanup in the PR, I assumed we want to preserve it

test/util/verifiers/osdisk_encryption.go
cadenmarchese: Done, let me know if this works.

test/util/framework/per_test_framework.go
cadenmarchese: Sorry about that - I didn't review the diff and missed that this was still on my branch. It's fixed now.

mgahagan73:
I've looked through the changes under /test and I don't see any issues.

/lgtm

PR #5683, opened 2026-06-16
fix: Add npm ci target and dependency on all npm run commands 

bennerv:
No jira

### What

Adds an `npm ci` dependency on any `npm run` commands as we need the dependencies installed in order to succeed first.

### Why

if dependencies are updated, this will force the user to have the latest dependencies for generation.  

### Testing

api validation should cover this from a prowjob
### PR Checklist
- [x] Self-reviewed the diff

### DIFF
--- a/api/.gitignore
+++ b/api/.gitignore
@@ -1,0 +1,1 @@
+.npm-install.stamp
--- a/api/Makefile
+++ b/api/Makefile
@@ -24,12 +24,12 @@ endif
 generate: swagger models
 
 .PHONY: swagger
-swagger:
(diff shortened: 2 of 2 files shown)


tsatam (approved):

bennerv:
/close

This is a good change, but the way we're building images for e2e in CI openshift/release is not right so this isn't ready to go.  I'd advise at some point doing: 
- build an image with all the tools we need in openshift/release
- do the go/node/etc. specific things for the PR in question (dependencies specific to the PR rather)
- then run `npm ci` during the verification tests that run in a job.

PR #5684, opened 2026-06-16
feature: adding GA application id to config

rachelvweber:
[ARO-27769](https://redhat.atlassian.net/browse/ARO-27769?atlOrigin=eyJpIjoiYTc5NmJkYTMwYmM0NDY0MzgyZDU0YjNiY2U1MTgyMDEiLCJwIjoiaiJ9)

### What

Adding empty app id to geneva action config

### Why

This unblocks reading geneva actions app id from config for generated geneva actions xmls. The placeholder value will be replaced with a real value in sdp-pipelines

### Testing

N/A just adding an empty value to config.

### PR Checklist
- [X] PR is scoped to a single task (no mixed concerns)
- [X] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [X] Summary explains the "Why" behind the change
- [X] Linked to relevant ticket/issue
- [X] Screenshots included (if graph/UI/metrics changes)
- [X] Self-reviewed the diff
- [X] Draft PR used for WIP (if applicable)
- [X] Tricky code blocks are commented
- [X] All comment threads resolved before merge

### DIFF
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -41,6 +41,9 @@
           "type": "string",
           "description": "The name of the application. Needs to be unique within the tenant. https://eng.ms/docs/microsoft-security/identity/auth/auth-sts-protocols/ests/ests-troubleshooting-guides/tsgs/3p-apps-build-out-automation/gotchas-and-known-issues"
         },
+        "applicationId": {
+          "type": "string"
+        },
         "useSNI": {
           "type": "boolean"
         },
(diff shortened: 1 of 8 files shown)


bennerv:
/lgtm

Talked with Rachel, the reason we're introducing this in our config (even though it's deployed as part of our pipelines) is because the GA extension is published independently of the deployment into prod and won't have access to the bicep to lookup the appid.  

Eventually it would be useful to remove this value and chain it to a bicep lookup but for now this will suffice.

PR #5686, opened 2026-06-16
Switch ServiceProvider reads in controllers to cached listers

deads2k:
Consumer controllers used to call database.GetOrCreateServiceProviderCluster / GetOrCreateServiceProviderNodePool on every sync, which both hit Cosmos and scattered the create path across many sites. Replace those calls with cached ServiceProvider{Cluster,NodePool}Lister Gets; when the document is absent the syncer returns nil and is re-enqueued by the existing ServiceProvider* informer event handlers once the document arrives.

A single dedicated controller per kind owns creation: CreateServiceProviderCluster and CreateServiceProviderNodePool consult the lister cache first and call GetOrCreate only on a miss, so the create-with-conflict-fallback pattern lives in exactly one place per resource.

### DIFF
--- a/backend/pkg/controllers/create_service_provider_cluster_controller.go
+++ b/backend/pkg/controllers/create_service_provider_cluster_controller.go
@@ -22,7 +22,6 @@ import (
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
 	"github.com/Azure/ARO-HCP/backend/pkg/informers"
 	"github.com/Azure/ARO-HCP/backend/pkg/listers"
-	controllerutil "github.com/Azure/ARO-HCP/internal/controllerutils"
 	"github.com/Azure/ARO-HCP/internal/database"
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
@@ -37,7 +36,6 @@ import (
 // have the document in hand to validate the request before any controller
 // has a chance to run.
 type createServiceProviderClusterSyncer struct {
-	cooldownChecker              controllerutil.CooldownChecker
 	resourcesDBClient            database.ResourcesDBClient
 	clusterLister                listers.ClusterLister
 	serviceProviderClusterLister listers.ServiceProviderClusterLister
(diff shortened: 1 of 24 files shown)


stevekuznetsov (approved):
/lgtm
/approve

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -343,26 +358,38 @@ func (c *controlPlaneDesiredVersionSyncer) desiredControlPlaneZVersion(ctx conte
 // frontend.newClusterAdmissionContext builds for cluster admission. The upgrade
 // controller passes the result to admission.AdmitClusterNodePoolsMinorVersionSkew
 // directly so that admission code stays free of any DB dependency.
-func (c *controlPlaneDesiredVersionSyncer) listClusterAdmissionNodePools(ctx context.Context, clusterResourceID *azcorearm.ResourceID) ([]admission.ClusterAdmissionNodePool, error) {
+//
+// Returns (_, false, nil) when any node pool is missing its
```
machi1990, line 375: The implication of this change is:
If a node pool just got created and its SPNP is missing for over 29s, we'll mark the cluster update operation as failed (see https://github.com/deads2k/ARO-HCP/blob/d445bb4846d55589f0bc9c8b0965b0ff7de6c095/backend/pkg/controllers/operationcontrollers/operation_cluster_update.go#L233 )

The creation of the [SPNP resync every 1m](https://github.com/deads2k/ARO-HCP/blob/d445bb4846d55589f0bc9c8b0965b0ff7de6c095/backend/pkg/controllers/create_service_provider_nodepool_controller.go#L65) and the relist of [nodepool happens every 30s](https://github.com/deads2k/ARO-HCP/blob/d445bb4846d55589f0bc9c8b0965b0ff7de6c095/backend/pkg/informers/informers.go#L55)

Meaning that, there a chance ( albeit a small one /edge case) that we could mark the control plane minor upgrade as failed if at the same time there is a node pool without the corresponding SPNP (when the node pool is created at about the same time).

Changefeed https://github.com/Azure/ARO-HCP/pull/5502 will make it better: until that lands, what is our stance on this issue? 
- Do we bump the timeout from 29s to 59s? 
- Increase the resync interval of the creation of SPNP? 
- Or Ignore the issue as its an edge case?

deads2k: > * Do we bump the timeout from 29s to 59s?

I think we'll do this.  I've found the bug that broke this, so I'm delivering a fix and bring it in along with this.

backend/pkg/controllers/create_service_provider_cluster_controller.go
```diff
@@ -0,0 +1,111 @@
...
+
+	_, err = c.serviceProviderClusterLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if err == nil {
+		return nil
+	}
+	if !database.IsNotFoundError(err) {
+		return utils.TrackError(fmt.Errorf("failed to get ServiceProviderCluster from lister: %w", err))
+	}
+
+	if _, err := database.GetOrCreateServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID()); err != nil {
```
machi1990, line 98: @deads2k what do you think about also initialising the .desired_version if the existingCluster has the X.Y.Z set in its version?

deads2k: > [@deads2k](https://github.com/deads2k) what do you think about also initialising the .desired_version if the existingCluster has the X.Y.Z set in its version?

starting empty is a sure way to get everything set in all failure modes.

PR #5701, opened 2026-06-17
include degraded message in operation condition

deads2k:
this will help our analysis tooling.


deads2k:
steve was ok via slack

/label lgtm

PR #5523, opened 2026-06-04
Shrink storage

deads2k:
This pays off some debt by eliminating duplicated functionality we created on the path to generic storage.

/hold  requires the finalization of our migration

PR #5449, opened 2026-05-30
allow customizing partitionkey

deads2k:
All customizing partition key to storage, so we can further align our storage and de-duplicate.  We can start this now because everything is using generic storage.

PR #5710, opened 2026-06-18
feat: add Cosmos DB migration controller to backend

deads2k:
Adds a subscription-watching controller that performs a read/write cycle on every document type in the Resources and KubeApplier containers to force Cosmos DB re-serialization. Each subscription is processed exactly once per process lifetime.

https://github.com/Azure/ARO-HCP/pull/5681 with CI coverage.


mbarnes:
/lgtm

The amount of logic repetition in the controller makes me wonder if there's a potential interface for database types that could shorten the cluster migration and its descendants into some kind of recursive pattern, maybe, but I'm not familiar enough with all the new types to suggest anything concrete at this time.

deads2k:
>The amount of logic repetition in the controller makes me wonder if there's a potential interface for database types that could shorten the cluster migration and its descendants into some kind of recursive pattern, maybe, but I'm not familiar enough with all the new types to suggest anything concrete at this time.

That's a good potential improvement. It's not strictly impossible, but would require a mapping from resourceType to CosmosType, plus a more complete untypedCRUD.  If you wanted to point @redhat-chai-bot to that in #internal-aro-hcp that would be a cool thing to do.

raelga:
/hold

This PR no longer builds against current `main` and fails every `e2e-parallel` batch at the images step (building the `aro-hcp-backend` image), so e2e never runs:

- https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/batch/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2067875575422259200
- https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/batch/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2067861664664915968

Both batches contain {5693, 5706, 5710, 5713} and do NOT include #5523, yet still fail:

```
backend/pkg/controllers/cosmos_migration.go:386:69: ... in argument to migrateControllers: need type assertion
make: *** [Makefile:33: aro-hcp-backend] Error 1
reason 'executing_graph:step_failed:building_project_image'
```

Root cause: the partition-key / CRUD refactor in #5449 ("allow customizing partitionkey") merged to main on 2026-06-18, enforcing CosmosMetadataAccessor on the CRUD type parameters and dropping the type-specific converters. This PR's `cosmos_migration.go` still targets the old single-type-param CRUD API. The PR was last green solo against main 15d70512 (before #5449); current main 2aae9a46 is after it.

Fix: rebase onto current main and update `cosmos_migration.go` to the new CRUD type-parameter API. Remove the hold once it builds.

nimrodshn:
A general comment about this approach -

Since we have moved the migration from being a synchronous process to an asynchronous background process we lost guarantees and tracking if the process finished and tracking for any mid-way work done.

One idea worth raising for future improvement is to persist the migration state to a document to allow for both tracking for when done as well as debugging and mid-way recovery if the backend leader is lost 
mid-way of the migration. 

This will also allow other advantages like:
- Detecting long-running or stalled migrations
- Progress visibility for operations
- Preventing duplicate work across backend replicas

@deads2k WDYT?

deads2k:
> One idea worth raising for future improvement is to persist the migration state to a document to allow for both tracking for when done as well as debugging and mid-way recovery if the backend leader is lost mid-way of the migration.
> 
> This will also allow other advantages like:
> 
>     * Detecting long-running or stalled migrations
> 
>     * Progress visibility for operations
> 
>     * Preventing duplicate work across backend replicas
> 
> 
> @deads2k WDYT?

No.

to determine doneness, build a kusto query against the datadump and see if the migration is finished.

deads2k:
/hold cancel

rebased to now compile

PR #5716, opened 2026-06-18
fix: ACR replication flag compat and Geography int/westus3 constraint

geoberle:
## Summary

- **ACR replication script**: PR #5551 replaced `--region-endpoint-enabled` with `--global-endpoint-routing`, but EV2 runs az CLI < 2.86.0 which doesn't have the new flag. Script now detects the supported flag from `--help` at startup, so it works on all az CLI versions.
- **Geography pipeline**: Added westus3 to int execution constraints. Kusto cluster creation was only running in uksouth, causing Region-level Kusto steps to fail in westus3.

## Test plan

- [x] Verify ACR replication script works in EV2 (az CLI < 2.86.0)
- [x] Verify ACR replication script works locally (az CLI 2.84.0)
- [ ] Run Global entrypoint pipeline for int/westus3 and confirm Geography steps execute

Ref: https://redhat.atlassian.net/browse/ARO-27812

### DIFF
--- a/dev-infrastructure/geography-pipeline.yaml
+++ b/dev-infrastructure/geography-pipeline.yaml
@@ -12,6 +12,7 @@ resourceGroups:
     - int
     regions:
     - uksouth
+    - westus3
   - clouds:
     - public
     environments:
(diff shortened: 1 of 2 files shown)


raelga (commented):
/lgtm
/approve

geoberle:
this change touches ACR replication, which does run during prow job
the other change is a runtime setting for INT allowing it to create a kusto cluster in NASA for the new INT region

PR #5726, opened 2026-06-19
ci: drop legacy role assignment adoption shim for dev shard0

roivaz:
Remove the temporary adoption shim that preserved Azure-generated GUID names for 45 role assignments on the first DEV E2E subscription (974ebd46). The old assignments (created imperatively via scripts) were deleted while the subscription was out of the CI pool, and the pipeline recreated them with deterministic guid(...) names.

* Remove legacyAssignmentIdsBySubscription param and all fallback logic from the RBAC Bicep templates
* All E2E subscriptions now use the same deterministic naming — no special-casing for shard0

Part of https://redhat.atlassian.net/browse/ARO-27253


raelga (approved):
/lgtm
/approve
/retest

PR #5744, opened 2026-06-23
fix: consume config.yaml prometheus shard/replica settings in prometheus helm chart

sclarkso:
https://redhat.atlassian.net/browse/AROSLSRE-1242

### What

Updates prometheus helm chart and tests to use replica/shard set in config.yaml

### Why

We already set replicas/shards to 2 in the config.yaml and sdp-pipelines, the prometheus chart currently ignores it.  Prometheus is getting OOMkilled in int and stage because due to WAL-replay.  Increasing the memory limit in int might work, but we are memory constrained in int and stage, and spreading the load (as intended by the existing config in aro-hcp and sdp-pipelines) gives us more capacity than just increasing the limit.

### Testing

personal dev env/e2e

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/dev-infrastructure/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_arohcp_monitor.yaml
+++ b/dev-infrastructure/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_arohcp_monitor.yaml
@@ -68585,7 +68585,7 @@ spec:
   probeSelector:
     matchLabels:
       release: arohcp-monitor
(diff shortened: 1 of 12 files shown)


inbharajmani:
dev env has the override in config.yaml. Should we perhaps remove them as well and use the default
https://github.com/Azure/ARO-HCP/blob/main/config/config.yaml#L1243
https://github.com/Azure/ARO-HCP/blob/main/config/config.yaml#L1272

PR #5745, opened 2026-06-23
fix: use configurable hcpNamespaceRegex for prometheus remote_write relabeling

inbharajmani:
https://redhat.atlassian.net/browse/AROSLSRE-1242

## Summary

- Adds a configurable `hcpNamespaceRegex` field to the Prometheus config to fix HCP metrics routing in CSPR
- In CSPR, the Clusters Service creates ephemeral OCM environments per PR run, and OCM assigns dynamic short IDs (e.g., `j-ah-846`) for namespace naming (`ocm-j-ah-846-<clusterID>`)
- The previous hardcoded regex `^ocm-arohcpcspr.*` never matched these namespaces, causing HCP metrics to be routed to the services workspace instead of the HCP workspace

## Why

The `services-cspr-usw3` Azure Monitor Workspace is at **124% events/min ingestion utilization** and actively dropping over 1M events per hour. Part of this overload is HCP metrics being sent to the wrong workspace because the `writeRelabelConfigs` regex doesn't match CSPR's dynamic namespace prefixes.

The HCP workspace (`hcps-cspr-usw3`) has been receiving **0 metrics** since the mgmt Prometheus was deployed, meaning HCP-specific dashboards and alerts are completely blind.

### Root cause

| Environment | CS `--environment` | OCM env ID | Namespace prefix | Regex match? |
|---|---|---|---|---|
| int/stg/prod | `arohcpint` | `arohcpint` (stable) | `ocm-arohcpint-*` | `^ocm-arohcpint.*` ✅ |
| CSPR | `arohcpcspr` | `j-ah-846` (dynamic per PR) | `ocm-j-ah-846-*` | `^ocm-arohcpcspr.*` ❌ |

In stable environments, OCM uses the environment name as the namespace prefix. In CSPR, OCM creates a new ephemeral environment per PR run and assigns a random short ID that doesn't match the `--environment` flag value.

### What this PR does

- Adds `hcpNamespaceRegex` to `prometheusSpec` in config, schema, and Helm values
- **Default** (all stable envs): `^ocm-arohcp<env>.*` — preserves existing behavior
- **CSPR override**: `^ocm-.*` — matches all OCM namespaces regardless of dynamic ID
- Updates the Prometheus template to use this configurable field instead of inline regex construction

## Test plan

- [x] `cd config && make materialize` passes
- [x] `make verify-schema` passes
- [x] Rendered configs verified:
  - CSPR: `hcpNamespaceRegex: ^ocm-.*`
  - dev: `hcpNamespaceRegex: ^ocm-arohcpdev.*`
  - pers: `hcpNamespaceRegex: ^ocm-arohcppers.*`
  - ci00: `hcpNamespaceRegex: ^ocm-arohcpci00.*`
- [ ] Deploy to CSPR and verify HCP metrics appear in `hcps-cspr-usw3` workspace

### DIFF
--- a/config/config.schema.json
+++ b/config/config.schema.json
@@ -787,6 +787,9 @@
             "retentionSize": {
               "type": "string",
               "pattern": "^$|^(0|([0-9]*[.])?[0-9]+((K|M|G|T|E|P)i?)?B)$"
+            },
+            "hcpNamespaceRegex": {
+              "type": "string"
             }
           }
         },
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -523,6 +523,7 @@ defaults:
             memory: NONE
         replicas: 2
         shards: 2
+        hcpNamespaceRegex: "^ocm-arohcp{{ .ctx.environment }}.*"
       prometheusConfigReloader:
         image:
           registry: mcr.microsoft.com/oss/v2
@@ -651,6 +652,7 @@ defaults:
             memory: NONE
         replicas: 2
         shards: 2
+        hcpNamespaceRegex: "^ocm-arohcp{{ .ctx.environment }}.*"
       prometheusConfigReloader:
         image:
           registry: mcr.microsoft.com/oss/v2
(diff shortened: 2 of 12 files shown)


inbharajmani:
/close

AMW are ephemeral and hence metrics are not important right now. Once we have a persistent AMW we can think about fixing the prometheus relabeling

PR #5746, opened 2026-06-23
test: add E2E test for AzureKmsKey update

ziccardi:
### What

This PR adds an E2E test that verifies updating an HCP cluster to use a new Azure KMS key version.

**The code is based on https://github.com/Azure/ARO-HCP/pull/5629/changes which needs to be merged before this.**

The test covers the full rotation flow:

- creates the test resource group and HCP cluster using the `v20260630preview` API;
- creates the required customer resources;
- creates a new version of the existing Azure Key Vault key;
- updates the cluster configuration to reference the new key version;
- waits for the cluster update operation to complete successfully;
- verifies that all `StorageVersionMigration` resources complete with `Succeeded=True`, ensuring that Kubernetes objects have been re-encrypted with the new key.

### Why

AzureKmsKey rotation must be validated end-to-end to ensure that updating the KMS key version on an existing cluster correctly triggers the re-encryption flow and leaves the cluster in a healthy state.

Checking only the ARM update result is not enough: the test also verifies the Kubernetes-side `StorageVersionMigration` resources so that we can confirm the re-encryption process actually completed.

The PR is intentionally split into multiple commits to make the implementation easier to review step by step. The history will be squashed before merge.

### PR Checklist
- [ ] Self-reviewed the diff


test/e2e/azure_kmskey_update.go
```diff
@@ -0,0 +1,151 @@
...
+				// TODO: Implementation
+
+				By("verifying the cluster is ready")
+				// TODO: Implementation
+			},
+		)
+	})
+
+	Context("Negative", func() {
+		It("creates a cluster with version < 4.22, rotates KMS key, and fails to update the cluster",
```
machi1990: Does this need to be an e2e? Can it be an integration test instead?

machi1990: I don't think we need a real cluster only to test that our validations are intact; we can achieve the same with a rapid feedback loop with an integration test

ahitacat: There are unit tests for this. I added them in the description so we covered all clusters version in the e2e. I have no strong opinion to only test cluster 4.22  and greater.

machi1990: Yes, I think we can content ourselves with that and drop the need for an e2e for the negative case

test/e2e/azure_kmskey_update.go
```diff
@@ -0,0 +1,151 @@
...
+
+				By("updating the cluster with the new KMS key")
+				// TODO: Implementation
+
+				By("verifying the cluster enters updating state")
+				// TODO: Implementation
+
+				By("waiting for re-encryption to complete")
+				// TODO: Implementation
+
```
machi1990: See https://redhat.atlassian.net/browse/ARO-27847?focusedCommentId=17359380 for some suggestion on how to the building blocks that can be used to achieve this

ziccardi: Mind a look?

test/e2e/azure_kmskey_update.go
```diff
@@ -0,0 +1,170 @@
+// Copyright 2025 Microsoft Corporation
```
machi1990:
```suggestion
// Copyright 2026 Microsoft Corporation
```

test/util/verifiers/storageversionmigration.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2025 Microsoft Corporation
```
machi1990:
```suggestion
// Copyright 2026 Microsoft Corporation
```

test/e2e/azure_kmskey_update.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azkeys"
+
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("AzureKmsKey Update", func() {
+	Context("Positive", func() {
```
machi1990:
```suggestion
```

test/e2e/azure_kmskey_update.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azkeys"
+
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("AzureKmsKey Update", func() {
```
machi1990:
```suggestion
var _ = Describe("Customer", func() {
```

test/e2e/azure_kmskey_update.go
```diff
@@ -0,0 +1,170 @@
...
+	Context("Positive", func() {
+		It("creates a cluster with version >= 4.22, rotates KMS key, and successfully updates the cluster",
```
machi1990:
```suggestion
		It("should be able to rotate KMS key for a cluster with version >= 4.22",
```

So that the test case leads like "Customer should be able to rotate KMS...."

See the example fixtures https://github.com/ziccardi/ARO-HCP/blob/7c737c8d8f9ff0117f3c41c97e43d93b6e166cf0/test/testdata/zz_fixture_TestMainListSuitesForEachSuite_prod_parallelprod_parallel.txt for how other test are named

test/e2e/kms_key_rotation.go
machi1990, line 1: Run ` UPDATE=true make test-unit ` to add this tests onto the list of fixtures

machi1990 (changes requested):

test/e2e/kms_key_rotation.go
machi1990, line 1: I wonder if the file name could use a better naming; 

kms_key_rotation or something close to the feature being tested

test/e2e/azure_kmskey_update.go
```diff
@@ -0,0 +1,170 @@
...
+				GinkgoLogr.Info("Successfully created new key version",
+					"keyVaultName", clusterParams.KeyVaultName,
+					"keyName", clusterParams.EtcdEncryptionKeyName,
+					"newVersion", newKeyVersion)
+
+				By("updating the cluster with the new KMS key")
+				// Update the cluster params with the new key version
+				clusterParams.EtcdEncryptionKeyVersion = newKeyVersion
+
+				// Rebuild the cluster object with the new key version
```
machi1990: The update + wait on the operations; can it use https://github.com/ziccardi/ARO-HCP/blob/7c737c8d8f9ff0117f3c41c97e43d93b6e166cf0/test/util/framework/hcp_helper.go#L1609 ?

test/e2e/azure_kmskey_update.go
```diff
@@ -59,6 +59,17 @@ var _ = Describe("AzureKmsKey Update", func() {
...
+
+				// Resolve to full install version if not using stable channel
+				if clusterParams.ChannelGroup != "stable" {
+					resolved, err := framework.GetLatestInstallVersion(ctx, clusterParams.ChannelGroup, clusterParams.OpenshiftVersionId)
+					if framework.IsVersionNotFoundError(err) {
+						Skip(fmt.Sprintf("no install version found for %s in %s channel (%v)", clusterParams.OpenshiftVersionId, clusterParams.ChannelGroup, err))
+					}
+					Expect(err).NotTo(HaveOccurred(), "failed to resolve install version for OpenShift %s in channel %s", clusterParams.OpenshiftVersionId, clusterParams.ChannelGroup)
+					clusterParams.OpenshiftVersionId = resolved
+				}
```
machi1990: This isn't strictly needed. I'd say we are better off not having it and keep this test straight forward

machi1990: The bot suggested it but I think it's one of those occasions were we are better off without the suggestion

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,229 @@
...
+			resourceGroup, err := tc.NewResourceGroup(ctx, "kms-key-rotate", tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for KMS key rotation test")
+
+			By("creating cluster parameters with version 4.22")
+			clusterParams := framework.NewDefaultClusterParams20260630()
+			clusterParams.ClusterName = clusterName
+			clusterParams.OpenshiftVersionId = "4.22"
+
+			By("resolving OpenShift version to installable release")
+			openShiftControlPlaneVersion, err := framework.GetLatestInstallVersion(ctx, clusterParams.ChannelGroup, "4.22")
```
machi1990: Why is this needed? `4.22` should be enough no? Once https://github.com/openshift/hypershift/pull/8790 is merged and the override for the already release 4.22.z versions available in the hypershift operator, the rotation feature should work for any 4.22+ version.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,229 @@
...
+				apiVersion  = "2026-06-30-preview"
+
+				// Timeout for StorageVersionMigration to complete re-encryption after KMS key rotation
+				storageVersionMigrationTimeout = 10 * time.Minute
+				verifierPollInterval           = 15 * time.Second
+			)
+
+			tc := framework.NewTestContext()
+
+			By("checking API version availability")
```
machi1990: What do you think of this https://github.com/Azure/ARO-HCP/blob/855eea85f3052237fc8528c6d280c0ac41ee4307/test/e2e/nodepool_ephemeral_osdisk_test.go#L133 approach?

i.e you try creating and if it fails, inspect the error that is api is not deployed and fail by the given time we expect the API to be there.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,229 @@
...
+			Expect(updateResult.Properties.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version).NotTo(BeNil(), "update result key Version was nil")
+			Expect(*updateResult.Properties.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version).To(Equal(newKeyVersion),
+				"cluster should reference the new KMS key version after update")
+
+			GinkgoLogr.Info("Cluster update completed successfully",
+				"clusterName", clusterName,
+				"provisioningState", *updateResult.Properties.ProvisioningState,
+				"newKeyVersion", newKeyVersion)
+
+			By("verifying StorageVersionMigration succeeded for re-encryption")
```
ahitacat, line 196: Most of the tests that requires to the AdminRestConfig does this step before making the update. If there is an issue getting this config it will fail early and avoid change more things. I think this can be done before rotating the key. So, if it is an issue, that will save us some az calls and update of the cluster.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,229 @@
...
+										ActiveKey: &hcpsdk20260630preview.KmsKeyUpdate{
+											Version: to.Ptr(newKeyVersion),
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+				framework.UpdateHCPClusterTimeout,
```
ahitacat, line 158: In my test the re-encryption took ~10minutes, I am a bit concern of this timeout as it might be trigger more times than we expect. Maybe we will need a new one ~15min? We should check this with the tests.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,229 @@
...
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to update cluster with new KMS key")
+
+			By("verifying the cluster is ready")
+			Expect(updateResult.Properties).NotTo(BeNil(), "update result Properties was nil")
+			Expect(updateResult.Properties.ProvisioningState).NotTo(BeNil(), "update result ProvisioningState was nil")
+			Expect(*updateResult.Properties.ProvisioningState).To(Equal("Succeeded"),
+				"cluster should be in Succeeded state after KMS key update completed")
+
+			By("verifying the cluster references the new KMS key version")
```
ahitacat, line 168: I would also add the step of that GETs the cluster an checks the key version.

test/util/verifiers/storageversionmigration.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+type verifyStorageVersionMigrationSucceeded struct{}
+
+func (v verifyStorageVersionMigrationSucceeded) Name() string {
+	return "VerifyStorageVersionMigrationSucceeded"
+}
+
+func (v verifyStorageVersionMigrationSucceeded) Verify(ctx context.Context, restConfig *rest.Config) error {
+	dynamicClient, err := dynamic.NewForConfig(restConfig)
```
ahitacat: It exists a typed client that I think is better to use.



```suggestion
	kubeClient, err := kubernetes.NewForConfig(adminRESTConfig)
```

test/util/verifiers/storageversionmigration.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// StorageVersionMigration is in storagemigration.k8s.io/v1beta1
+	svmGVR := schema.GroupVersionResource{
+		Group:    "storagemigration.k8s.io",
+		Version:  "v1beta1",
+		Resource: "storageversionmigrations",
+	}
+
+	svmList, err := dynamicClient.Resource(svmGVR).List(ctx, metav1.ListOptions{})
```
ahitacat: Then you can use the


```suggestion
svmGVR, err := kubeClient.StoragemigrationV1beta1().StorageVersionMigrations().List(ctx, metav1.ListOptions{})
```

test/util/verifiers/storageversionmigration.go
```diff
@@ -0,0 +1,106 @@
...
+			failedMigrations = append(failedMigrations, fmt.Sprintf("%s: failed to parse status.conditions: %v", name, err))
+			continue
+		}
+		if !found {
+			failedMigrations = append(failedMigrations, fmt.Sprintf("%s: status.conditions field not found", name))
+			continue
+		}
+
+		succeeded := false
+		for _, cond := range conditions {
```
ahitacat: With the client I suggest to use this will be simpler. Something similar as we do for nodes.go

test/e2e/kms_key_rotation.go
ziccardi: Done

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("Customer", func() {
+	// Deadline for v20260630preview API deployment in non-dev environments
+	timeBombDeadline := framework.Must(time.Parse(time.RFC3339, "2026-07-31T00:00:00Z"))
+
+	It("should be able to rotate KMS key for a cluster with version >= 4.22",
+		labels.RequireNothing, labels.High, labels.Positive,
```
ahitacat: I think it misses the` labels.AroRpApiCompatible`

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,206 @@
...
+				"keyVaultName", clusterParams.KeyVaultName,
+				"keyName", clusterParams.EtcdEncryptionKeyName,
+				"originalVersion", clusterParams.EtcdEncryptionKeyVersion)
+
+			// Create a new version of the key (this is the rotation)
+			kty := azkeys.KeyTypeRSA
+			createKeyResp, err := keyClient.CreateKey(ctx, clusterParams.EtcdEncryptionKeyName, azkeys.CreateKeyParameters{
+				Kty:     &kty,
+				KeySize: to.Ptr(int32(2048)),
+			}, nil)
```
mbukatov, line 124: The new key you are creating seem to be using the same name as used for the first key (created during cluster infra deployment via bicep)? Do I understand this right? Is this intentional?

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,204 @@
...
+								},
+							},
+						},
+					},
+				},
+				framework.UpdateHCPClusterTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to update cluster with new KMS key")
+
+			By("verifying the cluster is ready")
```
ahitacat, line 162: I think this step can be removed. If it is not successful the line 159 will finish this test. So the cluster should be ready.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,204 @@
...
+			Expect(updateResult.Properties).NotTo(BeNil(), "update result Properties was nil")
+			Expect(updateResult.Properties.ProvisioningState).NotTo(BeNil(), "update result ProvisioningState was nil")
+			Expect(*updateResult.Properties.ProvisioningState).To(Equal("Succeeded"),
+				"cluster should be in Succeeded state after KMS key update completed")
+
+			By("verifying the cluster references the new KMS key version")
```
ahitacat, line 168: This step can be also removed, I think it is covered by this one `By("confirming key version persists via GET (round-trip verification)")`

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,206 @@
...
+										ActiveKey: &hcpsdk20260630preview.KmsKeyUpdate{
+											Version: to.Ptr(newKeyVersion),
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+				framework.UpdateHCPClusterTimeout,
```
ahitacat, line 159: This timeout needs to change to something ~15min, I would create a new constant for that, similar to the one for upgrades.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,206 @@
...
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to update cluster with new KMS key")
+
+			By("verifying the cluster is ready")
+			Expect(updateResult.Properties).NotTo(BeNil(), "update result Properties was nil")
+			Expect(updateResult.Properties.ProvisioningState).NotTo(BeNil(), "update result ProvisioningState was nil")
+			Expect(*updateResult.Properties.ProvisioningState).To(Equal("Succeeded"),
```
ahitacat, line 166:
```suggestion
			Expect(*updateResult.Properties.ProvisioningState).To(Equal(hcpsdk20260630preview.ProvisioningStateSucceeded),
```

config/config.yaml
```diff
@@ -986,22 +976,18 @@ defaults:
...
   clustersService:
     image:
       registry: quay.io
       repository: app-sre/aro-hcp-clusters-service
-      # Manually pinned, see AROSLSRE-1395: the next digest (b17f6fe, vcs-ref
-      # ee741db) reproducibly breaks e2e-parallel — candidate 4.22/5.0 data-plane
-      # ingress never becomes reachable. Do not let the automated image bump
-      # advance this until fixed (source tag is pinned in image-updater/config.yaml).
-      digest: sha256:6a49b32884d56e707aafba6d73258cb57857087c261f606df71c74f2a9336bf7 # 18b5a25df4baa3fabe4a2dab97dcca51a5828c79 (2026-06-19 10:48)
+      digest: sha256:c7f2a35e844b0aa1c827aa700932ed52fa38b862731745cb0e1e1b6b9c2f6e5c
```
ahitacat, line 990: It has been updated, so we can remove this change. 
https://github.com/Azure/ARO-HCP/blob/main/config/config.yaml#L1000

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,201 @@
...
+			By("creating a resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "kms-key-rotate", tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for KMS key rotation test")
+
+			By("creating cluster parameters with version 4.22")
+			clusterParams := framework.NewDefaultClusterParams20260630()
+			clusterParams.ClusterName = clusterName
+			// TODO: Remove hardcode when latest changes are available in candidate channel
+			clusterParams.OpenshiftVersionId = "4.22.0-0.nightly-multi-2026-07-01-221503"
+			clusterParams.ChannelGroup = "nightly"
```
ahitacat, line 64: The new HO has the OCP override, so the hypershift change is there for all clusters with version 4.22 or greater. We can remove the hardcode, and use the `resolvedVersion, err := framework.GetLatestInstallVersion(ctx, framework.DefaultOCPChannelGroup, "4.22")`

ahitacat (commented):
I have rebased https://github.com/Azure/ARO-HCP/pull/5629 with main changes, can you please rebase this so we can see if the pipelines run successfully?

mbukatov:
This can be closed, since the E2E tests moved to https://github.com/Azure/ARO-HCP/pull/5629

ziccardi:
> This can be closed, since the E2E tests moved to #5629

Correct

PR #5751, opened 2026-06-23
Validate control plane operators permissions on user provided infra 

patilsuraj767:
[ARO-25131](https://redhat.atlassian.net/browse/ARO-25131)

### What
When creating an ARO-HCP Cluster, the end-users creating those Clusters need to pre-create a set of Azure Managed Identities which are then used by k8s control plane operators in the clusters.

This is the controller which validates the roles and permissions assigned to the managed identity provided by the user.


backend/pkg/azure/azurehelpers/role_definitions.go
```diff
@@ -82,6 +82,23 @@ func UnionDataActions(roleDefinitions []armauthorization.RoleDefinition) ([]stri
 	return dataActionsUnion.UnsortedList(), nil
 }
 
+// IntersectActions returns the role definition actions in a that are also present in b.
+// It assumes neither a nor b contains duplicate elements; this is not a true set
+// intersection, so repeated values in a are preserved in the result.
```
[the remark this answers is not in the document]
miguelsorianod: Is there some reason we can't make this a real set intersection? do we need the duplicates for some reason?

backend/pkg/azure/client/smi_client_builder.go
```diff
@@ -37,13 +41,16 @@ const (
...
-// ServiceManagedIdentityClientBuilder offers the ability tocreate Azure clients
-// authenticating as the the Cluster's Service Managed Identity, which is
+// ServiceManagedIdentityClientBuilder offers the ability to create Azure clients
+// authenticating as the Cluster's Service Managed Identity, which is
 // a cluster-scoped identity.
 type ServiceManagedIdentityClientBuilder interface {
 	BuilderType() ServiceManagedIdentityClientBuilderType
 	// UserAssignedIdentitiesClient returns a new User Assigned Identities client.
 	UserAssignedIdentitiesClient(ctx context.Context, clusterIdentityURL string, smiResourceID *azcorearm.ResourceID, subscriptionID string) (UserAssignedIdentitiesClient, error)
+
```
miguelsorianod: nitpick: no need for newline

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
```diff
@@ -15,16 +15,14 @@
...
-	// Name returns the name of the validation.
-	Name() string
-	// Validate validates the NodePool. It returns nil if the validation succeeds and an error otherwise.
-	Validate(ctx context.Context, cluster *api.HCPOpenShiftCluster, nodePoolSubscription *arm.Subscription, nodePool *api.HCPOpenShiftClusterNodePool) error
+// IdentityResourceMissingPermissions records the set of actions that an identity was denied or not granted on a specific Azure resource. It is only instantiated when at least one
+// action is missing; Decisions is always non-empty and contains only NotAllowed or Denied entries.
+type IdentityResourceMissingPermissions struct {
+	Resource  *azcorearm.ResourceID
+	Identity  *azcorearm.ResourceID
+	Decisions *[]azurecheckaccessv2client.AuthorizationDecision
```
miguelsorianod: This shouldn't be a pointer to slice of values. It should be a slice to pointers:
	Decisions []*azurecheckaccessv2client.AuthorizationDecision

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
```diff
@@ -0,0 +1,462 @@
...
+	for operatorName, identity := range cluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ControlPlaneOperators {
+		results, err := v.findMissingActionsForIdentity(ctx, checkAccessV2Client, cluster, operatorName, identity, &subnet.Subnet)
+		if err != nil {
+			return utils.TrackError(fmt.Errorf("failed to find missing actions for identity %q: %w", operatorName, err))
+		}
+		missingPermissions = append(missingPermissions, results...)
+	}
+
+	if len(missingPermissions) > 0 {
+		return utils.TrackError(fmt.Errorf("control plane operators missing required permissions: %s",
```
miguelsorianod: In this project we make use of long lines. We can put this in a single line

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+
+	return nil
+}
+
+func (v *ControlPlaneIdentitiesPermissionsValidation) roleActionsForOperator(ctx context.Context, operatorName string) ([]string, error) {
+	roleDefinitionsResourceIDs := v.clusterScopedIdentitiesConfig.ControlPlaneOperatorsIdentities[azure.ClusterOperatorIdentifier(operatorName)].RoleDefinitionsResourceIDs()
+	if len(roleDefinitionsResourceIDs) == 0 {
+		return nil, utils.TrackError(fmt.Errorf("no role definitions configured for operator identity %q", operatorName))
+	}
```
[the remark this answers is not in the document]
miguelsorianod: Yes, this feedback is accurate. This is dangerous. See https://github.com/Azure/ARO-HCP/pull/5751/changes#r3472696100

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
```diff
@@ -15,16 +15,14 @@
...
 
-// NodePoolValidation represents a validation that can be performed on a node pool.
-type NodePoolValidation interface {
-	// Name returns the name of the validation.
-	Name() string
-	// Validate validates the NodePool. It returns nil if the validation succeeds and an error otherwise.
-	Validate(ctx context.Context, cluster *api.HCPOpenShiftCluster, nodePoolSubscription *arm.Subscription, nodePool *api.HCPOpenShiftClusterNodePool) error
+// IdentityResourceMissingPermissions records the set of actions that an identity was denied or not granted on a specific Azure resource. It is only instantiated when at least one
+// action is missing; Decisions is always non-empty and contains only NotAllowed or Denied entries.
+type IdentityResourceMissingPermissions struct {
```
miguelsorianod: Make this unexported so other packages can't use it.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
```diff
@@ -0,0 +1,462 @@
...
+
+	return missingPermissions
+}
+
+// formatMissingPermissionsMessage builds a human-readable string from a set of authorization
+// failures. Each IdentityResourceMissingPermissions contributes one entry describing the
+// identity, resource, and the specific not-allowed and denied actions or dataActions, for example:
+//
+//	identity '.../cloud-controller' on resource '.../networkSecurityGroups/nsg': not allowed actions: [Microsoft.Network/networkSecurityGroups/write], denied dataActions: [Microsoft.Storage/storageAccounts/blobServices/containers/blobs/read]
+func formatMissingPermissionsMessage(results []*IdentityResourceMissingPermissions) string {
```
miguelsorianod: Thinking further about this: why don't we define our own struct with json tags and we simply print the json?

something like:

type identityResourceMissingPermissions struct {
	Resource  *azcorearm.ResourceID  `json:"XXXX"`
	Identity  *azcorearm.ResourceID  ...
	Decisions []*checkaccessv2AuthorizationDecisionData ...
}

type checkaccessv2AuthorizationDecisionData struct {
  ActionID ...
  IsDataAction ...
  AccessDecision ...
}

that would structure the data nicely in json format when outputting it and we wouldn't need this method at all

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
```diff
@@ -0,0 +1,462 @@
...
+		clusterScopedIdentitiesConfig:               clusterScopedIdentitiesConfig,
+		backendIdentityAzureCachedReaders:           backendIdentityAzureCachedReaders,
+		checkAccessV2ClientBuilder:                  checkAccessV2ClientBuilder,
+		miDataplaneBasedAccessTokenRetrieverBuilder: miDataplaneBasedAccessTokenRetrieverBuilder,
+		checkAccessV2Scope:                          checkAccessV2Scope,
+	}
+}
+
+func (v *ControlPlaneIdentitiesPermissionsValidation) Name() string {
+	return "ControlPlaneIdentitiesPermissionsValidation"
```
miguelsorianod: Do we expect in the future potentially to have validations that are associated to the nodepool? if so,  maybe we should end the name with ClusterValidation ?

This applies also to the type itself too (and the file names)

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	var roleActions []string
+	seenActions := map[string]struct{}{}
+	for _, roleDefinitionResourceID := range roleDefinitionsResourceIDs {
+		roleDefinitionResponse, err := v.backendIdentityAzureCachedReaders.RoleDefinitionsCachedReader.GetCachedByID(ctx, roleDefinitionResourceID.String(), nil)
+		if err != nil {
+			return nil, utils.TrackError(fmt.Errorf("failed to get role definition %q: %w", roleDefinitionResourceID.String(), err))
+		}
+
+		actions, err := azurehelpers.ActionsFromRoleDefinition(roleDefinitionResponse.RoleDefinition)
```
miguelsorianod: Then can we have an azure helper method that returns you the union, similar to others we have?

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
```diff
@@ -0,0 +1,462 @@
...
+		roleDefinitionResponse, err := v.backendIdentityAzureCachedReaders.RoleDefinitionsCachedReader.GetCachedByID(ctx, roleDefinitionResourceID.String(), nil)
+		if err != nil {
+			return nil, utils.TrackError(fmt.Errorf("failed to get role definition %q: %w", roleDefinitionResourceID.String(), err))
+		}
+
+		dataActions, err := azurehelpers.DataActionsFromRoleDefinition(roleDefinitionResponse.RoleDefinition)
+		if err != nil {
+			return nil, utils.TrackError(err)
+		}
+		for _, dataAction := range dataActions {
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/5751/changes#r3472840319

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
```diff
@@ -0,0 +1,462 @@
...
+// Only actions from roleDefinitionActions/roleDefinitionDataActions that are relevant to NSG resources are checked (via intersection with the known NSG action set); unrelated actions are
+// ignored. This avoids sending actions to CheckAccessV2 that a given operator role was never expected to hold on an NSG.
+// It returns:
+//   - (nil, nil) if the identity has all required permissions, or if none of the role's actions apply to NSG resources.
+//   - a non-nil slice of NotAllowed/Denied AuthorizationDecision entries, one per missing action, if any permission is absent.
+//   - (nil, error) if the CheckAccessV2 API call fails.
+func (v *ControlPlaneIdentitiesPermissionsValidation) checkNotAllowedAndDeniedActionsForNetworkSecurityGroup(ctx context.Context, checkAccessV2Client azureclient.CheckAccessV2Client, resourceID *azcorearm.ResourceID, roleDefinitionActions []string, roleDefinitionDataActions []string, token azcore.AccessToken) ([]azurecheckaccessv2client.AuthorizationDecision, error) {
+	// Union of all Microsoft.Network/networkSecurityGroups/* actions that appear across any operator role in dev-infrastructure/configurations/dev-operator-roles.tmpl.bicepparam
```
miguelsorianod: This comment is not correct. The source of truth is in `internal/azure/cluster_scoped_identities_config.go`. The same applies to all others.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
```diff
@@ -0,0 +1,462 @@
...
+	networkSecurityGroupActions := []string{
+		"Microsoft.Network/networkSecurityGroups/read",
+		"Microsoft.Network/networkSecurityGroups/write",
+		"Microsoft.Network/networkSecurityGroups/join/action",
+	}
+	var networkSecurityGroupDataActions []string
+
+	requiredActions := azurehelpers.IntersectActions(networkSecurityGroupActions, roleDefinitionActions)
+	requiredDataActions := azurehelpers.IntersectActions(networkSecurityGroupDataActions, roleDefinitionDataActions)
+	if len(requiredActions) == 0 && len(requiredDataActions) == 0 {
```
miguelsorianod: Is this check needed? checkNotAllowedAndDeniedActionsForResourceID seems to check that. Same comment applies to other similar methods

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,462 @@
...
+	if totalExpected == 0 {
+		return nil, nil
+	}
+
+	authRequest, err := checkAccessV2Client.CreateAuthorizationRequest(resourceID.String(), actions, token.Token)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+
+	for _, da := range dataActions {
```
miguelsorianod, line 317: This seems a bit odd. Why do we create the authrequest and then append dataactions afterwards? craft the values that are to be passed first and then call CreateAuthorizationRequest. Make sure you are not overwriting one of the original slices

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
patilsuraj767: Name is perfix with `ControlPlane` which itself says it is not nodepool.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,462 @@
...
+		return nil, utils.TrackError(err)
+	}
+	if authDecisionResponse == nil {
+		return nil, utils.TrackError(fmt.Errorf("authorization response for '%s' is unexpectedly nil, indicating a possible internal failure", resourceID.String()))
+	}
+
+	if totalExpected != len(authDecisionResponse.Value) {
+		logger.Error(fmt.Errorf("mismatch in authorization decision response for '%s': expected '%d' actions, got '%d' actions",
+			resourceID.String(), totalExpected, len(authDecisionResponse.Value)), "mismatch in authorization decision response")
+	}
```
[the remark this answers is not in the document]
miguelsorianod: This feedback is correct. We shouldn't log. We should error. Otherwise the validation will pass

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation_test.go
```diff
@@ -0,0 +1,839 @@
...
+	}
+
+	resourceID := api.Must(azcorearm.ParseResourceID("/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/testRG/providers/Microsoft.Network/networkSecurityGroups/testNSG"))
+
+	tests := []struct {
+		name               string
+		actions            []string
+		dataActions        []string
+		decisions          []checkaccessv2.AuthorizationDecision
+		noCheckAccess      bool
```
miguelsorianod: Instead of a bool, have a setupCheckAccessV2ClientMock that receives the mock and sets it up. If several test cases set up the same content, you can provide a helper (within this same test function) that different test cases call.

The reason we do it like this instead of the bool is because the bool approach doesn't scale over time if more needs appear (more bools, strings ,...) would be introduced making the test harder to reason and follow. Conceptually what we are doing is setting up the mock. If there are different mock config needs that will scale correctly.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
miguelsorianod: That doesn't necessarily imply that data/resources from a nodepool is not involved. For example, when creating a NodePool you can provide a different subnet than the parent cluster one. Should the control plane identity have permission to interact with that subnet?

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
patilsuraj767: CreateAuthorizationRequest currently does not populate the IsDataAction flag for the provided actions. That’s why we create the corresponding ActionInfo objects manually and append them to authRequest.Actions after calling CreateAuthorizationRequest.
Check CreateAuthorizationRequest definition - https://github.com/Azure/checkaccess-v2-go-sdk/blob/main/client/client.go#L113

>Make sure you are not overwriting one of the original slices

As we are appending to the slice, it will not overwrite the slice.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation_test.go
patilsuraj767: setupCheckAccessV2ClientMock functions so each case drives its own expected call behaviour

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
patilsuraj767: Removed the check

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
patilsuraj767: Changed it, Using union now.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
patilsuraj767: >Should the control plane identity have permission to interact with that subnet?

I don't have answer for this, but I have renamed the type and file name.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_validation.go
patilsuraj767: Done, Please take a look

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
```diff
@@ -0,0 +1,48 @@
...
+package validations
+
+import (
+	"encoding/json"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+)
+
+// checkaccessv2AuthorizationDecisionData captures a single CheckAccessV2 decision where the identity was not granted access.
+type checkaccessv2AuthorizationDecisionData struct {
```
miguelsorianod, line 23: nitpick: move this struct below identityResourceMissingPermissions as it's used by it

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
```diff
@@ -0,0 +1,48 @@
...
+
+// identityResourceMissingPermissions records the set of actions that an identity was denied or not granted on a specific Azure resource.
+// It is only instantiated when at least one action is missing; Decisions is always non-empty and contains only NotAllowed or Denied entries.
+type identityResourceMissingPermissions struct {
+	Resource  *azcorearm.ResourceID
+	Identity  *azcorearm.ResourceID
+	Decisions []*checkaccessv2AuthorizationDecisionData
+}
+
+func (m *identityResourceMissingPermissions) MarshalJSON() ([]byte, error) {
```
miguelsorianod: I am not sure why we have a custom marshal function. Can we not define the json tags in the struct and just marshal?

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"encoding/json"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+)
+
+// checkaccessv2AuthorizationDecisionData captures a single CheckAccessV2 decision where the identity was not granted access.
+type checkaccessv2AuthorizationDecisionData struct {
+	ActionID       string `json:"actionId"`
+	IsDataAction   bool   `json:"isDataAction,omitempty"`
```
miguelsorianod: remove omitempty. We want to always show it

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
```diff
@@ -0,0 +1,48 @@
...
+	AccessDecision string `json:"accessDecision"`
```
miguelsorianod: not sure why this is a string. use the library's corresponding data type

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
miguelsorianod: Interesting. Definitely document this as it's a surprising behavior

miguelsorianod: Have we also verified that it works correctly when providing data actions in this way?

backend/pkg/azure/azurehelpers/role_definitions.go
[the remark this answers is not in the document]
patilsuraj767: I don’t see any reason to preserve duplicates, so I refactored it to perform a true set intersection.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
miguelsorianod: >As we are appending to the slice, it will not overwrite the slice.

That is not strictly accurate. An append can modify the original slice.

Example:

```
arr := []int{1, 2, 3, 4} // len: 4, cap: 4
slice1 := arr[0:2]       // [1, 2] - len: 2, cap: 4

slice2 := append(slice1, 99) 

fmt.Println(slice1) // Output: [1, 2] (slice1 itself didn't change)
fmt.Println(slice2) // Output: [1, 2, 99]
fmt.Println(arr)    // Output: [1, 2, 99, 4] (The underlying array WAS modified!
```

Given that the code modifies the newly created array by the library we should be fine

patilsuraj767: No not verified because currently we do not check any data action permission for any resource.

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
patilsuraj767: The catch is what json.Marshal will produce for `*azcorearm.ResourceID`
`azcorearm.ResourceID` is a struct with fields like Parent, SubscriptionID, ResourceGroupName, Provider, Name, etc. and a recursive *ResourceID parent pointer.
If you just add json:"resource" tags to the field and call json.Marshal, we will get this instead of a clean path string.
```
{
  "resource": {
    "Parent": {
      "Parent": null,
      "SubscriptionID": "11111111-...",
      "ResourceGroupName": "myRG",
      "Provider": "Microsoft.Network",
      ...
    },
    "SubscriptionID": "11111111-...",
    "ResourceGroupName": "myRG",
    "Provider": "Microsoft.Network",
    "Name": "myNSG",
    ...
  }
}
```

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,391 @@
...
+func (v *ControlPlaneIdentitiesPermissionsClusterValidation) roleDataActionsForOperator(ctx context.Context, operatorName string) ([]string, error) {
+	roleDefinitionsResourceIDs := v.clusterScopedIdentitiesConfig.ControlPlaneOperatorsIdentities[azure.ClusterOperatorIdentifier(operatorName)].RoleDefinitionsResourceIDs()
+	if len(roleDefinitionsResourceIDs) == 0 {
+		return nil, nil
+	}
+	roleDefinitions, err := v.fetchRoleDefinitions(ctx, roleDefinitionsResourceIDs)
+	if err != nil {
+		return nil, err
+	}
+	return azurehelpers.UnionDataActions(roleDefinitions)
```
miguelsorianod, line 132: Now that we use UnionDataActions: that method never returns nil currently. What occurs on our code when we receive a non nil empty array? same question applies to all union methods usage

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
miguelsorianod: Let's verify. If the library turns out to not support that we shouldn't add code we don't know whether it works

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
patilsuraj767: Done

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
miguelsorianod: No this is not correct. Have you tried it?

`*azcorearm.ResourceID` should marshal into a resource id string, as it has a custom marshaltext function.

backend/pkg/azure/azurehelpers/role_definitions.go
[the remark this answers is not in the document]
miguelsorianod: :+1:

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	"github.com/Azure/ARO-HCP/internal/azure"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// --- Test helpers ---
+
+func newTestCluster() *api.HCPOpenShiftCluster {
+	cluster := api.MinimumValidClusterTestCase()
```
miguelsorianod: odd usage. As of now no tests in backend make use of api.MinimumValidClusterTestCase. That seems to be used for validations in frontend. I'd avoid using it here for now at least

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+	cluster.CustomerProperties.Platform.OperatorsAuthentication = api.OperatorsAuthenticationProfile{
+		UserAssignedIdentities: api.UserAssignedIdentitiesProfile{
+			ControlPlaneOperators: map[string]*azcorearm.ResourceID{
+				"CloudController": api.Must(azcorearm.ParseResourceID("/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/testResourceGroup/providers/Microsoft.ManagedIdentity/userAssignedIdentities/cloud-controller")),
```
miguelsorianod: Don't hardcode the operator name. Make use of existing ones (internal/azure/cluster_scoped_identities_config.go) if you intend to set a valid one. The same comment applies to all intended valid operator names

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+			},
+			ServiceManagedIdentity: api.Must(azcorearm.ParseResourceID("/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/testResourceGroup/providers/Microsoft.ManagedIdentity/userAssignedIdentities/smi")),
+		},
+	}
+	return cluster
+}
+
+func newTestSubscription() *arm.Subscription {
+	sub := api.CreateTestSubscription()
```
miguelsorianod: Same comment as https://github.com/Azure/ARO-HCP/pull/5751/changes#r3491514337

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+			ID: ptr.To("/subscriptions/11111111-1111-1111-1111-111111111111/providers/Microsoft.Authorization/roleDefinitions/test-role-def"),
+			Properties: &armauthorization.RoleDefinitionProperties{
+				Permissions: permissions,
+			},
+		},
+	}
+}
+
+// fakeToken returns a non-expired access token for use in tests.
+func fakeToken() azcore.AccessToken {
```
miguelsorianod: Too generic name. at the least newFakeAzureAccessToken

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := utils.ContextWithLogger(context.Background(), logr.Discard())
+			ctrl := gomock.NewController(t)
+
+			// --- CheckAccess builder and client ---
+			mockCheckAccessClient := azureclient.NewMockCheckAccessV2Client(ctrl)
```
miguelsorianod: Unnecessarily complext test body. Similar to feedback for other tests, do not set N different individual knobs. What this is doing conceptually is setting a mock for the check access client. That's what you need to have as a test arg

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
patilsuraj767: How does it matters?

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+									AccessDecision: checkaccessv2.Allowed,
+								}
+							}
+						}
+						return &checkaccessv2.AuthorizationDecisionResponse{Value: decisions}, nil
+					}).AnyTimes()
+			}
+
+			// --- SMI / subnets builder ---
+			mockSMIBuilder := azureclient.NewMockServiceManagedIdentityClientBuilder(ctrl)
```
miguelsorianod: Similar comment to https://github.com/Azure/ARO-HCP/pull/5751/changes#r3491544930

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+						Return(mockSimpleSubnetsClient(ctrl, tc.routeTable), nil)
+				}
+			}
+
+			// --- Identities config, role definition reader, and token retriever ---
+			var identitiesConfig *azure.ClusterScopedIdentitiesConfig
+			var readers *cachedreader.BackendIdentityAzureCachedReaders
+			var tokenRetrieverBuilder azureclient.MIDataplaneBasedIdentityAccessTokenRetrieverBuilder
+
+			if tc.emptyIdentitiesConfig {
```
miguelsorianod: Similar comment to the other ones.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+}
+
+// --- Tests for checkNotAllowedAndDeniedActionsForNetworkSecurityGroup ---
+
+func TestCheckNotAllowedAndDeniedActionsForNSG(t *testing.T) {
+	nsgResourceID := api.Must(azcorearm.ParseResourceID("/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/testRG/providers/Microsoft.Network/networkSecurityGroups/testNSG"))
+
+	setupCheckAccess := func(mock *azureclient.MockCheckAccessV2Client, decisions []checkaccessv2.AuthorizationDecision, err error) {
+		mock.EXPECT().CreateAuthorizationRequest(gomock.Any(), gomock.Any(), gomock.Any()).
+			Return(&checkaccessv2.AuthorizationRequest{}, nil)
```
miguelsorianod: We accept long lines. Applies to other similar statements

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+		setupCheckAccessV2ClientMock func(mock *azureclient.MockCheckAccessV2Client)
+		wantLen                      int
+		wantDecisions                []string
+		wantErr                      bool
+		wantErrContains              string
+	}{
+		{
+			name:        "all allowed",
+			roleActions: testAllNetworkActions(),
+			setupCheckAccessV2ClientMock: func(mock *azureclient.MockCheckAccessV2Client) {
```
miguelsorianod: I am unsure of the value of setupCheckAccess. Why not just defining what we want inside setupCheckAccessV2ClientMock ? if you have a specific payload that you reuse define that as a helper  function. Do not have complex logic in the fixture itself, making it hard to follow.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+		})
+	}
+}
+
+// --- Tests for checkNotAllowedAndDeniedActionsForVNet ---
+
+func TestCheckNotAllowedAndDeniedActionsForVNet(t *testing.T) {
+	vnetResourceID := api.Must(azcorearm.ParseResourceID("/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/testRG/providers/Microsoft.Network/virtualNetworks/testVNet"))
+
+	setupCheckAccess := func(mock *azureclient.MockCheckAccessV2Client, decisions []checkaccessv2.AuthorizationDecision, err error) {
```
miguelsorianod: Similar comment to https://github.com/Azure/ARO-HCP/pull/5751/changes#r3491577864 wherever I am seeing this

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+	azureclient "github.com/Azure/ARO-HCP/backend/pkg/azure/client"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	"github.com/Azure/ARO-HCP/internal/azure"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// --- Test helpers ---
+
+func newTestCluster() *api.HCPOpenShiftCluster {
```
miguelsorianod: Currently only used in one test, and it seems specific enough to the needs of that specific validation, so move to wherever used. Avoid unnecessary wide scoping

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+			ControlPlaneOperators: map[string]*azcorearm.ResourceID{
+				"CloudController": api.Must(azcorearm.ParseResourceID("/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/testResourceGroup/providers/Microsoft.ManagedIdentity/userAssignedIdentities/cloud-controller")),
+			},
+			ServiceManagedIdentity: api.Must(azcorearm.ParseResourceID("/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/testResourceGroup/providers/Microsoft.ManagedIdentity/userAssignedIdentities/smi")),
+		},
+	}
+	return cluster
+}
+
+func newTestSubscription() *arm.Subscription {
```
miguelsorianod: Currently used in a single place. Move. Avoid unnecessary wide scope. Unless you think it would be reussable between fully independent validatons and is not too coupled to specific tests / validations

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: I am curious how much this matters. I use an AI code editor to speed up routine edits, and it sometimes reformats code or wraps lines differently. Since our style guide allows long lines, I didn’t think it was worth spending additional time manually adjusting formatting unless it affects readability or violates our linting rules.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+	sub := api.CreateTestSubscription()
+	sub.Properties.TenantId = ptr.To(api.TestTenantID)
+	return sub
+}
+
+func newRoleDefinitionResponseWithDataActions(actions []string, dataActions []string) armauthorization.RoleDefinitionsClientGetByIDResponse {
```
miguelsorianod: Currently used in a single place. Move to where used. Avoid unnecessary wide scope

miguelsorianod: Additionally, there seems to be coupling with the test TestValidate, via hardcoded roledefinition ID attribute. Having that coupling while being at different scopes is bad design and ends up with brittle tests if that changes.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+	}
+}
+
+// fakeToken returns a non-expired access token for use in tests.
+func fakeToken() azcore.AccessToken {
+	return azcore.AccessToken{Token: "fake-token", ExpiresOn: time.Now().Add(time.Hour)}
+}
+
+// successfulRetriever returns a mock retriever that always yields fakeToken().
+func successfulRetriever(ctrl *gomock.Controller) *azureclient.MockMIDataplaneBasedIdentityAccessTokenRetriever {
```
miguelsorianod: Name too generic for being a package-level function.

Additionally, only used in one test. Move close.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+	mock := azureclient.NewMockMIDataplaneBasedIdentityAccessTokenRetriever(ctrl)
+	mock.EXPECT().GetToken(gomock.Any(), gomock.Any()).Return(fakeToken(), nil).AnyTimes()
+	return mock
+}
+
+// mockSuccessfulTokenRetrieverBuilder returns a builder whose Build always returns
+// a retriever that issues a valid fake token, with no MI Dataplane interaction.
+func mockSuccessfulTokenRetrieverBuilder(ctrl *gomock.Controller) *azureclient.MockMIDataplaneBasedIdentityAccessTokenRetrieverBuilder {
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/5751/changes#r3491611604

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+	mock.EXPECT().Build(gomock.Any(), gomock.Any()).DoAndReturn(
+		func(_ string, _ *azcorearm.ResourceID) (azureclient.MIDataplaneBasedIdentityAccessTokenRetriever, error) {
+			return successfulRetriever(ctrl), nil
+		}).AnyTimes()
+	return mock
+}
+
+// testIdentitiesConfig builds a ClusterScopedIdentitiesConfig with a single
+// "CloudController" operator referencing the given role definition resource ID.
+func testIdentitiesConfig(roleDefResourceID *azcorearm.ResourceID) *azure.ClusterScopedIdentitiesConfig {
```
miguelsorianod: Name too generic. Only used in one place. Move. I also notice it mentions "CloudController". If that's coupled to newTestCluster package-level function that's bad design as they are coupled and are independent functions

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// testAllNetworkActions returns the full set of network role actions used across tests.
+func testAllNetworkActions() []string {
```
miguelsorianod: Name is too generic as this is a package-level variable. Make it more concrete describing to what set of tests it applies

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+		"Microsoft.Network/virtualNetworks/write",
+		"Microsoft.Network/virtualNetworks/subnets/join/action",
+		"Microsoft.Network/virtualNetworks/subnets/read",
+		"Microsoft.Network/virtualNetworks/subnets/write",
+	}
+}
+
+// mockSimpleSubnetsClient returns a mock SubnetsClient that returns a subnet with
+// no attached resources, or with the optional RouteTable set.
+func mockSimpleSubnetsClient(ctrl *gomock.Controller, routeTable *armnetwork.RouteTable) *azureclient.MockSubnetsClient {
```
miguelsorianod: Only used in one place. Too wide. Move

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,779 @@
...
+				Properties: &armnetwork.SubnetPropertiesFormat{
+					RouteTable: routeTable,
+				},
+			},
+		}, nil).AnyTimes()
+	return mock
+}
+
+// testCheckAccessV2Scope is the CheckAccessV2 scope used in all Validate tests.
+const testCheckAccessV2Scope = "https://management.azure.com/.default"
```
miguelsorianod: Only used in one test. Too wide. Move

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: Fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: Fixed, now using `internal/azure/cluster_scoped_identities_config.go`

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: Fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: Fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: I have tried to reduce the complexity, please take a look.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: I have tried to reduce the complexity, please take a look.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: Removed `emptyIdentitiesConfig`.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: I have tried to reduce the complexity, please take a look.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: fixed

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
patilsuraj767: Verified by creating test script. It is working as expected.

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
patilsuraj767: You are correct. I have removed the `MarshalJSON()`

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
patilsuraj767: Done

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
patilsuraj767: There is no behavioral change in this flow, the downstream code uses `len()` checks, which treat nil and []string{} identically.

>What occurs on our code when we receive a non nil empty array?

`IntersectActions` function explicitly normalizes the empty case. So even if roleDefinitionActions is []string{}, the intersection with the hardcoded resource-specific actions will be empty, and IntersectActions returns nil.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
miguelsorianod: The feedback I received is that we want longer lines because shorter ones impact visibility. lengths of up to 150-200 characters are common. If you have an IDE that adds newlines on short length reconfigure it

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation.go
miguelsorianod: Thanks

backend/pkg/controllers/validationcontrollers/validations/identity_authorization_result.go
miguelsorianod: it's a nitpick. Visually you see first the data type and afterwards the data types that the datatype itself references

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,762 @@
...
+	type wantDecision struct {
+		actionId string
+		decision string
+	}
+
+	resourceID := api.Must(azcorearm.ParseResourceID("/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/testRG/providers/Microsoft.Network/networkSecurityGroups/testNSG"))
+
+	// setupCheckAccessWithCapture sets up the mock to mirror how the real client populates
+	// ActionInfo entries, and writes the final request to *capture for post-call assertions.
+	setupCheckAccessWithCapture := func(mock *azureclient.MockCheckAccessV2Client, decisions []checkaccessv2.AuthorizationDecision, capture *checkaccessv2.AuthorizationRequest) {
```
miguelsorianod: I am unsure about why this design here.

This seems code smell: non trivial logic within the mock, "capture" name, ...

The mock should just setup the mock.

Why not have:

wantErr (bool)
want []*checkaccessv2AuthorizationDecisionData

and then simply check the returned results are the expected ones?

miguelsorianod: I can see this maybe was from a desire of wanting to make sure that the check access api receives the expected data and also that it returns the data initially provided?

That's not a good approach. That approach is trying to reimplement the behavior of the check access client and trying to test integrate the function and the checkaccessclient. That's an integration test concern and not an unit test. It's also test integrating custom behavior and not the real one so it does not provide testing value. The function is effectively testing **how** the function talks to its dependency instead of **what** the function produces. That couples the test to implementation details, if someone refactors the function to build the request differently but prodces the same results, the test breaks for not good reason.

We are not interested in that. We are interested in unit testing the function contract: "given this input, this is what is returned".

The same applies to some other potential tests that have been designed in this way.

patilsuraj767: I have rewrote the unit test from scratch. Followed the simple pattern of input and expected output in the test case. Please take a look.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,1280 @@
...
+			dataActions: []string{"Microsoft.Storage/storageAccounts/blobServices/containers/blobs/read"},
+			setupMock: func(m *azureclient.MockCheckAccessV2Client) {
+				m.EXPECT().CreateAuthorizationRequest(testResourceID.String(), []string{"Microsoft.Network/networkSecurityGroups/read"}, fakeToken.Token).
+					Return(&azurecheckaccessv2client.AuthorizationRequest{
+						Actions: []azurecheckaccessv2client.ActionInfo{
+							{Id: "Microsoft.Network/networkSecurityGroups/read"},
+						},
+					}, nil)
+				m.EXPECT().CheckAccess(gomock.Any(), gomock.Any()).
+					DoAndReturn(func(_ context.Context, req azurecheckaccessv2client.AuthorizationRequest) (*azurecheckaccessv2client.AuthorizationDecisionResponse, error) {
```
miguelsorianod, line 229: In general the mocks shouldn't be performing testing or have complex logic. We don't really test the interaction with the internal dependency, but in this specific case I am fine keeping it as it's an isolated case and provides a bit of added value

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,1280 @@
...
+		roleDefinitionDataActions []string
+		setupMock                 func(*azureclient.MockCheckAccessV2Client)
+		wantResult                []*checkaccessv2AuthorizationDecisionData
+		wantErr                   bool
+	}{
+		{
+			name:                      "no overlap with VNet actions returns nil without API call",
+			roleDefinitionActions:     []string{"Microsoft.Compute/virtualMachines/read"},
+			roleDefinitionDataActions: nil,
+			setupMock:                 func(_ *azureclient.MockCheckAccessV2Client) {},
```
miguelsorianod: No need to explicitly define setupMock when not used. Instead of doing this, in the test body do something like

if t.setupMock != nil {
  t.setupMock()
}

in that way, it allows us to not needing to define it explicitly for every test case

Apply this to all tests

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,1280 @@
...
+						},
+					},
+				},
+			},
+			setupMock:   func(_ *cachedreader.MockRoleDefinitionsCachedReader) {},
+			wantActions: nil,
+			wantErr:     true,
+		},
+		{
+			name:         "operator with role definitions returns union of actions",
```
miguelsorianod: Here it would be more accurate to fully exercise the union by having a couple of role definitions with permissions in each

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,1280 @@
...
+						},
+					},
+				},
+			},
+			setupMock:       func(_ *cachedreader.MockRoleDefinitionsCachedReader) {},
+			wantDataActions: nil,
+			wantErr:         false,
+		},
+		{
+			name:         "operator with role definitions returns union of data actions",
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/5751/changes#r3504403988

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
miguelsorianod: and it would be useful to set dataactions too to make sure that those are not included

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
miguelsorianod: and it would be useful too to set actions too to make sure they are not included

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,1280 @@
...
+						return &azurecheckaccessv2client.AuthorizationDecisionResponse{Value: decisions}, nil
+					}).Times(2)
+			},
+			wantErr: false,
+		},
+		{
+			name: "missing permissions returns error with details",
+			setupCheckAccess: func(m *azureclient.MockCheckAccessV2Client) {
+				callCount := 0
+				m.EXPECT().CheckAccess(gomock.Any(), gomock.Any()).
```
miguelsorianod, line 1280: The mock contains too much complex logic. In unit tests we do not want to test integrate, even less with a fake dependency. Just return what you want that is coherent with the setup

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
```diff
@@ -0,0 +1,1280 @@
...
+		name             string
+		setupCheckAccess func(*azureclient.MockCheckAccessV2Client)
+		wantErr          bool
+		expectedError    string
+	}{
+		{
+			name: "all permissions granted returns nil",
+			setupCheckAccess: func(m *azureclient.MockCheckAccessV2Client) {
+				m.EXPECT().CheckAccess(gomock.Any(), gomock.Any()).
+					DoAndReturn(func(_ context.Context, req azurecheckaccessv2client.AuthorizationRequest) (*azurecheckaccessv2client.AuthorizationDecisionResponse, error) {
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/5751/changes#r3504432268

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: Done

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: Done. Added multiple roles to test union.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: Done. Added multiple roles to test union.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: removed DoAndReturn which contains the complex logic and now returning only what is needed.

backend/pkg/controllers/validationcontrollers/validations/control_plane_identities_permissions_cluster_validation_test.go
patilsuraj767: Done

tuxerrante (commented):
I found three issues that still look merge-relevant, and I think the safest fix direction in this PR is to keep them as small guardrails rather than broadening behavior.

1. `ControlPlaneIdentitiesPermissionsClusterValidation.roleActionsForOperator()` / `roleDataActionsForOperator()` assume every control-plane operator key exists in `clusterScopedIdentitiesConfig.ControlPlaneOperatorsIdentities` and call `RoleDefinitionsResourceIDs()` on the lookup result without checking it. That turns an unsupported operator key into a backend panic instead of a validation failure. The minimal safe fix is to guard the lookup and return a clear validation error for unknown operators here. I would avoid widening admission/API validation in this PR unless you want to intentionally expand scope.

2. `BackendOptions.validate()` does not require `CloudEnvironment`, but this PR now dereferences `b.options.CloudEnvironment.CheckAccessV2Scope()` during backend startup. That means `NewBackend()` can still succeed and fail later with a nil dereference in `Run()`. The smallest fix is to make `validate()` fail fast when `CloudEnvironment` is nil.

3. Route table handling still trusts nullable Azure SDK fields too much. In `checkMissingPermissionsForRouteTable()` and `checkNotAllowedAndDeniedActionsForRouteTable()` the code dereferences route-table IDs without guarding the nested pointers first. The minimal safe fix is to add nil checks and return tracked errors for malformed SDK responses instead of panicking.

PR #5752, opened 2026-06-23
feat: add cluster upgrade partial and complete SLO observability

machi1990:
### What

Add cluster version selection-to-active SLO recording rules, alerts & slo dashboard

### Why

feat: add UJ cluster upgrade partial and complete SLO observability

- Add an alert that pages when an upgrade has been stuck reaching partial for over 20 minutes (`for: 5m`, so paging starts after ~25 minutes)
- Add an alert that pages when an upgrade has been stuck completing for over 30 minutes (`for: 5m`, so paging starts after ~35 minutes)
- Add dashboards for the SLOs
Screenshot 2026-08-04 at 18 02 23

Screenshot 2026-08-07 at 14 30 47

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
@@ -53,3 +53,37 @@ resource arohcpClusterProvisionSloRecordingRules 'Microsoft.AlertsManagement/pro
     ]
   }
 }
+
+resource arohcpUserJourneyClusterUpgradeRecordingRules 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-03-01' = {
+  name: 'arohcp_user_journey_cluster_upgrade_recording_rules'
+  location: location
+  properties: {
+    scopes: [
+      azureMonitoring
+    ]
+    enabled: true
+    interval: 'PT1M'
+    rules: [
+      {
+        record: 'hosted_control_plane_upgrade:upgrade_eligible:info'
+        expression: '((count by (cluster, resource_id, subscription_id, cluster_uuid) (count by (cluster, resource_id, subscription_id, cluster_uuid, version) (backend_cluster_version_info == 1)) >= 2) and on (cluster, resource_id) (count by (cluster, resource_id) (backend_cluster_version_info{state="completed"} == 1) >= 1)) * 0 + 1'
+      }
+      {
+        record: 'hosted_control_plane_upgrade:version_state_first_seen:timestamp'
+        expression: 'min without (prometheus_replica) (min by (cluster, resource_id, subscription_id, cluster_uuid, version, state) ((hosted_control_plane_upgrade:version_state_first_seen:timestamp or (timestamp(backend_cluster_version_info{state=~"desired|partial"} == 1) and on (cluster, resource_id) (hosted_control_plane_upgrade:upgrade_eligible:info == 1))) unless on (cluster, resource_id, subscription_id, cluster_uuid, version) (max by (cluster, resource_id, subscription_id, cluster_uuid, version) (backend_cluster_version_info{state="completed"} == 1))))'
+      }
+      {
+        record: 'hosted_control_plane_upgrade:in_progress:count'
+        expression: 'count by (cluster) (count by (cluster, resource_id) ((max by (cluster, resource_id, subscription_id, cluster_uuid, version, state) (backend_cluster_version_info{state=~"desired|partial"} == 1) unless on (cluster, resource_id, subscription_id, cluster_uuid, version) max by (cluster, resource_id, subscription_id, cluster_uuid, version) (backend_cluster_version_info{state="completed"} == 1))) >= 1 and on (cluster, resource_id) (hosted_control_plane_upgrade:upgrade_eligible:info == 1)) or 0 * count by (cluster) (backend_cluster_version_info)'
+      }
+      {
+        record: 'hosted_control_plane_upgrade:duration_in_desired:seconds'
+        expression: '(time() - hosted_control_plane_upgrade:version_state_first_seen:timestamp{state="desired"}) and on (cluster, resource_id, subscription_id, cluster_uuid, version) (max by (cluster, resource_id, subscription_id, cluster_uuid, version, state) (backend_cluster_version_info{state="desired"} == 1 unless on (cluster, resource_id, subscription_id, cluster_uuid, version) (max by (cluster, resource_id, subscription_id, cluster_uuid, version) (backend_cluster_version_info{state="partial"} == 1 or backend_cluster_version_info{state="completed"} == 1)))) and on (cluster, resource_id) (hosted_control_plane_upgrade:upgrade_eligible:info == 1)'
+      }
+      {
+        record: 'hosted_control_plane_upgrade:duration_in_progress:seconds'
+        expression: '((time() - min without (state) (hosted_control_plane_upgrade:version_state_first_seen:timestamp{state=~"desired|partial"})) and on (cluster, resource_id) (hosted_control_plane_upgrade:upgrade_eligible:info == 1) unless on (cluster, resource_id, subscription_id, cluster_uuid, version) (max by (cluster, resource_id, subscription_id, cluster_uuid, version) (backend_cluster_version_info{state="completed"} == 1))) * on (cluster, resource_id, subscription_id, cluster_uuid, version) group_left (state) (max by (cluster, resource_id, subscription_id, cluster_uuid, version, state) (backend_cluster_version_info{state="partial"} == 1 or (backend_cluster_version_info{state="desired"} == 1 unless on (cluster, resource_id, subscription_id, cluster_uuid, version) max by (cluster, resource_id, subscription_id, cluster_uuid, version) (backend_cluster_version_info{state="partial"} == 1))))'
+      }
+    ]
+  }
+}
--- a/observability/alerts-rp-services.yaml
+++ b/observability/alerts-rp-services.yaml
@@ -2,6 +2,7 @@ prometheusRules:
   rulesFolders:
   - alerts/access-cluster-slo-prometheusRule.yaml
   - alerts/cluster-provision-slo-prometheusRule.yaml
+  - alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
   - alerts/nodepool-slo-prometheusRule.yaml
   untestedRules: []
   testDependencies:
--- a/observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
+++ b/observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
@@ -1,0 +1,57 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    app.kubernetes.io/name: kube-prometheus
+    app.kubernetes.io/part-of: kube-prometheus
+    prometheus: k8s
+    role: alert-rules
+  name: user-journey-cluster-upgrade-alerts
+  namespace: monitoring
+spec:
+  groups:
+  # Per-upgrade stuck alerts for cluster upgrade SLOs. Sev 4.
+  # At current upgrade volumes, alert directly on duration instead of fleet burn-rate.
+  #
+  # The ControlPlaneActiveVersion controller derives upgrade Partial/Completed state from
+  # HostedCluster version history (preferring status.controlPlaneVersion.history, else
+  # status.version.history), which feeds backend_cluster_version_info and these alerts.
+  # Control plane version history reports for versions >= 4.20, so completeness tracks the
+  # control plane independently of the data plane. While 4.19 is still around, those
+  # clusters fall back to status.version.history, where upgrade completeness depends on
+  # data plane state — e.g. no worker nodes means the upgrade will not reach completed;
+  # missing console component in external auth prevents CVO from marking the upgrade
+  # complete. The complete alert may therefore fire for 4.19 clusters due to missing
+  # data plane configuration rather than control plane failure.
+  - name: rp-user-journey-cluster-upgrade-monitor-rules
+    rules:
+    - alert: userJourneyClusterUpgradeStuckInDesired
+      expr: |
+        hosted_control_plane_upgrade:duration_in_desired:seconds > 1200
+      for: 5m
+      labels:
+        severity: "4"
+        slo: cluster-upgrade-reach-partial
+        component: slo
+      annotations:
+        summary: "{{ $labels.cluster }}: Cluster upgrade on {{ $labels.resource_id }} to version {{ $labels.version }} not progressing"
+        description: |
+          Cluster upgrade target version {{ $labels.version }} has been in desired for over 20 minutes without reaching partial. With the alert pending for 5 minutes, paging starts after ~25 minutes. Partial is when the target version becomes active (recognized as active) on the HCP cluster; the upgrade is in progress but not yet complete. Investigate backend_cluster_version_info on the affected cluster.
+          Service Cluster: {{ $labels.cluster }}
+        runbook_url: https://aka.ms/arohcp-runbook-cluster-upgrade
+        correlationId: "userJourneyClusterUpgradeStuckInDesired/{{ $labels.cluster }}{{ $labels.resource_id }}/{{ $labels.version }}"
+    - alert: userJourneyClusterUpgradeStuckInProgress
+      expr: |
+        hosted_control_plane_upgrade:duration_in_progress:seconds > 1800
+      for: 5m
+      labels:
+        severity: "4"
+        slo: cluster-upgrade-complete
+        component: slo
+      annotations:
+        summary: "{{ $labels.cluster }}: Cluster upgrade on {{ $labels.resource_id }} to version {{ $labels.version }} stuck"
+        description: |
+          Cluster upgrade target version {{ $labels.version }} has been in progress for over 30 minutes without reaching completed. With the alert pending for 5 minutes, paging starts after ~35 minutes. Completed is when the target version upgrade has finished and is reported as complete on the HCP cluster. Investigate backend_cluster_version_info on the affected cluster.
+          Service Cluster: {{ $labels.cluster }}
+        runbook_url: https://aka.ms/arohcp-runbook-cluster-upgrade
+        correlationId: "userJourneyClusterUpgradeStuckInProgress/{{ $labels.cluster }}{{ $labels.resource_id }}/{{ $labels.version }}"
--- a/observability/grafana-dashboards/sre/user-journey/cluster-upgrade-slo.json
+++ b/observability/grafana-dashboards/sre/user-journey/cluster-upgrade-slo.json
@@ -1,0 +1,1446 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 1,
+  "id": null,
+  "links": [],
+  "panels": [
+    {
+      "gridPos": {
+        "h": 4,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 1,
+      "options": {
+        "code": {
+          "language": "plaintext",
+          "showLineNumbers": false,
+          "showMiniMap": false
+        },
+        "content": "### Cluster Upgrade SLOs\n\nPer-upgrade observability from `backend_cluster_version_info` via `hosted_control_plane_upgrade:*` recording rules.\n\n| SLO | Target | Alert |\n|-----|--------|-------|\n| **Reach partial** | `partial` within 20m | `userJourneyClusterUpgradeStuckInDesired` (>20m in `desired`, `for: 5m` \u2192 pages after ~25m) |\n| **Upgrade complete** | `completed` within 30m | `userJourneyClusterUpgradeStuckInProgress` (>30m in progress, `for: 5m` \u2192 pages after ~35m) |\n\n**Upgrade SLOs** \u2014 in-progress and stuck upgrade health against the 20m / 30m thresholds (alerts add a 5-minute pending window before paging).\n\n**Upgrade and Version Visualization Across the Fleet** \u2014 share of clusters with an upgrade in progress, clusters by latest active z-stream, success rate, stuck rate by z-stream (threshold via Stuck Minutes), and latest z-stream per minor (ControlPlaneDesiredVersion gateway selection).\n\n**Runbook:** [https://aka.ms/arohcp-runbook-cluster-upgrade](https://aka.ms/arohcp-runbook-cluster-upgrade)",
+        "mode": "markdown"
+      },
+      "pluginVersion": "12.4.3",
+      "title": "Dashboard Info",
+      "type": "text"
+    },
+    {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 4
+      },
+      "id": 11,
+      "panels": [],
+      "title": "Upgrade SLOs",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "datasource",
+        "uid": "-- Mixed --"
+      },
+      "description": "Upgrade-eligible clusters with a version upgrade currently in progress (desired or partial, not yet completed).",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "blue"
+              }
+            ]
+          }
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 4,
+        "w": 8,
+        "x": 0,
+        "y": 5
+      },
+      "id": 3,
+      "options": {
+        "colorMode": "background",
+        "graphMode": "none",
+        "reduceOptions": {
+          "calcs": [
+            "lastNotNull"
+          ],
+          "fields": "",
+          "values": false
+        },
+        "textMode": "auto"
+      },
+      "pluginVersion": "12.4.3",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "sum(hosted_control_plane_upgrade:in_progress:count{cluster=~\"$cluster\"}) or vector(0)",
+          "instant": true,
+          "legendFormat": "Count of upgrades in progress",
+          "refId": "A"
+        }
+      ],
+      "title": "Count of Upgrades In Progress",
+      "type": "stat"
+    },
+    {
+      "datasource": {
+        "type": "datasource",
+        "uid": "-- Mixed --"
+      },
+      "description": "Upgrades stuck in desired for over 20 minutes without reaching partial (SLO threshold). Alert userJourneyClusterUpgradeStuckInDesired adds for: 5m and pages after ~25m.",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green"
+              },
+              {
+                "color": "red",
+                "value": 1
+              }
+            ]
+          }
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 4,
+        "w": 8,
+        "x": 8,
+        "y": 5
+      },
+      "id": 4,
+      "options": {
+        "colorMode": "background",
+        "graphMode": "none",
+        "reduceOptions": {
+          "calcs": [
+            "lastNotNull"
+          ],
+          "fields": "",
+          "values": false
+        },
+        "textMode": "auto"
+      },
+      "pluginVersion": "12.4.3",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "sum(count by (cluster) (hosted_control_plane_upgrade:duration_in_desired:seconds{cluster=~\"$cluster\"} > 1200) or (0 * hosted_control_plane_upgrade:in_progress:count{cluster=~\"$cluster\"})) or vector(0)",
+          "instant": true,
+          "legendFormat": "Stuck in desired",
+          "refId": "A"
+        }
+      ],
+      "title": "Stuck in Desired (>20m)",
+      "type": "stat"
+    },
+    {
+      "datasource": {
+        "type": "datasource",
+        "uid": "-- Mixed --"
+      },
+      "description": "Upgrades stuck in desired or partial for over 30 minutes without reaching completed (SLO threshold). Alert userJourneyClusterUpgradeStuckInProgress adds for: 5m and pages after ~35m.",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green"
+              },
+              {
+                "color": "red",
+                "value": 1
+              }
+            ]
+          }
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 4,
+        "w": 8,
+        "x": 16,
+        "y": 5
+      },
+      "id": 5,
+      "options": {
+        "colorMode": "background",
+        "graphMode": "none",
+        "reduceOptions": {
+          "calcs": [
+            "lastNotNull"
+          ],
+          "fields": "",
+          "values": false
+        },
+        "textMode": "auto"
+      },
+      "pluginVersion": "12.4.3",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "sum(count by (cluster) (hosted_control_plane_upgrade:duration_in_progress:seconds{cluster=~\"$cluster\"} > 1800) or (0 * hosted_control_plane_upgrade:in_progress:count{cluster=~\"$cluster\"})) or vector(0)",
+          "instant": true,
+          "legendFormat": "Stuck in progress",
+          "refId": "A"
+        }
+      ],
+      "title": "Stuck in Progress (>30m)",
+      "type": "stat"
+    },
+    {
+      "datasource": {
+        "type": "datasource",
+        "uid": "-- Mixed --"
+      },
+      "description": "Upgrade-eligible clusters with a version upgrade currently in progress (desired or partial, not yet completed).",
+      "fieldConfig": {
+        "defaults": {
+          "custom": {
+            "filterable": true
+          }
+        },
+        "overrides": [
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "Time In Progress (Minutes)"
+            },
+            "properties": [
+              {
+                "id": "custom.filterable",
+                "value": false
+              }
+            ]
+          }
+        ]
+      },
+      "gridPos": {
+        "h": 8,
+        "w": 24,
+        "x": 0,
+        "y": 9
+      },
+      "id": 2,
+      "options": {
+        "showHeader": true
+      },
+      "pluginVersion": "12.4.3",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "hosted_control_plane_upgrade:duration_in_progress:seconds{cluster=~\"$cluster\"} / 60",
+          "format": "table",
+          "instant": true,
+          "legendFormat": "",
+          "refId": "A"
+        }
+      ],
+      "title": "In-Progress Upgrade Durations",
+      "type": "table",
+      "transformations": [
+        {
+          "id": "filterFieldsByName",
+          "options": {
+            "include": {
+              "names": [
+                "cluster",
+                "resource_id",
+                "state",
+                "version",
+                "Value"
+              ]
+            }
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^cluster$",
+            "renamePattern": "Service Cluster"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^resource_id$",
+            "renamePattern": "Resource ID"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^version$",
+            "renamePattern": "Target Version"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^state$",
+            "renamePattern": "Current State"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^Value$",
+            "renamePattern": "Time In Progress (Minutes)"
+          }
+        }
+      ]
+    },
+    {
+      "datasource": {
+        "type": "datasource",
+        "uid": "-- Mixed --"
+      },
+      "description": "Clusters with upgrades stuck past the 20 minute reach-partial window.",
+      "fieldConfig": {
+        "defaults": {
+          "custom": {
+            "filterable": true
+          }
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 8,
+        "w": 24,
+        "x": 0,
+        "y": 17
+      },
+      "id": 6,
+      "options": {
+        "showHeader": true
+      },
+      "pluginVersion": "12.4.3",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "(\n  hosted_control_plane_upgrade:duration_in_desired:seconds{cluster=~\"$cluster\"}\n  and\n  hosted_control_plane_upgrade:duration_in_desired:seconds{cluster=~\"$cluster\"} > 1200\n) / 60",
+          "format": "table",
+          "instant": true,
+          "legendFormat": "",
+          "refId": "A"
+        }
+      ],
+      "title": "Stuck in Desired Detail",
+      "type": "table",
+      "transformations": [
+        {
+          "id": "filterFieldsByName",
+          "options": {
+            "include": {
+              "names": [
+                "cluster",
+                "resource_id",
+                "state",
+                "version",
+                "Value"
+              ]
+            }
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^cluster$",
+            "renamePattern": "Service Cluster"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^resource_id$",
+            "renamePattern": "Resource ID"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^version$",
+            "renamePattern": "Target Version"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^state$",
+            "renamePattern": "Current State"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^Value$",
+            "renamePattern": "Time In Desired (Minutes)"
+          }
+        }
+      ]
+    },
+    {
+      "datasource": {
+        "type": "datasource",
+        "uid": "-- Mixed --"
+      },
+      "description": "Clusters with upgrades stuck past the 30 minute upgrade-complete window.",
+      "fieldConfig": {
+        "defaults": {
+          "custom": {
+            "filterable": true
+          }
+        },
+        "overrides": [
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "Time In Progress (Minutes)"
+            },
+            "properties": [
+              {
+                "id": "custom.filterable",
+                "value": false
+              }
+            ]
+          }
+        ]
+      },
+      "gridPos": {
+        "h": 8,
+        "w": 24,
+        "x": 0,
+        "y": 25
+      },
+      "id": 7,
+      "options": {
+        "showHeader": true
+      },
+      "pluginVersion": "12.4.3",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "(\n  hosted_control_plane_upgrade:duration_in_progress:seconds{cluster=~\"$cluster\"}\n  and\n  hosted_control_plane_upgrade:duration_in_progress:seconds{cluster=~\"$cluster\"} > 1800\n) / 60",
+          "format": "table",
+          "instant": true,
+          "legendFormat": "",
+          "refId": "A"
+        }
+      ],
+      "title": "Stuck in Progress Detail",
+      "type": "table",
+      "transformations": [
+        {
+          "id": "filterFieldsByName",
+          "options": {
+            "include": {
+              "names": [
+                "cluster",
+                "resource_id",
+                "state",
+                "version",
+                "Value"
+              ]
+            }
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^cluster$",
+            "renamePattern": "Service Cluster"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^resource_id$",
+            "renamePattern": "Resource ID"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^version$",
+            "renamePattern": "Target Version"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^state$",
+            "renamePattern": "Current State"
+          }
+        },
+        {
+          "id": "renameByRegex",
+          "options": {
+            "regex": "^Value$",
+            "renamePattern": "Time In Progress (Minutes)"
+          }
+        }
+      ]
+    },
+    {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 33
+      },
+      "id": 12,
+      "panels": [],
+      "title": "Upgrade and Version Visualization Across the Fleet",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "datasource",
+        "uid": "-- Mixed --"
+      },
+      "description": "Percentage of clusters with a version upgrade currently in progress (desired or partial, not completed), divided by total clusters (backend_cluster_created_time_seconds).",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "decimals": 1,
+          "max": 100,
+          "min": 0,
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green"
+              },
+              {
+                "color": "yellow",
+                "value": 10
+              },
+              {
+                "color": "red",
+                "value": 25
+              }
+            ]
+          },
+          "unit": "percent"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 4,
+        "w": 8,
+        "x": 0,
+        "y": 34
+      },
+      "id": 8,
+      "options": {
+        "colorMode": "background",
+        "graphMode": "none",
+        "reduceOptions": {
+          "calcs": [
+            "lastNotNull"
+          ],
+          "fields": "",
+          "values": false
+        },
+        "textMode": "auto"
+      },
+      "pluginVersion": "12.4.3",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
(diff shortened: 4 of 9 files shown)


dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
```diff
@@ -1,3 +1,127 @@
...
+  properties: {
+    scopes: [
+      azureMonitoring
+    ]
+    enabled: true
+    interval: 'PT1M'
+    rules: [
+      {
+        record: 'hosted_control_plane_version:selection_to_partial:state_first_seen_timestamp'
+        expression: 'min without(prometheus_replica) ( min by (cluster, resource_id, subscription_id, cluster_uuid, version, state) ( hosted_control_plane_version:selection_to_partial:state_first_seen_timestamp or timestamp(backend_cluster_version_info == 1) ) )'
```
JameelB: the name of the metric `selection_to_partial` suggests that this is specific to a state. But in the expression, it seems like we're recording it for all states, 'desired', 'partial' and 'completed'. 

The way it's used, for example, below: `hosted_control_plane_version:selection_to_partial:state_first_seen_timestamp{state="desired"}` may be confusing. We could either remove this bit from the naming, something like 'hosted_control_plane_version:state_first_seen_timestamp'. Or make the recording specific to that state. I've no preference. wdyt?

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
```diff
@@ -1,3 +1,127 @@
...
+      }
+      {
+        record: 'hosted_control_plane_version:selection_to_partial:desired_pending_seconds'
+        expression: '( time() - hosted_control_plane_version:selection_to_partial:state_first_seen_timestamp{state="desired"} ) and on(cluster, resource_id, subscription_id, cluster_uuid, version) ( backend_cluster_version_info{state="desired"} == 1 )'
```
JameelB: Is the goal of the second part of this query to ensure we only capture metrics with the `desired` state to see how long it's been in that state (ignores when it moves to partial/completed).

Looking at the way the metrics are emitted ([PR](https://github.com/Azure/ARO-HCP/pull/5584/changes#diff-b88f0fcdcd73487f84c24fb7b83b829f8c1eba953d4b86536a59229ca847d008R57)), it seems to re-emit every sync and there wouldn't be a recording rule `hosted_control_plane_version:selection_to_partial:state_first_seen_timestamp{state="desired"}` for it anymore cause `backend_cluster_version_info == 1` for that state would not exist? If my understanding there is correct, is the second half of the query really needed?

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
machi1990: Thanks; I like this suggestion

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
machi1990: > backend_cluster_version_info{state="desired"} == 1

The intent of this was to capture "the given desired version is stuck on desired".
To only show up upgrades that are currently stuck as opposed to the version was ever stuck on desired.

Though I am wondering what's the best approach between the two differences; should we interest ourselves with the current stuck or current and any previous stuck?

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,60 @@
...
+    prometheus: k8s
+    role: alert-rules
+  name: uj-cluster-upgrade-version-selection-to-partial-alerts
+  namespace: monitoring
+spec:
+  groups:
+  # Multi-window burn-rate alerts for cluster upgrade version selection-to-partial SLO. Sev 4 (info).
+  - name: rp-uj-cluster-upgrade-version-selection-to-partial-monitor-rules
+    rules:
+    - alert: UJClusterUpgradeVersionSelectionToPartial5m1h
```
swiencki: Tier suffix should be long-window-first (`1h5m`, `6h30m`, `3d6h`) per the convention in the KAS and NodePool alerts.

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,60 @@
...
+      for: 5m
+      labels:
+        severity: info
+        slo: cluster-upgrade-version-selection-to-partial
+        short_window: 5m
+        long_window: 1h
+      annotations:
+        summary: "{{ $labels.cluster }}: Cluster upgrade version selection-to-partial error budget burning rapidly (>14.4x)"
+        description: "More than 14.4x of the 99.95% cluster upgrade version selection-to-partial error budget is being consumed over 5m and 1h windows. Cluster upgrades are stuck for over 20 minutes without the target version reaching partial. Investigate backend_cluster_version_info on affected clusters."
+        runbook_url: TBD
```
swiencki: RP lane alerts need a resolvable `aka.ms/arohcp-runbook-...` short-link before merge, even if the runbook is still high level.

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - name: rp-uj-cluster-upgrade-version-selection-to-partial-monitor-rules
+    rules:
+    - alert: UJClusterUpgradeVersionSelectionToPartial5m1h
+      expr: |
+        hosted_control_plane_version:selection_to_partial:stuck_over_20m:burnrate5m > 14.4
+        and
+        hosted_control_plane_version:selection_to_partial:stuck_over_20m:burnrate1h > 14.4
+      for: 5m
+      labels:
+        severity: info
```
swiencki: ADR recommends `warning` (Sev3) for fast/medium UJ tiers since they represent active customer-journey degradation; reserve `info` for the slow/ticket tier only.

machi1990: Okay. I don't want these to wake up anyone for now until at least some soaking time and a TSG is there; can you confirm that warning won't wake up anyone? :-D

machi1990: I'll keep it to Sev4 for now and adjust after soaking

observability/recording-rules-services.yaml
```diff
@@ -1,4 +1,6 @@
 prometheusRules:
-  rulesFolders: []
+  rulesFolders:
+  - alerts/UJClusterUpgradeVersionSelectionToPartialRecord-prometheusRule.yaml
```
machi1990: From @cgiradkar 

> Verify if we need to put it in a different folder e.g record rules

observability/alerts-rp-services.yaml
machi1990, line 1: I am unsure if this is the correct lane to use.  https://github.com/swiencki/architecture/blob/16fbc2fb1ffa04b61d21ae6f61f7a6b1845c1825/offerings/aro/decisions/adr-001-aro-hcp-alerting-recommendation-for-sre.md#routing-lanes-in-aro-hcp

Discussing with @swiencki to confirm this

machi1990: Simon has confirmed that's a good lane

observability/recording-rules-services.yaml
machi1990: The location of the recording rules is fine for now and follows the approach for already previous existing rules.
e.g HCPkasRecord-prometheusRule-KSM.yaml rules which are referenced here `observability/recording-rules-hcps.yaml`

Maybe in the long run, the project could choose a different pattern where all recording rules resides in a folder or a subfolder. The PR doesn't intend to change this

observability/alerts/userJourneyClusterUpgradeCompleteMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,70 @@
...
+  - name: rp-uj-cluster-upgrade-complete-monitor-rules
+    rules:
+    - alert: UJClusterUpgradeComplete1h5m
+      expr: |
+        hosted_control_plane_version:upgrade_complete:stuck_over_30m:burnrate5m > 14.4
+        and
+        hosted_control_plane_version:upgrade_complete:stuck_over_30m:burnrate1h > 14.4
+      for: 5m
+      labels:
+        severity: info
```
machi1990: intentionally as sev4 for now and over time it'll be graduated to the corresponding severity as par https://github.com/swiencki/architecture/blob/16fbc2fb1ffa04b61d21ae6f61f7a6b1845c1825/offerings/aro/decisions/adr-001-aro-hcp-alerting-recommendation-for-sre.md#burn-rate-tiers-timing-precision

observability/alerts/userJourneyClusterUpgradeCompleteMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,70 @@
...
+        description: "More than 14.4x of the 99.95% cluster upgrade complete error budget is being consumed over 5m and 1h windows. Cluster upgrades are stuck for over 30 minutes without reaching completed. Investigate backend_cluster_version_info on affected clusters."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-upgrade"
+    - alert: UJClusterUpgradeComplete6h30m
+      expr: |
+        hosted_control_plane_version:upgrade_complete:stuck_over_30m:burnrate30m > 6
+        and
+        hosted_control_plane_version:upgrade_complete:stuck_over_30m:burnrate6h > 6
+      for: 30m
+      labels:
+        severity: info
```
machi1990: same as https://github.com/Azure/ARO-HCP/pull/5752/changes#r3476800035 intentional as info for now and will be graduated over time

observability/alerts/userJourneyClusterUpgradeCompleteMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,70 @@
...
+        description: "More than 6x of the 99.95% cluster upgrade complete error budget is being consumed over 30m and 6h windows. Cluster upgrades are stuck for over 30 minutes without reaching completed. Investigate backend_cluster_version_info on affected clusters."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-upgrade"
+    - alert: UJClusterUpgradeComplete3d6h
+      expr: |
+        hosted_control_plane_version:upgrade_complete:stuck_over_30m:burnrate6h > 1
+        and
+        hosted_control_plane_version:upgrade_complete:stuck_over_30m:burnrate3d > 1
+      for: 30m
+      labels:
+        severity: info
```
machi1990: same as https://github.com/Azure/ARO-HCP/pull/5752/changes#r3476800035 to make it info and graduate over time

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,60 @@
...
+        description: "More than 14.4x of the 99.95% cluster upgrade version selection-to-partial error budget is being consumed over 5m and 1h windows. Cluster upgrades are stuck for over 20 minutes without the target version reaching partial. Investigate backend_cluster_version_info on affected clusters."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-upgrade"
+    - alert: UJClusterUpgradeVersionSelectionToPartial6h30m
+      expr: |
+        hosted_control_plane_version:selection_to_partial:stuck_over_20m:burnrate30m > 6
+        and
+        hosted_control_plane_version:selection_to_partial:stuck_over_20m:burnrate6h > 6
+      for: 30m
+      labels:
+        severity: info
```
machi1990: same comment as https://github.com/Azure/ARO-HCP/pull/5752/changes#r3476800035

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,60 @@
...
+        description: "More than 6x of the 99.95% cluster upgrade version selection-to-partial error budget is being consumed over 30m and 6h windows. Cluster upgrades are stuck for over 20 minutes without the target version reaching partial. Investigate backend_cluster_version_info on affected clusters."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-upgrade"
+    - alert: UJClusterUpgradeVersionSelectionToPartial3d6h
+      expr: |
+        hosted_control_plane_version:selection_to_partial:stuck_over_20m:burnrate6h > 1
+        and
+        hosted_control_plane_version:selection_to_partial:stuck_over_20m:burnrate3d > 1
+      for: 30m
+      labels:
+        severity: info
```
machi1990: same comment as https://github.com/Azure/ARO-HCP/pull/5752/changes#r3476800035

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,60 @@
...
+      for: 5m
+      labels:
+        severity: info
+        slo: cluster-upgrade-version-selection-to-partial
+        short_window: 5m
+        long_window: 1h
+      annotations:
+        summary: "{{ $labels.cluster }}: Cluster upgrade version selection-to-partial error budget burning rapidly (>14.4x)"
+        description: "More than 14.4x of the 99.95% cluster upgrade version selection-to-partial error budget is being consumed over 5m and 1h windows. Cluster upgrades are stuck for over 20 minutes without the target version reaching partial. Investigate backend_cluster_version_info on affected clusters."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-upgrade"
```
swiencki: Needs the `https://` prefix to render as a clickable link in IcM: `https://aka.ms/arohcp-runbook-cluster-upgrade`.

swiencki:
Before the alert goes live we need a runbook / tsg, and shorthand aka.ms url following the format in https://github.com/openshift-online/architecture/pull/79. I or another ARO Classic lead should be able to create that for you.

cgiradkar:
> Before the alert goes live we need a runbook / tsg, and shorthand aka.ms url following the format in [openshift-online/architecture#79](https://github.com/openshift-online/architecture/pull/79). I or another ARO Classic lead should be able to create that for you.

About the runbook and TSG, I have created a PR for AZure-Docs-Common. Sent you the link, PTAL

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
swiencki: Alerts need a `correlationId` annotation for IcM dedup, following the KAS pattern: `correlationId: "UJClusterUpgradeVersionSelectionToPartial/{{ $labels.cluster }}"`.

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,60 @@
...
+        and
+        hosted_control_plane_version:selection_to_partial:stuck_over_20m:burnrate1h > 14.4
+      for: 5m
+      labels:
+        severity: info
+        slo: cluster-upgrade-version-selection-to-partial
+        short_window: 5m
+        long_window: 1h
+      annotations:
+        summary: "{{ $labels.cluster }}: Cluster upgrade version selection-to-partial error budget burning rapidly (>14.4x)"
```
swiencki: Summary format should follow the PR 5765 pattern: `summary: "[HCPUpgradePartialBurn] {{ $labels.cluster }}"` with a structured description block including Window and Management Cluster fields.

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,60 @@
...
+    role: alert-rules
+  name: uj-cluster-upgrade-version-selection-to-partial-alerts
+  namespace: monitoring
+spec:
+  groups:
+  # Multi-window burn-rate alerts for cluster upgrade version selection-to-partial SLO. Sev 4 (info).
+  - name: rp-uj-cluster-upgrade-version-selection-to-partial-monitor-rules
+    rules:
+    - alert: UJClusterUpgradeVersionSelectionToPartial1h5m
+      expr: |
```
swiencki: Consider adding a minimum sample guard (e.g. `and hosted_control_plane_version:selection_to_partial:fleet_cluster_total > 0`) to prevent false-fires when no clusters are actively upgrading.

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,60 @@
...
+    prometheus: k8s
+    role: alert-rules
+  name: uj-cluster-upgrade-version-selection-to-partial-alerts
+  namespace: monitoring
+spec:
+  groups:
+  # Multi-window burn-rate alerts for cluster upgrade version selection-to-partial SLO. Sev 4 (info).
+  - name: rp-uj-cluster-upgrade-version-selection-to-partial-monitor-rules
+    rules:
+    - alert: UJClusterUpgradeVersionSelectionToPartial1h5m
```
swiencki: Alert naming should follow the PR 5765 convention: lowercase `userJourney` prefix instead of `UJ`, e.g. `userJourneyClusterUpgradeVersionSelectionToPartial1h5m`.

observability/alerts/UJClusterUpgradeCompleteMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,70 @@
...
+        and
+        hosted_control_plane_version:upgrade_complete:stuck_over_30m:burnrate1h > 14.4
+      for: 5m
+      labels:
+        severity: info
+        slo: cluster-upgrade-complete
+        short_window: 5m
+        long_window: 1h
+      annotations:
+        summary: "{{ $labels.cluster }}: Cluster upgrade complete error budget burning rapidly (>14.4x)"
```
swiencki: Same as the selection-to-partial alerts: needs `correlationId`, structured summary `[HCPUpgradeCompleteBurn] {{ $labels.cluster }}`, and `https://` on runbook_url.

observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule.yaml
machi1990: The recording rule `record: hosted_control_plane_version:selection_to_partial:stuck_over_20m_rate` already caters for this.
If there are no stuck clusters i.e no clusters actively being upgraded, then the rate will be `0` and there will be no alerts. 
there is a test case already for this https://github.com/machi1990/ARO-HCP/blob/cc93eab80e3f1a2dcab580983765371bb5792713/observability/alerts/UJClusterUpgradeVersionSelectionToPartialMonitor-prometheusRule_test.yaml#L101

observability/alerts/userJourneyClusterUpgradeCompleteRecord-prometheusRule.yaml
```diff
@@ -0,0 +1,101 @@
...
+    rules:
+    - record: hosted_control_plane_version:upgrade_complete:upgrade_pending_seconds
+      expr: |
+        (
+          time() - hosted_control_plane_version:state_first_seen_timestamp{state="desired"}
+        )
+        unless on(cluster, resource_id, subscription_id, cluster_uuid, version) (
+          backend_cluster_version_info{state="completed"} == 1
+        )
+    - record: hosted_control_plane_version:upgrade_complete:latency_seconds
```
JameelB: what's this used for?

observability/alerts/userJourneyClusterUpgradeCompleteRecord-prometheusRule.yaml
```diff
@@ -0,0 +1,101 @@
...
+  #
+  # SLO: 99.95% of cluster upgrades in the fleet complete within 30 minutes of selecting the target version.
+  # Error budget: 0.05% (1 - 0.9995)
+  #
+  # Success: desired first-seen to completed first-seen <= 1800s.
+  # In-flight: desired or partial without completed for the target version.
+  - name: user-journey-cluster-upgrade-complete-recording-rules
+    interval: 1m
+    rules:
+    - record: hosted_control_plane_version:upgrade_complete:upgrade_pending_seconds
```
JameelB: the naming here may be confusing 🤔 the metric name 'upgrade_complete' implies the state is finished, but the operation 'upgrade_pending_seconds' implies that it isn't. 

Might be better to change 'upgrade_complete' to reflect the actual state? maybe something like active_upgrade?

observability/alerts/userJourneyClusterUpgradeCompleteRecord-prometheusRule.yaml
JameelB: for naming, it's not really a delay that's being calculated but rather the duration of how long the upgrade completed? maybe `duration_seconds` would be a better fit?

observability/alerts/userJourneyClusterUpgradeVersionSelectionToPartialRecord-prometheusRule.yaml
```diff
@@ -0,0 +1,121 @@
...
+          max without(state) (backend_cluster_version_info{state="partial"}) == 1
+        )
+    - record: hosted_control_plane_version:selection_to_partial:stuck_over_20m
+      expr: |
+        hosted_control_plane_version:selection_to_partial:desired_pending_seconds > 1200
+    - record: hosted_control_plane_version:selection_to_partial:fleet_cluster_total
+      expr: |
+        count by (cluster) (
+          max by (cluster, resource_id) (backend_cluster_version_info == 1)
+        )
```
JameelB: what does this want to measure?

observability/alerts/userJourneyClusterUpgradeCompleteRecord-prometheusRule.yaml
machi1990: This is used in the dashboards. For how long the upgrade took to complete.

observability/alerts/userJourneyClusterUpgradeCompleteRecord-prometheusRule.yaml
machi1990: I see. The part `upgrade_complete` was to reflect that this recording rule if for the completeness state related alerts/dashboards and nothing to do with wether the upgrade was complete or not

observability/alerts/userJourneyClusterUpgradeVersionSelectionToPartialRecord-prometheusRule.yaml
machi1990: This counts the total number of HCP clusters par a given service `cluster`  regional deployment.

observability/grafana-dashboards/sre/user-journey/cluster-upgrade-slo-dashboard.json
```diff
@@ -0,0 +1,1490 @@
...
+      },
+      "pluginVersion": "11.6.9",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "(hosted_control_plane_version:duration_in_progress:seconds{cluster=~\"$cluster\"} > 1800) / 60",
```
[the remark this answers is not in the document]
cgiradkar: is this running in grafana? I think ther should be count clause added

observability/alerts/userJourneyClusterUpgradeRecord-prometheusRule.yaml
```diff
@@ -0,0 +1,152 @@
...
+    - record: hosted_control_plane_version:state_first_seen:timestamp
+      expr: |
+        min without(prometheus_replica) (
+          min by (cluster, resource_id, subscription_id, cluster_uuid, version, state) (
+            hosted_control_plane_version:state_first_seen:timestamp
+            or
+            timestamp(backend_cluster_version_info == 1)
+          )
+        )
+    - record: hosted_control_plane_version:fleet_cluster:total
```
JameelB: might be better using `:count` here? I seem to remember total is accumulative rather than the current number which may go up/down.

machi1990:
/hold 

Can be reviewed, I'll hold just before merge upon retesting the dashboards again. It can still be reviewed

dev-infrastructure/modules/storage/storage.bicep
```diff
@@ -1,92 +1 @@
-// Storage Account Module
```
JameelB, line 1: @machi1990 is this intentional?

observability/alerts/userJourneyClusterUpgradeReachPartialMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,74 @@
...
+      for: 5m
+      labels:
+        severity: info
+        slo: cluster-upgrade-reach-partial
+        short_window: 5m
+        long_window: 1h
+      annotations:
+        summary: "[HCPUpgradeReachPartialBurn] {{ $labels.cluster }}"
+        description: |
+          More than 14.4x of the 99.95% cluster upgrade reach-partial error budget is being consumed. Cluster upgrades are stuck for over 20 minutes without the target version reaching partial. Investigate backend_cluster_version_info on affected clusters.
```
JameelB: > Cluster upgrades are stuck for over 20 minutes without the target version reaching partial

Wdyt of having the context of what this means on the alert description itself rather than as a comment on the alert definition?

dev-infrastructure/modules/storage/storage.bicep
machi1990: An artefact of `make -C observability alerts recording`

observability/alerts/userJourneyClusterUpgradeReachPartialMonitor-prometheusRule.yaml
machi1990: Good idea

dev-infrastructure/modules/storage/storage.bicep
machi1990: or a git rebase gone wrong 🤷 Let me see

observability/alerts/userJourneyClusterUpgradeReachPartialMonitor-prometheusRule.yaml
machi1990: done

dev-infrastructure/modules/storage/storage.bicep
machi1990: removed this change, it was likely a rebase issue

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
```diff
@@ -27,3 +27,103 @@ resource arohcpAccessClusterSloRecordingRules 'Microsoft.AlertsManagement/promet
...
+  properties: {
+    scopes: [
+      azureMonitoring
+    ]
+    enabled: true
+    interval: 'PT1M'
+    rules: [
+      {
+        record: 'hosted_control_plane_version:state_first_seen:timestamp'
+        expression: 'min without (prometheus_replica) (min by (cluster, resource_id, subscription_id, cluster_uuid, version, state) (hosted_control_plane_version:state_first_seen:timestamp or (timestamp(backend_cluster_version_info == 1) and on (cluster, resource_id) (count by (cluster, resource_id) (backend_cluster_version_info{state="completed"} == 1) >= 1))))'
```
SudoBrendan: Just a note that self-referential (recursive `or`) on `state_first_seen` will lose state on prometheus restart. Not something to address I guess, but maybe we document it as a known limitation? We should also likely have some test cases for it.

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
```diff
@@ -27,3 +27,103 @@ resource arohcpAccessClusterSloRecordingRules 'Microsoft.AlertsManagement/promet
...
+        record: 'hosted_control_plane_version:desired:stuck_over_20m:count'
+        expression: 'count by (cluster) (hosted_control_plane_version:duration_in_desired:seconds > 1200) or 0 * hosted_control_plane_version:inflight_upgrade:count'
+      }
+      {
+        record: 'hosted_control_plane_version:desired:stuck_over_20m:ratio'
+        expression: 'hosted_control_plane_version:desired:stuck_over_20m:count / clamp_min(hosted_control_plane_version:inflight_upgrade:count, 1)'
+      }
+      {
+        record: 'hosted_control_plane_version:desired:stuck_over_20m:burnrate5m'
+        expression: 'round(avg_over_time(hosted_control_plane_version:desired:stuck_over_20m:ratio[5m]) / (1 - 0.9995), 0.01)'
```
SudoBrendan: why `round`? While not particularly large, this creates dead zones.

observability/alerts/userJourneyClusterUpgradeRecord-prometheusRule.yaml
```diff
@@ -0,0 +1,201 @@
...
+    # Share of in-flight upgrades stuck in desired over 20 minutes (reach-partial error rate).
+    - record: hosted_control_plane_version:desired:stuck_over_20m:ratio
+      expr: |
+        hosted_control_plane_version:desired:stuck_over_20m:count
+        /
+        clamp_min(hosted_control_plane_version:inflight_upgrade:count, 1)
+    # Reach-partial error-budget burn rate over 5 minutes (99.95% SLO).
+    - record: hosted_control_plane_version:desired:stuck_over_20m:burnrate5m
+      expr: |
+        round(
```
SudoBrendan: again - yea - I think we remove all `round` everywhere. That's helpful in maybe a dashboard, but not in what we record.

observability/alerts/userJourneyClusterUpgradeRecord-prometheusRule.yaml
```diff
@@ -0,0 +1,201 @@
...
+          avg_over_time(hosted_control_plane_version:in_progress:stuck_over_30m:ratio[6h])
+          / (1 - 0.9995),
+        0.01)
+    # Complete error-budget burn rate over 3 days.
+    - record: hosted_control_plane_version:in_progress:stuck_over_30m:burnrate3d
+      expr: |
+        round(
+          avg_over_time(hosted_control_plane_version:in_progress:stuck_over_30m:ratio[3d])
+          / (1 - 0.9995),
+        0.01)
```
SudoBrendan: here even - when we round, it's to a significant figure lower than the threshold; we'd need to round to 0.0001 or something to even measure what we're trying to detect.

I'll stop ranting here - no `round` please :)

observability/alerts/userJourneyClusterUpgradeRecord-prometheusRule.yaml
```diff
@@ -0,0 +1,201 @@
...
+      expr: |
+        hosted_control_plane_version:in_progress:stuck_over_30m:count
+        /
+        clamp_min(hosted_control_plane_version:inflight_upgrade:count, 1)
+    # Complete error-budget burn rate over 5 minutes (99.95% SLO).
+    - record: hosted_control_plane_version:in_progress:stuck_over_30m:burnrate5m
+      expr: |
+        round(
+          avg_over_time(hosted_control_plane_version:in_progress:stuck_over_30m:ratio[5m])
+          / (1 - 0.9995),
```
SudoBrendan: I'm really questioning averages / percentages in general here. If I'm understanding this whole concept - we're counting:

```
stuck clusters / in-flight clusters
```

and this is on a per-mgmt-cluster level, right? The max number of clusters even possible is 100 (your max denominator), so even if by some miracle all customers running HCPs in a MGMT cluster issue their upgrades simultaneously in a 30m window - _one failure_, 1/100 = 1% = fires at the highest possible burn rate - and all lower thresholds are irrelevant.

I think what likely makes more sense is something way simpler:

1. Single stuck cluster for a long time (like, 1h?) - info
2. Multiple clusters (idk, 3?) stuck for longer than 30m - warning (granted, I still think the liklihood of this happening is super low... unless we count SRE-initiated upgrades on z-stream... but I can also see this not scaling at all if we discover some way that customers find to wreck how customer-issued upgrades work and they are stuck perpetually until they action something - permissions (?) - which either happens on the order of days, if at all... so I can also see this over years being a perpetual-fire/never ending alert that is simply noise)

So - idk - maybe we just go with 1 - a stopgap alert to start. If it's not sufficient we add additionals later.

SudoBrendan (changes requested):
I have to stop here for the day - but a few important items to address.

> Runbook not yet there but they'll be created

don't be like me, learn from my mistakes. /hold until that runbook is live and the link resolves :)

cgiradkar:
lgtm

observability/alerts/userJourneyClusterUpgradeRecord-prometheusRule.yaml
machi1990: This takes onto account all sort of upgrades: user initiated and automated upgrades. 
I tried to follow the concern in this and I couldn't really grok it, what are is the actual issue about the alert knowing that you can have N upgrades triggered in parallel across the whole fleet (all regions, all management cluster)?

machi1990:
> I have to stop here for the day - but a few important items to address.
> 
> > Runbook not yet there but they'll be created
> 
> don't be like me, learn from my mistakes. /hold until that runbook is live and the link resolves :)

A runbook/tsg PR was shared in DM, it's being worked on parallel for now a link to it lands 404.

observability/alerts/userJourneyClusterUpgradeRecord-prometheusRule.yaml
machi1990: dropped the round all over

observability/alerts/userJourneyClusterUpgradeRecord-prometheusRule.yaml
SudoBrendan: Sorry - I had to do a double-take on this, I was partially wrong - because we're measuring a SVC cluster metric `backend_cluster_version_info`, not something measured from within the management cluster - but regardless - I think my point still stands on the math.

Let me see if I can explain it better.

---

In this PR - the change you suggest is that each Region's svc Azure Monitor Workspace (Prometheus) instance configures a RecordingRule. At a regional (not fleet) scope - each AMW individually evaluates your query for itself, for it's region alone, and publishes the aggregation as a new metric.

This means that this PR suggests a 99.95% SLO on a per region basis. Put another way - unless we sustain 2,000 cluster upgrades per region all the time (your suggested denominator - in-flight upgrades), 1 outlier cluster in a region failing to upgrade in time will fire past our highest threshold - it is not possible mathematically to burn slower than your highest threshold with less than 2,000 concurrent upgrades happening in a given region.

So I mean to say - until the point we anticipate 2,000 simultaneous upgrades concurrently at all times in all regions, if we expect a 99.95% success rate on this metric - I think it's better for us to actually alert on a single failure; it's the same alert. Alternatively - we can reduce the 99.95% rate, and say we only want 90% success (this means we'd expect 10 concurrent upgrades at all times to properly measure burn beyond a single cluster failure)...

But the real point I'm trying to make - even 10 concurrent upgrades - do we expect that to be happening a majority of the time in a given region? My guess would be a few a day in a region on average initiated by customers, and I'd even wager it'll be rare to get concurrent upgrades by customers at all for a while - and probably never in our smaller regions. SRE performing fleet wide maintenance would definitely give us a bigger denominator for brief times, but we're still going to be bound by the number of OCP releases possible - they don't happen every day or every hour.

TL;DR - I don't see there being enough upgrade traffic possible to warrant this metric as implemented (a ratio), it's equivalent to alerting on a single failure, and medium/slow burns are not possible mathematically.

machi1990: Thanks for the clarification. Right, at low traffic these will be equivalent to "alert if any upgrade duration is higher than the threshold". Do you reckon that we'll be better off with dropping the burn rate and take a similar approach like https://github.com/Azure/ARO-HCP/pull/5952 ?

machi1990: I am working on a second commit that contains the set of changes based on my understanding of this and I'll that kind of alerting matches your expectation in https://github.com/Azure/ARO-HCP/pull/5752#discussion_r3539885700

machi1990: Discussed via DM and a new commit has been added to reflect the agreed upon approach

observability/grafana-dashboards/sre/user-journey/cluster-upgrade-slo.json
```diff
@@ -0,0 +1,318 @@
...
+  "id": null,
+  "links": [],
+  "panels": [
+    {
+      "gridPos": { "h": 4, "w": 24, "x": 0, "y": 0 },
+      "id": 1,
+      "options": {
+        "code": { "language": "plaintext", "showLineNumbers": false, "showMiniMap": false },
+        "content": "### Cluster Upgrade SLOs\n\nPer-upgrade observability from `backend_cluster_version_info` via `hosted_control_plane_upgrade:*` recording rules.\n\n| SLO | Target | Alert |\n|-----|--------|-------|\n| **Reach partial** | `partial` within 20m | `userJourneyClusterUpgradeStuckInDesired` (>20m in `desired`) |\n| **Upgrade complete** | `completed` within 30m | `userJourneyClusterUpgradeStuckInProgress` (>30m in progress) |\n\n**Runbook:** [https://aka.ms/arohcp-runbook-cluster-upgrade](https://aka.ms/arohcp-runbook-cluster-upgrade)",
+        "mode": "markdown"
```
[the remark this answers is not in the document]
machi1990: The redirection is there and the PR is being worked on.

observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,49 @@
...
+      labels:
+        severity: info
+        slo: cluster-upgrade-reach-partial
+      annotations:
+        summary: "[HCPUpgradeStuckInDesired] {{ $labels.resource_id }}"
+        description: |
+          Cluster upgrade target version {{ $labels.version }} has been in desired for over 20 minutes without reaching partial. Partial is when the target version becomes active (recognized as active) on the HCP cluster; the upgrade is in progress but not yet complete. Investigate backend_cluster_version_info on the affected cluster.
+          Service Cluster: {{ $labels.cluster }}
+        runbook_url: https://aka.ms/arohcp-runbook-cluster-upgrade
+        correlationId: "userJourneyClusterUpgradeStuckInDesired/{{ $labels.resource_id }}/{{ $labels.version }}"
```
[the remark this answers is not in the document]
machi1990: The redirection is there and there is WIP PR being worked on

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
```diff
@@ -27,3 +27,37 @@ resource arohcpAccessClusterSloRecordingRules 'Microsoft.AlertsManagement/promet
...
+    enabled: true
+    interval: 'PT1M'
+    rules: [
+      {
+        record: 'hosted_control_plane_upgrade:upgrade_eligible:info'
+        expression: '((count by (cluster, resource_id, subscription_id, cluster_uuid) (count by (cluster, resource_id, subscription_id, cluster_uuid, version) (backend_cluster_version_info == 1)) >= 2) and on (cluster, resource_id) (count by (cluster, resource_id) (backend_cluster_version_info{state="completed"} == 1) >= 1)) * 0 + 1'
+      }
+      {
+        record: 'hosted_control_plane_upgrade:version_state_first_seen:timestamp'
+        expression: 'min without (prometheus_replica) (min by (cluster, resource_id, subscription_id, cluster_uuid, version, state) (hosted_control_plane_upgrade:version_state_first_seen:timestamp or (timestamp(backend_cluster_version_info{state=~"desired|partial"} == 1) and on (cluster, resource_id) (hosted_control_plane_upgrade:upgrade_eligible:info == 1)) or timestamp(backend_cluster_version_info{state="completed"} == 1)))'
```
SudoBrendan: If I'm reading this correctly - this persists series indefinitely due to the self-reference / recursion. After a single upgrade for a single cluster, we'll have:

{resource_id="foo", version="4.21.0", state="desired"} TIME_FIRST_SEEN
{resource_id="foo", version="4.21.0", state="partial"} TIME_FIRST_SEEN
{resource_id="foo", version="4.21.0", state="completed"} TIME_FIRST_SEEN

With two upgrades, we'll see double.

{resource_id="foo", version="4.21.0", state="desired"} TIME_FIRST_SEEN
{resource_id="foo", version="4.21.0", state="partial"} TIME_FIRST_SEEN
{resource_id="foo", version="4.21.0", state="completed"} TIME_FIRST_SEEN
{resource_id="foo", version="4.21.1", state="desired"} TIME_FIRST_SEEN
{resource_id="foo", version="4.21.1", state="partial"} TIME_FIRST_SEEN
{resource_id="foo", version="4.21.1", state="completed"} TIME_FIRST_SEEN

For 10000 clusters, each upgrading 1/wk (52 times) on z-stream, we'll have 1.5 million different series in a year's time - and nearly all of them will be irrelevant (old upgrades).

Regardless of if the upgrade completed or not - prometheus will continue to track the series and record the data in the interval specified - we'll be up to 1.5 million samples a minute (PT1M) - recording the same metric value for upgrades that completed months/years prior.

---

We should add a guard to kill old series. What about:

 hosted_control_plane_upgrade:version_state_first_seen:timestamp
            **unless on(cluster, resource_id, subscription_id, cluster_uuid, version) (
              backend_cluster_version_info{state="completed"} == 1
            )**...


In addition - we should probably remove this unused part

`or timestamp(backend_cluster_version_info{state="completed"} == 1`

This would make it so we only record the most recent upgrade, and only until it's done (no lingering artifacts).

observability/recording-rules/userJourneyClusterUpgrade-recordingRule_test.yaml
```diff
@@ -0,0 +1,145 @@
...
+  - expr: hosted_control_plane_upgrade:duration_in_desired:seconds{cluster="svc-1", resource_id="/sub/3/rg/c4", version="4.19.20", state="desired"}
+    eval_time: 35m
+    exp_samples: []
+  - expr: hosted_control_plane_upgrade:in_progress:count{cluster="svc-1"}
+    eval_time: 35m
+    exp_samples:
+    - labels: 'hosted_control_plane_upgrade:in_progress:count{cluster="svc-1"}'
+      value: 0
+# ========================================
+# Test 4: Complete — desired transitions to completed within 30m (upgrade)
```
SudoBrendan, line 123: Adding a test here to verify that the series disappears after upgrade completes would ensure we don't leak now or in the future.

observability/recording-rules/userJourneyClusterUpgrade-recordingRule_test.yaml
```diff
@@ -0,0 +1,145 @@
+rule_files:
+- userJourneyClusterUpgrade-recordingRule.yaml
+evaluation_interval: 1m
+tests:
+# ========================================
+# Test 1: Reach partial — desired transitions to partial within 20m (upgrade)
```
SudoBrendan, line 6: This could also use a test to ensure we get to the `state="desired"` clause, so we don't fire for "stuck in desired" for things actively progressing.

observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,49 @@
...
+  # At current upgrade volumes, alert directly on duration instead of fleet burn-rate.
+  #
+  # In absence of https://issues.redhat.com/browse/ARO-22686 (reporting of control plane
+  # version history for versions lower than 4.22), cluster upgrade completeness is directly
+  # impacted by data plane state — e.g. no worker nodes means the upgrade will not reach
+  # completed; missing console component in external auth prevents CVO from marking the
+  # upgrade complete. The complete alert may therefore fire due to missing data plane
+  # configuration rather than control plane failure. The runbook accommodates this.
+  - name: rp-user-journey-cluster-upgrade-monitor-rules
+    rules:
```
SudoBrendan, line 27: Alerts should probably define a `for`, even if it's brief, all others do.

SudoBrendan (changes requested):
We for sure need to fix the leak in the series we're recording.

observability/grafana-dashboards/sre/user-journey/cluster-upgrade-slo.json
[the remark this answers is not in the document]
SudoBrendan: be sure to hold this PR until the runbook merges - you are sending this alert to the production SRE queue. The runbook must resolve and be complete before ICM alerts go to the SRE team.

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
machi1990: Good catch

swiencki (commented):
observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml - Use numeric severity `"4"` instead of deprecated `info` string per the CEN standard established in PR #5717.

observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - alert: userJourneyClusterUpgradeStuckInDesired
+      expr: |
+        hosted_control_plane_upgrade:duration_in_desired:seconds > 1200
+      for: 5m
+      labels:
+        severity: "4"
+        slo: cluster-upgrade-reach-partial
```
swiencki, line 34: The title annotation includes cluster:{{ $labels.cluster }} but the IcM connector already prefixes the service cluster via IcM.Title, duplicating the cluster name in every incident title.

machi1990: > The title annotation includes cluster:{{ $labels.cluster }} but the IcM connector already prefixes the service cluster via IcM.Title, duplicating the cluster name in every incident title.

Is this " title: '[HCPUpgradeStuckInDesired] {{ $labels.resource_id }} version:{{ $labels.version }} cluster:{{ $labels.cluster }}'" in https://github.com/machi1990/ARO-HCP/blob/716c1c057d2b55edf16e30b44dd81b6ee16bb861/dev-infrastructure/modules/metrics/rules/generatedRPPrometheusAlertingRules.bicep what you are referring to? If so, that's a property of the generator not an artefact of this PR.

If that's what not the comment is about, perhaps are you referring to 
` summary: "[HCPUpgradeStuckInDesired] {{ $labels.resource_id }}"` in the file https://github.com/machi1990/ARO-HCP/blob/716c1c057d2b55edf16e30b44dd81b6ee16bb861/observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml ? if so, .resource_id is different from .cluster

observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: monitoring.coreos.com/v1
```
swiencki, line 1: Consider a fleet-level alert (e.g., "more than N upgrades stuck simultaneously across the fleet") so a z-stream rollout failure surfaces as a single incident rather than N separate per-cluster pages.

observability/recording-rules/userJourneyClusterUpgrade-recordingRule.yaml
```diff
@@ -0,0 +1,126 @@
+apiVersion: monitoring.coreos.com/v1
```
swiencki, line 1: Consider exposing a "upgrade completed" recording rule or dashboard signal so SRE can confirm all-clear after an incident and SLO reporting can measure success rate over time, not just failure duration.

machi1990: Yes; there will be a panel in the  dashboard that show all the clusters versions.

observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
machi1990: So myself and @SudoBrendan had a discussion about this here https://github.com/Azure/ARO-HCP/pull/5752#discussion_r3516185592 where we agreed to drop the fleet wide slo in favor of the approach taken here. Depending on traffic and how this initial v1 matures, other sort of alerts can be added as follow ups

swiencki: Have we ran a historical production report on what this looks like? CI might bump those values and allow for early detection.

machi1990: Just visibility for others, myself, @swiencki & @SudoBrendan are talking about this.

> Have we ran a historical production report on what this looks like? CI might bump those values and allow for early detection.

Shared gathered via kusto in DM. In a span of 30 minutes window, the current data show that the max concurrent traffic we've in prod is 30 upgrades in one given region.

machi1990:
/hold 

Until the runbook is available

observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,53 @@
...
+        severity: "4"
+        slo: cluster-upgrade-reach-partial
+        component: backend
+      annotations:
+        summary: "{{ $labels.cluster }}: Cluster upgrade on {{ $labels.resource_id }} to version {{ $labels.version }} not progressing"
+        description: |
+          Cluster upgrade target version {{ $labels.version }} has been in desired for over 20 minutes without reaching partial. With the alert pending for 5 minutes, paging starts after ~25 minutes. Partial is when the target version becomes active (recognized as active) on the HCP cluster; the upgrade is in progress but not yet complete. Investigate backend_cluster_version_info on the affected cluster.
+          Service Cluster: {{ $labels.cluster }}
+        runbook_url: https://aka.ms/arohcp-runbook-cluster-upgrade
+        correlationId: "userJourneyClusterUpgradeStuckInDesired/{{ $labels.cluster }}{{ $labels.resource_id }}/{{ $labels.version }}"
```
swiencki, line 42: Missing a `/` separator between cluster and resource_id; currently produces `svc-1/sub/2/rg/c2` which looks like one path instead of two fields. Should be `"userJourneyClusterUpgradeStuckInDesired/{{ $labels.cluster }}/{{ $labels.resource_id }}/{{ $labels.version }}"`.

machi1990: Interesting;

I did this intentionally to avoid it looking like *svc-1//sub/2/rg/c2*

I don't know how SRE's process is like, so I'll follow your guidance on how to structure this here

machi1990: Following internal discussion we sticked with `svc-1//sub/2/rg/c2` and I've thus reverted the change I did yesterday for this comment.

observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
machi1990: We decided to go with what's here

swiencki:
/lgtm
/approved
Let's get the TSG in before this merges.

machi1990:
/unhold

The runbook url is live https://aka.ms/arohcp-runbook-cluster-upgrade

observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,53 @@
...
+  groups:
+  # Per-upgrade stuck alerts for cluster upgrade SLOs. Sev 4.
+  # At current upgrade volumes, alert directly on duration instead of fleet burn-rate.
+  #
+  # In absence of https://issues.redhat.com/browse/ARO-22686 (reporting of control plane
+  # version history for versions lower than 4.22), cluster upgrade completeness is directly
+  # impacted by data plane state — e.g. no worker nodes means the upgrade will not reach
+  # completed; missing console component in external auth prevents CVO from marking the
+  # upgrade complete. The complete alert may therefore fire due to missing data plane
+  # configuration rather than control plane failure. The runbook accommodates this.
```
machi1990: There are two backports PRs:
- https://github.com/openshift/hypershift/pull/9055
- https://github.com/openshift/hypershift/pull/9054

Which once merged, will get us along in the right direction. I'll check the ETA of their merge today.

machi1990: Waiting on to be merged https://github.com/openshift/hypershift/pull/9211 and be bumped in aro-hcp to reduce chances of this alert firing because of customer misconfiguration.

machi1990:
/hold 

See https://github.com/Azure/ARO-HCP/pull/5752#discussion_r3713567895

machi1990:
Interesting that when I synced the dashboards, the `transformations` field is being ignored. 
It seems like a bug in the json marshalling round trip

machi1990:
> Interesting that when I synced the dashboards, the `transformations` field is being ignored. It seems like a bug in the json marshalling round trip

@swiencki will have a look at this

swiencki:
> > Interesting that when I synced the dashboards, the `transformations` field is being ignored. It seems like a bug in the json marshalling round trip
> 
> @swiencki will have a look at this

See https://github.com/Azure/ARO-Tools/pull/281

observability/alerts/userJourneyClusterUpgradeMonitor-prometheusRule.yaml
machi1990: We are good

machi1990:
/hold cancel 

https://github.com/openshift/hypershift/pull/9211 has been merged and bumped in aro-hcp via https://github.com/Azure/ARO-HCP/pull/6471

PR #5754, opened 2026-06-23
revert user journey aks apiserver alerts

stevekuznetsov:


bennerv:
/lgtm 

Currently paging wrong queue, queue ID is incorrect (hardcoded to a number), alert is changing for the same cluster flagging rather than mitigating / keeping the same incident open.

stevekuznetsov:
Prow queue is empty and we're passing the basic tests. Green-button for this one.

PR #5755, opened 2026-06-23
fix(observability): route KAS availability alerts to SRE lane at info severity

swiencki:
### What

Move UJ KAS availability alerts from the RP lane to the SRE lane and set severity to info (non-paging).

### Why

The RP lane did not have routing rules configured in IcM, causing alerts to page the wrong queue (MSFT triage). Routing through the SRE lane uses the established SRE-to-RP routing path.

### Testing

Lane reassignment and severity label change only; no functional change to alert expressions or recording rules.


SudoBrendan:
We were specifically asked by Angus to page SRE for HCP outages - we can update the routing, but unless we get signoff for the severity change by managers, let's keep it the same.

swiencki:
Superseded: rebasing on top of #5754 revert

PR #5758, opened 2026-06-23
chore(monitoring): remove RP IcM lane

swiencki:
Depends on #5754 and #5756.

### What

- Remove the RP IcM action group, bicep module, and observability config
- Remove RP routing ID from config schema and defaults
- The RP lane is no longer used; all SRE alerts route through the SRE lane

### Why

- The RP lane was added as a dormant lane but caused routing issues when alerts were sent to it
- All alerts that need SRE attention now route through ARO-HCP-SRE directly
- Removing dead infrastructure avoids future confusion

### Testing

- `cd config && make materialize` passes
- No RP references remain in bicep, config, or observability files


swiencki:
Not required.

PR #5760, opened 2026-06-23
fix: return 503 status code when CS provisioning shard is not yet registered

bennerv:
no jira

### What

Returns a 503 status code instead of 404 when a user attempts to create a cluster but CS has no active provisioning shard.

### Why

> The fleet-controller's 2-minute CosmosDB polling cycle created a ~117-second gap between the `fleet-registration` job writing the Stamp and the CS provision shard becoming active; the test attempted cluster creation during this window, CS returned 404 'Unable to find shard', and the frontend incorrectly propagated this as an ARM `ResourceNotFound` error.

By returning a 503, the [azure-sdk-for-go default retry policy](https://github.com/Azure/azure-sdk-for-go/blob/922efdda6cf2a775b8a76f292562fd5a2c0940dc/sdk/azcore/runtime/policy_retry.go#L52) will attempt retries for 5 minutes, which will alleviate the above while #5759 speeds up the registration. 

### Testing

unit tests added

### PR Checklist
- [ ] Self-reviewed the diff

### DIFF
--- a/internal/ocm/convert.go
+++ b/internal/ocm/convert.go
@@ -724,6 +724,12 @@ func CSErrorToCloudError(err error, resourceID *azcorearm.ResourceID) *arm.Cloud
 				arm.CloudErrorCodeInvalidRequestContent,
 				"", "%s", ocmError.Reason())
 		case http.StatusNotFound:
+			if strings.Contains(ocmError.Reason(), "Unable to find shard") {
+				return arm.NewCloudError(
+					http.StatusServiceUnavailable,
+					arm.CloudErrorCodeServiceUnavailable,
+					"", "Capacity is currently restricted, please try again later")
+			}
 			if resourceID != nil {
 				return arm.NewResourceNotFoundError(resourceID)
 			}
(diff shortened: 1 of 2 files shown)


internal/ocm/convert.go
```diff
@@ -724,6 +724,12 @@ func CSErrorToCloudError(err error, resourceID *azcorearm.ResourceID) *arm.Cloud
 				arm.CloudErrorCodeInvalidRequestContent,
 				"", "%s", ocmError.Reason())
 		case http.StatusNotFound:
+			if strings.Contains(ocmError.Reason(), "Unable to find shard") {
+				return arm.NewCloudError(
+					http.StatusServiceUnavailable,
+					arm.CloudErrorCodeServiceUnavailable,
+					"", "Service Unavailable")
```
geoberle: do we want to give some hint about the reason?
i've seen a lot of azure services disclosing that they are out of capacity

bennerv: `Capacity is currently restricted, please try again later`

PR #5761, opened 2026-06-23
dev-infrastructure: allow adding a ceiling to alert severity

stevekuznetsov:
We do not want to get high-severity alerting in INT and STG, regardless of what severity the actual alert has.


bennerv (approved):
/lgtm

stevekuznetsov:
OK, this needs to go in so it doesn't ruin my day. Sigh.

PR #5775, opened 2026-06-24
refactor(logging): standardize controller logger seeding across backend and kube-applier

deads2k:
Move LoggableKey + AddLoggerValues helper from
backend/pkg/controllers/controllerutils to internal/utils/context.go so both backend and kube-applier can share one contract. Give every kube-applier workqueue key a GetResourceID + AddLoggerValues method, add XxxControllerName constants for each kube-applier controller, and seed per-key loggers in every worker loop so each reconcile carries uniform structured-log fields (subscription_id, resource_group, resource_id, hcp_cluster_name) — matching what the backend's genericWatchingController already produces. Encode the rule in CLAUDE.md so future controllers inherit the pattern.


stevekuznetsov (approved):
/lgtm
/approve

PR #5778, opened 2026-06-24
fix: ToServiceProviderNodePoolResourceIDString embedded provider name…

deads2k:
…space twice

The helper concatenated NodePoolResourceType.String() (which already carries the Microsoft.RedHatOpenShift namespace prefix) into a path that already had the providers segment, producing keys like
.../providers/microsoft.redhatopenshift/hcpopenshiftclusters/<cluster>/microsoft.redhatopenshift/hcpopenshiftclusters/nodepools/<np>/...

The informer cache stores ServiceProviderNodePools under the canonical ARM resource ID, so serviceProviderNodePoolLister.Get always returned NotFound.

Switch to the same Types[-1] pattern the sibling helpers use, and add an integration test that seeds an SPNP, starts the informer with a short relist, and asserts the SPNP is observable via both lister.List and lister.Get within 30s. The test would have failed before this fix.

### DIFF
--- a/internal/api/types_cosmosdata.go
+++ b/internal/api/types_cosmosdata.go
@@ -25,6 +25,13 @@ import (
 
 type CosmosMetadata = arm.CosmosMetadata
 
+// The ToXxxResourceIDString helpers form ARM resource ID strings by
+// delegating to the parent helper and appending the final segment. The
+// segment uses the leaf type name (Types[len-1]) for nested types so that
+// the provider namespace is added exactly once at the top of the chain.
+// Authoring each level via its parent makes it impossible to accidentally
+// produce a non-canonical key — which the cache uses to look up objects.
+
 func ToResourceGroupResourceIDString(subscriptionName, resourceGroupName string) string {
 	return strings.ToLower(path.Join("/subscriptions", subscriptionName, "resourceGroups", resourceGroupName))
 }
@@ -39,8 +46,7 @@ func ToClusterResourceID(subscriptionName, resourceGroupName, clusterName string
 
 func ToClusterResourceIDString(subscriptionName, resourceGroupName, clusterName string) string {
 	return strings.ToLower(path.Join(
-		"/subscriptions", subscriptionName,
-		"resourceGroups", resourceGroupName,
+		ToResourceGroupResourceIDString(subscriptionName, resourceGroupName),
 		"providers", ClusterResourceType.String(), clusterName,
 	))
 }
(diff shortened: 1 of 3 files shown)


internal/api/types_cosmosdata.go
```diff
@@ -89,19 +89,23 @@ func ToOperationResourceIDString(subscriptionName, operationName string) string
...
 }
+
+// leafTypeName returns the trailing segment of an ARM ResourceType (the
+// part after the last slash). Using it in the per-level helpers prevents
+// callers from accidentally embedding the full `namespace/type/...` form
+// twice in the same ID — see the original ToServiceProviderNodePoolResourceIDString
+// bug.
+func leafTypeName(rt azcorearm.ResourceType) string {
+	return rt.Types[len(rt.Types)-1]
+}
```
machi1990, line 111: The type names are defined here
https://github.com/deads2k/ARO-HCP/blob/2a5e3f2447888a748cddb03f1956ea4130305e9f/internal/api/registry.go#L67 

Why don't we use them directly for each ToXXResourceID func?

internal/api/types_cosmosdata.go
```diff
@@ -51,10 +57,8 @@ func ToNodePoolResourceID(subscriptionName, resourceGroupName, clusterName, node
 
 func ToNodePoolResourceIDString(subscriptionName, resourceGroupName, clusterName, nodePoolName string) string {
 	return strings.ToLower(path.Join(
-		"/subscriptions", subscriptionName,
-		"resourceGroups", resourceGroupName,
-		"providers", ClusterResourceType.String(), clusterName,
-		NodePoolResourceType.Types[len(NodePoolResourceType.Types)-1], nodePoolName,
+		ToClusterResourceIDString(subscriptionName, resourceGroupName, clusterName),
+		leafTypeName(NodePoolResourceType), nodePoolName,
```
machi1990, line 61: In relation to comment https://github.com/Azure/ARO-HCP/pull/5778#discussion_r3472584709 can we use https://github.com/deads2k/ARO-HCP/blob/2a5e3f2447888a748cddb03f1956ea4130305e9f/internal/api/registry.go#L71 directly here? 

Same comment for all the other methods

deads2k: > In relation to comment [#5778 (comment)](https://github.com/Azure/ARO-HCP/pull/5778#discussion_r3472584709) can we use https://github.com/deads2k/ARO-HCP/blob/2a5e3f2447888a748cddb03f1956ea4130305e9f/internal/api/registry.go#L71 directly here?
> 
> Same comment for all the other methods

when I came to read the code, I found the singulars confusing.  I prefer a single name in the structured type

machi1990: I was suggesting to use the const `NodePoolResourceTypeName` direct here instead of `leafTypeName(NodePoolResourceType)`.

PR #5786, opened 2026-06-25
New e2e test for complex customer scenario.

miquelsi:
https://redhat.atlassian.net/browse/ARO-22744

### What

This new E2E test aggregates the following features in one cluster+nodepools scenario:
 - External OIDC provider via ExternalAuth
 - Cilium CNI with kube-proxy replacement
 - ETCD data encryption with customer-managed keys
 - ETCD disk-level encryption with platform-managed keys
 - Internal image registry disabled
 - API IP address access control (authorized CIDRs)
 - KeyVaultVisibility set to Private

### Why

Without this kind of testing, we can’t assume that the complex features work well together.

### Testing

N/A

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/e2e/cluster_create_feature_aggregation.go
+++ b/test/e2e/cluster_create_feature_aggregation.go
@@ -1,0 +1,495 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+
+	operatorclient "github.com/openshift/client-go/operator/clientset/versioned"
+
+	hcpsdk20240610preview "github.com/Azure/ARO-HCP/test/sdk/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
+	hcpsdk20251223preview "github.com/Azure/ARO-HCP/test/sdk/v20251223preview/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+// This test aggregates the following features in one cluster+nodepools scenario:
+// - External OIDC provider via ExternalAuth
+// - Cilium CNI with kube-proxy replacement
+// - ETCD data encryption with customer-managed keys
+// - ETCD disk-level encryption with platform-managed keys
+// - Internal image registry disabled
+// - API IP address access control (authorized CIDRs)
+// - KeyVaultVisibility set to Private
+var _ = Describe("Customer", func() {
+	It("should be able to create a cluster and node pools with aggregated advanced features",
+		labels.RequireNothing,
+		labels.Critical,
+		labels.Positive,
+		labels.AroRpApiCompatible,
+		labels.CreateCluster,
+		labels.MIContainers(1),
+		func(ctx context.Context) {
+			const (
+				customerClusterName       = "agg-cluster"
+				customerNodePoolNameA     = "agg-np-a"
+				customerNodePoolNameB     = "agg-np-b"
+				customerExternalAuthName  = "agg-extauth"
+				externalAuthSubjectPrefix = "prefix-"
+				ciliumNamespace           = "kube-system"
+				ciliumVersion             = "1.19.2"
+			)
+
+			tc := framework.NewTestContext()
+
+			if tc.UsePooledIdentities() {
+				err := tc.AssignIdentityContainers(ctx, 1, framework.IdentityContainerAssignmentRetryInterval)
+				Expect(err).NotTo(HaveOccurred(), "failed to assign pooled identity containers")
+			}
+
+			By("creating a resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "feature-aggregation", tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for feature aggregation test")
+
+			By("building cluster parameters for aggregated feature coverage")
+			clusterParams := framework.NewDefaultClusterParams20251223()
+			clusterParams.ClusterName = customerClusterName
+			clusterParams.ManagedResourceGroupName = framework.SuffixName(*resourceGroup.Name, "-managed", 64)
+			clusterParams.Network.NetworkType = "Other"
+			clusterParams.Network.PodCIDR = "10.255.0.0/16"
+			clusterParams.Network.ServiceCIDR = "172.30.0.0/16"
+			clusterParams.Network.MachineCIDR = "10.0.0.0/16"
+			clusterParams.Network.HostPrefix = 23
+			clusterParams.EncryptionKeyManagementMode = string(hcpsdk20251223preview.EtcdDataEncryptionKeyManagementModeTypeCustomerManaged)
+			clusterParams.EncryptionType = string(hcpsdk20251223preview.CustomerManagedEncryptionTypeKms)
+			clusterParams.KeyVaultVisibility = "Private"
+			clusterParams.ImageRegistryState = string(hcpsdk20251223preview.ClusterImageRegistryStateDisabled)
+
+			By("creating customer resources with private key vault support")
+			clusterParams, err = tc.CreateClusterCustomerResources20251223(
+				ctx,
+				resourceGroup,
+				clusterParams,
+				map[string]any{
+					"privateKeyVault": true,
+				},
+				TestArtifactsFS,
+				framework.RBACScopeResourceGroup,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create customer resources for aggregated feature cluster")
+
+			By("generating SSH key pair for authorized CIDR VM")
+			sshPublicKey, _, err := framework.GenerateSSHKeyPair()
+			Expect(err).NotTo(HaveOccurred(), "failed to generate SSH key pair for authorized CIDR VM")
+
+			By("deploying a VM to source the authorized public IP")
+			vmName := fmt.Sprintf("%s-test-vm", customerClusterName)
+			// Use a restriction-aware VM size selector to reduce SkuNotAvailable flakiness.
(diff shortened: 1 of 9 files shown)


test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,423 @@
...
+			)
+			Eventually(func(g Gomega) {
+				_, cmdErr := framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, disableKubeProxyCmd, 2*time.Minute)
+				g.Expect(cmdErr).NotTo(HaveOccurred(), "failed to disable kube-proxy from authorized VM")
+			}, 5*time.Minute, 10*time.Second).Should(Succeed(), "kube-proxy disable patch should succeed from authorized VM")
+
+			By("installing cilium with kube-proxy replacement enabled from the authorized VM")
+			installCiliumCmd := fmt.Sprintf(
+				`echo '%s' | base64 -d > /tmp/admin-kubeconfig; `+
+					`if ! command -v helm >/dev/null 2>&1; then curl -fsSL https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash; fi; `+
```
mbukatov: I' not fan of this kind of installation. But this is not the only place when we do this.

test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,423 @@
...
+			nodePoolErrB := tc.CreateNodePoolFromParam20251223(
+				ctx,
+				GinkgoLogr,
+				*resourceGroup.Name,
+				clusterParams.ManagedResourceGroupName,
+				customerClusterName,
+				nodePoolParamsB,
+				framework.NodePoolCreationTimeout,
+			)
+			Expect(errors.Join(nodePoolErrA, nodePoolErrB)).NotTo(HaveOccurred(), "failed to create aggregated feature node pools for cluster %q", customerClusterName)
```
mbukatov: If this fails on a CNI related issue, you won't be able to tell that.

test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,423 @@
...
+			_, err = framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, waitNodesReadyCmd, 16*time.Minute)
+			Expect(err).NotTo(HaveOccurred(), "all nodes should become Ready from authorized VM checks")
+			waitCiliumReadyCmd := fmt.Sprintf(
+				`echo '%s' | base64 -d > /tmp/admin-kubeconfig && kubectl --kubeconfig=/tmp/admin-kubeconfig -n %s rollout status daemonset/cilium --timeout=10m`,
+				adminKubeconfigB64,
+				ciliumNamespace,
+			)
+			_, err = framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, waitCiliumReadyCmd, 11*time.Minute)
+			Expect(err).NotTo(HaveOccurred(), "cilium daemonset should become Ready from authorized VM checks")
+		},
```
mbukatov, line 493: Hmm, writing more complex use case with a private cluster API endpoint like this one is pain to work with via the VM. Esp. since we have to either duplicate the cases, use error prone or insecure methods, or in worst case, don't run verification at all (eg. we have quite a few verifiers we could run here, but we don't, because we can't run them on the vm). This needs to be addressed in the future.

mbukatov (commented):
/lgtm

There are few things to be improved wrt better support for these kind of use cases, but having this E2E test case is better than not having this use case covered.

test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,423 @@
...
+				`echo '%s' | base64 -d > /tmp/admin-kubeconfig; `+
+					`if ! command -v helm >/dev/null 2>&1; then curl -fsSL https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash; fi; `+
+					`helm repo add cilium https://helm.cilium.io >/dev/null 2>&1 || true; `+
+					`helm repo update >/dev/null; `+
+					`helm upgrade --install cilium cilium/cilium --version %s --namespace %s --create-namespace --kubeconfig /tmp/admin-kubeconfig `+
+					`--set cni.uninstall=false `+
+					`--set cni.binPath=/var/lib/cni/bin `+
+					`--set cni.confPath=/var/run/multus/cni/net.d `+
+					`--set kubeProxyReplacement=true `+
+					`--set k8sServiceHost=172.20.0.1 `+
```
mbukatov: This should not be hardcoded, but passed as a parameter like the other values. I know that the other CNI test does this, but it's not a good idea, and here is ever worse, since we are combining multiple different features together and any major change in networking configuration would easily lead in a problem.

test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,401 @@
...
+// - Internal image registry disabled
+// - API IP address access control (authorized CIDRs)
+// - KeyVaultVisibility set to Private
+var _ = Describe("Customer", func() {
+	It("should be able to create a cluster and node pools with aggregated advanced features",
+		labels.RequireNothing,
+		labels.Critical,
+		labels.Positive,
+		labels.AroRpApiCompatible,
+		labels.CreateCluster,
```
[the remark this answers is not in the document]
mbukatov: This test case is not significantly slower compared to other test cases without such label.

test/e2e/cluster_create_feature_aggregation.go
mbukatov: This should be fixed:

- fetch the k8s api server IP address via a client query (cilium docs suggests `kubectl get ep kubernetes -o jsonpath='{$.subsets[0].addresses[0].ip}'` but I'm not sure if this is a correct approach here)
- pass it to the command in a similar way as it's done for `ipam.operator.clusterPoolIPv4PodCIDRList`

One more reason to fix this is the following discrepancy I noticed:

- we set `clusterParams.Network.ServiceCIDR = "172.30.0.0/16"`
- but the hard coded api server IP address value is 172.20.0.1
- moreover, value DefaultK8sServiceIP from test/util/framework/deployment_params.go is 172.30.0.1, matching the default ServiceCIDR value

That said, I'm not sure how to get the correct IP address value for Cilium in this case. The value from the other test doesn't seem to explain it, and we have no official documentation.

test/e2e/cluster_create_feature_aggregation.go
mbukatov: Could we at least split the commands into 2, so that:
- the first command will install helm
- the 2nd command will do the actual cilium installation

This way, we will be able to:
- retry helm installation (in case of github issues) few times before giving up
- quickly distinguish between helm and cilium installation failures

test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,423 @@
...
+			)
+			_, err = framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, waitNodesReadyCmd, 16*time.Minute)
+			Expect(err).NotTo(HaveOccurred(), "all nodes should become Ready from authorized VM checks")
+			waitCiliumReadyCmd := fmt.Sprintf(
+				`echo '%s' | base64 -d > /tmp/admin-kubeconfig && kubectl --kubeconfig=/tmp/admin-kubeconfig -n %s rollout status daemonset/cilium --timeout=10m`,
+				adminKubeconfigB64,
+				ciliumNamespace,
+			)
+			_, err = framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, waitCiliumReadyCmd, 11*time.Minute)
+			Expect(err).NotTo(HaveOccurred(), "cilium daemonset should become Ready from authorized VM checks")
```
mbukatov: We should extend this test case so that in the end we:

- disable authorized CIDRs filtering or add a IP address range used by openshift CI so that the test can reach the cluster directly again
- run all applicable verifiers at once (via `err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyNodesReady(), verifiers.VerifyCiliumOperational(ciliumNamespace, "k8s-app=cilium"), verifiers.VerifySimpleWebApp(), verifiers.VerifyCiliumConnectivityChecks("1.19.2"), ...)` so that we can have a comparable level of checking for the combined complex test case

Without doing this, we won't be actually sure that the combined cluster actually works in the same way as we verify in individual feature test cases. Since we don't run the existing verifiers from the VM, we don't have the comparable coverage.

mbukatov (commented):
There are multiple minor issues with this E2E test, most of it out of our control. That said I'm highlighting few which should be in my opinion fixed before merging, so that we:

- have better error reporting in case the test fails
- make the test coverage is closer to other single feature tests
- understand that we configure the features correctly

test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,447 @@
...
+			nodePoolErrB := tc.CreateNodePoolFromParam20251223(
+				ctx,
+				GinkgoLogr,
+				*resourceGroup.Name,
+				clusterParams.ManagedResourceGroupName,
+				customerClusterName,
+				nodePoolParamsB,
+				framework.NodePoolCreationTimeout,
+			)
+			Expect(errors.Join(nodePoolErrA, nodePoolErrB)).NotTo(HaveOccurred(), "failed to create aggregated feature node pools for cluster %q", customerClusterName)
```
mbukatov: I believe this Expect line should be dropped and is there by mistake, because the intention of this pattern is to delay the check and run expect together with a result of the verifier, like in the previous case.

test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,447 @@
...
+						"vmSize":       vmSize,
+					}),
+					framework.WithTimeout(30*time.Minute),
+				)
+				if deployErr == nil || !strings.Contains(deployErr.Error(), "SkuNotAvailable") {
+					break
+				}
+				time.Sleep(20 * time.Second)
+			}
+			Expect(deployErr).NotTo(HaveOccurred(), "failed to deploy authorized CIDR VM")
```
mbukatov, line 145: I see you added the test IP address to the list of allowed IP addresses, which makes sense and makes this combined test more robust and reusing the verifiers makes the scope comparable.

But in the current form, it looks like we don't need the vm at all? We could either:

- don't deploy the vm at all
- deploy the vm, but don't add it's IP address to the allow IP list so that then we can check that we can't access the cluster from the vm (via framework.RunVMCommand)

test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,490 @@
...
+// - Internal image registry disabled
+// - API IP address access control (authorized CIDRs)
+// - KeyVaultVisibility set to Private
+var _ = Describe("Customer", func() {
+	It("should be able to create a cluster and node pools with aggregated advanced features",
+		labels.RequireNothing,
+		labels.Critical,
+		labels.Positive,
+		labels.AroRpApiCompatible,
+		labels.CreateCluster,
```
mbukatov, line 63: With https://github.com/Azure/ARO-HCP/pull/6029 now merged, we now need to add `labels.MIContainers(1)`, don't we? I'm not sure why it's not raised via CI.

test/e2e/cluster_create_feature_aggregation.go
```diff
@@ -0,0 +1,490 @@
...
+			oidcRESTConfig.BearerToken = accessToken.Token
+			oidcRESTConfig.BearerTokenFile = ""
+			oidcClient, err := kubernetes.NewForConfig(oidcRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "failed to create kubernetes client with OIDC bearer token")
+			Eventually(func(g Gomega) {
+				nsList, listErr := oidcClient.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+				g.Expect(listErr).NotTo(HaveOccurred(), "external OIDC identity should be able to list namespaces")
+				g.Expect(nsList.Items).NotTo(BeEmpty(), "external OIDC identity should observe at least one namespace")
+			}, 5*time.Minute, 10*time.Second).Should(Succeed(), "external OIDC identity should be able to list namespaces through the Kubernetes API")
+		},
```
mbukatov, line 493: It would be good if we could run some final verifier in the very end utilizing most of the features.

Something like (assuming we can pass k8s client into VerifySimpleWebApp to make sure we can run a web app under different user):

``` .go
err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifySimpleWebApp(oidcClient), verifiers.VerifyCiliumConnectivityChecks(ciliumVersion))
```

But since this is not currently implemented in the original test case, I suggest to add it as a follow up, across all affected E2E test cases.

Sheer rerun of `verifiers.VerifyHCPCluster` would be enough.

mbukatov (commented):
the updates looks good

see my 2 minor suggestions below

test/e2e/cluster_create_feature_aggregation.go
mbukatov: Maybe it will be raised during a tide test rerun eventually.

mbukatov (commented):
/lgtm

thank you for adding the last changes

PR #5790, opened 2026-06-25
feat: introduce asynchronous nodepool creation in backend in a disabled state

miguelsorianod:
The backend logic to perform asynchronous nodepool creation is disabled for now: the added controller is only executed when ClusterServiceID is not set. For now in frontend we always store the ClusterServiceID so it is guaranteed to not be executed.

Additionally, the nodepool create operation has been updated to retrieve the ClusterServiceID from the NodePool cosmos resource instead of the operation. This allows us to stop relying on it for when we stop setting it in frontend.

### DIFF
--- a/backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
+++ b/backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
@@ -1,0 +1,186 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nodepoolcreationcontrollers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
+	ocmerrors "github.com/openshift-online/ocm-sdk-go/errors"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/backend/pkg/informers"
+	"github.com/Azure/ARO-HCP/backend/pkg/listers"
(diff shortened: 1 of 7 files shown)


backend/pkg/controllers/nodepoolcreationcontrollers/node_pool_cluster_service_create_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	// For the Cluster, we retrieve from the cache because we are not about to use its data to interact with cluster-service. At
+	// the moment we only use the ClusterServiceID to interact with cluster-service, which shouldn't change over time once set.
+	// If at some point this controller evolves to use other Cluster properties that will be sent to cluster-service and that
+	// can change over time, we will need to retrieve from the database instead.
+	cluster, err := c.clusterLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if err != nil {
+		return utils.TrackError(err)
+	}
```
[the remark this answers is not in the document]
miguelsorianod: This shouldn't be an issue because a nodepool existing requires a cluster to exist beforehand. At that point the cache should have the entry

backend/pkg/controllers/operationcontrollers/operation_state_v2.go
```diff
@@ -0,0 +1,271 @@
...
+	"slices"
+	"strings"
+
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+)
+
+// operationError is an internal error type decoupled from arm.CloudErrorBody.
+// It carries structured error details through the operation state pipeline
+// and converts to arm.CloudErrorBody at the boundary when persisting to the DB.
+type operationError struct {
```
deads2k: why not just use `arm.CloudErrorBody`?

backend/pkg/controllers/operationcontrollers/operation_state_v2.go
```diff
@@ -0,0 +1,271 @@
...
+			source = "<unset>"
+		}
+
+		for _, msg := range s.Messages {
+			allMessages = append(allMessages, fmt.Sprintf("[%s] %s", source, msg))
+		}
+
+		if s.ProvisioningState == worst {
+			worstErrors = append(worstErrors, s.Errors...)
+		}
```
deads2k: why not just all the errors?

backend/pkg/controllers/operationcontrollers/operation_state_v2.go
miguelsorianod: I wanted to keep the internal logic decoupled from details of arm.CloudErrorBody. Then, we would translate to an arm.CloudErrorBody only on the boundary where we are setting the actual ARM Operation cosmos resource

backend/pkg/controllers/operationcontrollers/operation_state_v2.go
miguelsorianod: We are picking the worst provisioning state, so we only return the errors associated to that state

PR #5794, opened 2026-06-25
ControlPlaneDesiredVersion: skip recompute while cluster create is inflight

deads2k:
Until the initial cluster create finishes, the desired-version syncer should not race the create flow by recomputing a z-stream. Gate the computation on:

  1. SPC.Spec.ControlPlaneVersion.DesiredVersion is empty — must run to seed it.
  2. cluster.SystemData.CreatedAt older than 2h — past that window the create is expected to be done.
  3. no active Create operation whose ExternalID is the cluster itself — without a create in flight there is nothing to race with.

If the active-operation lister errors we fail open (shouldRun=true) and surface the error, so a flaky lister doesn't pin the controller in skip-forever mode.

NewControlPlaneDesiredVersionController now takes a PassiveClock and the syncer stores the ActiveOperationLister directly (the cooldown checker already wrapped it).

### DIFF
--- a/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
+++ b/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/blang/semver/v4"
@@ -26,6 +27,7 @@ import (
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/operation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilsclock "k8s.io/utils/clock"
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 
@@ -47,6 +49,12 @@ import (
 	"github.com/Azure/ARO-HCP/internal/validation"
 )
 
+// clusterCreateGracePeriod is how long after a cluster's CreatedAt we
+// suppress automatic desired-version recomputation while an active Create
+// operation is still in flight. After this window the create is expected
+// to have finished, so resuming z-stream selection is safe.
+const clusterCreateGracePeriod = 2 * time.Hour
+
 // controlPlaneDesiredVersionControllerName is the Cosmos controller document ID for this syncer.
 const controlPlaneDesiredVersionControllerName = "ControlPlaneDesiredVersion"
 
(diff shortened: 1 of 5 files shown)


backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -557,3 +581,66 @@ func selectBestVersionFromCandidates(
...
+// flight) we skip so a freshly created cluster doesn't have its initial
+// desired version overwritten while creation is still in progress.
+func (c *controlPlaneDesiredVersionSyncer) shouldDetermineDesiredVersion(ctx context.Context, key controllerutils.HCPClusterKey, cluster *api.HCPOpenShiftCluster, spc *api.ServiceProviderCluster) (bool, error) {
+	if spc.Spec.ControlPlaneVersion.DesiredVersion == nil {
+		return true, nil
+	}
+	if c.clusterOlderThanGracePeriod(cluster) {
+		return true, nil
+	}
+	hasCreate, err := c.hasActiveClusterCreateOperation(ctx, key)
```
machi1990, line 606: Can we also have this check here https://github.com/deads2k/ARO-HCP/blob/709c453a7feeabf71a3285cae1fdd519d3140657/backend/pkg/controllers/upgradecontrollers/trigger_control_plane_upgrade_controller.go ? 

i.e not trigger the upgrade is the cluster is being created?

deads2k: > Can we also have this check here https://github.com/deads2k/ARO-HCP/blob/709c453a7feeabf71a3285cae1fdd519d3140657/backend/pkg/controllers/upgradecontrollers/trigger_control_plane_upgrade_controller.go ?
> 
> i.e not trigger the upgrade is the cluster is being created?

I think we should leave the check out of this controller.  If for some reason the desiredVersion is specified directly and differs, we should honor it. For instance, what if an adminAPI changed it?

machi1990: If we don't have that check there; we find ourselves in a situation where we trigger the upgrade based off the .desiredVersion computed from the initial initialisation.
i.e 
> . ServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion is unset
//     — we have nothing seeded yet, so we must run to fill it in.

This is the install case only; and once computed it should only be handled by the install controller. 

> or instance, what if an adminAPI changed it?

Is there a use case in SRE enforcing a version while the cluster is installing? I recall the use case of rollbacking to a previous installed version (where the cluster came from) but I've hard times imaging the SRE enforced version for install case;

Additionally, if we've an `SREEnforcedDesiredInstallVersion` we can add it as part of the `NeedsWork` check for the triggercontroller.
```
 If !hasActiveClusterCreateOperation() || sreEnforcedInstallVersion() {...do something}
```

deads2k: >Is there a use case in SRE enforcing a version while the cluster is installing?

clusters might end up perma-installing, so we should handle that.

deads2k: > If we don't have that check there; we find ourselves in a situation where we trigger the upgrade based off the .desiredVersion computed from the initial initialisation. i.e
> 
> > . ServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion is unset
> > //     — we have nothing seeded yet, so we must run to fill it in.
> 
> This is the install case only; and once computed it should only be handled by the install controller.

Urg.  I forgot that detail about install via cluster-service.  And it won't be fixed until the install call moves.  ok.

EDIT: once we move install to determine a version first, then we'll remove this extra bit.

PR #5795, opened 2026-06-25
test: make firing alerts block e2e

stevekuznetsov:
We would like to be able to catch PRs which add alerts that would be firing spuriously, before they merge. The mechanisms for this already exist, we just need to make them blocking.


deads2k:
Steve promised to guide us through this transition.

stevekuznetsov:

Allowing alerts to merge which fire in PR testing is a risky place to be, and we benefit greatly from plugging this gap.

stevekuznetsov:
/label lgtm

Removed a weird remnant alert that was firing and failing e2e periodically.

stevekuznetsov:
/label lgtm

for rebase

PR #5804, opened 2026-06-26
PoC cluster updates based on comparing CS state and cosmos desired stat with sha persistence

miguelsorianod:
Same as https://github.com/Azure/ARO-HCP/pull/5593 idea but with sha persistence to avoid hotlooping


miguelsorianod:
For now we are not going to introduce the sha persistence. Replaced by https://github.com/Azure/ARO-HCP/pull/6047 where we dynamically calculate it, compare cosmos desired state vs cs reported state, as well as add a minimum cooldown on reconciles to try to lower the number of CS get interactions.

We don't discard adding sha persistence at some point but for now we will keep it the described way.

PR #5806, opened 2026-06-26
fix: skip nodepool upgrade trigger if no active versions exist yet

JameelB:
### What

Skip creating an upgrade policy for a node pool if there are no active versions present. 

JIRA: [ARO-28029](https://redhat.atlassian.net/browse/ARO-28029)

### Why
With the move to reading the active versions from maestor/management cluster, there may be a chance that the active versions may not yet be present when a node pool is still being installed.

Previously, as soon as the desired version exist, the trigger would immediately attempt to create an upgrade policy since the actual version would be empty. This was causing the following errors and triggering unwanted alerts: 
```
(wrapped at trigger_node_pool_upgrade_controller.go:155) failed to create node pool upgrade policy: (wrapped at client.go:764) status is 400, identifier is '400', code is 'CLUSTERS-MGMT-400', at '2026-06-22T04:00:37Z' and operation identifier is 'xxx': Node pool new version '4.21.0' must be different from the old version
```

This change ensures that this controller doesn't trigger the upgrade in this case. 

### Testing
Unit tests added for the SyncOnce() to ensure the guards before creating the upgrade policy behave as expected. The full SyncOnce() path, including the creation of the upgrade policy, wasn't tested as there were already a separate test for this in the same file.

Required for https://github.com/Azure/ARO-HCP/pull/5805

### PR Checklist

- [ ] ~~Screenshots included (if graph/UI/metrics changes)~~
- [x] Self-reviewed the diff
- [ ] ~~Draft PR used for WIP (if applicable)~~

If E2E tests are included:

- [ ] ~~E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md)~~
- [ ] ~~If new E2E use case is covered (via a new test or new check/verifier),
  proper error message and logs which communicates nature of the problem.~~

### DIFF
--- a/backend/pkg/controllers/upgradecontrollers/trigger_node_pool_upgrade_controller.go
+++ b/backend/pkg/controllers/upgradecontrollers/trigger_node_pool_upgrade_controller.go
@@ -79,8 +79,9 @@ func (c *triggerNodePoolUpgradeSyncer) CooldownChecker() controllerutil.Cooldown
 //
 // High-level flow:
 //  1. Fetch the node pool and service provider node pool state
-//  2. Check if desiredVersion differs from latest actual version
-//  3. If different, create a NodePoolUpgradePolicy to trigger upgrade
+//  2. Skips upgrade trigger if no active versions exist yet (during installation)
+//  3. Check if desiredVersion differs from latest actual version
+//  4. If different, create a NodePoolUpgradePolicy to trigger upgrade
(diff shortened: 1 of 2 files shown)


machi1990 (approved):
/lgtm

machi1990 (approved):
/lgtm

PR #5807, opened 2026-06-26
feat: add peak threshold slider to metrics panel for series filtering

geoberle:
### What

Adds a vertical threshold slider to metrics panel charts. The slider sits alongside the Y axis and filters both visible series and legend entries to only those whose peak value exceeds the threshold. A dashed red line marks the threshold on the chart. The initial threshold can be configured per-query via `minPeakThreshold` in `queries.yaml`.

### Why

Metrics panels can have dozens of timeseries where only a few are interesting. The slider lets reviewers quickly isolate the series that actually spiked without manually toggling legend entries.

### Screenshot

Before

image

After

slider-demo


stevekuznetsov:
/lgtm
/approve

Amazing <3

PR #5814, opened 2026-06-26
feat: enable mise audit loging in msft environments

tony-schndr:
fixes https://redhat.atlassian.net/browse/ARO-26485

### What

Configures mise v2 audit logging following the established frontend / admin api pattern.

### Why

For comprehensive audit logging of all requests through MISE.

### Testing
Testing mise audit logging is not possible in non-Microsoft environment.  Deployment / configuration is naturally validated through e2e environment provisioning.

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/istio/deploy/charts/mise/templates/deployment.yaml
+++ b/istio/deploy/charts/mise/templates/deployment.yaml
@@ -46,6 +46,10 @@ spec:
           mountPath: /app/appsettings.json
           subPath: appsettings.json
           readOnly: true
+        {{- if .Values.audit.connectSocket }}
+        - name: mdsd-asa-run-vol
+          mountPath: /var/run/mdsd
+        {{- end }}
         env:
         - name: OTEL_EXPORTER_OTLP_ENDPOINT
           value: "{{ .Values.tracing.address }}"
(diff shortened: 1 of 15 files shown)


istio/deploy/charts/mise/templates/deployment.yaml
```diff
@@ -46,6 +46,10 @@ spec:
           mountPath: /app/appsettings.json
           subPath: appsettings.json
           readOnly: true
+        {{- if .Values.audit.connectSocket }}
+        - name: mdsd-asa-run-vol
```
tmstff, line 50: Comment by claude: "Volume name is slightly misleading. mdsd-asa-run-vol — "asa" suggests AzSecPack's specific socket
  (/var/run/mdsd/asa/default_fluent.socket), but the PR actually uses the plain fallback path
  Endpoint=unix:/var/run/mdsd/default_fluent.socket via CustomConnectionString. Frontend uses the same name mdsd-asa-run-vol, so it's
  at least consistent. Not a blocker, but naming could confuse future readers."
-> I'll leave it up to you :-)

mbarnes (commented):
/lgtm

This all looks reasonable to me but it lies entirely outside the scope of my code ownership, so someone else will have to approve.

bennerv:
/approve
/lgtm

There's not really a way to test this in dev since it requires mdsd integration which requires a onecert signed certificate.  This looks good, but true test will be int.  if we know of another way in the future to test in dev, we should try to.

PR #5816, opened 2026-06-26
make leader lease for kube-applier low pressure, but still reasonably…

deads2k:
… responsive based on OCP defaults


stevekuznetsov (approved):
/lgtm
/approve

PR #5845, opened 2026-06-29
fix: tolerating tenths place in credential refresher alerts and fixing dashboard

rachelvweber:
[ARO-28061](https://redhat.atlassian.net/browse/ARO-28061)
### What

Tolerates tenths place in credential refresher alerts. Also, explicitly calculating the diff between different buckets in the dashboard.

### Why

The buckets used to be whole numbers (-90, 0, 3, etc...), but now they have a tenths place (-90.0, 0.0, 3.0, etc...). This broke the alerts. For the dashboard, I thought the delta calculation was working, but lower buckets were still being included in the columns of higher buckets. I tried for a while to fix this in a way that's better to read and more flexible, but ultimately settled on this. Open to suggestions here.

### Testing

Updated the alerts tests to have both formats and tested the queries/dashboard update manually in the int grafana.

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:


stevekuznetsov (approved):
/lgtm
/approve

PR #5849, opened 2026-06-29
feat(monitoring): add per-lane ICM enabled config knob

SudoBrendan:
https://redhat.atlassian.net/browse/ARO-28074 (partial; also requires sdp-pipelines change)

### What

Add an enabled boolean to each IcM lane (SRE, SL, RP, MSFT) under monitoring.icm.<lane>. When false, that lane's action group is dropped from its alert rule wiring. Prometheus rules still evaluate (preserving the e2e alert gate) but do not generate IcM tickets.

Defaults to true for every lane; no behavior change until an environment overrides it.

Incorporates @mmazur's generalized approach from #5875, which extended the original RP-only knob to all four lanes using the shared icmConnection schema definition.

### Why

Non-production IcM tickets in the production queue create noise and misallocate on-call attention. The e2e alert gate catches alerting regressions and notifies engineering through failure messages, not IcM. Gating each lane independently lets any team suppress non-prod IcM for their lane without additional config work.

A follow-up in sdp-pipelines will set monitoring.icm.rp.enabled: false for INT and STG.

### Testing

**If you did not add tests, provide a clear justification.**

As a production-only infrastructural change, it is unfortunately not possible to test this feature. We are updating the bicep template - deployment (after updates to sdp-pipelines) is ultimately what determines if our production queue no longer sees ICMs for INT/STG environments. Public documentation cites `[]` as a valid action group. **Defaults in this repo preserve existing state, and effectively render this PR standalone as a no-op.** SDP-Pipelines is what may cause differences in how we deploy, and this PR enables that future change.

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:


komidore64 (approved):

SudoBrendan:
@mmazur I'm not quite sure how to proceed on this between our two PRs and their different acceptance criteria :sweat_smile: but I saw you closed yours so I went ahead and incorporated your changes here and kept your credit as an author. Your solution is the correct one given the expanded acceptance criteria you proposed - all of which I agree with.

PR #5859, opened 2026-06-30
Fix: Remove unneeded error message check

mgahagan73:
https://redhat.atlassian.net/browse/ARO-27850

### What

Fix test failure due to error message string mismatch by removing the error message content check.

### Why

The exact string check on the error message was causing false test failures and there isn't much value in knowing the exact error message, the test only needs to know that the connection failed.

### Testing

Pre-merge e2e is sufficient here.

- Unit tests

### PR Checklist

- [ ] Screenshots included (if graph/UI/metrics changes) (N/A)
- [x] Self-reviewed the diff

If E2E tests are included:

- [ ] E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md) N/A
  proper error message and logs which communicates nature of the problem. N/A

### DIFF
--- a/test/e2e/cluster_authorized_cidrs_connectivity.go
+++ b/test/e2e/cluster_authorized_cidrs_connectivity.go
@@ -187,7 +187,7 @@ var _ = Describe("Authorized CIDRs", func() {
 				// Try to connect from the test runner (which is not in authorized CIDRs)
 				err = testAPIConnectivity(apiURL, 5*time.Second)
 				Expect(err).To(HaveOccurred(), "Connection from unauthorized IP should be blocked")
-				GinkgoWriter.Printf("Unauthorized connection error: %v\n", err)
+				GinkgoWriter.Printf("Connection from unauthorized IP address failed as expected on error: %v\n", err)
 
 				By("verifying VM can access cluster API with credentials")
 				adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
--- a/test/e2e/cluster_authorized_cidrs_connectivity.go
+++ b/test/e2e/cluster_authorized_cidrs_connectivity.go
@@ -187,7 +187,7 @@ var _ = Describe("Authorized CIDRs", func() {
 				// Try to connect from the test runner (which is not in authorized CIDRs)
 				err = testAPIConnectivity(apiURL, 5*time.Second)
 				Expect(err).To(HaveOccurred(), "Connection from unauthorized IP should be blocked")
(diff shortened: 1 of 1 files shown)


test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -187,8 +187,7 @@ var _ = Describe("Authorized CIDRs", func() {
 				// Try to connect from the test runner (which is not in authorized CIDRs)
 				err = testAPIConnectivity(apiURL, 5*time.Second)
 				Expect(err).To(HaveOccurred(), "Connection from unauthorized IP should be blocked")
-				// Verify it's a connection error (EOF indicates connection was closed by server/network)
-				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("Get \"%s/healthz\": EOF", apiURL)), "Should fail with EOF error indicating blocked connection")
+				// We only need to verify that the connection is blocked, not the specific error message
```
mbukatov: Dropping this check makes sense.

That said, there is still a small possibility that the connection will fail for unrelated reason while we introduce a regression, and this check won't be able to distinguish between these two states.

Maybe we can check that the error is of expected class of an error via https://pkg.go.dev/errors#Is instead.

mbukatov: But let's not complicate this too much. For now, let's just log the error (via info or debug level) so that we can audit this if needed.

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -187,8 +187,7 @@ var _ = Describe("Authorized CIDRs", func() {
 				// Try to connect from the test runner (which is not in authorized CIDRs)
 				err = testAPIConnectivity(apiURL, 5*time.Second)
 				Expect(err).To(HaveOccurred(), "Connection from unauthorized IP should be blocked")
-				// Verify it's a connection error (EOF indicates connection was closed by server/network)
-				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("Get \"%s/healthz\": EOF", apiURL)), "Should fail with EOF error indicating blocked connection")
+				GinkgoWriter.Printf("Unauthorized connection error: %v\n", err)
```
mbukatov:
```suggestion
				GinkgoWriter.Printf("Connection from unauthorized IP address failed as expected on error: %v\n", err)
```

mbukatov (commented):
/lgtm

PR #5862, opened 2026-06-30
tooling/prometheus-rules: add context back to correlation IDs

stevekuznetsov:
The previous change to remove all the specificity from correlation IDs was done in error. We *need* the ability to differentiate between a specific controller having high error retry volume, to be able to know if one node went down or 30 in a cluster, etc. The specificity is a hard prerequisite to having useful alerting.

It is clear and well-understood that the IcM user interface makes it difficult to triage lots of alerts, but we can dump the raw data from the IcM database and trivially build our own UIs on top of it, so no line of reasoning about the DX is persuasive to lose the ability to distinguish our failures.

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/generatedHCPPrometheusAlertingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedHCPPrometheusAlertingRules.bicep
@@ -32,12 +32,12 @@ resource kubernetesApps 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-03
           severity: 'warning'
         }
         annotations: {
-          correlationId: 'KubePodCrashLooping/{{ $labels.cluster }}'
+          correlationId: 'KubePodCrashLooping/{{ $labels.cluster }}/{{ $labels.namespace }}/{{ $labels.pod }}/{{ $labels.container }}'
           description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is in waiting state (reason: "CrashLoopBackOff").'
           info: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is in waiting state (reason: "CrashLoopBackOff").'
           runbook_url: 'https://runbooks.prometheus-operator.dev/runbooks/kubernetes/kubepodcrashlooping'
           summary: 'Pod is crash looping.'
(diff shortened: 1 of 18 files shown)


dev-infrastructure/modules/metrics/rules/generatedHCPPrometheusAlertingRules.bicep
```diff
@@ -32,12 +32,12 @@ resource kubernetesApps 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-03
...
         }
         annotations: {
-          correlationId: 'KubePodCrashLooping/{{ $labels.cluster }}'
+          correlationId: 'KubePodCrashLooping/{{ $labels.cluster }}/{{ $labels.namespace }}/{{ $labels.pod }}/{{ $labels.container }}'
           description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is in waiting state (reason: "CrashLoopBackOff").'
           info: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is in waiting state (reason: "CrashLoopBackOff").'
           runbook_url: 'https://runbooks.prometheus-operator.dev/runbooks/kubernetes/kubepodcrashlooping'
           summary: 'Pod is crash looping.'
-          title: 'Pod is crash looping.'
+          title: 'Pod is crash looping. namespace:{{ $labels.namespace }} pod:{{ $labels.pod }} container:{{ $labels.container }}'
```
stevekuznetsov, line 40: It strikes me that the `description` is often what I would want to see from these IcMs, not the title. A generic title is not very useful to me. These edits seem to come from newer code that had a similar insight, but I don't know if we should prefer to edit the title like this, or just use the description. For what it's worth, the `ns/podname (container)` formatting is IMO more k8s-native and easier to understand than this key-value pairing.

rachelvweber: I sent this in a message, but will put this here for posterity. I think some of the descriptions are too long to function as a title, which is why I originally pulled these out of the description and put them on the title, but I agree that namespace/name is more intuitive.

stevekuznetsov:
/label lgtm

had to rebase

stevekuznetsov:
/label lgtm

had to rebase

stevekuznetsov:
/label lgtm

after rebase

PR #5872, opened 2026-07-01
chore: aling ci mgmt cluster sizing with provision script

roivaz:
[This](https://github.com/openshift/release/blob/30ca9730b6f4e3b5f8b00ef16e31aa2772863f44/ci-operator/step-registry/aro-hcp/provision/environment/aro-hcp-provision-environment-commands.sh#L135-L137) override was set to reollout single-wave e2e. This PR sets those values in the config so the override can be removed.

https://redhat.atlassian.net/browse/ARO-27792

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1662,12 +1662,12 @@ clouds:
                 vmSize: Standard_D4ds_v6
               userAgentPool:
                 maxCount: 14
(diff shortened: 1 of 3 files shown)


raelga (approved):
/lgtm
/approve

PR #5874, opened 2026-07-01
Revert "Merge pull request #5858 from miguelsorianod/reintroduce-backendnodepoolcreate-async-disabled

miguelsorianod:
This reverts commit 9daef27c91b88cc90fb6e65899ab640b25f81054, reversing changes made to 92e7c1177062b0c89bbd032da7ee4d20d2d20d85.

We found out that when running Cosmos transactions our transaction layer logic doesn't propagate precondition failed errors.

### Testing

Testing is required for feature completion and tests should be part of the pull
request along with the feature changes.

Describe the testing provided. If you did not add tests, provide a clear
justification.

- Unit tests
- [Integration tests](https://github.com/Azure/ARO-HCP/tree/main/test-integration)
- [E2E tests](https://github.com/Azure/ARO-HCP/blob/main/test/e2e/README.md)

### PR Checklist

- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
- [ ] Screenshots included (if graph/UI/metrics changes)
- [ ] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [ ] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

If E2E tests are included:

- [ ] E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md)
- [ ] If new E2E use case is covered (via a new test or new check/verifier),
  demonstrate that the test is able to detect a defect/error and fail with
  proper error message and logs which communicates nature of the problem.

PR #5876, opened 2026-07-01
feat: reintroduce MR 5858 and propagate db transaction step errors with http status code information

miguelsorianod:
Reintroduces https://github.com/Azure/ARO-HCP/pull/5858 by reverting the revert MR https://github.com/Azure/ARO-HCP/pull/5874.

Additionally, it also introduces the transactionStepError Go error type and returns
it from transaction.Execute when a db transaction step fails. It also
extends the general isResponseError function to match both
azcore.ResponseError and transactionStepError. This allows the usage
of the database.IsPreconditionFailedError to recognize whether an error
returned because of a DB transaction step failure failed because of a
HTTP 412 status code.


geoberle:
/hold
until we inspected the resulting backend metrics

raelga (approved):
/lgtm
/approve

PR #5894, opened 2026-07-02
test: add e2e test for private ingress cluster creation

zgalor:
[ARO-27802](https://redhat.atlassian.net/browse/ARO-27802)

## Summary
- Adds an e2e test that creates a cluster with `ingress.type: Private` via the `v2026-06-30-preview` API
- Verifies private ingress at three levels:
  1. **Control plane**: ARM GET returns `ingress.type == Private`
  2. **Data plane**: `IngressController` CR in `openshift-ingress-operator` has `loadBalancer.scope == Internal`
  3. **Network**: Ingress endpoint is reachable from a VM inside the VNet but not from outside
- Regenerates v2026 test SDK to include `IngressProfile`/`IngressType` types
- Adds `IngressType` field to `ClusterParams20260630` framework params (defaults to `Public`)

## Changes
- **`test/e2e/cluster_create_private_ingress.go`** (new) — The e2e test, following `cluster_create_complex_cilium_kv.go` and `cluster_authorized_cidrs_connectivity.go` patterns
- **`test/util/framework/deployment_params.go`** — Added `IngressType` to `ClusterParams20260630`
- **`test/util/framework/hcp_helper.go`** — Wire `IngressType` into `BuildHCPClusterFromParams20260630`
- **`test/sdk/v20260630preview/...`** — Regenerated test SDK

## Depends on
- PR #5643 (merged) — adds `ingress.type` field to ARM API

## Test plan
- [x] `go vet --tags E2Etests ./test/e2e/...` — compilation passes
- [x] `make -C test` — test binary builds
- [x] `./test/aro-hcp-tests list` — test appears with correct labels
- [x] `./test/aro-hcp-tests list --suite "integration/parallel"` — appears in correct suite
- [x] `go test ./test/util/framework/...` — framework unit tests pass
- [x] `make record-nonlocal-e2e` — test registration passes
- [ ] CI e2e-parallel — will validate against live environment


zgalor:
@mbukatov PTAL

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,276 @@
...
+					if rt.ResourceType == nil || !strings.EqualFold(*rt.ResourceType, "hcpOpenShiftClusters") {
+						continue
+					}
+					for _, v := range rt.APIVersions {
+						if v != nil && strings.EqualFold(*v, apiVersion) {
+							available = true
+							break
+						}
+					}
+				}
```
ziccardi:
```suggestion
				available := slices.ContainsFunc(provider.ResourceTypes, func(rt *armresources.ProviderResourceType) bool {
					if rt.ResourceType == nil || !strings.EqualFold(*rt.ResourceType, "hcpOpenShiftClusters") {
						return false
					}

					return slices.ContainsFunc(rt.APIVersions, func(v *string) bool {
						return v != nil && strings.EqualFold(*v, apiVersion)
					})
				})	
```

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,276 @@
...
+			Eventually(func(g Gomega) {
+				route, getErr := rtClient.RouteV1().Routes("openshift-console").Get(ctx, "console", metav1.GetOptions{})
+				g.Expect(getErr).NotTo(HaveOccurred(), "failed to get console Route")
+				g.Expect(route.Spec.Host).NotTo(BeEmpty(), "console Route host should not be empty")
+				consoleURL = "https://" + route.Spec.Host
+			}, 5*time.Minute, 15*time.Second).Should(Succeed())
+
+			GinkgoLogr.Info("Console URL resolved from Route", "url", consoleURL)
+
+			By("verifying ingress is reachable from VM inside the VNet")
```
machi1990, line 189: Add a negative case; ingress is not reachable outside the VM

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,276 @@
...
+	operatorclient "github.com/openshift/client-go/operator/clientset/versioned"
+	routeclient "github.com/openshift/client-go/route/clientset/versioned"
+
+	hcpsdk20260630preview "github.com/Azure/ARO-HCP/test/sdk/v20260630preview/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+)
+
+var _ = Describe("Customer", func() {
+	It("should create a cluster with private ingress and verify the ingress is internal",
```
machi1990: I wonder if we can avoid creating an additional cluster and re-use https://github.com/zgalor/ARO-HCP/blob/eee3bf7dc2399802d89994c25625b2689aed0809/test/e2e/cluster_create_v20260630preview.go ?

machi1990: Or drop that test case; which ever you prefer

machi1990: Doesn't this e2e depend on CS https://github.com/openshift-online/aro-hcp-clusters-service/commit/ee741db6b5592f0cab9dd4da4b23f7b7180136a6 being deployed to ARO-HCP ? I think so, in that case, let's /hold the PR until that happens; well I don't expect the e2e to pass either until then

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,276 @@
...
+				framework.GetAdminRESTConfigTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to get admin REST config for private ingress cluster %q", customerClusterName)
+
+			By("verifying IngressController has internal load balancer scope")
+			opClient, err := operatorclient.NewForConfig(adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "failed to create operator client from admin REST config")
+
+			var ic *operatorv1.IngressController
+			Eventually(func(g Gomega) {
```
machi1990: Can you create a verifier, similar to the ones in https://github.com/zgalor/ARO-HCP/tree/eee3bf7dc2399802d89994c25625b2689aed0809/test/util/verifiers

test/e2e/cluster_create_private_ingress.go
machi1990: Ah it's in By("verifying ingress is NOT reachable from outside the VNet") and I missed it

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,276 @@
...
+				output, runErr := framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, curlCmd, 2*time.Minute)
+				g.Expect(runErr).NotTo(HaveOccurred(), "RunVMCommand failed for ingress connectivity test")
+
+				httpCode := strings.TrimSpace(output)
+				if httpCode != previousOutput {
+					GinkgoLogr.Info("VM ingress connectivity check", "httpCode", httpCode)
+					previousOutput = httpCode
+				}
+				g.Expect(httpCode).NotTo(Equal("000"),
```
machi1990: Doesn't the console availability depends on the external auth client configuration?

Are we better off deploying a sample app and checking that's available from within the VM's Vnet and not available from outside the Vnet?

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,276 @@
...
+			rtClient, err := routeclient.NewForConfig(adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "failed to create route client from admin REST config")
+
+			var consoleURL string
+			Eventually(func(g Gomega) {
+				route, getErr := rtClient.RouteV1().Routes("openshift-console").Get(ctx, "console", metav1.GetOptions{})
+				g.Expect(getErr).NotTo(HaveOccurred(), "failed to get console Route")
+				g.Expect(route.Spec.Host).NotTo(BeEmpty(), "console Route host should not be empty")
+				consoleURL = "https://" + route.Spec.Host
+			}, 5*time.Minute, 15*time.Second).Should(Succeed())
```
machi1990: See https://github.com/Azure/ARO-HCP/pull/5894/changes#r3514165965 about a possible suggestion but if you want the test to use the console then the value of it can be obtained from arm directly: See https://github.com/zgalor/ARO-HCP/blob/eee3bf7dc2399802d89994c25625b2689aed0809/test/e2e/cluster_tls_endpoints.go#L142

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,276 @@
...
+			Expect(err).To(HaveOccurred(),
+				"private ingress should not be reachable from outside the VNet, but connection succeeded")
+			GinkgoLogr.Info("Confirmed ingress is not reachable from outside the VNet")
+		},
+	)
+})
+
+// testIngressConnectivity attempts an HTTPS connection to the given URL.
+// Returns nil if the connection succeeds, or an error if it fails.
+func testIngressConnectivity(url string, timeout time.Duration) error {
```
machi1990: See the sample app verifier `verifySimpleWebApp` https://github.com/zgalor/ARO-HCP/blob/eee3bf7dc2399802d89994c25625b2689aed0809/test/util/verifiers/serving_app.go which can do the job here

test/e2e/cluster_create_private_ingress.go
machi1990: This is the verifier https://github.com/zgalor/ARO-HCP/blob/eee3bf7dc2399802d89994c25625b2689aed0809/test/util/verifiers/serving_app.go I am talking about (we need a fravor of it that's run from within a VM)

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,276 @@
...
+			const (
+				customerClusterName  = "private-ingress"
+				customerNodePoolName = "np-1"
+				apiVersion           = "2026-06-30-preview"
+			)
+
+			tc := framework.NewTestContext()
+
+			By("checking API version availability")
+			if !framework.IsDevelopmentEnvironment() {
```
machi1990: I am wondering if we can follow this approach here https://github.com/ziccardi/ARO-HCP/blob/2be789e8eff4ea073c7374f3fd7b04e4950afd37/test/e2e/kms_key_rotation.go#L89

i.e we always create the cluster, if creation fail we inspect it as such https://github.com/ziccardi/ARO-HCP/blob/2be789e8eff4ea073c7374f3fd7b04e4950afd37/test/e2e/kms_key_rotation.go#L89 and if the API is not yet deployed we fail with the timebomb.
It's straightforward that way

test/e2e/cluster_create_private_ingress.go
zgalor: CS support for private ingress ([MR !353](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/353)) was merged on June 16 and is included in the current prod CS version. Specifically, the prod version (`ea24fe5ee4`, June 17) is 2 commits ahead of !353's merge commit (`07dc9dacb5c3`), so the private ingress code is deployed to all environments.

Additionally, the e2e-parallel CI on the merged frontend PR (#5643) already creates clusters with the `ingresses` field set and passes successfully — if CS didn't support it, cluster creation would fail with a rejection error.

test/util/verifiers/ingress.go
```diff
@@ -0,0 +1,67 @@
...
+	return nil
+}
+
+// VerifyIngressControllerScope returns a verifier that checks the default
+// IngressController's load balancer scope matches the expected value.
+// Use operatorv1.InternalLoadBalancer for private ingress clusters
+// or operatorv1.ExternalLoadBalancer for public ingress clusters.
+func VerifyIngressControllerScope(expectedScope operatorv1.LoadBalancerScope) HostedClusterVerifier {
+	return verifyIngressControllerScope{expectedScope: expectedScope}
+}
```
machi1990: nitpick: missing newline at EOF

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	return nil
+}
```
machi1990: np: missing newline

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,426 @@
...
+			return nsName, "", fmt.Errorf("timed out waiting for route host to be assigned")
+		case <-time.After(5 * time.Second):
+		}
+	}
+
+	return nsName, host, nil
+}
+
+// cleanupSampleApp removes the namespace created by deploySampleApp.
+func cleanupSampleApp(ctx context.Context, dynamicClient dynamic.Interface, namespace string) {
```
machi1990: is this needed if the test cleans up everything?

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,426 @@
...
+				"private ingress should not be reachable from outside the VNet, but connection succeeded")
+			GinkgoLogr.Info("Confirmed ingress is not reachable from outside the VNet")
+		},
+	)
+})
+
+// deploySampleApp creates a simple web app (agnhost serve-hostname) with a
+// Service and Route in a new namespace. Returns the namespace name and the
+// route host, or an error.
+func deploySampleApp(ctx context.Context, dynamicClient dynamic.Interface) (string, string, error) {
```
machi1990: Maybe this can go in the helper pkg?

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,426 @@
...
+			Expect(err).NotTo(HaveOccurred(), "failed to verify cluster health or IngressController scope for cluster %q", customerClusterName)
+			GinkgoLogr.Info("Cluster health and IngressController scope verified")
+
+			By("deploying a sample web app to verify ingress connectivity")
+			dynamicClient, err := dynamic.NewForConfig(adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "failed to create dynamic client from admin REST config")
+
+			appNamespace, appRouteHost, err := deploySampleApp(ctx, dynamicClient)
+			Expect(err).NotTo(HaveOccurred(), "failed to deploy sample web app for ingress connectivity test")
+			defer cleanupSampleApp(ctx, dynamicClient, appNamespace)
```
machi1990: Do we need to cleanup if the test cleansup everything at the end of the run?

test/e2e/cluster_create_private_ingress.go
zgalor: We can, but shouldn't we keep the fail-upfront approach?

It fails fast and avoids wasting CI time and Azure resources on a cluster that can't be created(resource group, managed identities, infra)

And it also separates "is the API available?" from "does cluster creation work?" which provides cleaner test semantics

machi1990: I prefer [that approach](https://github.com/Azure/ARO-HCP/pull/5894#discussion_r3514235595) because it keeps the test focused on what's supposed to test; cluster creation of a private cluster works making it easier for the reader to follow what's being tested.

The resources are released as soon as the test ends that's why it wasn't the biggest concern for me.

test/util/verifiers/ingress.go
zgalor: Fixed — added trailing newline.

test/e2e/cluster_create_private_ingress.go
zgalor: Fixed — added trailing newline.

test/e2e/cluster_create_private_ingress.go
zgalor: Removed — the cluster gets deleted at test end, so the namespace goes with it.

test/e2e/cluster_create_private_ingress.go
zgalor: Moved to `test/util/framework/sample_app_helper.go` as `framework.DeploySampleApp()`.

test/e2e/cluster_create_private_ingress.go
zgalor: Removed — same as above, cluster deletion cleans up the namespace.

test/e2e/cluster_create_private_ingress.go
zgalor: Good point — switched to the try-and-inspect pattern using `isAPINotDeployedError()` (same as `kms_key_rotation.go` and `nodepool_ephemeral_osdisk.go`). The upfront API version check is removed.

test/util/framework/sample_app_helper.go
machi1990, line 1: What do you think of re-using the sample app logic in https://github.com/zgalor/ARO-HCP/blob/79f05b13556036e11be3734062d102049e26432a/test/util/verifiers/serving_app.go#L69 ? to consolidate one way of creating a sample app; that's called from that verifier and here?

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,246 @@
...
+			Expect(err).NotTo(HaveOccurred(), "failed to create dynamic client from admin REST config")
+
+			_, appRouteHost, err := framework.DeploySampleApp(ctx, dynamicClient)
+			Expect(err).NotTo(HaveOccurred(), "failed to deploy sample web app for ingress connectivity test")
+
+			appURL := "https://" + appRouteHost
+			GinkgoLogr.Info("Sample app deployed", "url", appURL)
+
+			By("verifying ingress is reachable from VM inside the VNet")
+			curlCmd := fmt.Sprintf("curl -k -s -o /dev/null -w '%%{http_code}' --connect-timeout 10 %s", appURL)
```
machi1990, line 192: nitpick: this can follow[ the verifiers pattern](https://github.com/zgalor/ARO-HCP/blob/79f05b13556036e11be3734062d102049e26432a/test/util/verifiers/)

test/util/framework/sample_app_helper.go
zgalor: I looked into reusing `verifySimpleWebApp` but the use cases are quite different:

- `verifySimpleWebApp` deploys, verifies TLS reachability (both insecure and strict), and cleans up atomically in a single `Verify()` call. We need the app to stay deployed for the VM connectivity test and the negative external connectivity check.
- It uses embedded YAML artifacts with `nodeSelector` support, while we just need a minimal probe target.
- Refactoring `serving_app.go` to separate deployment from verification would broaden this PR's scope and risk breaking the TLS endpoint tests.

Happy to consolidate in a follow-up if a third use case emerges and the shared pattern becomes clearer. For now keeping them separate avoids coupling two different concerns (TLS cert validation vs connectivity probing).

test/e2e/cluster_create_private_ingress.go
zgalor: `testIngressConnectivity` doesn't fit the `HostedClusterVerifier` interface - it makes an external HTTP connection from outside the VNet rather than querying the data plane via `adminRESTConfig`. Making it a verifier would mean forcing a different signature or ignoring the `restConfig` parameter.

It's a 15-line helper specific to this test's negative assertion (proving the ingress is *not* reachable externally), so I think keeping it local to the test file is the right call.

test/util/framework/sample_app_helper.go
zgalor: Done - extracted `verifiers.DeploySampleApp()` from `verifySimpleWebApp`. Both the serving app verifier and the private ingress test now use the same function and embedded YAML artifacts. Deleted `framework/sample_app_helper.go`.

test/util/framework/sample_app.go
machi1990, line 1: I think it can go in the `framework` pkg.

machi1990 (commented):
Left one suggestion otherwise lgtm but I'd love @mbukatov or @miquelsi with approval rights to have a look at the change as well

test/util/framework/sample_app.go
zgalor: Moved to `framework.DeploySampleApp()` with the embedded YAML artifacts. `verifySimpleWebApp` now calls `framework.DeploySampleApp()` and the private ingress test does the same.

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,241 @@
...
+				nil,
+				framework.ClusterCreationTimeout,
+			)
+			if isAPINotDeployedError(err) {
+				if time.Now().Before(timeBombDeadline) {
+					Skip(fmt.Sprintf("v20260630preview API not yet deployed; skipping until %s", timeBombDeadline.Format(time.RFC3339)))
+				}
+				Fail(fmt.Sprintf("v20260630preview API still not deployed as of %s deadline", timeBombDeadline.Format(time.RFC3339)))
+			}
+			Expect(err).NotTo(HaveOccurred(), "failed to create HCP cluster %q with private ingress", customerClusterName)
```
mbukatov, line 97: I may be better to try to deploy a hosted cluster before trying the test VM, since if the hosted cluster deployment fails on the API not being available, we won't have to spend time and resources on the test vm deployment.

test/e2e/cluster_create_private_ingress.go
mbukatov: I tend to agree with @zgalor here, we don't have a good framework functions for the testing via the external test-vm here. That said, the problem is not with the verifier interface, but the fact that we can't easily execute them from the vm. If we were not using a test vm, I would make sense to just use verifier as usual.

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,241 @@
...
+
+			By("deploying a sample web app to verify ingress connectivity")
+			sampleApp, err := framework.DeploySampleApp(ctx, adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "failed to deploy sample web app for ingress connectivity test")
+
+			appURL := "https://" + sampleApp.RouteHost
+			GinkgoLogr.Info("Sample app deployed", "url", appURL)
+
+			By("verifying ingress is reachable from VM inside the VNet")
+			curlCmd := fmt.Sprintf("curl -k -s -o /dev/null -w '%%{http_code}' --connect-timeout 10 %s", appURL)
```
mbukatov, line 192: Why do we ignore TLS validation here? The url clearly uses TLS.

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,241 @@
...
+				output, runErr := framework.RunVMCommand(ctx, tc, *resourceGroup.Name, vmName, curlCmd, 2*time.Minute)
+				g.Expect(runErr).NotTo(HaveOccurred(), "RunVMCommand failed for ingress connectivity test")
+
+				httpCode := strings.TrimSpace(output)
+				if httpCode != previousOutput {
+					GinkgoLogr.Info("VM ingress connectivity check", "httpCode", httpCode)
+					previousOutput = httpCode
+				}
+				g.Expect(httpCode).NotTo(Equal("000"),
+					"should receive HTTP response from internal LB via VM, got curl error code 000")
```
mbukatov: Shouldn't we check for HTTP 200 code instead?

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,241 @@
...
+			}, 10*time.Minute, 15*time.Second).Should(Succeed())
+
+			By("verifying ingress is NOT reachable from outside the VNet")
+			err = testIngressConnectivity(ctx, appURL, 10*time.Second)
+			Expect(err).To(HaveOccurred(),
+				"private ingress should not be reachable from outside the VNet, but connection succeeded")
+			GinkgoLogr.Info("Confirmed ingress is not reachable from outside the VNet")
+		},
```
mbukatov, line 212: Could you also delete the sample web app in a similar way how the VerifySimpleWebApp does the cleanup?

mbukatov (commented):
Providing few comments related to the E2E test case.

test/e2e/cluster_create_private_ingress.go
zgalor: I removed sample app cleanup per @machi1990's earlier review feedback — the cluster gets deleted at test end, so the namespace goes with it. Happy to add it back if you both prefer explicit cleanup.

test/e2e/cluster_create_private_ingress.go
zgalor: TLS validation is skipped (`-k`) because we're testing network connectivity to the internal LB, not cert validity. Cert provisioning (OneCert → Key Vault → IngressController) has variable latency and is already covered by `VerifySimpleWebApp`. I'll add a comment to make this explicit.

test/e2e/cluster_create_private_ingress.go
zgalor: Good point - reordered to deploy the VM after cluster creation. Both are synchronous calls, so the VM deploys after the cluster is ready. This avoids wasting ~5-10 min on the VM if the API isn't available and the test skips early.

test/e2e/cluster_create_private_ingress.go
zgalor: Good point - changed to check for `200` specifically. The sample app (`agnhost serve-hostname`) always returns 200, so this proves the full path (router → service → pod) works, not just that something responded.

test/e2e/cluster_create_private_ingress.go
```diff
@@ -0,0 +1,243 @@
...
+						"subnetName":   clusterParams.SubnetName,
+						"sshPublicKey": sshPublicKey,
+						"vmSize":       vmSize,
+					}),
+					framework.WithTimeout(30*time.Minute),
+				)
+				if deployErr == nil || strings.Contains(deployErr.Error(), "SkuNotAvailable") {
+					break
+				}
+				time.Sleep(20 * time.Second)
```
ziccardi: This waits 20 seconds even after the last attempt

zgalor: Fixed. Thanks

test/e2e/cluster_create_private_ingress.go
mbukatov: I'm ok with keeping it this way, since this code is not part of a reusable function and the cleanup itself is less likely to be related to the feature under test.

test/e2e/cluster_create_private_ingress.go
mbukatov: Ok, this is not ideal, but explicitly stating this along with a reasoning is good for now (as it will make clear that this is not omission but intention).

PR #5840, opened 2026-06-29
Make kubeapplier outages visible

tmstff:
https://redhat.atlassian.net/browse/ARO-28003

### What

Make kubeapplier outages visible

### Why

to have a trigger for investigating and fixing the problem.

### Testing

TBD

### PR Checklist

- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
- [ ] Screenshots included (if graph/UI/metrics changes)
- [ ] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [ ] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

If E2E tests are included:

- [ ] E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md)
- [ ] If new E2E use case is covered (via a new test or new check/verifier),
  demonstrate that the test is able to detect a defect/error and fail with
  proper error message and logs which communicates nature of the problem.

PR #5901, opened 2026-07-02
Add kube applier desires metric

tmstff:
https://redhat.atlassian.net/browse/ARO-28003

### What

add a metric for the number of kube-applier desires

### Why

To later use it in alerting for checking wether kube-applier has work to do which is not done.

### Testing

There's an alert when the metric is not present - which in effect serves as e2e-test, as it will fail in case of absence

the check if actual work is done when desires are present will be a follow up PR (to be able to play around with "real life" metrics before creating the alert)

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/generatedSLHCPPrometheusAlertingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedSLHCPPrometheusAlertingRules.bicep
@@ -109,3 +109,43 @@ resource hcpClusterOperatorsRules 'Microsoft.AlertsManagement/prometheusRuleGrou
     ]
   }
 }
+
+resource kubeApplier 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-03-01' = {
+  name: 'kube-applier'
+  location: location
+  properties: {
+    interval: 'PT1M'
+    rules: [
+      {
+        actions: [
+          for g in actionGroups: {
+            actionGroupId: g
+            actionProperties: {
+              'IcM.Title': '#$.labels.cluster#: #$.annotations.title#'
+              'IcM.CorrelationId': '#$.annotations.correlationId#'
+            }
+          }
+        ]
+        alert: 'KubeApplierDesiresMetricNotPresent'
+        enabled: true
+        labels: {
+          severity: 'warning'
+        }
+        annotations: {
+          correlationId: 'KubeApplierDesiresMetricNotPresent/{{ $labels.cluster }}'
+          description: 'kube-applier on cluster {{ $labels.cluster }} lacks kube_applier_desires metric for 5 minutes.'
+          info: 'kube-applier on cluster {{ $labels.cluster }} lacks kube_applier_desires metric for 5 minutes.'
+          runbook_url: 'TBD'
+          summary: 'kube-applier on cluster {{ $labels.cluster }} lacks kube_applier_desires metric'
+          title: 'kube-applier on cluster {{ $labels.cluster }} lacks kube_applier_desires metric'
+        }
+        expression: 'kube_applier_health == 1 unless on (cluster, namespace) kube_applier_desires'
+        for: 'PT5M'
+        severity: severityCeiling > 0 ? max(3, severityCeiling) : 3
+      }
+    ]
+    scopes: [
+      azureMonitoring
+    ]
+  }
+}
--- a/kube-applier/alerts/kube-applier-prometheusRule.yaml
+++ b/kube-applier/alerts/kube-applier-prometheusRule.yaml
@@ -13,7 +13,7 @@ spec:
   - name: kube-applier
     rules:
     - alert: KubeApplierDesiresMetricNotPresent
-      # Fires when at least instance of kube-applier is running but the kube_applier_desires metric is not present for 5 minutes.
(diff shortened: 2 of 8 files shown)


kube-applier/pkg/app/kube_applier.go
```diff
@@ -143,6 +143,12 @@ func (o *Options) runControllersUnderLeaderElection(
 	applyInformer := informers.NewApplyDesireInformer(listers.ApplyDesires())
 	readInformer := informers.NewReadDesireInformer(listers.ReadDesires())
 
+	collector := newDesireCollector(
+		applyInformer.GetStore(),
+		readInformer.GetStore(),
+		o.metricsRegisterer(),
+	)
```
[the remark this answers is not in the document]
tmstff: fixed

kube-applier/pkg/app/desire_metrics.go
```diff
@@ -0,0 +1,111 @@
...
+	applyStore  cache.Store
+	deleteStore cache.Store
+	readStore   cache.Store
+	total       *prometheus.GaugeVec
+}
+
+func newDesireCollector(
+	applyStore, deleteStore, readStore cache.Store,
+	registerer prometheus.Registerer,
+) *desireCollector {
```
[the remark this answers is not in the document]
tmstff: fixed

kube-applier/alerts/kube-applier-prometheusRule.yaml
```diff
@@ -0,0 +1,26 @@
...
+    prometheus: k8s
+    role: alert-rules
+  name: kube-applier-monitoring-rules
+  namespace: monitoring
+spec:
+  groups:
+  - name: kube-applier
+    rules:
+    - alert: KubeApplierDesiresMetricNotPresent
+      # Fires when at least instance of kube-applier is running but the kube_applier_desires metric is not present for 5 minutes.
```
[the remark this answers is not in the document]
tmstff: done

dev-infrastructure/modules/metrics/rules/generatedSLHCPPrometheusAlertingRules.bicep
```diff
@@ -109,3 +109,43 @@ resource hcpClusterOperatorsRules 'Microsoft.AlertsManagement/prometheusRuleGrou
...
+          }
+        ]
+        alert: 'KubeApplierDesiresMetricNotPresent'
+        enabled: true
+        labels: {
+          severity: 'warning'
+        }
+        annotations: {
+          correlationId: 'KubeApplierDesiresMetricNotPresent/{{ $labels.cluster }}'
+          description: 'kube-applier on cluster {{ $labels.cluster }} lacks kube_applier_desires metric for 5 minutes.'
```
stevekuznetsov, line 136: Is this separate from the `up` checks? If we have the `up` checks for scraping, do we need this?

stevekuznetsov: Also FYI the way this rolls out will deploy the alert first, then deploy the updated kube-applier component, so we will expect to see this fire during rollout. IMO not an issue for this alert today, but something to keep in mind

dev-infrastructure/modules/metrics/rules/generatedSLHCPPrometheusAlertingRules.bicep
```diff
@@ -109,3 +109,43 @@ resource hcpClusterOperatorsRules 'Microsoft.AlertsManagement/prometheusRuleGrou
...
+        expression: 'kube_applier_health == 1 unless on (cluster, namespace) kube_applier_desires'
+        for: 'PT5M'
+        severity: severityCeiling > 0 ? max(3, severityCeiling) : 3
+      }
+    ]
+    scopes: [
+      azureMonitoring
+    ]
+  }
+}
```
stevekuznetsov, line 151: Were we expecting to see a more involved alert for progression here?

dev-infrastructure/modules/metrics/rules/generatedSLHCPPrometheusAlertingRules.bicep
tmstff: @stevekuznetsov 

> Is this separate from the up checks? If we have the up checks for scraping, do we need this?

As far as I can tell there is no such alert (at least I didn't find one). I could create another PR to introduce that.

The intention for the KubeApplierDesiresMetricNotPresent alert was to substitute an e2e test that verifies that the metric is present. This way we even have the whole chain including prometheus scraping. In case all goes well no harm is done. In case this metric somehow breaks, the e2e test fails with because of the alert.

dev-infrastructure/modules/metrics/rules/generatedSLHCPPrometheusAlertingRules.bicep
tmstff: I planned to do that as a follow up PR, as I can then try it out with real int, stating or prod data (depending on how much patience is there to wait :-D ) The idea how to do it is here: https://github.com/Azure/ARO-HCP/pull/5840/changes#diff-22842587d8e3162e734a327c5c3b2e1b8049c0f229742762a71c09d84be3d1d0

PR #5913, opened 2026-07-03
feat(e2e): add make target to apply slot-managed identity pool

roivaz:
Add a `apply-identity-pool` target to test/Makefile that builds the aro-hcp-tests binary first, which regenerates the Bicep-derived ARM artifacts (e.g. msi-pools.json) from the source-of-truth Bicep before they are embedded. Running the CLI via `go run` or a hand-built binary could embed a stale, git-ignored artifact, causing resource groups to be deleted and recreated instead of updated in place.

Update the identity-leasing CI docs to point at the new target and explain why the raw binary invocation is unsafe.

Related to https://redhat.atlassian.net/browse/ARO-27549


raelga (approved):
/lgtm
/approve

PR #5923, opened 2026-07-03
observability: shove all alerts into Kusto

stevekuznetsov:
observability: shove all alerts into Kusto

We want to be able to query and analyze all of our alerting, and we
likely don't even want all of them to go to humans. Using the built-in
support for EventHub and ingestion into Kusto gives us a low-lift
approach for this.

---

based on https://github.com/Azure/ARO-HCP/pull/5862


dev-infrastructure/templates/audit-logs-eventhub.bicep
```diff
@@ -66,6 +72,20 @@ resource eventHubNamespace 'Microsoft.EventHub/namespaces@2024-01-01' = if (kust
...
+    }
+
+    // Consumer group for alert events Kusto data connection
+    resource kustoConsumerGroup 'consumergroups@2024-01-01' = {
+      name: alertEventsKustoConsumerGroupName
+    }
+  }
 }
 
 var eventHubDataReceiverRole = 'a638d3c7-ab3a-418d-83e6-5f17a39d4fde'
```
geoberle, line 91: i assume we will need a role assigbment also for the `alertEventsEventHub` similar to what we do for `eventHub`

stevekuznetsov:
/label lgtm

This works as expected, confirmed with Sajeel on Slack that we want to send it in.

PR #5903, opened 2026-07-02
Revert "Merge pull request #5751 from patilsuraj767/supatil/ARO-25131-cp-operators-permission"

miguelsorianod:
This reverts commit 71ddfd7ea0badf2af4a2761e1bd61ba5ce44d0cb, reversing changes made to 42521ea3a792d92f8c947b3172e8b94c5e1389d2.

PR #5562, opened 2026-06-08
feat: add containerRegistry to ARM API for ACR pull via managed identity (ARO-24037)

twolff-gh:
## Summary

Add `properties.platform.containerRegistry.managedIdentity` to the v20261001preview ARM API, allowing customers to configure a user-assigned managed identity for worker node ACR image pulls.

[ARO-24037](https://redhat.atlassian.net/browse/ARO-24037)

## What changes

- **ARM API**: `ContainerRegistryProfile` on `PlatformProfile` (v20261001preview). Day-2 mutable via PATCH. Set, change, or clear (null).
- **Admission**: OCP >= 4.22 version gate.
- **Validation**: Resource type check, managed resource group rejection, struct-level platform immutability preserved.
- **Backend**: `ContainerRegistryPullCredentialsPermissionValidation` — checks CAPZ has assign/action on the MI. Re-validates on day-2 MI changes via `InputKeyedClusterValidation`.
- **Dispatch + CS conversion**: Wired through `clusterUpdateDispatchConfig` for day-2 updates and clearing.
- **Cross-version preservation**: Older API versions preserve the field through round-trips.

## Testing

- **E2E** (v20261001preview, skips until deployed): Create cluster with MI → pull from private ACR → day-2 update to different MI → verify pull with new MI.
- **Integration**: PATCH to add containerRegistry, PATCH null to clear. Declarative artifact tests with v20261001preview.
- **Unit**: Validation (create, update, clear, wrong type, same-MRG), admission version gate (4.17/4.21/4.22), backend permission validation (9 scenarios), OCM conversion round-trip, dispatch config, immutability enforcement for all platform fields.

## Companion changes

- HyperShift CPO backport (openshift/hypershift#9211)
- OCM model (#1164)
- Clusters Service

PR #5930, opened 2026-07-06
Reapply "Merge pull request #5751 from patilsuraj767/supatil/ARO-2513…

patilsuraj767:
…1-cp-operators-permissions"

This reverts commit 778af7ccb6bfcccb3df00488c6f2c33627ee22be.

Reapplying reverted changes in https://github.com/Azure/ARO-HCP/pull/5903


patilsuraj767:
The previous [PR](https://github.com/Azure/ARO-HCP/pull/5751) was [reverted](https://github.com/Azure/ARO-HCP/pull/5903) because it triggered the following CI alert:
```
Backend controller workqueue clustervalidationcontrolplaneidentitiespermissionsclustervalidation has a retry ratio of > 50% sustained over 10 minutes, indicating most queue activity is failed retries rather than fresh work.
```
The controller was repeatedly failing and requeuing for two reasons:

1. Cosmos DB precondition failure -  This is a false-positive error rather than an actual issue. The Cluster Validation controller currently treats Cosmos DB precondition failures as errors. This can be addressed in a separate PR.
2. CheckAccess API call failure:
    ```
    (wrapped at control_plane_identities_permissions_cluster_validation.go:95) failed to find missing actions for identity \"cloud-controller-manager\": (wrapped at control_plane_identities_permissions_cluster_validation.go:212) (wrapped at control_plane_identities_permissions_cluster_validation.go:326) invalid character 'T' looking for beginning of value
    ```
    This error originates from the Azure [checkaccess-v2-go-sdk](https://github.com/Azure/checkaccess-v2-go-sdk/blob/717f2142b5cab1cd2e523ae7dce6da6543aee46d/client/client.go#L101). The SDK attempts to unmarshal the response as JSON, but the API returned a plain-text response instead.
    To investigate further, I added debug logging to capture the raw response body. The response was:
    ```
    POST https://westus3.authorization.azure.net/providers/Microsoft.Authorization/checkAccess?api-version=2023-05-01
   Authorization: REDACTED
   Content-Length: 397
   User-Agent: backend/ azsdk-go-aro-pdpclient/0.0.1 (go1.25.11; linux)
   RESPONSE Status: 429 Too Many Requests
   Client-Request-Id: REDACTED
   Content-Length: 17
   Content-Type: text/HTML
   Date: Tue, 07 Jul 2026 05:45:22 GMT
   Retry-After: 3
   Too Many Requests
   ```
    The root cause is that the CheckAccess API is rate-limited (HTTP 429) and returned a plain-text response (Content-Type: text/HTML) instead of JSON. Since the SDK assumes a JSON response, it fails during unmarshalling and returns the misleading error:
    ```
    invalid character 'T' looking for beginning of value
    ```
    As a result, the controller treats the request as a failure and requeues it, eventually causing the workqueue retry ratio alert.

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,393 @@
...
+// Only actions from roleActions/roleDataActions that are relevant to route table resources are checked (via intersection with the known route table action set).
+// It returns:
+//   - (nil, nil) if the subnet has no attached route table, if the identity has all required permissions, or if none of the role's actions apply to route table resources.
+//   - a non-nil *IdentityResourceMissingPermissions populated with the route table resource ID, the identity, and the slice of NotAllowed/Denied decisions, if any permission is missing.
+//   - (nil, error) if the route table resource ID cannot be parsed or the CheckAccessV2 API call fails.
+func (v *ControlPlaneIdentitiesPermissionsClusterValidation) checkMissingPermissionsForRouteTable(ctx context.Context, checkAccessV2Client azureclient.CheckAccessV2Client, clusterSubnet *armnetwork.Subnet, identity *azcorearm.ResourceID, roleActions []string, roleDataActions []string, token azcore.AccessToken) (*identityResourceMissingPermissions, error) {
+	if clusterSubnet.Properties.RouteTable == nil {
+		return nil, nil
+	}
+	routeTableResourceID, err := azcorearm.ParseResourceID(*clusterSubnet.Properties.RouteTable.ID)
```
zgalor, line 290: This is already called later inside `checkNotAllowedAndDeniedActionsForRouteTable`.
You can either pass it in as a parameter or add it as a return value, but we can avoid parsing it twice

backend/pkg/azure/client/check_access_v2_client_throttle_retry.go
```diff
@@ -0,0 +1,76 @@
...
+		func(attempt int, delay time.Duration) {
+			logger.Info("Too many requests, retrying CheckAccessV2 API call after delay", "resourceID", resourceID, "attempt", attempt, "delay", delay.String())
+		},
+		func() (*checkaccessv2.AuthorizationDecisionResponse, error) {
+			return checkAccessV2Client.CheckAccess(ctx, authRequest)
+		},
+	)
+
+	var responseErr *azcore.ResponseError
+	if errors.As(err, &responseErr) && responseErr.StatusCode == http.StatusTooManyRequests {
```
zgalor: Move that block inside the `if err != nil` lock (line 69). 
It is more readable: "if there's an error, check if it's a 429 we should skip; otherwise, propagate."

zgalor: But I have a deeper concern. 

If we've exhausted all retries and the API is still throttled, that means the retry mechanism failed to resolve the issue. Silently skipping the check means we have no observability into how often this happens - if it's persistent, the validation is effectively permanently disabled for affected clusters.                                                                                                                                             
                                                                
Can we add a counter metric (e.g., `arohcp_checkaccess_v2_throttle_exhausted_total`) that gets incremented here? That way we can alert on it and distinguish "occasional throttle resolved by retry" from "systemic rate limit problem."

zgalor:
The existing tests for the retry mechanism all use mocks that simulate 429 responses on command. Can we add an e2e test that actually triggers rate limiting against the MockFPA? For example, fire 100+ CheckAccessV2 calls in a tight window to guarantee we hit the real rate limit, then verify the retry mechanism handles the real 429s with real `Retry-After` headers and eventually succeeds.

nimrodshn:
LGTM :100: 
TY!

miguelsorianod:
We should not follow this approach, as this is going to block the controller threads.

What we need is to improve the validation controllers so we can signal requeuing after a desired time. Then the validation can return whatever intent is desired

backend/cmd/root.go
```diff
@@ -358,6 +359,12 @@ func (f *BackendRootCmdFlags) ToBackendOptions(ctx context.Context, cmd *cobra.C
 			fpaTokenCredRetriever, azureConfig.CloudEnvironment.CheckAccessV2Endpoint(f.AzureLocation),
 			azureConfig.CloudEnvironment.CheckAccessV2Scope(), azureConfig.CloudEnvironment.AZCoreClientOptions(),
 		)
+
+		// Throttle our aggregate call rate to stay under the real FPA identity's documented CheckAccessV2 limit of 500 requests/second, rather than reacting to HTTP 429 responses after the fact.
```
miguelsorianod: "rather than reacting to HTTP 429 responses after the fact." this code seems unnecessary once this is merged

backend/cmd/root.go
```diff
@@ -373,6 +380,13 @@ func (f *BackendRootCmdFlags) ToBackendOptions(ctx context.Context, cmd *cobra.C
 			azureConfig.CloudEnvironment.CheckAccessV2Scope(), azureConfig.CloudEnvironment.AZCoreClientOptions(),
 		)
 
+		// Throttle our aggregate call rate to stay under the Azure Permissions Manager identity's documented CheckAccessV2 limit of 25 requests per 5 seconds, rather than reacting to HTTP 429
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/5930/changes/3925c039ae01bd1f21fa5f5152de2711ffad0e30#r3628571728

backend/pkg/azure/client/check_access_v2_client_rate_limiter.go
```diff
@@ -0,0 +1,92 @@
...
+// CheckAccess waits for a token from the shared rate limiter (returning its error, e.g. from context cancellation, without calling the API if one isn't available) and then delegates to the inner client.
+func (c *rateLimitedCheckAccessV2Client) CheckAccess(ctx context.Context, authzReq checkaccessv2.AuthorizationRequest) (*checkaccessv2.AuthorizationDecisionResponse, error) {
+	if err := c.rateLimiter.Wait(ctx); err != nil {
+		return nil, err
+	}
+	return c.inner.CheckAccess(ctx, authzReq)
+}
+
+// CreateAuthorizationRequest is local request construction rather than an API call, so it is not subject to rate limiting; it delegates directly to the inner client.
+func (c *rateLimitedCheckAccessV2Client) CreateAuthorizationRequest(resourceId string, actions []string, jwtToken string) (*checkaccessv2.AuthorizationRequest, error) {
```
miguelsorianod: resourceId -> resourceID.

In this project we uppercase all acronyms

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: Can we change the file name to rate_limited_check_access_v2_client.go ?

backend/cmd/root.go
```diff
@@ -356,6 +359,12 @@ func (f *BackendRootCmdFlags) ToBackendOptions(ctx context.Context, cmd *cobra.C
 			fpaTokenCredRetriever, azureConfig.CloudEnvironment.CheckAccessV2Endpoint(f.AzureLocation),
 			azureConfig.CloudEnvironment.CheckAccessV2Scope(), azureConfig.CloudEnvironment.AZCoreClientOptions(),
 		)
+
+		// Throttle our aggregate call rate to stay under the real FPA identity's documented CheckAccessV2 limit of 500 requests/second, rather than reacting to HTTP 429 responses after the fact.
+		checkAccessV2ClientBuilder = azureclient.NewRateLimitedCheckAccessV2ClientBuilder(
+			checkAccessV2ClientBuilder,
+			flowcontrol.NewTokenBucketRateLimiter(azureclient.CheckAccessV2RealFPARateLimiterQPS, azureclient.CheckAccessV2RealFPARateLimiterBurst),
```
patilsuraj767: @miguelsorianod I think the token bucket rate limiter would introduce unnecessary overhead in Stage and Prod, where we use the 1P app with a much higher rate limit. Should we skip creating the RateLimitedCheckAccessV2ClientBuilder in those higher environments? Thoughts?

backend/pkg/azure/client/check_access_v2_client_rate_limiter.go
```diff
@@ -0,0 +1,92 @@
...
+// tenants/callers rather than per-client.
+type rateLimitedCheckAccessV2ClientBuilder struct {
+	inner       CheckAccessV2ClientBuilder
+	rateLimiter flowcontrol.RateLimiter
+}
+
+var _ CheckAccessV2ClientBuilder = (*rateLimitedCheckAccessV2ClientBuilder)(nil)
+
+// NewRateLimitedCheckAccessV2ClientBuilder wraps inner so that every CheckAccessV2Client it builds shares rateLimiter, throttling calls to CheckAccess to stay within the CheckAccessV2 API's rate limit. Callers
+// should construct rateLimiter once (e.g. via flowcontrol.NewTokenBucketRateLimiter) and share a single instance across the process, since Build may be called many times (once per tenant) and the API's rate
```
miguelsorianod: It is not fully clear as of now if we should be sharing the rate-limiter cross-tenant. We currently do not have a full idea of how the rate-limit applies. The documentation mentions: "The API has a default rate limit of 500 RPS per application/tenant/region" but that's somehow ambiguous.

The current questions that need an answer, which will help us take a decision are:
* Do we have different FPAs per regions?
* Does the "application" in that context means the app used to interact with the CheckAccessAPI, or the identity/app that permissions are checked for?
* When having a multi-tenant app (like the FPA), what is considered "application/tenant" ? does it apply to the tenant where the app is created, or to the tenants it interacts with?

backend/pkg/azure/client/tenant_rate_limiter_cache.go
```diff
@@ -0,0 +1,80 @@
...
+// Rounded up to absorb allocator overhead and insulate against the estimate drifting slightly as dependencies change.
+const estimatedBytesPerCachedTenant = 400
+
+// tenantRateLimiterCache is a memory-bounded, least-recently-used cache mapping tenant IDs to their own flowcontrol.RateLimiter. It lets rateLimitedCheckAccessV2ClientBuilder keep a dedicated rate
+// limiter per tenant without letting its memory footprint grow without bound for the lifetime of the process: once the cache's entry-count capacity (derived from its memory budget) is reached, caching
+// a new tenant evicts the least-recently-used tenant's rate limiter to make room. The LRU eviction mechanics themselves are delegated to k8s.io/utils/lru (the same LRU cache already used elsewhere in
+// this codebase, e.g. internal/controllerutils.TimeBasedCooldownChecker and the *ClusterServiceDeleteDispatchSyncer controllers), which is thread-safe for individual Get/Add calls; tenantRateLimiterCache
+// only adds a lock around the check-then-create sequence in getOrCreate, since the cache has no atomic "get existing or else compute and store" operation.
+type tenantRateLimiterCache struct {
+	cache *lru.Cache
```
miguelsorianod: There is no need for the mutex. The lru cache is already thread-safe.

We can also keep it simple. An arbitrary number of items should be ok. Let's put 10000 for example

miguelsorianod: Unless there's some reason for the additional `tenantRateLimiterCache`, let's remove it keeping it simple. Use the lru cache directly

miguelsorianod: Wherever we end up placing the cache, document the key and the value (types, meaning, ...)

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,103 @@
...
+	return c.inner.CreateAuthorizationRequest(resourceID, actions, jwtToken)
+}
+
+// rateLimitedCheckAccessV2ClientBuilder wraps a CheckAccessV2ClientBuilder so that every client it builds for a given tenant ID shares that tenant's own rateLimiter, throttling the CheckAccessV2 call rate
+// per tenant rather than across all tenants/callers, since the CheckAccessV2 API's documented rate limits apply per calling application/tenant/region, not to the caller as a whole. Per-tenant rate limiters
+// are kept in a memory-bounded tenantRateLimiterCache so memory use stays bounded regardless of how many distinct tenants are seen over the process's lifetime.
+type rateLimitedCheckAccessV2ClientBuilder struct {
+	inner          CheckAccessV2ClientBuilder
+	newRateLimiter func() flowcontrol.RateLimiter
+	rateLimiters   *tenantRateLimiterCache
```
miguelsorianod: Here is where the lru would be

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+var _ CheckAccessV2ClientBuilder = (*rateLimitedCheckAccessV2ClientBuilder)(nil)
+
+// NewRateLimitedCheckAccessV2ClientBuilder wraps inner so that every CheckAccessV2Client it builds for a given tenant ID shares a rate limiter dedicated to that tenant, throttling calls to CheckAccess to
+// stay within the CheckAccessV2 API's per-tenant rate limit. newRateLimiter is called at most once per distinct tenant ID (the first time Build is invoked for it) to lazily create and cache that tenant's
+// rate limiter, e.g. `func() flowcontrol.RateLimiter { return flowcontrol.NewTokenBucketRateLimiter(qps, burst) }`. maxCacheBytes bounds how much memory the per-tenant rate limiter cache may use at
+// once, evicting the least-recently-used tenant once exhausted (see CheckAccessV2RateLimiterMaxCacheBytes). The returned builder must be reused across calls (rather than reconstructed per Build call)
+// for the per-tenant caching to take effect.
+func NewRateLimitedCheckAccessV2ClientBuilder(inner CheckAccessV2ClientBuilder, newRateLimiter func() flowcontrol.RateLimiter, maxCacheBytes int) CheckAccessV2ClientBuilder {
```
miguelsorianod: Here we would instantiate the lru

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,103 @@
...
+		newRateLimiter: newRateLimiter,
+		rateLimiters:   newTenantRateLimiterCache(maxCacheBytes),
+	}
+}
+
+func (b *rateLimitedCheckAccessV2ClientBuilder) Build(tenantID string) (CheckAccessV2Client, error) {
+	inner, err := b.inner.Build(tenantID)
+	if err != nil {
+		return nil, err
+	}
```
miguelsorianod, line 107: here we would make sure we have the rate-limiter per tenant (get from cache, add to cache if it doesn't exist and return)

backend/pkg/azure/client/rate_limited_check_access_v2_client_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+		client := &rateLimitedCheckAccessV2Client{inner: mockInner, rateLimiter: flowcontrol.NewFakeNeverRateLimiter()}
+
+		result, err := client.CheckAccess(context.Background(), fakeAuthRequest)
+		require.Error(t, err)
+		assert.Nil(t, result)
+	})
+}
+
+func TestRateLimitedCheckAccessV2Client_CreateAuthorizationRequest(t *testing.T) {
```
miguelsorianod: This test is not needed. That method doesn't use the rate limit so testing that no rate limit is used is unnecessary

miguelsorianod:
@patilsuraj767 if https://github.com/Azure/ARO-HCP/pull/5562 is merged before this PR this PR should change that validation to use the rate-limiter

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,130 @@
...
+// to k8s.io/utils/lru (the same LRU cache already used elsewhere in this codebase, e.g. internal/controllerutils.TimeBasedCooldownChecker and the *ClusterServiceDeleteDispatchSyncer controllers),
+// which is thread-safe for individual Get/Add calls; rateLimitedCheckAccessV2ClientBuilder only adds mu around the check-then-create sequence in getOrCreateRateLimiter, since the cache has no atomic
+// "get existing or else compute and store" operation.
+type rateLimitedCheckAccessV2ClientBuilder struct {
+	inner          CheckAccessV2ClientBuilder
+	newRateLimiter func() flowcontrol.RateLimiter
+	// rateLimiters caches each tenant's own flowcontrol.RateLimiter, keyed by tenant ID (string) and storing a flowcontrol.RateLimiter as the value.
+	rateLimiters *lru.Cache
+	// mu guards access to rateLimiters, ensuring that only one goroutine can add or evict a tenant's rate limiter at a time.
+	mu sync.Mutex
```
patilsuraj767: lru.Cache is thread-safe for individual Get/Add calls in isolation, the mutex here isn't protecting the LRU's internals, it protects "check cache, and if missing, create + store" sequence in getOrCreate. 
Check the test case - `creates only one rate limiter for a tenant even when getOrCreateRateLimiter is called concurrently for it`

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,132 @@
...
+	CheckAccessV2InsecureARMPermissionsManagerRateLimiterQPS = 5
+
+	// CheckAccessV2InsecureARMPermissionsManagerRateLimiterBurst is the maximum number of CheckAccessV2 requests allowed to run back-to-back, under the ARMPermissionsManager identity's rate limit,
+	// before throttling starts, matching its 25-request burst window.
+	CheckAccessV2InsecureARMPermissionsManagerRateLimiterBurst = 25
+
+	// CheckAccessV2RateLimiterCacheSize bounds how many distinct tenants' rate limiters rateLimitedCheckAccessV2ClientBuilder caches at once. 10000 is comfortably enough to cache every distinct
+	// Azure tenant expected to have HCP clusters validated by a single backend replica, while still capping worst-case memory growth instead of letting the cache grow for the lifetime of the
+	// process. Once the limit is exceeded, building a client for a new tenant evicts the least-recently-used tenant's rate limiter.
+	CheckAccessV2RateLimiterCacheSize = 10000
```
miguelsorianod: No need to have this as exported as it's not used anywhere else.

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"k8s.io/client-go/util/flowcontrol"
+	"k8s.io/utils/lru"
+
+	checkaccessv2 "github.com/Azure/checkaccess-v2-go-sdk/client"
+)
+
+const (
+	// CheckAccessV2RealFPARateLimiterQPS is the sustained request rate, in queries per second, allowed against the CheckAccessV2 API when calling it as the real FPA identity, which has a documented
+	// limit of 500 requests per second.
```
miguelsorianod: Put references to doc for both the high and the low limits if possible

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,132 @@
...
+	return c.inner.CheckAccess(ctx, authzReq)
+}
+
+// CreateAuthorizationRequest is local request construction rather than an API call, so it is not subject to rate limiting; it delegates directly to the inner client.
+func (c *rateLimitedCheckAccessV2Client) CreateAuthorizationRequest(resourceID string, actions []string, jwtToken string) (*checkaccessv2.AuthorizationRequest, error) {
+	return c.inner.CreateAuthorizationRequest(resourceID, actions, jwtToken)
+}
+
+// rateLimitedCheckAccessV2ClientBuilder wraps a CheckAccessV2ClientBuilder so that every client it builds for a given tenant ID shares that tenant's own rateLimiter, throttling the CheckAccessV2 call rate
+// per tenant rather than across all tenants/callers, since the CheckAccessV2 API's documented rate limits apply per calling application/tenant/region, not to the caller as a whole. Per-tenant rate limiters
```
miguelsorianod: Put reference to doc where it explains what is being described here

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,132 @@
...
+// are kept in a size-bounded, least-recently-used rateLimiters cache (see CheckAccessV2RateLimiterCacheSize) so memory use stays bounded regardless of how many distinct tenants are seen over the
+// process's lifetime: once the cache's capacity is reached, caching a new tenant evicts the least-recently-used tenant's rate limiter to make room. The LRU eviction mechanics themselves are delegated
```
miguelsorianod: >  The LRU eviction mechanics themselves are delegated
// to k8s.io/utils/lru (the same LRU cache already used elsewhere in this codebase, e.g. internal/controllerutils.TimeBasedCooldownChecker and the *ClusterServiceDeleteDispatchSyncer controllers),
// which is thread-safe for individual Get/Add calls; rateLimitedCheckAccessV2ClientBuilder only adds mu around the check-then-create sequence in getOrCreateRateLimiter, since the cache has no atomic
// "get existing or else compute and store" operation.

all this part seems unnecessary. Remove

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
miguelsorianod: 👍 . Rename to rateLimitersMu.

backend/pkg/azure/client/rate_limited_check_access_v2_client_test.go
```diff
@@ -0,0 +1,331 @@
...
+
+		builder := NewRateLimitedCheckAccessV2ClientBuilder(mockInnerBuilder, 1, 1)
+
+		result, err := builder.Build("tenant-id")
+		require.Error(t, err)
+		assert.Equal(t, wantErr, err)
+		assert.Nil(t, result)
+	})
+
+	t.Run("builds a client whose rate limiter admits requests up to the given burst before throttling", func(t *testing.T) {
```
miguelsorianod, line 84: This test will be flaky. Even though the qps is set to a very low level nothing guarantees it will always execute as expected (cpu scheduling, cpu throttling, ci delays, ...). See if you can leverage a fake clock to  do what you want to exercise

backend/pkg/azure/client/rate_limited_check_access_v2_client_test.go
```diff
@@ -0,0 +1,331 @@
...
+
+		newRateLimiterCalls := 0
+		builder := newTestBuilder(func() flowcontrol.RateLimiter {
+			newRateLimiterCalls++
+			return flowcontrol.NewFakeAlwaysRateLimiter()
+		}, mockInnerBuilder)
+
+		// Fill the cache to its capacity with distinct tenants, then request one more (tenant-0, the least-recently-used, since it was inserted first and never re-accessed): this evicts tenant-0's
+		// cached rate limiter to make room.
+		for i := range CheckAccessV2RateLimiterCacheSize {
```
miguelsorianod: Unnecessary to go over the 10k items. Create a builder with the minimum cache size to do what you want to do

backend/pkg/azure/client/rate_limited_check_access_v2_client_test.go
```diff
@@ -0,0 +1,331 @@
...
+		builder.getOrCreateRateLimiter("tenant-a")
+
+		// tenant-c exceeds maxSize, so tenant-b (now least-recently-used) is evicted, not tenant-a.
+		builder.getOrCreateRateLimiter("tenant-c")
+
+		rateLimiterAAgain := builder.getOrCreateRateLimiter("tenant-a")
+		assert.Same(t, rateLimiterA, rateLimiterAAgain, "tenant-a should still be cached and not require a new rate limiter")
+	})
+
+	t.Run("creates only one rate limiter for a tenant even when getOrCreateRateLimiter is called concurrently for it", func(t *testing.T) {
```
miguelsorianod, line 318: For this we have to be aware that there's no guarantee this will be exercised as intended as it's non deterministic. It's a best effort. Let's add a note to it

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
patilsuraj767: Done, renamed the variable.

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
patilsuraj767: Done

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
patilsuraj767: We don't have the reference doc for low limits (for 3P app) but I added the doc for high limit for 1P app.

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
patilsuraj767: Removed

backend/pkg/azure/client/rate_limited_check_access_v2_client_test.go
patilsuraj767: It is not possible to inject a fake clock, So instead of racing Wait(ctx) against a context deadline, I switched to asserting bucket exhaustion via `TryAccept()`. `TryAccept() `makes a single, non-blocking, immediate decision, returns false immediately if token not available.

backend/pkg/azure/client/rate_limited_check_access_v2_client_test.go
patilsuraj767: Documented it.

backend/pkg/azure/client/rate_limited_check_access_v2_client_test.go
miguelsorianod: Are you sure it's not possible to inject a fake clock? I see https://pkg.go.dev/k8s.io/client-go/util/flowcontrol provides the ability to receive a clock

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// CheckAccessV2InsecureARMPermissionsManagerRateLimiterQPS is the sustained request rate, in queries per second, allowed against the CheckAccessV2 API when calling it as the (insecure,
+	// non-production) ARMPermissionsManager identity, which has a documented limit of 25 requests per 5 seconds (25/5 = 5 QPS).
+	CheckAccessV2InsecureARMPermissionsManagerRateLimiterQPS = 5
+
+	// CheckAccessV2InsecureARMPermissionsManagerRateLimiterBurst is the maximum number of CheckAccessV2 requests allowed to run back-to-back, under the ARMPermissionsManager identity's rate limit,
+	// before throttling starts, matching its 25-request burst window.
+	CheckAccessV2InsecureARMPermissionsManagerRateLimiterBurst = 25
+
+	// CheckAccessV2RateLimiterCacheSize bounds how many distinct tenants' rate limiters rateLimitedCheckAccessV2ClientBuilder caches at once. 10000 is comfortably enough to cache every distinct
```
miguelsorianod: nitpick: code comments still contain the uppercase on this in several places

backend/pkg/azure/client/rate_limited_check_access_v2_client_test.go
patilsuraj767: Yes we do have [newTokenBucketRateLimiterWithClock](https://github.com/kubernetes/client-go/blob/master/util/flowcontrol/throttle.go#L98) where clock can be passed but only `TryAccept` and `Accept` are the methods that touch clock which we pass. `Wait` just forwards straight to the *rate.Limiter's own Wait(ctx). `rate.Limiter.Wait` is itself hardcoded to real time - https://github.com/golang/time/blob/master/rate/rate.go#L249

backend/pkg/azure/client/rate_limited_check_access_v2_client.go
patilsuraj767: Fixed

backend/pkg/azure/client/rate_limited_check_access_v2_client_test.go
miguelsorianod: I see. Unfortunately rate.Limit is not an interface. Even with the approach you mention though that doesn't seem to be deterministic as you still depend on time and scheduling. Can we combine the fake clock with that approach?

patilsuraj767: Done, I have added the fake clock.

miguelsorianod:
/lgtm
/approve

let's keep the hold until we see e2e results and then we check both:
* kusto logs
* metrics like (controller requeues ensuring there's no surges, timings, ...)

miguelsorianod:
Let's keep the hold for now because we observe rate-limits still occur. This signals that there's no understanding on how the checkaccess2 rate limits work, or how the token bucket rate limit works/behaves.

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,425 @@
...
+}
+
+var _ ClusterValidation = (*ControlPlaneIdentitiesPermissionsClusterValidation)(nil)
+
+func (v *ControlPlaneIdentitiesPermissionsClusterValidation) Validate(ctx context.Context, clusterSubscription *coreapi.Subscription, cluster *coreapi.HCPOpenShiftCluster) ValidationResult {
+	checkAccessV2Client, err := v.checkAccessV2ClientBuilder.Build(*clusterSubscription.Properties.TenantId)
+	if err != nil {
+		return UnknownValidation(
+			"InternalError",
+			"Failed to build CheckAccessV2 client.",
```
miguelsorianod: This is the public message. It exposes too many internal details

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,425 @@
...
+			ControllerReportingPolicyTypeError,
+		)
+	}
+
+	subnetsClient, err := v.smiClientBuilder.SubnetsClient(ctx, cluster.ServiceProviderProperties.ManagedIdentitiesDataPlaneIdentityURL,
+		cluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ServiceManagedIdentity, cluster.ID.SubscriptionID)
+	if err != nil {
+		return UnknownValidation(
+			"InternalError",
+			"Failed to get subnets client.",
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/5930/changes#r3750730392

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,425 @@
...
+		)
+	}
+
+	// Fetch the subnet details to validate attached subnet devices permissions.
+	subnetResourceID := cluster.CustomerProperties.Platform.SubnetID
+	subnet, err := subnetsClient.Get(ctx, subnetResourceID.ResourceGroupName, subnetResourceID.Parent.Name, subnetResourceID.Name, nil)
+	if err != nil {
+		return UnknownValidation(
+			"InternalError",
+			"Failed to get subnet.",
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/5930/changes#r3750730392

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,425 @@
...
+		)
+	}
+
+	var missingPermissions []*identityResourceMissingPermissions
+	for operatorName, identity := range cluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ControlPlaneOperators {
+		results, err := v.findMissingActionsForIdentity(ctx, checkAccessV2Client, cluster, operatorName, identity, &subnet.Subnet)
+		if err != nil {
+			return UnknownValidation(
+				"InternalError",
+				"Failed to find missing actions for identity.",
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/5930/changes#r3750730392

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,425 @@
...
+		}
+		missingPermissions = append(missingPermissions, results...)
+	}
+
+	if len(missingPermissions) > 0 {
+		jsonBytes, err := json.Marshal(missingPermissions)
+		if err != nil {
+			return UnknownValidation(
+				"InternalError",
+				"Failed to marshal missing permissions.",
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/5930/changes#r3750730392

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
```diff
@@ -0,0 +1,425 @@
...
+				fmt.Sprintf("failed to marshal missing permissions: %v", err),
+				ControllerReportingPolicyTypeError,
+			)
+		}
+		return FailedValidation(
+			"InternalError",
+			"Control plane operators missing required permissions.",
```
miguelsorianod: Here we can expose the missing permissions as that's useful to the user

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
patilsuraj767: fixed

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
patilsuraj767: fixed

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
patilsuraj767: fixed

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
patilsuraj767: fixed

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
patilsuraj767: fixed

backend/pkg/utils/validationutils/control_plane_identities_permissions_cluster_validation.go
patilsuraj767: fixed

PR #5931, opened 2026-07-06
feat: add autoscaler ReadDesire and update operation gate for cluster autoscaler 

shitaljante:
### What

-  Adds a ReadDesire for the cluster-autoscaler ControlPlaneComponent (in addition to the existing HostedCluster ReadDesire).
- Read the autoscaler ControlPlaneComponent from ReadDesire cache and checks Available + RolloutComplete readiness.
- operationClusterUpdate waits for autoscaler readiness 

### Testing
Created cluster using frontend service. Then update cluster autoscaler properties using request payload as - 
```
payload - {
  "properties": {
    "autoscaling": {
      "maxNodesTotal": 200,
      "maxPodGracePeriodSeconds": 1200,
      "podPriorityThreshold": -10
    }
  }
}
```

> curl -s -X PATCH \
>  "http://localhost:8443/${CLUSTER_RESOURCE_ID}?api-version=${API_VERSION}" \
>  -H "Content-Type: application/json" \
>  -H "X-Ms-Arm-Resource-System-Data: {\"createdBy\":\"aro-hcp-local-testing\",\"createdByType\":\"User\",\"createdAt\":\"$(date -u +%Y-%m-%dT%H:%M:%SZ)\",\"lastModifiedBy\":\"aro-hcp-local-testing\",\"lastModifiedByType\":\"User\",\"lastModifiedAt\":\"$(date -u +%Y-%m-%dT%H:%M:%SZ)\"}" \
>  -d @ auto_scalar.json | jq .

After Successful update - 

image

Read desire created with "controlplanecomponents" as resource for a cluster.
image (1)

- Unit tests

E2E tests are included as part of https://redhat.atlassian.net/browse/ARO-27730

### DIFF
--- a/backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
+++ b/backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
@@ -16,6 +16,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -27,6 +28,7 @@ import (
 	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
 	"github.com/Azure/ARO-HCP/backend/pkg/informers"
 	"github.com/Azure/ARO-HCP/backend/pkg/listers"
+	"github.com/Azure/ARO-HCP/backend/pkg/maestrohelpers"
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/api/kubeapplier"
 	controllerutil "github.com/Azure/ARO-HCP/internal/controllerutils"
@@ -143,13 +145,6 @@ func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key
 	}
 	csClusterID := existingCluster.ServiceProviderProperties.ClusterServiceID.ID()
 
-	target := hostedClusterTarget(c.hostedClusterNamespaceEnvIdentifier, csClusterID, csClusterDomainPrefix)
-	desired := buildReadDesire(
-		kubeapplier.ToClusterScopedReadDesireResourceIDString(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, readDesireNameReadonlyHostedCluster),
-		mcResourceID,
-		target,
-	)
-
 	kaClient := c.kubeApplierDBClients.For(ctx, mcResourceID)
 	if kaClient == nil {
 		// Registry doesn't have an entry yet for this MC (e.g. the fleet
@@ -160,27 +155,29 @@ func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key
 	}
 	crud, err := kaClient.ReadDesiresForCluster(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
 	if err != nil {
-		return utils.TrackError(fmt.Errorf("get ReadDesire CRUD: %w", err))
+		return utils.TrackError(fmt.Errorf("failed to get ReadDesire CRUD: %w", err))
 	}
-	existing, err := getExistingReadDesire(ctx, crud, readDesireNameReadonlyHostedCluster)
-	if err != nil {
-		return err
+
+	desiredReadDesires := []*kubeapplier.ReadDesire{
+		buildReadDesire(
+			kubeapplier.ToClusterScopedReadDesireResourceIDString(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, readDesireNameReadonlyHostedCluster),
+			mcResourceID,
+			hostedClusterTarget(c.hostedClusterNamespaceEnvIdentifier, csClusterID, csClusterDomainPrefix),
+		),
+		buildReadDesire(
+			kubeapplier.ToClusterScopedReadDesireResourceIDString(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, maestrohelpers.ReadDesireNameReadonlyHypershiftControlPlaneComponentClusterAutoscaler),
+			mcResourceID,
+			clusterAutoscalerTarget(c.hostedClusterNamespaceEnvIdentifier, csClusterID, csClusterDomainPrefix),
+		),
 	}
-	if !readDesireNeedsWork(existing, desired) {
-		return nil
-	}
-	if existing == nil {
-		if _, err := crud.Create(ctx, desired, nil); err != nil {
-			return utils.TrackError(fmt.Errorf("create ReadDesire: %w", err))
+
+	var errs []error
+	for _, desired := range desiredReadDesires {
+		if err := c.ensureReadDesire(ctx, crud, desired); err != nil {
+			errs = append(errs, err)
 		}
-		return nil
 	}
-	replacement := existing.DeepCopy()
-	replacement.Spec = *desired.Spec.DeepCopy()
-	if _, err := crud.Replace(ctx, replacement, nil); err != nil {
-		return utils.TrackError(fmt.Errorf("replace ReadDesire: %w", err))
-	}
-	return nil
+	return errors.Join(errs...)
 }
 
 // readDesireNameReadonlyHostedCluster is the well-known ReadDesire name
@@ -202,6 +199,18 @@ func hostedClusterTarget(envIdentifier, csClusterID, csClusterDomainPrefix strin
 		Resource:  "hostedclusters",
 		Namespace: hostedClusterNamespace(envIdentifier, csClusterID),
 		Name:      csClusterDomainPrefix,
+	}
+}
+
+// clusterAutoscalerTarget builds the ResourceReference for the cluster-autoscaler
+// ControlPlaneComponent in the HCP control plane namespace.
+func clusterAutoscalerTarget(envIdentifier, csClusterID, csClusterDomainPrefix string) kubeapplier.ResourceReference {
+	return kubeapplier.ResourceReference{
+		Group:     hsv1beta1.SchemeGroupVersion.Group,
+		Version:   hsv1beta1.SchemeGroupVersion.Version,
+		Resource:  "controlplanecomponents",
+		Namespace: hostedControlPlaneNamespace(envIdentifier, csClusterID, csClusterDomainPrefix),
+		Name:      "cluster-autoscaler",
 	}
 }
 
@@ -231,7 +240,7 @@ func getExistingReadDesire(
 		return nil, nil
 	}
 	if err != nil {
-		return nil, utils.TrackError(fmt.Errorf("get ReadDesire: %w", err))
+		return nil, utils.TrackError(fmt.Errorf("failed to get ReadDesire: %w", err))
 	}
 	return existing, nil
 }
@@ -247,4 +256,36 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
 		return true
 	}
 	return existing.Spec.TargetItem != desired.Spec.TargetItem
+}
+
+// ensureReadDesire creates or updates a ReadDesire when the desired spec differs from cosmos.
+func (c *createClusterScopedReadDesiresSyncer) ensureReadDesire(ctx context.Context, crud database.ResourceCRUD[kubeapplier.ReadDesire, *kubeapplier.ReadDesire], desired *kubeapplier.ReadDesire) error {
+	existing, err := getExistingReadDesire(ctx, crud, desired.ResourceID.Name)
+	if err != nil {
+		return err
+	}
+	if !readDesireNeedsWork(existing, desired) {
+		return nil
+	}
+	name := desired.ResourceID.Name
+	if existing == nil {
+		_, err := crud.Create(ctx, desired, nil)
+		if database.IsConflictError(err) {
+			return nil
+		}
+		if err != nil {
+			return utils.TrackError(fmt.Errorf("failed to create ReadDesire %q: %w", name, err))
+		}
+		return nil
+	}
+	replacement := existing.DeepCopy()
+	replacement.Spec = *desired.Spec.DeepCopy()
+	_, err = crud.Replace(ctx, replacement, nil)
+	if database.IsPreconditionFailedError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to replace ReadDesire %q: %w", name, err))
+	}
+	return nil
 }
--- a/backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
+++ b/backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
@@ -1,0 +1,262 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controllers
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/go-logr/logr/testr"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+
+	hsv1beta1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/backend/pkg/listertesting"
+	"github.com/Azure/ARO-HCP/backend/pkg/maestrohelpers"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	"github.com/Azure/ARO-HCP/internal/api/kubeapplier"
+	"github.com/Azure/ARO-HCP/internal/databasetesting"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+const (
+	readDesireTestSubscriptionID    = "00000000-0000-0000-0000-000000000000"
+	readDesireTestResourceGroupName = "test-rg"
+	readDesireTestClusterName       = "test-cluster"
+	readDesireTestEnvIdentifier     = "int"
+	readDesireTestDomainPrefix      = "cluster1"
+	readDesireTestClusterServiceID  = "/api/clusters_mgmt/v1/clusters/abc123"
+)
+
+var readDesireTestManagementClusterResourceID = api.Must(azcorearm.ParseResourceID(
+	"/providers/microsoft.redhatopenshift/stamps/1/managementclusters/default",
+))
+
+func readDesireTestKey() controllerutils.HCPClusterKey {
+	return controllerutils.HCPClusterKey{
+		SubscriptionID:    readDesireTestSubscriptionID,
+		ResourceGroupName: readDesireTestResourceGroupName,
+		HCPClusterName:    readDesireTestClusterName,
+	}
+}
+
+func newTestCluster(opts ...func(*api.HCPOpenShiftCluster)) *api.HCPOpenShiftCluster {
+	resourceID := api.Must(azcorearm.ParseResourceID(
+		"/subscriptions/" + readDesireTestSubscriptionID +
+			"/resourceGroups/" + readDesireTestResourceGroupName +
+			"/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/" + readDesireTestClusterName,
+	))
+	cluster := &api.HCPOpenShiftCluster{
+		CosmosMetadata: arm.CosmosMetadata{
+			ResourceID:   resourceID,
+			PartitionKey: strings.ToLower(resourceID.SubscriptionID),
+		},
+		TrackedResource: arm.TrackedResource{
+			Resource: arm.Resource{
+				ID:   resourceID,
+				Name: readDesireTestClusterName,
+				Type: resourceID.ResourceType.String(),
+			},
+		},
+		ServiceProviderProperties: api.HCPOpenShiftClusterServiceProviderProperties{
+			ClusterServiceID: api.Ptr(api.Must(api.NewInternalID(readDesireTestClusterServiceID))),
+		},
+		CustomerProperties: api.HCPOpenShiftClusterCustomerProperties{
+			DNS: api.CustomerDNSProfile{
+				BaseDomainPrefix: readDesireTestDomainPrefix,
+			},
+			Version: api.VersionProfile{
+				ID: "4.20.0",
+			},
+		},
+	}
+	for _, opt := range opts {
+		opt(cluster)
+	}
+	return cluster
+}
+
+func newTestSPC(mcResourceID *azcorearm.ResourceID) *api.ServiceProviderCluster {
+	spcResourceID := api.Must(azcorearm.ParseResourceID(
+		"/subscriptions/" + readDesireTestSubscriptionID +
+			"/resourceGroups/" + readDesireTestResourceGroupName +
+			"/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/" + readDesireTestClusterName +
+			"/serviceProviderClusters/" + api.ServiceProviderClusterResourceName,
+	))
+	return &api.ServiceProviderCluster{
+		CosmosMetadata: api.CosmosMetadata{
+			ResourceID:   spcResourceID,
+			PartitionKey: strings.ToLower(spcResourceID.SubscriptionID),
+		},
+		Status: api.ServiceProviderClusterStatus{
+			ManagementClusterResourceID: mcResourceID,
+		},
+	}
+}
+
+func TestCreateClusterScopedReadDesires_SyncOnce(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name                         string
+		resources                    []any
+		cachedServiceProviderCluster *api.ServiceProviderCluster
+		kubeApplierDesires           []any
+		wantErr                      bool
+		verifyDB                     func(t *testing.T, ctx context.Context, kaClient *databasetesting.MockKubeApplierDBClient)
+	}{
+		{
+			name: "creates HostedCluster and cluster-autoscaler ReadDesires",
+			resources: []any{
+				newTestCluster(),
+			},
+			cachedServiceProviderCluster: newTestSPC(readDesireTestManagementClusterResourceID),
+			verifyDB: func(t *testing.T, ctx context.Context, kaClient *databasetesting.MockKubeApplierDBClient) {
+				t.Helper()
+				crud, err := kaClient.ReadDesiresForCluster(readDesireTestSubscriptionID, readDesireTestResourceGroupName, readDesireTestClusterName)
+				require.NoError(t, err)
+
+				hostedClusterRD, err := crud.Get(ctx, readDesireNameReadonlyHostedCluster)
+				require.NoError(t, err)
+				assert.Equal(t, hostedClusterTarget(readDesireTestEnvIdentifier, "abc123", readDesireTestDomainPrefix), hostedClusterRD.Spec.TargetItem)
+
+				autoscalerRD, err := crud.Get(ctx, maestrohelpers.ReadDesireNameReadonlyHypershiftControlPlaneComponentClusterAutoscaler)
+				require.NoError(t, err)
+				assert.Equal(t, clusterAutoscalerTarget(readDesireTestEnvIdentifier, "abc123", readDesireTestDomainPrefix), autoscalerRD.Spec.TargetItem)
+				assert.Equal(t, "controlplanecomponents", autoscalerRD.Spec.TargetItem.Resource)
+				assert.Equal(t, "cluster-autoscaler", autoscalerRD.Spec.TargetItem.Name)
+				assert.Equal(t, hostedControlPlaneNamespace(readDesireTestEnvIdentifier, "abc123", readDesireTestDomainPrefix), autoscalerRD.Spec.TargetItem.Namespace)
+				assert.Equal(t, hsv1beta1.SchemeGroupVersion.Group, autoscalerRD.Spec.TargetItem.Group)
+				assert.Equal(t, hsv1beta1.SchemeGroupVersion.Version, autoscalerRD.Spec.TargetItem.Version)
+			},
+		},
+		{
+			name: "creates cluster-autoscaler ReadDesire even when cluster version is below 4.20",
+			resources: []any{
+				newTestCluster(func(c *api.HCPOpenShiftCluster) {
+					c.CustomerProperties.Version.ID = "4.19.0"
+				}),
+			},
+			cachedServiceProviderCluster: newTestSPC(readDesireTestManagementClusterResourceID),
+			verifyDB: func(t *testing.T, ctx context.Context, kaClient *databasetesting.MockKubeApplierDBClient) {
+				t.Helper()
+				crud, err := kaClient.ReadDesiresForCluster(readDesireTestSubscriptionID, readDesireTestResourceGroupName, readDesireTestClusterName)
+				require.NoError(t, err)
+
+				_, err = crud.Get(ctx, readDesireNameReadonlyHostedCluster)
+				require.NoError(t, err)
+
+				autoscalerRD, err := crud.Get(ctx, maestrohelpers.ReadDesireNameReadonlyHypershiftControlPlaneComponentClusterAutoscaler)
+				require.NoError(t, err)
+				assert.Equal(t, clusterAutoscalerTarget(readDesireTestEnvIdentifier, "abc123", readDesireTestDomainPrefix), autoscalerRD.Spec.TargetItem)
+			},
+		},
+		{
+			name: "skips when domain prefix is not yet synced",
+			resources: []any{
+				newTestCluster(func(c *api.HCPOpenShiftCluster) {
+					c.CustomerProperties.DNS.BaseDomainPrefix = ""
+				}),
+			},
+			cachedServiceProviderCluster: newTestSPC(readDesireTestManagementClusterResourceID),
+			verifyDB: func(t *testing.T, ctx context.Context, kaClient *databasetesting.MockKubeApplierDBClient) {
+				t.Helper()
+				crud, err := kaClient.ReadDesiresForCluster(readDesireTestSubscriptionID, readDesireTestResourceGroupName, readDesireTestClusterName)
+				require.NoError(t, err)
+				_, err = crud.Get(ctx, readDesireNameReadonlyHostedCluster)
+				require.Error(t, err)
+			},
+		},
+		{
+			name: "skips when management cluster is not placed",
+			resources: []any{
+				newTestCluster(),
+			},
+			verifyDB: func(t *testing.T, ctx context.Context, kaClient *databasetesting.MockKubeApplierDBClient) {
+				t.Helper()
+				crud, err := kaClient.ReadDesiresForCluster(readDesireTestSubscriptionID, readDesireTestResourceGroupName, readDesireTestClusterName)
+				require.NoError(t, err)
+				_, err = crud.Get(ctx, readDesireNameReadonlyHostedCluster)
+				require.Error(t, err)
+			},
+		},
+		{
+			name: "replaces cluster-autoscaler ReadDesire when target namespace changes",
+			resources: []any{
+				newTestCluster(),
+			},
+			cachedServiceProviderCluster: newTestSPC(readDesireTestManagementClusterResourceID),
+			kubeApplierDesires: []any{
+				buildReadDesire(
+					kubeapplier.ToClusterScopedReadDesireResourceIDString(
+						readDesireTestSubscriptionID, readDesireTestResourceGroupName, readDesireTestClusterName, maestrohelpers.ReadDesireNameReadonlyHypershiftControlPlaneComponentClusterAutoscaler),
+					readDesireTestManagementClusterResourceID,
+					clusterAutoscalerTarget(readDesireTestEnvIdentifier, "abc123", "old-prefix"),
+				),
+			},
+			verifyDB: func(t *testing.T, ctx context.Context, kaClient *databasetesting.MockKubeApplierDBClient) {
+				t.Helper()
+				crud, err := kaClient.ReadDesiresForCluster(readDesireTestSubscriptionID, readDesireTestResourceGroupName, readDesireTestClusterName)
+				require.NoError(t, err)
+				autoscalerRD, err := crud.Get(ctx, maestrohelpers.ReadDesireNameReadonlyHypershiftControlPlaneComponentClusterAutoscaler)
+				require.NoError(t, err)
+				assert.Equal(t, hostedControlPlaneNamespace(readDesireTestEnvIdentifier, "abc123", readDesireTestDomainPrefix), autoscalerRD.Spec.TargetItem.Namespace)
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			ctx := utils.ContextWithLogger(context.Background(), testr.New(t))
+
+			mockResourcesDBClient, err := databasetesting.NewMockResourcesDBClientWithResources(ctx, tt.resources)
+			require.NoError(t, err)
+
+			mockKubeApplierDBClients := databasetesting.NewMockKubeApplierDBClients()
+			mockKubeApplierClient, err := databasetesting.NewMockKubeApplierDBClientWithResources(ctx, tt.kubeApplierDesires)
+			require.NoError(t, err)
+			mockKubeApplierDBClients.Register(readDesireTestManagementClusterResourceID, mockKubeApplierClient)
+
+			serviceProviderClusterListerStub := &listertesting.SliceServiceProviderClusterLister{}
+			if tt.cachedServiceProviderCluster != nil {
+				serviceProviderClusterListerStub.ServiceProviderClusters = []*api.ServiceProviderCluster{tt.cachedServiceProviderCluster}
+			}
+
+			syncer := &createClusterScopedReadDesiresSyncer{
+				resourcesDBClient:                   mockResourcesDBClient,
+				kubeApplierDBClients:                mockKubeApplierDBClients,
+				serviceProviderClusterLister:        serviceProviderClusterListerStub,
+				hostedClusterNamespaceEnvIdentifier: readDesireTestEnvIdentifier,
+			}
+
+			err = syncer.SyncOnce(ctx, readDesireTestKey())
+			if tt.wantErr {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			if tt.verifyDB != nil {
+				tt.verifyDB(t, ctx, mockKubeApplierClient)
+			}
+		})
+	}
+}
--- a/backend/pkg/controllers/create_nodepool_scoped_read_desires_controller.go
+++ b/backend/pkg/controllers/create_nodepool_scoped_read_desires_controller.go
@@ -147,7 +147,7 @@ func (c *createNodePoolScopedReadDesiresSyncer) SyncOnce(ctx context.Context, ke
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("get ReadDesire CRUD: %w", err))
 	}
-	existing, err := getExistingReadDesire(ctx, crud, readDesireNameReadonlyNodePool)
+	existing, err := getExistingReadDesire(ctx, crud, desired.ResourceID.Name)
 	if err != nil {
 		return err
 	}
--- a/backend/pkg/controllers/cs_maestro_utils.go
+++ b/backend/pkg/controllers/cs_maestro_utils.go
@@ -23,3 +23,10 @@ import "fmt"
 func hostedClusterNamespace(envIdentifier, csClusterID string) string {
 	return fmt.Sprintf("ocm-%s-%s", envIdentifier, csClusterID)
 }
+
+// hostedControlPlaneNamespace returns the management-cluster namespace that hosts a
+// given HCP's control plane workloads (including ControlPlaneComponent CRs).
+// Hypershift names it "ocm-<envIdentifier>-<csClusterID>-<csClusterDomainPrefix>".
+func hostedControlPlaneNamespace(envIdentifier, csClusterID, csClusterDomainPrefix string) string {
+	return fmt.Sprintf("ocm-%s-%s-%s", envIdentifier, csClusterID, csClusterDomainPrefix)
+}
--- a/backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
@@ -220,6 +220,11 @@ func (c *operationClusterUpdate) determineOperationState(ctx context.Context, op
 	} else {
 		operationStates = append(operationStates, operationState.withSource("hypershiftHostedCluster"))
 	}
+	if operationState, autoscalerErr := c.hypershiftControlPlaneClusterAutoscalerState(ctx, existingCluster, existingServiceProviderCluster); autoscalerErr != nil {
+		errs = append(errs, utils.TrackError(autoscalerErr))
+	} else {
+		operationStates = append(operationStates, operationState.withSource("hypershiftControlPlaneClusterAutoscaler"))
+	}
 
 	if err := errors.Join(errs...); err != nil {
 		return nil, err
@@ -240,8 +245,8 @@ func (c *operationClusterUpdate) determineOperationState(ctx context.Context, op
 	return picked, nil
 }
 
-func (c *operationClusterUpdate) desiredVersionResolutionOperationState(ctx context.Context, operation *api.Operation, existingCluster *api.HCPOpenShiftCluster, existingServiceProviderCluster *api.ServiceProviderCluster) (*operationState, error) {
-	resultingDesiredVersion := existingServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion
+func (c *operationClusterUpdate) desiredVersionResolutionOperationState(ctx context.Context, operation *api.Operation, existingCluster *api.HCPOpenShiftCluster, spc *api.ServiceProviderCluster) (*operationState, error) {
+	resultingDesiredVersion := spc.Spec.ControlPlaneVersion.DesiredVersion
 	if resultingDesiredVersion == nil {
 		return nil, utils.TrackError(fmt.Errorf("service provider cluster has no desired version"))
 	}
--- a/backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
@@ -20,6 +20,11 @@ import (
 	"slices"
 	"time"
 
+	"github.com/blang/semver/v4"
+
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
 	"github.com/openshift/hypershift/api/hypershift/v1beta1"
 
@@ -28,6 +33,7 @@ import (
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/ocm"
 	"github.com/Azure/ARO-HCP/internal/utils"
+	"github.com/Azure/ARO-HCP/internal/utils/apihelpers"
 )
 
 // Cluster update operation state calculation for the cluster update operation controller.
@@ -476,4 +482,73 @@ func (c *operationClusterUpdate) clusterServiceClusterNodeDrainTimeoutSpecMatche
 		return false, fmt.Sprintf("Cluster Service nodeDrainGracePeriod is %d minutes, want %d", got, desired)
 	}
 	return true, ""
+}
+
+// hypershiftControlPlaneClusterAutoscalerState gates on the
+// cluster-autoscaler ControlPlaneComponent status (Available + RolloutComplete)
+// when the active control plane is 4.20+. HostedCluster autoscaling Spec matching
+// is owned by hypershiftHostedClusterOperationState.
+func (c *operationClusterUpdate) hypershiftControlPlaneClusterAutoscalerState(ctx context.Context, existingCluster *api.HCPOpenShiftCluster, spc *api.ServiceProviderCluster) (*operationState, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(spc.Status.ControlPlaneVersion.ActiveVersions)
+	if lowest == nil {
+		return newOperationState(arm.ProvisioningStateUpdating, "control plane active versions not yet reported"), nil
+	}
+	// Compare major.minor only so pre-release builds (e.g. nightlies like
+	// 4.20.0-0.nightly-...) still satisfy the 4.20+ autoscaler gate.
+	lowestMajorMinor := semver.Version{Major: lowest.Major, Minor: lowest.Minor}
+	if !lowestMajorMinor.GTE(semver.Version{Major: 4, Minor: 20}) {
+		msg := fmt.Sprintf(
+			`lowest active control plane version %q does not support ControlPlaneComponent cluster-autoscaler (requires 4.20+)`,
+			lowest.String(),
+		)
+		return newOperationState(arm.ProvisioningStateSucceeded, msg), nil
+	}
+
+	controlPlaneComponent, err := maestrohelpers.GetCachedControlPlaneClusterAutoscalerForCluster(
+		ctx, c.readDesireLister,
+		existingCluster.ID.SubscriptionID, existingCluster.ID.ResourceGroupName, existingCluster.ID.Name,
+	)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+	if controlPlaneComponent == nil {
+		return newOperationState(arm.ProvisioningStateUpdating, "cluster autoscaler state not cached yet"), nil
+	}
+	if !c.isControlPlaneClusterAutoscalerReady(controlPlaneComponent) {
+		message := c.controlPlaneClusterAutoscalerNotReadyMessage(controlPlaneComponent)
+		logger.Info("cluster autoscaler ControlPlaneComponent is not ready", "message", message)
+		return newOperationState(arm.ProvisioningStateUpdating, message), nil
+	}
+	return newOperationState(arm.ProvisioningStateSucceeded, ""), nil
+}
+
+func (c *operationClusterUpdate) isControlPlaneClusterAutoscalerReady(controlPlaneComponent *v1beta1.ControlPlaneComponent) bool {
+	return apimeta.IsStatusConditionTrue(controlPlaneComponent.Status.Conditions, string(v1beta1.ControlPlaneComponentAvailable)) &&
+		apimeta.IsStatusConditionTrue(controlPlaneComponent.Status.Conditions, string(v1beta1.ControlPlaneComponentRolloutComplete))
+}
+
+const (
+	clusterAutoscalerNotAvailableMsg       = "cluster autoscaler not available"
+	clusterAutoscalerRolloutNotCompleteMsg = "cluster autoscaler rollout not complete"
+	clusterAutoscalerNotReadyMsg           = "cluster autoscaler not ready"
+)
+
+func (c *operationClusterUpdate) controlPlaneClusterAutoscalerNotReadyMessage(controlPlaneComponent *v1beta1.ControlPlaneComponent) string {
+	available := apimeta.FindStatusCondition(controlPlaneComponent.Status.Conditions, string(v1beta1.ControlPlaneComponentAvailable))
+	rollout := apimeta.FindStatusCondition(controlPlaneComponent.Status.Conditions, string(v1beta1.ControlPlaneComponentRolloutComplete))
+	if available == nil || available.Status != metav1.ConditionTrue {
+		if available != nil && len(available.Message) > 0 {
+			return fmt.Sprintf("%s: %s: %s", clusterAutoscalerNotAvailableMsg, available.Reason, available.Message)
+		}
+		return clusterAutoscalerNotAvailableMsg
+	}
+	if rollout == nil || rollout.Status != metav1.ConditionTrue {
+		if rollout != nil && len(rollout.Message) > 0 {
+			return fmt.Sprintf("%s: %s: %s", clusterAutoscalerRolloutNotCompleteMsg, rollout.Reason, rollout.Message)
+		}
+		return clusterAutoscalerRolloutNotCompleteMsg
+	}
+	return clusterAutoscalerNotReadyMsg
 }
--- a/backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation_test.go
+++ b/backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation_test.go
@@ -16,18 +16,26 @@ package operationcontrollers
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
 	"testing"
 
+	"github.com/blang/semver/v4"
 	"github.com/go-logr/logr/testr"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/ptr"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 
 	arohcpv1alpha1 "github.com/openshift-online/ocm-sdk-go/arohcp/v1alpha1"
 	"github.com/openshift/hypershift/api/hypershift/v1beta1"
 
+	"github.com/Azure/ARO-HCP/backend/pkg/maestrohelpers"
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/api/arm"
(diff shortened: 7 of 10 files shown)


backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -155,6 +161,13 @@ func (c *operationClusterUpdate) determineOperationState(ctx context.Context, op
 	} else {
 		operationStates = append(operationStates, operationState)
 	}
+	if operation.AutoscalingUpdate {
```
machi1990: Why does this matter?

frontend/pkg/frontend/cluster.go
```diff
@@ -684,6 +685,10 @@ func (f *Frontend) updateHCPClusterInCosmos(ctx context.Context, writer http.Res
 	}
 
 	transaction := f.resourcesDBClient.NewTransaction(oldInternalCluster.ID.SubscriptionID)
+	autoscalingUpdate := !equality.Semantic.DeepEqual(
+		oldInternalCluster.CustomerProperties.Autoscaling,
+		newInternalCluster.CustomerProperties.Autoscaling,
+	)
```
machi1990: I don't think this is needed

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -155,6 +161,13 @@ func (c *operationClusterUpdate) determineOperationState(ctx context.Context, op
...
+		if operationState, autoscalerErr := c.clusterAutoscalerOperationState(ctx, operation); autoscalerErr != nil {
```
machi1990: I believe this is only supported for clusters that are 4.20+, maybe we can gate it

internal/api/types_operation.go
```diff
@@ -92,6 +92,11 @@ type Operation struct {
 	// This will be removed once all clusters whose deletion was triggered before the new approach is fully rolled out have been
 	// fully deleted in all ARO-HCP permanent environments, for all regions.
 	UsesNewClusterDeletionApproach bool `json:"usesNewClusterDeletionApproach"`
+
+	// AutoscalingUpdate is true for cluster update operations where the customer changed
+	// CustomerProperties.Autoscaling. The backend uses this to gate operation completion on
+	// cluster-autoscaler ControlPlaneComponent readiness on the management cluster.
+	AutoscalingUpdate bool `json:"autoscalingUpdate,omitempty"`
```
machi1990: Let's drop this, we should always check that the autoscaler is ready in an update operation

internal/api/types_serviceprovider_cluster.go
```diff
@@ -299,4 +299,8 @@ const (
 	// MaestroBundleInternalNameReadonlyHypershiftHostedCluster is the internal name of the Maestro Bundle that represents
 	// the Cluster's Hypershift's HostedCluster K8s resource.
 	MaestroBundleInternalNameReadonlyHypershiftHostedCluster MaestroBundleInternalName = "readonlyHypershiftHostedCluster"
+
+	// MaestroBundleInternalNameReadonlyControlPlaneClusterAutoscaler is the internal name of the ReadDesire that mirrors
+	// the cluster-autoscaler ControlPlaneComponent on the management cluster control plane namespace.
+	MaestroBundleInternalNameReadonlyControlPlaneClusterAutoscaler MaestroBundleInternalName = "readonlyControlPlaneClusterAutoscaler"
```
machi1990: poorly named; it doesn't rely on maestro anymore.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -205,6 +202,20 @@ func hostedClusterTarget(envIdentifier, csClusterID, csClusterDomainPrefix strin
 	}
 }
 
+const clusterAutoscalerControlPlaneComponentName = "cluster-autoscaler"
```
machi1990: inline

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +271,48 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
...
+	}
+	if readDesire.Status.KubeContent == nil || len(readDesire.Status.KubeContent.Raw) == 0 {
+		return newOperationState(arm.ProvisioningStateUpdating, "cluster autoscaler state not cached yet"), nil
+	}
+
+	cpc, err := maestrohelpers.GetCachedControlPlaneClusterAutoscalerForCluster(
+		ctx, c.readDesireLister,
+		operation.ExternalID.SubscriptionID,
+		operation.ExternalID.ResourceGroupName,
+		operation.ExternalID.Name,
```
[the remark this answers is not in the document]
machi1990: This comment makes sense

backend/pkg/maestrohelpers/controlplanecomponents.go
```diff
@@ -0,0 +1,93 @@
...
+	if database.IsNotFoundError(err) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to get ReadDesire for cluster-autoscaler ControlPlaneComponent: %w", err))
+	}
+	if readDesire.Status.KubeContent == nil || len(readDesire.Status.KubeContent.Raw) == 0 {
+		return nil, nil
+	}
+	cpc := &v1beta1.ControlPlaneComponent{}
```
machi1990: Can we name this `controlPlaneComponent` ? same throughout

backend/pkg/controllers/cs_maestro_utils.go
```diff
@@ -23,3 +23,10 @@ import "fmt"
 func hostedClusterNamespace(envIdentifier, csClusterID string) string {
 	return fmt.Sprintf("ocm-%s-%s", envIdentifier, csClusterID)
 }
+
+// controlPlaneNamespace returns the management-cluster namespace that hosts a
+// given HCP's control plane workloads (including ControlPlaneComponent CRs).
+// Hypershift names it "ocm-<envIdentifier>-<csClusterID>-<csClusterDomainPrefix>".
+func controlPlaneNamespace(envIdentifier, csClusterID, csClusterDomainPrefix string) string {
```
machi1990: hostedControlPlaneNamespace

machi1990: or controlPlaneResourcesNamespace

internal/api/types_operation.go
miguelsorianod: +1

backend/pkg/maestrohelpers/controlplanecomponents.go
```diff
@@ -0,0 +1,93 @@
...
+	if cpc == nil {
+		return false
+	}
+	return meta.IsStatusConditionTrue(cpc.Status.Conditions, string(v1beta1.ControlPlaneComponentAvailable)) &&
+		meta.IsStatusConditionTrue(cpc.Status.Conditions, string(v1beta1.ControlPlaneComponentRolloutComplete))
+}
+
+// ControlPlaneClusterAutoscalerNotReadyMessage returns a human-readable reason when
+// the cluster-autoscaler ControlPlaneComponent is not ready.
+func ControlPlaneClusterAutoscalerNotReadyMessage(cpc *v1beta1.ControlPlaneComponent) string {
```
miguelsorianod: This doesn't belong here, it belongs to the operation files

internal/api/types_serviceprovider_cluster.go
miguelsorianod: readonlyControlPlaneClusterAutoscaler -> readonlyHypershiftControlPlaneComponentClusterAutoscaler

it's a hypershift api resource so let's be explicit in the names

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -25,16 +25,19 @@ import (
 
 	"github.com/blang/semver/v4"
 
-	apimeta "k8s.io/apimachinery/pkg/api/meta"
```
miguelsorianod, line 28: I am not sure why the package aliasing has been removed

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +271,48 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
 	}
 	return newOperationState(newOperationStatus, msg), nil
 }
+
+func (c *operationClusterUpdate) clusterAutoscalerOperationState(ctx context.Context, operation *api.Operation) (*operationState, error) {
```
miguelsorianod: Do not pass the operation. Pass the cluster resource id. In that way we decouple this method from the operation type.

backend/pkg/maestrohelpers/controlplanecomponents.go
```diff
@@ -0,0 +1,93 @@
...
+	cpc := &v1beta1.ControlPlaneComponent{}
+	if err := json.Unmarshal(readDesire.Status.KubeContent.Raw, cpc); err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to unmarshal ControlPlaneComponent from ReadDesire kubeContent: %w", err))
+	}
+	return cpc, nil
+}
+
+// IsControlPlaneClusterAutoscalerReady reports whether the cluster-autoscaler
+// ControlPlaneComponent has completed rollout and is available.
+func IsControlPlaneClusterAutoscalerReady(cpc *v1beta1.ControlPlaneComponent) bool {
```
miguelsorianod: This seems to belong to the operation files itself

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -248,3 +259,31 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
 	}
 	return existing.Spec.TargetItem != desired.Spec.TargetItem
 }
+
+// ensureReadDesire creates or updates a ReadDesire when the desired spec differs from cosmos.
+func ensureReadDesire(
```
miguelsorianod: This scope doesn't seem correct.

Shouldn't this be done in createClusterScopedReadDesiresSyncer, and the operation controllers just reference it?

miguelsorianod: also consider what happens with versions

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: I will gate at createClusterScopedReadDesiresSyncer so that it won't create the autoscaler ReadDesire below 4.20, and at clusterAutoscalerOperationState to return success without waiting when the cluster version is below 4.20.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: unintentionally removed during the cherry-pick from the upstream main branch to the fork branch. I'll restore

internal/api/types_serviceprovider_cluster.go
shitaljante: MaestroBundleInternalName was kept for consistency with the existing HostedCluster and NodePool constants.
The documentation already indicates that this is a ReadDesire :

// MaestroBundleInternalNameReadonlyControlPlaneClusterAutoscaler is the internal name of the ReadDesire.

Let me know if you think we should remove the Maestro prefix from this name.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -158,23 +151,25 @@ func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key
...
 		return err
 	}
-	if !readDesireNeedsWork(existing, desired) {
-		return nil
-	}
-	if existing == nil {
-		if _, err := crud.Create(ctx, desired, nil); err != nil {
-			return utils.TrackError(fmt.Errorf("create ReadDesire: %w", err))
+
+	if api.ClusterVersionSupportsHypershiftControlPlaneComponent(existingCluster.CustomerProperties.Version.ID) {
```
miguelsorianod: This will make us need to duplicate code each time we add a new one. This is not going to be the last one we add. Lets' refactor this so new ones can be added easily.

If one fails, accumulate errors instead of failing immediately. This will allow to do a best-effort instead of failing when one fails.

Take into account different ones might have different preconditions

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -248,3 +259,31 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
...
+	}
+	if existing == nil {
+		if _, err := crud.Create(ctx, desired, nil); err != nil {
+			return utils.TrackError(fmt.Errorf("create ReadDesire %q: %w", name, err))
+		}
+		return nil
+	}
+	replacement := existing.DeepCopy()
+	replacement.Spec = *desired.Spec.DeepCopy()
+	if _, err := crud.Replace(ctx, replacement, nil); err != nil {
```
miguelsorianod: We want to treat failures due to outdated DB contents as non errors:

change to:
_, err := replace
if error is precondition failed return without error
if error is not nil return error
...

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -248,3 +259,31 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
 	}
 	return existing.Spec.TargetItem != desired.Spec.TargetItem
 }
+
+// ensureReadDesire creates or updates a ReadDesire when the desired spec differs from cosmos.
+func (c *createClusterScopedReadDesiresSyncer) ensureReadDesire(
+	ctx context.Context,
```
miguelsorianod: No need to newline here. We accept long lines in the project

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -155,6 +162,16 @@ func (c *operationClusterUpdate) determineOperationState(ctx context.Context, op
 	} else {
 		operationStates = append(operationStates, operationState)
 	}
+	if operationState, autoscalerErr := c.clusterAutoscalerOperationState(
+		ctx,
```
miguelsorianod: No need to newline here. We accept long line

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -155,6 +162,16 @@ func (c *operationClusterUpdate) determineOperationState(ctx context.Context, op
...
+		operation.ExternalID.SubscriptionID,
```
miguelsorianod: No need to separate the parameters. Just pass the resource id.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +275,64 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
...
 	return newOperationState(newOperationStatus, msg), nil
 }
+
+func (c *operationClusterUpdate) clusterAutoscalerOperationState(
+	ctx context.Context,
+	subscriptionID, resourceGroupName, clusterName string,
+) (*operationState, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	existingCluster, err := c.resourcesDBClient.HCPClusters(subscriptionID, resourceGroupName).Get(ctx, clusterName)
```
miguelsorianod: Don't retrieve this again. In desiredVersionResolutionOperationState it is also being retrieved. Move retrieval one level up and pass down the shared retrieved cluster

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +275,64 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
...
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+	if !api.ClusterVersionSupportsHypershiftControlPlaneComponent(existingCluster.CustomerProperties.Version.ID) {
```
miguelsorianod: This creates the operationstate and reports is as succeeded. A message would be useful to debug

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +275,64 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
...
+		return newOperationState(arm.ProvisioningStateUpdating, message), nil
+	}
+	return newOperationState(arm.ProvisioningStateSucceeded, ""), nil
+}
+
+func isControlPlaneClusterAutoscalerReady(controlPlaneComponent *v1beta1.ControlPlaneComponent) bool {
+	if controlPlaneComponent == nil {
+		return false
+	}
+	return apimeta.IsStatusConditionTrue(controlPlaneComponent.Status.Conditions, string(v1beta1.ControlPlaneComponentAvailable)) &&
```
miguelsorianod: How do we know that the conditions here are in that state because of the desired state being applied? it seems to me this doesn't necessarily guarantee the desired state has been applied

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +275,64 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
...
+		apimeta.IsStatusConditionTrue(controlPlaneComponent.Status.Conditions, string(v1beta1.ControlPlaneComponentRolloutComplete))
+}
+
+func controlPlaneClusterAutoscalerNotReadyMessage(controlPlaneComponent *v1beta1.ControlPlaneComponent) string {
+	if controlPlaneComponent == nil {
```
miguelsorianod: Why is this check needed? isn't this called when we know it's not nil?

backend/pkg/controllers/operationcontrollers/operation_cluster_update_test.go
```diff
@@ -54,6 +60,7 @@ func TestOperationClusterUpdate_SynchronizeOperation(t *testing.T) {
 		serviceProviderClusterStatusConditions         []metav1.Condition
 		controlPlaneDesiredVersionControllerConditions []metav1.Condition
 		seedMismatchFirstSeenAt                        time.Time
+		readDesires                                    []*kubeapplier.ReadDesire
```
miguelsorianod: Change to be a parameter of the specific of the resource instead of array. That way we will be more explicit instead of in an array of N elements which will be harder to follow.

If not nil, the test body adds it to the internal readdesirelister

backend/pkg/controllers/operationcontrollers/operation_cluster_update_test.go
```diff
@@ -252,11 +259,16 @@ func TestOperationClusterUpdate_SynchronizeOperation(t *testing.T) {
 				Return(clusterStatus, nil)
 
 			fakeClock := clocktesting.NewFakeClock(testClockNow)
+			readDesires := tt.readDesires
+			if readDesires == nil && api.ClusterVersionSupportsHypershiftControlPlaneComponent(tt.customerVersionID) {
```
miguelsorianod: Don't complicate test body logic and make assumptions.
This will evolve poorly when there are needs to create a different shapre of the control plane autoscaler readdesire for the tests

Follow https://github.com/Azure/ARO-HCP/pull/5931/changes#r3537885390. Pass explicitly your information as test args. Reuse helpers to instantiate the same shape in the test args. Don't instantiate it in the args if it's not needed.

internal/api/hypershift_version.go
```diff
@@ -12,29 +12,16 @@
...
-		if err != nil {
-			return nil, err
-		}
-		cmd.AddCommand(c)
+// ClusterVersionSupportsHypershiftControlPlaneComponent reports whether the
+// cluster version has the cluster-autoscaler ControlPlaneComponent (4.20+).
+func ClusterVersionSupportsHypershiftControlPlaneComponent(versionID string) bool {
+	v, err := semver.ParseTolerant(versionID)
+	if err != nil {
+		return false
```
miguelsorianod: Returning false instead of an error when we can't parse doesn't seem correct.

internal/api/hypershift_version.go
```diff
@@ -12,29 +12,16 @@
...
-	}
-	for _, newCmd := range commands {
-		c, err := newCmd()
-		if err != nil {
-			return nil, err
-		}
-		cmd.AddCommand(c)
+// ClusterVersionSupportsHypershiftControlPlaneComponent reports whether the
+// cluster version has the cluster-autoscaler ControlPlaneComponent (4.20+).
+func ClusterVersionSupportsHypershiftControlPlaneComponent(versionID string) bool {
```
miguelsorianod: cc @machi1990 is this the way we are approaching checks of this kind?

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: We need Version.ID along with Resource ID , changed the signature to take existingCluster *api.HCPOpenShiftCluster instead of separate params

backend/pkg/controllers/operationcontrollers/operation_cluster_update_test.go
shitaljante: Removed the implicit default that created a ready autoscaler ReadDesire for 4.20+ cases. Each test case that needs the autoscaler gate to pass now sets explicitly with helper.

internal/api/hypershift_version.go
shitaljante: This is aligned with the existing pattern in desiredVersionResolutionOperationState, which uses semver.ParseTolerant and returns an error on unparseable versions

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: Hypershift sets ControlPlaneComponentRolloutComplete when rollout of the current spec generation completes, and ControlPlaneComponentAvailable when the component is serving.  This rollout completion criteria was provided by Hypershift team.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: Moved retrieval above

internal/api/hypershift_version.go
shitaljante: Updated

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
shitaljante: So should ClusterVersionSupportsHypershiftControlPlaneComponent be treated as the precondition for the cluster-autoscaler ReadDesire handler? Also should we have a generic mechanism for expressing supported versions?

miguelsorianod: I think we can have the cluster-autoscaler ReadDesire anyway. It will be useful at the least to view its state in the Cosmos DB

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
miguelsorianod: You don't know if that rollout corresponds to the current desired spec

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
miguelsorianod: reopening. Not addressed

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +271,65 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
 	}
 	return newOperationState(newOperationStatus, msg), nil
 }
+
+func (c *operationClusterUpdate) clusterAutoscalerOperationState(ctx context.Context, existingCluster *api.HCPOpenShiftCluster) (*operationState, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	supported, err := api.ClusterVersionSupportsHypershiftControlPlaneComponent(existingCluster.CustomerProperties.Version.ID)
```
machi1990: This should use the the ServiceProviderCluster actual lowest versions and not the desired version.

machi1990: Why is `api.ClusterVersionSupportsHypershiftControlPlaneComponent` in the api package? I'd have kept it here as a private method

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -158,23 +151,29 @@ func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key
...
+		mcResourceID,
+		hostedClusterTarget(c.hostedClusterNamespaceEnvIdentifier, csClusterID, csClusterDomainPrefix),
+	)
+	if err := c.ensureReadDesire(ctx, crud, readDesireNameReadonlyHostedCluster, hostedClusterDesired); err != nil {
 		return err
 	}
-	if !readDesireNeedsWork(existing, desired) {
-		return nil
+
+	supportsHypershiftControlPlaneComponent, err := api.ClusterVersionSupportsHypershiftControlPlaneComponent(existingCluster.CustomerProperties.Version.ID)
```
machi1990: Maybe for the case of creation we can always create it;
So that when a 4.19 cluster gets upgrade to 4.20, it'll have it automatically

internal/api/hypershift_version.go
```diff
@@ -12,29 +12,20 @@
...
+	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
-func NewCommand() (*cobra.Command, error) {
-	cmd := &cobra.Command{
-		Use:   "istio",
-		Short: "Manage AKS-managed Istio service mesh upgrades",
+// ClusterVersionSupportsHypershiftControlPlaneComponent reports whether the
+// cluster version has the cluster-autoscaler ControlPlaneComponent (4.20+).
+func ClusterVersionSupportsHypershiftControlPlaneComponent(versionID string) (bool, error) {
```
machi1990: I left a comment suggesting to move this to the only place I think it should be used

internal/api/hypershift_version.go
```diff
@@ -12,29 +12,20 @@
...
+	v, err := semver.ParseTolerant(versionID)
```
machi1990: See my comment about using the ServiceProviderCluster lowest active version; there are some utilies for it https://github.com/shitaljante/ARO-HCP/blob/d82abb31a31a66d9b86cf2e92f3706a213fd8459/internal/utils/apihelpers/version.go#L25

machi1990: That way you don't need to parse it here;

internal/api/hypershift_version.go
```diff
@@ -12,29 +12,20 @@
...
-	}
-	for _, newCmd := range commands {
-		c, err := newCmd()
-		if err != nil {
-			return nil, err
-		}
-		cmd.AddCommand(c)
-	}
-	return cmd, nil
+	return v.Major > 4 || (v.Major == 4 && v.Minor >= 20), nil
```
machi1990: This can be 
```
minVersionWithAutoscalerComponent := semver.Version{Major: 4, Minor: 20}
return v.GTE(minVersionWithAutoscalerComponent)
```

internal/api/hypershift_version.go
machi1990: This can be improved a bit; I left some suggestions.
The biggest one is to use the service provider lowest version.
And the second biggest is to use it only during read of the desired and that's in the operation update and make it a private function or it can be inlined logic the place we need it

internal/api/types_serviceprovider_cluster.go
machi1990: Let's drop the maestro prefix.
I like this https://github.com/Azure/ARO-HCP/pull/5931#discussion_r3527925652 naming suggestion

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
machi1990: No, only gate here during read. Let's have the ReadDesire created; it'll be synced when a 4.19 cluster upgrades to 4.20. 

PS: I don't think we've 4.19 anymore but our code paths allows someone with an AFEC flag to still create them so allowing the readDesire to be craeted always will ensure that it's there from the get go and re-used during upgrade.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
miguelsorianod: +1

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
miguelsorianod: I agree. Also because of what I mention in https://github.com/Azure/ARO-HCP/pull/5931#discussion_r3550174831

internal/api/hypershift_version.go
shitaljante: Updated: the operation gate now loads ServiceProviderCluster and uses active CP version

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: Moved to a private method

internal/api/hypershift_version.go
shitaljante: Updated

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -31,19 +31,34 @@ import (
...
 )
 
+var minVersionWithClusterAutoscalerControlPlaneComponent = semver.Version{Major: 4, Minor: 20}
+
+func clusterSupportsClusterAutoscalerControlPlaneComponent(lowestActiveVersion *semver.Version) bool {
+	if lowestActiveVersion == nil {
+		return false
+	}
+	return lowestActiveVersion.GTE(minVersionWithClusterAutoscalerControlPlaneComponent)
+}
```
machi1990: This can be inlined to where it is used

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +281,71 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
 	}
 	return newOperationState(newOperationStatus, msg), nil
 }
+
+func (c *operationClusterUpdate) clusterAutoscalerOperationState(ctx context.Context, existingCluster *api.HCPOpenShiftCluster) (*operationState, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	spc, err := c.resourcesDBClient.ServiceProviderClusters(
```
machi1990: This method https://github.com/shitaljante/ARO-HCP/blob/65ec255feb1ec9ddf57bd3686d5ca2ef33bb6429/backend/pkg/controllers/operationcontrollers/operation_cluster_update.go#L205 also retrieves the ServiceProvider cluster, perhaps we can extract the retrieval to the caller and pass the retrieve value to both that method and the one you are adding here?

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +281,71 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
...
+		logger.Info("cluster autoscaler ControlPlaneComponent is not ready", "message", message)
+		return newOperationState(arm.ProvisioningStateUpdating, message), nil
+	}
+	return newOperationState(arm.ProvisioningStateSucceeded, ""), nil
+}
+
+func isControlPlaneClusterAutoscalerReady(controlPlaneComponent *v1beta1.ControlPlaneComponent) bool {
+	if controlPlaneComponent == nil {
+		return false
+	}
```
machi1990: The caller already checks for this

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: we have this method 
```
// GTE checks if v is greater than or equal to o.
func (v Version) GTE(o Version) bool {
	return (v.Compare(o) >= 0)
}
```
which can be used inline as 

> lowest.GTE(semver.Version{Major: 4, Minor: 20})

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +277,59 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
...
 }
+
+func (c *operationClusterUpdate) clusterAutoscalerOperationState(ctx context.Context, existingCluster *api.HCPOpenShiftCluster, spc *api.ServiceProviderCluster) (*operationState, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(spc.Status.ControlPlaneVersion.ActiveVersions)
+	if lowest == nil {
+		return newOperationState(arm.ProvisioningStateSucceeded, "control plane active versions not yet reported"), nil
+	}
+	if !lowest.GTE(semver.Version{Major: 4, Minor: 20}) {
```
machi1990: form the lowest, we need to extract the major.minor and only compare 

so that nightly versions can also succeed, other nightly 4.20 full sember will be considered as not greater than 4.20.

This needs to be 


```suggestion
    lowestMajorMinor := semver.Version{Major: lowest.Major, Minor: lowest.Minor}.
	if !lowestMajorMinor.GTE(semver.Version{Major: 4, Minor: 20}) {
```

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -258,3 +277,59 @@ func (c *operationClusterUpdate) clusterServiceUpdateOperationState(ctx context.
 	}
 	return newOperationState(newOperationStatus, msg), nil
 }
+
+func (c *operationClusterUpdate) clusterAutoscalerOperationState(ctx context.Context, existingCluster *api.HCPOpenShiftCluster, spc *api.ServiceProviderCluster) (*operationState, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	lowest, _ := apihelpers.FindLowestAndHighestClusterVersion(spc.Status.ControlPlaneVersion.ActiveVersions)
+	if lowest == nil {
+		return newOperationState(arm.ProvisioningStateSucceeded, "control plane active versions not yet reported"), nil
```
miguelsorianod: Should this be succeeded?

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
miguelsorianod: This comment is still pending

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -248,3 +271,28 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
 	}
 	return existing.Spec.TargetItem != desired.Spec.TargetItem
 }
+
+// ensureReadDesire creates or updates a ReadDesire when the desired spec differs from cosmos.
+func (c *createClusterScopedReadDesiresSyncer) ensureReadDesire(ctx context.Context, crud database.ResourceCRUD[kubeapplier.ReadDesire, *kubeapplier.ReadDesire], name string, desired *kubeapplier.ReadDesire) error {
```
miguelsorianod: It is redundant to pass `name` here. The own readdesire has the name

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -58,6 +59,11 @@ type createClusterScopedReadDesiresSyncer struct {
 	hostedClusterNamespaceEnvIdentifier string
 }
 
+type clusterScopedReadDesireEntry struct {
+	name    string
```
miguelsorianod: Is `name` needed? the own readdesire should contain the name

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -248,3 +271,28 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
...
+	}
+	if existing == nil {
+		if _, err := crud.Create(ctx, desired, nil); err != nil {
+			return utils.TrackError(fmt.Errorf("create ReadDesire %q: %w", name, err))
+		}
+		return nil
+	}
+	replacement := existing.DeepCopy()
+	replacement.Spec = *desired.Spec.DeepCopy()
+	if _, err := crud.Replace(ctx, replacement, nil); database.IsPreconditionFailedError(err) {
```
miguelsorianod: redundant and unnecessary if - else if nesting. Change to:

```
err = crud.Replace(ctx, replacement, nil)
if database.IsPreconditionFailedError(err) {
  return nil
}
if err != nil {
  return utils.TrackError(....)
}

return nil
```

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+func TestCreateClusterScopedReadDesires_SyncOnce(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name        string
+		resources   []any
+		kubeApplier []any
+		wantErr     bool
+		verify      func(t *testing.T, ctx context.Context, kaClient *databasetesting.MockKubeApplierDBClient)
```
miguelsorianod: rename to verifyDB to be consistent with other tests and to make explicit that the only type of checks within the function are DB checks

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+}
+
+func TestCreateClusterScopedReadDesires_SyncOnce(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name        string
+		resources   []any
+		kubeApplier []any
```
miguelsorianod: rename to kubeApplierDesires

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
```diff
@@ -0,0 +1,253 @@
...
+		wantErr     bool
+		verify      func(t *testing.T, ctx context.Context, kaClient *databasetesting.MockKubeApplierDBClient)
+	}{
+		{
+			name: "creates HostedCluster and cluster-autoscaler ReadDesires",
```
miguelsorianod, line 127: Is this design correct? here `resources` are being defined but it contains readdesires, which don't belong to that container

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
```diff
@@ -0,0 +1,253 @@
...
+				assert.Equal(t, "controlplanecomponents", autoscalerRD.Spec.TargetItem.Resource)
+				assert.Equal(t, "cluster-autoscaler", autoscalerRD.Spec.TargetItem.Name)
+				assert.Equal(t, hostedControlPlaneNamespace(readDesireTestEnvIdentifier, "abc123", readDesireTestDomainPrefix), autoscalerRD.Spec.TargetItem.Namespace)
+				assert.Equal(t, hsv1beta1.SchemeGroupVersion.Group, autoscalerRD.Spec.TargetItem.Group)
+				assert.Equal(t, hsv1beta1.SchemeGroupVersion.Version, autoscalerRD.Spec.TargetItem.Version)
+			},
+		},
+		{
+			name: "creates cluster-autoscaler ReadDesire even when cluster version is below 4.20",
+			resources: []any{
```
miguelsorianod, line 153: Same comment as https://github.com/Azure/ARO-HCP/pull/5931/changes#r3570545103

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
miguelsorianod: I've just seen that newReadDesireTestCluster and newReadDesireTestSPC are **not** readdesires. Why are we naming them that way?

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
miguelsorianod: I've just seen that newReadDesireTestCluster and newReadDesireTestSPC are not readdesires. Why are we naming them that way?

backend/pkg/controllers/operationcontrollers/operation_cluster_update_test.go
```diff
@@ -51,9 +57,11 @@ func TestOperationClusterUpdate_SynchronizeOperation(t *testing.T) {
 		name                                           string
 		clusterState                                   arohcpv1alpha1.ClusterState
 		customerVersionID                              string
+		activeVersions                                 []api.HCPClusterActiveVersion
 		serviceProviderClusterStatusConditions         []metav1.Condition
 		controlPlaneDesiredVersionControllerConditions []metav1.Condition
 		seedMismatchFirstSeenAt                        time.Time
+		controlPlaneClusterAutoscalerReadDesire        *kubeapplier.ReadDesire
```
miguelsorianod: call it cachedControlPlaneClusterAutoscalerReadDesire.

It is a cached view.

Also document that when set, wires a ReadDesiredLister containing this cached readdesire

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
machi1990: The chances of this branch being hit are zero but we want a different state here than succeeded

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
shitaljante: yes, renamed it to simply newTestCluster and newTestSPC

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: good catch, thank you

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: We could validate the HostedCluster resource to confirm that the update has completed. I noticed we already have similar logic here:
https://github.com/Azure/ARO-HCP/blob/3d168cf9d5f2b7fb9ff120138dea76ed52846e85/backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go#L77

Do you think it would make sense to move this check into clusterAutoscalerOperationState?

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
shitaljante: renamed

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
shitaljante: Updated

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -313,3 +321,79 @@ func (c *operationClusterUpdate) clusterServiceClusterStatusOperationState(ctx c
 
 	return newOperationState(newOperationStatus, msg), nil
 }
+
+func (c *operationClusterUpdate) clusterAutoscalerOperationState(ctx context.Context, existingCluster *api.HCPOpenShiftCluster, spc *api.ServiceProviderCluster) (*operationState, error) {
```
miguelsorianod: Move this to the backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go file as that's where we put the calculations now. Same for the corresponding unit tests of this function

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -313,3 +321,79 @@ func (c *operationClusterUpdate) clusterServiceClusterStatusOperationState(ctx c
...
+		existingCluster.ID.ResourceGroupName,
+		existingCluster.ID.Name,
+	)
+	if err != nil {
+		return nil, utils.TrackError(err)
+	}
+	if hostedCluster == nil {
+		return newOperationState(arm.ProvisioningStateUpdating, "Hypershift HostedCluster has not been observed yet"), nil
+	}
+	if matches, message := c.hypershiftHostedClusterAutoscalingSpecMatchesDesired(existingCluster.CustomerProperties.Autoscaling, &hostedCluster.Spec.Autoscaling); !matches {
```
miguelsorianod: Redundant, already calculated in the hostedcluster operation calculation since last rebase

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
miguelsorianod: If this only checks the autoscaling control plane component and only its status, rename the method appropriately

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
miguelsorianod: Since last rebase the HostedCluster autoscaler section is being checked already in hypershiftHostedClusterAutoscalingSpecMatchesDesired. Doublecheck that that method checks all we need and if so no further changes needed

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -56,6 +57,10 @@ type createClusterScopedReadDesiresSyncer struct {
 	hostedClusterNamespaceEnvIdentifier string
 }
 
+type clusterScopedReadDesireEntry struct {
```
miguelsorianod: Why a struct for this if it's only one attribute

backend/pkg/controllers/create_nodepool_scoped_read_desires_controller.go
```diff
@@ -141,7 +141,7 @@ func (c *createNodePoolScopedReadDesiresSyncer) SyncOnce(ctx context.Context, ke
 	if err != nil {
 		return utils.TrackError(fmt.Errorf("get ReadDesire CRUD: %w", err))
 	}
-	existing, err := getExistingReadDesire(ctx, crud, readDesireNameReadonlyNodePool)
+	existing, err := getExistingReadDesire(ctx, crud, desired)
```
miguelsorianod: No need to pass the whole desired when you only need the name

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
shitaljante: removed struct and directly initialised at assignment

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: removed duplicate function call

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
shitaljante: hypershiftHostedClusterAutoscalingSpecMatchesDesired coveres required checks , removed redundant function call

machi1990 (commented):
Mostly lgtm. I've not reviewed to see if @miguelsorianod comments are addressed.

Before I tag this, can you share the example kusto log dump of the new readdesired for the autoscaler component? Thanks I want be sure it is correctly retrieved on top of the passing e2e. @shitaljante

shitaljante:
I've attached the Kusto log dump for the new ReadDesire created for the autoscaler component in my personal dev environment to the PR description. 
Attaching the Kusto logs showing the ReadDesire created during the E2E test run here.
https://privatebin.corp.redhat.com/?c2dc32a2da65eefe#Ban2hTC92uGjDRVttNgJgb1Sm1fD6543oz9Fkvi9ex9z

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation_test.go
```diff
@@ -90,72 +98,7 @@ func TestHypershiftHostedClusterOperationState(t *testing.T) {
...
-			readDesires: []*kubeapplier.ReadDesire{
-				newHostedClusterReadDesire(t, &v1beta1.HostedCluster{
-					Spec: func() v1beta1.HostedClusterSpec {
-						spec := testClusterUpdateMatchingHostedClusterSpec()
-						spec.Autoscaling.MaxNodeProvisionTime = "15m"
-						return spec
-					}(),
-				}),
-			},
-			wantState: arm.ProvisioningStateSucceeded,
```
nimrodshn, line 158: Why are these removed?

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
```diff
@@ -434,3 +440,66 @@ func (c *operationClusterUpdate) clusterServiceClusterNodeDrainTimeoutSpecMatche
...
+		}
+		return "cluster autoscaler not available"
+	}
+	if rollout == nil || rollout.Status != metav1.ConditionTrue {
+		if rollout != nil && len(rollout.Message) > 0 {
+			return fmt.Sprintf("cluster autoscaler rollout not complete: %s: %s", rollout.Reason, rollout.Message)
+		}
+		return "cluster autoscaler rollout not complete"
+	}
+	return "cluster autoscaler not ready"
```
nimrodshn: Nit: mind declaring all the messages here as constants, for readability, reusability, etc.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -241,3 +261,31 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
...
+	if !readDesireNeedsWork(existing, desired) {
+		return nil
+	}
+	name := desired.ResourceID.Name
+	if existing == nil {
+		if _, err := crud.Create(ctx, desired, nil); err != nil {
+			return utils.TrackError(fmt.Errorf("create ReadDesire %q: %w", name, err))
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
nimrodshn: @shitaljante Worth considering in case of concurrent runs of the controller.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -182,6 +178,10 @@ func (c *createClusterScopedReadDesiresSyncer) SyncOnce(ctx context.Context, key
 // ManagementClusterContent document path stays stable across the migration.
 var readDesireNameReadonlyHostedCluster = strings.ToLower(string(api.MaestroBundleInternalNameReadonlyHypershiftHostedCluster))
 
+// readDesireNameReadonlyHypershiftControlPlaneComponentClusterAutoscaler is the well-known ReadDesire name
+// for the cluster-autoscaler ControlPlaneComponent mirror.
+var readDesireNameReadonlyHypershiftControlPlaneComponentClusterAutoscaler = strings.ToLower(string(api.ReadonlyHypershiftControlPlaneComponentClusterAutoscaler))
```
zgalor: Can we reuse `ReadDesireNameReadonlyHypershiftControlPlaneComponentClusterAutoscaler` here?

https://github.com/Azure/ARO-HCP/blob/52d405823c/backend/pkg/maestrohelpers/controlplanecomponents.go#L36

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation_test.go
shitaljante: Similar functionality is tested via new coverage TestHypershiftControlPlaneClusterAutoscalerOperationState , But I see that test case is complementary to above tests and not replacement , re-added above test cases.

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation_test.go
```diff
@@ -90,72 +98,7 @@ func TestHypershiftHostedClusterOperationState(t *testing.T) {
...
+			wantMessageSubstr: "hypershift HostedCluster autoscaling maxNodesTotal is 5, want 10",
 		},
 		{
 			name: "imageContentSources missing desired source returns Updating",
```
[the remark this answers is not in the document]
shitaljante: re-added

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
shitaljante: Added constants

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
shitaljante: Updated

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -241,3 +250,34 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
...
+	if err != nil {
+		return err
+	}
+	if !readDesireNeedsWork(existing, desired) {
+		return nil
+	}
+	name := desired.ResourceID.Name
+	if existing == nil {
+		if _, err := crud.Create(ctx, desired, nil); err != nil {
+			if database.IsConflictError(err) || database.IsPreconditionFailedError(err) {
```
miguelsorianod: database.IsPreconditionFailedError shouldn't be possible on create, as by definition you get that once the ETag changed, here we are creating a new resource

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -241,3 +250,34 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
...
+				return nil
+			}
+			return utils.TrackError(fmt.Errorf("create ReadDesire %q: %w", name, err))
```
miguelsorianod: create ReadDesire ... -> failed to create ReadDesire ...

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -241,3 +250,34 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
...
+	existing, err := getExistingReadDesire(ctx, crud, desired.ResourceID.Name)
+	if err != nil {
+		return err
+	}
+	if !readDesireNeedsWork(existing, desired) {
+		return nil
+	}
+	name := desired.ResourceID.Name
+	if existing == nil {
+		if _, err := crud.Create(ctx, desired, nil); err != nil {
```
miguelsorianod: Unnecessary nesting. Change to:

err := create(...)
if database.IsConflictError(err) {
  return nil
}
if err != nil {
  return whatevererror
}
return nil

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
```diff
@@ -241,3 +250,34 @@ func readDesireNeedsWork(existing, desired *kubeapplier.ReadDesire) bool {
...
+		return nil
+	}
+	replacement := existing.DeepCopy()
+	replacement.Spec = *desired.Spec.DeepCopy()
+	_, err = crud.Replace(ctx, replacement, nil)
+	if database.IsPreconditionFailedError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("replace ReadDesire %q: %w", name, err))
```
miguelsorianod: replace ReadDesire -> failed to replace ReadDesire

backend/pkg/maestrohelpers/controlplanecomponents.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to get ReadDesire for cluster-autoscaler ControlPlaneComponent: %w", err))
+	}
+	if !apimeta.IsStatusConditionTrue(readDesire.Status.Conditions, kubeapplier.ConditionTypeSuccessful) {
+		return nil, nil
+	}
+	if readDesire.Status.KubeContent == nil || len(readDesire.Status.KubeContent.Raw) == 0 {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
shitaljante: fixed as per suggestion, GetCachedHostedClusterForCluster and GetCachedNodePoolForNodePool has same logic

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
shitaljante: Updated

backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go
shitaljante: Updated all three scenarios with proper err message

backend/pkg/controllers/operationcontrollers/operation_cluster_update_test.go
miguelsorianod, line 97: We miss a test here that ensures that when the autoscaler is supported, and all the operation state calculations pass except from the autoscaler one we are in updating state.

We also seem to miss a test when autoscaler is supported, and all the operation state calculations pass, including the autoscaler one, where we want succeeded state

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
```diff
@@ -434,3 +440,72 @@ func (c *operationClusterUpdate) clusterServiceClusterNodeDrainTimeoutSpecMatche
...
+		apimeta.IsStatusConditionTrue(controlPlaneComponent.Status.Conditions, string(v1beta1.ControlPlaneComponentRolloutComplete))
+}
+
+const (
+	clusterAutoscalerNotAvailableMsg       = "cluster autoscaler not available"
+	clusterAutoscalerRolloutNotCompleteMsg = "cluster autoscaler rollout not complete"
+	clusterAutoscalerNotReadyMsg           = "cluster autoscaler not ready"
+)
+
+func controlPlaneClusterAutoscalerNotReadyMessage(controlPlaneComponent *v1beta1.ControlPlaneComponent) string {
```
miguelsorianod: This should be a method of the operationClusterUpdate type

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation.go
```diff
@@ -434,3 +440,72 @@ func (c *operationClusterUpdate) clusterServiceClusterNodeDrainTimeoutSpecMatche
...
+	}
+	if !isControlPlaneClusterAutoscalerReady(controlPlaneComponent) {
+		message := controlPlaneClusterAutoscalerNotReadyMessage(controlPlaneComponent)
+		logger.Info("cluster autoscaler ControlPlaneComponent is not ready", "message", message)
+		return newOperationState(arm.ProvisioningStateUpdating, message), nil
+	}
+	return newOperationState(arm.ProvisioningStateSucceeded, ""), nil
+}
+
+func isControlPlaneClusterAutoscalerReady(controlPlaneComponent *v1beta1.ControlPlaneComponent) bool {
```
miguelsorianod: This should be a method of the operationClusterUpdate type

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation_test.go
```diff
@@ -1437,3 +1445,173 @@ func TestClusterServiceClusterSpecOperationState(t *testing.T) {
...
+		controlPlaneComponent *v1beta1.ControlPlaneComponent
+		want                  bool
+	}{
+		{
+			name:                  "both Available and RolloutComplete true is ready",
+			controlPlaneComponent: readyControlPlaneClusterAutoscaler(),
+			want:                  true,
+		},
+		{
+			name: "Available false is not ready",
```
miguelsorianod, line 1780: add more test cases:
* rollout is not ready
* condition does not exist (for each)
* condition is unknown (for each)

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation_test.go
```diff
@@ -1437,3 +1445,173 @@ func TestClusterServiceClusterSpecOperationState(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tt.want, isControlPlaneClusterAutoscalerReady(tt.controlPlaneComponent))
+		})
+	}
+}
+
+func TestControlPlaneClusterAutoscalerNotReadyMessage(t *testing.T) {
```
miguelsorianod, line 1859: This unit test seems doesn't really cover exhaustively.

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation_test.go
```diff
@@ -1437,3 +1445,173 @@ func TestClusterServiceClusterSpecOperationState(t *testing.T) {
...
+			wantState:      arm.ProvisioningStateUpdating,
+			wantMessage:    "cluster autoscaler state not cached yet",
+		},
+		{
+			name: "treats 4.20 nightly as supporting autoscaler ControlPlaneComponent",
+			activeVersions: []api.HCPClusterActiveVersion{{
+				Version: ptr.To(semver.MustParse("4.20.0-0.nightly-2026-01-01-000000")),
+			}},
+			wantState:   arm.ProvisioningStateUpdating,
+			wantMessage: "cluster autoscaler state not cached yet",
```
miguelsorianod: Is this test case correct? this message seems to show when the readdesire is not cached yet, but the test title is about something completely different.

Also improve title, this is not about "4.20" in particular, but rather the treatment of nightlies when the autoscaler has a version that supports autoscaler right?

backend/pkg/controllers/operationcontrollers/operation_cluster_update_state_calculation_test.go
```diff
@@ -1437,3 +1445,173 @@ func TestClusterServiceClusterSpecOperationState(t *testing.T) {
...
+		},
+		{
+			name:           "succeeds when autoscaler ControlPlaneComponent is ready",
+			activeVersions: []api.HCPClusterActiveVersion{{Version: ptr.To(semver.MustParse("4.20.0"))}},
+			cachedControlPlaneClusterAutoscalerReadDesire: newControlPlaneClusterAutoscalerReadDesire(t, readyControlPlaneClusterAutoscaler()),
+			wantState: arm.ProvisioningStateSucceeded,
+		},
+	}
```
miguelsorianod, line 2011: I find a basic test case missing: updating when the controlplanecomponent is not ready

shitaljante:
Added test cases as per above comments

backend/pkg/controllers/operationcontrollers/operation_cluster_update_test.go
shitaljante: Added two cases:
"autoscaler all checks pass but autoscaler ReadDesire not cached yet keeps operation updating", "autoscaler all checks pass, operation transitions to succeeded",

shitaljante:
This is weird `make test` runs successfully locally, also a different test fails each time in the pipeline with error - "msg":"cosmos migration encountered errors for subscription","subscriptionID":"sub-new","err":"subscription doc: failed to get subscription \"sub-new\": simulated subscription fetch failure" followed by nil pointer exception. 

I'm going to try one more time and see if I can figure out what's missing.
/test test-unit

miguelsorianod:
> This is weird `make test` runs successfully locally, also a different test fails each time in the pipeline with error - "msg":"cosmos migration encountered errors for subscription","subscriptionID":"sub-new","err":"subscription doc: failed to get subscription "sub-new": simulated subscription fetch failure" followed by nil pointer exception.
> 
> I'm going to try one more time and see if I can figure out what's missing. /test test-unit

It seems it failed again. The issue is a nil panic:
```
...
 --- FAIL: TestCreateClusterScopedReadDesires_SyncOnce (0.00s)
    --- FAIL: TestCreateClusterScopedReadDesires_SyncOnce/creates_HostedCluster_and_cluster-autoscaler_ReadDesires (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1127be0]
goroutine 103 [running]:
testing.tRunner.func1.2({0x1325ea0, 0x21c1ad0})
	/usr/lib/golang/src/testing/testing.go:1872 +0x237
testing.tRunner.func1()
	/usr/lib/golang/src/testing/testing.go:1875 +0x35b
panic({0x1325ea0?, 0x21c1ad0?})
	/usr/lib/golang/src/runtime/panic.go:783 +0x132
github.com/Azure/ARO-HCP/backend/pkg/controllers.(*createClusterScopedReadDesiresSyncer).SyncOnce(0xc0004adef8, {0x17b7220, 0xc00012d4a0}, {{0x156fd7c, 0x24}, {0x1545680, 0x7}, {0x154c10d, 0xc}})
	/go/src/github.com/Azure/ARO-HCP/backend/pkg/controllers/create_cluster_scoped_read_desires_controller.go:125 +0x4e0
github.com/Azure/ARO-HCP/backend/pkg/controllers.TestCreateClusterScopedReadDesires_SyncOnce.func8(0xc000102a80)
	/go/src/github.com/Azure/ARO-HCP/backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go:243 +0x29b
testing.tRunner(0xc000102a80, 0xc000452120)
	/usr/lib/golang/src/testing/testing.go:1934 +0xea
created by testing.(*T).Run in goroutine 86
	/usr/lib/golang/src/testing/testing.go:1997 +0x465
FAIL	github.com/Azure/ARO-HCP/backend/pkg/controllers	0.038s
...
```

shitaljante:
Independently of my PR, an unrelated refactor of this controller was merged into main for "ServiceProvider reads in controller," which is the root cause of the nil panic.

After syncing my local branch with the latest main, I was able to reproduce the issue in the unit tests and fixed the test cases accordingly.

backend/pkg/controllers/create_cluster_scoped_read_desires_controller_test.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+}
+
+func TestCreateClusterScopedReadDesires_SyncOnce(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name                    string
+		resources               []any
+		serviceProviderClusters []*api.ServiceProviderCluster
```
miguelsorianod: No need to define a whole array. We are only going to retrieve a single SPC as the real logic only retrieves one and it doesn't seem likely the controller will evolve to need SPCs from other clusters other than the one being processed

miguelsorianod: nitpick: rename to cachedServiceProviderCluster (it makes it clear it comes from a lister)

PR #5333, opened 2026-05-19
feat: support FIPS mode via feature flags & cluster tags

mzazrivec:
https://redhat.atlassian.net/browse/ARO-27735

### What

This pull request adds the ability to provision a new ARO-HCP cluster in FIPS mode using Azure feature flags & new optional cluster tag `aro-hcp.experimental.cluster.fips-enabled: true / false`. This implementation is to validate that the provisioning of a fips-enabled ARO-HCP cluster works end to end. Eventually, this implementation will be removed in favor of specifying the FIPS flag via standard ARO-HCP API.

### Why

This change allows for creating an ARO-HCP cluster with fips-enabled node pools (the ARO-HCP control plane pods are always deployed on fips-enabled clusters). 

### Testing

The PR includes end2end test for the feature. To manually test the feature, enable the `microsoft.redhatopenshift/experimentalreleasefeatures` AFEC for your subscription and add the `aro-hcp.experimental.cluster.fips-enabled: true` tag to the newly created cluster. The tag needs to be set during cluster installation and changing it later won't have any effect (this is a day-1 operation).

### PR Checklist
- [x] PR is scoped to a single task (no mixed concerns)
- [x] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [x] Summary explains the "Why" behind the change
- [x] Linked to relevant ticket/issue
- [ ] Screenshots included (if graph/UI/metrics changes)
- [x] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [x] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

PR #5935, opened 2026-07-06
feat: FIPS for ARO-HCP

mzazrivec:
https://redhat.atlassian.net/browse/ARO-17753
https://redhat.atlassian.net/browse/ARO-14930
https://redhat.atlassian.net/browse/ARO-16233

### What

This pull request adds the ability to provision a new ARO-HCP cluster in FIPS mode using a new v20260630 API model `cryptoRestrictions`. This implementation replaces the PoC from https://github.com/Azure/ARO-HCP/pull/5333

### Why

This change allows for creating an ARO-HCP cluster with fips-enabled node pools (the ARO-HCP control plane pods are always deployed on fips-enabled clusters).

### Testing

This pull request contains unit tests for the modified code and makes some modification to the existing e2e FIPS test from https://github.com/Azure/ARO-HCP/pull/5333

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:


api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -134,6 +134,11 @@ model HcpOpenShiftClusterProperties {
   @visibility(Lifecycle.Read)
   @added(Versions.v2026_06_30_preview)
   status?: ResourceStatus;
+
+  /** Cluster uses FIPS validated cryptographic libraries */
```
miguelsorianod: Can we find out exactly what does this mean and perform? this documentation doesn't seem to really explain what is this about.

Internally, this ends up setting hypershift's HostedCluster `.spec.FIPS` attribute set to true. Even when looking at the Hypershift doc, that might not be good enough. Let's make sure we can properly describe the functionality.

internal/admission/admit_cluster.go
```diff
@@ -125,7 +124,7 @@ func mutateClusterExperimentalFeatures(_ context.Context, admissionContext *Clus
 	var errs field.ErrorList
 
 	// Reject unrecognized experimental tags.
-	knownTags := sets.New(api.TagClusterSingleReplica, api.TagClusterSizeOverride, api.TagClusterCPOImageOverride, api.TagClusterFIPSEnabled)
+	knownTags := sets.New(api.TagClusterSingleReplica, api.TagClusterSizeOverride, api.TagClusterCPOImageOverride)
```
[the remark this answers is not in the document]
miguelsorianod: Are we aware of this?

internal/api/v20260630preview/generated/time_rfc3339.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by Microsoft (R) AutoRest Code Generator (autorest: 3.10.9, generator: @autorest/go@4.0.0-preview.74)
+// Code generated by Microsoft (R) AutoRest Code Generator (autorest: 3.10.8, generator: @autorest/go@4.0.0-preview.74)
```
miguelsorianod: It seems there's a discrepancy here of generators used. Something went not quite right here when running the tools. This applies to other generated files

internal/validation/validate_cluster.go
```diff
@@ -270,6 +273,9 @@ func validateClusterCustomerProperties(ctx context.Context, op operation.Operati
 		validateImageDigestMirror,
 	)...)
 
+	// FIPSEnabled
+	errs = append(errs, immutableByCompare(ctx, op, fldPath.Child("fipsEnabled"), &newObj.FIPSEnabled, safe.Field(oldObj, toFIPSEnabled))...)
```
miguelsorianod: Because this has been introduced in a new api-version, we shouldn't accept fipsEnabled set to true for older versions. We need to perform version-aware checks. @ahitacat is working on introducing mechanisms to do that. Please sync with her around this.

internal/api/types_cluster.go
```diff
@@ -63,6 +63,7 @@ type HCPOpenShiftClusterCustomerProperties struct {
 	Etcd                    EtcdProfile                 `json:"etcd,omitempty"`
 	ClusterImageRegistry    ClusterImageRegistryProfile `json:"clusterImageRegistry,omitempty"`
 	ImageDigestMirrors      []ImageDigestMirror         `json:"imageDigestMirrors,omitempty"`
+	FIPSEnabled             bool                        `json:"fipsEnabled"`
```
miguelsorianod: @SudoBrendan @bennerv @mbarnes you that are more familiar with introduction of new api versions and fields: is the way it's being added here correct? bool (non pointer), no omitempty, we rely on the Go data type's default (false) which means older API versions will have it as disabled (which is what we want) and no explicit setting to false and so on

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mzazrivec: @miguelsorianod sure. I updated the doc string here, let me know if it's sufficient.

internal/api/v20260630preview/generated/time_rfc3339.go
mzazrivec: Yep, older autorest installation in my devel setup. Fixed now.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -134,6 +134,15 @@ model HcpOpenShiftClusterProperties {
...
   status?: ResourceStatus;
+
+  /** fipsEnabled is a install-time flag to enable FIPS mode for cluster worker nodes.
+   *
+   * When set to true, the nodes will be provisioned and booted in FIPS mode, which won't
+   * allow the nodes to use non-FIPS validated cryptographic libraries
+   */
+  @added(Versions.v2026_06_30_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  fipsEnabled?: boolean = false;
```
machi1990: Let's follow the API convention of not using booleans for the API; inspired by https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#do-not-use-boolean-fields

machi1990 (changes requested):

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mzazrivec: @machi1990 is this an Azure API thing? Because you're pointing to an openshift documentation. And I do see booleans being used in other places of ARO-HCP API.

mzazrivec: @machi1990 actually, I talked to @SudoBrendan yesterday about this and agreed on not using the bool here. So I removed the `fipsEnabled` boolean in favor of a model of its own (pushed now).

internal/api/types_cluster.go
mzazrivec: I also modified now the API and the internal struct so that the bool would go away.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -165,6 +170,32 @@ union ExternalAuthProvisioningState {
   "AwaitingSecret",
 }
 
+/** FIPS mode configuration for cluster nodes */
+@added(Versions.v2026_06_30_preview)
+model FipsProfile {
+  /** state is an install-time flag to enable FIPS mode for cluster worker nodes.
+   * This can only be set during cluster creation and cannot be changed after cluster creation.
+   * When set to "Enabled", the nodes will be provisioned and booted in FIPS mode, which won't allow
+   * the nodes to use non-FIPS validated cryptographic libraries. When set to "Disabled", the cluster
```
miguelsorianod: What does it mean that "nodes can't use non-FIPS validated cryptographic libraries" ? at what level?

mzazrivec: So in general, the idea is that if a machine is booted in FIPS mode (i.e. there's a kernel boot argument), the libraries installed on that machine (for example OpenSSL) won't allow anyone to use non-fips compliant algorithms (MD5 being a good example). The reality can be more complicated, because apps running in the container may be using its own implementation of crypto algorithms and not care about the FIPS switch at all (but that's a different problem really).

In short, if the worker node is booted with fips=1, the kernel and the libraries will do the right thing and good behaving apps will follow.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/examples/2026-06-30-preview/HcpOpenShiftClusters_CreateOrUpdate_MaximumSet_Gen.json
```diff
@@ -87,7 +87,10 @@
         "nodeDrainTimeoutMinutes": 20,
         "clusterImageRegistry": {
           "state": "Enabled"
-        }
+        },
+        "fips": {
+	  "state": "Disabled"
```
bennerv: formatting is wrong, please update formatting here.

test/e2e/cluster_fips_mode.go
```diff
@@ -69,13 +72,13 @@ var _ = Describe("FIPS Mode Support", func() {
 				Expect(err).NotTo(HaveOccurred(), "failed to create cluster customer resources")
 
 				By("creating the ARO-HCP cluster with FIPS enabled")
```
bennerv: need to skip & timebomb cluster installation if the 2026-06-30-preview API isn't available.  Please look at the existing example we have to check against it.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/examples/2026-06-30-preview/HcpOpenShiftClusters_CreateOrUpdate_MaximumSet_Gen.json
mzazrivec: Fixed

test/e2e/cluster_fips_mode.go
mzazrivec: Yep, added.

mbarnes:
Informal LGTM.

I assume migrating any existing clusters that were using the experimental flag isn't a concern?

mzazrivec:
> I assume migrating any existing clusters that were using the experimental flag isn't a concern?

@mbarnes There should not be any clusters like that alive anywhere. The ones I used for the devel work are deleted and so should be the ones created during the e2e tests.

internal/validation/validate_cluster.go
mzazrivec: I might be missing something here, but I can think of 3 basic scenarios only:
**1st. scenario:** v2026 cluster with fips disabled (or enabled), then v2026 patch to fips enabled (or disabled): we don't need a version-aware check here, the `immutableByCompare()` function will catch this.
**2nd scenario:** v2026 cluster with fips disabled (or enabled), then v2025 patch to fips enabled (or disabled): we don't need a version-aware check here, because this scenario will fail earlier during the request unmarshalling (unknown field "fips").
**3rd scenario:** v2025 (or even older) cluster with fips disabled (by default), then v2026 patch to fips enabled: we don't need a version-aware check here, the `immutableByCompare()` function will catch this.

So I don't think we need a version-aware check here. Or is there any other scenario which I need to consider?

miguelsorianod: @ahitacat can you chime in here?

internal/ocm/convert.go
```diff
@@ -136,6 +136,17 @@ func convertOutboundTypeRPToCS(outboundTypeRP api.OutboundType) (string, error)
 	}
 }
 
+func convertFipsState(fipsState api.FIPSState) (bool, error) {
```
miguelsorianod: Fips -> FIPS

internal/validation/validate_cluster.go
```diff
@@ -578,6 +578,19 @@ func validateCustomerIngressProfile(ctx context.Context, op operation.Operation,
 	return errs
 }
 
+var (
+	toFipsState = func(oldObj *api.FIPSProfile) *api.FIPSState { return &oldObj.State }
+)
+
+func validateFipsProfile(ctx context.Context, op operation.Operation, fldPath *field.Path, newObj, oldObj *api.FIPSProfile) field.ErrorList {
```
miguelsorianod: validateFipsProfile -> validateFIPSProfile

internal/validation/validate_cluster_comprehensive_test.go
```diff
@@ -1827,6 +1827,66 @@ func TestValidateClusterUpdate(t *testing.T) {
 				{Message: "field is immutable", FieldPath: "customerProperties.clusterImageRegistry.state"},
 			},
 		},
+		{
+			name: "immutable FIPS enabled - update (Disabled to Enabled)",
```
miguelsorianod: The test case titles are confusing. What does it mean "immutable Fips enabled" ? Similar comment for the other test case titles.

For example I would word this one something like:

immutable FIPS - update (Disabled to Enabled)

internal/validation/validate_cluster_comprehensive_test.go
```diff
@@ -1827,6 +1827,66 @@ func TestValidateClusterUpdate(t *testing.T) {
...
+				c := createValidCluster()
+				c.CustomerProperties.FIPS.State = api.FIPSStateDisabled
+				return c
+			}(),
+			expectErrors: []utils.ExpectedError{
+				{Message: "field is immutable", FieldPath: "customerProperties.fips.state"},
+			},
+		},
+		{
+			name: "immutable FIPS enabled - update (Enabled to Disabled)",
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/5935/changes#r3623395677

internal/validation/validate_cluster_comprehensive_test.go
```diff
@@ -1827,6 +1827,66 @@ func TestValidateClusterUpdate(t *testing.T) {
...
+				c := createValidCluster()
+				c.CustomerProperties.FIPS.State = api.FIPSStateEnabled
+				return c
+			}(),
+			expectErrors: []utils.ExpectedError{
+				{Message: "field is immutable", FieldPath: "customerProperties.fips.state"},
+			},
+		},
+		{
+			name: "FIPS enabled unchanged - update (both Disabled)",
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/5935/changes#r3623395677. The "enabled" part seems doesn't belong.

Maybe you wanted to say FIPS unchanged - update (both Disabled) ?

internal/validation/validate_cluster_comprehensive_test.go
```diff
@@ -1827,6 +1827,66 @@ func TestValidateClusterUpdate(t *testing.T) {
...
+			}(),
+			oldCluster: func() *api.HCPOpenShiftCluster {
+				c := createValidCluster()
+				c.CustomerProperties.FIPS.State = api.FIPSStateDisabled
+				return c
+			}(),
+			expectErrors: []utils.ExpectedError{},
+		},
+		{
+			name: "FIPS enabled unchanged - update (both Enabled)",
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/5935/changes#r3623403123

test/util/framework/deployment_params.go
```diff
@@ -134,6 +134,7 @@ type ClusterParams20260630 struct {
 	ChannelGroup                  string
 	AuthorizedCIDRs               []*string
 	Autoscaling                   *hcpsdk20260630preview.ClusterAutoscalingProfile
+	Fips                          *hcpsdk20260630preview.FipsProfile
```
miguelsorianod: Fips -> FIPS

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -165,6 +170,32 @@ union ExternalAuthProvisioningState {
   "AwaitingSecret",
 }
 
+/** FIPS mode configuration for cluster nodes */
+@added(Versions.v2026_06_30_preview)
+model FipsProfile {
```
miguelsorianod: FipsProfile -> FIPSProfile

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -165,6 +170,32 @@ union ExternalAuthProvisioningState {
...
+   * nodes will be using all standard cryptographic libraries. The default is Disabled.
+   */
+  @added(Versions.v2026_06_30_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  state?: FipsState = FipsState.Disabled;
+}
+
+/** The desired FIPS mode state for cluster nodes */
+@added(Versions.v2026_06_30_preview)
+union FipsState {
```
miguelsorianod: FipsState -> FIPSState

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
```diff
@@ -274,6 +280,15 @@ func newResourceStatus(from []metav1.Condition) *generated.ResourceStatus {
 	}
 }
 
+func newFipsProfile(from *api.FIPSProfile) generated.FipsProfile {
```
miguelsorianod: newFipsProfile -> newFIPSProfile

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
```diff
@@ -692,6 +711,10 @@ func normalizeImageDigestMirrors(p []*generated.ImageDigestMirror, out *[]api.Im
 	*out = slice
 }
 
+func normalizeFips(p generated.FipsProfile, out *api.FIPSProfile) {
```
miguelsorianod: normalizeFips -> normalizeFIPS

internal/validation/validate_cluster.go
```diff
@@ -578,6 +578,19 @@ func validateCustomerIngressProfile(ctx context.Context, op operation.Operation,
 	return errs
 }
 
+var (
+	toFipsState = func(oldObj *api.FIPSProfile) *api.FIPSState { return &oldObj.State }
```
miguelsorianod: toFipsState -> toFIPSState

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
miguelsorianod: To be consistent with the other normalize functions, pass a pointer instead of a value: `p *generated.FIPSProfile`

internal/validation/validate_cluster.go
miguelsorianod: @mzazrivec

>  2nd scenario: v2026 cluster with fips disabled (or enabled), then v2025 patch to fips enabled (or disabled): we don't need a version-aware check here, because this scenario will fail earlier during the request unmarshalling (unknown field "fips").

Could you point me to where this occurs?

What about the scenario where older versions pass fips state enabled? is that possible?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mbarnes: TypeSpec validation might reject that, but you can [override the name for client SDKs](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/decorators/#@Azure.ClientGenerator.Core.clientName).
```
@clientName("FIPSState")
```
I did that for several other fields in this spec for things like `CIDR` instead of `Cidr`, `CA` (certificate authority) instead of `ca`, etc.

internal/validation/validate_cluster.go
ahitacat: I think we don't require the version validation. As what @mzazrivec said, we can use the unmarshalling so older versions will no set this variable to enabled, and the immutableByCompare validation to no update it.

For scenario 2: During a patch we create the new cluster with `decodeDesiredClusterPatch` it creates a hcpOpenShiftCluster with the v2025 model and using the current cosmos cluster (the cluster specifies the FIPS but this new hcpOpenShiftCluster doesn't have it). Then calls to `ApplyRequestBody`  that applies the new payload to the hcpOpenShiftCluster.  This method calls json.Unmarshal(modifiedData, v) which unmarshals the merged JSON back into the v2025 struct -> here is where it fails.

So, for new fields like this one I think we are fine by not validating if the call is done with an older api-version, as unmarshall will fail. 

This validation should be done in cases where the field was already in older versions and we change the method that can be used in that field (i.e. that field is updatable in newer versions)

internal/ocm/convert.go
mzazrivec: Fixed.

internal/validation/validate_cluster.go
mzazrivec: Changed as suggested.

internal/validation/validate_cluster_comprehensive_test.go
mzazrivec: Yep, changed as suggested.

internal/validation/validate_cluster_comprehensive_test.go
mzazrivec: Changed as suggested.

internal/validation/validate_cluster_comprehensive_test.go
mzazrivec: Yep, removed the extraneous `enabled`.

internal/validation/validate_cluster_comprehensive_test.go
mzazrivec: Changed as suggested.

test/util/framework/deployment_params.go
mzazrivec: Changed as suggested.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mzazrivec: Changed as suggested.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mzazrivec: I actually used the following directive:
```
/** The desired FIPS mode state for cluster nodes */
#suppress "@azure-tools/typespec-azure-core/casing-style"
union FIPSState {
   ...
}
```
and it seems to have the same (or similar) effect.

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
mzazrivec: Changed as suggested.

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
mzazrivec: Changed as suggested.

internal/validation/validate_cluster.go
mzazrivec: Changed as suggested.

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
mzazrivec: Changed as suggested.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
miguelsorianod: @mbarnes does that have some unintended/undesired side effect?

miguelsorianod: I still see some references to `Fips` in the PR changes in general, are we missing some change?

miguelsorianod: > and it seems to have the same (or similar) effect.

Can we understand what are the differences and whether it's desirable to "suppress" or not? (whatever that means and implies on the type it's attached to and potentially to its subtypes)

mzazrivec: @miguelsorianod That's the generated attr. from the top level api `fips` attribute. Do you want that as `FIPS` as well? I think that would need the `@clientName` directive Matt is mentioning above (or is using `FIPS` as the top level api attr. ok?)

mzazrivec: > Can we understand what are the differences and whether it's desirable to "suppress" or not? (whatever that means and implies on the type it's attached to and potentially to its subtypes)

`@clientName` does not suppress model name validation (i.e. you would still need to the PascalCase). It instructs the generator tools to use an explicit type when generating the SDK code from it. So you're going to have `FipsProfile` in API.tsp and `FIPSProfile` in SDK.

`#suppress "@azure-tools/typespec-azure-core/casing-style"` suppresses the PascalCase on model names and whatever model name you used is what will be propagated by the generator to the SDK code (unless it is overriden by `@clientName()`. So you're going to have `FIPSProfile` in API.tsp and the same in SDK (unless you override it for SDK).

miguelsorianod: I see. If we are consistent with what we've done for CIDR, Ca and such let's do the following:
Let's leave Fips, FipsProfile in tsp, with the clientName annotation with the uppercasing. In that way we are consistent with the other changes and we have the Go types with the uppercase acronyms

internal/api/enums.go
```diff
@@ -258,3 +258,18 @@ var (
 		MirrorSourcePolicyAllowContactingSource,
 	)
 )
+
+// CryptoRestrictions represents cryptographic restrictions for a cluster.
+type CryptoRestrictions string
+
+const (
+	CryptoRestrictionsNone CryptoRestrictions = "None"
```
miguelsorianod, line 267: Document the constants

mzazrivec: Added

miguelsorianod:
/lgtm
/approve

on the non e2e tests part

PR #5851, opened 2026-06-30
fix(infra): provision postgres access via Go pipeline step (ARO-28051)

raelga:
[ARO-28051](https://issues.redhat.com/browse/ARO-28051)

### What

Replaces the clusters-service and maestro Postgres `deploymentScripts`
(`postgres-sql.bicep` / `postgres-access.bicep`) with a Go program
(`dev-infrastructure/scripts/postgres-access`) run from a pipeline Shell step.
The binary registers each service managed identity as a db-level `pgaadauth`
principal and grants it the application database — the same end-state the old
script produced.

The Go module is pre-built by a pipeline `buildStep` and invoked from the
`cs-postgres-access` / `maestro-postgres-access` Shell steps, following the
`recreate-system-pool` (#5397) and `cleanup-pko-resources` (#5149) precedent.

### Why

ARM `Microsoft.Resources/deploymentScripts` mount their script content into ACI
through Azure Files, which requires shared-key storage. AME Azure Policy
**SFI-ID4.2.1** forbids `allowSharedKeyAccess` storage accounts, so the
deploymentScript path is not viable.

The old deploymentScript ran `psql` inside its own ACI image. The EV2 Shell
runner image has **no postgres client**, so the grant is implemented in Go with
the pure-Go `lib/pq` driver and an Entra (`oss-rdbms`) token — no `psql`/`libpq`
needed.

Unlike the ARM-only dual-admin alternative (#5833), this keeps registering the
MI as a **db-level** principal rather than a second **server-level** Entra
admin, so it works on already-deployed servers without the
`Updating an existing principal isn't supported` collision on the upgrade path.

```text
EV2:   shellIdentity MSI (server Entra admin) -> token -> lib/pq -> grants
local: az-login operator -> enrolled as Entra admin -> token -> lib/pq -> grants
```

### Testing

- **Unit tests** (`main_test.go`): env parsing, idempotent grant/schema SQL with
  hyphenated-identifier quoting, DSN escaping, JWT user-vs-MSI classification.
- **Manual end-to-end in a personal dev environment** against both flex servers
  on the upgrade path (servers already deployed from `main`, MIs already
  registered as db-roles):
  - `clusters-service` and `maestro` grants both succeeded (`rc=0`), **no
    principal collision**.
  - Verified via `psql` that the role carries the correct
    `pgaadauth` security label (`aadauth,oid=<mi-principal-id>,type=service`)
    and has `CONNECT` on the database.
[...]

PR #5945, opened 2026-07-07
fix(infra): handle service-principal auth in postgres-access step

inbharajmani:
https://redhat.atlassian.net/browse/AROSLSRE-1417

## Summary

- The `postgres-access` Go script (introduced in #5851, [ARO-28051](https://redhat.atlassian.net/browse/ARO-28051)) fails in Prow CI with a PostgreSQL `pgaadauth` OID mismatch because it only handles EV2 (managed identity) and local operator (UPN) auth paths
- Add service-principal detection: when the token has no UPN and no `xms_mirid` (MSI marker), enrol the CI SP as an Entra admin and connect as its `appid`
- Distinguish MSI from SP tokens via the `xms_mirid` claim to preserve the EV2 path

## Context

The script has three callers:
1. **EV2** — `shellIdentity` runs as the rollout MSI → token OID matches postgres role → works (unchanged)
2. **Local operator** — has UPN → connects as UPN, enrols as admin → works (unchanged)
3. **Prow CI** — templatize ignores `shellIdentity`, runs as CI service principal → **was broken**, now fixed

Ephemeral CI environments (`ci00`/`ci01`) are unaffected — they use containerized postgres (`deploy: false`), so the step is skipped entirely. The bug only manifests in CSPR, which use Azure Postgres Flexible Server.

This was masked because every CSPR postsubmit since the script landed (June 30) failed earlier at the EventGrid `authenticationName` step before reaching `cs-postgres-access`.

## Failed CI runs

- [2074360206455738368](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/branch-ci-Azure-ARO-HCP-main-cspr-pipeline-postsubmit/2074360206455738368) — `cs-postgres-access` OID mismatch

## Test plan

- [x] Existing unit tests pass (10/10)
- [x] New `TestIsManagedIdentityToken` and `TestServicePrincipalAppID` tests added and passing
- [ ] CSPR postsubmit passes `cs-postgres-access` step after merge

### DIFF
--- a/dev-infrastructure/scripts/postgres-access/main.go
+++ b/dev-infrastructure/scripts/postgres-access/main.go
@@ -189,6 +189,20 @@ func run(ctx context.Context) error {
 		if err := ensureEntraAdmin(ctx, cred, cfg, claims); err != nil {
 			return fmt.Errorf("enroll operator as Entra admin: %w", err)
 		}
+	} else if !isManagedIdentityToken(claims) {
+		// Running as a service principal (Prow CI). templatize ignores
+		// shellIdentity, so the token belongs to the CI SP, not the rollout
+		// MSI. Enrol the SP as an Entra admin and connect as its appId.
+		// MSI tokens also carry appid but are distinguished by the
+		// xms_mirid claim — those take the default adminUserName path.
+		appID := servicePrincipalAppID(claims)
+		if appID == "" {
+			return fmt.Errorf("token has no UPN, appid, or xms_mirid — cannot determine caller identity")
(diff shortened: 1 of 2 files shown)


raelga (approved):
/lgtm
/approve

geoberle:
i'm working on a runmode for templatize that honors the shell step `identityFrom`
consider doing low effort work on this to keep it from failing in prow, especially because we don't have postgres there

just make sure the step does not run when azure postgres is disabled

PR #5952, opened 2026-07-07
backend: add alerting for stuck async operations

stevekuznetsov:
We have end-to-end test assertions that our async operations never take longer than $timeout. We have seen a few cases where customer interactions with the ARM API ended up stuck, and had no visibility into this, even though our end-to-end test suite, had it run on that region at that time, would have clearly caught this.

This new set of alerts is an assertion from the backend that async operations never spend more than a certain amount of time in a particular phase. We still benefit from having our async operations not having a clear timeout in the backend, as fixing the underlying problem will cause them to get un-stuck and finish, but we want to know if we're failing to progress.

Existing SLO alerting is for failed/terminal phases of async operations, which do not trigger for 'stuck' operations which never exit 'progressing.'

### DIFF
--- a/backend/alerts/backend-async-operations-prometheusRule.yaml
+++ b/backend/alerts/backend-async-operations-prometheusRule.yaml
@@ -1,0 +1,361 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    app.kubernetes.io/name: kube-prometheus
+    app.kubernetes.io/part-of: kube-prometheus
+    prometheus: k8s
+    role: alert-rules
+  name: backend-async-operations-rules
+  namespace: monitoring
+spec:
+  groups:
+  # ========================================
+  # Backend Async Operation Stuck Alerts
+  # ========================================
+  # These alerts fire when an async operation has been running longer than
+  # the corresponding e2e test timeout. Thresholds are anchored to the
+  # canonical timeout constants in test/util/framework/constants.go.
+  #
+  # Metric: backend_resource_operation_start_time_seconds (total wall-clock
+  # time since the operation was created). Combined with phase_info to
+  # restrict to operations still in non-terminal phases.
+  #
+  # No `for` clause: the phase filter already ensures we only alert on
+  # in-flight operations. Once an operation reaches succeeded/failed/canceled,
+  # the phase no longer matches and the alert resolves immediately.
+  #
+  # All expressions use max by(...) to deduplicate across Prometheus
+  # replicas/shards, preventing duplicate alerts for the same operation.
+  - name: backend-async-operations
+    rules:
+    # ----------------------------------------
+    # Cluster operations
+    # ----------------------------------------
+    # Cluster create: e2e timeout = 20 minutes (ClusterCreationTimeout)
+    - alert: BackendAsyncOperationClusterCreateStuck
+      expr: |
+        (
+          max by (resource_id, subscription_id, resource_type, operation_type, phase, cluster) (
+            time() - backend_resource_operation_start_time_seconds{
+              resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+              operation_type="create"
+            }
+          )
+          and
+          max by (resource_id, subscription_id, resource_type, operation_type, phase, cluster) (
+            backend_resource_operation_phase_info{
+              resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+              operation_type="create",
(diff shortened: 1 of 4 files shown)


machi1990 (commented):
lgtm, intentionally not tagging it to give a member of SL team to see if the wiring is done correctly and wether the missing runbook is going to be a problem.

stevekuznetsov:
Will rebase on https://github.com/Azure/ARO-HCP/pull/5955

backend/alerts/backend-async-operations-prometheusRule.yaml
```diff
@@ -0,0 +1,298 @@
...
+        (
+          (time() - backend_resource_operation_start_time_seconds{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="revokecredentials"
+          })
+          and
+          backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="revokecredentials",
+            phase=~"accepted|provisioning"
```
SudoBrendan: Revocation lifecycle appears to be `accepted -> deleting -> succeeded/failed`

backend/alerts/backend-async-operations-prometheusRule.yaml
```diff
@@ -0,0 +1,298 @@
...
+      annotations:
+        summary: 'External auth delete operation stuck'
+        description: 'External auth delete operation for {{ $labels.resource_id }} has been in deleting phase for over 25 minutes (e2e timeout: ExternalAuthDeletionTimeout).'
+        runbook_url: TBD
+    # ----------------------------------------
+    # Backstop: any async operation not covered above
+    # ----------------------------------------
+    # Catches any future operation types that don't have explicit alerts.
+    # Threshold (60 min) is higher than the longest specific alert (45 min)
+    # to avoid double-firing before the specific alert has had time to fire.
```
SudoBrendan: nit: the test case that validates this backstop is validation that we get a 2x fire even for existing alerts (cluster create). I think this can be ok to preserve a true backstop, but the comment is misleading.

However, I could see a 2x queue administrative cost of these tickets getting burdensome quickly. Are you aware of a way we'd be able to reconcile the duplication at the ICM connector layer perhaps? Maybe we can get creative with correlationIds? What do you think?

backend/alerts/backend-async-operations-prometheusRule.yaml
```diff
@@ -0,0 +1,298 @@
...
+            phase=~"accepted|provisioning"
+          } == 1
+        ) > 1200
+      labels:
+        severity: "3"
+      annotations:
+        summary: 'Node pool create operation stuck'
+        description: 'Node pool create operation for {{ $labels.resource_id }} has been in {{ $labels.phase }} phase for over 20 minutes (e2e timeout: NodePoolCreationTimeout).'
+        runbook_url: TBD
+    # Node pool update: e2e timeout = 45 minutes (NodePoolVersionUpgradeTimeout)
```
SudoBrendan, line 200: You pulled this straight from the test case, but I wonder if that test itself is wrong and conflating various kinds of "update", maybe it was from the before-times when we hadn't decided to fire-and-forget on OCP version upgrades? As I understand from a backend Operation perspective, the actual LRO should go basically from Accepted -> Succeeded/Failed immediately on an OCP upgrade, right? We should maybe consider a shorter timeout, so that when customers bump their autoscale max from 8 -> 9 and that gets locked it doesn't take 45m to fire (here, and in the test, honestly)

However... while investigating this - I did discover that there are some LROs that we could expect to go longer than 45m or even an hour - [internal discussion here](https://redhat-external.slack.com/archives/C075PHEFZKQ/p1783548434945969).

SudoBrendan (commented):
Thanks for getting this Steve - been on my list of top alerts for a while; lots of value here.

backend/alerts/backend-async-operations-prometheusRule.yaml
stevekuznetsov: Yes, good catch - we can use the same correlation ID for all of them so they just double-hit one alert.

backend/alerts/backend-async-operations-prometheusRule.yaml
stevekuznetsov: Yeah, if/when we change the e2e, we can change this - and you're 100% right that there are possible updates that could take a long time. If and unless we update the metric to classify the 'type' of update LRO, we can't subdivide. Time will tell if it is meaningful or not.

PR #5954, opened 2026-07-07
feat: alert on stale int and stg HCP clusters

mmazur:
https://redhat.atlassian.net/browse/AROSLSRE-1421

## Why
Int and stg HCP cluster objects should not remain around indefinitely. This adds an SL services alert for any HCP cluster object older than 3 hours in those environments.

## What
- Adds `HCPClusterOlderThan3Hours` in `observability/alerts/hcp-test-clusters-prometheusRule.yaml`.
- Wires the rule into the SL services alert config.
- Uses `max by (cluster, resource_id, environment)` over `backend_cluster_created_time_seconds{environment=~"int|stg"}`.
- Fires after `for: 5m` when cluster object age exceeds 3h.
- Includes `environment` in the summary and description.
- Sets the correlation ID to `IntStgClusterOlderThan3Hours{{ $labels.resource_id }}`.
- Adds rule coverage for int, stg, prod exclusion, and pre-threshold behavior.

## Testing
- `make -C tooling/prometheus-rules run-sl-services`

No screenshots; alert rule only.

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/generatedPrometheusAlertingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedPrometheusAlertingRules.bicep
@@ -1642,6 +1642,48 @@ This may indicate that finalizers are stuck or resources are failing to cleanup.
   }
 }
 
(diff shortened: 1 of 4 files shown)


stevekuznetsov (approved):
/lgtm
/approve

PR #5962, opened 2026-07-07
feat(stg-global): re-encrypt global secrets for arohcpstg2-global (AROSLSRE-1314)

hbhushan3:
[AROSLSRE-1314](https://redhat.atlassian.net/browse/AROSLSRE-1314)

### What

Adds a new `https://arohcpstg2-global.vault.azure.net:` block to `dev-infrastructure/data/encryptedsecrets.yaml` containing all 12 global secrets re-encrypted against the new dedicated STG-global Key Vault `arohcpstg2-global`.

Purely additive: **+53 / −0**, one new vault block. The existing `dev` / `int` / `prod` / `stg` blocks are untouched.

Secrets included (same set as the existing `arohcpstg-global` block): `acm-d-password`, `acm-d-username`, `component-sync-pull-secret`, `ocmirror-pull-secret`, `prow-token`, `quay-io-bearer`, `quay-password`, `quay-username`, `redhat-pending-quay-password`, `redhat-pending-quay-username`, `redhat-prod-quay-password`, `redhat-prod-quay-username`.

### Why

The STG dedicated-subscription migration (epic [AROSLSRE-525](https://redhat.atlassian.net/browse/AROSLSRE-525)) stands up a parallel "V2" global stack in a dedicated `hcp-stg-global` subscription, which includes a **new** global Key Vault `arohcpstg2-global`. The V2 global pipeline's `decrypt-and-ingest-secrets` (SecretSync) step targets that vault but fails because `encryptedsecrets.yaml` has no block for it:

```
keyvault https://arohcpstg2-global.vault.azure.net does not exist in encryption config
```

Envelope encryption is **per-vault**: each secret's `dataEncryptionKey` is the AES DEK wrapped with *that vault's* `secretSyncKey` public key (RSA-OAEP-256). The existing `arohcpstg-global` block's wrapped DEKs are therefore useless to `arohcpstg2-global` — the secrets must be re-registered (re-wrapped) against the new vault's own key rather than copied. Source plaintext was taken from `arohcpdev-global` (the Red Hat–tenant global vault, source-of-truth confirmed with the secret-sync OWNER).

### Testing

Full local `test-decrypt` against the new vault isn't runnable from a Red Hat–tenant machine: `arohcpstg2-global` lives in the AME tenant, and Key Vault `Decrypt` is tenant-scoped, so the private-key unwrap can only be exercised by the pipeline's `decrypt-and-ingest-secrets` step in-environment (the sanctioned register-here / populate-in-env split from [`docs/secret-sync.md`](https://github.com/Azure/ARO-HCP/blob/main/docs/secret-sync.md)). To compensate, the block was validated as far as is possible offline plus the dev source values were validated end-to-end:

- **`make test-decrypt` (dev vault): all 12 PASS** — every source value in the `arohcpdev-global` block decrypts and passes its content validator (JWT parse on `prow-token`, docker-config JSON on the two pull secrets, non-empty on the rest). Confirms the source shapes are valid.
- **KEK authenticity (offline):** SHA-256 of the `keyEncryptionKey` embedded in the new block == SHA-256 of the `secretSyncKey` public key downloaded directly from the `arohcpstg2-global` vault. Proves `register` wrapped against the authentic vault key.
- **Wrapped-DEK sizes (offline):** all 12 `dataEncryptionKey` values decode to exactly 512 bytes = well-formed RSA-4096 OAEP ciphertext for the vault's key.
- **Integrity (offline):** every `data` / `dataEncryptionKey` base64-decodes cleanly; file parses as YAML with 5 vault blocks; `git diff` confirms no deletions/edits outside the new block.
- **Deploy-time:** the private-key unwrap will be exercised by the V2 global `decrypt-and-ingest-secrets` step on the next Step-3 rollout ([AROSLSRE-1200](https://redhat.atlassian.net/browse/AROSLSRE-1200)).

- The diff is intentionally ciphertext-only and additive; no existing block changes.
- After merge, this is pulled downstream into `sdp-pipelines` via a `Revision.mk` bump and promotion to `release/hcp/public/stg`, which unblocks the V2 global deploy's secret ingestion.


raelga (approved):
/lgtm
/approve

PR #5963, opened 2026-07-07
fix: redacting CreatedBy and LastModifiedBy from SystemData

rachelvweber:
[ARO-28164](https://redhat.atlassian.net/browse/ARO-28164)

What
Redacts CreatedBy and LastModifiedBy from HCP cluster data dump

Why
These can include customer email addresses

### Testing

Added unit tests to confirm the fields we need are redacted from hcp cluster, nodepool, and external auth.

I wanted a more general solution than this, so I was planning on using the same library that AKS uses, https://github.com/Azure/redact, but I ran into a number of issues. The library redacts by default, which I think is reasonable. However, some of our objects have types that are in external libraries, like arm.ResourceId, and we can't add the appropriate tags to that type to ensure they are unredacted. I'm planning on making a PR on the redact repo to add the functionality I need, but for now, this will redact the fields that we need.

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/internal/serverutils/dump_data.go
+++ b/internal/serverutils/dump_data.go
@@ -16,15 +16,20 @@ package serverutils
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 
 	"github.com/Azure/ARO-HCP/internal/database"
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
+
+const RedactStr = "REDACTED"
 
 // DumpDataToLogger writes a structured-log entry for every document related
 // to resourceID. It covers three storage layers:
@@ -61,6 +66,10 @@ func DumpDataToLogger(
 	if err != nil {
 		return utils.TrackError(err)
 	}
+	err = redactTypedDocument(startingCosmosRecord)
+	if err != nil {
+		return utils.TrackError(err)
+	}
 	logger.Info(fmt.Sprintf("dumping resourceID %v", startingCosmosRecord.ResourceID),
 		"currentResourceID", resourceIDToString(startingCosmosRecord.ResourceID),
 		"content", startingCosmosRecord,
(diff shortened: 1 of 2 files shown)


internal/serverutils/dump_data.go
```diff
@@ -213,3 +225,39 @@ func DumpBillingToLogger(ctx context.Context, resourcesDBClient database.Resourc
...
+func redactTypedDocument(d *database.TypedDocument) (*database.TypedDocument, error) {
+	if d == nil {
+		return nil, fmt.Errorf("typed document is nil")
+	}
+
+	if len(d.Properties) == 0 {
+		return d, nil
+	}
+
+	var props map[string]any
```
stevekuznetsov: I believe David will want `unstructured.Unstructured` for this kind of manipulation

internal/serverutils/dump_data.go
```diff
@@ -213,3 +225,39 @@ func DumpBillingToLogger(ctx context.Context, resourcesDBClient database.Resourc
...
+			sd["lastModifiedBy"] = RedactStr
+		}
+	}
+
+	redactedProps, err := json.Marshal(props)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal redacted typed document properties for %s: %w", resourceIDToString(d.ResourceID), err)
+	}
+
+	output := *d
```
stevekuznetsov: Do we need the shallow copy? Seems like this would ballon memory use, and if we're expected to mutate the input, might not be necessary.

stevekuznetsov (approved):
/lgtm
/approve

internal/serverutils/dump_data.go
```diff
@@ -61,6 +66,10 @@ func DumpDataToLogger(
 	if err != nil {
 		return utils.TrackError(err)
 	}
+	err = redactTypedDocument(startingCosmosRecord)
```
miguelsorianod, line 69: I understand the desire to redact that information.

If we were to need that information, how would we access it?

machi1990 (approved):
/lgtm

internal/serverutils/dump_data.go
stevekuznetsov: Using the SAW and JIT.

PR #5970, opened 2026-07-08
test: add E2E test for KmsKey update with new version

ahitacat:
### What

Created PR in favor of https://github.com/Azure/ARO-HCP/pull/5746

This PR adds an E2E test that verifies updating an HCP cluster to use a new Azure KMS key version.

**The code is based on https://github.com/Azure/ARO-HCP/pull/5629/changes which needs to be merged before this.**

The test covers the full rotation flow:

creates the test resource group and HCP cluster using the v20260630preview API;
creates the required customer resources;
creates a new version of the existing Azure Key Vault key;
updates the cluster configuration to reference the new key version;
waits for the cluster update operation to complete successfully;
verifies that all StorageVersionMigration resources complete with Succeeded=True, ensuring that Kubernetes objects have been re-encrypted with the new key.

### Why

AzureKmsKey rotation must be validated end-to-end to ensure that updating the KMS key version on an existing cluster correctly triggers the re-encryption flow and leaves the cluster in a healthy state.

Checking only the ARM update result is not enough: the test also verifies the Kubernetes-side StorageVersionMigration resources so that we can confirm the re-encryption process actually completed.

### Testing

This is the E2E test for the work in https://github.com/Azure/ARO-HCP/pull/5629

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


test/util/verifiers/storageversionmigration.go
```diff
@@ -0,0 +1,156 @@
...
+func (v verifyStorageVersionMigrationSucceeded) Verify(ctx context.Context, restConfig *rest.Config) error {
+	dynClient, err := dynamic.NewForConfig(restConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create dynamic client: %w", err)
+	}
+
+	svmList, err := dynClient.Resource(svmGVR).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list StorageVersionMigration resources: %w", err)
+	}
```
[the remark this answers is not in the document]
ahitacat: StorageVersionMigration should have been succeeded by this time, if not the error is during re-encryption so it should be OK to surface it directly.

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,196 @@
...
+
+			By("creating a resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "kms-key-rotate", tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for KMS key rotation test")
+
+			By("resolving latest 4.22 version from candidate channel")
+			resolvedVersion, err := framework.GetLatestInstallVersion(ctx, framework.DefaultOCPChannelGroup, "4.22")
+			if err != nil {
+				Skip(fmt.Sprintf("No install version found for 4.22 in %s channel: %s", framework.DefaultOCPChannelGroup, err.Error()))
+			}
```
machi1990, line 60: I don't think this matters if hypershift provides the override.
Let's just put 4.22 in this case and remove the roundtrip to cincinatti

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+const HCPClusterReencryptionUpgradeTimeout = 15 * time.Minute
+
+var _ = Describe("Customer", func() {
+	// Deadline for v20260630preview API deployment in non-dev environments
+	timeBombDeadline := framework.Must(time.Parse(time.RFC3339, "2026-07-31T00:00:00Z"))
+
+	It("should be able to rotate KMS key for a cluster with version >= 4.22",
```
machi1990, line 40: While I don't have the number of the duration impact this test is going to have vis-a-vis creation of cluster and needing MIs etc I permitted myself to throw out this idea; 
What do you think of consolidating this test case with https://github.com/ahitacat/ARO-HCP/blob/9fcb5539bd2e9c753f873a85e759ca87cac06aab/test/e2e/cluster_update.go#L118 ?

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,196 @@
...
+				"keyName", clusterParams.EtcdEncryptionKeyName,
+				"originalVersion", clusterParams.EtcdEncryptionKeyVersion)
+
+			createKeyResp, err := keyClient.CreateKey(ctx, clusterParams.EtcdEncryptionKeyName, azkeys.CreateKeyParameters{
+				Kty:     to.Ptr(azkeys.KeyTypeRSA),
+				KeySize: to.Ptr(int32(2048)),
+			}, nil)
+			Expect(err).NotTo(HaveOccurred(), "failed to create new key version (rotation)")
+			Expect(createKeyResp.Key).NotTo(BeNil(), "created key response was nil")
+			Expect(createKeyResp.Key.KID).NotTo(BeNil(), "created key ID was nil")
```
mbukatov, line 126: Ah, I see. Running [keyClient.CreateKey](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azkeys#Client.CreateKey) with existing key name creates it's new version, effectively rotating it:

> If the named key already exists, Azure Key Vault creates a new version of the key. It requires the keys/create permission.

And if I'm reading the docs right, this doesn't need any new permission. Sheer create should be enough, which we should have, since this key was created during `tc.CreateClusterCustomerResources20260630()` call few lines above.

mbukatov: That said, Azure KeyVault also supports [RotateKey](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azkeys#Client.RotateKey) API call, which requires a key rotation policy. But I'm not sure if and how we support auto rotation. 

We should consider covering it if we support this use case as well.

[Key rotation and versioning](https://learn.microsoft.com/en-us/azure/key-vault/keys/secure-keys#key-rotation-and-versioning)

test/e2e/kms_key_rotation.go
```diff
@@ -0,0 +1,196 @@
...
+			Expect(fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged).NotTo(BeNil(), "fetched cluster CustomerManaged was nil")
+			Expect(fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged.Kms).NotTo(BeNil(), "fetched cluster Kms was nil")
+			Expect(fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey).NotTo(BeNil(), "fetched cluster ActiveKey was nil")
+			Expect(fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version).NotTo(BeNil(), "fetched cluster key Version was nil")
+			Expect(*fetchedCluster.Properties.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version).To(Equal(newKeyVersion),
+				"cluster should reference the new KMS key version after round-trip GET")
+
+			By("verifying StorageVersionMigration succeeded for re-encryption")
+			err = verifiers.VerifyStorageVersionMigrationSucceeded().Verify(ctx, adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "all StorageVersionMigration resources should reach Succeeded state after KMS key rotation")
```
mbukatov, line 192: We should also disable the old key and verify that cluster is still fully operational when we are sure that the rotation is fully complete.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2026-06-30-preview/openapi.json
```diff
@@ -0,0 +1,3862 @@
+{
```
mbukatov, line 1: Don't we need the API changes to be merged before the feature itself?

mbukatov (commented):
Providing few comments wrt E2E test for the key rotation.

That said, I don't understand why this test code is not part of https://github.com/Azure/ARO-HCP/pull/5629/ and at the same time, this E2E test contain updates of the API spec.

test/e2e/kms_key_rotation.go
ahitacat: Asked this to @mbukatov, he suggested to keep this separated. The key rotation requires the latest API. And we may want to consider add other test cases into key rotation (I have several in mind)

test/e2e/kms_key_rotation.go
ahitacat: Good idea, do you know how to test this? Maybe:

1. Rotate the key
2. On completion, disable old key 
3. Rotate again the key (second rotation)

That will be enough? I want to test consecutive rotations. To have more confidence and catch regression bugs once this is fixed https://redhat.atlassian.net/browse/OCPBUGS-98086

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2026-06-30-preview/openapi.json
ahitacat: Yes, these commits are from the API + RP change, so we can test it and have more confidence of merging those.

test/e2e/kms_key_rotation.go
ahitacat: We don't support auto-rotation. Key rotation should be something that the customer starts, as per how hypershift requires the version during creation o update the new key version.

This should require any new permission. However, the test is failing because of lack of permissions.

```
fail [github.com/Azure/ARO-HCP/test/e2e/kms_key_rotation.go:124]: failed to create new key version (rotation)
Unexpected error:
    <*exported.ResponseError | 0xc001301830>: 
    POST https://cust-kv-3t5qlposar5yi.vault.azure.net/keys/etcd-data-kms-encryption-key/create
    RESPONSE 403: 403 Forbidden
    ERROR CODE: Forbidden
    {
      "error": {
        "code": "Forbidden",
        "message": "Caller is not authorized to perform action on resource.\r\nIf role assignments, deny assignments or role definitions were changed recently, please observe propagation time.\r\nCaller: appid=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;oid=c209f8df-52ae-48fb-98ea-380f58b04652;iss=https://sts.windows.net/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/\r\nAction: 'Microsoft.KeyVault/vaults/keys/create/action'\r\nResource: '/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourcegroups/kms-key-rotate-bdhzsf/providers/microsoft.keyvault/vaults/cust-kv-3t5qlposar5yi/keys/etcd-data-kms-encryption-key'\r\nAssignment: (not found)\r\nDenyAssignmentId: null\r\nDecisionReason: null \r\nVault: cust-kv-3t5qlposar5yi;location=westus3\r\n",
        "innererror": {
          "code": "ForbiddenByRbac"
        }
      }
    }
```

In my local setup I had configure the

ahitacat:
Closed in favor of this https://github.com/Azure/ARO-HCP/pull/5629

PR #5971, opened 2026-07-08
Revert "feat: support FIPS mode via feature flags & cluster tags"

raelga:
Reverts Azure/ARO-HCP#5333

### Why

The new FIPS E2E test added in #5333 is failing the **Prod** and **Stage** periodic `e2e-parallel` jobs.

**What's failing**

`FIPS Mode Support with ExperimentalReleaseFeatures AFEC registered` (both specs) in the persistent e2e suite:

```
RESPONSE 400: 400 Bad Request
Invalid value: "aro-hcp.experimental.cluster.fips-enabled": unrecognized experimental tag
target: tags[aro-hcp.experimental.cluster.fips-enabled]
```

Prod periodic `periodic-prod-e2e-parallel` — 3 consecutive FAILURE runs:
- [2074674924374462464](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-Azure-ARO-HCP-main-periodic-prod-e2e-parallel/2074674924374462464) (2026-07-08)
- [2074312532201312256](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-Azure-ARO-HCP-main-periodic-prod-e2e-parallel/2074312532201312256) (2026-07-07)
- [2073950128078589952](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-Azure-ARO-HCP-main-periodic-prod-e2e-parallel/2073950128078589952) (2026-07-06)

Stage periodic `periodic-stage-e2e-parallel` — [2074674924420599808](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-Azure-ARO-HCP-main-periodic-stage-e2e-parallel/2074674924420599808) (2026-07-08), same FIPS spec failing.

**Root cause**

The test (`test/e2e/cluster_fips_mode.go`) is labelled `labels.RequireNothing`, so it runs in **every** environment, and it was registered into the `prod_parallel`, `stage_parallel` and `integration_parallel` suites (the `zz_fixture_TestMainListSuitesForEachSuite_*` files in #5333). But it depends on the experimental FIPS feature — the `microsoft.redhatopenshift/experimentalreleasefeatures` AFEC and the `aro-hcp.experimental.cluster.fips-enabled` cluster tag — which is **not enabled in Prod or Stage**. Cluster creation therefore fails with `InvalidRequestContent: unrecognized experimental tag` (cluster_fips_mode.go:84). Where creation gets further, the verifier then fails to launch its check pod (`test/util/verifiers/fips.go`): `pods "fips-check-" is forbidden: serviceaccount "default" not found`.

In short: the test was added to the Prod/Stage/Integration suites before the feature flag was activated in those environments, so it can only fail there.

**Follow-up**

The feature and its E2E test can be re-introduced once (a) the experimental AFEC is enabled in the target environments, or (b) the test is gated to only run where the flag is active (an environment label rather than `RequireNothing`) and is demonstrated green in Integration, Stage and Prod first.


raelga:
/hold

As discussed, we are going to wait for the Canary region to confirm the status of this feature before reverting.

raelga:
Prod E2E is passing this test, with the race condition flakiness that will be resolved in https://github.com/Azure/ARO-HCP/pull/5973.

/close

PR #5974, opened 2026-07-08
proposal timeout on operation controllers

miguelsorianod:
This PR builds on top of https://github.com/Azure/ARO-HCP/pull/5593 and adds a commit that contains a proposal on adding a timeout in the operation cluster update controller that ends up marking the operation as failed if after an hour the operation is not in a terminal state


backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -114,46 +144,98 @@ func (c *operationClusterUpdate) SynchronizeOperation(ctx context.Context, key c
...
 	return nil
 }
 
-func (c *operationClusterUpdate) determineOperationState(ctx context.Context, operation *api.Operation) (*operationState, error) {
+func (c *operationClusterUpdate) determineOperationState(ctx context.Context, operation *api.Operation, existingCluster *api.HCPOpenShiftCluster) (*operationState, error) {
 	logger := utils.LoggerFromContext(ctx)
+
+	// Fail fast on operation timeout before calling downstream checks. If the operation has timed out, we
+	// return the timeout state instead of the other states.
+	timeoutState, err := c.operationTimeoutOperationState(operation)
```
miguelsorianod, line 196: I considered having this as another operationState calculation with the others like desiredVersionResolutionOperationState, but I decided to have it intentionally separate and before those: if we determine there's a timeout it's not worth calculating the other states as we know we will fail anyway.

backend/pkg/controllers/operationcontrollers/operation_cluster_update.go
```diff
@@ -171,19 +253,29 @@ func (c *operationClusterUpdate) determineOperationState(ctx context.Context, op
...
-func (c *operationClusterUpdate) desiredVersionResolutionOperationState(ctx context.Context, operation *api.Operation) (*operationState, error) {
-	existingCluster, err := c.resourcesDBClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).Get(ctx, operation.ExternalID.Name)
-	if err != nil {
-		return nil, utils.TrackError(err)
+// operationTimeoutOperationState checks whether the cluster update operation
+// has exceeded clusterUpdateOperationTimeout. Returns (nil, nil) while the
+// operation is still within the deadline, (Failed, nil) once timed out, or an
+// error if StartTime is unset. Called before downstream status checks so a
+// timed-out operation is marked as Failed and not evaluated further in that case.
+func (c *operationClusterUpdate) operationTimeoutOperationState(operation *api.Operation) (*operationState, error) {
```
miguelsorianod, line 265: I considered returning "Succeeded" operation state if there's no timeout but semantically didn't seem correct so we return nil instead to signal "no timeout yet"

davidffrench:
Hi @miguelsorianod — this PR has had no activity for 55 days. Is it still valid and something you plan to continue, or should it be closed? Please let me know. Thanks!

PR #5977, opened 2026-07-08
hcpctl: snapshot: handle batches

stevekuznetsov:


tooling/hcpctl/pkg/snapshot/queries/maestro/mgmtAuditLogs/query.kql
```diff
@@ -3,9 +3,6 @@ let bundleResources = bundleMap
 | summarize groupVersion = take_any(groupVersion), resourceKind = take_any(resourceKind), label = take_any(label) by resourceNamespace, resourceName;
 cluster('{{ .ClusterURI }}').database('{{ .ServiceDatabase }}').table('aksEvents')
 | where timestamp between ({{ kqlDatetime .PhaseStartTime }} .. {{ kqlDatetime .PhaseEndTime }})
-{{- if and .ServiceClusterName .ManagementClusterName }}
-| where cluster in ('{{ .ServiceClusterName }}', '{{ .ManagementClusterName }}')
-{{- end }}
```
deads2k, line 8: I'm not clear on why we're removing this.

stevekuznetsov: This database had no `cluster` column, so we never ran this query. The bundle IDs we're looking up are sufficiently precise (random cluster ID inside, etc) that not subdividing all e2e audit logs by cluster is unlikely to bring up false positives.

PR #5984, opened 2026-07-09
fix: correct off-by-one error in GetInstallVersionForZStreamUpgrade

brucebarrera:
### Jira

Fixes [ARO-27955](https://redhat.atlassian.net/jira/software/c/projects/ARO/boards/11822?selectedIssue=[ARO-27955](https://redhat.atlassian.net/browse/ARO-27955))

### What

  `GetInstallVersionForZStreamUpgrade` had two bugs that combined to return the wrong install version for z-stream
  upgrade testing:

  1. **Wrong Cincinnati channel**: The function queried the *next* minor's channel (`candidate-4.22` when testing 4.21)
  as `maxVersion`. `GetUpgradeCandidatesInMaxMinorFromCincinnati` filters results to `maxVersion`'s minor, so it only
  found **y-stream** (cross-minor) upgrade targets — never z-stream (same-minor) ones. `hasUpgradePath=true` meant "can
  upgrade to the next minor," but the test waited for a same-minor z-stream upgrade that would never happen.

  2. **Off-by-one in the loop**: The loop started at `i=0` (the latest version in the minor). Since the latest version
  can't z-stream upgrade further, it should start at `i=1`.

  The fix:
  - Uses `candidates[0].String()` (current minor's latest) as `maxVersion` so upgrade target lookups find z-stream
  targets
  - Starts the loop at `i=1`, skipping the latest version which has no z-stream upgrade available
  - Removes the unused next-minor lookup (`GetLatestVersionInMinor` call)

### Why

The hosted control plane z-stream upgrade-only verification is failing across dev and stg. 4 contributing tests covering z-stream upgrade verification for 4.x hosted control plane versions. The upgrade path validation or version constraint logic may have a defect.

  All three z-stream upgrade e2e tests (4.19, 4.20, 4.21) timed out after 45 minutes in Prow CI with:
  VerifyHostedControlPlaneZStreamUpgradeOnly(initial=X.Y.Z) failed:
  no version in clusterversion/version status.history is greater than initial "X.Y.Z"
  The function returned the latest version in each minor (4.19.37, 4.20.29, 4.21.0) with `hasUpgradePath=true`, but no
  newer z-stream version existed within the same minor to upgrade to.

### Testing

 The fix will be automatically tested by CI when:
- Prow CI runs the E2E tests
- The `VerifyHostedControlPlaneZStreamUpgradeOnly` test executes for 4.19, 4.20, 4.21, 4.22  
- Monitor the results at: https://prow.ci.openshift.org/?repo=Azure%2FARO-HCP

### PR Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/test/util/framework/version_helper.go
+++ b/test/util/framework/version_helper.go
@@ -122,11 +122,12 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
 	// upgrade targets, not y-stream (next-minor) ones.
 	maxVersion := candidates[0].String()
 
-	// When candidates[i] has z-stream upgrade targets, install candidates[i+1] (the
-	// version just before it). This ensures the upgrade target is candidates[i] — a
-	// well-established version — rather than candidates[0] which may be too freshly
-	// released for HyperShift to pick up.
-	for i := 0; i < len(candidates)-1; i++ {
+	// Skip candidates[0] (the latest) — it has no same-minor upgrade targets since
+	// maxVersion equals it. For each remaining candidates[i] that has z-stream upgrade
+	// targets, install candidates[i+1] (the version just before it). This ensures the
+	// upgrade target is candidates[i] — a well-established version — rather than
+	// candidates[0] which may be too freshly released for HyperShift to pick up.
+	for i := 1; i < len(candidates)-1; i++ {
 		upgradeTargets, err := GetUpgradeCandidatesInMaxMinorFromCincinnati(ctx, channelGroup, maxVersion, candidates[i].String())
 		if err != nil {
 			return "", false, err
--- a/test/util/framework/version_helper.go
+++ b/test/util/framework/version_helper.go
@@ -101,14 +101,14 @@ func isRetryableVersionError(err error) bool {
 	return true
 }
 
+var getAllVersionsInMinor = GetAllVersionsInMinorStartingWith
+var getUpgradeCandidates = GetUpgradeCandidatesInMaxMinorFromCincinnati
+
 // GetInstallVersionForZStreamUpgrade returns the version to install the cluster with when testing
 // a z-stream upgrade, and whether that version has an available z-stream upgrade path. It uses
 // configuredVersionID and queries Cincinnati for the given channelGroup (e.g. "candidate", "stable").
 // When no version with an upgrade path is found, it still returns the configured version so the
 // caller can install and optionally skip upgrade assertions.
-var getAllVersionsInMinor = GetAllVersionsInMinorStartingWith
-var getUpgradeCandidates = GetUpgradeCandidatesInMaxMinorFromCincinnati
-
 func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string, configuredVersionID string) (installVersion string, hasUpgradePath bool, err error) {
 	candidates, err := getAllVersionsInMinor(ctx, channelGroup, configuredVersionID)
 	if err != nil {
--- a/test/util/framework/version_helper.go
+++ b/test/util/framework/version_helper.go
@@ -134,7 +134,7 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
 			return "", false, err
 		}
 		if len(upgradeTargets) > 0 {
-			return candidates[i+1].String(), true, nil
+			return candidates[i].String(), true, nil
 		}
 	}
 	return candidates[0].String(), false, nil
--- a/test/util/framework/version_helper.go
+++ b/test/util/framework/version_helper.go
@@ -122,15 +122,17 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
 	// upgrade targets, not y-stream (next-minor) ones.
 	maxVersion := candidates[0].String()
 
-	// Start from candidates[1] (second-latest) because candidates[0] is already the
-	// latest in the minor and cannot z-stream upgrade further.
-	for i := 1; i < len(candidates); i++ {
+	// When candidates[i] has z-stream upgrade targets, install candidates[i+1] (the
+	// version just before it). This ensures the upgrade target is candidates[i] — a
+	// well-established version — rather than candidates[0] which may be too freshly
+	// released for HyperShift to pick up.
+	for i := 0; i < len(candidates)-1; i++ {
 		upgradeTargets, err := GetUpgradeCandidatesInMaxMinorFromCincinnati(ctx, channelGroup, maxVersion, candidates[i].String())
 		if err != nil {
 			return "", false, err
 		}
 		if len(upgradeTargets) > 0 {
-			return candidates[i].String(), true, nil
+			return candidates[i+1].String(), true, nil
 		}
 	}
 	return candidates[0].String(), false, nil
--- a/test/util/framework/version_helper.go
+++ b/test/util/framework/version_helper.go
@@ -118,17 +118,13 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
 		return candidates[0].String(), false, nil
 	}
 
-	nextMinorStr := fmt.Sprintf("%d.%d", candidates[0].Major, candidates[0].Minor+1)
-	maxVersion, err := GetLatestVersionInMinor(ctx, channelGroup, nextMinorStr)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


test/util/framework/version_helper.go
```diff
@@ -118,23 +118,19 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
 		return candidates[0].String(), false, nil
 	}
 
-	nextMinorStr := fmt.Sprintf("%d.%d", candidates[0].Major, candidates[0].Minor+1)
-	maxVersion, err := GetLatestVersionInMinor(ctx, channelGroup, nextMinorStr)
```
machi1990, line 122: With the removal of this call;

How do we ensure that the upgrade path to the next minor is maintained when picking an install candidate?

test/util/framework/version_helper.go
```diff
@@ -118,23 +118,19 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
...
+	for i := 1; i < len(candidates); i++ {
 		upgradeTargets, err := GetUpgradeCandidatesInMaxMinorFromCincinnati(ctx, channelGroup, maxVersion, candidates[i].String())
 		if err != nil {
 			return "", false, err
 		}
 		if len(upgradeTargets) > 0 {
-			return candidates[i+1].String(), true, nil
+			return candidates[i].String(), true, nil
 		}
 	}
```
machi1990: I am trying to be sure I understand the issue with the code that was present and I'd like to understand what's the problematic case that is solved by the change? Can you share an example of the observed / potential failure, and how this PR is fixing it?

test/util/framework/version_helper.go
```diff
@@ -118,16 +118,14 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
...
-
+	// Use the current minor's latest as maxVersion so we find z-stream (same-minor)
+	// upgrade targets, not y-stream (next-minor) ones.
+	maxVersion := candidates[0].String()
+
+	// When candidates[i] has z-stream upgrade targets, install candidates[i+1] (the
+	// version just before it). This ensures the upgrade target is candidates[i] — a
+	// well-established version — rather than candidates[0] which may be too freshly
+	// released for HyperShift to pick up.
 	for i := 0; i < len(candidates)-1; i++ {
```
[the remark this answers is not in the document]
brucebarrera: addressed

test/util/framework/version_helper.go
```diff
@@ -118,17 +118,16 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
...
+	// Use the current minor's latest as maxVersion so we find z-stream (same-minor)
+	// upgrade targets, not y-stream (next-minor) ones.
+	maxVersion := candidates[0].String()
+
+	// Skip candidates[0] (the latest) — it has no same-minor upgrade targets since
+	// maxVersion equals it. For each remaining candidates[i] that has z-stream upgrade
+	// targets, install candidates[i+1] (the version just before it). This ensures the
+	// upgrade target is candidates[i] — a well-established version — rather than
+	// candidates[0] which may be too freshly released for HyperShift to pick up.
+	for i := 1; i < len(candidates)-1; i++ {
```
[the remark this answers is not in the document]
brucebarrera: Copilot's feedback is a false positive. Here's why:                                                                

When len(candidates) == 2, the only possible upgrade path is candidates[1] → candidates[0]. But candidates[0] is the absolute latest version — the one that may be too freshly released for HyperShift to pick up. That's exactly the scenario this PR exists to avoid.

Even if we could check candidates[1], we'd want to return candidates[i+1] = candidates[2], which doesn't exist — that's why the loop bound is i < len(candidates)-1.

The original code on main behaves identically for len == 2:

for i := 0; i < len(candidates)-1; i++   // i=0: queries candidates[0], always empty → loop ends
return candidates[0], false               // same result

Returning hasUpgradePath=false for a minor with only 2 patch releases is correct — there's no safe version to install
where the upgrade target is a well-established (non-latest) version. The test sees false and skips the upgrade
assertion, which is the right thing to do.

swiencki:
Consider adding a focused unit test for the helper so the same-minor and candidates[i+1] behavior stays locked.

test/util/framework/version_helper.go
```diff
@@ -106,8 +106,11 @@ func isRetryableVersionError(err error) bool {
 // configuredVersionID and queries Cincinnati for the given channelGroup (e.g. "candidate", "stable").
 // When no version with an upgrade path is found, it still returns the configured version so the
 // caller can install and optionally skip upgrade assertions.
+var getAllVersionsInMinor = GetAllVersionsInMinorStartingWith
+var getUpgradeCandidates = GetUpgradeCandidatesInMaxMinorFromCincinnati
+
 func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string, configuredVersionID string) (installVersion string, hasUpgradePath bool, err error) {
```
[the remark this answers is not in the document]
brucebarrera: The file is correct — the doc comment is on lines 107-111, directly above the func on line 112. The var declarations
 are on lines 104-105 above it, with a blank line separating them. Copilot is giving stale feedback from the previous revision.

brucebarrera:
> Consider adding a focused unit test for the helper so the same-minor and candidates[i+1] behavior stays locked.

@swiencki the tests have been added.

test/util/framework/version_helper.go
brucebarrera: We don't need to — this function is for z-stream (same-minor) upgrades only. The removed code was actually counterproductive: GetUpgradeCandidatesInMaxMinorFromCincinnati builds its Cincinnati channel from maxVersion's  major.minor, so when maxVersion was from the next minor (e.g. 4.20.x), it queried candidate-4.20 instead of  candidate-4.19 and returned y-stream targets, not z-stream ones. Using candidates[0] (same-minor latest) queries the  correct channel and finds actual z-stream upgrade paths.

test/util/framework/version_helper.go
brucebarrera: Concrete example from the CI failure that prompted this PR:

Minor 4.19 had these candidates (descending): 4.19.39, 4.19.38, 4.19.37, ...

Old code (for i := 0, returning candidates[i]): The loop started at i=0, queried candidates[0] (4.19.39) — which always returned empty (it's the latest, nothing to upgrade to within the same minor). Then at i=1, candidates[1]  (4.19.38) had upgrade targets, so it returned candidates[1] = 4.19.38 as the install version. The only z-stream  upgrade target from 4.19.38 was 4.19.39 — the absolute latest. But 4.19.39 was freshly released and not yet available  in HyperShift/ACR, so the auto-upgrade never triggered. The cluster sat at 4.19.38 "Partial" until the test timed out.

New code (for i := 1, returning candidates[i+1]): Skips candidates[0] (useless query). At i=1, candidates[1] (4.19.38)  has upgrade targets, so it returns candidates[i+1] = 4.19.37 as the install version. Now the upgrade target is  4.19.38 — a well-established version that HyperShift reliably picks up. The auto-upgrade succeeds.

The key shift: install one version older so the upgrade target is the second-latest (established), not the absolute  latest (possibly too new).

adalrsjr1:
/lgtm
looks sane to me

DrakeSeifert:
@brucebarrera can you revert this please?

DrakeSeifert:
nvm I did it here: https://github.com/Azure/ARO-HCP/pull/6275

test/util/framework/version_helper.go
machi1990: Hmm @brucebarrera take a look here https://github.com/brucebarrera/ARO-HCP/blob/4a922c62dee8806f0db9f0acbecb69d42f09066a/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go#L517 on how the next minor existence is important to z-stream upgrade. 

The test is setup in a way that it takes the next minor existence into consideration so that it's sure that the controller will be able to work with it and select a new desired version for the upgrade trigger.

test/util/framework/version_helper.go
machi1990: > Concrete example from the CI failure that prompted this PR:

Thanks, a kind ask for next time to remember to  also include a prow link so that myself as a reviewer can walk along the same analysis path.

PR #5991, opened 2026-07-09
fix: use underlay inventory for service alerts

mmazur:
https://redhat.atlassian.net/browse/AROSLSRE-1444

## Summary
- Switch services-only Prometheus, Arobit, and MISE absence alerts to use `underlay_clusters` as the desired-cluster source of truth.
- Add Sev 3 underlay inventory integrity alerts for disappeared entries and too-small or absent inventory in the SL services queue.
- Add `docs/alerts/underlay_clusters.md` as the runbook for the new inventory alerts.
- Update the corresponding YAML rule tests.
- Leave generated Bicep unchanged for now.

## Testing
- Validated touched observability alert tests with the prometheus-rules generator using temporary output outside the repo.
- Validated touched MISE alert test with the prometheus-rules generator using temporary output outside the repo.

### DIFF
--- a/docs/alerts/underlay_clusters.md
+++ b/docs/alerts/underlay_clusters.md
@@ -1,0 +1,42 @@
+# underlay_clusters alerts
+
+The `underlay_clusters` metric is emitted by the static recording rule defined in
+`dev-infrastructure/modules/metrics/underlay-clusters-metric.bicep`.
+Each service and management cluster deployment instantiates that module with its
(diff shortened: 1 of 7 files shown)


swiencki (commented):
observability/alerts/prometheus-prometheusRule.yaml - `UnderlayClusterInventoryTooSmall` will fire as a permanent false positive in INT/dev environments that legitimately have 1-2 clusters.

PR #5993, opened 2026-07-09
fix: weight PrometheusUptime by sample count

mmazur:
Jira: https://redhat.atlassian.net/browse/AROSLSRE-1446

## Why

`PrometheusUptime` is intended to measure whether Prometheus self-`up` samples are healthy for at least 95% of the last 24h, while `PrometheusUptimeSampleCount` separately covers missing samples / ingestion gaps.

The current expression calculates `avg_over_time(up[1d])` independently for each target series and then averages those per-series results equally:

```promql
avg by (job, namespace, cluster) (
  avg_over_time(up{job="prometheus/prometheus",namespace="prometheus"}[1d])
) < 0.95
```

That means each target identity gets equal weight regardless of how many samples it contributed. **For example, one series with 5 minutes of `up=0` and another series with 24 hours of `up=1` are averaged as `(0 + 1) / 2 = 50%` availability, even though the failed series only contributed 5 minutes of data.**

## What

This changes `PrometheusUptime` to weight by actual present sample count:

```promql
sum_over_time(up[1d]) / count_over_time(up[1d]) < 0.95
```

aggregated by `job`, `namespace`, and `cluster`.

With this shape:
- `up=0` samples still count as unhealthy.
- Missing samples do not count here; `PrometheusUptimeSampleCount` owns that signal.
- Short-lived stale target identities only contribute their actual sample count instead of getting equal weight with full-day series.

Also updates the Prometheus alert runbook wording and adds a regression test for a stale short-lived failed target series.

## Verification

- Corrected alert doesn't go below 99.95% across all regions in the past 30 days.
- Current alert is firing in multiple regions due to the short timeseries of only 0s having the same weight as 24h series of ones.

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/generatedPrometheusAlertingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedPrometheusAlertingRules.bicep
@@ -66,19 +66,21 @@ Check the status of the Prometheus pods, service endpoints, and network connecti
         }
         annotations: {
           correlationId: 'PrometheusUptime/{{ $labels.cluster }}'
-          description: '''Prometheus has been unreachable for more than 5% of the time over the past 24 hours.
+          description: '''Fewer than 95% of present Prometheus self-up samples were healthy over the past 24 hours.
 This may indicate that the Prometheus server is down, experiencing network issues, or stuck in a crash loop.
+Missing samples are covered separately by PrometheusUptimeSampleCount.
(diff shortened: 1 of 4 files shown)


avollmer-redhat (approved):
Reviewed against PR standards and technical correctness. All standards met — clear title, linked ticket, excellent "Why" summary, consistent changes across all four files, and a targeted regression test for the stale short-lived series bug.

The `e2e-parallel` CI failure is unrelated to alerting rule changes.

/lgtm

PR #5998, opened 2026-07-09
fix: add region to AMW incident titles

mmazur:
## Why

Direct AMW metric alerts currently rely on the Azure Monitor/IcM default incident title format, which does not include the ARO HCP region. This makes incidents like AMW ingestion alerts harder to triage when the same alert exists in multiple regions and is not in line with how prom incidents look like.

## What

- Pass the required ARO HCP `region` value into the monitoring deployment.
- Set `IcM.Title` for all direct AMW metric alerts to `<region>: <alert name>`.
- Keep existing Azure metric alert resource names unchanged to avoid replacing alert resources just for title formatting.

## Validation

- `bicep build dev-infrastructure/templates/monitoring.bicep --stdout`


stevekuznetsov (approved):
/lgtm
/approve

PR #6004, opened 2026-07-09
feat(frontend): reject duplicate subnet, NSG, and MRG on cluster create

JakobGray:
### What

- Prefetch subscription clusters and node pools in newClusterAdmissionContext
- Add admitClusterManagedResourceGroupName, admitClusterSubnetResourceID, and admitClusterNetworkSecurityGroupResourceID (CREATE only)
- Add NSG same-subscription and not-in-MRG rules to static validation (parity with validateAroHcpClusterNetworkSecurityGroupResourceId)

This change closes the following gaps in validation during cluster create between CS and ARO-HCP:

- Duplicate MRG in subscription
- Duplicate subnet on another cluster
- Duplicate subnet on another cluster’s node pool
- Duplicate NSG on another cluster
- NSG in wrong subscription
- NSG in managed RG

These mirror checks done in CS before a cluster create is accepted:
- `validateAroHcpClusterNetworkSecurityGroupResourceId`
- `validateManagedResourceGroupUniqueness`
- `validateSubnetUniqueness`
- `validateNsgUniqueness`

These checks are best-effort. Race conditions are possible for clusters admitted at the same time.

### Why

Cluster Service no longer returns synchronous 400s for platform uniqueness checks after async CS create migration. Move those checks to the frontend so invalid cluster PUTs fail at admission time with the same error messages CS used in performSpecValidation.

### Testing

Added unit tests for new functions. Modified integration tests where new errors were occurring because they are now being caught by the admission validation. Tested against existing E2E tests around subnet/NSG/MRG reuse between clusters (`test/e2e/cluster_nsg_subnet_reuse.go` and `test/e2e/clusters_sharing_resgroup.go`) 

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/internal/admission/admit_cluster.go
+++ b/internal/admission/admit_cluster.go
@@ -53,6 +53,16 @@ type ClusterAdmissionContext struct {
 	// ClusterNodePools is the list of node pools belonging to the cluster, used
 	// for minor-version skew checks against the desired cluster version.
 	ClusterNodePools []ClusterAdmissionNodePool
+	// SubscriptionClusters lists cluster documents in the same subscription
+	// (not including the current cluster being admitted), used
+	// for cross-cluster platform resource uniqueness on CREATE.
+	// The list is empty on UPDATE.
+	SubscriptionClusters []*api.HCPOpenShiftCluster
+	// SubscriptionNodePools lists node pool documents under SubscriptionClusters,
+	// used to ensure a cluster subnet is not already assigned to another cluster's
+	// node pool on CREATE.
+	// The list is empty on UPDATE.
+	SubscriptionNodePools []*api.HCPOpenShiftClusterNodePool
 }
 
 // ClusterAdmissionNodePool is a single node pool plus its prefetched service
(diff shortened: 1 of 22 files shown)


internal/admission/admit_cluster.go
```diff
@@ -53,6 +53,15 @@ type ClusterAdmissionContext struct {
 	// ClusterNodePools is the list of node pools belonging to the cluster, used
 	// for minor-version skew checks against the desired cluster version.
 	ClusterNodePools []ClusterAdmissionNodePool
+	// SubscriptionClusters lists cluster documents in the same subscription, used
```
miguelsorianod: Document whether it includes the cluster being processed itself or not

internal/admission/admit_cluster.go
```diff
@@ -222,6 +231,110 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
 	errs := field.ErrorList{}
 
 	errs = append(errs, admitClusterVersionProfile(ctx, admissionContext, op, fldPath.Child("version"), &newObj.Version, safe.Field(oldObj, validation.ToClusterCustomerPropertiesVersion))...)
+	platformPath := fldPath.Child("platform")
```
miguelsorianod: Move all of this to admitClusterPlatform

internal/admission/admit_cluster.go
```diff
@@ -53,6 +53,15 @@ type ClusterAdmissionContext struct {
...
 	ClusterNodePools []ClusterAdmissionNodePool
+	// SubscriptionClusters lists cluster documents in the same subscription, used
+	// for cross-cluster platform resource uniqueness on CREATE.
+	// The list is empty on UPDATE.
+	SubscriptionClusters []*api.HCPOpenShiftCluster
+	// SubscriptionNodePools lists node pool documents under SubscriptionClusters,
+	// used to ensure a cluster subnet is not already assigned to another cluster's
+	// node pool on CREATE.
+	// The list is empty on UPDATE.
+	SubscriptionNodePools []*api.HCPOpenShiftClusterNodePool
```
miguelsorianod, line 65: Do we need to check this? don't we check that for nodepools the subnet must be part of the VNet of the cluster?

internal/admission/admit_cluster.go
```diff
@@ -222,6 +231,110 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
...
+}
+
+// admitClusterSubnetResourceID ensures that the subnet ID is not already in use by any other
+// cluster or node pool within the same subscription when creating a new cluster.
+//
+// Best-effort only: compares against SubscriptionClusters and SubscriptionNodePools
+// prefetched before admission runs. Concurrent creates (or a create racing with a
+// node pool create) using the same subnet can both succeed.
+func admitClusterSubnetResourceID(_ context.Context, admissionContext *ClusterAdmissionContext, op operation.Operation, fldPath *field.Path, newObj *api.CustomerPlatformProfile) field.ErrorList {
+	if op.Type != operation.Create || newObj.SubnetID == nil {
```
miguelsorianod: can newObj.SubnetID be nil at this point? if not, remove the check

internal/admission/admit_cluster.go
```diff
@@ -222,6 +231,110 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
...
+	return errs
+}
+
+// admitClusterManagedResourceGroupName ensures the managed resource group name
+// is unique within the subscription on CREATE.
+//
+// Best-effort only: compares against SubscriptionClusters prefetched before
+// admission runs. Concurrent creates with the same MRG name can both succeed.
+func admitClusterManagedResourceGroupName(_ context.Context, admissionContext *ClusterAdmissionContext, op operation.Operation, fldPath *field.Path, newObj *api.CustomerPlatformProfile) field.ErrorList {
+	if op.Type != operation.Create || len(newObj.ManagedResourceGroup) == 0 {
```
miguelsorianod: Can managedresourcegroup be nil at this point? if not, remove the check

internal/admission/admit_cluster.go
```diff
@@ -222,6 +231,110 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
...
+	return errs
+}
+
+// admitClusterNetworkSecurityGroupResourceID ensures that the network security group ID is not already in use by any other
+// cluster within the same subscription when creating a new cluster.
+//
+// Best-effort only: compares against SubscriptionClusters prefetched before
+// admission runs. Concurrent creates with the same NSG can both succeed.
+func admitClusterNetworkSecurityGroupResourceID(_ context.Context, admissionContext *ClusterAdmissionContext, op operation.Operation, fldPath *field.Path, newObj *api.CustomerPlatformProfile) field.ErrorList {
+	if op.Type != operation.Create || newObj.NetworkSecurityGroupID == nil {
```
miguelsorianod: can newObj.SubnetID be nil at this point? if not, remove the check

internal/admission/admit_cluster.go
JakobGray: This replicates the `ValidateSubnetNotUsedByAnotherClusterNodePools()` validation in CS. Node pool subnet must be in the same VNet as its parent cluster, but VNet sharing across clusters is not prevented

internal/admission/admit_cluster.go
JakobGray: In a regular end-to-end flow it can be expected that the field is set and caught further upstream. In unit or integration tests it could be that a cluster is defined that is not fully formed because it is isolating focus on something else. If the subnet isn't set it is safe to say there is no uniqueness concern and the validation can return early.

internal/admission/admit_cluster.go
JakobGray: Same as https://github.com/Azure/ARO-HCP/pull/6004#discussion_r3573047361

internal/admission/admit_cluster.go
JakobGray: Same as https://github.com/Azure/ARO-HCP/pull/6004#discussion_r3573047361

internal/admission/admit_cluster.go
JakobGray: Done

internal/admission/admit_cluster.go
miguelsorianod: We shouldn't condition the logic to unit tests/integration tests

internal/admission/admit_cluster.go
miguelsorianod: We shouldn't condition the logic to unit tests/integration tests

internal/admission/admit_cluster.go
miguelsorianod: We shouldn't condition the logic to unit tests/integration tests

internal/admission/admit_cluster.go
miguelsorianod: Regarding subnets, does the following summarize the behavior?

in CS:
* We allow reusing VNets between clusters
* We do not allow reusing Subnets between clusters
* We enforce that the node pool subnets must belong to the parent cluster VNet
* We allow reusing Subnets between Node Pools within the same cluster
* We do not allow reusing Subnets between Node Pools across clusters

internal/validation/validate_cluster.go
```diff
@@ -175,6 +175,8 @@ func validateResourceIDsAgainstClusterID(ctx context.Context, op operation.Opera
 	errs = append(errs, DifferentResourceGroupName(ctx, op, field.NewPath("customerProperties", "platform", "managedResourceGroup"), &newCluster.CustomerProperties.Platform.ManagedResourceGroup, nil, newCluster.ID.ResourceGroupName)...)
 	errs = append(errs, SameSubscription(ctx, op, field.NewPath("customerProperties", "platform", "subnetId"), newCluster.CustomerProperties.Platform.SubnetID, nil, newCluster.ID.SubscriptionID)...)
 	errs = append(errs, DifferentResourceGroupNameFromResourceID(ctx, op, field.NewPath("customerProperties", "platform", "subnetId"), newCluster.CustomerProperties.Platform.SubnetID, nil, newCluster.CustomerProperties.Platform.ManagedResourceGroup)...)
+	errs = append(errs, SameSubscription(ctx, op, field.NewPath("customerProperties", "platform", "networkSecurityGroupId"), newCluster.CustomerProperties.Platform.NetworkSecurityGroupID, nil, newCluster.ID.SubscriptionID)...)
```
miguelsorianod, line 182: For future readers, document in the validations that we can't relax the "same subscription" validation for subnets and nsgs, because relaxing them would break the admission where we list only the clusters/nodepools within the same subscription. If that is considered the other part would need to be fixed/solved

internal/admission/admit_cluster.go
JakobGray: That all sounds correct

internal/admission/admit_cluster.go
miguelsorianod: It seems we run admission even when we have validation failures. Because of this, this needs a check to avoid panicking. To be safe we can return an error if the assumption of the subnet being set is broken.

internal/api/enums.go
```diff
@@ -164,13 +164,17 @@ const (
...
 	EtcdDataEncryptionKeyManagementModeTypePlatformManaged EtcdDataEncryptionKeyManagementModeType = "PlatformManaged"
 )
 
 var (
 	ValidEtcdDataEncryptionKeyManagementModeType = sets.New[EtcdDataEncryptionKeyManagementModeType](
 		EtcdDataEncryptionKeyManagementModeTypeCustomerManaged,
-		EtcdDataEncryptionKeyManagementModeTypePlatformManaged,
+		// TODO: re-enable once platform-managed etcd encryption is supported.
+		// EtcdDataEncryptionKeyManagementModeTypePlatformManaged,
 	)
```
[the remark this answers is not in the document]
miguelsorianod: This should be fine. Even though we accept it at the frontend level, CS returns an error when that's the case. We are simply moving the error to the frontend layer

internal/api/enums.go
```diff
@@ -164,13 +164,17 @@ const (
 	// EtcdDataEncryptionKeyManagementModeTypeCustomerManaged - Customer managed encryption key management mode type.
 	EtcdDataEncryptionKeyManagementModeTypeCustomerManaged EtcdDataEncryptionKeyManagementModeType = "CustomerManaged"
 	// EtcdDataEncryptionKeyManagementModeTypePlatformManaged - Platform managed encryption key management mode type.
+	// Not currently supported by Cluster Service; left defined so EnsureDefaults / Cosmos defaults keep
```
miguelsorianod: no need to mention "by Cluster Service". Our service in general doesn't support it.

internal/validation/validate_cluster.go
```diff
@@ -734,6 +734,44 @@ func validateUserAssignedIdentitiesProfile(ctx context.Context, op operation.Ope
...
+				identityPath,
+				identity.String(),
+				fmt.Sprintf("managed identity with resource id '%s' must be unique within the cluster", identity.String()),
+			))
+			return
+		}
+		observed[key] = identityPath
+	}
+
+	for operatorName, identity := range newObj.ControlPlaneOperators {
```
miguelsorianod: Without sorting, are the tests deterministic?

internal/admission/admit_cluster_test.go
```diff
@@ -684,3 +684,167 @@ func TestAdmitCluster_Update(t *testing.T) {
...
+			subscriptionClusters: []*api.HCPOpenShiftCluster{
+				makeCluster("existing-cluster", "shared-mrg", subnetID, nsgID),
+			},
+			newCluster: makeCluster("new-cluster", "shared-mrg", otherSubnetID, otherNsgID),
+			expectErrors: []utils.ExpectedError{
+				{FieldPath: "properties.platform.managedResourceGroup", Message: "please provide a unique managed resource group name"},
+			},
+		},
+		{
+			name: "create rejects duplicate subnet used by node pool",
```
miguelsorianod, line 778: Is this correct? We should allow using the same subnet by another nodepool if the nodepool belongs to the same cluster

miguelsorianod: By definition that should always be the case during cluster creation, which is the only point in time for now where the admission is executed, but the test might be inconsistent

miguelsorianod: I mixed new-cluster and existing-cluster. That seems good. Maybe the test name can be a bit more accurate:
create rejects duplicate subnet in use by another cluster's nodepool

JakobGray:
Work merged in https://github.com/Azure/ARO-HCP/pull/6121

miguelsorianod:

Closing this.

PR #6025, opened 2026-07-10
fix: make frontend cluster create idempotent to prevent CS duplicate collisions

cssjr:
## Summary

- Fixes [ARO-27951](https://issues.redhat.com/browse/ARO-27951): "InvalidRequestContent — Duplicate ARO-HCP cluster name collision" hitting ~7.2% of CI E2E runs
- Adds idempotency check to the frontend `createHCPCluster()` path before calling `PostCluster`, preventing split-brain when Cosmos write fails after CS create succeeds
- Extracts `FindClusterByAzureInfo` into a shared `ocm` package function, deduplicating logic between frontend and backend controller

## Root Cause

The frontend's `createHCPCluster()` performs a two-phase write that is **not atomic**:

1. **Phase 1**: `PostCluster` to Cluster Service (CS) — creates the cluster record
2. **Phase 2**: Cosmos DB transactional batch — creates cluster + operation documents

If Phase 2 fails (transient Cosmos error), the system enters a **split-brain state**: CS has the cluster, Cosmos does not. When the Azure SDK automatically retries the failed 500:

1. `CreateOrUpdateHCPCluster` checks Cosmos → not found → routes to `createHCPCluster`
2. `PostCluster` to CS → CS rejects: "Duplicate ARO-HCP cluster name" (HTTP 400)
3. Error mapped to `InvalidRequestContent` → test/request fails

This was confirmed by an identical Stage incident documented in [ARO-17939](https://issues.redhat.com/browse/ARO-17939) (comment by Mike Gahagan, 2026-05-20).

## Fix

Before calling `PostCluster`, the frontend now searches CS for an existing cluster with matching Azure metadata (subscription, resource group, resource name, tenant, managed resource group) via `ocm.FindClusterByAzureInfo`. If an orphaned CS cluster from a prior failed attempt is found, it is reused instead of creating a duplicate.

The backend's `ClusterClusterServiceCreate` controller already had this idempotency pattern — this PR extracts it into a shared function and applies it to the frontend as well.

## Test plan

- [x] `go build` for all affected packages (internal/ocm, frontend, backend)
- [x] `go test ./internal/ocm/...` — passes
- [x] `go test ./frontend/...` — passes
- [x] `go test ./backend/...` — all 20+ controller test packages pass
- [x] `make lint` — passes
- [x] Integration tests (`make test-integration`) 
- [ ] E2E validation: monitor `branch-ci-Azure-ARO-HCP-main-e2e-integration-e2e-parallel` success rate after merge

### DIFF
--- a/frontend/pkg/frontend/cluster.go
+++ b/frontend/pkg/frontend/cluster.go
@@ -369,24 +369,46 @@ func (f *Frontend) createHCPCluster(writer http.ResponseWriter, request *http.Re
 		tenantID = *subscription.Properties.TenantId
 	}
 
-	initialClusterProperties := map[string]string{}
-	if len(f.clusterServiceProvisionShard) != 0 {
-		initialClusterProperties[ocm.CSPropertyProvisionShardID] = f.clusterServiceProvisionShard
-	}
-	if f.clusterServiceNoopProvision {
-		initialClusterProperties[ocm.CSPropertyNoopProvision] = ocm.CSPropertyEnabled
-	}
-	if f.clusterServiceNoopDeprovision {
-		initialClusterProperties[ocm.CSPropertyNoopDeprovision] = ocm.CSPropertyEnabled
-	}
-	newClusterServiceClusterBuilder, newClusterServiceAutoscalerBuilder, err := ocm.BuildCSCluster(newInternalCluster.ID, tenantID, newInternalCluster, initialClusterProperties, nil, nil)
-	if err != nil {
-		return utils.TrackError(err)
-	}
 	logger.Info(fmt.Sprintf("creating resource %s", newInternalCluster.ID))
-	resultingClusterServiceCluster, err := f.clusterServiceClient.PostCluster(ctx, newClusterServiceClusterBuilder, newClusterServiceAutoscalerBuilder)
+
+	// Check if CS already has a cluster with matching Azure metadata. This
+	// handles the retry-after-partial-failure scenario: a prior attempt may
+	// have created the CS cluster but failed to write the Cosmos documents,
+	// leaving an orphan in CS that would cause PostCluster to reject with
+	// "Duplicate ARO-HCP cluster name" (see ARO-27951).
+	resultingClusterServiceCluster, err := ocm.FindClusterByAzureInfo(
+		ctx, f.clusterServiceClient,
+		newInternalCluster.ID.SubscriptionID,
+		newInternalCluster.ID.ResourceGroupName,
+		newInternalCluster.Name,
+		tenantID,
+		newInternalCluster.CustomerProperties.Platform.ManagedResourceGroup,
+	)
 	if err != nil {
 		return utils.TrackError(err)
+	}
+
+	if resultingClusterServiceCluster == nil {
+		initialClusterProperties := map[string]string{}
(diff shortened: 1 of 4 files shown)


swiencki (commented):
backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go - A second cluster in a different bucket asserting no manifest is created would prove the bucketing logic actually excludes foreign work.

frontend/pkg/frontend/cluster.go
```diff
@@ -369,26 +369,48 @@ func (f *Frontend) createHCPCluster(writer http.ResponseWriter, request *http.Re
...
-		initialClusterProperties[ocm.CSPropertyNoopProvision] = ocm.CSPropertyEnabled
-	}
-	if f.clusterServiceNoopDeprovision {
-		initialClusterProperties[ocm.CSPropertyNoopDeprovision] = ocm.CSPropertyEnabled
-	}
-	newClusterServiceClusterBuilder, newClusterServiceAutoscalerBuilder, err := ocm.BuildCSCluster(newInternalCluster.ID, tenantID, newInternalCluster, initialClusterProperties, nil, nil)
-	if err != nil {
-		return utils.TrackError(err)
-	}
 	logger.Info(fmt.Sprintf("creating resource %s", newInternalCluster.ID))
```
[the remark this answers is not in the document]
cssjr: Addressed in latest

cssjr:
> backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go - A second cluster in a different bucket asserting no manifest is created would prove the bucketing logic actually excludes foreign work.

@swiencki good call out.  Added test "found among non-matching clusters" to address.

deads2k:
We are removing all cluster-service calls from the frontend. The PR doing so has tested green and is tagged for merge: https://github.com/Azure/ARO-HCP/pull/6121

/close

frontend/pkg/frontend/cluster.go
```diff
@@ -369,24 +369,45 @@ func (f *Frontend) createHCPCluster(writer http.ResponseWriter, request *http.Re
...
+		newClusterServiceClusterBuilder, newClusterServiceAutoscalerBuilder, err := ocm.BuildCSCluster(newInternalCluster.ID, tenantID, newInternalCluster, initialClusterProperties, nil, nil)
+		if err != nil {
+			return utils.TrackError(err)
+		}
+		resultingClusterServiceCluster, err = f.clusterServiceClient.PostCluster(ctx, newClusterServiceClusterBuilder, newClusterServiceAutoscalerBuilder)
+		if err != nil {
+			return utils.TrackError(err)
+		}
+	} else {
+		logger.Info("Reusing existing Cluster Service cluster found by Azure metadata", "csClusterHREF", resultingClusterServiceCluster.HREF())
```
deads2k, line 410: we don't need this PR anymore, but I am curious what you think should happen if hte content in the existing cluster-service instance does not match the new create call.

cssjr: If there wasn't a 100% match, it was going to create a new one (not reuse the old one).  My understanding was a partial match was not duplicative so a new one could be created without conflict/idempotentcy issues.

PR #6026, opened 2026-07-10
feat(observability): hcp kube-apiserver availability alerts to SRE

SudoBrendan:
https://redhat.atlassian.net/browse/ARO-26202

### What

Re-introduce the alerts reverted by #5754

**Why this isn't a "revert the revert" PR:** The underlying structure of the repository has fundamentally changed - files moving, generators shifting, etc. In addition, the feedback provided was "the PR is just too big" - it would be more work (and more confusing to review) a full revert-the-revert. Here's my proposed alternative:

1. I did a `git checkout` of the slim number of files we actually care about for this PR so we have a byte-for-byte comparison without the overhead of a full revert (see Commit 1)
2. I have proposed changes in Commit 2/3 << what needs your review
3. I have generated output of our final results in Commit 4

### Why

While not our final or a perfect solution, SREs need to be alerted to HCP KAS outages. We can iterate towards a better state in future work, which will take alignment across many teams, even outside the ARO org, to get right - ref [RFE-9299](https://redhat.atlassian.net/browse/RFE-9299).

### Testing

- Promtool tests

Thank you to @swiencki for his contributions to this PR - many of the changes here were adopted from his work in #5765.

### PR Checklist

- [x] Self-reviewed the diff


dev-infrastructure/templates/monitoring.bicep
```diff
@@ -206,6 +206,15 @@ module rpAlerts '../modules/metrics/rp-rules.bicep' = {
...
 }
 
+module rpHcpAlerts '../modules/metrics/rp-hcp-rules.bicep' = {
+  name: 'rpHcpAlerts'
+  params: {
+    azureMonitoringWorkspaceId: hcpAzureMonitoringWorkspaceId
+    actionGroups: rpActionGroups
+    severityCeiling: alertSeverityCeiling
+  }
+}
```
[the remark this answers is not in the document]
SudoBrendan: this is confusing for humans too - I'll give you a pass. Queue is correct.

observability/alerts/UserJourneyKubeApiserverAvailabilityMonitor-prometheusRule-KSM.yaml
```diff
@@ -0,0 +1,103 @@
...
+          1 - (hostedClusterAPI_kubeapiserver_available:sli_sum_3d / hostedClusterAPI_kubeapiserver_available:sli_count_3d)
+          > (1 * (1 - 0.9995))
+          and on (name, namespace, _id, resource_id, cluster)
+          hostedClusterAPI_kubeapiserver_available:sli_count_3d > 130
+        )
+      for: 3h
+      labels:
+        long_window: 3d
+        short_window: 6h
+        severity: "3"
```
swiencki, line 93: The 3d/6h slow-burn tier uses severity "3" but per the ADR, slow-burn tiers should be severity "4" (ticket) since they represent degradation, not immediate customer impact.

observability/alerts/UserJourneyKubeApiserverAvailabilityMonitor-prometheusRule-KSM.yaml
```diff
@@ -0,0 +1,103 @@
...
+        )
+      for: 10m
+      labels:
+        long_window: 1h
+        short_window: 5m
+        severity: "3"
+        slo: kas-availability
+      annotations:
+        summary: "[HCPKASAvailableBurn] {{ $labels.cluster }} / {{ $labels.namespace }} (1h/5m)"
+        description: |
```
swiencki, line 35: The generator auto-extracts resource_id and _id from the description and appends them to the IcM title; since the summary already includes cluster and namespace, consider removing those two from the description to avoid duplicating them in the generated title, or accept the current title format if the extra context is useful for IcM triage.

observability/alerts/UserJourneyKubeApiserverAvailabilityMonitor-prometheusRule-KSM.yaml
SudoBrendan: In the long term - won't we want these to be 2/2/3? Medium burn exhausts in ~2 days. Slow in ~5d

swiencki: Second thought, since it's KAS I'm good with it being 2/2/3.

observability/alerts/UserJourneyKubeApiserverAvailabilityMonitor-prometheusRule-KSM.yaml
SudoBrendan: Yea - I don't see a great solution here, especially considering that Description (which seems like it should be the body of the Ticket) doesn't actually translate to anything right now in ICM. I think we _want_ it to be the ICM ticket body eventually, right? Today - all ticket bodies are blank.

Current generator logic will duplicate Description content to Title, and this change must be minimal to merge so I think a "fix" there is beyond the scope of this PR.

Which would we rather have?

1. We remove this content from Description (thus Title too). The only way SREs can see ResourceID of the ticket is to scroll all the way to the bottom for the Labels.
2. Leave as-is, accept the duplication (out of scope), SREs don't have to scroll past all comments for cluster identifiers, it's in the Title

FWIW I think my desired future state (out of scope here) for our team is:

1. Title has nothing to do with Description; completely independent
2. Description == body of the ICM

SudoBrendan:
/hold

Ensuring we verify this with alert-tester before merge.

observability/alerts/UserJourneyKubeApiserverAvailabilityMonitor-prometheusRule-KSM.yaml
```diff
@@ -0,0 +1,102 @@
...
+          hostedClusterAPI_kubeapiserver_available:sli_count_1h > 54
+        )
+      for: 10m
+      labels:
+        long_window: 1h
+        short_window: 5m
+        severity: "3"
+        slo: kas-availability
+      annotations:
+        summary: "[HCPKASAvailableBurn] {{ $labels.cluster }} / {{ $labels.namespace }} (1h/5m) resource_id:{{ $labels.resource_id }} _id:{{ $labels._id }}"
```
swiencki: `_id` is the HostedCluster `spec.clusterID`, so it should remain in the PromQL joins for series identity, but could we remove it and `cluster` from the responder-facing summary while retaining `resource_id`, `namespace`, and the burn window?

SudoBrendan: mm, I would like that.

The generator is limiting here, because it's enforcing other team's rules for us too. If we want bodies of tickets to read as they do currently:

```
Resource ID: {{ $labels.resource_id }}
Management Cluster: {{ $labels.cluster }}
Namespace: {{ $labels.namespace }}
```

then the generator is going to put cluster in the Summary regardless - [at the beginning](https://github.com/Azure/ARO-HCP/blob/cb0ee635a846e2ac65a48124a6e52022f66f27a0/tooling/prometheus-rules/internal/generator.go#L688), and also [somewhere else](https://github.com/Azure/ARO-HCP/blob/cb0ee635a846e2ac65a48124a6e52022f66f27a0/tooling/prometheus-rules/internal/generator.go#L542).

Do we want the cluster to appear as labeled metadata in the body of the ICM (rather than a random string in a title only)? If we do, it has to be in the title too (twice, yuck). I think the correct fix for this is changing the [forced cluster prefix](https://github.com/Azure/ARO-HCP/blob/cb0ee635a846e2ac65a48124a6e52022f66f27a0/tooling/prometheus-rules/internal/generator.go#L688), but I know that's going to mess up someone else's day.

It's a shame every team's rules are enforced in a single generator. Ideally - we could shape our own norms/values for our team in an independent one.

SudoBrendan: I went ahead and removed _id since that's not in the description anymore.

observability/grafana-dashboards/kas-monitor/kas_golden_signals.json
SudoBrendan, line 1: cc @swiencki for the other PR discussion we had on the TSG/UJ docs.

SudoBrendan:
/unhold

I believe we've met the original purpose of the hold - we got docs merged, we ran alert-tester against this update.

PR #73, opened 2024-04-24
Validation refactoring

mbarnes:
### What this PR does

These are miscellaneous tweaks and refactoring commits related to both field visibility enforcement and preflight.  I factored these "fluff" commits out so that subsequent pull requests will be shorter and easier to review.

Related Jiras:

- [ARO-6347: Validate metadata from api structs](https://issues.redhat.com/browse/ARO-6347)
- [ARO-6156: Create a Preflight Endpoint for ARO HCP](https://issues.redhat.com/browse/ARO-6156)

### Special notes for your reviewer

I recommend reviewing one commit at a time.

### Checklist

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [ ] Deployment: The deployment process was considered and addressed if required
- [ ] Testing: New code requires new unit tests.
- [ ] Documentation: Is the documentation updated? Either in the doc located in focus area, in the README or in the code itself.
- [ ] Customers: Is this change affecting customers? Is the release plan considered?

PR #6029, opened 2026-07-11
feat: add MI resource aware scheduler to e2e

mgahagan73:
https://redhat.atlassian.net/browse/ARO-26724

### What

1. **Add `labels.MIContainers(N)` to every E2E test spec** — replaces the qualitative `MIDemandHigh`/`MIDemandMedium` labels with a quantitative function that declares exact MI container demand per test (0, 1, 2, or 3). Applied to all 42 `It()` blocks and 4 `DescribeTable` blocks across 39 files.

2. **Add `verify-mi-containers` pre-merge CI check** — new AST-based Go tool (`hack/verify-mi-containers/`) that walks `test/e2e/` and rejects any `It()` or `DescribeTable()` missing an `MIContainers(N)` label, or where the declared count doesn't match the `AssignIdentityContainers(ctx, N, ...)` call in the body. Added to the `make verify` aggregate target alongside `verify-gomega-assertions`.

3. **Wire up OTE resource-aware scheduler** — bumps `openshift-tests-extension` dependency to pick up generic resource pool support ([OTE PR #73](https://github.com/openshift-eng/openshift-tests-extension/pull/73)). The test runner (`main.go`) now:
   - Walks all specs at startup, parses `MIContainers:N` labels, and sets `spec.Resources.ResourcePools["mi-containers"] = N` so the OTE scheduler enforces capacity before dispatching each test
   - Sets `Suite.ResourcePools` on all 9 suite definitions from `LEASED_MSI_CONTAINERS` env var (falls back to `DefaultMIContainerCount = 15`)
   - Refuses to start if any test is missing its `MIContainers` label (runtime fail-fast)
   - Logs pool size, source, and demand distribution at startup: `[scheduler] pool mi-containers=50 (source: LEASED_MSI_CONTAINERS), 82 specs (5×0, 75×1, 2×2+)`
   - Priority sort gated by `ARO_HCP_DISABLE_MI_SORT` env var (kept on initially for validation)

4. **Fix MI container cleanup bugs** — two fixes to the identity pool broker:
   - `release()` now skips cleanup callbacks for assigned-only entries (entries that were reserved but never used don't have FICs or role assignments to delete)
   - Identity container release moved before the `skipCleanup` gate in `per_test_framework.go` so containers are always returned to the pool, even when resource cleanup is skipped

5. **Suppress misleading scheduler warnings in non-test post-steps** — `parseMIContainerCount()` runs eagerly in `setupCli()` before cobra resolves the subcommand, so non-test operations (gather-observability, cleanup) no longer emit spurious pool warnings

### Why

E2E tests compete for a finite pool of shared managed identity (MI) containers (15–50 depending on environment). Without capacity-aware scheduling, multi-container tests get starved by single-container tests that consume all available capacity, causing long retry loops, timeouts, and flaky failures. The previous mitigation (priority-based sorting) was a workaround that couldn't prevent pool exhaustion.

The resource-aware scheduler gates test dispatch on actual pool availability — a test needing 3 containers won't be dispatched until 3 are free. This eliminates starvation and retry-loop contention. The `MIContainers(N)` labels + `verify-mi-containers` checker ensure every new test must declare its demand, preventing regressions where a new test is added without proper labeling.

The cleanup fixes prevent pool leaks (containers stuck in assigned/busy state) that further reduce effective pool capacity.

### Testing

- Dev and INT CI testing during Draft phase
- Validated under 15-container hard-cap (build `2077786113480593408`): scheduler handled 63 pool exhaustion events correctly with no deadlocks, 1:1 dispatch-on-complete, ~22% slower than 50-container baseline
- `make verify` passes (including new `verify-mi-containers` check)
- Unit tests for `release()` state transitions and `verify-mi-containers` AST checker

- The `MIDemandHigh`/`MIDemandMedium` labels are removed (replaced by `MIContainers(2)` and `MIContainers(3)`)
- Existing test code (`AssignIdentityContainers()` / `NextLeasedIdentityPool()`) is unchanged — the scheduler gates capacity before dispatch, so the broker's retry loop succeeds near-instantly
- Priority sort is kept initially behind `ARO_HCP_DISABLE_MI_SORT` — will be removed after 1-2 validated CI runs
- `helmtest/internal/config_test.go` and helm fixture changes are from `make materialize`/unrelated cleanup

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:


mgahagan73:
I think this is now ready for broader review. I checked the test results and none of the failures are related to this PR. The MI scheduler seemed to work fine in all environments.

test/util/framework/identities_helper_test.go
```diff
@@ -0,0 +1,132 @@
...
+				}
+			} else if cleanups != nil {
+				origCleanups := cleanups
+				cleanups = make([]func() error, len(origCleanups))
+				for i, fn := range origCleanups {
+					cleanups[i] = func() error {
+						cleanupsCalled++
+						return fn()
+					}
+				}
```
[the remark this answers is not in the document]
mgahagan73: This should not be an issue with recent golang versions.

test/cmd/aro-hcp-tests/main.go
```diff
@@ -66,14 +67,39 @@ func parseSuiteParallelismOverride() *int {
 	return &n
 }
 
-func miDemandPriority(spec *et.ExtensionTestSpec) int {
-	if spec.Labels.Has(labels.MIDemandHigh[0]) {
-		return 2
+const DefaultMIContainerCount = 15
```
mbukatov, line 76: Could you add go doc comment for this constant? Also I wonder if this is necessary as a transition mechanism. If not, I'm not sure it's a good idea to have such default hidden in the test runner.

test/util/labels/labels.go
```diff
@@ -54,12 +57,16 @@ var (
...
-	MIDemandHigh   = ginkgo.Label("MIDemand:High")
-	MIDemandMedium = ginkgo.Label("MIDemand:Medium")
-)
+// MIContainers declares how many managed identity containers a test needs.
+// Every It() and DescribeTable() must include this label. The verify-mi-containers
+// CI check and the runtime scheduler enforce this — unlabeled tests are rejected.
+func MIContainers(n int) ginkgo.Labels {
+	if n < 0 {
+		fmt.Fprintf(os.Stderr, "FATAL: MIContainers: n must be >= 0, got %d\n", n)
+		os.Exit(1)
```
mbukatov: I see that we can't pass error value directly given the return type, but I'm not sure if this is a correct error handling. How does a failure look like? Could we use panic instead?

hack/verify-mi-containers/main_test.go
```diff
@@ -0,0 +1,110 @@
...
+			file:           "testdata/zero_but_calls.go",
+			wantViolations: 1,
+			wantSubstring:  "MIContainers(0) but calls AssignIdentityContainers",
+		},
+		{
+			name:           "nonzero without AssignIdentityContainers call is flagged",
+			file:           "testdata/nonzero_no_call.go",
+			wantViolations: 1,
+			wantSubstring:  "MIContainers(2) but does not call AssignIdentityContainers",
+		},
```
mbukatov, line 65: What would happen if we have multiple labels in the same section? Is that worth covering in this test suite?

tooling/helmtest/internal/config_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	result := ReplaceImageDigestExcluding(cfg, []string{
+		"hypershift.sharedIngressImage.digest",
+	})
+
+	assert.Equal(t, "sha256:1234567890", result["hypershift"].(map[string]any)["image"].(map[string]any)["digest"])
+	assert.Equal(t, "sha256:bbb", result["hypershift"].(map[string]any)["sharedIngressImage"].(map[string]any)["digest"])
+	assert.Equal(t, "sha256:1234567890", result["other"].(map[string]any)["digest"])
+}
```
mbukatov: Isn't this file included by mistake? It doesn't seem to be related to MI resource scheduler.

test/cmd/aro-hcp-tests/main.go
```diff
@@ -282,13 +312,45 @@ func setupCli() *cobra.Command {
...
 	// })
 
-	// Sort specs so tests with higher managed identity container demand are
-	// dispatched first. This prevents starvation: multi-container tests get
-	// dispatched while the pool is full, before single-container tests can
-	// consume all available capacity.
-	sort.SliceStable(specs, func(i, j int) bool {
-		return miDemandPriority(specs[i]) > miDemandPriority(specs[j])
+	var missingLabel []string
+	var demand0, demand1, demandN int
```
mbukatov, line 449: Why do we count demand this way? Does it follow from the openshift-tests-extension resource scheduler? Could you add a go doc string about it if possible?

mbukatov (commented):
This looks good, thanks for implementing it. I like that this PR includes linter to make sure labels are used correctly along with unit tests.

Providing few minor comments.

test/util/labels/labels.go
mgahagan73: For this case (which I think is very unlikely given the pre-commit checks) I thought an exit with a fatal error would be easier handle/debug than filling up the logs with a long call trace, but if you think we are better off panic'ing then I can make it panic.

test/cmd/aro-hcp-tests/main.go
mgahagan73: Comment added, I added this constant to make it easier for use with personal dev environments. It is only used If LEASED_MSI_CONTAINERS is empty or not set (comment on line 109).

tooling/helmtest/internal/config_test.go
mgahagan73: removed

test/cmd/aro-hcp-tests/main.go
mgahagan73: Comments added.

hack/verify-mi-containers/main_test.go
mgahagan73: parseMIContainersLabel should now reject multiple labels. We could extend the verifier to detect duplicate AST-level labels.MIContainers() calls as a follow-up, we will already catch this at runtime.

test/cmd/aro-hcp-tests/main.go
mbukatov: Thank you, this makes sense.

test/util/labels/labels.go
mbukatov: I was concerned about a proper error reporting, and wondered if we used panic, we get ginkgo to report more context.

I tried the proposed approach, added a test case which uses invalid MI value, compiled the test runner and got these errors without even trying to run a test:

```
$ ./aro-hcp-tests
FATAL: MIContainers: n must be >= 0, got -1
```

```
$ ./aro-hcp-tests -h
FATAL: MIContainers: n must be >= 0, got -1
```

But when I used panic:

```.diff
--- a/test/util/labels/labels.go
+++ b/test/util/labels/labels.go
@@ -16,7 +16,6 @@ package labels

 import (
        "fmt"
-       "os"

        "github.com/onsi/ginkgo/v2"
 )
@@ -66,8 +65,7 @@ var (
 // CI check and the runtime scheduler enforce this — unlabeled tests are rejected.
 func MIContainers(n int) ginkgo.Labels {
        if n < 0 {
-               fmt.Fprintf(os.Stderr, "FATAL: MIContainers: n must be >= 0, got %d\n", n)
-               os.Exit(1)
+               panic(fmt.Sprintf("FATAL: MIContainers: n must be >= 0, got %d\n", n))
        }
        return ginkgo.Label(fmt.Sprintf("MIContainers:%d", n))
 }
```

I got:

```
$ ./aro-hcp-tests
panic: couldn't build extension test specs from ginkgo: couldn't build ginkgo tree: Assertion or Panic detected during tree construction
        var _ = Describe("Martin", func() {
        /home/martin/projects/ARO-HCP/test/e2e/cluster_martin.go:25
          Ginkgo detected a panic while constructing the spec tree.
          You may be trying to make an assertion in the body of a container node
          (i.e. Describe, Context, or When).

          Please ensure all assertions are inside leaf nodes such as BeforeEach,
          It, etc.

          Here's the content of the panic that was caught:
          FATAL: MIContainers: n must be >= 0, got -1


          Learn more at: http://onsi.github.io/ginkgo/#no-assertions-in-container-nodes


goroutine 1 [running]:
main.setupCli()
        /home/martin/projects/ARO-HCP/test/cmd/aro-hcp-tests/main.go:190 +0x28f4
main.main()
        /home/martin/projects/ARO-HCP/test/cmd/aro-hcp-tests/main.go:507 +0x13
```

This is slightly more verbose, but it points to the source code file and line number, along with the assertion.

hack/verify-mi-containers/main_test.go
mbukatov: If you could extend the unit tests in this PR directly, it would be better, as the change under test is done here already, and spinning such a small change into a separate PR would require more work (such as review and E2E reruns).

test/cmd/aro-hcp-tests/main.go
mbukatov: Thank you.

test/util/framework/per_test_framework.go
```diff
@@ -239,6 +239,14 @@ func (tc *perItOrDescribeTestContext) deleteCreatedResources(ctx context.Context
 		tc.RecordTestStep("Delete created resources", startTime, finishTime)
 	}()
 
+	// Always release identity containers — they are shared pool resources,
+	// not test-created resources subject to skipCleanup.
+	defer func() {
+		if err := tc.releaseLeasedIdentities(ctx); err != nil {
+			ginkgo.GinkgoLogr.Error(err, "failed to release leased identities")
+		}
+	}()
```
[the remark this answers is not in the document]
mgahagan73: Adding this would potentially leak MI containers so I'm not sure we would want this.

hack/verify-mi-containers/main_test.go
mgahagan73: This should be covered now.

test/util/labels/labels.go
mgahagan73: Done, we panic now.

mbukatov (commented):
/lgtm

thank you for the updates

roivaz:
Great job!
/approve

PR #6032, opened 2026-07-12
backend/metricscontrollers: scope operation metric cleanup to operati…

stevekuznetsov:
…on_type

Previously, the operation phase metrics handler called deleteByResourceID before emitting new metric labels, which cleared ALL series for a given resource_id regardless of operation_type. When multiple operations of different types existed for the same ARM resource (e.g. a completed 'create' and an in-flight 'delete'), the informer's unordered relist iteration caused whichever operation was processed last to clobber the other's metrics. This race produced false 'stuck operation' alerts: an in-flight delete operation's start_time could be overwritten by the older create operation's start_time on one relist, then restored on the next.

Scope the pre-emit DeletePartialMatch to resource_id + operation_type. Operations of different types (create, update, delete, requestcredential, revokecredentials) now coexist as independent series on the same resource. Operations of the same type still collapse to one series (last writer wins), which is correct since only one operation of a given type can be active on a resource at a time.

Observed in production: an ExternalAuth delete operation completed at 05:07:04 UTC but the metric continued showing phase=deleting because the completed create operation (same resource_id, different operation_type) was intermittently winning the relist race and clobbering the delete series. This caused a BackendAsyncOperation- ExternalAuthDeleteStuck alert to fire 4 minutes after the operation had already succeeded.

### DIFF
--- a/backend/pkg/controllers/metricscontrollers/operation_phase_metrics_controller.go
+++ b/backend/pkg/controllers/metricscontrollers/operation_phase_metrics_controller.go
@@ -39,17 +39,16 @@ var operationMetricLabelNames = []string{"resource_id", "subscription_id", "reso
 // resources. This matches the format already used by the sibling
 // backend_resource_provision_state metric family.
 //
-// Multiple operations on the same ARM resource id collapse to one
-// series. On the AllOperations() informer's unordered iteration
-// (every relist / resync, or backend restart) whichever operation
-// is processed last wins for that resource_id; this may temporarily
-// or persistently reflect an older terminal op until a later emit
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
/label lgtm

What we lose in extra cardinality is fine (even 10x on 100 clusters per prometheus node is nothing) when what we gain is the ability to do useful things at all :)

stevekuznetsov:
Image job failures are due to DPTP issues - chatting with Deep, it looks like it will take a while to fix that. 

deads2k ack'ed this change, and without it we will see spurious e2e failures on the alerts. Going to FM this so that folks running their CI in EU morning have working e2e

PR #6036, opened 2026-07-12
backend: add a controller to auto-scale the Azure Monitor Workspace

stevekuznetsov:
We have two (and only ever two) Azure Monitor Workspaces in a region - the `hcp` and the `svc` ones. Each of those needs to have their ingestion limit and max number of time series managed by us, since we can't a priori ask for enough capacity to run an end-to-end test and we do not care much to have alerts going off in prod where the only recourse is to bump the limit in config and re-deploy. We can only ask for more capacity when we're using more than 50% of the current limit, up to 20M:

https://learn.microsoft.com/en-us/azure/azure-monitor/metrics/azure-monitor-workspace-monitor-ingest-limits?tabs=portal#request-for-an-increase-in-ingestion-limits-preview


config/rendered/dev/pers/westus3.yaml
```diff
@@ -102,6 +102,8 @@ automationDryRun: false
 azureGeoShortId: us
 azureRegionAvailabilityZoneCount: 4
 backend:
+  amwScaling:
+    pollInterval: 1m
```
stevekuznetsov: Set this higher in prod, needs to be tight in dev.

backend/pkg/controllers/amwscaling/controller.go
```diff
@@ -0,0 +1,368 @@
...
+}
+
+func (c *Controller) readUtilization(ctx context.Context, subscriptionID, resourceID string) (*AMWUtilization, error) {
+	client, err := c.metricsClientFactory(subscriptionID)
+	if err != nil {
+		return nil, fmt.Errorf("creating metrics client: %w", err)
+	}
+
+	metricNames := activeTimeSeriesMetric + "," + eventsPerMinuteMetric
+	timespan := "PT5M" // last 5 minutes
```
stevekuznetsov: unclear from docs what timespan they use to check for the 50% utilization target, may need to change

PR #6040, opened 2026-07-13
chore: route kernel (kmsg) logs to Kusto systemdLogs for node-freeze investigation

roivaz:
## Why

ARO-HCP management-cluster `userswft` nodes intermittently **hard-freeze** — the whole VM goes silent at once (kubelet, containerd and node-problem-detector all stop in the same instant) and the Azure platform then declares the VM unavailable. In dev CI this trips `KubeNodeNotReady` on ~12% of `e2e-parallel` runs (**[AROSLSRE-1465](https://redhat.atlassian.net/browse/AROSLSRE-1465)**), and it has also been seen in persistent environments.

When it happens we have **no window into the guest kernel** in the seconds before the freeze:

- AKS nodepools expose no settable boot/serial `diagnosticsProfile`, so boot diagnostics are off and cannot be turned on.
- The OS disk is wiped on reimage, so nothing is recoverable after the fact.

We need the guest kernel ring buffer (`hung_task` / soft-lockup / RCU stall / OOM / MCE / panic traces) streamed **off the node continuously**, so the pre-freeze lines survive even when the node itself does not. This is generally useful forensic data for node-failure triage, not just for this one bug.

## What

**1. Capture kernel (kmsg) logs.** Add `Systemd_Filter _TRANSPORT=kernel` to the existing arobit fluent-bit `systemd` input.

- Kernel journal records live in the **same** system journal fluent-bit already reads and carry no `_SYSTEMD_UNIT`; they are matched by `_TRANSPORT=kernel`.
- `Systemd_Filter` entries are OR'd, so this **only adds** kernel lines alongside the existing kubelet/containerd capture.
- fluent-bit streams continuously (`Flush 1s`), so lines emitted before the freeze instant ship off-node; only the final sub-second is lost.

**2. Drop kernel audit noise.** The kernel journal also carries SELinux/audit-subsystem records routed through kmsg — lines beginning with `audit:` (audit types `1300/1325/1327/1130/1131` — syscall/netfilter/service-state audits) and `kauditd_printk_skb: N callbacks suppressed` throttle notices. These are process/security audit events, not kernel-health signals. A `grep` filter drops them so only actionable kernel diagnostics are ingested. kubelet/containerd records and non-audit kernel lines are unaffected.

### No Kusto schema change

Kusto is shared, so this deliberately does **not** touch any schema. Kernel lines flow to the **existing** `systemdLogs` table via its existing ingestion mapping:

- The full journal record is preserved in the `log` dynamic column → `_TRANSPORT` / `SYSLOG_IDENTIFIER` / `PRIORITY` stay queryable.
- `message` maps from `$.log.MESSAGE`.
- `systemd_unit` is simply empty for kernel lines.

Query them with e.g. `ServiceLogs.systemdLogs | where log.TRANSPORT == "kernel"`.

## Volume / cost

Measured on real `e2e-parallel` runs (mgmt cluster, ~1.4h active, 21 `userswft` nodes), as full ingested payload:

| Feed | Volume | Share of existing `systemdLogs` |
|---|---:|---:|
| Existing (kubelet + containerd) | ~2,205 MB | baseline |
| Kernel — audit noise (**dropped by this PR**) | ~50.6 MB | — |
| **Kernel — diagnostics (kept)** | **~7.5 MB** | **+0.34%** |

The audit filter takes the added volume from +2.6% to **+0.34%** on top of the existing feed — a rounding error next to the kubelet/containerd and `containerLogs` feeds the clusters already sustain through the same pipeline. Verified live: on a run with the filter, audit records ingested = **0**, with the genuine kernel diagnostics retained.

## Testing

- Validated end-to-end on `e2e-parallel` runs from this PR: kernel records land in `systemdLogs` with full fidelity (captured complete kernel stack traces incl. registers/modules/call-chain), and the audit filter drops 100% of audit noise while keeping diagnostics.
- Helm golden fixtures regenerated via `UPDATE=true go test -count=1 ./...` in `tooling/helmtest`; verify-mode `go test` passes.
- No daemonset / RBAC change required (kernel records are already in the mounted journal).

### DIFF
--- a/dev-infrastructure/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_arobit.yaml
+++ b/dev-infrastructure/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_arobit.yaml
@@ -120,6 +120,16 @@ data:
         Tag             systemd.*
         Systemd_Filter  _SYSTEMD_UNIT=containerd.service
         Systemd_Filter  _SYSTEMD_UNIT=kubelet.service
+        # AROSLSRE-1465: capture kernel ring-buffer (kmsg) messages so we can see
+        # the guest kernel stack (hung_task/soft-lockup/panic) in the seconds
+        # before a userswft node hard-freezes into NodeNotReady. Kernel journal
+        # records carry no _SYSTEMD_UNIT; they are matched by _TRANSPORT=kernel.
+        # Systemd_Filter entries are OR'd, so this only adds kernel lines. They
+        # flow to the existing systemdLogs Kusto table (no schema change): the
+        # full record is preserved in the `log` dynamic column, so _TRANSPORT /
+        # SYSLOG_IDENTIFIER / PRIORITY stay queryable, and `message` maps from
+        # $.log.MESSAGE. systemd_unit is simply empty for kernel lines.
+        Systemd_Filter  _TRANSPORT=kernel
         Read_From_Tail  On
         Strip_Underscores On
         DB              /var/fluent-bit/flb-storage/flb-systemdb.db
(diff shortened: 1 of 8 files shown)


raelga (approved):
/lgtm
/approve

PR #6041, opened 2026-07-13
feat: move and enable asynchronous CS create interaction for externalauths

miguelsorianod:
Move Cluster Service interaction out of the frontend create path and hand it off to the backend's async externalauth creation controllers.

External Auth create operations no longer set an InternalID as it is not needed anymore when using the new external auth creation approach in backend.


miguelsorianod:

Closing this.

PR #6049, opened 2026-07-13
Generated controller flow documentation

deads2k:
attempt to automatically generate some useful documentation about how clusters are actually created.


mbarnes:
This is good reference material.  Only minor gripe is I would personally prefer all docs go under `/docs`.

deads2k:
rebased and moved dir

/label lgtm
/hold cancel

PR #6058, opened 2026-07-13
Add deletes to change feeds

deads2k:
Add deletes to changefeeds

### DIFF
--- a/internal/database/crud_helpers.go
+++ b/internal/database/crud_helpers.go
@@ -20,7 +20,10 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/utils/ptr"
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
@@ -29,6 +32,8 @@ import (
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
+
+const SoftDeleteTTLSeconds = 30
 
 func OldResourceIDToCosmosID(resourceID *azcorearm.ResourceID) (string, error) {
 	if resourceID == nil {
@@ -76,7 +81,19 @@ func getByItemID[InternalAPIType, CosmosAPIType any](ctx context.Context, contai
 		return nil, utils.TrackError(err)
 	}
 
+	if IsSoftDeleted(responseItem.Value) {
+		return nil, NewNotFoundError()
+	}
+
 	return responseItemToInternalObj[InternalAPIType, CosmosAPIType](ctx, cosmosID, responseItem)
+}
+
+func IsSoftDeleted(data []byte) bool {
+	var doc TypedDocument
+	if err := json.Unmarshal(data, &doc); err != nil {
+		return false
+	}
+	return doc.DeletionTimestamp != nil
 }
 
 func get[InternalAPIType, CosmosAPIType any](ctx context.Context, containerClient *azcosmos.ContainerClient, partitionKeyString string, completeResourceID *azcorearm.ResourceID) (*InternalAPIType, error) {
@@ -101,9 +118,9 @@ func list[InternalAPIType, CosmosAPIType any](ctx context.Context, containerClie
 		PageSizeHint: -1,
 	}
 	if prefix == nil {
-		query = "SELECT * FROM c WHERE LENGTH(c.resourceID) > 0"
+		query = "SELECT * FROM c WHERE LENGTH(c.resourceID) > 0 AND (NOT IS_DEFINED(c.deletionTimestamp))"
 	} else {
-		query = "SELECT * FROM c WHERE STARTSWITH(c.resourceID, @prefix, true)"
+		query = "SELECT * FROM c WHERE STARTSWITH(c.resourceID, @prefix, true) AND (NOT IS_DEFINED(c.deletionTimestamp))"
 		queryOptions = azcosmos.QueryOptions{
 			PageSizeHint: -1,
 			QueryParameters: []azcosmos.QueryParameter{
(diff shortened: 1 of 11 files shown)


internal/database/informers/changefeed_list_watch.go
```diff
@@ -404,10 +404,16 @@ func (c *ChangeFeedWatcher[InternalAPIType, InternalAPITypePointer, CosmosAPITyp
 	}
 
 	objDeleted := false
-	if c.shouldDeliverItemFn != nil && !c.shouldDeliverItemFn(internalObj) {
+	if objAsTypedDocument.DeletionTimestamp != nil {
+		if objPreviouslySeen {
+			objDeleted = true
+		} else {
+			logger.Info("skipping soft-deleted document not previously seen", "content", cosmosObj)
+			return nil
```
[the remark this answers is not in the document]
deads2k: Wrong, we want to do this for datadump purposes.

internal/database/crud_helpers.go
```diff
@@ -347,20 +363,79 @@ func deleteResource(ctx context.Context, containerClient *azcosmos.ContainerClie
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal soft-deleted document: %w", err)
+	}
+
+	opts := &azcosmos.ItemOptions{
+		IfMatchEtag: &responseItem.ETag,
+	}
+	_, err = containerClient.ReplaceItem(ctx, pk, cosmosID, modified, opts)
 	if err != nil {
 		return utils.TrackError(err)
```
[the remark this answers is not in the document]
deads2k: This is a better outcome.

internal/database/crud_helpers.go
```diff
@@ -347,20 +363,79 @@ func deleteResource(ctx context.Context, containerClient *azcosmos.ContainerClie
...
 	}
-	return nil
-}
 
-func deleteByCosmosID(ctx context.Context, containerClient *azcosmos.ContainerClient, partitionKeyString, cosmosID string) error {
-	_, err := containerClient.DeleteItem(ctx, azcosmos.NewPartitionKeyString(partitionKeyString), cosmosID, nil)
+	var typedDoc TypedDocument
+	if err := json.Unmarshal(responseItem.Value, &typedDoc); err != nil {
+		return fmt.Errorf("failed to unmarshal document for soft delete: %w", err)
+	}
```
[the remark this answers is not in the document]
deads2k: I wonder if this is true. I bet the bot is wrong since this mostly worked last e2e run.

deads2k:
pods not running again.  Need to reach the next level on that.

/retest

internal/databasetesting/mock_resources_crud.go
```diff
@@ -303,7 +308,37 @@ func (m *mockResourceCRUD[InternalAPIType, InternalAPITypePointer, CosmosAPIType
...
+	modified, err := json.Marshal(typedDoc)
+	if err != nil {
+		return fmt.Errorf("failed to marshal soft-deleted document: %w", err)
+	}
+
+	dataWithETag, _, err := injectETag(json.RawMessage(modified))
+	if err != nil {
+		return fmt.Errorf("failed to inject etag: %w", err)
+	}
+	store.StoreDocument(cosmosID, dataWithETag)
```
machi1990, line 341: Should we also honor the TTL and sets a goroutine that wipes out the doc after the TTL? https://github.com/deads2k/ARO-HCP/blob/11f0b43b0c2f2e058619661f4ea8f89179c1304b/internal/database/crud_helpers.go#L417

internal/database/crud_helpers.go
```diff
@@ -347,20 +363,79 @@ func deleteResource(ctx context.Context, containerClient *azcosmos.ContainerClie
...
+							cmBytes, _ := json.Marshal(cm)
+							props["cosmosMetadata"] = cmBytes
+							propsBytes, _ := json.Marshal(props)
+							doc.Properties = propsBytes
+						}
+					}
+				}
+			}
+		}
+	}
```
machi1990: This will be easier to follow if we use `unstructured` similar to how we do for the `redactTypedDocument ` in the data dumper.
See suggestion below

```suggestion
	if len(doc.Properties) > 0 {
		var props unstructured.Unstructured
		if err := json.Unmarshal(doc.Properties, &props.Object); err != nil {
			return fmt.Errorf("unmarshal properties: %w", err)
		}
		iv, found, err := unstructured.NestedInt64(props.Object, "cosmosMetadata", "instanceVersion")
		if err != nil {
			return fmt.Errorf("read cosmosMetadata.instanceVersion: %w", err)
		}
		if found {
			if err := unstructured.SetNestedField(props.Object, iv+1, "cosmosMetadata", "instanceVersion"); err != nil {
				return fmt.Errorf("set cosmosMetadata.instanceVersion: %w", err)
			}
			updated, err := json.Marshal(props.Object)
			if err != nil {
				return fmt.Errorf("marshal properties: %w", err)
			}
			doc.Properties = updated
		}
	}
```

machi1990 (commented):
left two suggestions, other lgtm

internal/databasetesting/mock_resources_crud.go
mbarnes: IMO, not necessary.  The deletion after TTL is undetectable through the DB client interfaces.
_Maybe_ simulate it as a unit test in the database or changefeed packages if we feel it needs to be tested.

mbarnes:
LGTM as well, nothing to add beyond @machi1990's comments.

internal/databasetesting/mock_resources_crud.go
deads2k: Agree, we'll defer.

internal/database/crud_helpers.go
deads2k: it didn't like unstructured, but I did update to use the unstructure helper functions.

deads2k:
comments addressed.  apply label base don't on previous comment and in slack
/label lgtm

machi1990:
/retest 

The test failed with 
```
{fail [github.com/Azure/ARO-HCP/test/e2e/cluster_pullsecret.go:291]: failed to create pull-secret-test pod in pullsecret-image-test
Unexpected error:
    <*errors.StatusError | 0xc0011fa280>: 
    pods "pull-secret-test" is forbidden: error looking up service account pullsecret-image-test/default: serviceaccount "default" not found
    ...
occurred  fail [github.com/Azure/ARO-HCP/test/e2e/cluster_pullsecret.go:291]: failed to create pull-secret-test pod in pullsecret-image-test
Unexpected error:
    <*errors.StatusError | 0xc0011fa280>: 
    pods "pull-secret-test" is forbidden: error looking up service account pullsecret-image-test/default: serviceaccount "default" not found
    ...
occurred}
```

perhaps we need something like https://github.com/Azure/ARO-HCP/pull/5973/changes I'll open a separate PR for that

PR #6060, opened 2026-07-14
fix: flappy nodepoolstuckoperation alert fix

sclarkso:
### What

Wrapped the UJNodePoolStuckOperation expression in max_over_time(...)[6h:5m] so that when a backend retry resets start_time_seconds and briefly drops the age below the 7200s threshold, the alert stays firing instead of auto-resolving and creating a duplicate IcM incident. Added a comment block explaining the rationale.

### Why

 When a node pool operation is genuinely stuck, the backend periodically retries it, which resets backend_resource_operation_start_time_seconds. This briefly drops (time() - start_time) below the 7200s threshold, causing the alert to auto-resolve. When the age re-crosses the threshold a few hours later, a brand-new IcM incident is opened for the same wedged resource. In practice this produces roughly one duplicate incident every ~4 hours for a single stuck operation.

### Testing

  - Existing tests 1-12 pass unchanged (the max_over_time wrapper doesn't change behavior for continuously-stuck or non-stuck operations)
  - Added Test 13: validates a stuck delete where start_time is reset by a backend retry at t=240m — confirms the alert stays firing through the gap when the raw condition is false
  - Added Test 14: validates that the 6h hold is bounded — after the operation completes, the alert resolves once there has been no stuck sample for 6 hours
  - make alerts passes and regenerates the bicep correctly

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


raelga (approved):
/lgtm
/approve

PR #6067, opened 2026-07-14
feat: move and enable asynchronous CS update interactions

miguelsorianod:
This PR removes CS update interactions from frontend, which makes the update dispatch controllers in backend fully take ownership of interacting with CS.


miguelsorianod:

Closing this.

PR #6068, opened 2026-07-14
feat: add more externalauth validations in frontend

miguelsorianod:
I did a general review of static validations that are performed in CS. This PR contains validations that are being performed in CS, implemented in Frontend.

This should not impact existing resources because CS is already enforcing the validation.

### DIFF
--- a/internal/validation/validate_externalauth.go
+++ b/internal/validation/validate_externalauth.go
@@ -179,13 +179,13 @@ func validateTokenIssuerProfile(ctx context.Context, op operation.Operation, fld
 	errs = append(errs, validate.RequiredSlice(ctx, op, fldPath.Child("audiences"), newObj.Audiences, safe.Field(oldObj, toTokenIssuerProfileAudiences))...)
 	errs = append(errs, MinItems(ctx, op, fldPath.Child("audiences"), newObj.Audiences, safe.Field(oldObj, toTokenIssuerProfileAudiences), 1)...)
 	errs = append(errs, MaxItems(ctx, op, fldPath.Child("audiences"), newObj.Audiences, safe.Field(oldObj, toTokenIssuerProfileAudiences), 10)...)
+	errs = append(errs, validate.EachSliceVal(
+		ctx, op, fldPath.Child("audiences"),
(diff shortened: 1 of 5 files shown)


internal/validation/validate_externalauth.go
```diff
@@ -179,17 +180,17 @@ func validateTokenIssuerProfile(ctx context.Context, op operation.Operation, fld
...
 	//	newObj.Audiences, safe.Field(oldObj, toTokenIssuerProfileAudiences),
 	//	nil, nil,
-	//	validate.RequiredValue,
-	//)...)
-	//errs = append(errs, validate.EachSliceVal(
-	//	ctx, op, fldPath.Child("audiences"),
-	//	newObj.Audiences, safe.Field(oldObj, toTokenIssuerProfileAudiences),
-	//	nil, nil,
 	//	NoExtraWhitespace,
 	//)...)
```
JameelB, line 195: Shall we add this (note we're currently not doing this in cs)? from the authentication [type](https://github.com/openshift/api/blob/master/config/v1/types_authentication.go#L302-L311), it seems like it's not really validated but we could enforce it. This is usually a uri, a guid or a string usually not containing a whitespace.

If we do decide to add this though, it must also be done for the client id. For oidc auth type, the client id must always be included as part of the audience.

JameelB (commented):
overall lgtm, matches the validations in cs and requirements in authentication [type](https://github.com/openshift/api/blob/master/config/v1/types_authentication.go). 

Just one comment on an existing to do - though not strictly required for this change.

internal/validation/validate_externalauth.go
miguelsorianod: Do you know if extra whitespaces are supported once that setting reaches Hypershift/OCP? if not, we can add the enforcement if we want. If they are supported, then because we don't have any validation nothing prevents that someone already has one with such scenario, in which case we can't enforce it as it would break the update flow, unless we check that noone is really having it in any permanent environment, and even in that case there's a potential amount of time where inbetween our enforcement merging and the rollout someone does it

PR #6079, opened 2026-07-14
e2e: issue certificates from delegated DNS zone from LetsEncrypt

bennerv:
### What

An e2e test to prove that we can issue TLS certificates from a publicly trusted CA (LetsEncrypt in this case) 

### Why

We plan to shift the control of the ingress domain more in the customer's direction, and this proves that it's possible to do so by ensuring the customer can request certificates for the delegated domain.  

### Testing

This test proves this is possible.  

Currently this uses the LetsEncrypt staging CA rather than prod due to rate limits.


test/e2e/cluster_ingress_cert_letsencrypt.go
```diff
@@ -0,0 +1,531 @@
...
+			break
+		}
+	}
+	if !sanFound {
+		return fmt.Errorf("leaf cert SANs %v do not include expected %q", leaf.DNSNames, expectedSAN)
+	}
+
+	issuerOrg := strings.Join(leaf.Issuer.Organization, " ")
+	issuerCN := leaf.Issuer.CommonName
+	isLetsEncrypt := strings.Contains(issuerCN, "Encrypt") ||
```
bennerv: Rather than doing this, we should validate the certificate chain up to the root CA of let's encrypt similarly to how we do in `cluster_tls_endpoints.go`.

test/e2e/cluster_ingress_cert_letsencrypt.go
```diff
@@ -0,0 +1,642 @@
...
+// ACME DNS-01 challenges with the Azure DNS solver.
+func unstructuredClusterIssuer(name, email, server, hostedZoneName, resourceGroupName, subscriptionID, uamiClientID string) *unstructured.Unstructured {
+	return &unstructured.Unstructured{Object: map[string]interface{}{
+		"apiVersion": "cert-manager.io/v1",
+		"kind":       "ClusterIssuer",
+		"metadata":   map[string]interface{}{"name": name},
+		"spec": map[string]interface{}{
+			"acme": map[string]interface{}{
+				"email":  email,
+				"server": server,
```
bennerv, line 410: Why staging?  Their [prod service](https://letsencrypt.org/docs/rate-limits/) only allows issuing 50 certificates every 7 days, based on the TLD.  Since we run more than 50 e2e tests in 7 days, we adhere to the [staging service](https://letsencrypt.org/docs/staging-environment/) which has higher rate limits.
> Up to 50 certificates can be issued per registered domain (or IPv4 address, or IPv6 /64 range) every 7 days. This is a global limit, and all new order requests, regardless of which account submits them, count towards this limit. The ability to issue new certificates for the same registered domain refills at a rate of 1 certificate every 202 minutes.

test/e2e/cluster_ingress_cert_letsencrypt.go
```diff
@@ -0,0 +1,642 @@
...
+			Expect(msiResp.Properties).NotTo(BeNil(), "UAMI properties was nil")
+			Expect(msiResp.Properties.ClientID).NotTo(BeNil(), "UAMI ClientID was nil")
+			Expect(msiResp.Properties.PrincipalID).NotTo(BeNil(), "UAMI PrincipalID was nil")
+			uamiClientID := *msiResp.Properties.ClientID
+			uamiPrincipalID := *msiResp.Properties.PrincipalID
+
+			tenantID := tc.TenantID()
+			if len(tenantID) == 0 {
+				tenantID = *msiResp.Properties.TenantID
+			}
```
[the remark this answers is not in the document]
bennerv: fixed.

test/e2e/cluster_ingress_cert_letsencrypt.go
```diff
@@ -0,0 +1,639 @@
...
+// Pinned upstream cert-manager release. Bump deliberately; do not float to :latest.
+const certManagerManifestURL = "https://github.com/cert-manager/cert-manager/releases/download/v1.20.2/cert-manager.yaml"
+
+var _ = Describe("Customer", func() {
+	It("should be able to issue the default ingress serving certificate from Let's Encrypt using cert-manager with ACME DNS-01 challenges",
+		labels.RequireNothing,
+		labels.High,
+		labels.Positive,
+		labels.CreateCluster,
+		labels.AroRpApiCompatible,
```
mgahagan73, line 75: Now that the MI Container aware scheduler has merged, you will need to add a label depicting how many Managed Identity containers this test will use. e.g. labels.MIContainers(n) where n is the number of MI Containers. Please also rebase to pick up the container checks that were added to make verify to avoid any last minute surprises.

test/e2e/cluster_ingress_cert_letsencrypt.go
```diff
@@ -0,0 +1,640 @@
...
+				ingressNamespace     = "openshift-ingress"
+				issuerName           = "letsencrypt-staging" // we're using staging as prod has strict limits on cert signing
+				acmeEmail            = "noreply@redhat.com"
+				acmeServer           = "https://acme-staging-v02.api.letsencrypt.org/directory"
+			)
+			tc := framework.NewTestContext()
+			if tc.UsePooledIdentities() {
+				err := tc.AssignIdentityContainers(ctx, 1, framework.IdentityContainerAssignmentRetryInterval)
+				Expect(err).NotTo(HaveOccurred(), "failed to assign pooled identity containers")
+			}
```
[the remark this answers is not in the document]
bennerv: it's available everywhere, not applicable.

mzazrivec (approved):

PR #6080, opened 2026-07-14
feat(observability): hcp kube-apiserver error rate and latency alerts to SRE

SudoBrendan:
### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


observability/alerts/UserJourneyKubeApiserverErrorsMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,112 @@
...
+        summary: "[HCPKASErrorsBurn] {{ $labels.cluster }} / {{ $labels.namespace }} (1h/5m)"
+        description: |
+          Management Cluster: {{ $labels.cluster }}
+          Namespace: {{ $labels.namespace }}
+          Lookup: use hcpctl to map namespace to cluster resource ID.
+        runbook_url: https://aka.ms/arohcp-runbook-ujkasavailable
+        correlationId: "userJourneyKubeApiserverErrors/{{ $labels.cluster }}/{{ $labels.namespace }}"
+    - alert: userJourneyKubeApiserverErrors6h30m
+      expr: |
+        (
```
swiencki, line 47: The comment says a follow-up will align slow-burn to severity "4"; add a tracking Jira reference so it doesn't get lost.

observability/alerts/UserJourneyKubeApiserverErrorsMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,112 @@
...
+          kas:apiserver_request_5xx:rate_avg_5m / kas:apiserver_request_total:rate_avg_5m
+          > (14.4 * (1 - 0.9995))
+          and on (namespace, cluster)
+          kas:apiserver_request_total:rate_avg_5m > 1
+        )
+        and on (namespace, cluster)
+        (
+          kas:apiserver_request_5xx:rate_avg_1h / kas:apiserver_request_total:rate_avg_1h
+          > (14.4 * (1 - 0.9995))
+          and on (namespace, cluster)
```
swiencki, line 22: The correlationId uses namespace while the availability alerts use _id; consider aligning for consistent IcM incident grouping across all KAS alert types.

SudoBrendan: I like the idea, but this is actually a bit complicated. To get `_id` requires a join of some kind, because the KAS itself (where this metric comes from) is _id (hypershift) unaware. To do this, we'd introduce dependencies across our various SLIs.

As an alternative, we _could_ update the other HCP alert to use `namespace` as our identifier instead of `_id`. It's worth knowing that those alerts would have correlation IDs like `cluster-1/ocm-env-abc123-foo` and these would have `cluster-1/ocm-env-abc123`, but the prefixes already distinguish different alerts from each other, and honestly I think this identifies the alerts a bit better (where is the SLI coming from, rather than some random GUID from OCM that I don't see SRE relying on) anyways? WDYT?

swiencki: Using `namespace` is more inline with what SRE would require. Question is do we want to make this the standard?

observability/alerts/UserJourneyKubeApiserverLatencyMonitor-prometheusRule.yaml
```diff
@@ -0,0 +1,108 @@
...
+  name: userjourney-kubeapiserver-latency-alerts
+  namespace: prometheus
+spec:
+  groups:
+  - name: rp-userjourney-kas-latency-monitor-rules
+    rules:
+    - alert: userJourneyKubeApiserverLatency1h5m
+      expr: |
+        (
+          (1 - kas:apiserver_request_latency:sli_ratio:rate5m)
```
sankur-codes, line 13: All the `kas:apiserver_request_latency:sli_ratio:` are dependent on and blockd by upstream [bug](https://github.com/openshift/hypershift/pull/8715). 

The bug has been merged and backported till 4.20 (our default version). Once rolled out for our HCp clusters, should emit the metrics for use.

PR #5709, opened 2026-06-18
feat: migrate clusters service cluster create from frontend to backend

JakobGray:
### What

This PR combines the frontend and backend changes for the cluster service create migration. 

### Why

It is intended to test the changes together and not be merged.

### PR Checklist
- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
- [ ] Screenshots included (if graph/UI/metrics changes)
- [ ] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [ ] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

PR #6082, opened 2026-07-14
Cluster install migration

JakobGray:
https://redhat.atlassian.net/browse/ARO-24823

### What

This continues migrating HCP cluster installation so Cluster Service (CS) provisioning happens asynchronously in the backend instead of synchronously in the frontend, and backfills the validation CS used to do synchronously so users still get fast, equivalent errors.

**Defer CS provisioning to the backend**

- Frontend no longer calls PostCluster during cluster create; it persists the cluster with an empty ClusterServiceID and enqueues the ARM operation for the backend controller to provision CS async. Removes now-unused frontend provision-shard/noop provision flags.

**Backfill validation CS used to enforce synchronously**

- Reject duplicate managed resource group, subnet, and NSG across a subscription's clusters/node pools (admitCluster* in internal/admission), plus static NSG same-subscription/not-in-MRG checks — mirrors CS's performSpecValidation uniqueness checks.
- Validate the SWIFT vnetIntegrationSubnetId shares the cluster's VNet (mirrors validateAroHcpSwiftSubnetSameVnet).
- Enforce unique managed identities within a cluster across service/control-plane/data-plane operators (mirrors validateAzureOperatorsAuthenticationManagedIdentitiesUniqueWithinCluster).
- Reject PlatformManaged etcd encryption key management mode; only CustomerManaged is accepted for now (mirrors validateAroHcpEtcdEncryptionDataEncryptionKeyManagementMode).

**Test tooling**

- Integration tests can now stamp a ClusterServiceID deterministically (setClusterServiceID + loadCosmos ServiceProviderCluster) instead of relying on a real mocked PostCluster call, since the frontend no longer makes that call during create.

### Why

Cluster Service no longer returns synchronous 400s for these checks once cluster creation became async, since the frontend no longer waits on a PostCluster response. Moving the checks into frontend admission/validation preserves today's fail-fast behavior (same error messages CS used to return) instead of silently deferring rejection to a later, harder-to-surface backend failure. These are best-effort checks; races between concurrently-admitted clusters are still possible, same as before.

### Testing

- Unit tests added for each new validation/admission function (internal/admission, internal/validation).
- Updated integration test fixtures/expectations where the new admission/validation now catches requests that previously passed the frontend and failed only later.

See https://github.com/Azure/ARO-HCP/pull/5709 for review of frontend migration in greater detail
See https://github.com/Azure/ARO-HCP/pull/6004 for review of validation changes in greater detail
See https://github.com/Azure/ARO-HCP/pull/4821 for backend changes. These should be rolled out to production before frontend changes are added.

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


internal/admission/admit_cluster.go
```diff
@@ -53,6 +53,16 @@ type ClusterAdmissionContext struct {
 	// ClusterNodePools is the list of node pools belonging to the cluster, used
 	// for minor-version skew checks against the desired cluster version.
 	ClusterNodePools []ClusterAdmissionNodePool
+	// SubscriptionClusters lists cluster documents in the same subscription
+	// (not including the current cluster being admitted), used
+	// for cross-cluster platform resource uniqueness on CREATE.
```
stevekuznetsov, line 62: This is a huge database load on create and racy. Either we enforce this at the DB layer and return the error on attempted create, or we don't do this IMO.

stevekuznetsov: I don't know the CosmosDB schema well - if we can't enforce uniqueness constraints at the schema layer, we need to do it by using a transaction on the create, and a SELECT in there to ensure no other matches exist.

internal/api/enums.go
```diff
@@ -164,13 +164,17 @@ const (
...
+	// filling the historic value, but excluded from ValidEtcdDataEncryptionKeyManagementModeType until
+	// platform-managed etcd encryption is supported.
 	EtcdDataEncryptionKeyManagementModeTypePlatformManaged EtcdDataEncryptionKeyManagementModeType = "PlatformManaged"
 )
 
 var (
 	ValidEtcdDataEncryptionKeyManagementModeType = sets.New[EtcdDataEncryptionKeyManagementModeType](
 		EtcdDataEncryptionKeyManagementModeTypeCustomerManaged,
-		EtcdDataEncryptionKeyManagementModeTypePlatformManaged,
+		// TODO: re-enable once platform-managed etcd encryption is supported.
```
stevekuznetsov, line 176: Why must this PR orphan this feature?

internal/validation/validate_cluster.go
```diff
@@ -726,6 +736,44 @@ func validateUserAssignedIdentitiesProfile(ctx context.Context, op operation.Ope
...
 	errs = append(errs, RestrictedResourceIDWithResourceGroup(ctx, op, fldPath.Child("serviceManagedIdentity"), newObj.ServiceManagedIdentity, safe.Field(oldObj, toUserAssignedIdentitiesServiceManagedIdentity), "Microsoft.ManagedIdentity/userAssignedIdentities")...)
 
+	// Managed identity resource IDs must be unique across control-plane operators,
+	// data-plane operators, and the service managed identity within a cluster
+	errs = append(errs, validateManagedIdentitiesUniqueWithinCluster(fldPath, newObj)...)
+
+	return errs
+}
+
+// validateManagedIdentitiesUniqueWithinCluster ensures that each managed identity
```
stevekuznetsov, line 746: Why do we care?

internal/api/enums.go
miguelsorianod: This PR doesn't orphan the feature. It enforces the only allowed value on the frontend side instead of in CS, which was already enforcing it. See the `Why` section for some detail

internal/validation/validate_cluster.go
miguelsorianod: This moves the validation from CS to RP, which was already enforcing it. See `Why` for some additional context

internal/admission/admit_cluster.go
miguelsorianod: We are aware of the raciness and potential load, which is documented as a best-effort. This is a potential first step to move the check from CS to the RP to unblock the movement of the CS call to async. We have in mind exploring leveraging the DB. In case it races CS would detect it for now. We have some potential options in mind to improve on this that hopefully we can explore sooner than later.

test-integration/utils/databasemutationhelpers/step_httpcreate.go
```diff
@@ -70,6 +72,13 @@ func (l *httpCreateStep) StepID() StepID {
 }
 
 func (l *httpCreateStep) RunTest(ctx context.Context, t *testing.T, stepInput StepInput) {
+	err := integrationutils.EnsureParentClusterServiceID(
```
miguelsorianod: Why is this needed? I am unsure about the correctness of this. This is to interact with HTTP Frontend, and it shouldn't imply setting always the parent cluster service id. If a cluster needs the CSID, use the appropriate artifact and step

internal/admission/admit_cluster.go
stevekuznetsov: If there's no way to get this from the DB, then add a cache to the frontend. N+1 read on CREATE is a terrible thing to do for performance. 

What bad thing happens when our service ends up with two clusters with the same MRG? If nothing bad happens, why bother checking?

internal/validation/validate_cluster.go
stevekuznetsov: No, but why do these need to be unique? What's the point of the check?

internal/api/enums.go
stevekuznetsov: Thanks, was not clear

internal/admission/admit_cluster.go
miguelsorianod: > If there's no way to get this from the DB

We are getting it from the DB

> What bad thing happens when our service ends up with two clusters with the same MRG? If nothing bad happens, why bother checking?

This is to move the check from CS to frontend, because of the context in `Why`, catching most of the cases which we deemed as good enough for now.

internal/validation/validate_cluster.go
miguelsorianod: This was discussed on a thread a while ago with details that I do not recall, for now we are simply moving what CS is enforcing,

internal/admission/admit_cluster.go
stevekuznetsov: >We are getting it from the DB

Sorry, I was imprecise. What I meant to say is "if there is no way to get this uniqueness constraint from the database itself, meaning we _have_ to do it client-side and hope we win the race, ..." - so, this PR. N+1 is a no-go, so at minimum let's add the changefeed-driven cache (lister).

> This is to move the check from CS to frontend,

Yes, and I am asking for

1. justification that this is actually a good idea and something we care to do - just because it happened to exist in CS before is strictly not enough to justify this
2. if the duplicates are fatal for a cluster, then this PR or follow up work is needed to add a controller to the backend that would detect races that lost in the frontend and expose it in status

internal/validation/validate_cluster.go
stevekuznetsov: At minimum, could we please add a GoDoc comment for why unique identities are needed? I actually remember that thread going exactly the other way, and would appreciate it if someone suggesting to merge this actually justifies their need.

miguelsorianod: As discussed through other channels:

@JakobGray let's add a code comment mentioning that we are going to revisit this after GA, linking the https://redhat.atlassian.net/browse/ARO-21615 jira.

We can put a summary of a conversation that happened around it: the last discussions were around the will to remove the restriction, but we needed to perform a set of verifications and decisions to ensure that would be feasible and work with our requirements. Work started on it but priorities shifted and we need to retake that

test-integration/utils/databasemutationhelpers/step_httpcreate.go
JakobGray: It exists because the frontend still gates node pool/external auth create when `parent.ClusterServiceID == nil`, while cluster create no longer sets that ID synchronously. We can remove it and add more artifacts though

internal/admission/admit_cluster.go
deads2k: I accept the changefeed based cache is a good improvement. For our current scale predicted for the rest of 2026, this will be tolerable.  We will improve later this  year.

miguelsorianod:
As a note just in case: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6082/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2077606785388122112 we had an e2e passing before further changes in the integration-tests.

internal/admission/admit_cluster.go
```diff
@@ -267,6 +232,143 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
...
+		if existingSubnet == nil {
+			errs = append(errs, field.InternalError(subnetPath, errors.New("existing cluster is missing subnetId")))
+			continue
+		}
+		if strings.EqualFold(subnetID, existingSubnet.String()) {
+			errs = append(errs, field.Invalid(
+				subnetPath,
+				subnetID,
+				fmt.Sprintf("Subnet '%s' is already in use by another cluster", subnetID),
+			))
```
deads2k, line 361: why just reuse instead of overlapping?

miguelsorianod:
/hold we can't merge until the phase1 PR is rolled out to all prod regions.

This PR depends on rollout of phase1 PR 50e62e128 (Wed Jul 8 20:52:35 2026 +0000).

At the moment of writing this (2026-07-16):
* the prod regions that have the latest PR commit have b8aa3e0 (Mon Jul 13 06:36:17 2026 +0000)
* some other prod regions still have an older PR commit ec119d1 (Tue Jul 7 14:27:47 2026 +0200)

JakobGray:
Work merged in https://github.com/Azure/ARO-HCP/pull/6121

miguelsorianod:

Closing this.

PR #6083, opened 2026-07-14
stop mutating customer ingress

deads2k:
/hold

This cannot merge until we have a test that demonstrates how a customer can create their own certificate and manage it on the HostedCluster directly.


deads2k:
If this works, we expect to see only the custom ingress testing fail.

bennerv:
/lgtm

Verified testing in a dev environment.  results [here](https://gist.github.com/bennerv/dc284f98db46fa8f884cc15b953bbc8d)

Still held on: testing a certificate issued through key vault (with a digicert CA) and then reflected into the cluster via ESO or csi-secrets-store.

PR #6088, opened 2026-07-15
Fix: Rbac least privilege scoping

hlipsig:
ARO-28263  
  ARO-28264 

### What

 Velero ClusterRole replaces wildcard with explicit verb list
hypershift-installer made SA and ClusterRoleBinding ephemeral via Helm hooks

### Why
Follows principal of least privilege.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


weshayutin:
@shubham-pampattiwar @kaovilai @sseago  please help @hlipsig with getting the appropiate settings for just deploying velero .

shubham-pampattiwar (commented):
One thing to flag from the Velero side, the install Job runs `velero install --dry-run -o yaml`, which always generates a ClusterRoleBinding that binds the velero SA to cluster-admin.
Since Kubernetes RBAC is additive, the velero SA ends up with cluster-admin at runtime regardless of the scoping in velero-cluster-role. The verb tightening in this PR is correct in isolation, but it doesn't change the effective runtime permissions.

To get the actual least-privilege benefit, you'd need to strip the cluster-admin ClusterRoleBinding from the generated manifest before applying it (for example, in the kustomize patch or the install script).

Let me know if my understanding is correct.

hlipsig:
> One thing to flag from the Velero side, the install Job runs `velero install --dry-run -o yaml`, which always generates a ClusterRoleBinding that binds the velero SA to cluster-admin. Since Kubernetes RBAC is additive, the velero SA ends up with cluster-admin at runtime regardless of the scoping in velero-cluster-role. The verb tightening in this PR is correct in isolation, but it doesn't change the effective runtime permissions.
> 
> To get the actual least-privilege benefit, you'd need to strip the cluster-admin ClusterRoleBinding from the generated manifest before applying it (for example, in the kustomize patch or the install script).
> 
> Let me know if my understanding is correct.

@shubham-pampattiwar Would that be a breaking change for velero? Is there anything it does that explicitly requires cluster-admin?

shubham-pampattiwar:
> > One thing to flag from the Velero side, the install Job runs `velero install --dry-run -o yaml`, which always generates a ClusterRoleBinding that binds the velero SA to cluster-admin. Since Kubernetes RBAC is additive, the velero SA ends up with cluster-admin at runtime regardless of the scoping in velero-cluster-role. The verb tightening in this PR is correct in isolation, but it doesn't change the effective runtime permissions.
> > To get the actual least-privilege benefit, you'd need to strip the cluster-admin ClusterRoleBinding from the generated manifest before applying it (for example, in the kustomize patch or the install script).
> > Let me know if my understanding is correct.
> 
> @shubham-pampattiwar Would that be a breaking change for velero? Is there anything it does that explicitly requires cluster-admin?

Velero doesn't explicitly require cluster-admin, it uses it as a default because Velero needs broad permissions to backup and restore arbitrary resources across namespaces. But it's not a hard requirement.

OADP has never used cluster-admin for the Velero SA. We deploy Velero with scoped permissions through the operator's CSV and it works fine. So stripping the cluster-admin binding from the generated manifest would not be a breaking change for Velero.

The scoped velero-cluster-role you already have in the Helm chart covers what Velero needs for your use case. If you strip the cluster-admin binding, Velero will run with just that scoped role. The main thing to watch for is making sure the scoped role covers all the resource types you need to backup and restore, if you hit a Forbidden error during backup/restore, it means a permission is missing from the scoped role.

twolff-gh:
> > > One thing to flag from the Velero side, the install Job runs `velero install --dry-run -o yaml`, which always generates a ClusterRoleBinding that binds the velero SA to cluster-admin. Since Kubernetes RBAC is additive, the velero SA ends up with cluster-admin at runtime regardless of the scoping in velero-cluster-role. The verb tightening in this PR is correct in isolation, but it doesn't change the effective runtime permissions.
> > > To get the actual least-privilege benefit, you'd need to strip the cluster-admin ClusterRoleBinding from the generated manifest before applying it (for example, in the kustomize patch or the install script).
> > > Let me know if my understanding is correct.
> > 
> > 
> > @shubham-pampattiwar Would that be a breaking change for velero? Is there anything it does that explicitly requires cluster-admin?
> 
> Velero doesn't explicitly require cluster-admin, it uses it as a default because Velero needs broad permissions to backup and restore arbitrary resources across namespaces. But it's not a hard requirement.
> 
> OADP has never used cluster-admin for the Velero SA. We deploy Velero with scoped permissions through the operator's CSV and it works fine. So stripping the cluster-admin binding from the generated manifest would not be a breaking change for Velero.
> 
> The scoped velero-cluster-role you already have in the Helm chart covers what Velero needs for your use case. If you strip the cluster-admin binding, Velero will run with just that scoped role. The main thing to watch for is making sure the scoped role covers all the resource types you need to backup and restore, if you hit a Forbidden error during backup/restore, it means a permission is missing from the scoped role.

@hlipsig Was there anything left to do for this here?

hlipsig:
@twolff-gh not exactly from this PR perspective. @tony-schndr let me know this wasn't fully implemented and we should keep in mind. for the second half. I don't know if that's done yet or not to be able to test.

PR #6090, opened 2026-07-15
fix: validate node pool osDisk sizeGiB maximum by disk type

oshrirao:
Jira issue :
https://redhat.atlassian.net/browse/ARO-22496

### What

- Add maximum sizeGiB validation for node pool OS disks: 4095 GiB for Managed (persistent) disks and 2040 GiB for Ephemeral disks
- Document limits in TypeSpec and OpenAPI (@maxValue(4095) with description noting the ephemeral cap)
- Add unit tests for at-max and over-max cases for both disk types

### Why

Users could create node pools with sizeGiB values above Azure’s limits (e.g. 5000 with StandardSSD_LRS). The request passed RP validation and failed later during VM provisioning with Azure errors like:

  - Managed: The value 5000 of parameter 'osDisk.diskSizeGB' is out of range. The value must be between '1' and '4095', inclusive.
  - Ephemeral: The value 4000 of parameter 'osDisk.diskSizeGB' is out of range. The value must be between '1' and '2040', inclusive.

This change rejects invalid sizes at the ARM API during node pool create and preflight, so customers get an immediate validation error instead of an async provisioning failure.

Limits align with Azure documentation:

  - Managed OS disk max: 4,095 GiB — [Managed disks overview (OS disk)](https://learn.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview#os-disk)
  - Ephemeral OS disk max: 2,040 GiB (or VM local cache/temp/NVMe size, whichever is smaller) — [Ephemeral OS disks](https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks)

Fixes [ARO-22496](https://redhat.atlassian.net/browse/ARO-22496)

### Testing
- Added unit tests and followed below plan:

1. go test -run TestValidateNodePoolCreate ./internal/validation/...

 -  Managed at max (4095) — valid 
 -  Managed over max (4096) — rejected
 -  Ephemeral at max (2040) — valid
 -  Ephemeral over max (2041) — rejected

2. Manual test scenarios:

 -  sizeGiB: 5000 (Managed) → must be less than or equal to 4095
 -  sizeGiB: 4000 (Ephemeral) → must be less than or equal to 2040
 -  Valid sizes → preflight Succeeded

3.  make -C api swagger — OpenAPI regenerated and matches TypeSpec
 
### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
+++ b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
@@ -831,9 +831,13 @@ model NodePoolPlatformProfile {
 
 /** The settings and configuration options for OSDisk */
 model OsDiskProfile {
-  /** The OS disk size in GiB */
+  /** The OS disk size in GiB. Maximum is 4095 GiB for Managed disks. For Ephemeral disks,
+   * the maximum is 2040 GiB; Azure may enforce a lower effective limit based on the
+   * selected VM size's local cache, temp, or NVMe capacity.
+   */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   @minValue(64)
+  @maxValue(4095)
   sizeGiB?: int32 = 64;
 
   /** The type of the disk storage account
--- a/internal/validation/validate_nodepools.go
+++ b/internal/validation/validate_nodepools.go
@@ -38,6 +38,17 @@ const (
 	nodePoolK8sLabelKeyNodeRoleWorker = "node-role.kubernetes.io/worker"
 	nodePoolK8sLabelKeyMachineRole    = "machine.openshift.io/cluster-api-machine-role"
 	nodePoolK8sLabelKeyMachineType    = "machine.openshift.io/cluster-api-machine-type"
+
+	// MaxManagedOSDiskSizeGiB is the maximum allowed OS disk size (GiB) for Managed (persistent) OS disks.
+	// See https://learn.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview#os-disk
+	MaxManagedOSDiskSizeGiB int32 = 4095
+	// MaxEphemeralOSDiskSizeGiB is the absolute maximum allowed OS disk size (GiB) for Ephemeral OS disks.
+	// Azure may enforce a lower effective limit per VM size based on local cache/temp/NVMe capacity.
+	// Values above this absolute cap are rejected here; values within this cap but above the selected
+	// VM size's local disk capacity are accepted by this validator and fail later during Azure
+	// provisioning.
+	// See https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks
+	MaxEphemeralOSDiskSizeGiB int32 = 2040
 )
 
 // nodePoolForbiddenK8sLabelValuesByKey maps Kubernetes label keys to forbidden values.
@@ -331,6 +342,13 @@ func validateOSDiskProfile(ctx context.Context, op operation.Operation, fldPath 
 	//DiskType               OsDiskType             `json:"diskType"`
 	errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("diskType"), &newObj.DiskType, safe.Field(oldObj, toOSDiskProfileDiskType))...)
 	errs = append(errs, validate.Enum(ctx, op, fldPath.Child("diskType"), &newObj.DiskType, safe.Field(oldObj, toOSDiskProfileDiskType), api.ValidOsDiskTypes, nil)...)
+
+	switch newObj.DiskType {
+	case api.OsDiskTypeManaged:
+		errs = append(errs, Maximum(ctx, op, fldPath.Child("sizeGiB"), newObj.SizeGiB, safe.Field(oldObj, toOSDiskProfileSizeGiB), MaxManagedOSDiskSizeGiB)...)
+	case api.OsDiskTypeEphemeral:
+		errs = append(errs, Maximum(ctx, op, fldPath.Child("sizeGiB"), newObj.SizeGiB, safe.Field(oldObj, toOSDiskProfileSizeGiB), MaxEphemeralOSDiskSizeGiB)...)
+	}
 
 	//EncryptionSetID        string                 `json:"encryptionSetId,omitempty"`
 	errs = append(errs, RestrictedResourceIDWithResourceGroup(ctx, op, fldPath.Child("encryptionSetId"), newObj.EncryptionSetID, safe.Field(oldObj, toOSDiskProfileEncryptionSetID), "Microsoft.Compute/diskEncryptionSets")...)
--- a/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2024-06-10-preview/openapi.json
+++ b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2024-06-10-preview/openapi.json
@@ -3062,9 +3062,10 @@
         "sizeGiB": {
           "type": "integer",
           "format": "int32",
-          "description": "The OS disk size in GiB",
+          "description": "The OS disk size in GiB. Maximum is 4095 GiB for Managed disks. For Ephemeral disks,\nthe maximum is 2040 GiB; Azure may enforce a lower effective limit based on the\nselected VM size's local cache, temp, or NVMe capacity.",
           "default": 64,
           "minimum": 64,
+          "maximum": 4095,
           "x-ms-mutability": [
             "read",
             "create"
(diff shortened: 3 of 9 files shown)


internal/validation/validate_nodepools.go
```diff
@@ -38,6 +38,14 @@ const (
 	nodePoolK8sLabelKeyNodeRoleWorker = "node-role.kubernetes.io/worker"
 	nodePoolK8sLabelKeyMachineRole    = "machine.openshift.io/cluster-api-machine-role"
 	nodePoolK8sLabelKeyMachineType    = "machine.openshift.io/cluster-api-machine-type"
+
+	// MaxManagedOSDiskSizeGiB is the maximum allowed OS disk size (GiB) for Managed (persistent) OS disks.
+	// See https://learn.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview#os-disk
+	MaxManagedOSDiskSizeGiB int32 = 4095
+	// MaxEphemeralOSDiskSizeGiB is the maximum allowed OS disk size (GiB) for Ephemeral OS disks.
+	// Azure may enforce a lower effective limit per VM size based on local cache/temp/NVMe capacity.
```
miguelsorianod, line 46: > Azure may enforce a lower effective limit per VM size based on local cache/temp/NVMe capacity

Does that mean that if those lower effective limits occur the request would go through and it would fail later?

If so, document

internal/validation/validate_nodepools.go
```diff
@@ -38,6 +38,14 @@ const (
 	nodePoolK8sLabelKeyNodeRoleWorker = "node-role.kubernetes.io/worker"
 	nodePoolK8sLabelKeyMachineRole    = "machine.openshift.io/cluster-api-machine-role"
 	nodePoolK8sLabelKeyMachineType    = "machine.openshift.io/cluster-api-machine-type"
+
+	// MaxManagedOSDiskSizeGiB is the maximum allowed OS disk size (GiB) for Managed (persistent) OS disks.
+	// See https://learn.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview#os-disk
+	MaxManagedOSDiskSizeGiB int32 = 4095
```
miguelsorianod, line 44: The doc mentions that some might have a lower effective limit. In that case, would the request go through and fail later? if so, document

internal/validation/validate_nodepools.go
```diff
@@ -332,6 +340,12 @@ func validateOSDiskProfile(ctx context.Context, op operation.Operation, fldPath
 	errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("diskType"), &newObj.DiskType, safe.Field(oldObj, toOSDiskProfileDiskType))...)
 	errs = append(errs, validate.Enum(ctx, op, fldPath.Child("diskType"), &newObj.DiskType, safe.Field(oldObj, toOSDiskProfileDiskType), api.ValidOsDiskTypes, nil)...)
 
+	maxOSDiskSizeGiB := MaxManagedOSDiskSizeGiB
```
miguelsorianod: Do not assume it will be managed if it's not ephemeral. Add two different checks.

internal/validation/validate_nodepools.go
oshrirao: Yes, it will fail later, so documented here, thanks.

internal/validation/validate_nodepools.go
oshrirao: Yes added different checks. Thanks

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -831,9 +831,13 @@ model NodePoolPlatformProfile {
...
 /** The settings and configuration options for OSDisk */
 model OsDiskProfile {
-  /** The OS disk size in GiB */
+  /** The OS disk size in GiB. Maximum is 4095 GiB for Managed disks. For Ephemeral disks,
+   * the maximum is 2040 GiB; Azure may enforce a lower effective limit based on the
+   * selected VM size's local cache, temp, or NVMe capacity.
+   */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   @minValue(64)
+  @maxValue(4095)
```
mbarnes, line 840: This needs to be versioned so as not to change the Swagger for older API versions.  Do it like this:

```suggestion
  /** The OS disk size in GiB */
  @removed(Versions.v2026_06_30_preview)
  @renamedFrom(Versions.v2026_06_30_preview, "sizeGiB")
  @visibility(Lifecycle.Read, Lifecycle.Create)
  @minValue(64)
  noMaxSizeGiB?: int32 = 64;

  /** The OS disk size in GiB. Maximum is 4095 GiB for Managed disks. For Ephemeral disks,
   * the maximum is 2040 GiB; Azure may enforce a lower effective limit based on the
   * selected VM size's local cache, temp, or NVMe capacity.
   */
  @added(Versions.v2026_06_30_preview)
  @visibility(Lifecycle.Read, Lifecycle.Create)
  @minValue(64)
  @maxValue(4095)
```

mbarnes (changes requested):

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
oshrirao: Implemented the suggested changes. Thanks for the suggestion @mbarnes !

mbarnes (approved):
Informal approval to clear my requested TypeSpec change.
I see an outstanding documentation request from Miguel yet.

internal/validation/validate_nodepools.go
oshrirao: Yes, it will fail later, so documented here, thanks.

shitaljante:
/hold 
Could you please fix the failure? - "Uncommitted changes detected. Please run 'make generate' and 'make fmt', then commit."

oshrirao:

shitaljante:
/unhold 
changes LGTM.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -884,8 +832,20 @@ model NodePoolPlatformProfile {
 /** The settings and configuration options for OSDisk */
 model OsDiskProfile {
   /** The OS disk size in GiB */
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "sizeGiB")
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  @minValue(64)
+  noMaxSizeGiB?: int32 = 64;
```
miguelsorianod: If we are adding these validations in v20260630 api version only (which was requested by @mbarnes in https://github.com/Azure/ARO-HCP/pull/6090#discussion_r3614381821), doesn't that mean that we should only enforce the limits on that version? Right now the validation logic does it independently on the api version.

On the other hand, those seem to be hard limits for which if you pass that threshold you get effectively a non functional nodepool.

I think the options we have are:
* Change the validation to only enforce it in the api version we document the validations are introduced
* Keep the validation as is, but in that case we should also update the openapi of the older api versions

Because it's something that results in a non functional nodepool I am inclined towarsd maybe we can go ahead with retroactively imposing the limit on older api versions (modifying their openapi definitions etc.)

miguelsorianod: I would like the view from people more familiar with the api release process: @mbarnes , @bennerv ?

bennerv: If it's impossible to create a nodepool with a size greater than what is advertised my vote is to apply it uniformly to all versions regardless of API version used by the customer.

oshrirao: Applied changes to all api versions, Thanks for the confirmation @bennerv @miguelsorianod !
@bennerv @miguelsorianod Could you PTAL?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -883,9 +831,13 @@ model NodePoolPlatformProfile {
 
 /** The settings and configuration options for OSDisk */
 model OsDiskProfile {
-  /** The OS disk size in GiB */
+  /** The OS disk size in GiB. Maximum is 4095 GiB for Managed disks. For Ephemeral disks,
```
bennerv, line 834: I'm not super fond of putting the limitations in the description, rather as decorator expressions here, but I understand why we put it here since it varies based on the disk type.

bennerv:
/lgtm
/approve
/retest

Testing failing on build cluster due to insufficient quota.  if it continues, we should raise a thread with the testplatform folks.

PR #6092, opened 2026-07-15
observability: Add cluster deletion SLO, prometheus alerting rules and grafana dashboard (ARO-26216)

kartikaysharma27:
Adds Prometheus alerting and recording rules for Cluster Deletion monitoring as part of [ARO-26216](https://redhat.atlassian.net/browse/ARO-26216). Rules follow [ADR-001: ARO HCP Alerting Recommendation](https://github.com/openshift-online/architecture/pull/79).

runbook_url: aka.ms/arohcp-runbook-cluster-deletion — placeholder short-link; will be pointed to the TSG once published.

Alerts

userJourneyClusterDeletionErrors1h5m — Fast burn, >72% failure rate for 5m (14.4× budget burn)
userJourneyClusterDeletionErrors6h30m — Medium burn, >30% failure rate for 30m (6× budget burn)
userJourneyClusterDeletionErrors3d — Slow burn, >5% failure rate for 6h (1× budget burn)
userJourneyClusterDeletionErrorsDegradation — Early warning, >15% failure rate for 30m
userJourneyClusterDeletionLatency1h5m — Timeliness fast burn, >72% of deletes exceed 30m for 5m
userJourneyClusterDeletionLatency6h30m — Timeliness medium burn, >30% of deletes exceed 30m for 30m
userJourneyClusterDeletionLatency3d — Timeliness slow burn, >5% of deletes exceed 30m for 6h
userJourneyClusterDeletionStuckOperation — Operation stuck in same phase >1 hour; HA deduplication via max without(prometheus_replica)
userJourneyClusterDeletionQueueDepth — Workqueue depth >10 for 5m
All alerts use CEN numeric severity (Sev 3: Errors1h5m, Errors6h30m, Latency1h5m, Latency6h30m (fast + medium burns) Sev 4: Errors3d, ErrorsDegradation, Latency3d, StuckOperation, QueueDepth (slow burns + stuck + saturation)) and include correlationId per [ADR-001](https://redhat.atlassian.net/browse/ADR-001).

Recording Rules

errors:backend_cluster_deletion_operation:error_rate
errors:backend_cluster_deletion_operation:latency_error_rate
Registered in recording-rules-services.yaml and regenerated into generatedRecordingRules.bicep.

Dashboard

Grafana dashboard cluster-deletion.json — fleet-wide SLO visibility covering success SLO, timeliness SLO, stuck operations, duration percentiles (p50/p95/p99), failure ratio trend, and per-subscription breakdown. Includes cluster template variable for per-cluster drill-down.

Testing

25 promtool test scenarios covering all 9 alerts (fire + no-fire pairs), recording rule math, and a combined healthy baseline.
make -C tooling/prometheus-rules run-rp-services and run-recording-rules-services pass. Generated Bicep committed.

Note: HCPdeletionStuck-prometheusRule.yaml is a pre-existing unregistered legacy file. userJourneyClusterDeletionStuckOperation supersedes it with a 1h threshold and CEN-compliant severity.
Jira: [ARO-26216](https://redhat.atlassian.net/browse/ARO-26216)
ADR: ARO HCP Alerting Recommendation ([ADR-001](https://redhat.atlassian.net/browse/ADR-001))

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
@@ -1,3 +1,4 @@
+
 param azureMonitoring string
 
 param location string = resourceGroup().location
@@ -49,6 +50,28 @@ resource arohcpClusterProvisionSloRecordingRules 'Microsoft.AlertsManagement/pro
       {
         record: 'errors:backend_cluster_provision:error_rate'
         expression: '(count by (cluster) (backend_resource_operation_phase_info{operation_type="create",phase="failed",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"}) or 0 * count by (cluster) (backend_resource_operation_phase_info{operation_type="create",phase=~"succeeded|failed|canceled",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"})) / clamp_min(count by (cluster) (backend_resource_operation_phase_info{operation_type="create",phase=~"succeeded|failed|canceled",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"}), 1)'
+      }
+    ]
+  }
+}
+
+resource arohcpClusterDeletionSloRecordingRules 'Microsoft.AlertsManagement/prometheusRuleGroups@2023-03-01' = {
+  name: 'arohcp_cluster_deletion_slo_recording_rules'
+  location: location
+  properties: {
+    scopes: [
+      azureMonitoring
+    ]
+    enabled: true
+    interval: 'PT1M'
+    rules: [
+      {
+        record: 'errors:backend_cluster_deletion_operation:error_rate'
+        expression: '(count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase=~"failed|canceled",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"}) or 0 * count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase=~"succeeded|failed|canceled",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"})) / clamp_min(count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase=~"succeeded|failed|canceled",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"}), 1)'
+      }
+      {
+        record: 'errors:backend_cluster_deletion_operation:timeliness_error_rate'
+        expression: '(count by (cluster) ((backend_resource_operation_last_transition_time_seconds{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"} - backend_resource_operation_start_time_seconds{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"}) >= 1800 and backend_resource_operation_phase_info{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"} == 1) or 0 * count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"} == 1)) / clamp_min(count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"} == 1), 1)'
       }
     ]
   }
--- a/observability/alerts-rp-services.yaml
+++ b/observability/alerts-rp-services.yaml
@@ -3,6 +3,7 @@ prometheusRules:
   - alerts/access-cluster-slo-prometheusRule.yaml
   - alerts/cluster-provision-slo-prometheusRule.yaml
   - alerts/nodepool-slo-prometheusRule.yaml
+  - alerts/HCPClusterDeletion-prometheusRule.yaml
   untestedRules: []
   testDependencies:
   - recording-rules-services.yaml
--- a/observability/alerts/HCPClusterDeletion-prometheusRule.yaml
+++ b/observability/alerts/HCPClusterDeletion-prometheusRule.yaml
@@ -1,0 +1,329 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    app.kubernetes.io/name: kube-prometheus
+    app.kubernetes.io/part-of: kube-prometheus
+    prometheus: k8s
+    role: alert-rules
+  name: cluster-deletion-slo-rules
+  namespace: monitoring
+spec:
+  groups:
+  # ========================================
+  # Cluster Deletion SLO Recording Rules
+  # ========================================
+  # These rules track the Errors and Timeliness SLOs for cluster delete operations.
+  # SLO: 95% of cluster delete operations should succeed (not reach "failed" or "canceled" terminal phase)
+  # Error Budget: 5% (1 - 0.95)
+  # Scope: covers cluster delete operations only.
+  - name: arohcp_cluster_deletion_slo_recording_rules
+    interval: 1m
+    rules:
+    # Instantaneous error rate: fraction of terminal operations that failed or were canceled.
+    # Uses count by (cluster) to preserve the cluster label per nodepool production pattern.
+    # The "or 0 * count by (cluster)" fallback returns zero with the cluster label when
+    # there are no failures, avoiding lost label context (vs "or vector(0)").
+    - record: errors:backend_cluster_deletion_operation:error_rate
+      expr: |
+        (
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"failed|canceled"
+          })
+          or
+          0 * count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          })
+        )
+        /
+        clamp_min(
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          }),
+          1
+        )
+
+    # Timeliness error rate: fraction of succeeded operations that exceeded the 30-minute target.
+    # Uses count by (cluster) to preserve the cluster label per nodepool production pattern.
+    - record: errors:backend_cluster_deletion_operation:timeliness_error_rate
+      expr: |
+        (
+          count by (cluster) (
+            (
+              backend_resource_operation_last_transition_time_seconds{
+                resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+                operation_type="delete",
+                phase="succeeded"
+              }
+              -
+              backend_resource_operation_start_time_seconds{
+                resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+                operation_type="delete",
+                phase="succeeded"
+              }
+            ) >= 1800
+            and
+            backend_resource_operation_phase_info{
+              resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+              operation_type="delete",
+              phase="succeeded"
+            } == 1
+          )
+          or
+          0 * count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase="succeeded"
+          } == 1)
+        )
+        /
+        clamp_min(
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase="succeeded"
+          } == 1),
+          1
+        )
+
+  # ========================================
+  # Cluster Deletion SLO Alerts — Errors
+  # ========================================
+  # Multi-window burn-rate-style alerting adapted for gauge metrics.
+  #
+  # Burn rate thresholds calculated from 95% SLO (5% error budget):
+  #   Fast (14.4x):  14.4 * 0.05 = 0.72 → 72% error rate
+  #   Medium (6x):   6   * 0.05 = 0.30 → 30% error rate
+  #   Slow (1x):     1   * 0.05 = 0.05 →  5% error rate
+  #
+  # The 'for' duration enforces sustained observation, approximating the
+  # short window of multi-window burn rate alerting.
+  - name: arohcp_cluster_deletion_slo_error_alerts
+    rules:
+    # ----------------------------------------
+    # Fast Burn Alert (info)
+    # ----------------------------------------
+    # 14.4x burn rate → >72% failure rate sustained for 5 minutes
+    # Would exhaust 7-day error budget in ~12 hours
+    - alert: userJourneyClusterDeletionErrors1h5m
+      expr: |
+        errors:backend_cluster_deletion_operation:error_rate > 0.72
+        and
+        clamp_min(
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          }) or vector(0),
+          0
+        ) >= 5
+      for: 5m
+      labels:
+        severity: info
+        slo: cluster-deletion-errors
+        short_window: 5m
+        long_window: 1h
+      annotations:
+        title: "Cluster deletion operation error rate high (>72%)"
+        summary: "Cluster deletion operation error rate high (>72%)"
+        description: "More than 72% of cluster delete operations are in failed or canceled state, indicating a fast error budget burn (14.4x) that would exhaust the 95% SLO budget in ~12 hours."
+        runbook_url: "TBD"
+
+    # ----------------------------------------
+    # Medium Burn Alert (info)
+    # ----------------------------------------
+    # 6x burn rate → >30% failure rate sustained for 30 minutes
+    # Would exhaust 7-day error budget in ~28 hours
+    - alert: userJourneyClusterDeletionErrors6h30m
+      expr: |
+        errors:backend_cluster_deletion_operation:error_rate > 0.30
+        and
+        clamp_min(
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          }) or vector(0),
+          0
+        ) >= 5
+      for: 30m
+      labels:
+        severity: info
+        slo: cluster-deletion-errors
+        short_window: 30m
+        long_window: 6h
+      annotations:
+        title: "Cluster deletion operation error rate elevated (>30%) for 30+ minutes"
+        summary: "Cluster deletion operation error rate elevated (>30%) for 30+ minutes"
+        description: "More than 30% of cluster delete operations are in failed or canceled state sustained over 30 minutes, indicating a medium error budget burn (6x) that would exhaust the 95% SLO budget in ~28 hours."
+        runbook_url: "TBD"
+
+    # ----------------------------------------
+    # Slow Burn Alert (info — creates Jira ticket, no page)
+    # ----------------------------------------
+    # Slow burn alerts generate Jira ticket for investigation.
+    # 1x burn rate → >5% failure rate sustained for 6 hours
+    # Would exhaust 7-day error budget in ~7 days
+    - alert: userJourneyClusterDeletionErrors3d
+      expr: |
+        errors:backend_cluster_deletion_operation:error_rate > 0.05
+        and
+        clamp_min(
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          }) or vector(0),
+          0
+        ) >= 5
+      for: 6h
+      labels:
+        severity: info
+        slo: cluster-deletion-errors
+        long_window: 3d
+      annotations:
+        title: "Cluster deletion operation error rate exceeds SLO target (>5%) for 6+ hours"
+        summary: "Cluster deletion operation error rate exceeds SLO target (>5%) for 6+ hours"
+        description: "More than 5% of cluster delete operations are in failed or canceled state sustained over 6 hours, indicating persistent degradation at the 95% SLO boundary that would exhaust the error budget in ~7 days."
+        runbook_url: "TBD"
+
+    # ----------------------------------------
+    # Degradation Alert (info)
+    # ----------------------------------------
+    # With a 95% SLO, the fast-burn threshold only fires at ~72% failure rate.
+    # This provides early warning at 15% failure rate before burn-rate alerts fire.
+    - alert: userJourneyClusterDeletionErrorsDegradation
+      expr: |
+        errors:backend_cluster_deletion_operation:error_rate > 0.15
+        and
+        clamp_min(
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          }) or vector(0),
+          0
+        ) >= 5
+      for: 30m
+      labels:
+        severity: info
+        slo: cluster-deletion-errors
+      annotations:
+        title: "Cluster deletion operation failure rate exceeds 15% for 30 minutes"
+        summary: "Cluster deletion operation failure rate exceeds 15% for 30 minutes"
+        description: "The cluster deletion operation failure rate has been above 15% for 30 minutes. This provides early warning of degradation before SLO-based burn rate alerts fire."
+        runbook_url: "TBD"
+
+    # ----------------------------------------
+    # Timeliness SLO Breach (info)
+    # ----------------------------------------
+    # Fires when >5% of successful deletes exceeded the 30-minute latency target,
+    # sustained for 30 minutes.
+    - alert: userJourneyClusterDeletionLatencySLOBreach
+      expr: |
+        errors:backend_cluster_deletion_operation:timeliness_error_rate > 0.05
+      for: 30m
+      labels:
+        severity: info
+        slo: cluster-deletion-timeliness
+      annotations:
+        title: "Cluster deletion Timeliness SLO breach: >5% of deletes exceed 30 minutes"
+        summary: "Cluster deletion Timeliness SLO breach: >5% of deletes exceed 30 minutes"
+        description: "More than 5% of successful cluster delete operations took longer than 30 minutes, sustained over 30 minutes. This violates the Timeliness SLO target of ≥95% completion within 30 minutes."
+        runbook_url: "TBD"
+
+    # ----------------------------------------
+    # Stuck Operations Alert (info)
+    # ----------------------------------------
+    # Operations stuck in non-terminal phases are invisible to success/failure SLIs.
+    # Uses last_transition_time_seconds (time in current phase) to match the dashboard
+    # panel exactly and avoid premature firing when earlier phases are still in progress.
+    #
+    # This alert fires when time in the CURRENT phase
+    #     (last_transition_time_seconds) > 1h, monitors all 5 active phases.
+    # 
+    - alert: userJourneyClusterDeletionStuckOperation
+      expr: |
+        (
+          max without(prometheus_replica) (
+            backend_resource_operation_phase_info{
+              resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+              operation_type="delete",
+              phase=~"accepted|awaitingsecret|provisioning|updating|deleting"
+            } == 1
+          )
+          and
+          (
+            time() - max without(prometheus_replica) (
+              backend_resource_operation_last_transition_time_seconds{
+                resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+                operation_type="delete",
+                phase=~"accepted|awaitingsecret|provisioning|updating|deleting"
+              }
+            )
+          ) > 3600
+        )
+      for: 15m
+      labels:
+        severity: info
+        slo: cluster-deletion-stuck
+      annotations:
+        title: "Cluster deletion operation stuck in non-terminal phase for over 1 hour"
+        summary: "Cluster deletion operation stuck in non-terminal phase for over 1 hour"
+        description: "Cluster delete operation for {{ $labels.resource_id }} has been in {{ $labels.phase }} phase for over 1 hour. Stuck operations are invisible to success/failure SLIs and require investigation."
+        runbook_url: "TBD"
+        correlationId: "userJourneyClusterDeletionStuckOperation/{{ $labels.cluster }}/{{ $labels.resource_id }}"
+
+  # ========================================
+  # Cluster Deletion Saturation Alerts — Saturation
+  # ========================================
+  # Threshold-based alerting for component-level metrics per ADR.
+  - name: arohcp_cluster_deletion_saturation_alerts
+    rules:
+    # Queue depth: cluster deletion controller workqueue accumulating work
+    - alert: userJourneyClusterDeletionQueueDepth
+      expr: |
+        max by (name, cluster) (
+          max without(prometheus_replica) (
+            workqueue_depth{namespace="aro-hcp", name="OperationClusterDelete"}
+          )
+        ) > 10
+      for: 5m
+      labels:
+        severity: info
+      annotations:
+        title: "Cluster deletion controller workqueue depth is high"
+        summary: "Cluster deletion controller workqueue depth is high"
+        description: "Cluster deletion controller workqueue {{ $labels.name }} has had a depth > 10 for more than 5 minutes, indicating work is accumulating faster than it can be processed."
+        runbook_url: "TBD"
+
+    # Retry hot loop: high retry ratio on cluster deletion workqueue
+    - alert: userJourneyClusterDeletionRetryHotLoop
+      expr: |
+        (
+          sum by (name, cluster) (
+            max without(prometheus_replica) (
+              rate(workqueue_retries_total{namespace="aro-hcp", name="OperationClusterDelete"}[10m])
+            )
+          )
+          /
+          sum by (name, cluster) (
+            max without(prometheus_replica) (
+              rate(workqueue_adds_total{namespace="aro-hcp", name="OperationClusterDelete"}[10m])
+            )
+          )
+        ) > 0.5
+      for: 10m
+      labels:
+        severity: info
+      annotations:
+        title: "Cluster deletion controller workqueue in retry hot loop"
+        summary: "Cluster deletion controller workqueue in retry hot loop"
+        description: "Cluster deletion controller workqueue {{ $labels.name }} has a retry ratio > 50% sustained over 10 minutes, indicating most queue activity is failed retries rather than fresh work."
+        runbook_url: "TBD"
--- a/observability/alerts/HCPClusterDeletion-prometheusRule.yaml
+++ b/observability/alerts/HCPClusterDeletion-prometheusRule.yaml
@@ -11,88 +11,6 @@ metadata:
 spec:
   groups:
   # ========================================
-  # Cluster Deletion SLO Recording Rules
-  # ========================================
-  # These rules track the Errors and Timeliness SLOs for cluster delete operations.
-  # SLO: 95% of cluster delete operations should succeed (not reach "failed" or "canceled" terminal phase)
-  # Error Budget: 5% (1 - 0.95)
-  # Scope: covers cluster delete operations only.
-  - name: arohcp_cluster_deletion_slo_recording_rules
-    interval: 1m
-    rules:
-    # Instantaneous error rate: fraction of terminal operations that failed or were canceled.
-    # Uses count by (cluster) to preserve the cluster label per nodepool production pattern.
-    # The "or 0 * count by (cluster)" fallback returns zero with the cluster label when
-    # there are no failures, avoiding lost label context (vs "or vector(0)").
-    - record: errors:backend_cluster_deletion_operation:error_rate
-      expr: |
-        (
-          count by (cluster) (backend_resource_operation_phase_info{
-            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
-            operation_type="delete",
-            phase=~"failed|canceled"
-          })
-          or
-          0 * count by (cluster) (backend_resource_operation_phase_info{
-            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
-            operation_type="delete",
-            phase=~"succeeded|failed|canceled"
-          })
-        )
-        /
-        clamp_min(
-          count by (cluster) (backend_resource_operation_phase_info{
-            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
-            operation_type="delete",
-            phase=~"succeeded|failed|canceled"
-          }),
-          1
-        )
-
-    # Timeliness error rate: fraction of succeeded operations that exceeded the 30-minute target.
-    # Uses count by (cluster) to preserve the cluster label per nodepool production pattern.
-    - record: errors:backend_cluster_deletion_operation:timeliness_error_rate
-      expr: |
-        (
-          count by (cluster) (
-            (
-              backend_resource_operation_last_transition_time_seconds{
-                resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
-                operation_type="delete",
-                phase="succeeded"
-              }
-              -
-              backend_resource_operation_start_time_seconds{
-                resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
-                operation_type="delete",
-                phase="succeeded"
-              }
-            ) >= 1800
-            and
-            backend_resource_operation_phase_info{
-              resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
-              operation_type="delete",
-              phase="succeeded"
-            } == 1
-          )
-          or
-          0 * count by (cluster) (backend_resource_operation_phase_info{
-            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
-            operation_type="delete",
-            phase="succeeded"
-          } == 1)
-        )
-        /
-        clamp_min(
-          count by (cluster) (backend_resource_operation_phase_info{
-            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
-            operation_type="delete",
-            phase="succeeded"
-          } == 1),
-          1
-        )
-
-  # ========================================
   # Cluster Deletion SLO Alerts — Errors
   # ========================================
   # Multi-window burn-rate-style alerting adapted for gauge metrics.
@@ -125,16 +43,16 @@ spec:
         ) >= 5
       for: 5m
       labels:
-        severity: info
+        severity: "3"
         slo: cluster-deletion-errors
         short_window: 5m
         long_window: 1h
       annotations:
         title: "Cluster deletion operation error rate high (>72%)"
-        summary: "Cluster deletion operation error rate high (>72%)"
+        summary: "{{ $labels.cluster }}: Cluster deletion operation error rate high (>72%)"
         description: "More than 72% of cluster delete operations are in failed or canceled state, indicating a fast error budget burn (14.4x) that would exhaust the 95% SLO budget in ~12 hours."
-        runbook_url: "TBD"
-
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionErrors1h5m/{{ $labels.cluster }}"
     # ----------------------------------------
     # Medium Burn Alert (info)
     # ----------------------------------------
@@ -154,16 +72,16 @@ spec:
         ) >= 5
       for: 30m
       labels:
-        severity: info
+        severity: "3"
         slo: cluster-deletion-errors
         short_window: 30m
         long_window: 6h
       annotations:
         title: "Cluster deletion operation error rate elevated (>30%) for 30+ minutes"
-        summary: "Cluster deletion operation error rate elevated (>30%) for 30+ minutes"
+        summary: "{{ $labels.cluster }}: Cluster deletion operation error rate elevated (>30%) for 30+ minutes"
         description: "More than 30% of cluster delete operations are in failed or canceled state sustained over 30 minutes, indicating a medium error budget burn (6x) that would exhaust the 95% SLO budget in ~28 hours."
-        runbook_url: "TBD"
-
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionErrors6h30m/{{ $labels.cluster }}"
     # ----------------------------------------
     # Slow Burn Alert (info — creates Jira ticket, no page)
     # ----------------------------------------
@@ -184,15 +102,15 @@ spec:
         ) >= 5
       for: 6h
       labels:
-        severity: info
+        severity: "3"
         slo: cluster-deletion-errors
         long_window: 3d
       annotations:
         title: "Cluster deletion operation error rate exceeds SLO target (>5%) for 6+ hours"
-        summary: "Cluster deletion operation error rate exceeds SLO target (>5%) for 6+ hours"
+        summary: "{{ $labels.cluster }}: Cluster deletion operation error rate exceeds SLO target (>5%) for 6+ hours"
         description: "More than 5% of cluster delete operations are in failed or canceled state sustained over 6 hours, indicating persistent degradation at the 95% SLO boundary that would exhaust the error budget in ~7 days."
-        runbook_url: "TBD"
-
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionErrors3d/{{ $labels.cluster }}"
     # ----------------------------------------
     # Degradation Alert (info)
     # ----------------------------------------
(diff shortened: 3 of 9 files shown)


observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,329 @@
+apiVersion: monitoring.coreos.com/v1
```
swiencki, line 1: Recording rules are defined here but not registered in recording-rules-services.yaml or regenerated into generatedRecordingRules.bicep; the deployed alerts will reference metrics that never get created.

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
```diff
@@ -0,0 +1,329 @@
...
+    interval: 1m
+    rules:
+    # Instantaneous error rate: fraction of terminal operations that failed or were canceled.
+    # Uses count by (cluster) to preserve the cluster label per nodepool production pattern.
+    # The "or 0 * count by (cluster)" fallback returns zero with the cluster label when
+    # there are no failures, avoiding lost label context (vs "or vector(0)").
+    - record: errors:backend_cluster_deletion_operation:error_rate
+      expr: |
+        (
+          count by (cluster) (backend_resource_operation_phase_info{
```
swiencki: Alerts use severity: info and runbook_url: "TBD"; use numeric CEN severity ("3"/"4") and point to the cluster deletion TSG per ADR.

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
```diff
@@ -0,0 +1,329 @@
...
+
+    # Timeliness error rate: fraction of succeeded operations that exceeded the 30-minute target.
+    # Uses count by (cluster) to preserve the cluster label per nodepool production pattern.
+    - record: errors:backend_cluster_deletion_operation:timeliness_error_rate
+      expr: |
+        (
+          count by (cluster) (
+            (
+              backend_resource_operation_last_transition_time_seconds{
+                resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
```
swiencki: userJourneyClusterDeletionLatencySLOBreach is a single-threshold alert without MWMBR tiers; add multi-window confirmation or document the exception.

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
```diff
@@ -0,0 +1,1896 @@
+{
```
swiencki, line 1: Dashboard needs a cluster-level variable for drill-down once a problematic delete is identified.

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
```diff
@@ -0,0 +1,329 @@
...
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"failed|canceled"
+          })
+          or
+          0 * count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          })
```
swiencki: Most alerts here omit correlationId annotations; ADR-001 requires correlationId for IcM deduplication on every RP-lane alert.

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
kartikaysharma27: Updated the CEN severity ("3"/"4") .

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
kartikaysharma27: Added it, please check.

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
kartikaysharma27: Added cluster level variable.

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
kartikaysharma27: Added correlationID

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: registered the rules.

observability/alerts-rp-services.yaml
```diff
@@ -3,6 +3,7 @@ prometheusRules:
   - alerts/access-cluster-slo-prometheusRule.yaml
   - alerts/cluster-provision-slo-prometheusRule.yaml
   - alerts/nodepool-slo-prometheusRule.yaml
+  - alerts/HCPClusterDeletion-prometheusRule.yaml
```
ehvs: Lets follow the convention of prometheusrule naming like the previous ones.

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
```diff
@@ -0,0 +1,295 @@
...
+        long_window: 6h
+        component: slo
+      annotations:
+        title: "Cluster deletion operation error rate elevated (>30%) for 30+ minutes"
+        summary: "{{ $labels.cluster }}: Cluster deletion operation error rate elevated (>30%) for 30+ minutes"
+        description: "More than 30% of cluster delete operations are in failed or canceled state sustained over 30 minutes, indicating a medium error budget burn (6x) that would exhaust the 95% SLO budget in ~28 hours."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionErrors6h30m/{{ $labels.cluster }}"
+    # ----------------------------------------
+    # Slow Burn Alert (info — creates Jira ticket, no page)
```
ehvs: "create Jiras ticket" is made automatically? Can you elaborate where this is defined?

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,295 @@
...
+  # ========================================
+  # Cluster Deletion Saturation Alerts — Saturation
+  # ========================================
+  # Threshold-based alerting for component-level metrics per ADR.
+  - name: arohcp_cluster_deletion_saturation_alerts
+    rules:
+    # Queue depth: cluster deletion controller workqueue accumulating work
+    - alert: userJourneyClusterDeletionQueueDepth
+      expr: |
+        max by (name, cluster) (
```
ehvs, line 266: Please review this PR https://github.com/Azure/ARO-HCP/pull/6346, and inspect if your current logic would suffer of the same problem.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,295 @@
...
+        ) > 0.5
+      for: 10m
+      labels:
+        severity: "4"
+        component: slo
+      annotations:
+        title: "Cluster deletion controller workqueue in retry hot loop"
+        summary: "{{ $labels.cluster }}: Cluster deletion controller workqueue {{ $labels.name }} retry hot loop"
+        description: "Cluster deletion controller workqueue {{ $labels.name }} has a retry ratio > 50% sustained over 10 minutes, indicating most queue activity is failed retries rather than fresh work."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
```
ehvs: This does not seem to be resolving to the tsg. 
I checked the TSG from the PR opened, and it is still pending on the new template.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,295 @@
...
+        description: "The cluster deletion operation failure rate has been above 15% for 30 minutes. This provides early warning of degradation before SLO-based burn rate alerts fire."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionErrorsDegradation/{{ $labels.cluster }}"
+    # ----------------------------------------
+    # Timeliness SLO — Fast Burn (info)
+    # ----------------------------------------
+    # 14.4x burn rate → >72% of successful deletes exceeded 30m, sustained 5m
+    - alert: userJourneyClusterDeletionLatency1h5m
+      expr: |
+        errors:backend_cluster_deletion_operation:timeliness_error_rate > 0.72
```
ehvs: This seems to be missing the guard of  `>= 5`. Meaning, it will trigger if 100% of total amount of 1 is failed, instead of considering at least 5.
So they only trigger when there are actually enough deletions for the percentage to mean something.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,295 @@
...
+        description: "More than 30% of successful cluster deletes exceeded 30 minutes, sustained for 30 minutes. Medium burn rate (6x) against the 95% Timeliness SLO — on track to exhaust the budget in ~28 hours."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionLatency6h30m/{{ $labels.cluster }}"
+    # ----------------------------------------
+    # Timeliness SLO — Slow Burn (info)
+    # ----------------------------------------
+    # 1x burn rate → >5% of successful deletes exceeded 30m, sustained 6h
+    - alert: userJourneyClusterDeletionLatency3d
+      expr: |
+        errors:backend_cluster_deletion_operation:timeliness_error_rate > 0.05
```
ehvs: Same here. This seems to be missing the guard of  `>= 5`. Meaning, it will trigger if 100% of total amount of 1 is failed, instead of considering at least 5.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,295 @@
...
+        description: "More than 72% of successful cluster deletes exceeded 30 minutes, sustained for 5 minutes. Fast burn rate (14.4x) against the 95% Timeliness SLO — on track to exhaust the budget in ~12 hours."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionLatency1h5m/{{ $labels.cluster }}"
+    # ----------------------------------------
+    # Timeliness SLO — Medium Burn (info)
+    # ----------------------------------------
+    # 6x burn rate → >30% of successful deletes exceeded 30m, sustained 30m
+    - alert: userJourneyClusterDeletionLatency6h30m
+      expr: |
+        errors:backend_cluster_deletion_operation:timeliness_error_rate > 0.30
```
ehvs: Same here. This seems to be missing the guard of  `>= 5`. Meaning, it will trigger if 100% of total amount of 1 is failed, instead of considering at least 5.

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
```diff
@@ -0,0 +1,295 @@
...
+            phase=~"succeeded|failed|canceled"
+          }) or vector(0),
+          0
+        ) >= 5
+      for: 30m
+      labels:
+        severity: "3"
+        slo: cluster-deletion-errors
+        short_window: 30m
+        long_window: 6h
```
ehvs: I noticed that `long_window` is present on the alert, but I do not see it used anywhere.  Can we drop this variable? Since the expression only accounts for `short_window` and the value in `for`.  
I see the naming also coming with it, but imo this is misleading to what the expression actually is doing.

ehvs (changes requested):

observability/alerts-rp-services.yaml
kartikaysharma27: Done, added naming convention as per previous ones.

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
kartikaysharma27: This was mainly for task purpose, but could be misleading to oncall, removed the line for code.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: aka.ms/arohcp-runbook-cluster-deletion is a placeholder pointing to where the TSG will live once published. Same pattern used in all other SLO alerts. Will update the URL when the TSG is live.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: This is valid check, I added it in code.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: This is valid check, I added it in code.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: This is valid check, I added it in code.

observability/alerts/HCPClusterDeletion-prometheusRule.yaml
kartikaysharma27: removed.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: We already use max without(prometheus_replica) which strips the replica label before the alert evaluates.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,306 @@
...
+        title: "Cluster deletion operation stuck in non-terminal phase for over 1 hour"
+        summary: "{{ $labels.cluster }}: Cluster deletion operation {{ $labels.resource_id }} stuck in {{ $labels.phase }} for over 1 hour"
+        description: "Cluster delete operation for {{ $labels.resource_id }} has been in {{ $labels.phase }} phase for over 1 hour. Stuck operations are invisible to success/failure SLIs and require investigation."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionStuckOperation/{{ $labels.cluster }}/{{ $labels.resource_id }}"
+  # ========================================
+  # Cluster Deletion Saturation Alerts — Saturation
+  # ========================================
+    - name: arohcp_cluster_deletion_saturation_alerts
+    rules:
```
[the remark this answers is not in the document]
kartikaysharma27: Good catch here, fixed the indentation issue.

observability/recording-rules/cluster-deletion-slo-recordingRule.yaml
```diff
@@ -0,0 +1,94 @@
...
+  #
+  # Metrics are gauge-based (backend_resource_operation_phase_info) from Cosmos DB,
+  # so we use count() aggregations rather than rate() on counters.
+  - name: arohcp_cluster_deletion_slo_recording_rules
+    interval: 1m
+    rules:
+    # Instantaneous error rate: fraction of terminal operations that failed or were canceled.
+    # The "or 0 * count by (cluster)" fallback returns zero with the cluster label when
+    # there are no failures, avoiding lost label context (vs "or vector(0)").
+    - record: errors:backend_cluster_deletion_operation:error_rate
```
swiencki, line 36: These alerts evaluate one retained-gauge ratio with only a `for` duration, so can we either calculate the documented long and short windows for each tier or explicitly document that the RP lane is accepting the same gauge approximation used by the existing journeys?

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
```diff
@@ -0,0 +1,1921 @@
...
...openshift/hcpopenshiftclusters\", operation_type=\"delete\", phase=\"failed\", cluster=~\"$cluster\", subscription_id!~\"${exclude_subs:raw}\"}) < 3600)) or vector(0)) / clamp_min(((count((backend_resource_operation_phase_info{resource_type=\"microsoft.redhatopenshift/hcpopenshiftclusters\", operation_type=\"delete\", phase=\"succeeded\", cluster=~\"$cluster\", subscription_id!~\"${exclude_subs:raw}\"} == 1) and ((time() - backend_resource_operation_last_transition_time_seconds{resource_type=\"microsoft.redhatopenshift/hcpopenshiftclusters\", operation_type=\"delete\", phase=\"succeeded\", cluster=~\"$cluster\", subscription_id!~\"${exclude_subs:raw}\"}) < 3600)) or vector(0)) + (count((backend_resource_operation_phase_info{resource_type=\"microsoft.redhatopenshift/hcpopenshiftclusters\", operation_type=\"delete\", phase=\"failed\", cluster=~\"$cluster\", subscription_id!~\"${exclude_subs:raw}\"} == 1) and ((time() - backend_resource_operation_last_transition_time_seconds{resource_type=\"microsoft.redhatopenshift/hcpopenshiftclusters\", operation_type=\"delete\", phase=\"failed\", cluster=~\"$cluster\", subscription_id!~\"${exclude_subs:raw}\"}) < 3600)) or vector(0))), 1)) * 100",
```
swiencki: Please include canceled deletes in both the numerator and denominator of this failure ratio so it matches the recording rule and the dashboard's other failed/canceled panels.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,306 @@
...
+          0
+        ) >= 5
+      for: 5m
+      labels:
+        severity: "3"
+        slo: cluster-deletion-errors
+        component: slo
+      annotations:
+        title: "Cluster deletion operation error rate high (>72%)"
+        summary: "{{ $labels.cluster }}: Cluster deletion operation error rate high (>72%)"
```
swiencki: Please remove `{{ $labels.cluster }}` from every summary because RP generation already prefixes the IcM title with the service cluster, which currently produces the cluster twice.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,306 @@
...
+      labels:
+        severity: "3"
+        slo: cluster-deletion-stuck
+        component: slo
+      annotations:
+        title: "Cluster deletion operation stuck in non-terminal phase for over 1 hour"
+        summary: "{{ $labels.cluster }}: Cluster deletion operation {{ $labels.resource_id }} stuck in {{ $labels.phase }} for over 1 hour"
+        description: "Cluster delete operation for {{ $labels.resource_id }} has been in {{ $labels.phase }} phase for over 1 hour. Stuck operations are invisible to success/failure SLIs and require investigation."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionStuckOperation/{{ $labels.cluster }}/{{ $labels.resource_id }}"
```
swiencki: Please correlate this customer-facing stuck-operation alert by `subscription_id` rather than `resource_id` so suppression works per customer without fragmenting incidents per operation.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,306 @@
...
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          }) or vector(0),
+          0
+        ) >= 5
+      for: 6h
+      labels:
+        severity: "3"
```
swiencki: Please align `Errors3d`, `ErrorsDegradation`, `Latency3d`, and `StuckOperation` to Sev 4 so slow-burn, precursor, and stuck signals match the production RP-lane access and provisioning conventions.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,306 @@
...
+                resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+                operation_type="delete",
+                phase=~"accepted|awaitingsecret|provisioning|updating|deleting"
+              }
+            )
+          ) > 3600
+        )
+      for: 15m
+      labels:
+        severity: "3"
```
swiencki: Please add `exclude_internal_subscriptions: "true"` to the stuck-operation alert because its output retains `subscription_id` and the comparable RP-lane access and node pool alerts exclude internal test operations.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,306 @@
...
+        severity: "4"
+        component: slo
+      annotations:
+        title: "Cluster deletion controller workqueue depth is high"
+        summary: "{{ $labels.cluster }}: Cluster deletion controller workqueue {{ $labels.name }} depth is high"
+        description: "Cluster deletion controller workqueue {{ $labels.name }} has had a depth > 10 for more than 5 minutes, indicating work is accumulating faster than it can be processed."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionQueueDepth/{{ $labels.cluster }}/{{ $labels.name }}"
+    # Retry hot loop: high retry ratio on cluster deletion workqueue
+    - alert: userJourneyClusterDeletionRetryHotLoop
```
swiencki: Can we remove this retry-ratio alert unless production evidence shows that `OperationClusterDelete` retries predict customer-visible impact, since it measures the same generic rate-limiting workqueue behavior removed for node pools and the stuck-operation alert already covers the actionable outcome?

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
```diff
@@ -0,0 +1,1921 @@
...
+          },
+          "editorMode": "code",
+          "expr": "count((backend_resource_operation_phase_info{resource_type=\"microsoft.redhatopenshift/hcpopenshiftclusters\", operation_type=\"delete\", phase=~\"succeeded|failed|canceled\", cluster=~\"$cluster\", subscription_id!~\"${exclude_subs:raw}\"} == 1) and ((time() - backend_resource_operation_last_transition_time_seconds{resource_type=\"microsoft.redhatopenshift/hcpopenshiftclusters\", operation_type=\"delete\", phase=~\"succeeded|failed|canceled\", cluster=~\"$cluster\", subscription_id!~\"${exclude_subs:raw}\"}) < $__range_s)) or vector(0)",
+          "instant": true,
+          "legendFormat": "Completed",
+          "range": false,
+          "refId": "A"
+        }
+      ],
+      "title": "Completed Delete Attempts (Window)",
```
swiencki, line 218: The PR does not currently define an alertable Traffic signal, so can we either add an actionable delete-request surge or throughput signal or document why deletion traffic has no meaningful alert threshold for this journey?

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,306 @@
...
+        description: "Cluster delete operation for {{ $labels.resource_id }} has been in {{ $labels.phase }} phase for over 1 hour. Stuck operations are invisible to success/failure SLIs and require investigation."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionStuckOperation/{{ $labels.cluster }}/{{ $labels.resource_id }}"
+  # ========================================
+  # Cluster Deletion Saturation Alerts — Saturation
+  # ========================================
+  - name: arohcp_cluster_deletion_saturation_alerts
+    rules:
+    # Queue depth: cluster deletion controller workqueue accumulating work
+    - alert: userJourneyClusterDeletionQueueDepth
```
swiencki, line 264: Please add a queue-depth dashboard panel, and retry/adds panels only if the retry alert remains, so responders can inspect the saturation signal that triggered the alert.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,306 @@
...
+      for: 5m
+      labels:
+        severity: "3"
+        slo: cluster-deletion-errors
+        component: slo
+      annotations:
+        title: "Cluster deletion operation error rate high (>72%)"
+        summary: "{{ $labels.cluster }}: Cluster deletion operation error rate high (>72%)"
+        description: "More than 72% of cluster delete operations are in failed or canceled state, indicating a fast error budget burn (14.4x) that would exhaust the 95% SLO budget in ~12 hours."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
```
swiencki, line 53: Before enabling these RP-lane alerts, please ensure the TSG Alert Triage table maps every retained alert name to a dedicated diagnostic section with mitigation and validation steps.

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
kartikaysharma27: Deletion traffic doesn't have a meaningful alertable surge threshold at this stage. The QueueDepth alert (workqueue_depth > 10 for 5m) covers the actionable outcome. We can add a rate-based alert once we have production baseline data to calibrate a meaningful threshold. I will document it on jira as on track work.

observability/recording-rules/cluster-deletion-slo-recordingRule.yaml
kartikaysharma27: RP lane is accepting the same gauge approximation as existing user journeys, documented in cluster-deletion-slo-recordingRule.yaml comment.

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
kartikaysharma27: Updated to phase=~"failed|canceled" in both numerator and denominator, matching the recording rule and other dashboard panels.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Removed {{ $labels.cluster }} prefix from all summaries and titles.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Done ,correlationId updated to userJourneyClusterDeletionStuckOperation/{{ $labels.cluster }}/{{ $labels.subscription_id }}.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Done, Errors3d, ErrorsDegradation, Latency3d, and StuckOperation all set to severity "4".

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Added exclude_internal_subscriptions: "true".

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Following what nodepool alert removed, I removed the retry ratio alert from our deletion code.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Done, added a new panel.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: We have drafted the tsg and it will follow and maps the alerts.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,280 @@
...
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          }) or vector(0),
+          0
+        ) >= 5
+      for: 30m
+      labels:
+        severity: "3"
```
swiencki, line 74: Errors1h5m and Errors6h30m are both Sev3 with distinct correlationIds (same for Latency1h5m/6h30m), so a single sustained burn opens two concurrent Sev3 IcMs in the RP queue; please run the alert tester across the full cluster-deletion set and collapse these to a shared correlationId or one page tier if it does.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,280 @@
...
+        and
+        clamp_min(
+          count by (cluster) (backend_resource_operation_phase_info{
+            resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+            operation_type="delete",
+            phase=~"succeeded|failed|canceled"
+          }) or vector(0),
+          0
+        ) >= 5
+      for: 5m
```
swiencki, line 44: The `count by (cluster) >= 5` guard scopes error paging to management clusters with at least five terminal deletes in-window, a good low-volume noise guard, but quiet clusters then get no error paging, so please confirm that is intended.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,280 @@
...
+        summary: "Cluster deletion Timeliness SLO slow burn (>5% slow for 6h)"
+        description: "More than 5% of successful cluster deletes exceeded 30 minutes, sustained for 6 hours. Slow burn rate (1x) against the 95% Timeliness SLO — budget will be exhausted in ~7 days."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionLatency3d/{{ $labels.cluster }}"
+    # ----------------------------------------
+    # Stuck Operations Alert (info)
+    # ----------------------------------------
+    # This alert fires when time in the CURRENT phase
+    #     (last_transition_time_seconds) > 1h, monitors all 5 active phases.
+    - alert: userJourneyClusterDeletionStuckOperation
```
swiencki, line 225: This is the only stuck-delete signal in the RP lane (the existing BackendAsyncOperationClusterDeleteStuck lives in the SL lane at 25m/Sev3), so please confirm the RP 1h/Sev4 threshold is intentionally a longer, complementary view rather than an accidental duplicate.

observability/recording-rules/cluster-deletion-slo-recordingRule.yaml
```diff
@@ -0,0 +1,101 @@
...
+  # the way true MWMBR does. The `for:` duration in each alert acts as the short window.
+  # As is accepting the same gauge approximation as used by existing user journeys.
+  # ================================================
+  - name: arohcp_cluster_deletion_slo_recording_rules
+    interval: 1m
+    rules:
+    # Instantaneous error rate: fraction of terminal operations that failed or were canceled.
+    # The "or 0 * count by (cluster)" fallback returns zero with the cluster label when
+    # there are no failures, avoiding lost label context (vs "or vector(0)").
+    - record: errors:backend_cluster_deletion_operation:error_rate
```
swiencki, line 36: These SLIs are instantaneous gauge ratios (error-rate and timeliness only) with no delete-throughput signal, so a drop in delete submissions to zero would go undetected by this chain; please confirm a traffic-drop SLI is intentionally out of scope.

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
```diff
@@ -0,0 +1,1999 @@
...
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "count by (phase) (backend_cluster_provision_state{cluster=~\"$cluster\"} == 1)",
+          "instant": false,
+          "legendFormat": "{{phase}}",
```
swiencki, line 1005: The recorded series aggregate `by (cluster)` only, so any panel legending by `{{phase}}` or `{{subscription_id}}` should target the raw gauge or ALERTS instead of the recorded SLI series or those labels render empty.

observability/recording-rules/cluster-deletion-slo-recordingRule.yaml
```diff
@@ -0,0 +1,101 @@
...
+                operation_type="delete",
+                phase="succeeded"
+              }
+              -
+              backend_resource_operation_start_time_seconds{
+                resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+                operation_type="delete",
+                phase="succeeded"
+              }
+            ) >= 1800
```
swiencki: The timeliness SLI counts deletes that finish in exactly 1800 seconds as failures even though the SLO text says exceeded 30 minutes, so use greater-than 1800 if the target is inclusive.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,280 @@
...
+        severity: "4"
+        slo: cluster-deletion-stuck
+        component: slo
+        exclude_internal_subscriptions: "true"
+      annotations:
+        title: "Cluster deletion operation stuck in non-terminal phase for over 1 hour"
+        summary: "Cluster deletion operation {{ $labels.resource_id }} stuck in {{ $labels.phase }} for over 1 hour"
+        description: "Cluster delete operation for {{ $labels.resource_id }} has been in {{ $labels.phase }} phase for over 1 hour. Stuck operations are invisible to success/failure SLIs and require investigation."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionStuckOperation/{{ $labels.cluster }}/{{ $labels.subscription_id }}"
```
swiencki, line 257: The stuck-operation correlationId omits resource_id and phase even though the alert preserves those labels, so multiple stuck deletes in the same cluster or subscription can collapse into one IcM; include resource_id and phase like the summary does.

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
```diff
@@ -54,6 +55,28 @@ resource arohcpClusterProvisionSloRecordingRules 'Microsoft.AlertsManagement/pro
...
+  properties: {
+    scopes: [
+      azureMonitoring
+    ]
+    enabled: true
+    interval: 'PT1M'
+    rules: [
+      {
+        record: 'errors:backend_cluster_deletion_operation:error_rate'
+        expression: '(count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase=~"failed|canceled",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"}) or 0 * count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase=~"succeeded|failed|canceled",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"})) / clamp_min(count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase=~"succeeded|failed|canceled",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"}), 1)'
```
pepedocs, line 70: Is there any reason why a canceled delete is treated as failure? It can succeed too.

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
```diff
@@ -54,6 +55,28 @@ resource arohcpClusterProvisionSloRecordingRules 'Microsoft.AlertsManagement/pro
...
+      }
+      {
+        record: 'errors:backend_cluster_deletion_operation:timeliness_error_rate'
```
pepedocs: Can we please rename this to `latency` rather than `timeliness` to keep it standard across our metrics and SLOs?

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,280 @@
...
+        slo: cluster-deletion-errors
+        component: slo
+      annotations:
+        title: "Cluster deletion operation failure rate exceeds 15% for 30 minutes"
+        summary: "Cluster deletion operation failure rate exceeds 15% for 30 minutes"
+        description: "The cluster deletion operation failure rate has been above 15% for 30 minutes. This provides early warning of degradation before SLO-based burn rate alerts fire."
+        runbook_url: "aka.ms/arohcp-runbook-cluster-deletion"
+        correlationId: "userJourneyClusterDeletionErrorsDegradation/{{ $labels.cluster }}"
+    # ----------------------------------------
+    # Timeliness SLO — Fast Burn (info)
```
pepedocs: Can we use Latency instead of Timeliness to keep this standard? Latency is the jargon for how long something takes, this is used across the industry. Using this common term leverages familiarity. Timeliness can mean more than just latency.

pepedocs (changes requested):

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: A sustained burn now opens one IcM per cluster per dimension.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Intentional. BackendAsyncOperationClusterDeleteStuck (SL lane, 25m/Sev3) is an early internal signal for the component team. The RP lane alert (1h/Sev4) uses a longer window to filter transient stalls that the SL alert catches and recovers from quickly, escalating only when the operation is genuinely hung. Different thresholds, different severity, different audience.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Yes, that's intended,  the threshold is a noise guard to avoid alert fatigue on low-activity clusters. Coverage for quiet clusters is provided by userJourneyClusterDeletionStuckOperation, which has no volume guard and fires on any individual stuck delete.

observability/recording-rules/cluster-deletion-slo-recordingRule.yaml
kartikaysharma27: This PR covers quality SLOs (error rate and latency) on operations that do reach a terminal phase. Traffic-drop monitoring can be added as a follow-up task.

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
kartikaysharma27: the recorded series carry only the cluster label by design. All dashboard panels that legend by {{phase}} or {{subscription_id}} target the raw gauge directly; the recorded SLI series are only used in alert expressions.

observability/recording-rules/cluster-deletion-slo-recordingRule.yaml
kartikaysharma27: Fixed : changed to > 1800. A delete completing in exactly 1800 s has met the 30-minute target, not exceeded it.

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
kartikaysharma27: A canceled delete means the operation stopped before the cluster was removed,  the resource still exists and the customer's intent was not fulfilled, which is an SLO failure regardless of who initiated the cancellation. Unlike succeeded, a canceled delete never results in the cluster being deleted, so it belongs in the error numerator

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
kartikaysharma27: Done — renamed throughout: recording rule (latency_error_rate), alert names (Latency1h5m, Latency6h30m, Latency3d), SLO labels (cluster-deletion-latency), all annotations, and the Grafana dashboard.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Done — renamed to Latency throughout: recording rule, alert names, SLO labels, all annotations, and Grafana dashboard.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: just want to flag that an earlier comment in this PR (now resolved) moved the correlationId to subscription_id and away from resource_id. Could you clarify which scoping you'd prefer so we can align on the right deduplication boundary: per-subscription (current) or per-resource.?

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
pepedocs: @kartikaysharma27 Could you verify this with BU or @swiencki ? 

This may skew the SLO if there are too many cancellations by the customer but the service is simply serving the intent of the customer.

pepedocs (changes requested):
@kartikaysharma27 +1 for using `Latency`, the only thing I think needs to be consulted is the canceled part whether this can be treated as a Service error or not. When the user cancels and the cancelling results to an error then this is definitely a Service error, but if it was successful then logically there's no Service error. Additionally, if we treat cancelling whatever its result is as a Service error, then we may run into the risk of a skewed SLO because some customers keep on cancelling for some reason (e.g. bad automation).

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
kartikaysharma27: Simon reviewed this PR and explicitly requested that canceled deletes be included in both the numerator and denominator of the failure ratio (his Aug 11 comment). The reasoning is that a canceled delete still leaves the resource behind  from an SLO perspective, the customer's original intent (cluster deletion) was not fulfilled. That said, if customer-initiated cancellations become a significant volume, we can revisit and exclude them with a label filter. For now, this aligns with how the existing user journeys treat incomplete operations.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
swiencki: Per sub is fine, if we need to adjust it later we can.

edisonLcardenas (approved):

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
```diff
@@ -0,0 +1,1999 @@
...
+      },
+      "pluginVersion": "11.6.9",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "count(backend_cluster_created_time_seconds{cluster=~\"$cluster\"}) or vector(0)",
```
yadneshk: During testing I noticed that metrics are being duplicated by prometheus's replicas. For eg, here is duped metric for the same cluster reported by two prom agents which also then duplicate the count
```
backend_cluster_created_time_seconds{cluster="cspr-westus3-svc-1",container="aro-hcp-backend",endpoint="metrics",environment="cspr",instance="10.128.64.241:8081",job="aro-hcp-backend-metrics","microsoft.amwresourceid"="/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/resourcegroups/hcp-underlay-cspr-westus3/providers/microsoft.monitor/accounts/services-cspr-usw3",namespace="aro-hcp",pod="aro-hcp-backend-584c977dfd-tb6x4",prometheus="prometheus/prometheus",prometheus_replica="prom-agent-prometheus-1",region="westus3",resource_id="/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/resourcegroups/testcluster10-02-net-rg-03/providers/microsoft.redhatopenshift/hcpopenshiftclusters/testcluster10-02",service="aro-hcp-backend-metrics",subscription_id="1d3378d3-5a3f-4712-85a1-2485495dfc4b"}

backend_cluster_created_time_seconds{cluster="cspr-westus3-svc-1",container="aro-hcp-backend",endpoint="metrics",environment="cspr",instance="10.128.64.241:8081",job="aro-hcp-backend-metrics","microsoft.amwresourceid"="/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/resourcegroups/hcp-underlay-cspr-westus3/providers/microsoft.monitor/accounts/services-cspr-usw3",namespace="aro-hcp",pod="aro-hcp-backend-584c977dfd-tb6x4",prometheus="prometheus/prometheus",prometheus_replica="prom-agent-prometheus-0",region="westus3",resource_id="/subscriptions/1d3378d3-5a3f-4712-85a1-2485495dfc4b/resourcegroups/testcluster10-02-net-rg-03/providers/microsoft.redhatopenshift/hcpopenshiftclusters/testcluster10-02",service="aro-hcp-backend-metrics",subscription_id="1d3378d3-5a3f-4712-85a1-2485495dfc4b"}
```

This should be addressed with `max without (prometheus_replica)` function. Other dashboards too seem to be doing it like cluster-upgrade-slo.json

yadneshk: image

kartikaysharma27: Thanks for putting efforts on testing.  Added max without(prometheus_replica) wrapper to 8 dashboard panels where absolute counts or table rows would double-count under Prometheus HA. It should fix the duplicate issue.

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
```diff
@@ -54,6 +54,28 @@ resource arohcpClusterProvisionSloRecordingRules 'Microsoft.AlertsManagement/pro
...
+      }
+      {
+        record: 'errors:backend_cluster_deletion_operation:latency_error_rate'
+        expression: '(count by (cluster) ((backend_resource_operation_last_transition_time_seconds{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"} - backend_resource_operation_start_time_seconds{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"}) > 1800 and backend_resource_operation_phase_info{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"} == 1) or 0 * count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"} == 1)) / clamp_min(count by (cluster) (backend_resource_operation_phase_info{operation_type="delete",phase="succeeded",resource_type="microsoft.redhatopenshift/hcpopenshiftclusters"} == 1), 1)'
```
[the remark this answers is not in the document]
kartikaysharma27: it's hallucinated.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,280 @@
...
+        correlationId: "userJourneyClusterDeletionLatency3d/{{ $labels.cluster }}"
+    # ----------------------------------------
+    # Stuck Operations Alert (Sev 4)
+    # ----------------------------------------
+    # This alert fires when time in the CURRENT phase
+    #     (last_transition_time_seconds) > 1h, monitors all 5 active phases.
+    - alert: userJourneyClusterDeletionStuckOperation
+      expr: |
+        (
+          max without(prometheus_replica) (
```
swiencki: This `max without(prometheus_replica)` strips only `prometheus_replica` and keeps the per-replica `pod`/`instance` labels, so a single stuck delete opens one incident per Prometheus replica; should it adopt the `max by (cluster, environment, region, subscription_id, resource_id, resource_type, operation_type, phase)` identity aggregation that #6346 applied to the equivalent nodepool stuck alert?

observability/alerts/cluster-deletion-slo-prometheusRule_test.yaml
```diff
@@ -0,0 +1,834 @@
...
+# time() at eval_time Nm = N * 60 seconds (promtool starts at Unix epoch 0).
+# last_transition_time = 0 → at eval_time 61m, time()-0 = 3660 > 3600.
+# for: 15m → alert fires at eval_time 61+15 = 76m. Test at 80m.
+# ─────────────────────────────────────────────────────────────────────────────
+- name: stuck_operation_fires_after_1_hour
+  interval: 1m
+  input_series:
+  # Operation stuck in "deleting" phase since epoch (last_transition_time=0)
+  # Two Prometheus replicas — max without(prometheus_replica) deduplicates to one alert
+  - series: 'backend_resource_operation_phase_info{resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",operation_type="delete",phase="deleting",resource_id="stuck1",subscription_id="sub1",cluster="svc-1",prometheus_replica="prom-0"}'
```
swiencki, line 595: This dedup test only varies `prometheus_replica`, which `without(prometheus_replica)` already collapses, so it passes without exercising the `pod`/`instance` case that duplicates in production; can we mirror #6346's test by feeding two series that differ only by `instance`?

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
```diff
@@ -0,0 +1,1999 @@
+{
```
swiencki, line 1: The sibling cluster-provisioning-slo dashboard includes an `SLO Alert Status` row but this one does not; should we add that row so the nine new deletion SLO alerts' firing state is visible alongside the SLOs?

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,280 @@
...
+            backend_resource_operation_phase_info{
+              resource_type="microsoft.redhatopenshift/hcpopenshiftclusters",
+              operation_type="delete",
+              phase=~"accepted|awaitingsecret|provisioning|updating|deleting"
```
swiencki: A delete operation transits `accepted` to `deleting` to terminal, so `provisioning|updating|awaitingsecret` in this stuck-phase regex never matches for `operation_type="delete"`; is the broader phase set intentional defensiveness or carried over from the create/update template?

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Switched to max by(cluster, environment, region, subscription_id, resource_id, resource_type, operation_type, phase) in commit 94b5576.

observability/grafana-dashboards/sre/user-journey/cluster-deletion.json
kartikaysharma27: Will add in a follow-up PR  work to keep this one focused.

observability/alerts/cluster-deletion-slo-prometheusRule_test.yaml
kartikaysharma27: Added stuck_dedup_instance_label test that feeds two series differing only by instance label and asserts one alert.

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
kartikaysharma27: Narrowed to accepted|deleting - verified from source code that delete operations only transit accepted → deleting → terminal.

kartikaysharma27:
> Kartikay-redhat (@kartikaysharma27) +1 for using `Latency`, the only thing I think needs to be consulted is the canceled part whether this can be treated as a Service error or not. When the user cancels and the cancelling results to an error then this is definitely a Service error, but if it was successful then logically there's no Service error. Additionally, if we treat cancelling whatever its result is as a Service error, then we may run into the risk of a skewed SLO because some customers keep on cancelling for some reason (e.g. bad automation).

Canceled deletes are included as errors by design for this UJ, answered in above reviews.

vinaybommana:
validated:

`make -C tooling/prometheus-rules run-rp-services run-recording-rules-services   → PASS`

  All 26 promtool assertions across 20 scenarios pass, and all 9 alerts are present in generatedRPPrometheusAlertingRules.bicep (lines 1149–1394) with the generator's unless on (subscription_id) internal_subscription:info correctly injected into the stuck alert.Metrics, label sets, and phase values all check out against `backend/pkg/controllers/metrics/operation_phase_metrics_controller.go `and `internal/api/coreapi/resource.go, and workqueue_depth{name="OperationClusterDelete"}` matches the real controller name.

/lgtm

observability/alerts/cluster-deletion-slo-prometheusRule.yaml
```diff
@@ -0,0 +1,299 @@
...
+  # The 'for' duration enforces sustained observation, approximating the
+  # short window of multi-window burn rate alerting.
+  - name: arohcp_cluster_deletion_slo_error_alerts
+    rules:
+    # ----------------------------------------
+    # Fast Burn Alert (Sev 3)
+    # ----------------------------------------
+    # 14.4x burn rate → >72% failure rate sustained for 5 minutes
+    # Would exhaust 7-day error budget in ~12 hours
+    - alert: userJourneyClusterDeletionErrors1h5m
```
swiencki: All three tiers gate the same point-in-time error_rate with only for, so the 1h5m/6h30m/3d names do not map to real windows; please template the per-tier *_over_time recording rules and volume floors from cluster-provision-slo, which solves the same gauge-windowing this PR's comment calls impossible.

PR #6096, opened 2026-07-15
fix(test): wrap Graph writes in Eventually

mvacula02:
[ARO-27423](https://redhat.atlassian.net/browse/ARO-27423)

### What

Wrap Graph writes in Eventually with a 2‑minute budget and 5‑second poll interval, matching the existing convention. 

### Why

A newly‑created application object is not immediately visible on every Graph replica, write operations against it can return 404 ResourceNotFound or 403 RequestDenied fo tens of seconds (or even longer) after the create.

### Testing

The PR modifies existing test framework functions and their 2 callers. 

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/util/framework/per_test_framework.go
+++ b/test/util/framework/per_test_framework.go
@@ -795,10 +795,12 @@ func (tc *perItOrDescribeTestContext) NewAppRegistrationWithServicePrincipal(ctx
 		return nil, nil, fmt.Errorf("failed to get graph client: %w", err)
 	}
 
-	app, err := graphClient.CreateApplication(ctx, appName, []string{})
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to create app registration: %w", err)
-	}
+	var app *graphutil.Application
+	gomega.Eventually(func() error {
+		var err error
+		app, err = graphClient.CreateApplication(ctx, appName, []string{})
+		return err
+	}, 2*time.Minute, 5*time.Second).Should(gomega.Succeed(), "failed to create app registration")
 
 	func() {
 		tc.contextLock.Lock()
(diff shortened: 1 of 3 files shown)


test/util/framework/per_test_framework.go
```diff
@@ -807,14 +809,30 @@ func (tc *perItOrDescribeTestContext) NewAppRegistrationWithServicePrincipal(ctx
 		tc.knownAppRegistrationIDs = append(tc.knownAppRegistrationIDs, app.ID)
 	}()
 
-	sp, err := graphClient.CreateServicePrincipal(ctx, app.AppID)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to create service principal: %w", err)
-	}
+	var sp *graphutil.ServicePrincipal
+	gomega.Eventually(func() error {
```
mbukatov, line 813: Here as well.

test/util/framework/per_test_framework.go
```diff
@@ -795,10 +795,12 @@ func (tc *perItOrDescribeTestContext) NewAppRegistrationWithServicePrincipal(ctx
 		return nil, nil, fmt.Errorf("failed to get graph client: %w", err)
 	}
 
-	app, err := graphClient.CreateApplication(ctx, appName, []string{})
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to create app registration: %w", err)
-	}
+	var app *graphutil.Application
+	gomega.Eventually(func() error {
```
mbukatov, line 799: We should pass the context to Eventually as well, so that if the context is cancelled, Eventually won't try to retry to run the func again.

See and example [how a context is used with Eventually in the docs](https://pkg.go.dev/github.com/onsi/gomega#Eventually) (it's passed to both Eventually as well as the function it runs inside):

``` .go
Eventually(ctx, func() int {
    return client.FetchCount(ctx, "/users")
}).Should(BeNumerically(">=", 17))
```

test/util/framework/per_test_framework.go
```diff
@@ -807,14 +809,30 @@ func (tc *perItOrDescribeTestContext) NewAppRegistrationWithServicePrincipal(ctx
...
+		return err
+	}, 2*time.Minute, 5*time.Second).Should(gomega.Succeed(), "failed to create service principal")
 
 	return app, sp, nil
 }
 
+func AddAppRegistrationPassword(ctx context.Context, graphClient *graphutil.Client, appID, displayName string) string {
+	var passSecret string
+	gomega.Eventually(func() error {
+		pass, err := graphClient.AddPassword(ctx, appID, displayName, time.Now(), time.Now().Add(24*time.Hour))
```
mbukatov, line 825: Here as well.

mbukatov (commented):
Providing a feedback wrt context handling.

That said, I realized that the client we are using here is from [internal/graph/util/serviceprincipal.go](https://github.com/Azure/ARO-HCP/blob/main/internal/graph/util/serviceprincipal.go), which means it is under our control.

I wonder if in some cases, it would not be better to fix the problem (increase retry budget in these functions directly, or at least expose them so that the test code can increase them). Esp. [`func (c *Client) AddPassword(ctx context.Context, appID, displayName string, startTime, endTime time.Time) (*PasswordCredential, error)`](https://github.com/Azure/ARO-HCP/blob/dbb4684b883c86aa09f4516b8516cf5d8bc834c0/internal/graph/util/applications.go#L86) code may be extended to log errors better (so that we don't get just ` polling error: error status code received from the API`.

Let's ask @bennerv for suggestion.

mvacula02:
closing in favor of #6325

PR #6101, opened 2026-07-15
test: add E2E tests to validate node pool creation for various availability zone scenarios

oshrirao:
Jira issue: 
https://redhat.atlassian.net/browse/ARO-13378

### What

Adds ARO-HCP E2E tests for node pool creation for various availability zone behavior.

**Scenarios covered:**
- **Positive:** nodepool without AZ; nodepool with a valid AZ (Updated existing tests to add Explicit assertion for AZ to avoid reduntant tests)
- **Negative:** non-existing AZ (`"99"`) fails inflight with expected error
- **Negative:** VM size that does not support AZ deployment fails inflight (Skips when no such SKU exists in the region)

### Testing

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

- [x] E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.mhttps://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md)

### DIFF
--- a/test/e2e/cluster_create_nodepool_osdisk.go
+++ b/test/e2e/cluster_create_nodepool_osdisk.go
@@ -122,5 +122,7 @@ var _ = Describe("Customer", func() {
 			Expect(created.Properties.Platform).ToNot(BeNil(), "nodepool Properties.Platform was nil")
 			Expect(created.Properties.Platform.OSDisk).ToNot(BeNil(), "nodepool Properties.Platform.OSDisk was nil")
 			Expect(*created.Properties.Platform.OSDisk.SizeGiB).To(Equal(customerNodeOsDiskSizeGiB), "nodepool OS disk size should be %d GiB", customerNodeOsDiskSizeGiB)
+			Expect(created.Properties.Platform.AvailabilityZone).To(BeNil(),
+				"expected availability zone to be unset when not specified on nodepool create")
 		})
 })
--- a/test/e2e/nodepool_availability_zone.go
+++ b/test/e2e/nodepool_availability_zone.go
@@ -1,0 +1,166 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	hcpsdk20240610preview "github.com/Azure/ARO-HCP/test/sdk/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+)
+
+const nonExistentAvailabilityZone = "99"
+
+type availabilityZoneTestCluster struct {
+	resourceGroupName        string
+	managedResourceGroupName string
+	clusterName              string
+	nodePoolsClient          *hcpsdk20240610preview.NodePoolsClient
+	tc                       *framework.E2ETestContext
+}
+
+func createAvailabilityZoneTestCluster(ctx context.Context) availabilityZoneTestCluster {
+	const customerClusterName = "np-az-cluster"
+
+	tc := framework.NewTestContext()
+
+	if tc.UsePooledIdentities() {
+		err := tc.AssignIdentityContainers(ctx, 1, framework.IdentityContainerAssignmentRetryInterval)
+		Expect(err).NotTo(HaveOccurred(), "failed to assign pooled identity containers")
+	}
+
+	By("creating a resource group")
+	resourceGroup, err := tc.NewResourceGroup(ctx, "np-az", tc.Location())
+	Expect(err).NotTo(HaveOccurred(), "failed to create resource group for nodepool availability zone tests")
+
+	clusterParams := framework.NewDefaultClusterParams20240610()
+	clusterParams.ClusterName = customerClusterName
+	managedResourceGroupName := framework.SuffixName(*resourceGroup.Name, "-managed", 64)
+	clusterParams.ManagedResourceGroupName = managedResourceGroupName
+
+	By("creating customer resources")
+	clusterParams, err = tc.CreateClusterCustomerResources20240610(ctx,
+		resourceGroup,
+		clusterParams,
+		map[string]any{},
+		TestArtifactsFS,
+		framework.RBACScopeResourceGroup,
+	)
+	Expect(err).NotTo(HaveOccurred(), "failed to create customer resources for availability zone tests")
+
+	By("creating the HCP cluster")
+	err = tc.CreateHCPClusterFromParam20240610(ctx,
+		GinkgoLogr,
+		*resourceGroup.Name,
+		clusterParams,
+		framework.ClusterCreationTimeout,
+	)
+	Expect(err).NotTo(HaveOccurred(), "failed to create HCP cluster %s", customerClusterName)
+
+	return availabilityZoneTestCluster{
+		resourceGroupName:        *resourceGroup.Name,
+		managedResourceGroupName: managedResourceGroupName,
+		clusterName:              customerClusterName,
+		nodePoolsClient:          tc.Get20240610ClientFactoryOrDie(ctx).NewNodePoolsClient(),
+		tc:                       tc,
+	}
+}
+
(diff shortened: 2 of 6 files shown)


test/e2e/nodepool_availability_zone.go
```diff
@@ -0,0 +1,226 @@
...
+			noAZNodePoolParams.NodePoolName = "np-no-az"
+			err = tc.CreateNodePoolFromParam20240610(ctx,
+				GinkgoLogr,
+				resourceGroupName,
+				managedResourceGroupName,
+				clusterName,
+				noAZNodePoolParams,
+				framework.NodePoolCreationTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create nodepool without availability zone")
```
miguelsorianod: Have we checked if we have existing e2e tests around this? I see here we exercise nodepool with and without availability zone. I am quite sure at least one of it is already being exercised in other e2e tests so it seems redundant

oshrirao: Yes I checked existing E2E coverage before adding this.
Overlap:
Many tests create nodepools without setting AvailabilityZone (default params), so “create without AZ succeeds” is indirectly exercised (e.g. cluster_create_nodepool_osdisk.go, simple_negative_cases.go).
nodepool_autoscaling.go creates an AZ nodepool when usable zones exist, but only to test autoscaling, it doesn’t assert the AZ field round-trip.

Not covered elsewhere:
Explicit assertion that Properties.Platform.AvailabilityZone is unset on no-AZ create and matches on with-AZ create.
Negative inflight validation: invalid zone ("99") and VM size that doesn’t support zone deployment (with expected error strings + ProvisioningStateFailed).

This PR adds dedicated AZ create coverage. Happy to trim the positive “without AZ” case if we consider implicit coverage sufficient, I’d keep the with-AZ assertion and both negative scenarios as the net-new value. 

Would appreciate your thoughts on this, happy to adjust the tests based on your feedback.

test/util/framework/vm_sku_selection.go
```diff
@@ -112,6 +112,77 @@ func (tc *perItOrDescribeTestContext) SelectVMSize(ctx context.Context, selector
...
+	}
+	if selector.NamePattern != nil && !selector.NamePattern.MatchString(*sku.Name) {
+		return "", false
+	}
+	if !skuUsable(sku, location, selector) {
+		return "", false
+	}
+	if skuAdvertisesAvailabilityZones(sku, location) {
+		return "", false
+	}
```
LiorKGOW, line 169:
```suggestion
	if sku == nil || sku.Name == nil || selector.NamePattern != nil && !selector.NamePattern.MatchString(*sku.Name) || !skuUsable(sku, location, selector) || skuAdvertisesAvailabilityZones(sku, location) {
		return "", false
	}
```

LiorKGOW: Just a nit, 
As all of those conditions return the same, we could keep it in a single if, and keep the order (in case it matters), but I am fine with your suggesiton if you wish to make it more explicit 😇

oshrirao: > Just a nit, As all of those conditions return the same, we could keep it in a single if, and keep the order (in case it matters), but I am fine with your suggesiton if you wish to make it more explicit 😇

Thanks for the comment @LiorKGOW ! I separated them because I find it cleaner and easier to read. Since each check is independent, the explicit version felt more readable to me.

test/e2e/nodepool_availability_zone.go
oshrirao: @miguelsorianod Removed redundant positive scenarios and added Explicit assertion for AZ in existing tests.

test/e2e/cluster_create_nodepool_osdisk.go
```diff
@@ -122,5 +122,9 @@ var _ = Describe("Customer", func() {
 			Expect(created.Properties.Platform).ToNot(BeNil(), "nodepool Properties.Platform was nil")
 			Expect(created.Properties.Platform.OSDisk).ToNot(BeNil(), "nodepool Properties.Platform.OSDisk was nil")
 			Expect(*created.Properties.Platform.OSDisk.SizeGiB).To(Equal(customerNodeOsDiskSizeGiB), "nodepool OS disk size should be %d GiB", customerNodeOsDiskSizeGiB)
+			if created.Properties.Platform.AvailabilityZone != nil {
+				Expect(*created.Properties.Platform.AvailabilityZone).To(BeEmpty(),
+					"expected availability zone to be unset when not specified on nodepool create")
```
shitaljante: As the AvailabilityZone is built via pointer , you can do this instead 
```
Expect(created.Properties.Platform.AvailabilityZone).To(BeNil(), "expected availability zone to be unset when not specified on nodepool create")
```

test/e2e/nodepool_availability_zone.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+var _ = Describe("Nodepool Availability Zone", func() {
+	It("should reject nodepool creation for invalid availability zone configurations",
+		labels.RequireNothing,
+		labels.Medium,
+		labels.Negative,
+		labels.AroRpApiCompatible,
+		func(ctx context.Context) {
+			resourceGroupName, managedResourceGroupName, clusterName, nodePoolsClient, tc := createAvailabilityZoneTestCluster(ctx)
```
shitaljante: Nit: we are returning three strings in order (resourceGroupName, managedResourceGroupName, clusterName) from "createAvailabilityZoneTestCluster", consider adding a small result struct to avoid assigning values to wrong order.

test/e2e/nodepool_availability_zone.go
```diff
@@ -0,0 +1,154 @@
...
+
+			workerVMSize, err := tc.SelectVMSize(ctx, framework.DefaultWorkerVMSizeSelector())
```
shitaljante: are we sure that selected workerVMSize supports the availability zones ?
we should confirm that before assigning a nonExistentAvailabilityZone.

oshrirao:

test/e2e/cluster_create_nodepool_osdisk.go
oshrirao: Added suggestion. Thanks.

test/e2e/nodepool_availability_zone.go
oshrirao: Added result stuct. Thanks for the suggestion @shitaljante

test/e2e/nodepool_availability_zone.go
oshrirao: Good catch!! we shouldn’t assume the default worker SKU always has usable zones. I updated this to select via DefaultWorkerVMSizeSelector with **RequireZones**: **true**, so we only proceed with a VM size that has availability zone support before assigning the non-existent zone.

test/e2e/nodepool_availability_zone.go
```diff
@@ -0,0 +1,166 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: Have we verified if this works as expected?

If some of the validations are being performed as CS inflights, those occur asynchronously and not on the frontend, so checking for an error on the frontend call is not going to return those errors. It could be that the operation returns a message that we might be able to match, but right now we are exposing internal details that we might not end up returning when we redesign the operation reporting area to only include user-facing messages (or when the corresponding inflight is fully removed from CS side)

I wonder if at this point the best is to wait until we have finished the validations reporting logic in the RP, we migrate the validations there and we implement the test checking the validation result. The alternative is to try to implement it with what's here, but the issue will be that once the operation message surfacing design changes and/or the inflights are removed from CS this will start to fail and will need to be redesigned at that point

oshrirao: Yes, I ran this locally and it passed. The helper waits on the LRO, so we do see the failed async operation (and currently the CS inflight message in the error), not a sync frontend validation reject.

That said, you’re right that matching those messages is brittle and will need a rewrite once RP validation reporting is in place / CS inflights move. We already cover this in the CS [MR](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/372), so I’m fine merging that and holding this ARO E2E for now.

shitaljante: Sounds good to close this PR and focus on the API tests in CS.
@miguelsorianod Do we already have a plan to migrate the E2E tests from CS to RP along with codebase, or will that be handled as follow-up work after the functionality migration?

test/e2e/nodepool_availability_zone.go
```diff
@@ -0,0 +1,166 @@
...
+		tc:                       tc,
+	}
+}
+
+var _ = Describe("Nodepool Availability Zone", func() {
+	It("should reject nodepool creation for invalid availability zone configurations",
+		labels.RequireNothing,
+		labels.Medium,
+		labels.Negative,
+		labels.AroRpApiCompatible,
```
mgahagan73, line 93: Now that the MI Container aware scheduler has merged, you will need to add a label depicting how many Managed Identity containers this test will use. e.g. labels.MIContainers(n) where n is the number of MI Containers. Please also rebase to pick up the container checks that were added to make verify to avoid any last minute surprises.

oshrirao: Thanks for the update! We've decided to drop the plan to add this E2E test to ARO-HCP for now. We've already added it to CS, so I'll go ahead and close this PR.

PR #6102, opened 2026-07-15
test/aro-hcp-tests: pass MonitoringEventsDatabase to Kusto queries

stevekuznetsov:
test/aro-hcp-tests: emit failed jUnit when snapshot gathering errors

Previously, when gatherer.Gather() failed (e.g. due to a Kusto query
error), the error was logged but no jUnit test case was emitted. This
meant CI saw 0 tests / 0 failures and reported success, completely
hiding the problem.

Synthesize a failed VerificationReport for each Gather failure so it
flows through the normal jUnit aggregation path and surfaces in CI.

---

test/aro-hcp-tests: pass MonitoringEventsDatabase to Kusto queries

The GatherInput struct was missing MonitoringEventsDatabase, leaving it
as an empty string. The alert queries use this database via the
alertEvents.kql template partial, which rendered as
`.database('')` — causing Kusto to look up 'NetDefaultDB' which
does not exist, failing every alert query.

Read the value from the rendered config key
kusto.monitoringEventsDatabase and pass it through.

---

### DIFF
--- a/test/cmd/aro-hcp-tests/gather-snapshot/options.go
+++ b/test/cmd/aro-hcp-tests/gather-snapshot/options.go
@@ -61,12 +61,13 @@ type ValidatedOptions struct {
 }
 
 type completedOptions struct {
-	OutputDir      string
-	KustoEndpoint  string
-	ServiceDB      string
-	HCPDB          string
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
This is a correctness issue, it must go in.

PR #6104, opened 2026-07-15
chore: remove duplicate stuck deletion alert

sclarkso:
### What

Removes HCPdeletionStuck alert.

### Why

2 alerts exist for this issue.  Removing mine.  The other alert remains:
  │ BackendAsyncOperationClusterDeleteStuck │ backend/alerts/backend-async-operations-prometheusRule.yaml │ 

### Testing

e2e
### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


roivaz:
Don't we want this alert as well?
* this one detects clusters stuck in deletion (including customer HCPs)
* the other one detects HCPs hanging around in our e2e subs for longer than they should

Maybe just filter the alert in this PR to non-e2e subscriptions?

roivaz:
Reply from @sclarkso 

>No, that e2e sub one still exists; the 2 I'm talking about are the same expression.  Mine was the original one and used a different metric.  That metric was deleted, so then I updated it with the new metric, but it took forever to merge.  While waiting for it to merge, they added the same alert to the backend queries:
https://github.com/Azure/ARO-HCP/blob/main/backend/alerts/backend-async-operations-prometheusRule.yaml
It's the same expression, just different evaluation times.  And the one I'm leaving is bundled in with other backend alerts, so it made more sense to delete mine.

PR #6110, opened 2026-07-16
fix: call cincinatti API directly to check for next-minor channel existence in control plane desired version selection

machi1990:
### What

fix: call cincinatti API directly to check for next-minor channel existence in control plane desired version selection

### Why

selectBestVersionFromCandidates in the control plane desired version controller checked whether the next-minor Cincinnati channel exists by calling GetUpdates from a candidate version on that channel. That probes whether a specific version has upgrade edges in the next-minor graph — not whether the channel itself is published.

This can misclassify channel availability e.g. treat a channel as missing when it exists but the candidate has no edges yet.

Fixes https://redhat.atlassian.net/browse/ARO-28443

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -566,6 +560,10 @@ func selectBestVersionFromCandidates(
 	if preferLatestOverGateway {
 		return &candidates[0], nil
 	}
+
+	// TODO: If the next minor exists but none of the candidates have a gateway to it,
+	// and none of the active versions have a gateway to it either, prefer the latest
+	// candidate instead of returning nil — there is no existing next-minor path to preserve.
```
machi1990, line 566: @deads2k  @stevekuznetsov  what your thoughts on this? 
The scenario described in the comment can occur and we are off maintaining z-stream upgrade if a new y + 1 is available but we don't have an upgrade path to it yet from the current y.z or candidates y.z+1; In this case, we return y.z+1 and not nil.

What do you reckon?

machi1990: Also worth noting that; I left it as a TODO with a plan to fast follow on it once the current PR merges if we agree

internal/cincinnati/graph_client.go
```diff
@@ -0,0 +1,122 @@
...
+		return false, fmt.Errorf("query graph for %s returned %s: %s", channel, resp.Status, strings.TrimSpace(string(body)))
+	}
+
+	var payload struct {
+		Nodes []json.RawMessage `json:"nodes"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&payload); err != nil {
+		return false, fmt.Errorf("decode graph response for %s: %w", channel, err)
+	}
+	return len(payload.Nodes) > 0, nil
```
deads2k, line 121: Provide an artifact we can load in testing that shows what this looks like.  I think we need to know if at least one of the nodes is for the previous 4.y

internal/cincinnati/graph_client_test.go
```diff
@@ -0,0 +1,179 @@
...
+				w.WriteHeader(http.StatusNotFound)
+			case "stable-4.97":
+				w.WriteHeader(http.StatusInternalServerError)
+				_, _ = w.Write([]byte("boom"))
+			case "stable-4.96":
+				w.WriteHeader(http.StatusOK)
+				_, _ = w.Write([]byte(`{not-json`))
+			default:
+				t.Fatalf("unexpected channel: %s", r.URL.Query().Get("channel"))
+			}
```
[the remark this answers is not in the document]
deads2k: needs fixing.

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
deads2k: Yeah, I think that's ok.

internal/cincinnati/graph_client.go
deads2k: > think we need to know if at least one of the nodes is for the previous 4.y
I found https://github.com/Azure/ARO-HCP/pull/6110/changes#r3593914880.  I see this part doesn't apply. The example content is still important.

deads2k:
test problems need fixing.

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
machi1990: Cool, thanks. I'll open a separate followup PR

internal/cincinnati/graph_client_test.go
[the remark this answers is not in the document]
machi1990: Yes, this was a good catch. It was fixed already. We now return the http error code instead of failing

backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
machi1990: I opened https://github.com/Azure/ARO-HCP/pull/6122

PR #6111, opened 2026-07-16
feat:cache SKUs API reponse for VM validations

shitaljante:
https://redhat.atlassian.net/browse/ARO-15594 
### What

- Adds Azure clients for Compute Resource SKUs and Usage, built via the FPA client builder.
- Introduces FirstPartyApplicationAzureCachedReaders that caches per-subscription VM Resource SKU lists in memory which is filtered based on location. The cache key is subscriptionID. 
- Success TTL is 20m, error TTL 5m (errors cached on the entry)
-  Least Recently Used (LRU) eviction policy is implemented, to evict the entry that has been Least Recently Used at the point in time of eviction.
- Wires the cached reader into backend startup for new VM validations.
- Adds unit coverage 

### Why
Node pool update inflight checks need Resource SKUs data from Azure. Without caching, repeated list calls are slow and can hit rate limits. Caching (aligned with clusters-service) keeps SKU lookups fast and reusable across validations while still refreshing on a bounded TTL.

### Testing

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,198 @@
...
+
+package cachedreader
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/golang-lru/v2/expirable"
```
miguelsorianod: We shouldn't add a new dependency. Stick to K8s related libraries.

backend/pkg/app/backend.go
```diff
@@ -93,6 +93,7 @@ type BackendOptions struct {
 	FPAClientBuilder                   azureclient.FirstPartyApplicationClientBuilder
 	BackendIdentityAzureClients        *azureclient.BackendIdentityAzureClients
 	BackendIdentityAzureCachedReaders  *cachedreader.BackendIdentityAzureCachedReaders
+	ResourceSKUsCachedReader           cachedreader.ResourceSKUsCachedReader
```
miguelsorianod: We must design it in a way where it's clear from the types and names that it can't be misused with another identity.

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,198 @@
...
+				continue
+			}
+			skus = append(skus, sku)
+		}
+	}
+	return skus, nil
+}
+
+func cloneResourceSKUSlice(skus []*armcompute.ResourceSKU) []*armcompute.ResourceSKU {
+	return append([]*armcompute.ResourceSKU(nil), skus...)
```
miguelsorianod: How is this a clone? this just copies the pointers.

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,198 @@
...
+	var noEvictionCallback expirable.EvictCallback[string, *cachedResourceSKUsEntry]
+	return &resourceSKUsCachedReader{
+		clientBuilder: clientBuilder,
+		clock:         utilsclock.RealClock{},
+		cache:         expirable.NewLRU(maxEntries, noEvictionCallback, resourceSKUsCacheKeySuccessTTL),
+	}
+}
+
+// SetClock substitutes the time source used for TTL checks. Intended for tests.
+func (c *resourceSKUsCachedReader) SetClock(clock utilsclock.PassiveClock) {
```
miguelsorianod: This seems like a hack for tests. This should be a dependency on the type and the tests should provide at instantiation time

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+const (
+	// resourceSKUsCacheKeySuccessTTL defines how long a cached successful Resource SKUs list is valid.
+	// Also used as the expirable LRU entry TTL since creation.
+	resourceSKUsCacheKeySuccessTTL = 20 * time.Minute
+	// resourceSKUsCacheKeyErrorTTL defines how long a cached Resource SKUs list error is valid.
+	resourceSKUsCacheKeyErrorTTL = 5 * time.Minute
+	// defaultResourceSKUsCacheMaxEntries is the LRU capacity for per-subscription SKU cache entries.
+	defaultResourceSKUsCacheMaxEntries = 30
```
miguelsorianod, line 52: Can we make some numbers on consumption and document it? 30 might be little depending on that

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,198 @@
...
+func NewResourceSKUsCachedReader(clientBuilder resourceSKUsClientBuilder) ResourceSKUsCachedReader {
+	return newResourceSKUsCachedReader(clientBuilder, defaultResourceSKUsCacheMaxEntries)
+}
+
+func newResourceSKUsCachedReader(clientBuilder resourceSKUsClientBuilder, maxEntries int) *resourceSKUsCachedReader {
+	var noEvictionCallback expirable.EvictCallback[string, *cachedResourceSKUsEntry]
+	return &resourceSKUsCachedReader{
+		clientBuilder: clientBuilder,
+		clock:         utilsclock.RealClock{},
+		cache:         expirable.NewLRU(maxEntries, noEvictionCallback, resourceSKUsCacheKeySuccessTTL),
```
miguelsorianod: Here it seems we are incorrectly conflating the maximum time until the key is removed from cache automatically vs the maximum time the key is kept in cache when it was a success since the last request and when a new request comes.
It is not correct to reuse and name them equally

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,198 @@
...
+	return c.clock.Since(entry.lastUpdate) > ttl
+}
+
+func (c *resourceSKUsCachedReader) listVirtualMachineSKUsFromAzure(ctx context.Context, tenantID, subscriptionID string) ([]*armcompute.ResourceSKU, error) {
+	client, err := c.clientBuilder.ResourceSKUsClient(tenantID, subscriptionID)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to create Resource SKUs client: %w", err))
+	}
+
+	pager := client.NewListPager(nil)
```
ahitacat: In CS we list by location, I think we should do the same here. We only care about the same region where this service is deployed.

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		skus, listErr := c.listVirtualMachineSKUsFromAzure(ctx, tenantID, subscriptionID)
+		entry := &cachedResourceSKUsEntry{
+			skus:       skus,
+			err:        listErr,
+			lastUpdate: c.clock.Now().UTC(),
+		}
+		c.cache.Add(cacheKey, entry)
+		return entry, nil
```
ahitacat: Shouldn't this return the listErrs as the error instead of nil?

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,198 @@
...
+	})
+	if err != nil {
```
ahitacat: err is always nil, we should consider returning the error if it exists no?

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: I used github.com/hashicorp/golang-lru/v2/expirable because it provides an LRU cache with time-based expiration (TTL).

Currently, backend uses k8s.io/utils/lru, which provides size-bounded LRU caching but does not support TTL. I also saw a [comment](https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/9727#note_15568263) indicating that the HashiCorp library has a valid license, so I used it here.

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: We store `listErr` in the cache entry which allows failed list requests to be cached for the error TTL and returned on next requests without calling Azure again. This is by design. 

Returning the same error from Do would be redundant.

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: I removed the unused outer error check, which was always nil.

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: There is also [k8s.io/apimachinery/pkg/util/cache.LRUExpireCache](http://k8s.io/apimachinery/pkg/util/cache.LRUExpireCache), which is already pulled in backend. I can switch to that and check if it meets our requirements.

backend/pkg/app/backend.go
shitaljante: Changed to 	- FPAAzureCachedReaders              *cachedreader.FirstPartyApplicationAzureCachedReaders

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader.go
shitaljante: Documented the consumption values - 
	// Testing with one Resource SKUs cache entry per subscription, filtered by the region, results in
	// an approximate size of 4 MB per entry. With defaultResourceSKUsCacheMaxEntries = 30, the cache
	// could reach approximately 120 MB in the worst case. This value can be adjusted after measuring
	// the actual memory footprint with real filtered lists.

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: Filtered based on the region where the backend is deployed: https://github.com/Azure/ARO-HCP/pull/6111/changes#diff-c95b499a6d49148bf931d975dbe397922dad9c458fe9de5cba4a99db7d7b5a7bR344

Should we instead filter based on the cluster region? I believe both regions should be the same, but does it matter which one we use here? @miguelsorianod

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: Updated function to deep copy fields.

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: Updated to use k8s library. 7d95bc0a5b240455361691a373c640020ab42923 commit contains the changes to use `cache.LRUExpireCache`

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: Resolved by switching to the new library.

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader.go
```diff
@@ -0,0 +1,223 @@
...
+
+// GetVirtualMachineSKU looks up one VM size in the cached list.
+func (c *resourceSKUsCachedReader) GetVirtualMachineSKU(ctx context.Context, tenantID, subscriptionID, vmSize string) (*armcompute.ResourceSKU, error) {
+	skus, err := c.ListVirtualMachineSKUs(ctx, tenantID, subscriptionID)
+	if err != nil {
+		return nil, err
+	}
+	for _, sku := range skus {
+		if sku != nil && sku.Name != nil && *sku.Name == vmSize {
+			return sku, nil
```
ahitacat, line 190: Shouldn't this return a copy of the sku as well?

shitaljante: GetVirtualMachineSKU calls `ListVirtualMachineSKUs`, which already returns deepCopyResourceSKUSlice(entry.skus)  So the SKU returned by this is already a copy, transitively.

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,219 @@
...
+		return nil, err
+	}
+	for _, sku := range skus {
+		if sku != nil && sku.Name != nil && *sku.Name == vmSize {
+			return sku, nil
+		}
+	}
+	return nil, utils.TrackError(fmt.Errorf("VM size %q not found in Resource SKUs for subscription %q in location %q", vmSize, subscriptionID, c.location))
+}
```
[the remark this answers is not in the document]
shitaljante: Added test cases

backend/pkg/azure/cachedreader/first_party_application_cached_readers.go
```diff
@@ -0,0 +1,41 @@
...
+	// ResourceSKUsCachedReader returns Microsoft.Compute Resource SKUs for virtualMachines
+	// with in-memory caching on read paths, authenticated via the FPA identity.
+	ResourceSKUsCachedReader ResourceSKUsCachedReader
+}
+
+// NewFirstPartyApplicationAzureCachedReaders constructs FPA-scoped cached readers from the
+// FPA client builder so the identity used for Azure calls is explicit in the type system.
+// location is the Azure location (region) this backend is deployed in, used to scope
+// Resource SKUs lookups to that region.
+func NewFirstPartyApplicationAzureCachedReaders(
```
miguelsorianod: It's unclear to me what would be the lifetime of the underlying clients. An important aspect is that we can't keep an underlying FPA client created for a long time because the token will expire.

backend/pkg/azure/cachedreader/first_party_application_cached_readers.go
```diff
@@ -0,0 +1,41 @@
...
+	fpaClientBuilder azureclient.FirstPartyApplicationClientBuilder,
```
miguelsorianod: nitpick: in this project we accept long lines

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,223 @@
...
+	// per-entry TTL for error entries.
+	resourceSKUsCacheErrorFreshnessTTL = 5 * time.Minute
+	// defaultResourceSKUsCacheMaxEntries is the LRU capacity for per-subscription SKU cache entries.
+	// Testing with one Resource SKUs cache entry per subscription, filtered by the region, results in
+	// an approximate size of 4 MB per entry. With defaultResourceSKUsCacheMaxEntries = 30, the cache
+	// could reach approximately 120 MB in the worst case. This value can be adjusted after measuring
+	// the actual memory footprint with real filtered lists.
+	defaultResourceSKUsCacheMaxEntries = 30
+
+	virtualMachinesResourceType = "virtualMachines"
```
miguelsorianod: this is package-level variable. The name is too generic. Name it so there's no confusion about what this isa bout

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,223 @@
...
+
+func (c *resourceSKUsCachedReader) ensureCached(ctx context.Context, tenantID, subscriptionID string) (*cachedResourceSKUsEntry, error) {
+	cacheKey := strings.ToLower(subscriptionID)
+	if value, ok := c.cache.Get(cacheKey); ok {
+		return value.(*cachedResourceSKUsEntry), nil
+	}
+
+	// Detach cancel/deadline from the caller so the singleflight winner cannot
+	// poison concurrent callers or cache a context-cancellation error.
+	azureCtx := context.WithoutCancel(ctx)
```
miguelsorianod: what does this mean? If the parent context is cancelled we should honor it

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,223 @@
...
+// resourceSKUsCachedReader wraps a Resource SKUs client builder. List results are cached in memory
+// per subscription ID with a TTL, LRU eviction, and singleflight deduplication. All lookups are
+// scoped to a single, fixed Azure location set at construction time.
+type resourceSKUsCachedReader struct {
+	clientBuilder resourceSKUsClientBuilder
+	// location is the Azure location (region) this backend is deployed in. Every Resource SKUs
+	// list call is filtered to this location, since we only care about SKUs available where the
+	// service itself runs.
+	location string
+	cache    *lrucache.LRUExpireCache
```
miguelsorianod: Document cache key and value

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: This detaches the context from the parent context. In case of timeout or cancellation, if multiple goroutines are trying to list SKUs for the same subscription at the same time, singleflight will execute only one call at a time. For each goroutine, a new copy of the context is created with this. The other goroutines will block and receive result from cache data after the single go routine is complete.

If we use the actual caller's context, which has cancellation attached, there is chance that particular caller disconnects or times out, the shared Azure call will be cancelled with context canceled. This can happen even though the other's contexts are still valid.

backend/pkg/azure/cachedreader/first_party_application_cached_readers.go
shitaljante: Each call to `ResourceSKUsClient()` (https://github.com/Azure/ARO-HCP/pull/6111/changes#diff-1c7a411c20e7c7f2d996cac8dd139374a52a4cd94a3ca50143af506209a126e3R167) calls `fpaTokenCredRetriever.RetrieveCredential(tenantID)` to get a token credential, then a new `armcompute.ResourceSKUsClient` created from it. https://github.com/Azure/ARO-HCP/pull/6111/changes#diff-c58dcb5f45f5f7f5635c41deb0acd66b8886f6f952bd31dc50e96b6a22f51810R86. 

In short - a new token is obtained from fpaClientBuilder on every Azure call.

backend/pkg/azure/cachedreader/resource_skus.go
miguelsorianod: After discussing this with other stakeholders, let's do the following:
keep using the region where backend is deployed.
However, change the interface methods to also require a location, and at this point, check that if the provided location doesn't match the deployed location then return an error. This allows us to already change the interface but not need to implement the internal changes at the cache level. An alternative is to implement the internal caching level changes so it is per location per tenant.

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,228 @@
...
+			}
+			skus = append(skus, sku)
+		}
+	}
+	return skus, nil
+}
+
+// deepCopyResourceSKUSlice returns a deep copy of the SKU slice so callers cannot
+// mutate nested fields of values held in the cache.
+func deepCopyResourceSKUSlice(skus []*armcompute.ResourceSKU) ([]*armcompute.ResourceSKU, error) {
```
miguelsorianod: No need to perform deepcopies. Just return as is. Document that consumers shouldn't mutate the returned results

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,228 @@
...
+// resourceSKUsClientBuilder builds subscription-scoped Resource SKUs clients (e.g. FPA builder).
+type resourceSKUsClientBuilder interface {
+	ResourceSKUsClient(tenantID string, subscriptionID string) (azureclient.ResourceSKUsClient, error)
+}
+
+// ResourceSKUsCachedReader exposes cached reads of Microsoft.Compute Resource SKUs for virtualMachines,
+// scoped to the single Azure location (region) this backend is deployed in.
+type ResourceSKUsCachedReader interface {
+	// ListVirtualMachineSKUs returns the cached VM Resource SKU list for the subscription.
+	ListVirtualMachineSKUs(ctx context.Context, tenantID, subscriptionID string) ([]*armcompute.ResourceSKU, error)
```
miguelsorianod: In general we shouldn't add abstractions on top of the cache. However, for this specific case I can see a bit of added value of returning the resources so we can keep it like this.

backend/pkg/azure/cachedreader/resource_skus.go
miguelsorianod: That sounds good

backend/pkg/azure/cachedreader/first_party_application_cached_readers.go
miguelsorianod: Understood. That seems would cover the case where we get a new token when refresh occurs

backend/pkg/azure/cachedreader/resource_skus.go
miguelsorianod: but isn't the context different per request to the cache interface?

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,228 @@
...
+	// GetVirtualMachineSKU looks up one VM size in the cached list.
+	GetVirtualMachineSKU(ctx context.Context, tenantID, subscriptionID, vmSize string) (*armcompute.ResourceSKU, error)
+}
+
+type cachedResourceSKUsEntry struct {
+	skus []*armcompute.ResourceSKU
```
miguelsorianod: this cache entry are not "skus" (general). This are compute resource SKUs. And within that, of the Virtual Machine type

should the interface be VirtualMachineResourceSKUsCachedReader?

backend/pkg/azure/cachedreader/resource_skus.go
miguelsorianod: By adding abstracted methods on top, as well as returning different errors this is not a pull-through cache like the role definitions cached reader anymore. We should also be clear about this in the documentation if we want to go that route

backend/pkg/azure/cachedreader/resource_skus.go
shitaljante: make sense, I'll update all related fields(struct, interfaces) ResourceSKUs -> VirtualMachineResourceSKUs

backend/pkg/azure/cachedreader/resource_skus.go
miguelsorianod: This means however that we stop honoring the cancellation if the parent context is cancelled. If the parent's context signals that there should be cancellation then the cancellation should be honored

backend/pkg/app/backend.go
```diff
@@ -933,6 +902,7 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
 				go externalAuthMetricsController.Run(ctx, 1)
 				go placementSyncController.Run(ctx, 20)
 				go cosmosMigrationController.Run(ctx, 5)
+				go virtualMachineResourceSKUsCachedReaderController.Run(ctx, 5)
```
miguelsorianod: Set threadiness to 20

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,325 @@
...
+	// service itself runs.
+	location string
+	// cache maps a lowercased subscription ID (strings.ToLower(subscriptionID)) to a
+	// *cachedResourceSKUsEntry holding the location-filtered VM Resource SKU list and/or the
+	// cached list error for that subscription.
+	cache *lrucache.LRUExpireCache
+
+	// queue is where incoming work is placed to de-dup refresh requests for the same
+	// subscription and to allow "easy" rate limited requeues on errors.
+	queue workqueue.TypedRateLimitingInterface[VirtualMachineSKUKey]
```
miguelsorianod: It seems incorrect that in one hand we have the key of the queue subscription+tenant but on the other hand the key of the cache is just subscription. Be consistent with whatever decision is taken

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,325 @@
...
+
+type cachedResourceSKUsEntry struct {
+	vmSKUs []*armcompute.ResourceSKU
+	// err contains the error returned by the Resource SKUs list call. nil if the list succeeded.
+	err error
+}
+
+// VirtualMachineSKUKey is the workqueue key used to drive on-demand refreshes of a subscription's
+// cached VM Resource SKUs.
+type VirtualMachineSKUKey struct {
```
miguelsorianod: unexport as we don't want to expose it outside

miguelsorianod: virtualMachineResourceSKUKey

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,325 @@
...
+	queue workqueue.TypedRateLimitingInterface[VirtualMachineSKUKey]
+}
+
+// NewVirtualMachineResourceSKUsCachedReaderController returns a
+// VirtualMachineResourceSKUsCachedReaderController that caches VM Resource SKUs lists from
+// clients created by clientBuilder, filtered to location. Cache keys are lowercased subscription
+// IDs. The returned controller must be started with Run before any caller relying on a cache miss
+// being refreshed will receive a response; callers hitting an already-warm cache do not require
+// Run to have been called.
+func NewVirtualMachineResourceSKUsCachedReaderController(clientBuilder resourceSKUsClientBuilder, location string) *VirtualMachineResourceSKUsCachedReaderController {
```
miguelsorianod: Don't receive a clientBuilder resourceSKUsClientBuilder, receive a  azureclient.FirstPartyApplicationClientBuilder instead. And also name the types and constructors so it's clear it's limited to FPA. This will intentionally prevent being used by other identities

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,325 @@
...
+
+	azureclient "github.com/Azure/ARO-HCP/backend/pkg/azure/client"
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+const (
+	// resourceSKUsCacheSuccessFreshnessTTL is how long a successful cached SKU list may be
+	// returned without refreshing from Azure when a new request arrives. Also used as the
+	// LRUExpireCache per-entry TTL for successful entries.
```
miguelsorianod: I am not sure I understand this description. This is how long a successfully cached entry remains after being cached.

Similar comment applies to failure

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,325 @@
...
+)
+
+// resourceSKUsClientBuilder builds subscription-scoped Resource SKUs clients (e.g. FPA builder).
+type resourceSKUsClientBuilder interface {
+	ResourceSKUsClient(tenantID string, subscriptionID string) (azureclient.ResourceSKUsClient, error)
+}
+
+// VirtualMachineResourceSKUsCachedReader exposes cached reads of Microsoft.Compute Resource SKUs for virtualMachines,
+// scoped to the single Azure location (region) this backend is deployed in.
+// Unlike RoleDefinitionsCachedReader, this is not a pull-through cache: it does not mirror a
```
miguelsorianod: A follow-up task for you will be to change RoleDefinitionsCachedReader to also be controller-based, as it suffers from the same issues as we had with the pull-through based for this one.
This also means this part of the comment will stop to apply.

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,325 @@
...
+type VirtualMachineResourceSKUsCachedReader interface {
+	// ListVirtualMachineSKUs returns the cached VM SKU list. The returned slice and its entries are shared with the cache; callers must
+	// not mutate them.
+	ListVirtualMachineSKUs(ctx context.Context, tenantID, subscriptionID string) ([]*armcompute.ResourceSKU, error)
+	// GetVirtualMachineSKU looks up one VM size in the cached list. The returned value is shared
+	// with the cache; callers must not mutate it.
+	GetVirtualMachineSKU(ctx context.Context, tenantID, subscriptionID, vmSize string) (*armcompute.ResourceSKU, error)
+}
+
+type cachedResourceSKUsEntry struct {
```
miguelsorianod: cachedResourceSKUsEntry -> cachedComputeVMResourceSKUsEntry? the attribute within seems to indicate this is designed to be specific to vm compute skus

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,325 @@
...
+// on the underlying ResourceSKUsClient. Error shapes also differ from calling ResourceSKUsClient
+// directly: list errors are wrapped, and GetVirtualMachineSKU synthesizes a "not found" error
+// that the raw client would never return.
+//
+// The returned *armcompute.ResourceSKU values (and slices of them) are shared with the cache and
+// with other concurrent callers. Callers must treat them as read-only and must not mutate them.
+type VirtualMachineResourceSKUsCachedReader interface {
+	// ListVirtualMachineSKUs returns the cached VM SKU list. The returned slice and its entries are shared with the cache; callers must
+	// not mutate them.
+	ListVirtualMachineSKUs(ctx context.Context, tenantID, subscriptionID string) ([]*armcompute.ResourceSKU, error)
```
miguelsorianod: https://github.com/Azure/ARO-HCP/pull/6111#discussion_r3665892616 This doesn't seem to have been addressed for the methods

backend/pkg/azure/cachedreader/resource_skus_test.go
```diff
@@ -0,0 +1,462 @@
...
+
+// testListOptions is the ResourceSKUsClientListOptions every list call is expected to use,
+// filtered to testLocation.
+func testListOptions() *armcompute.ResourceSKUsClientListOptions {
+	return &armcompute.ResourceSKUsClientListOptions{
+		Filter: ptr.To(resourceSKUsListFilterForLocation(testLocation)),
+	}
+}
+
+type fakeResourceSKUsClientBuilder struct {
```
miguelsorianod: Don't, use the mock fpa one (generate with mockgen if it doesn't exist already) once you've changed the design as mentioned in https://github.com/Azure/ARO-HCP/pull/6111/changes#r3684251733

backend/pkg/azure/cachedreader/resource_skus_test.go
```diff
@@ -0,0 +1,462 @@
...
+		return nil, f.err
+	}
+	client, ok := f.clients[subscriptionID]
+	if !ok {
+		return nil, errors.New("no client for subscription")
+	}
+	return client, nil
+}
+
+func makeVMResourceSKU(name, family string) *armcompute.ResourceSKU {
```
miguelsorianod: makeVMResourceSKU, makeDiskResourceSKU, skuListPager are package-level functions, intended for tests. The names are too generic. Make clear it's for tests. makeTest...

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,325 @@
...
+	// GetVirtualMachineSKU looks up one VM size in the cached list. The returned value is shared
+	// with the cache; callers must not mutate it.
+	GetVirtualMachineSKU(ctx context.Context, tenantID, subscriptionID, vmSize string) (*armcompute.ResourceSKU, error)
```
miguelsorianod: With this design of the Get, a consumer of  `GetVirtualMachineSKU` won't be able to differentiate between the machine not being found vs another error in the api call being returned. For now this is ok, but if we end up detecting that the consumers of GetVirtualMachineSKU need to differentiate between those issues, we will need to adapt this. Maybe returning something like `ComputeResourceSKUVirtualMachineNotFoundError` and documenting it so consumers of it can check that specific error.
Let's keep it as it is for now, but let's keep this in mind

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,325 @@
...
+	azureclient "github.com/Azure/ARO-HCP/backend/pkg/azure/client"
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+const (
+	// resourceSKUsCacheSuccessFreshnessTTL is how long a successful cached SKU list may be
+	// returned without refreshing from Azure when a new request arrives. Also used as the
+	// LRUExpireCache per-entry TTL for successful entries.
+	resourceSKUsCacheSuccessFreshnessTTL = 20 * time.Minute
```
miguelsorianod: Let's increase the SKU freshness to for example 1 hour. The sku data doesn't seem likely to change that frequently

backend/pkg/azure/cachedreader/resource_skus.go
miguelsorianod: After discussing further the specific behavior of the sku cache, we want to change the way it behaves:

* We only cache the error if we never had data for that specific cache key. This includes caching context cancelled errors and deadline errors.
* Once we have a first non error value successfully stored for that key, we always return the latest valid retrieved one

The cache essentially becomes a "while there's no entry we pull-through cache, when there's an entry already, we only update the cache when there's a successful retrieval from the source".

This achieves several things:
* We minimize the risk that the consumer will be blocked a big amount of time because of the cache for that key being initially empty for a long time
* We allow consumers to compute with the latest valid data

The downside of it is that if the source keeps returning errors in a prolonged amount of time we will have older and older data. The idea is that we will deal with anomalies where we don't get entries for a prolonged amount of time via metrics+alerts: to mitigate this, we want to add alerting so that when the controller error ratio of the cache is elevated we trigger an alert. The controllers already provide the metrics to support this alerting. This has to be done for all environments. We are not going to block the merging of the PR/functionality but it must be performed. While we don't have alerting, because we have the success/failures TTLs that should mitigate it + we expect to rollout new code relatively frequently that will empty the caches anyway and if those errors persist we will notice as we cache the errors initially until the key has a valid entry.

Finally, I want to reiterate that the behavior described here is particular to the SKU cache, adapted to how frequently the nature of the data changes, whether we consider the consequences of having a relatively long time outdated data acceptable and so on. Each cache must be designed adapted to its needs and the data it caches. For other caches our decisions might be completely different and opposite to this one. Therefore, the design of each cache must be evaluated carefuly considering those aspects, and copying other existing ones without going over that process is not a good approach.

shitaljante: Thank you for details. I agree that this behavior is specific to the SKU cache and that other caches should be evaluated independently based on the nature of their data and the impact of serving stale values.
I'll incorporate the required changes for the SKU cache. 
My understanding is that we will:

- Cache errors only until the first successful value is available for a given key.
- Once a valid value exists, always serve the latest valid cached value and only update it when a subsequent retrieval succeeds.
- Use metrics and alerting to detect prolonged failures and ensure we are aware when the cached data may become stale. (Follow-up task)

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,385 @@
...
+	// cachedAt is when vmSKUs was last successfully refreshed. It drives ensureCached's
+	// staleness check for scheduling a background refresh; it does not affect whether a
+	// successful entry is returned while it remains in the cache (up to
+	// resourceSKUsCacheSuccessEntryTTL).
+	cachedAt time.Time
+}
+
+// virtualMachineResourceSKUKey identifies a tenant/subscription pair's cached VM Resource SKUs.
+// It is used, unmodified, both as the workqueue key that drives on-demand refreshes and as the
+// cache key itself, so the two always agree. Lookups are case-sensitive: callers passing the
```
miguelsorianod: No. Treat them as case-insensitive

backend/pkg/azure/cachedreader/resource_skus.go
```diff
@@ -0,0 +1,385 @@
...
+// past resourceSKUsCacheSuccessEntryTTL), there is nothing useful to serve: this queues a refresh
+// and polls using ctx, the caller's own context, so an individual caller's cancellation or
+// timeout only stops that caller's wait; it can never cancel or affect the underlying Azure call,
+// which runs under the controller's own Run context and may be shared with other concurrent
+// callers refreshing the same subscription.
+func (c *FPAVirtualMachineResourceSKUsCachedReaderController) ensureCached(ctx context.Context, tenantID, subscriptionID string) (*cachedComputeVMResourceSKUsEntry, error) {
+	skuKey := virtualMachineResourceSKUKey{TenantID: tenantID, SubscriptionID: subscriptionID}
+	if value, ok := c.cache.Get(skuKey); ok {
+		entry := value.(*cachedComputeVMResourceSKUsEntry)
+		if entry.vmSKUs != nil && c.clock.Since(entry.cachedAt) > resourceSKUsCacheSuccessFreshnessTTL {
```
miguelsorianod: When we get a hit we don't want to renew the freshness of the cache.

We have a single TTL the 1 hour one.

If the entry is in the cache, we just return it.
If the entry it's not in the cache then we enqueue the key in the workqueue and wait until we have it in the cache (PollUntil ...) at which point will be added with the 1hour ttl by the worker.

Remove this. Keep only resourceSKUsCacheSuccessFreshnessTTL which is already being used in the worker thread.

Here, when we get a hit we return whatever is in the cache (both result and err).

Update the doc to reflect the new reality.

backend/pkg/azure/cachedreader/resource_skus.go
miguelsorianod: > Once a valid value exists, always serve the latest valid cached value and only update it when a subsequent retrieval succeeds.

Once a valid value exists, we will always serve the latest valid cached value. Once the cache automatically expires the valid value, the cache will be empty so it will be a pull-through cache in that case until it has an entry. I think we might want to do this:
* https://github.com/Azure/ARO-HCP/pull/6111#discussion_r3691855625
* https://github.com/Azure/ARO-HCP/pull/6111#discussion_r3692184666

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader.go
```diff
@@ -0,0 +1,385 @@
...
+//
+// If ref already has a successful cached entry, a failed refresh here never overwrites it: the
+// cache is left untouched so the last known-good SKU list keeps being served by ensureCached
+// until a later refresh succeeds or the entry ages past resourceSKUsCacheSuccessEntryTTL. If ref
+// has never had a successful entry, the error itself is cached (including
+// context-cancellation/deadline errors) for resourceSKUsCacheErrorFreshnessTTL so repeat callers
+// fail fast instead of blocking, and so the short TTL naturally triggers a retry once it elapses.
+func (c *FPAVirtualMachineResourceSKUsCachedReaderController) SyncOnce(ctx context.Context, ref virtualMachineResourceSKUKey) error {
+	skus, listErr := c.listVirtualMachineSKUsFromAzure(ctx, ref.TenantID, ref.SubscriptionID)
+	if listErr != nil {
```
miguelsorianod, line 265: SyncOnce is only called when:
* SyncOnce itself returns an error
* The ensureCached method enqueues the key. Which should happen when the cache doesn't have an entry.

With that, if SyncOnce itself never returns an error we know that SyncOnce will only execute when the cache is empty. Which means it should transparently cache what it receives.

Only SyncOnce should write into the cache.

This conceptually has to be:

```
	cacheKey := strings.ToLower(ref.SubscriptionID)
	if _, ok := c.cache.Get(cacheKey); ok {
		return nil
	}

	skus, listErr := c.listVirtualMachineSKUsFromAzure(ctx, ref.TenantID, ref.SubscriptionID)
	entry := &cachedResourceSKUsEntry{
		skus: skus,
		err:  listErr,
	}
	ttl := resourceSKUsCacheSuccessFreshnessTTL
	if listErr != nil {
		ttl = resourceSKUsCacheErrorFreshnessTTL
	}
	c.cache.Add(cacheKey, entry, ttl)

	return nil
```

backend/pkg/azure/cachedreader/resource_skus.go
miguelsorianod: Additionally, depending on how https://github.com/Azure/ARO-HCP/pull/6111#discussion_r3691872290 evolves we might actually not need to do anything around controller requeue for this specific case as SyncOnce should never return an error. However it will still be good that we work in figuring out how to do it because we might want it for all controllers

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader.go
```diff
@@ -0,0 +1,385 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: The filename is outdated (and the test one) as now the types are more specific. Update

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader.go
shitaljante: With this approach, the statement from https://github.com/Azure/ARO-HCP/pull/6111#discussion_r3685164023 "while there's no entry, we pull-through cache; when there's an entry already, we only update the cache when there's a successful retrieval from the source" would not be fully valid. 
We are currently returning the cached entry even when it contains only an error. I think the check should be based on whether we have a successful cached SKU list, so that we continue retrying until we get a successful response:
```
    if cachedEntry, ok := c.cache.Get(cacheKey); ok && cachedEntry.(*cachedComputeVMResourceSKUsEntry).err == nil {
        return nil
    }
    
	skus, listErr := c.listVirtualMachineSKUsFromAzure(ctx, ref.TenantID, ref.SubscriptionID)
	entry := &cachedResourceSKUsEntry{
		skus: skus,
		err:  listErr,
	}
	ttl := resourceSKUsCacheSuccessFreshnessTTL
	if listErr != nil {
		ttl = resourceSKUsCacheErrorFreshnessTTL
	}
	c.cache.Add(cacheKey, entry, ttl)

	return nil
```

shitaljante: As Discussed,  Removed check for 
```
  if cachedEntry, ok := c.cache.Get(cacheKey); ok && cachedEntry.(*cachedComputeVMResourceSKUsEntry).err == nil {
        return nil
    }
```

backend/go.mod
```diff
@@ -18,6 +19,7 @@ require (
 	github.com/go-logr/logr v1.4.3
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
+	github.com/openshift-online/maestro v0.0.0-20260213014104-081c1f6df17b
```
miguelsorianod: Why has this changed? it seems unrelated

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader_test.go
```diff
@@ -0,0 +1,514 @@
...
+		ctrl := gomock.NewController(t)
+		mockClient := azureclient.NewMockResourceSKUsClient(ctrl)
+		mockClient.EXPECT().NewListPager(testListOptions()).Return(makeTestSKUListPager([]*armcompute.ResourceSKU{vmSKU}, nil)).Times(1)
+		mockBuilder := azureclient.NewMockFirstPartyApplicationClientBuilder(ctrl)
+		mockBuilder.EXPECT().ResourceSKUsClient(testTenantID, testSubscriptionID).Return(mockClient, nil).Times(1)
+		controller := NewFPAVirtualMachineResourceSKUsCachedReaderController(mockBuilder, testLocation)
+		key := newVirtualMachineResourceSKUKey(testTenantID, testSubscriptionID)
+
+		require.NoError(t, controller.SyncOnce(ctx, key))
+		require.NoError(t, controller.SyncOnce(ctx, key), "second SyncOnce must early-return without calling Azure")
```
miguelsorianod, line 299: This test seems incorrect and/or outdated. It's at the least non deterministic. There's no guarantee with the current logic that the second SyncOnce will early-return without calling azure

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader_test.go
```diff
@@ -0,0 +1,514 @@
...
+		ctrl := gomock.NewController(t)
+		mockClient := azureclient.NewMockResourceSKUsClient(ctrl)
+		mockClient.EXPECT().NewListPager(testListOptions()).Return(makeTestSKUListPager(nil, errors.New("transient"))).Times(1)
+		mockBuilder := azureclient.NewMockFirstPartyApplicationClientBuilder(ctrl)
+		mockBuilder.EXPECT().ResourceSKUsClient(testTenantID, testSubscriptionID).Return(mockClient, nil).Times(1)
+		controller := NewFPAVirtualMachineResourceSKUsCachedReaderController(mockBuilder, testLocation)
+		key := newVirtualMachineResourceSKUKey(testTenantID, testSubscriptionID)
+
+		require.NoError(t, controller.SyncOnce(ctx, key))
+		require.NoError(t, controller.SyncOnce(ctx, key), "second SyncOnce must early-return even when the cached entry is an error")
```
miguelsorianod, line 317: Same as https://github.com/Azure/ARO-HCP/pull/6111/changes#r3703571827

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader_test.go
```diff
@@ -0,0 +1,514 @@
...
+		_, err := controller.ListVirtualMachineSKUs(ctx, testTenantID, testSubscriptionID, testLocation)
+		require.Error(t, err)
+		assert.ErrorContains(t, err, "transient")
+	})
+}
+
+// TestFPAVirtualMachineResourceSKUsCachedReaderController_SuccessTTLExpiryRefreshes verifies that
+// after the success TTL expires the next SyncOnce pulls through from Azure again (no stale-while-
+// revalidate): ensureCached only serves an entry while it remains in the LRUExpireCache.
+func TestFPAVirtualMachineResourceSKUsCachedReaderController_SuccessTTLExpiryRefreshes(t *testing.T) {
```
miguelsorianod, line 328: Same as https://github.com/Azure/ARO-HCP/pull/6111/changes#r3703571827. This seems probably outdated

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader_test.go
```diff
@@ -0,0 +1,514 @@
...
+		require.NoError(t, controller.SyncOnce(ctx, newVirtualMachineResourceSKUKey(testTenantID, testSubscriptionID)))
+
+		got, err := controller.GetVirtualMachineSKU(ctx, testTenantID, testSubscriptionID, testLocation, testVMSize)
+		require.NoError(t, err)
+		require.NotNil(t, got)
+		assert.Equal(t, testVMSize, *got.Name)
+		assert.Equal(t, testVMFamily, *got.Family)
+	})
+
+	t.Run("returns an error when the VM size is not found", func(t *testing.T) {
```
miguelsorianod, line 568: Missing test case that covers when the own API call returns an error

shitaljante: Added test cases to cover the internal errors

backend/pkg/azure/cachedreader/virtual_machine_resource_skus_cached_reader_test.go
shitaljante: This is still valid after new changes added

PR #6116, opened 2026-07-16
docs(ci): cap prod e2e slots at 3/sub for deny-assignment limit

roivaz:
## Why

The ARO-HCP RP creates *sharded* (per-cluster) Azure **deny assignments** in PROD (~21 per HCP). The Authorization RP caps deny assignments at **2000 per subscription**, which limits a subscription to **~92 concurrent HCP clusters**, independent of role-assignment quota or identity-pool size.

With the PROD identity-container pool sizing (`identity_container_count: 25`, the max HCPs a single suite run provisions concurrently), the safe slot count is `floor(92 / 25) = 3` per subscription.

This constraint is **AME-only**.

## What

- Reduce both PROD pools in `test/e2e-config/e2e-slots.yaml` from `slot_count: 12` to `3`, with an inline comment explaining the cap.
- Document the constraint as **Bottleneck 3** in `docs/ci/identity-leasing.md` "Scaling Constraints".

## Depends on

- openshift/release#82011 — reduces the corresponding Boskos slot pools (`aro-hcp-prod-shard{0,1}-slot`) from 12 to 3. **That PR must merge first** so Boskos never advertises more slots than this catalog backs.

Holding this PR until the release-side change lands.

### DIFF
--- a/docs/ci/identity-leasing.md
+++ b/docs/ci/identity-leasing.md
@@ -221,7 +221,7 @@ This applies to **AME environments only (STG and PROD)**: the RP only creates de
 max-slots-per-sub = floor(max-hcps-per-sub / identity-container-count-per-slot)
 ```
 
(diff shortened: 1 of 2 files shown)


roivaz:
/hold

Depends on openshift/release#82011 (Boskos slot reduction), which must merge first. Remove the hold once that lands.

PR #6118, opened 2026-07-16
Gracefully handle soft-delete of getOrCreate resources

deads2k:
With a soft-delete, we can end up in a case where GET indicates 404, a create indicates a 409/conflict, then a GET indicates 404.  We've tried to fix in two ways.
1. when the resource we're running a sync on doesn't exist, do nothing.
2. when a 404, 409, 404 loop happens, delay for just longer than TTL and
   try again.

### DIFF
--- a/internal/database/service_provider_nodepool.go
+++ b/internal/database/service_provider_nodepool.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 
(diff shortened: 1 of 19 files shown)


internal/database/service_provider_nodepool.go
```diff
@@ -58,32 +60,44 @@ func GetOrCreateServiceProviderNodePool(
...
+	case err == nil:
+		return existingServiceProviderNodePool, nil
+	case IsNotFoundError(err):
+		if len(secondAttempt) >= 1 && secondAttempt[0] {
+			return nil, utils.TrackError(fmt.Errorf("second NotFound, Conflict, NotFound error: %w", err))
+		}
+		select {
+		case <-ctx.Done():
+			return nil, utils.TrackError(ctx.Err())
+		case <-time.After((SoftDeleteTTLSeconds + 1) * time.Second):
```
stevekuznetsov, line 94: Ideally this happens by re-queueing at the controller workqueue level, so we don't block the entire reconcile goroutine for 31s, but if that's not doable, c'est la vie

stevekuznetsov (approved):
/lgtm
/approve
/hold for one comment

internal/database/service_provider_nodepool.go
deads2k: We're very near the https://github.com/Azure/ARO-HCP/pull/5686.  Let's tolerate it for a few days.

PR #6122, opened 2026-07-16
Prefer latest if no gateway from minor for active versions and all candidates

machi1990:
### What

fix: prefer latest z-stream candidate when no next-minor path exists

### Why

When the next-minor channel exists but neither candidates nor active
versions are gateways, return the latest candidate instead of nil —
there is no existing next-minor path to preserve. Preferring the latest
candidate prioritizes security and bug fixes over withholding an upgrade
that cannot unlock the next minor anyway.

### Testing

justification.

- Unit tests

Review the second commit only; the first commit comes from https://github.com/Azure/ARO-HCP/pull/6110

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
+++ b/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
@@ -460,6 +460,12 @@ func FindAllUpgradeTargetVersionsInMinor(
 // This method implements the core version selection logic for all upgrade scenarios (both Y-stream and Z-stream).
 // It prioritizes versions that have an upgrade path to the next minor version (gateway versions).
 //
+// When preferLatestOverGateway is true (y-stream upgrades), the latest candidate is returned even if
+// no gateway to the next minor exists. When false (z-stream upgrades), nil is returned only when an
+// active version already has a next-minor path that would be lost by moving to a non-gateway
+// candidate; if neither candidates nor active versions can reach the next minor, the latest
+// candidate is returned instead — there is no existing next-minor path to preserve.
+//
 // Version selection algorithm:
 //  1. Query Cincinnati for all available updates from EACH active version in the target minor channel
 //  2. Filter candidates: only include versions within the target minor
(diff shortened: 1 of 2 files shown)


backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go
```diff
@@ -561,10 +577,20 @@ func selectBestVersionFromCandidates(
...
+	// prefer the latest candidate — there is no existing path to preserve.
+	for _, active := range activeVersions {
+		isGateway, err := isGatewayToNextMinor(ctx, active, cincinnatiClient, channelGroup, nextMinor)
+		if err != nil {
+			return nil, utils.TrackError(err)
+		}
+		if isGateway {
+			return nil, nil
+		}
+	}
```
deads2k, line 591: seeing it written in code, would this mean that if we somehow end up at a 4.y.z with no path to 4.y+1, then we can keep upgrading to the very latest 4.y.z that *still* has path to 4.y+1 because it's so new?  That would be unfortunate.

machi1990: > would this mean that if we somehow end up at a 4.y.z with no path to 4.y+1

For z-stream upgrade, If we reach this branch, it is because neither of the  upgrade candidates in 4.y have an upgrade path to 4.y+1. 

> then we can keep upgrading to the very latest 4.y.z that still has path to 4.y+1 because it's so new? 

Here we know that 4.y+1 is still new and/or all 4.y active versions in the cluster plus 4.y.z+1 upgrade candidates do not have an upgrade to 4.y+1; in this case yes we continue honoring z-stream upgrade within 4.y; maintaining our priority to security & bug fixes within 4.y. We can do this as we don't have the upgrade path to 4.y+1 that we'd want to preserve

> That would be unfortunate.

Yes, I don't have the data how often this happened when a new minor is released but while I was at https://github.com/Azure/ARO-HCP/pull/6110 this came up as a gap. If this at all happens without this improvement ; the following will occur.
- For existing clusters, we'll stop z-stream upgrades until the new 4.y+1 is reachable from 4.y and possible candidates. Maybe in O(week) as edges are added? 
- For initial install, all clusters will be install at a 4.y.0 final resolved version: https://github.com/machi1990/ARO-HCP/blob/b528d137b5efe7fbfdad50f533dfe8cf468752a1/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go#L279. For this case, I wonder why we chose to set _preferLatestOverGateway=false,_ https://github.com/machi1990/ARO-HCP/blob/b528d137b5efe7fbfdad50f533dfe8cf468752a1/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go#L266  thinking about it _preferLatestOverGateway=true_ will be most appropriate for initial

@deads2k thoughts?

machi1990:
Superseded by https://github.com/Azure/ARO-HCP/pull/6192

PR #6123, opened 2026-07-16
remove unnecessary comparison exclusions

deads2k:
We've completed our storage migration, so we can tidy up our comparisons now.


deads2k:
/label lgtm

simplification I owed.

PR #6124, opened 2026-07-16
[wip] admin creds via backend

deads2k:


docs/system-admin-credentials/PLAN.md
```diff
@@ -0,0 +1,302 @@
...
+cluster's system admin credentials. Nested under the cluster ARM resource ID and
+created when a `RevokeCredentials` operation fires.
+
+**Why a separate, first-class type** (review feedback from deads2k): The original
+implementation had the dispatch controller do everything inline — flip every
+credential to `AwaitingRevocation`, create the CRR desires, and poll for
+completion. deads2k asked for this to be modeled as a nested type with dedicated
+controllers instead, because:
+
+1. Revocation is a cluster-scoped event (revoking *all* active credentials), not
```
stevekuznetsov, line 83: there are two signers, so two revocation buckets. if this also models revocation for SRE signer, we need to distinguish

deads2k: > there are two signers, so two revocation buckets. if this also models revocation for SRE signer, we need to distinguish

There's a neat thing. Is it currently in use so we can vet a replacement flow?

PR #6125, opened 2026-07-16
test: track deletion times, use step IDs

stevekuznetsov:
We want to track also the time taken to delete as seen by the client and to use machine-readable markers, so we can do post-hoc analysis on this.

/label lgtm


stevekuznetsov:
/hold cancel
/retest

???

raelga:
/hold cancel
/retest-required

The only failing required check is `ci/prow/e2e-parallel`, and it's not this change. On the last run 2080626868242878464 `aro-hcp-test-local` passed, the failure is in the `aro-hcp-gather-snapshot` step: `hypershift/hostedClusterConditions` and `hypershift/nodePoolConditions` both returned no results, so the observability snapshot gate flakes. Retesting.

PR #6128, opened 2026-07-17
fix: reject missing JWT claim headers in sessiongate proxy authorization

hlipsig:
Add an explicit empty-header rejection before the owner comparison as defense in depth, and add unit tests for the middleware. The header pattern only exists when local dev is true, but we can still harden this to prevent any accidental situations in the future.

Ref:  ARO-28371

### What

Makes local dev more secure.

### Why

Defense in depth hardening.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/sessiongate/pkg/server/middleware/authorization_test.go
+++ b/sessiongate/pkg/server/middleware/authorization_test.go
@@ -1,0 +1,118 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package middleware
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


sessiongate/pkg/server/middleware/authorization_test.go
```diff
@@ -0,0 +1,125 @@
+// Copyright 2025 Microsoft Corporation
```
twolff-gh: Nit 2026

sessiongate/pkg/server/middleware/authorization_test.go
```diff
@@ -0,0 +1,125 @@
...
+		w.WriteHeader(http.StatusOK)
+	})
+
+	tests := []struct {
+		name       string
+		owner      sessiongatev1alpha1.Principal
+		headers    map[string]string
+		wantStatus int
+	}{
+		{
```
twolff-gh, line 36: Good test cases. Maybe one more for the explicit case:
``` 
 {
      name:  "empty owner name with missing header rejects",
      owner: sessiongatev1alpha1.Principal{
          Type: sessiongatev1alpha1.PrincipalTypeAzureUser,
          Name: "",
      },
      headers:    map[string]string{},
      wantStatus: http.StatusUnauthorized,
  },
```

sessiongate/pkg/server/middleware/authorization_test.go
hlipsig: Updated the new test file copyright to 2026 in 170f76835.

sessiongate/pkg/server/middleware/authorization_test.go
hlipsig: Added the empty owner name with missing header regression case in 170f76835.

PR #6132, opened 2026-07-17
Fix: Replace bash shell-out with Go-native parameter expansion in templati…

hlipsig:
…ze settings

In local dev, the Resolve function passed RegionShortSuffix and RegionShortOverride values into bash -c "echo <value>" unquoted, allowing shell metacharacter injection from the settings YAML. Replace with a Go-native expandBashSubstring function that handles ${VAR:offset:length} substring extraction using os.Expand, eliminating the shell dependency entirely.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/tooling/templatize/pkg/settings/settings.go
+++ b/tooling/templatize/pkg/settings/settings.go
@@ -15,10 +15,9 @@
 package settings
 
 import (
-	"context"
 	"fmt"
 	"os"
-	"os/exec"
+	"strconv"
 	"strings"
 
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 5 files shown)


tooling/templatize/pkg/settings/settings.go
```diff
@@ -94,22 +93,92 @@ func Resolve(ctx context.Context, environment Environment) (EnvironmentParameter
...
 	}
 	return out, nil
 }
+
+// expandBashSubstring expands a limited subset of bash-style parameter
+// expansions. It supports plain variable expansion ($VAR, ${VAR}) and
+// substring extraction (${VAR:offset}, ${VAR:offset:length}).
+// Note: negative offsets require a space after the colon (${VAR: -3});
+// ${VAR:-3} is bash default-value syntax and is not supported here.
+func expandBashSubstring(s string) (string, error) {
```
stevekuznetsov, line 117: If we don't want to use shell, we should just invent our own mechanism here - 

```yaml
regionShortSuffixConfig:
    baseVar: "whatever"
    offset: 4
    length: 3
```

Instead of trying to keep the bash interface and hope we implement something that is close enough to it that we're right. 

If we design our own interface it will not be ambiguous and we will be able to be certain we implemented it fully and correctly.

tooling/templatize/pkg/settings/settings.go
```diff
@@ -94,22 +93,92 @@ func Resolve(ctx context.Context, environment Environment) (EnvironmentParameter
...
+		}
+
+		colonIdx := strings.IndexByte(key, ':')
+		if colonIdx == -1 {
+			return os.Getenv(key)
+		}
+
+		varName := key[:colonIdx]
+		substringExpr := key[colonIdx+1:]
+		value := os.Getenv(varName)
```
twolff-gh, line 131: Does this and the one on L125 mean that whatever is passed gets used downstream and/or printed?

Low likelihood and a preexisting issue, but the template values are passed in as is an used directly. If something changes with the template and maps to an legitimate env var, that could theoretically create an exposure of that env var.

Do we want to add any allowlisting here or elsewhere?

PR #6137, opened 2026-07-17
fix: apply minor offset on OpenShift major version jumps

machi1990:
### What

fix: apply minor offset on OpenShift major version jumps

### Why

The "minor+1" approach breaks at the 4→5 boundary (4.23 jumps to 5.x, not 4.24). Add a utility that accounts for that and use it in the control plane desired version controller

### Testing

justification.

- Unit tests

https://redhat.atlassian.net/browse/ARO-28474

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -16,6 +16,7 @@ package api
 
 import (
 	"encoding/json"
+	"fmt"
 	"iter"
 	"net/http"
 	"reflect"
@@ -23,6 +24,7 @@ import (
 	"strings"
 
 	"dario.cat/mergo"
+	"github.com/blang/semver/v4"
 	jsonpatch "github.com/evanphx/json-patch"
 
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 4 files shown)


internal/api/utils.go
```diff
@@ -64,6 +66,25 @@ var AllowedChannelGroups = sets.New("stable", "fast")
...
+// major.minor.0 (patch and pre-release metadata cleared).
+//
+// Within a major, that is major.(minor+1).0. At major boundaries that do not
+// continue as major.(minor+1), a bridged next-major line is returned instead
+// (for example 4.23.x → 5.1.0).
+func NextMinorReleaseLine(v semver.Version) semver.Version {
+	// Maps OpenShift minor release lines ("x.y") that do not continue as
+	// major.(minor+1) onto the next-major release line they bridge to.
+	// Add entries when new major-boundary bridges are supported.
+	bridgedNextMinorReleaseLines := map[string]string{
```
ahitacat: Maybe instead of considering this a minor release treat it as an special case of major release and reuse `var AllowMajorUpgradePaths ` ?

I am still not sure if this upgrade will be painless, or if it will need some special treatment.

machi1990: it's a different use case to `AllowMajorUpgradePaths `.

If we re-use that, for 4.22 we might consider 5.0 as the next minor after 4.22 while in reality 4.22 can land to 4.23 or 5.0. 

The use case here is more like; what's the final minor version in a given major and where does it point to in the next major

ahitacat: Fair point, can then consider add it as a variable in the same file than AllowMajorUpgradePaths, there are similar variables there. In case of change it will be easy to handle.

machi1990: Yes, I added the variable above but didn't export it. Only a package private variable.

machi1990: This has been updated to use AllowMajorUpgradePaths following internal discussions with @deads2k

internal/api/utils_test.go
```diff
@@ -227,3 +228,47 @@ func TestNonNilSliceValues(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			got := NextMinorReleaseLine(Must(semver.ParseTolerant(tt.in)))
+			assert.Equal(t, Must(semver.ParseTolerant(tt.want)), got)
+		})
+	}
```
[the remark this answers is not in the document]
machi1990: Not needed.

machi1990:
/test e2e-parallel

The failure seems to be on KSM rotation. I pinged @ahitacat on it.

PR #6151, opened 2026-07-17
explicitly track the two HCP namespaces in the management cluster

deads2k:
We need this to more easily write things like admincredentials request.


backend/pkg/controllers/clusterpropertiescontroller/serviceprovidercluster_properties_sync.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil
+	}
+
+	hostedClusterNamespace := hostedCluster.Namespace
+	hostedClusterName := hostedCluster.Name
+	if len(hostedClusterNamespace) == 0 || len(hostedClusterName) == 0 {
+		return nil
+	}
+
+	controlPlaneNamespace := fmt.Sprintf("%s-%s", hostedClusterNamespace, strings.ReplaceAll(hostedClusterName, ".", "-"))
```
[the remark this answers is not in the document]
deads2k: that's a silly dep to copy.

backend/pkg/controllers/clusterpropertiescontroller/serviceprovidercluster_properties_sync.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/Azure/ARO-HCP/backend/pkg/kubeapplierhelpers"
+	"github.com/Azure/ARO-HCP/backend/pkg/listers"
+	"github.com/Azure/ARO-HCP/internal/database"
+	dblisters "github.com/Azure/ARO-HCP/internal/database/listers"
+	unionkubeapplierinformers "github.com/Azure/ARO-HCP/internal/database/unioninformers/kubeapplier"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// serviceProviderClusterPropertiesSyncer synchronizes namespace fields on the
+// ServiceProviderCluster from the observed HostedCluster ReadDesire content:
```
stevekuznetsov, line 36: How do you know what namespace to put the HostedCluster readdesire into if one of the fields you get out of it is the namespace for the object itself? This smells weird

deads2k: > How do you know what namespace to put the HostedCluster readdesire into if one of the fields you get out of it is the namespace for the object itself? This smells weird

Currently, the backend is NOT the source of truth for which namespace the hostedcluster is written to.  Once it is, we'll change where this is set.

stevekuznetsov: I was making the narrow point that whatever source of truth you use for the readdesire namespace is already the source for this ns

deads2k: >I was making the narrow point that whatever source of truth you use for the readdesire namespace is already the source for this ns

I'm not wiring this up to cluster-service.  Once we build it differently, We can update this.

PR #6020, opened 2026-07-10
Add changefeed support for create/modify

deads2k:
This is a new PR to avoid the noise of the old PR since it had so many resolved, but unmarked bot comments.

PR #6155, opened 2026-07-18
revert(backend): switch informers off the Cosmos change feed to recover prod (AROSLSRE-1521)

raelga:
Jira: [AROSLSRE-1521](https://redhat.atlassian.net/browse/AROSLSRE-1521)

### What

Surgical revert of the backend informers back to the Cosmos **list** path, undoing the informer switch introduced by #6020. Scope is limited to the `backend/pkg/informers` package plus its callers/tests; the shared changefeed infrastructure is intentionally left untouched.

Reverted to pre-#6020:
- `backend/pkg/informers/informers.go` (list + `ExpiringWatcher`, no `changeFeedClient`)
- `backend/pkg/informers/types.go` (`NewBackendInformers` drops the `changeFeedClient` arg)
- `backend/pkg/informers/informers_test.go`
- `backend/pkg/app/backend.go` (drops the `ResourcesDBClient` arg to `NewBackendInformers`)
- `test-integration/backend/launch/notification_test.go`, `test-integration/frontend/informer_test.go`
- `test-integration/backend/changefeed/changefeed_test.go` — removed only `TestActiveOperationInformer` (asserted near-real-time deletes, unachievable on the list/relist path). `TestChangeFeedListWatcher` and all shared helpers are kept.

### Why

Prod uksouth nodepool creates hung at `provisioningState=Accepted` service-wide. Root cause is a regression from #6020 (`Add changefeed support for create/modify`, merge `b231113aa`), specifically the commit that wired the backend nodepool/external-auth/controller informers onto the change feed.

The old list path filters documents server-side:

```sql
SELECT * FROM c WHERE LENGTH(c.resourceID) > 0 AND (...)
```

so it never observed legacy documents with a nil `resourceID`. The change feed has **no such filter**:

```text
change feed reads poison doc (resourceID == nil)
  -> processDocument returns "missing resourceID"
  -> readFeedRange returns WITHOUT advancing the continuation token
  -> that feed range is re-read forever -> permanent stall
```

Four feeds were poisoned (NodePool / ExternalAuth / Controller / ServiceProviderCluster); Cluster and Operation feeds were healthy, which matches the observed 45-pass / 32-fail split.

### Why not a full `git revert -m 1`

A full revert of #6020 is not viable and does not compile:
- It deletes `changefeed_list_watch.go`, which #6140 now depends on (fleet informers were moved onto the shared change feed).
- It collides with the kube-applier constructor signature changes and the controller removals in #6140 / #6149 / #6127.

Trialed: `git revert -m 1 b231113aa` produces 7 conflicts across 132 files. Hence this package-scoped surgical revert that restores service without disturbing the shared infra that fleet now relies on.

### Testing

Run locally (`GOTOOLCHAIN=auto`):
- `backend`: `go build ./...` ✅, `go vet ./...` ✅, full `go test ./...` ✅ (no failures)
- `test-integration`: `go build ./...` ✅, `go vet ./backend/... ./frontend/...` ✅
- `go test ./backend/changefeed/...` (shared `TestChangeFeedListWatcher`) ✅ 46.9s
- `go test ./frontend/...` ✅ 52.3s, `go test ./backend/launch/...` ✅
- `gofmt -l` on all changed files: clean

- This is a **mitigation** to recover prod. The proper forward fix (recover `resourceID` from the legacy pipe-delimited cosmos `ID` inside `CosmosGenericToInternal`, mirroring what `CosmosToInternal` already does for bare `TypedDocument`) lands as a separate PR so the change feed can be safely re-enabled afterward.
- Shared changefeed infra (`internal/database/informers`, `changefeed_list_watch.go`) and the fleet informers are deliberately **not** touched.
- The removed `TestActiveOperationInformer` cannot pass on the list path (it asserts sub-90s Deleted events; the list path only detects deletes on relist). It is replaced in spirit by the forward fix, not by this revert.

### PR Checklist

- [x] Tests added/updated where applicable (removed a test that is invalid on the restored path)
- [x] Verified locally (build, vet, tests, gofmt)
- [x] Linked Jira issue

### DIFF
--- a/backend/pkg/app/backend.go
+++ b/backend/pkg/app/backend.go
@@ -376,7 +376,6 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context, 
 
 	backendInformers := informers.NewBackendInformers(ctx,
 		b.options.ResourcesDBClient.ResourcesGlobalListers(),
-		b.options.ResourcesDBClient,
 		b.options.BillingDBClient.BillingGlobalListers(),
 	)
 
--- a/backend/pkg/informers/informers.go
+++ b/backend/pkg/informers/informers.go
@@ -24,7 +24,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
-	utilsclock "k8s.io/utils/clock"
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 
@@ -32,7 +31,6 @@ import (
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/database"
-	dbinformers "github.com/Azure/ARO-HCP/internal/database/informers"
 	"github.com/Azure/ARO-HCP/internal/utils"
 	"github.com/Azure/ARO-HCP/internal/utils/armhelpers"
 )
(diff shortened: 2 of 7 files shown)


raelga:
Closing in favor of #6156 (merged), which is the mitigation we shipped for [AROSLSRE-1521](https://redhat.atlassian.net/browse/AROSLSRE-1521).

**Why we did not go with this surgical revert (from the incident review):**

- A true `git revert -m 1` of #6020 entangles ~6 interconnected (non-conflicting but coupled) PRs, and there's low confidence the result is coherent when reassembled.
- This surgical scope compiles, but leaves later behavioral changes in place that don't belong on the list path. Concrete example raised in review: #6020 removed the informer **cooldowns** because they're unnecessary on the change-feed path; this revert switches the informers back to the list path but leaves the cooldowns removed, which would produce mass, synchronized triggers.
- We could not roll a previous rollout/commit either (untested path, no weekend on-call).

**What we shipped instead:** #6156 makes the change feed skip exactly the documents the list path already excludes server-side (`WHERE LENGTH(c.resourceID) > 0`), so it restores the pre-#6020 observable behavior without a revert and without changing informer semantics.

**Follow-ups tracked under [AROSLSRE-1521](https://redhat.atlassian.net/browse/AROSLSRE-1521):** identify/clean the rogue Cosmos record (Monday, via query), investigate why the repeated relist didn't force forward progress, add per-informer `delivered==0 && errors>0` alerting, and intentionally exercise a region down-level so rollback becomes a tested path.

PR #6157, opened 2026-07-19
feat: add new API field nodeSshPublicKeys

tmstff:
https://redhat.atlassian.net/browse/ARO-26939

### What

New (ARO-HCP) API version exposing the nodeSshPublicKeys field to customers, as part of an effort to provide ssh access to cluster nodes for customers ( https://redhat.atlassian.net/browse/ARO-26935 ).

ARM API publication and transmission to CS not part of this story, will be handled in a follow-up.

### Why

To allow the customer to provide ssh public keys, which can then be forwarded to HyperShift via Backend and CS - to ultimately enable the customer to log into cluster nodes via ssh.

### Testing

- Unit tests (for helper methods and validations)
- [Integration tests](https://github.com/Azure/ARO-HCP/tree/main/test-integration) (to make sure the API functions as expected, also with different API versions)
- [E2E tests](https://github.com/Azure/ARO-HCP/blob/main/test/e2e/README.md) (which can later on be extended to test ssh login)
    - run test against personal-dev-env: `make e2e-local/run-test TEST_NAME="Customer should persist nodeSshPublicKeys set at cluster creation and return them via ARM GET"`

* limit to 5 keys as per Jeromes request (can be changed later, as this is not part of the public API
* parse SSH keys to make sure they are valid (allows for accepting newer algorithms by using the latest crypto lib version)

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
+++ b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
@@ -137,6 +137,20 @@ model HcpOpenShiftClusterProperties {
   @visibility(Lifecycle.Read, Lifecycle.Create)
   clusterImageRegistry?: ClusterImageRegistryProfile;
 
+  /** One or more SSH public keys used for secure access to worker nodes. When set,
+   * these keys are distributed to all nodes, enabling SSH access for debugging
+   * and maintenance purposes. If changed, all keys previously set will be replaced.
+   *
+   * WARNING: Setting or changing this field currently causes the recreation of all worker nodes.
+   * This behavior is subject to change in a future OpenShift release.
+   *
+   * SSH public key format: <algorithm> <base64-encoded-key> [comment]
+   * Key constraints: algorithm + space + base64 blob required; comment optional; no newlines within key.
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  @added(Versions.v2026_09_01_preview)
+  nodeSshPublicKeys?: string[];
+
   /** Status of the cluster resource */
   @visibility(Lifecycle.Read)
   @added(Versions.v2026_06_30_preview)
--- a/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
+++ b/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
@@ -2400,6 +2400,18 @@
             "create"
           ]
         },
+        "nodeSshPublicKeys": {
+          "type": "array",
+          "description": "One or more SSH public keys used for secure access to worker nodes. When set,\nthese keys are distributed to all nodes, enabling SSH access for debugging\nand maintenance purposes. If changed, all keys previously set will be replaced.\n\nWARNING: Setting or changing this field currently causes the recreation of all worker nodes.\nThis behavior is subject to change in a future OpenShift release.\n\nSSH public key format: <algorithm> <base64-encoded-key> [comment]\nKey constraints: algorithm + space + base64 blob required; comment optional; no newlines within key.",
+          "items": {
+            "type": "string"
+          },
+          "x-ms-mutability": [
+            "read",
+            "update",
+            "create"
+          ]
+        },
         "status": {
           "$ref": "#/definitions/ResourceStatus",
           "description": "Status of the cluster resource",
@@ -2502,6 +2514,18 @@
           "description": "nodeDrainTimeoutMinutes is the grace period for how long Pod Disruption Budget-protected workloads will be\nrespected during any node draining operation. After this grace period, any workloads protected by Pod Disruption\nBudgets that have not been successfully drained from a node will be forcibly evicted. This is\nespecially relevant to cluster upgrades.\n\nValid values are in minutes and from 0 to 10080 minutes (1 week).\n0 means that the MachinePool can be drained without any time limitation.\n\nThis is the value is used a default for all NodePools. It can be overridden\nby specifying nodeDrainTimeoutMinutes for a given NodePool",
           "minimum": 0,
           "maximum": 10080,
+          "x-ms-mutability": [
+            "read",
+            "update",
+            "create"
+          ]
+        },
+        "nodeSshPublicKeys": {
+          "type": "array",
+          "description": "One or more SSH public keys used for secure access to worker nodes. When set,\nthese keys are distributed to all nodes, enabling SSH access for debugging\nand maintenance purposes. If changed, all keys previously set will be replaced.\n\nWARNING: Setting or changing this field currently causes the recreation of all worker nodes.\nThis behavior is subject to change in a future OpenShift release.\n\nSSH public key format: <algorithm> <base64-encoded-key> [comment]\nKey constraints: algorithm + space + base64 blob required; comment optional; no newlines within key.",
+          "items": {
+            "type": "string"
+          },
           "x-ms-mutability": [
             "read",
             "update",
--- a/docs/cosmos-data-flow.md
+++ b/docs/cosmos-data-flow.md
@@ -1321,6 +1321,16 @@ Each entry links to every actor that writes the field.
 | [Frontend: PUT Cluster (Create)](#put-cluster-create) | Sets from request body |
 | [ClusterBaseDomainPrefixSync](#clusterbasedomainprefixsync) | Backfills from CS if empty |
 
+### `HCPOpenShiftCluster.CustomerProperties.NodeSshPublicKeys`
+
+| Actor | When |
+|-------|------|
+| [Frontend: PUT Cluster (Create)](#put-cluster-create) | Sets from request body (requires API version `2026-09-01-preview` or later) |
+| [Frontend: PUT Cluster (Update)](#put-cluster-update) | Replaces from request body; older API versions preserve existing value via `preserveUnknownClusterFields` |
+| [Frontend: PATCH Cluster (Update)](#patch-cluster-update) | Merged from PATCH body; older API versions preserve existing value via `preserveUnknownClusterFields` |
+
+Single writer (frontend only). Stored in Cosmos but not dispatched to Cluster Service — no backend controller forwards this field. Exposed only in API version `2026-09-01-preview` and later.
+
 ### `HCPOpenShiftCluster.Identity.UserAssignedIdentities`
 
 | Actor | When |
--- a/internal/azureapi/v20240610preview/hcpopenshiftclusters_methods.go
+++ b/internal/azureapi/v20240610preview/hcpopenshiftclusters_methods.go
@@ -440,6 +440,8 @@ func preserveUnknownClusterFields(from, to *coreapi.HCPOpenShiftCluster) {
 	to.CustomerProperties.Platform.VnetIntegrationSubnetID = from.CustomerProperties.Platform.VnetIntegrationSubnetID
 	// Ingress was added in v2026_06_30_preview.
 	to.CustomerProperties.Ingress = from.CustomerProperties.Ingress
+	// NodeSshPublicKeys was added in v2026_09_01_preview.
+	to.CustomerProperties.NodeSshPublicKeys = from.CustomerProperties.NodeSshPublicKeys
 	// Visibility was added in v2025_12_23_preview.
 	if from.CustomerProperties.Etcd.DataEncryption.CustomerManaged != nil && from.CustomerProperties.Etcd.DataEncryption.CustomerManaged.Kms != nil {
 		if to.CustomerProperties.Etcd.DataEncryption.CustomerManaged == nil {
--- a/internal/azureapi/v20251223preview/hcpopenshiftclusters_methods.go
+++ b/internal/azureapi/v20251223preview/hcpopenshiftclusters_methods.go
@@ -494,6 +494,8 @@ func preserveUnknownClusterFields(from, to *coreapi.HCPOpenShiftCluster) {
 	to.CustomerProperties.Ingress = from.CustomerProperties.Ingress
 	// CryptoRestrictions was added in v2026_06_30_preview
 	to.CustomerProperties.CryptoRestrictions = from.CustomerProperties.CryptoRestrictions
+	// NodeSshPublicKeys was added in v2026_09_01_preview.
+	to.CustomerProperties.NodeSshPublicKeys = from.CustomerProperties.NodeSshPublicKeys
 }
 
 func normalizeManagedIdentity(identity *generated.ManagedServiceIdentity) *coreapi.ManagedServiceIdentity {
--- a/internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
+++ b/internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
@@ -397,6 +397,7 @@ func (v version) NewHCPOpenShiftCluster(from *coreapi.HCPOpenShiftCluster) corea
 				Autoscaling:       metadataapi.PtrOrNil(newClusterAutoscalingProfile(&from.CustomerProperties.Autoscaling)),
 				// Use Ptr (not PtrOrNil) to ensure int32 zero value is preserved in JSON response.
 				NodeDrainTimeoutMinutes: metadataapi.Ptr(from.CustomerProperties.NodeDrainTimeoutMinutes),
+				NodeSSHPublicKeys:       metadataapi.StringSliceToStringPtrSlice(from.CustomerProperties.NodeSshPublicKeys),
 				ClusterImageRegistry:    metadataapi.PtrOrNil(newClusterImageRegistryProfile(&from.CustomerProperties.ClusterImageRegistry)),
 				Etcd:                    metadataapi.PtrOrNil(newEtcdProfile(&from.CustomerProperties.Etcd)),
 				ImageDigestMirrors:      newImageDigestMirrors(from.CustomerProperties.ImageDigestMirrors),
@@ -522,6 +523,7 @@ func (c *HcpOpenShiftCluster) ConvertToInternal(existing *coreapi.HCPOpenShiftCl
 			normalizeAutoscaling(c.Properties.Autoscaling, &out.CustomerProperties.Autoscaling)
 		}
 		out.CustomerProperties.NodeDrainTimeoutMinutes = metadataapi.Deref(c.Properties.NodeDrainTimeoutMinutes)
+		out.CustomerProperties.NodeSshPublicKeys = metadataapi.StringPtrSliceToStringSlice(c.Properties.NodeSSHPublicKeys)
 		if c.Properties.ClusterImageRegistry != nil {
 			normalizeClusterImageRegistry(c.Properties.ClusterImageRegistry, &out.CustomerProperties.ClusterImageRegistry)
 		}
@@ -544,9 +546,9 @@ func (c *HcpOpenShiftCluster) ConvertToInternal(existing *coreapi.HCPOpenShiftCl
 }
 
 // preserveUnknownClusterFields copies customer-facing fields from existing that
-// this API version doesn't know about. Currently empty — no cross-version
-// customer fields exist yet between v20240610preview and v20260630preview.
-func preserveUnknownClusterFields(from, to *coreapi.HCPOpenShiftCluster) {
+// this API version doesn't know about.
+func preserveUnknownClusterFields(_, _ *coreapi.HCPOpenShiftCluster) {
+	// All fields introduced up to v2026_06_30_preview are known to this version.
 }
 
 func normalizeManagedIdentity(identity *generated.ManagedServiceIdentity) *coreapi.ManagedServiceIdentity {
--- a/internal/validation/validators.go
+++ b/internal/validation/validators.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/blang/semver/v4"
 	"github.com/google/uuid"
+	"golang.org/x/crypto/ssh"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/operation"
@@ -383,6 +384,26 @@ func MatchesRegex(_ context.Context, _ operation.Operation, fldPath *field.Path,
 		return nil
 	}
 	return field.ErrorList{field.Invalid(fldPath, *value, errorString)}
+}
+
+func ValidateSSHPublicKey(_ context.Context, _ operation.Operation, fldPath *field.Path, value, _ *string) field.ErrorList {
+	if value == nil {
+		return nil
+	}
+	if len(strings.TrimSpace(*value)) == 0 {
+		return field.ErrorList{field.Invalid(fldPath, *value, "SSH public key must not be empty")}
+	}
+	if len(*value) > 8192 {
+		return field.ErrorList{field.TooLong(fldPath, *value, 8192)}
+	}
+	_, _, _, remainder, err := ssh.ParseAuthorizedKey([]byte(*value))
+	if err != nil {
+		return field.ErrorList{field.Invalid(fldPath, *value, err.Error())}
+	}
+	if len(strings.TrimSpace(string(remainder))) > 0 {
+		return field.ErrorList{field.Invalid(fldPath, *value, "SSH public key must contain exactly one key")}
+	}
+	return nil
 }
 
 func ValidateUUID(_ context.Context, _ operation.Operation, fldPath *field.Path, value, _ *string) field.ErrorList {
--- a/test-integration/frontend/cross_version_roundtrip_test.go
+++ b/test-integration/frontend/cross_version_roundtrip_test.go
@@ -415,7 +415,7 @@ func clusterCreatePayload(clusterName, apiVersion string) []byte {
 }`, clusterName, subscriptionID, subscriptionID, subscriptionID))
 
 	case v20260901:
-		// v20260901 payload
+		// v20260901 payload — v20260630 plus nodeSshPublicKeys (the new field introduced in this version)
 		return []byte(fmt.Sprintf(`{
   "identity": {
     "type": "UserAssigned",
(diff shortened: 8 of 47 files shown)


api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-06-30-preview/openapi.json
```diff
@@ -2332,6 +2332,17 @@
             "create"
           ]
         },
+        "nodeSshPublicKey": {
+          "type": "string",
+          "description": "The SSH public key used for secure access to cluster nodes. When set, this key\nis distributed to all nodes, enabling SSH access for debugging\nand maintenance purposes.",
+          "maxLength": 8192,
+          "pattern": "^(?<type>[A-Za-z0-9._@-]+)[ \t]+(?<key>[A-Za-z0-9+/]+=*)(?:[ \t]+(?<comment>.*))?$",
```
miguelsorianod: Where do the pattern and maxlength come from?

miguelsorianod: Does this mean this field accepts one and only one SSH public key?

In the Hypershift API, the corresponding attribute in the HostedCluster:
```
	// sshKey is a local reference to a Secret that must have a "id_rsa.pub" key whose content must be the public part of 1..N SSH keys.
	// If the reference is set but none of the above requirements are met, the HostedCluster will enter a degraded state.
	// TODO(alberto): Signal this in a condition.
	// When sshKey is set, the controllers will generate a machineConfig with the sshAuthorizedKeys https://coreos.github.io/ignition/configuration-v3_2/ populated with this value.
	// This MachineConfig will be part of every payload generated by the controllers for any NodePool of the HostedCluster.
	// Changing this value will trigger a rollout for all existing NodePools in the cluster.
	// +rollout
	// +optional
	SSHKey corev1.LocalObjectReference `json:"sshKey"`
```

Which if I understand correctly technically allows providing multiple (confirm this as I just checked superficially)

Were/are we aware of this and is our design where it seems to differ intentional?

tmstff: Hey @miguelsorianod !
Thanks for the note. I verified, that you can really supply multiple ssh keys:

Created secret with 2 keys, referenced it in the HostedCluster CR and after a while checked one of the nodes:

```
kubectl --kubeconfig=/tmp/hc-kubeconfig \
    debug node/s3r5y1v5g8f9o5q-np-1-dx4tj-9jkzx \
    -it --image=busybox -- sh -c 'cat /host/home/core/.ssh/authorized_keys'
--profile=legacy is deprecated and will be removed in the future. It is recommended to explicitly specify a profile, for example "--profile=general".
Creating debugging pod node-debugger-s3r5y1v5g8f9o5q-np-1-dx4tj-9jkzx-pmnh6 with container debugger on node s3r5y1v5g8f9o5q-np-1-dx4tj-9jkzx.
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIH7jrlvxZvr2Mm5i2LaAcSdGt1PHoi5LXRWPQKsY3mJv tsteffen@tsteffen-mac
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIP0Rv/ZxC9jZUIeqd5p5Z8lFJT6IZKY0p46YCcgDSJ0T redmon-tsteffens@TIMSTEFFENSA0C4
```

-> will allow multiple ssh keys in the ARM API!

tmstff: done ✅

tmstff: removed the pattern. Checks are now in the validate.go and validate_clusters.go.
Limit to 5 keys is wish from Gerome. Limit to 8192 bytes / key is a sensible option to prevent buffer overrun exploits etc. . Instead of regexp, we parse key with ssh lib.
All can be changed if we see need, not part of the api.

internal/validation/validators.go
```diff
@@ -384,6 +385,19 @@ func MatchesRegex(_ context.Context, _ operation.Operation, fldPath *field.Path,
...
+		return nil
+	}
+	if len(*value) > 8192 {
+		return field.ErrorList{field.TooLong(fldPath, *value, 8192)}
+	}
+	if _, _, _, _, err := ssh.ParseAuthorizedKey([]byte(*value)); err != nil {
+		return field.ErrorList{field.Invalid(fldPath, *value, err.Error())}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
tmstff: should be fixed, feel free to re-check.

test/e2e/cluster_node_ssh.go
```diff
@@ -0,0 +1,116 @@
...
+	// Deadline for v20260901preview API deployment in non-dev environments
+	timeBombDeadline := framework.Must(time.Parse(time.RFC3339, "2026-10-31T00:00:00Z"))
+
+	It("should persist nodeSshPublicKeys set at cluster creation and return them via ARM GET",
+		labels.RequireNothing,
+		labels.Critical,
+		labels.Positive,
+		labels.AroRpApiCompatible,
+		labels.CreateCluster,
+		func(ctx context.Context) {
```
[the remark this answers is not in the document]
tmstff: outdated

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
```diff
@@ -2507,6 +2519,18 @@
             "update",
             "create"
           ]
+        },
+        "nodeSshPublicKeys": {
+          "type": "array",
+          "description": "One or more SSH public keys used for secure access to cluster nodes. When set,\nthese keys are distributed to all nodes, enabling SSH access for debugging\nand maintenance purposes.",
```
miguelsorianod: We should document the format of a value. "As a user, how should I specify a SSH Public key here?"

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
```diff
@@ -2507,6 +2519,18 @@
             "update",
             "create"
           ]
+        },
+        "nodeSshPublicKeys": {
```
miguelsorianod, line 2523: Will changing this have an effect on the existing nodepools? will new nodepools be created and the previous ones removed? or will only affect when new nodes are created?
This should be clear in the documentation

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
tmstff: added.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
tmstff: @miguelsorianod added a note - please check if that's fine

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -137,6 +137,27 @@ model HcpOpenShiftClusterProperties {
...
+   * these keys are distributed to all nodes, enabling SSH access for debugging
+   * and maintenance purposes.
+   *
+   * WARNING: Changing the value of nodeSshPublicKeys will trigger a rollout for all nodepools.
+   *          In this process fresh nodes will be created that eventually replace the existing ones.
+   *
+   * SSH public key format: <algorithm> <base64-encoded-key> [comment]
+   *
+   * Example:
+   * ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOMqqnkVzrm0SdG6UOoqKLsabgH5C9okWi0dh2l9GKJl user@host
```
bennerv: we don't need to provide the example in the description, we should be able to do it in our example swagger.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -137,6 +137,27 @@ model HcpOpenShiftClusterProperties {
...
+   *
+   * Key constraints:
+   * - Algorithm + space + base64 blob required
+   * - Comment optional (typically user@host)
+   * - No newlines within key
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  @added(Versions.v2026_09_01_preview)
+  nodeSshPublicKeys?: string[];
```
bennerv, line 152: should we set a `@maxLength` on this?

bennerv, line 152: The only thing that gives me hesistation is that Azure has a resource type of `Microsoft.Compute/sshPublicKeys`.  If we ever want to support that in the future, we may want to pull this into a sshKeys struct and support string[] and resourceID[] and combine them in the future.  Thoughts?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
tmstff: ok!

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
tmstff: We had a team discussion on this and the preferred solution was to define limits but not make them visible in the API spec. This way we can change the limits later on (e.g. when Gerome deicides so) just by changing the RP-backend. No ARM-api release necessary. The general opinion was that there's a little bit of smell to it, but we found it still OK.

Currently we've got limits for the key count (5) and the key length (8192) ion validate_cluster.go and validators.go .

What do you think @bennerv ?

tmstff: As it seems very unclear to me whether or not this will be required at some point, I would suggest leave it as is for now - and add a second field nodeSshPublicKeyResourceIds when required. Can still combine them, as we can access both fields from the same document in comos. Would still feel like a clean solution to me. This way those 2 possibilities have an independent api-lifecycle and can be retired independently. Also it shines in terms of simplicity (especially if the resourceIds don't become relevant).

Alright with you @bennerv ?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
tmstff: doc should explain the behaviour now

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
tmstff: After discussion with @bennerv the current idea is to change it to a nested structure like
```
nodeAccess {
   sshPublicKeys: ["key one", "key two"]
}
```
so in future we can extend it without having to add to level entries, e.g. for `Microsoft.Compute/sshPublicKeys` in case that becomes relevant:
```
nodeAccess {
   sshPublicKeys: ["key one", "key two"],
   sshPublicKeyResourceIds: ["id 1", "id 2"]
}
```
awaiting team feedback.
Changes nothing for the OCM & CS implementation, as we would still pass a single `nodeSshPublicKeys` field (which would then contain a combination of those).

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -137,6 +137,20 @@ model HcpOpenShiftClusterProperties {
...
+   *
+   * WARNING: Setting or changing this field currently causes the recreation of all worker nodes.
+   * This behavior is subject to change in a future OpenShift release.
+   *
+   * SSH public key format: <algorithm> <base64-encoded-key> [comment]
+   * Key constraints: algorithm + space + base64 blob required; comment optional; no newlines within key.
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  @added(Versions.v2026_09_01_preview)
+  nodeSshPublicKeys?: string[];
```
mbarnes, line 152: `Ssh` should be `SSH` in client SDKs.  Use [@clientName](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/decorators/#@Azure.ClientGenerator.Core.clientName) to override.
(See elsewhere in the spec where I changed `Cidr` to `CIDR`.)

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -137,6 +137,20 @@ model HcpOpenShiftClusterProperties {
   @visibility(Lifecycle.Read, Lifecycle.Create)
   clusterImageRegistry?: ClusterImageRegistryProfile;
 
+  /** One or more SSH public keys used for secure access to worker nodes. When set,
+   * these keys are distributed to all nodes, enabling SSH access for debugging
+   * and maintenance purposes. If changed, all keys previously set will be replaced.
+   *
```
machi1990, line 143: I suppose changing the keys causes nodes to be rebooted, correct? If so, let's document it.
See the corresponding description from Hypershift https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/hostedcluster_types.go#L702

Something similar to https://github.com/tmstff/ARO-HCP/blob/a398fcbe50f49ae17e9d0a16ee83ca56a932a01c/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp#L109

Also let's make sure that the doc team is aware of this

tmstff: @machi1990 is line 144 sufficient? ( WARNING: Setting or changing this field currently causes the recreation of all worker nodes.)

machi1990: You are correct. I totally missed it, sorry for the noise

tmstff: no problem :-)

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mbarnes: Discussed in Slack but for the record here, I suggested using a [discriminated type](https://typespec.io/docs/standard-library/discriminated-types/) for the array elements instead of having multiple arrays.

So you could keep a single array field
```
nodeSshPublicKeys?: sshPublicKey[];
```

where `sshPublicKey` is defined to be [something like this](https://typespec.io/playground/?c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtxhVjb3JlIjsKCkBkaXNjcmltaW5hdGVkKCN7ZW52ZWxvcGU6ICJub25lIn0pCnVuaW9uIHNzaFB1YmxpY0tleSB7CiAgaW5saW5lOs0ZScUUOwogIHJlc291cmNlSUTOIlLJGDsKfQoKbW9kZWzTQ8VicMgUOiBzdHJpbmfXM8pUxTfMe0HkAO4uQ29yZS5hcm3JKWRlbnRpZmllcjsKfQ%3D%3D&tspconfig=ZW1pdDoKICAtICJAdHlwZXNwZWMvb3BlbmFwaTMiCg%3D%3D&vs=%7B%7D).

(Realizing now that model names should be capitalized, but you get the point.)

tmstff: the JSON would then look like
```
"nodeSshPublicKeys": [
    {
        "kind": "inline",
        "publicKey": "ssh-rsa ..."
    },
    {
        "kind": "resourceID",
        "resourceID": "/subscriptions/..."
    }
]
```
see also [slack discussion](https://redhat-internal.slack.com/archives/C07SQ27Q3QR/p1787074930624059)

test-integration/frontend/cross_version_roundtrip_test.go
```diff
@@ -423,8 +423,75 @@ func clusterCreatePayload(clusterName, apiVersion string) []byte {
...
+    }
+  },
+  "tags": {
+    "env": "test"
+  },
+  "type": "Microsoft.RedHatOpenShift/hcpOpenShiftClusters"
+}`, clusterName, subscriptionID, subscriptionID, subscriptionID))
+
+	case v20261001:
+		// v20261001 payload — identical copy of v20260901 that does not (yet) carry nodeSshPublicKeys
```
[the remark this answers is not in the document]
tmstff: just force-pushed a new version

internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
```diff
@@ -544,9 +546,9 @@ func (c *HcpOpenShiftCluster) ConvertToInternal(existing *coreapi.HCPOpenShiftCl
 }
 
 // preserveUnknownClusterFields copies customer-facing fields from existing that
-// this API version doesn't know about. Currently empty — no cross-version
-// customer fields exist yet between v20240610preview and v20260630preview.
-func preserveUnknownClusterFields(from, to *coreapi.HCPOpenShiftCluster) {
+// this API version doesn't know about.
+func preserveUnknownClusterFields(_, _ *coreapi.HCPOpenShiftCluster) {
+	// All fields introduced up to v2026_06_30_preview are known to this version.
```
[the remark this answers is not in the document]
tmstff: just force-pushed a new version

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -137,6 +137,21 @@ model HcpOpenShiftClusterProperties {
...
+   * WARNING: Setting or changing this field currently causes the recreation of all worker nodes.
+   * This behavior is subject to change in a future OpenShift release.
+   *
+   * SSH public key format: <algorithm> <base64-encoded-key> [comment]
+   * Key constraints: algorithm + space + base64 blob required; comment optional; no newlines within key.
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  @added(Versions.v2026_09_01_preview)
+  @removed(Versions.v2026_10_01_preview)
+  nodeSshPublicKeys?: SshPublicKey[];
```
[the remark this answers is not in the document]
tmstff: will be addressed later on, when moving the field to the actual target version (currently just temporary target version)

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
tmstff: done

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
tmstff: done

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -137,6 +137,22 @@ model HcpOpenShiftClusterProperties {
...
+   * WARNING: Setting or changing this field currently causes the recreation of all worker nodes.
+   * This behavior is subject to change in a future OpenShift release.
+   *
+   * SSH public key format: <algorithm> <base64-encoded-key> [comment]
+   * Key constraints: algorithm + space + base64 blob required; comment optional; no newlines within key.
+   */
+  @clientName("nodeSSHPublicKeys")
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  @added(Versions.v2026_09_01_preview)
+  @removed(Versions.v2026_10_01_preview)
```
mbarnes, line 153: Versioning decorators look wrong.  This is claiming the field was added in **2026-09-01-preview** (it wasn't) and is removed in the next version.  I think you just want:
```suggestion
  @added(Versions.v2026_10_01_preview)
```

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -155,6 +171,37 @@ model HcpOpenShiftClusterProperties {
...
 }
 
+/** An SSH public key used for secure access to worker nodes.
+ *
+ * This is a discriminated type: the `kind` property selects the form in which
+ * the key is provided. Currently only the `inline` form is supported.
+ */
+@clientName("SSHPublicKey")
+@added(Versions.v2026_09_01_preview)
+@removed(Versions.v2026_10_01_preview)
```
mbarnes, line 181: Fix versioning decorators.  See above.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -155,6 +171,37 @@ model HcpOpenShiftClusterProperties {
...
+@discriminated(#{ envelope: "none" })
+union SshPublicKey {
+  /** The SSH public key material is provided directly in the request. */
+  inline: InlineSshPublicKey,
+}
+
+/** An SSH public key provided inline. */
+@clientName("InlineSSHPublicKey")
+@added(Versions.v2026_09_01_preview)
+@removed(Versions.v2026_10_01_preview)
```
mbarnes, line 191: Fix versioning decorators.  See above.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -155,6 +171,37 @@ model HcpOpenShiftClusterProperties {
...
+model InlineSshPublicKey {
```
mbarnes, line 192: Add a versioning decorator to the new model.
```suggestion
@added(Versions.v2026_10_01_preview)
model InlineSshPublicKey {
```

docs/cosmos-data-flow.md
```diff
@@ -1538,6 +1538,16 @@ Each entry links to every actor that writes the field.
...
+### `HCPOpenShiftCluster.CustomerProperties.NodeSshPublicKeys`
+
+| Actor | When |
+|-------|------|
+| [Frontend: PUT Cluster (Create)](#put-cluster-create) | Sets from request body (only API version `2026-09-01-preview`; removed in `2026-10-01-preview`) |
+| [Frontend: PUT Cluster (Update)](#put-cluster-update) | Replaces from request body; older API versions preserve existing value via `preserveUnknownClusterFields` |
+| [Frontend: PATCH Cluster (Update)](#patch-cluster-update) | Merged from PATCH body; older API versions preserve existing value via `preserveUnknownClusterFields` |
+
+Single writer (frontend only). Stored in Cosmos but not dispatched to Cluster Service — no backend controller forwards this field. Exposed only in API version `2026-09-01-preview` (removed in `2026-10-01-preview`). Each entry is a discriminated union (`kind: inline` with a `key` field).
+
```
mbarnes, line 1550: I think this is AI generated?  Maybe sure this is updated after the above corrections.

internal/azureapi/v20240610preview/hcpopenshiftclusters_methods.go
```diff
@@ -440,6 +440,8 @@ func preserveUnknownClusterFields(from, to *coreapi.HCPOpenShiftCluster) {
 	to.CustomerProperties.Platform.VnetIntegrationSubnetID = from.CustomerProperties.Platform.VnetIntegrationSubnetID
 	// Ingress was added in v2026_06_30_preview.
 	to.CustomerProperties.Ingress = from.CustomerProperties.Ingress
+	// NodeSshPublicKeys was added in v2026_09_01_preview.
```
mbarnes, line 443: Correct the comment.

internal/azureapi/v20251223preview/hcpopenshiftclusters_methods.go
```diff
@@ -494,6 +494,8 @@ func preserveUnknownClusterFields(from, to *coreapi.HCPOpenShiftCluster) {
 	to.CustomerProperties.Ingress = from.CustomerProperties.Ingress
 	// CryptoRestrictions was added in v2026_06_30_preview
 	to.CustomerProperties.CryptoRestrictions = from.CustomerProperties.CryptoRestrictions
+	// NodeSshPublicKeys was added in v2026_09_01_preview.
```
mbarnes, line 497: Correct the comment.

mbarnes (changes requested):
Code changes generally look ok but this is being added to the wrong API version.

2026-09-01-preview is already released.  Corrections needed in the TypeSpec versioning decorators.  Also back out the generated changes in `/internal/azureapi/v20260901preview/` and regenerate for `v20261001preview`.

PR #6160, opened 2026-07-20
fix: skip queue.Add when a key is already backing off after sync errors

patilsuraj767:
## Summary

Routine enqueue paths (informer resyncs, relist replays, and periodic tickers) call `queue.Add`, which bypasses an item already waiting on `AddRateLimited` backoff. After a sync failure, that can pull a key back into the queue immediately and hotloop retries instead of honoring the rate limiter.

This change adds a `NumRequeues(key) > 0` guard before those non-urgent `Add` calls across backend, fleet, kube-applier, mgmt-agent, sessiongate, and internal union informer controllers. Real changes (`changed=true`, `AddFunc`, `DeleteFunc`) still enqueue immediately; only resync/ticker paths are skipped while a key is backing off.

Below sample code shows that queue.Add(key) bypasses the rate-limiting behavior of AddRateLimited(). The queue.Add() call does not check whether the item is already waiting in the rate limiter. https://gist.github.com/patilsuraj767/b28df1e8d042ce94300d6032066223e3

### DIFF
--- a/backend/pkg/controllers/billingcontrollers/orphaned_billing_cleanup.go
+++ b/backend/pkg/controllers/billingcontrollers/orphaned_billing_cleanup.go
@@ -152,7 +152,13 @@ func (c *orphanedBillingCleanup) Run(ctx context.Context, threadiness int) {
 	}
 
 	// TODO before switching to a regular informer, build a basic LRU "don't fire unless your cooldown is over"
-	go wait.JitterUntilWithContext(ctx, func(ctx context.Context) { c.queue.Add("default") }, 60*time.Minute, 0.1, true)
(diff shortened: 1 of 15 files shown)


davidffrench:
Hi @patilsuraj767 — this PR has had no activity for 31 days. Is it still valid and something you plan to continue, or should it be closed? Please let me know. Thanks!

patilsuraj767:
> Hi Suraj Patil (@patilsuraj767) — this PR has had no activity for 31 days. Is it still valid and something you plan to continue, or should it be closed? Please let me know. Thanks!

Not a valid PR. Closing it.

PR #6163, opened 2026-07-20
test: skip 5.0 candidate install until HCCO ingress fix lands

zgalor:
## Summary

- Skip the 5.0 candidate install e2e test until a 5.0 EC build includes the HCCO ingress LB scope fix ([OCPBUGS-98571](https://issues.redhat.com/browse/OCPBUGS-98571))
- The fix ([openshift/hypershift#8992](https://github.com/openshift/hypershift/pull/8992)) merged to HyperShift main on 2026-07-14, but 5.0.0-ec.4 (built 2026-07-03) does not include it
- When CS sets `topology=PublicAndPrivate` on Swift clusters ([ARO-26913](https://redhat.atlassian.net/browse/ARO-26913)), the HCCO creates an IngressController with `InternalLoadBalancer` scope, causing `*.apps` routes to point to an unreachable internal LB IP (10.0.0.5) instead of the shared ingress public IP
- 4.22 is covered by the CPO override in [openshift/hypershift#9010](https://github.com/openshift/hypershift/pull/9010)
- Uses time-bomb pattern: skips before 2026-08-10, fails after — ensuring the skip doesn't persist silently once the fix should be available (5.0.0-ec.5 expected ~July 27)

## Test plan

- [x] Verify the skip follows existing time-bomb patterns (`nodepool_ephemeral_osdisk.go`, `cluster_create_private_ingress.go`)
- [x] e2e-parallel CI passes with the 5.0 test skipped

### DIFF
--- a/test/e2e/complete_cluster_create_multiversion.go
+++ b/test/e2e/complete_cluster_create_multiversion.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


zgalor:
@miquelsi @patriksuba, this is required to unblock the merging of private KAS support into CS. PTAL

zgalor:
@bennerv PTAL as well

PR #6167, opened 2026-07-20
fix(e2e): treat all-zone-restricted SKUs as unavailable in region

roivaz:
## Why

The E2E VM SKU selector (`SelectVMSize` / `skuRestrictedInLocation` in `test/util/framework/vm_sku_selection.go`) only filtered out **Location-type** SKU restrictions. It ignored the case where a SKU advertises availability zones in a region but **Zone-type** restrictions remove *all* of them.

This is not theoretical: Azure commonly expresses a full subscription/region ban as a **Zone-type** restriction listing every zone (reason `NotAvailableForSubscription`) rather than a Location-type restriction. Verified against the STG E2E subscription in `uksouth`: of 904 `virtualMachines` SKUs there were **0** Location-type restrictions but **5** SKUs whose every advertised zone was Zone-restricted (e.g. `Standard_D15_v2` — zones `[1,2,3]` all `NotAvailableForSubscription`).

Because selectors like `default-worker` do not set `RequireZones`, such a SKU passed filtering, got selected, and then failed at deploy time with `SkuNotAvailable`.

## What

- `skuRestrictedInLocation()` now also returns `true` when a SKU advertises zones in the location but has zero available zones after applying Zone-type restrictions.
- Guarded on `len(advertisedZones) > 0` so genuinely **non-zonal** SKUs are not misclassified as restricted.
- Extracted an `advertisedZones()` helper shared with `availableZones()`. The subset relationship is handled naturally by subtraction, so a restriction listing zones beyond those advertised (Azure sometimes does this, e.g. `NV12s_v3` advertises `[2,3]` but the restriction lists `[1,2,3]`) still empties the set.
- Added unit tests: 3 new `TestSelectVMSize` table cases + a new `TestSkuRestrictedInLocation` (none / location / all-zones / partial / over-listing / non-zonal / other-location).

https://redhat.atlassian.net/browse/AROSLSRE-1390

### DIFF
--- a/test/util/framework/per_test_framework.go
+++ b/test/util/framework/per_test_framework.go
@@ -1161,7 +1161,8 @@ func (tc *perItOrDescribeTestContext) AvailableZones(ctx context.Context, vmSize
 		if skuRestrictedInLocation(sku, location) {
 			return nil, nil
 		}
-		return availableZones(sku, location), nil
+		_, available := zonesInLocation(sku, location)
+		return available, nil
 	}
 	return nil, fmt.Errorf("VM size %q not found in Resource SKUs for location %q", vmSize, location)
 }
--- a/test/util/framework/vm_sku_selection.go
+++ b/test/util/framework/vm_sku_selection.go
@@ -25,6 +25,8 @@ import (
 
 	"github.com/onsi/ginkgo/v2"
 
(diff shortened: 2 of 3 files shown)


raelga (approved):
/lgtm
/approve

PR #6168, opened 2026-07-20
feat: add CosmosDB read-only query endpoint to the admin API

bennerv:
## Summary
- Adds `POST /admin/v1/cosmos/query` endpoint to the admin API for direct CosmosDB read access in production environments
- Validates queries against mutating keywords (UPDATE, DELETE, INSERT, CREATE, etc.) with string-literal-aware word-boundary matching
- Accepts container name, SQL query, optional partition key, and max items limit; returns raw JSON documents
- MISE-gated via the existing Istio AuthorizationPolicy — no additional config needed

## Test plan
- [x] Unit tests for keyword validation (35 cases covering edge cases: string literals, field name substrings, case insensitivity)
- [x] E2E tests verified locally against dev environment:
  - Valid SELECT query returns results from Resources container
  - Mutating keywords are rejected with 400
  - Partition key scoping works
  - Empty result sets return correctly
- [ ] CI passes

### DIFF
--- a/admin/server/handlers/cosmosquery/query.go
+++ b/admin/server/handlers/cosmosquery/query.go
@@ -1,0 +1,231 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cosmosquery
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"regexp"
+	"strings"
+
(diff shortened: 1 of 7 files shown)


bennerv:
This PR was mostly AI-authored. I'll take it out of draft once I've fully reviewed it.

admin/server/handlers/cosmosquery/query.go
```diff
@@ -0,0 +1,231 @@
...
+const (
+	defaultMaxItems = 100
+	maxMaxItems     = 1000
+	maxRequestBody  = 1 * 1024 * 1024 // 1MB
+)
+
+var mutatingKeywordPattern *regexp.Regexp
+
+func init() {
+	keywords := []string{
```
bennerv, line 41: todo: figure out if the query client we use actually allows mutating records.  if it doesn't no need to filter out key words here.

admin/server/handlers/cosmosquery/query.go
```diff
@@ -0,0 +1,231 @@
...
+	if results == nil {
+		results = []json.RawMessage{}
+	}
+
+	resp := QueryResponse{
+		Results: results,
+		Count:   len(results),
+	}
+
+	_, err = arm.WriteJSONResponse(w, http.StatusOK, resp)
```
bennerv, line 190: todo: we should stream the result instead of limiting max items so we can see the full response from the client side?

bennerv:
/close

will be implemented as part of: https://redhat.atlassian.net/browse/ARO-29536

PR #6170, opened 2026-07-20
mark a create operation timedout even if cluster-service and cosmos disagree on state

deads2k:
Turns out that in the practical sense, when we had "operation didn't finish" it was because `backend` detected that HCP had not reported that it reached a level.  Cluster-service just assumed the cluster was ready. And so there was an error before we reached the code that timed out and failed.  This adds a test for it and does the checks in the other order.


stevekuznetsov (approved):
/lgtm
/approve

PR #6189, opened 2026-07-21
docs: add Maestro agent CloudEvents desync scenario and scale-down strategy

avollmer-redhat:
## Why

During [AROSLSRE-1520](https://redhat.atlassian.net/browse/AROSLSRE-1520) (50 HCP clusters stuck deleting in brazilsouth after a scale test), we discovered a new Maestro failure mode that isn't covered by the existing runbook scenarios. The Maestro Agent continued recreating ManifestWorks even though the ResourceBundles were already soft-deleted on the server side — a CloudEvents sync desync caused by [ARO-28432](https://redhat.atlassian.net/browse/ARO-28432). Resolution required scaling the agent down (not just restarting) to eliminate a race condition during cleanup.

We also hit a significant debugging delay because the `deleted_at` field lives at the root level of the Maestro ResourceBundle JSON, not inside `.metadata` — existing `jq` queries were checking the wrong path and showing bundles as "clean" when they weren't.

### `docs/ops/cleanup-stuck-cluster-deletion.md`
- **Scenario 7**: Maestro Agent ignores soft-deleted ResourceBundles (CloudEvents desync) — distinct from Scenario 2 (active bundles) and Scenario 3 (orphaned namespace bundles), with criteria to distinguish between them
- **Strategy 3**: Scale down Maestro Agent, clean ManifestWorks/AppliedManifestWorks, scale back up — safer than restart because it eliminates the race condition where the agent recreates resources during cleanup
- **GOTCHA note**: `deleted_at` is at the JSON root level, not `.metadata.deleted_at` — incorrect queries caused hours of debugging delay
- **Port-forward quick reference**: `kubectl port-forward -n maestro svc/maestro 8002:8000` is more reliable than `oc exec` for API inspection
- Fixed the existing Step 6 query that incorrectly used `.metadata.deleted_at`
- Renumbered old Strategy 3 → Strategy 4; updated all cross-references

### `docs/ops/fix-maestro-stale-resource-bundle.md`
- Added "Related Scenarios and Gotchas" section with the `deleted_at` field location warning
- Cross-reference to Scenario 7 in the stuck deletion runbook

## References
- [AROSLSRE-1520](https://redhat.atlassian.net/browse/AROSLSRE-1520) (primary incident — Closed/Done)
- [ARO-28432](https://redhat.atlassian.net/browse/ARO-28432) (root cause bug: Maestro orphaned DELETE)
- [AROSLSRE-1537](https://redhat.atlassian.net/browse/AROSLSRE-1537) (secondary: CS resource contention during mass deletion)

### DIFF
--- a/docs/ops/cleanup-stuck-cluster-deletion.md
+++ b/docs/ops/cleanup-stuck-cluster-deletion.md
@@ -231,10 +231,16 @@ kubectl exec -n clusters-service deployment/clusters-service -- \
 
 ```bash
 # Search for resource bundles by cluster ID (search in manifest content)
+#
+# TIP: Use port-forward instead of oc exec for more reliable Maestro API
+# inspection -- oc exec can truncate responses or drop fields:
+#   kubectl port-forward -n maestro svc/maestro 8002:8000
+#   curl -s 'http://localhost:8002/api/maestro/v1/resource-bundles?size=2900' | ...
+#
 kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
   "curl -s 'http://localhost:8000/api/maestro/v1/resource-bundles?size=2900'" | \
   jq --arg cid "${CLUSTER_ID}" '[.items[] | select(.manifests | tostring | test($cid)) |
-  {id, name, created_at, consumer_name, deleted_at: .metadata.deleted_at,
+  {id, name, created_at, consumer_name, deleted_at,
    labels: .metadata.labels}]'
 ```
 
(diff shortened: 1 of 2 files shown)


docs/ops/cleanup-stuck-cluster-deletion.md
```diff
@@ -279,11 +285,66 @@ kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
...
+Connect to the management cluster:
+
+```bash
+hcpctl mc breakglass <mc-name>
+export KUBECONFIG=<path-from-output>
+```
+
+```bash
+# 1. Scale down the Maestro Agent to stop ManifestWork recreation
+kubectl scale deployment -n maestro maestro-agent --replicas=0
```
weherdh, line 328: The prerequisite should be: verify that the Maestro Agent scale-down will not affect ANY customer operations — not just existing clusters, but also any in-flight or incoming create/update/delete requests targeting this management cluster. While the Agent is down, any ARM operation that requires a ManifestWork to be applied on this MC will fail or time out. The safest precondition is confirming the MC has no customer-facing traffic (no customer clusters AND the region is not actively serving new requests). If the region is live with customers, this procedure requires a maintenance window.

avollmer-redhat: Good call — added a prerequisite warning block in 50ae3fbe1 covering the full blast radius (all ManifestWork delivery blocked, in-flight ARM ops fail/timeout, need to confirm no customer traffic or take a maintenance window). Also added a fallback reference to Strategy 4 (manual finalizer removal) for cases where a maintenance window isn't available, since it patches individual resources without stopping the agent.

PR #6194, opened 2026-07-21
feat: add controller to identify and clean up cosmos records without resourceID

deads2k:
Old records in Cosmos may lack a resourceID field. This adds a MissingResourceID controller that runs every hour, lists all such records via a new ListMissingResourceID method on ResourcesDBClient, dumps each to the log, and deletes items whose resourceType contains "controller" or "operation".


stevekuznetsov (approved):
/lgtm
/approve

PR #6196, opened 2026-07-21
[wip] feat: add SystemAdminCredentialRequest type, store credentials in Cos…

deads2k:
…mos on issuance, add 48h GC

Brings in SystemAdminCredentialRequest API type and CRUD from cs-198-admincreds, plus BuildKubeconfig in internal/systemadmincredential. When the break-glass credential is issued by cluster service, operation_request_credential now creates a SystemAdminCredentialRequest in Cosmos (name = operationID) and updates the operation's InternalID to the ARM resource ID. The frontend reads the credential back from Cosmos via assembleAdminCredentialFromCosmos.

A new CredentialGC controller runs every hour, iterates all clusters per subscription, and deletes any SystemAdminCredentialRequest older than 48 hours.

Alternative to a larger port. This would try to simply copy admin credentials to the API we eventually want and use that for fast retrieval.  It is a stepping stone to https://github.com/Azure/ARO-HCP/pull/6124 which is a stepping stone to a new retrieval API that requires client-side integration to build the credential itself.


frontend/pkg/frontend/frontend.go
```diff
@@ -1021,13 +1022,12 @@ func (f *Frontend) OperationResult(writer http.ResponseWriter, request *http.Req
 	var responseBody []byte
 
 	switch {
-	case operation.InternalID.Kind() == cmv1.BreakGlassCredentialKind:
-		csBreakGlassCredential, err := f.clusterServiceClient.GetBreakGlassCredential(ctx, operation.InternalID)
+	case operation.InternalID.Kind() == api.SystemAdminCredentialRequestKind:
+		adminCred, err := f.assembleAdminCredentialFromCosmos(ctx, operation)
```
miguelsorianod, line 1034: We need to figure out how to rollout this without impact on pre-existing resources. An issue we have now is that for older operations this would not return any result as the creation of the resource exists for the active operations which would have ended at that point.

Independently on that, the backend code (whatever we end up implementing) should be rolled out before the frontend code as it will depend on that

backend/pkg/controllers/operationcontrollers/operation_request_credential.go
```diff
@@ -144,3 +168,109 @@ func (opsync *operationRequestCredential) SynchronizeOperation(ctx context.Conte
...
+	_, err = credCRUD.Create(ctx, cred, nil)
+	if err != nil {
+		if database.IsConflictError(err) {
+			return credResourceID, nil
+		}
+		return nil, utils.TrackError(fmt.Errorf("failed to create SystemAdminCredentialRequest: %w", err))
+	}
+
+	keyVaultSecretClient, err := opsync.keyVaultSecretClientFactory.KeyVaultSecretClient(
+		managementCluster.Status.HostedClustersSecretsKeyVaultManagedIdentityClientID,
```
stevekuznetsov, line 250: are you allowed to do this? I was expecting to see the backend MSI given Azure RBAC permissions over those keyvaults. if there's no link between the backend pod/MSI and this other client ID, I don't think you'll be allowed to mint a token

backend/cmd/root.go
```diff
@@ -330,12 +330,12 @@ func (f *BackendRootCmdFlags) ToBackendOptions(ctx context.Context, cmd *cobra.C
 		return nil, utils.TrackError(fmt.Errorf("failed to create FPA client builder: %w", err))
 	}
 
-	backendIdentityAzureClients, err := app.NewBackendIdentityAzureClients(ctx, azureConfig)
+	backendIdentityClientBuilder, err := app.NewBackendIdentityClientBuilder(ctx, azureConfig)
```
miguelsorianod, line 333: That seems to have changed the behavior of existing backend identity clients:
Before this PR we had some that were instantiated at startup time (like the blob one, because we always use that one instance), and now we have some that are instantiated on demand (as the keyvaults are different per mgmt cluster). This has changed  and we are not instantiating at startup time the blob and the role one. We should have both.

backend/pkg/app/backend_identity_wiring.go
```diff
@@ -45,33 +43,22 @@ func NewBackendIdentityAzureClients(ctx context.Context, azureConfig *azureconfi
...
+	return azureclient.NewBackendIdentityClientBuilder(
 		defaultAzureCredential,
-		&azblob.ClientOptions{
-			ClientOptions: *azureConfig.CloudEnvironment.AZCoreClientOptions(),
-		},
-	)
-	if err != nil {
-		return nil, utils.TrackError(fmt.Errorf("failed to create dataplane identities OIDC configuration blob storage client: %w", err))
-	}
+		azureConfig.CloudEnvironment.AZCoreClientOptions(),
```
miguelsorianod, line 48: It is redundant to pass both AZCoreClientOptions and ARMClientOptions.
ARMClientOptions should be enough. Within it it has an attribute policy.ClientOptions which is essentially AZCoreClientOptions (they are type alias)

backend/pkg/azure/client/keyvault_client.go
```diff
@@ -0,0 +1,31 @@
...
+
+package client
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azsecrets"
+)
+
+type KeyVaultSecretClient interface {
```
miguelsorianod, line 23: nitpick: Secret -> Secrets. The https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azsecrets doc refers to it as "Azure Key Vault Secrets client module for Go" (plural)

backend/cmd/root.go
```diff
@@ -437,13 +437,14 @@ func (f *BackendRootCmdFlags) ToBackendOptions(ctx context.Context, cmd *cobra.C
...
 		FPAClientBuilder:                   fpaClientBuilder,
-		BackendIdentityAzureClients:        backendIdentityAzureClients,
+		BackendIdentityClientBuilder:       backendIdentityClientBuilder,
 		BackendIdentityAzureCachedReaders:  backendIdentityAzureCachedReaders,
 		ExitOnPanic:                        f.ExitOnPanic,
 		FPAMIDataplaneClientBuilder:        fpaMIDataplaneClientBuilder,
 		SMIClientBuilder:                   smiClientBuilder,
 		CheckAccessV2ClientBuilder:         checkAccessV2ClientBuilder,
 		ClusterScopedIdentitiesConfig:      clusterScopedIdentitiesConfig,
+		KeyVaultSecretClientFactory:        azureclient.NewManagedIdentityKeyVaultSecretClientFactory(azCoreClientOptions),
```
miguelsorianod, line 447: Why do we need KeyVaultSecretClientFactory in backend? We should be able to use BackendIdentityClientBuilder which is what we want when we want to use the backend identity. It uses the default azure credential chain, which appropriately selects the backend identity

frontend/pkg/frontend/keyvault_client.go
```diff
@@ -0,0 +1,62 @@
...
+var _ KeyVaultSecretClientFactory = (*managedIdentityKeyVaultSecretClientFactory)(nil)
+
+func NewManagedIdentityKeyVaultSecretClientFactory(clientOptions azcore.ClientOptions) KeyVaultSecretClientFactory {
+	return &managedIdentityKeyVaultSecretClientFactory{
+		clientOptions: clientOptions,
+	}
+}
+
+func (f *managedIdentityKeyVaultSecretClientFactory) KeyVaultSecretClient(managedIdentityClientID string, vaultURL string) (KeyVaultSecretReader, error) {
+	cred, err := azidentity.NewManagedIdentityCredential(&azidentity.ManagedIdentityCredentialOptions{
```
miguelsorianod, line 48: If the intention of this factory is to use the pod's identity, in this case the frontend identity, we should use the default azure credential https://learn.microsoft.com/en-us/azure/developer/go/sdk/authentication/credential-chains#defaultazurecredential-overview. We want to end up leveraging the pod's workload identity. (It also has the nice bonus where you can technically run it using other authentication methods transparently in other environments like local dev, ...)

miguelsorianod: At this point, I would just have the concept of a FrontendIdentityClientsBuilder similar to backend. That builder would be instantiated containing the defaultazurecredentials similar to what it's being done in backend

frontend/pkg/frontend/frontend.go
```diff
@@ -1073,6 +1081,104 @@ func (f *Frontend) OperationResult(writer http.ResponseWriter, request *http.Req
...
+
+	keyVaultSecretClient, err := f.keyVaultSecretClientFactory.KeyVaultSecretClient(
+		managementCluster.Status.HostedClustersSecretsKeyVaultManagedIdentityClientID,
+		managementCluster.Status.HostedClustersSecretsKeyVaultURL,
+	)
+	if err != nil {
+		return "", utils.TrackError(fmt.Errorf("failed to create Key Vault secret client: %w", err))
+	}
+
+	resp, err := keyVaultSecretClient.GetSecret(ctx, secretName, "", nil)
```
miguelsorianod, line 1170: Does the Frontend identity have enough permissions to perform the desired actions in the desired key vault? If we are going to go to that route we need to make sure we do, in all aro-hcp environments

miguelsorianod, line 1170: To be aware of, this means that now the frontend for this endpoint on admincredentials operationresult now has more DB interactions (to figure out the vault url) as well as Azure API interaction for keyvault interaction, which adds latency

backend/pkg/controllers/operationcontrollers/operation_request_credential.go
miguelsorianod: Here we shouldn't be using keyVaultSecretClientFactory. We should be using BackendIdentityClientsBuilder and then get a keyvault secrets client from it. That will ensure that the Backend's identity is what's being used, which is what we want.

backend/pkg/controllers/operationcontrollers/operation_request_credential.go
```diff
@@ -144,3 +168,109 @@ func (opsync *operationRequestCredential) SynchronizeOperation(ctx context.Conte
...
+	}, nil)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to store kubeconfig in Key Vault: %w", err))
+	}
+
+	return credResourceID, nil
+}
+
+func keyVaultSecretNameForCredential(credResourceID *azcorearm.ResourceID) string {
+	hash := sha256.Sum256([]byte(strings.ToLower(credResourceID.String())))
```
miguelsorianod, line 274: Is there some reason we use the sha of the credResourceID? hashing works but it will make discoverability harder just from the name. I see that we are setting tags in the secret which would help but that's a bit more jumps to find it out.
Have we considered having a fixed prefix that signals that it's an admin credential + the resource name which is the operation id which is a uuid with fixed length?
something like system-admin-credential-kubeconfig-<operation_uuid> (35+36 = 71 chars)

As a note, the maximum number of characters for a key vault secret name is 127 characters (with alphanumeric characters and hyphens) https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftkeyvault

backend/pkg/controllers/operationcontrollers/operation_request_credential.go
```diff
@@ -144,3 +168,109 @@ func (opsync *operationRequestCredential) SynchronizeOperation(ctx context.Conte
...
+		),
+	}, nil)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to store kubeconfig in Key Vault: %w", err))
+	}
+
+	return credResourceID, nil
+}
+
+func keyVaultSecretNameForCredential(credResourceID *azcorearm.ResourceID) string {
```
miguelsorianod, line 273: Let's add a fixed length prefix to indicate just from the name that it's about the kubeconfig of a cluster's admin-credentials. For example:
`system-admin-credential-kubeconfig-`

This helps identify quickly just from the name that it's about that concept. That keyvault also contains other secrets that are not related to system admin credentials kubeconfig so this also allows to differentiate them from those others just from the name

backend/pkg/controllers/operationcontrollers/operation_request_credential.go
```diff
@@ -144,3 +168,109 @@ func (opsync *operationRequestCredential) SynchronizeOperation(ctx context.Conte
...
+
+	keyVaultSecretClient, err := opsync.keyVaultSecretClientFactory.KeyVaultSecretClient(
+		managementCluster.Status.HostedClustersSecretsKeyVaultManagedIdentityClientID,
+		managementCluster.Status.HostedClustersSecretsKeyVaultURL,
+	)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to create Key Vault secret client: %w", err))
+	}
+
+	_, err = keyVaultSecretClient.SetSecret(ctx, keyVaultSecretName, azsecrets.SetSecretParameters{
```
miguelsorianod, line 257: In Azure some API interactions are asynchronous. Do we know if SetSecret guarantees that once the call is returned the secret is there? I am guessing that yes because the result is the secret response and when it's asynchronous it's usually some kind of poller that is returned as well as the method name is usually called Begin*.

Raising awareness about the different kind of behaviors as I think it's useful to be aware of it.

backend/pkg/controllers/operationcontrollers/operation_request_credential.go
```diff
@@ -144,3 +168,109 @@ func (opsync *operationRequestCredential) SynchronizeOperation(ctx context.Conte
...
+		operation.ExternalID.SubscriptionID,
+		operation.ExternalID.ResourceGroupName,
+		operation.ExternalID.Name,
+		credName,
+	)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to build credential request resource ID: %w", err))
+	}
+
+	serviceProviderCluster, err := opsync.resourcesDBClient.ServiceProviderClusters(
```
miguelsorianod, line 191: Would using SPC lister be more appropriate here? or are we concerned about caching and it being outdated?

internal/api/types_systemadmincredential.go
```diff
@@ -0,0 +1,92 @@
...
+
+// SystemAdminCredentialRequest represents a temporary admin credential request
+// tracked in Cosmos, replacing the cluster-service break-glass credential flow.
+//
+// Lifecycle is tracked via metav1.Conditions rather than an explicit Phase enum
+// so that individual aspects of the request (issuance, revocation, cleanup) can
+// progress independently and callers can reason about each concern separately.
+//
+// +k8s:deepcopy-gen=true
+type SystemAdminCredentialRequest struct {
```
miguelsorianod, line 29: What's the reasoning of containing the "System" and the "Request" words?

miguelsorianod: Would having the word "Cluster" as part of the name make sense? or do we think it's redundant because we will ever have cluster ones and it's implied?

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -363,6 +386,115 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
...
+	managementCluster, err := c.managementClusterLister.Get(ctx, stampIdentifier)
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to get ManagementCluster for stamp %q: %w", stampIdentifier, err))
+	}
+
+	if managementCluster.Status.HostedClustersSecretsKeyVaultURL == "" || managementCluster.Status.HostedClustersSecretsKeyVaultManagedIdentityClientID == "" {
+		return nil, nil
+	}
+
+	kvClient, err := c.keyVaultSecretClientFactory.KeyVaultSecretClient(
```
miguelsorianod, line 405: Similar to https://github.com/Azure/ARO-HCP/pull/6196/changes#r3633971629, here we shouldn't be using keyVaultSecretClientFactory. We should be using BackendIdentityClientsBuilder

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -363,6 +386,115 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
...
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("failed to create Key Vault secret client: %w", err))
+	}
+	return kvClient, nil
+}
+
+func (c *clusterChildResourcesCleanupController) ensureKeyVaultSecretsDeleted(ctx context.Context, clusterResourceID *azcorearm.ResourceID) error {
+	logger := utils.LoggerFromContext(ctx)
+
+	spc, err := c.resourcesDBClient.ServiceProviderClusters(clusterResourceID.SubscriptionID, clusterResourceID.ResourceGroupName, clusterResourceID.Name).Get(ctx, api.ServiceProviderClusterResourceName)
```
miguelsorianod, line 418: Is usage of direct Cosmos intentional for SPC, or could we leverage lister? or are we worried about caching content being outdated here?

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -363,6 +386,115 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
...
+
+	kvClient, err := c.getKeyVaultSecretClient(ctx, spc)
+	if err != nil {
+		return err
+	}
+	if kvClient == nil {
+		return nil
+	}
+
+	pager := kvClient.NewListSecretPropertiesPager(nil)
```
miguelsorianod, line 434: Instead of listing, which could have a non small amount of elements, would retrieving the admincredentials from cosmos, calculating the name for each one of them and then doing a delete would be an alternative?

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -363,6 +386,115 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
...
+				azureclient.KeyVaultBinarySourceBackend,
+				clusterResourceID.SubscriptionID,
+				clusterResourceID.ResourceGroupName,
+				clusterResourceID.Name,
+			) {
+				continue
+			}
+			secretName := secret.ID.Name()
+			logger.Info("deleting Key Vault secret for cluster", "secretName", secretName)
+			if _, err := kvClient.DeleteSecret(ctx, secretName, nil); err != nil {
```
miguelsorianod, line 454: Does DeleteSecret perform a soft or hard delete? I am wondering what are the consequences if it's a soft one

miguelsorianod: Also wondering about this because if I recall correctly trying to create a secret with the same name as one that is still soft deleted would fail with a conflict

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -363,6 +386,115 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
...
+				return utils.TrackError(fmt.Errorf("failed to delete Key Vault secret %q: %w", secretName, err))
+			}
+		}
+	}
+
+	return nil
+}
+
+func (c *clusterChildResourcesCleanupController) countKeyVaultSecretsForCluster(ctx context.Context, spc *api.ServiceProviderCluster, serviceProviderClusterResourceID *azcorearm.ResourceID) (int, error) {
```
miguelsorianod, line 463: nitpick: this doesn't count all the secrets in the keyvault, just a subset of some. We should be clear in the name of the method about this

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -280,6 +292,17 @@ func (c *clusterChildResourcesCleanupController) extraDeleteGateShouldDeleteServ
 		return true, nil
 	}
 
+	remaining, err := c.countKeyVaultSecretsForCluster(ctx, spc, serviceProviderClusterResourceID)
```
miguelsorianod, line 295: Is this logic needed? If ensureKeyVaultSecretsDeleted succeeds, doesn't that mean we've eliminated all of them?

backend/pkg/controllers/clusterdeletion/cluster_child_resources_cleanup_controller_test.go
```diff
@@ -26,18 +26,57 @@ import (
...
 	"github.com/Azure/ARO-HCP/internal/api/kubeapplier"
 	"github.com/Azure/ARO-HCP/internal/database"
+	dblistertesting "github.com/Azure/ARO-HCP/internal/database/listertesting"
 	"github.com/Azure/ARO-HCP/internal/databasetesting"
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
+type emptyKeyVaultSecretClient struct{}
+
+func (e *emptyKeyVaultSecretClient) GetSecret(_ context.Context, _ string, _ string, _ *azsecrets.GetSecretOptions) (azsecrets.GetSecretResponse, error) {
```
miguelsorianod, line 49: Have we evaluated using mockgen for this? It will also allow unit tests in general to set the desired inputs/outputs. We do that with other azure clients already like for example `RoleDefinitionsClient` (see the file for the mockgen marker)

backend/pkg/controllers/mismatchcontrollers/credential_gc.go
```diff
@@ -0,0 +1,199 @@
...
+	errs := []error{}
+
+	for _, cluster := range allClusters.Items(ctx) {
+		credCRUD := c.resourcesDBClient.SystemAdminCredentialRequests(
+			cluster.ID.SubscriptionID,
+			cluster.ID.ResourceGroupName,
+			cluster.ID.Name,
+		)
+
+		credIterator, err := credCRUD.List(ctx, nil)
```
miguelsorianod, line 88: Is there some reason we want to use the DB directly instead of a lister in this context?

frontend/cmd/cmd.go
```diff
@@ -255,10 +255,17 @@ func (opts *FrontendOpts) Run() error {
 		utils.TracerName,
 	)
```
miguelsorianod, line 257: To improve at some point: on frontend right now we hardcode the azure cloud type to public. Once we need to support more clouds that will fall short. In backend we already make that configurable (part of the azureconfig file that it receives). Here we will want to do something similar (at the least a cli flag with the value)

frontend/pkg/frontend/frontend.go
```diff
@@ -1073,6 +1081,104 @@ func (f *Frontend) OperationResult(writer http.ResponseWriter, request *http.Req
...
+	} else if len(cred.Status.Kubeconfig) > 0 {
+		kubeconfig = cred.Status.Kubeconfig
+	} else {
+		cluster, err := f.resourcesDBClient.HCPClusters(op.ExternalID.SubscriptionID, op.ExternalID.ResourceGroupName).Get(ctx, op.ExternalID.Name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get cluster: %w", err)
+		}
+		kubeconfigBytes, err := systemadmincredential.BuildKubeconfig(
+			cred.Status.SignedCertificate,
+			cred.Spec.PrivateKeyPEM,
```
miguelsorianod, line 1126: Even though we store kubeconfig in key vault, PrivateKeyPEM would still need to be available somehow

PR #6216, opened 2026-07-22
PoC: remove operationresult get breakglass credential by syncing information from CS into Cosmos

miguelsorianod:
This PR PoC showcases a potential solution to remove the CS Get BreakGlass credential call that is performed in Frontend during an operation result request for an operation that is an admin-credential operation.

The proposal does the following:
* Create a new Cosmos direct subresource of Cluster named .../adminCredentials/<adminCredentialName>. <adminCredentialName> is the breakglass credential CS ID. The cosmos resource also contains the corresponding operationid that triggered its creation, so it can be retrieved when having that information only
* In the backend controller that dispatches the creation of breakglass credential in CS, a corresponding  .../adminCredentials/<adminCredentialName> Cosmos document is created
* A new backend controller SyncClusterAdminCredentials is created. It is a cluster-watching controller. It iterates over the cosmos admin credentials of the cluster, retrieves the information on CS side and it syncs the reported state from CS to the corresponding admin credentials cosmos document. This includes the status, expirationtimestamp and kubeconfig. If CS side does not contain the breakglass credential it removes the entry from Cosmos. This covers the cases where CS deletes the breakglass credential after some time of it having expired and/or revoked
  * The controller also introduces an additional cooldown timer of 30 seconds. This is because each time the controller calls SyncOnce it interacts with the CS API for all admincredentials of the cluster. To be conservative about the load we put on CS we introduce this cooldown
  * We update the admin credential request operation controller to rely on the Cosmos adminCredentials document instead of CS. The admin credential revoke operation controller remains unchanged. TODO should we change the revoke one to depend on cosmos? if so, only on cosmos or both cosmos and cs (potential orphans on cs side?)
* There's no handling of potentially orphan breakglass credentials on CS side. We are fine with this because in aro-hcp all the CS breakglass credentials have an expiration timestamp set, with a maximum value of 24h, and CS takes care of deleting them some time after they expire
* The frontend logic of operation result request for an operation that is an admin-credential operation now retrieves the corresponding admin-credential document from Cosmos instead of interacting with CS
* It updates the datadumper to redact the kubeconfig attribute

If we were to want to rollout this approach at some point we would need to figure out a safe rollout approach. It will need to be at least in two phases as frontend can't be rolled out until the backend code is there. We should also consider impact on pre-existing resources/operations, they shouldn't be negatively impacted.


davidffrench:
Hi @miguelsorianod — this PR has had no activity for 31 days. Is it still valid and something you plan to continue, or should it be closed? Please let me know. Thanks!

miguelsorianod:
Closing. This was one of the design proposals discussed. The different design proposals were discussed, which ended up with the current RP design around systemadmincredentials.

PR #6217, opened 2026-07-22
test: add e2e test for private KAS (api.visibility: Private)

zgalor:
## Summary

Adds an e2e test that creates a cluster with `api.visibility: Private` and verifies:
- ARM GET returns `api.visibility=Private` and `ingress.type=Public` (independence check)
- KAS is reachable from a VM inside the VNet (`kubectl version` via internal LB)
- KAS is NOT reachable from outside the VNet
- Public ingress remains functional despite private KAS (console URL reachable from outside)

Also extracts shared e2e helpers from private KAS, private ingress, and authorized CIDRs tests:
- `framework.DeployTestVM()` — SSH key gen, VM size selection, bicep deploy with retry
- `framework.RunKubectlOnVM()` — kubeconfig write + kubectl execution on VM
- `framework.TestHTTPSConnectivity()` — unified `testAPIConnectivity` and `testIngressConnectivity`
- `framework.SampleAppManifests()` — returns sample app YAML for `kubectl apply -f -` on VMs

### Ingress verification approach

Instead of deploying a sample app via `kubectl apply` from the VM (which requires authenticated kubectl through the internal LB — unreliable due to breakglass token auth issues via Swift proxy), the test verifies public ingress by checking the OpenShift console URL. The console is deployed by default on every cluster and served through the default ingress router, so if it's reachable from outside the VNet, the default ingress is public.

### KAS reachability approach

KAS reachability from inside the VNet is verified using `kubectl version` (unauthenticated `/version` endpoint) from a VM via the internal LB. This proves the private KAS network path (VM → customer subnet → internal LB → Swift → KAS pods) is functional without depending on authenticated access through the internal LB.

## Test plan

- [x] `go build -tags E2Etests ./test/e2e/` compiles
- [x] `golangci-lint` passes
- [x] `make record-nonlocal-e2e` passes
- [x] Suite fixtures updated
- [x] Test passes locally against pers env (all 5 verifications succeed)
- [ ] CI e2e pass

Ref: [ARO-23333](https://redhat.atlassian.net/browse/ARO-23333)

### DIFF
--- a/test/e2e/cluster_create_private_kas.go
+++ b/test/e2e/cluster_create_private_kas.go
@@ -75,38 +75,9 @@ var _ = Describe("Customer", func() {
 			)
 			Expect(err).NotTo(HaveOccurred(), "failed to create customer resources for private KAS cluster")
 
-			By("generating SSH key pair and deploying test VM in customer VNet")
-			sshPublicKey, _, err := framework.GenerateSSHKeyPair()
-			Expect(err).NotTo(HaveOccurred(), "failed to generate SSH key pair for test VM")
-
-			vmName := fmt.Sprintf("%s-test-vm", customerClusterName)
-			vmSize, err := tc.SelectVMSize(ctx, framework.JumpboxVMSizeSelector())
-			Expect(err).NotTo(HaveOccurred(), "failed to resolve a jumpbox VM size for private KAS test")
-
-			var deployErr error
-			for attempt := 0; attempt < 3; attempt++ {
-				if attempt > 0 {
-					time.Sleep(20 * time.Second)
-				}
-				_, deployErr = tc.CreateBicepTemplateAndWait(ctx,
-					framework.WithTemplateFromFS(TestArtifactsFS, "test-artifacts/generated-test-artifacts/modules/test-vm.json"),
-					framework.WithDeploymentName("test-vm"),
-					framework.WithScope(framework.BicepDeploymentScopeResourceGroup),
-					framework.WithClusterResourceGroup(*resourceGroup.Name),
-					framework.WithParameters(map[string]any{
-						"vmName":       vmName,
-						"vnetName":     clusterParams.VnetName,
-						"subnetName":   clusterParams.SubnetName,
-						"sshPublicKey": sshPublicKey,
-						"vmSize":       vmSize,
-					}),
-					framework.WithTimeout(30*time.Minute),
-				)
-				if deployErr == nil || strings.Contains(deployErr.Error(), "SkuNotAvailable") {
-					break
-				}
-			}
-			Expect(deployErr).NotTo(HaveOccurred(), "failed to deploy test VM for private KAS verification")
+			By("deploying test VM in customer VNet")
+			vmName, _, err := tc.DeployTestVM(ctx, TestArtifactsFS, *resourceGroup.Name, customerClusterName, clusterParams.VnetName, clusterParams.SubnetName)
+			Expect(err).NotTo(HaveOccurred(), "failed to deploy test VM for private KAS verification")
 
 			By("creating the HCP cluster with private KAS via v20260630preview")
 			err = tc.CreateHCPClusterFromParam20260630(ctx,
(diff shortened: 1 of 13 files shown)


zgalor:
@nimrodshn I did some refactoring to extract the VM deployment and kubectl execution used in all three tests. Please take a look, specifically at the authorized CIDRs test

nimrodshn:
@zgalor  Changes for authorized cidr blocks look OK, TY :100:

zgalor:
@machi1990 @miquelsi PTAL

test/e2e/cluster_create_private_kas.go
machi1990, line 1: Can we somehow combine private kas and private ingress test cases? 
Or we absolute need them to be separate test cases?

machi1990: i.e we setup the test case in a way that it creates a private kas cluster with private ingress and run the two verifications

zgalor: I wanted to validate all 4 combinations of KAS/Ingress visibility since they should be fully independent,
And to make sure ingress visibility is no longer derived from KAS visibility (as it used to be).

Public KAS + Public Ingress (Covered by any other test)
Public KAS + Private Ingress (Covered in cluster_create_private_ingress.go)
Private KAS + Public Ingress (covered in this PR - cluster_create_private_kas.go)
Private KAS + Private Ingress (Not covered yet) - I plan to cover it in a follow-up PR

machi1990: Ah makes sense. Thank you for clarifying

zgalor:
@machi1990 I changed the test to make it more stable, not basing it on fragile VM-based kubectl operations (prone to conflicts and timeouts)

Would you please take another look?

@miquelsi @mbukatov I would also appreciate QE input

machi1990 (approved):
/lgtm

test/util/framework/vm_helper.go
```diff
@@ -32,8 +34,161 @@ import (
...
+		kubeconfigB64, kubectlArgs,
+	)
+	return RunVMCommand(ctx, tc, resourceGroup, vmName, cmd, pollTimeout)
+}
+
+// TestHTTPSConnectivity attempts an HTTPS connection to the given URL.
+// Returns nil if the connection succeeds, or an error if it fails (DNS, timeout,
+// connection refused, etc.). TLS validation is skipped — this tests network
+// reachability, not certificate validity. Redirects are not followed.
+func TestHTTPSConnectivity(ctx context.Context, url string, timeout time.Duration) error {
```
machi1990: np: Why is this in vm_helper.go?
it should be in a file that doesn't have to do with VMs

machi1990 (commented):
Approved but left a nitpick

test/util/framework/vm_helper.go
zgalor: moved to `connectivity_helper.go`

test/util/framework/vm_helper.go
```diff
@@ -32,8 +33,129 @@ import (
...
+		kubeconfigB64, kubectlArgs,
+	)
+	return RunVMCommand(ctx, tc, resourceGroup, vmName, cmd, pollTimeout)
+}
+
+// GetPrivateKASInternalIP finds the private IP address of the internal load
+// balancer created by HyperShift for a private KAS cluster. The internal LB
+// is in the cluster's managed resource group and has a frontend IP on the
+// customer's subnet. Returns the IP address or an error if not found.
+func GetPrivateKASInternalIP(ctx context.Context, tc interface {
```
machi1990: this can live in a separate helper file

zgalor: moved to `resource_group_helper.go`

test/e2e/cluster_create_private_kas.go
```diff
@@ -0,0 +1,202 @@
...
+
+	hcpsdk20260630preview "github.com/Azure/ARO-HCP/test/sdk/v20260630preview/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+)
+
+// This test creates a cluster with private KAS (api.visibility: Private) using
+// the v2026-06-30-preview API and verifies that:
+//   - The Kubernetes API server is only reachable from within the VNet
+//   - The default ingress remains public (independence of KAS and ingress visibility)
```
mbukatov, line 34: You will need to add a hack to the test code and skip this test case in higher environments until given date, in a similar way how done here: https://github.com/Azure/ARO-HCP/pull/6271

This is the reason the test case PR was reverted in [#6349](https://github.com/Azure/ARO-HCP/pull/6349). The reintroduction of the PR is tracked in https://redhat.atlassian.net/browse/AROSLSRE-1674

test/util/framework/connectivity_helper.go
```diff
@@ -0,0 +1,54 @@
...
+
+	req, err := http.NewRequestWithContext(timeoutCtx, "GET", url, nil)
+	if err != nil {
+		return err
+	}
+
+	client := &http.Client{
+		Timeout: timeout,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
mbukatov, line 40: Since you moved this from private test case function into a framework library, we can no longer have disabled TLS as a default.

We should either keep TLS check enabled, or move this option into a function argument, so that caller can explicitly decide if tls check should be ignored or not and then your test cases can set skipverify true to preserve the original behavior.

mbukatov (commented):
Initial review for the E2E test. Both comments should be addressed in the next PR reintroducing the test.

PR #6230, opened 2026-07-22
feat(kube-applier): redact Secret data before persisting ReadDesire status

deads2k:
ReadDesire targets that point at core/v1 Secrets now have their kubeContent redacted before it is written to Cosmos. Only metadata and known-safe data keys (currently "tls.crt") are preserved; all other data keys and stringData are stripped.

This prevents private keys, passwords, and tokens from leaking into the service cluster database when a ReadDesire watches a Secret.

This is a build block to pull the public part of a serving cert into the cosmos for use building a kubeconfig.

### DIFF
--- a/internal/api/kubeapplier/types_read_desire.go
+++ b/internal/api/kubeapplier/types_read_desire.go
@@ -69,5 +69,13 @@ type ReadDesireStatus struct {
 	// nil means the target either has not been observed yet or is absent
 	// from the cluster — distinguish those two via the "Successful"
 	// condition (Unknown vs. True).
+	//
+	// When TargetItem is a core/v1 Secret, KubeContent is redacted before
+	// storage: structural fields (apiVersion, kind, type) and metadata are
+	// preserved, but annotations that can embed the full Secret (e.g.
+	// kubectl.kubernetes.io/last-applied-configuration) are stripped. In the
+	// data map only known-safe keys (currently "tls.crt") survive; all other
+	// data keys and stringData are removed to prevent private keys,
+	// passwords, and tokens from leaking into Cosmos.
 	KubeContent *runtime.RawExtension `json:"kubeContent,omitempty"`
 }
--- a/kube-applier/pkg/controllers/read_desire_kubernetes/redact.go
+++ b/kube-applier/pkg/controllers/read_desire_kubernetes/redact.go
@@ -1,0 +1,92 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package read_desire_kubernetes
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/Azure/ARO-HCP/internal/api/kubeapplier"
+)
+
+// safeSecretDataKeys lists the data keys that are safe to mirror into
+// ReadDesire.status.kubeContent when the target is a v1 Secret. All other
+// data keys are stripped before the object is persisted to Cosmos. This
+// prevents private keys, passwords, and tokens from leaking into the
+// service cluster's database.
+var safeSecretDataKeys = map[string]struct{}{
+	"tls.crt": {},
+}
+
+// isSecret returns true when the ResourceReference points at a core/v1 Secret.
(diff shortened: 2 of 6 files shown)


kube-applier/pkg/controllers/read_desire_kubernetes/redact.go
```diff
@@ -0,0 +1,92 @@
...
+// annotations that can embed the full Secret (e.g.
+// kubectl.kubernetes.io/last-applied-configuration). In the data map, only
+// keys listed in safeSecretDataKeys survive; on any error reading data the
+// field is removed entirely (fail closed). The object is modified in place;
+// callers must pass a deep copy if the original must be preserved (e.g.
+// informer cache objects).
+func redactSecret(obj *unstructured.Unstructured) {
+	data, found, err := unstructured.NestedMap(obj.Object, "data")
+	if err != nil {
+		unstructured.RemoveNestedField(obj.Object, "data")
```
miguelsorianod, line 55: Would here returning an error instead be better? if the object is a secret it should always have the `data` as a map right?

kube-applier/pkg/controllers/read_desire_kubernetes/redact.go
```diff
@@ -0,0 +1,92 @@
...
+	stripUnsafeAnnotations(obj)
+}
+
+// stripUnsafeAnnotations removes annotations that can embed a full Secret
+// manifest. If the annotations map becomes empty it is removed entirely so
+// the serialized object stays clean.
+func stripUnsafeAnnotations(obj *unstructured.Unstructured) {
+	annotations, found, err := unstructured.NestedStringMap(obj.Object, "metadata", "annotations")
+	if !found || err != nil {
+		return
```
miguelsorianod, line 82: Similar question as https://github.com/Azure/ARO-HCP/pull/6230/changes#r3644126385

internal/api/kubeapplier/types_read_desire.go
```diff
@@ -69,5 +69,13 @@ type ReadDesireStatus struct {
 	// nil means the target either has not been observed yet or is absent
 	// from the cluster — distinguish those two via the "Successful"
 	// condition (Unknown vs. True).
+	//
+	// When TargetItem is a core/v1 Secret, KubeContent is redacted before
+	// storage: structural fields (apiVersion, kind, type) and metadata are
+	// preserved, but annotations that can embed the full Secret (e.g.
+	// kubectl.kubernetes.io/last-applied-configuration) are stripped. In the
+	// data map only known-safe keys (currently "tls.crt") survive; all other
+	// data keys and stringData are removed to prevent private keys,
```
miguelsorianod, line 78: Did we consider as an alternative to preserve all keys but redact their corresponding values (unless whitelisted/safe) ? this would allow preserving the keys so we would know what other keys are there

kube-applier/pkg/controllers/read_desire_kubernetes/redact_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+func TestSyncOnce_SecretTarget_RedactsUnsafeKeys(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	target := secretTarget("my-tls-secret")
+	desire := newReadDesire(t, target)
+	dyn := dynamicForTestdata(t, "testdata/secret_present")
+
+	c, w := startSyncedController(t, ctx, target, desire, dyn)
```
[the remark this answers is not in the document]
miguelsorianod: This seems might be legit because the unit test is failing. Maybe some locally uncommited files?

internal/api/kubeapplier/types_read_desire.go
```diff
@@ -69,5 +69,13 @@ type ReadDesireStatus struct {
 	// nil means the target either has not been observed yet or is absent
 	// from the cluster — distinguish those two via the "Successful"
 	// condition (Unknown vs. True).
+	//
```
miguelsorianod, line 72: When we start to leverage the CS resources endpoints functionality I wonder if we will hit some case where we actually need the values. But maybe in that case it will be an applydesire and not a readdesire? I guess worst case it can be whitelisted or we do needed changes so we don't need a secret for whatever it's originally related to

kube-applier/pkg/controllers/read_desire_kubernetes/redact.go
deads2k: > Would here returning an error instead be better? if the object is a secret it should always have the data as a map right?

I'd rather be resilient here.  Weird things happen.

internal/api/kubeapplier/types_read_desire.go
deads2k: > Did we consider as an alternative to preserve all keys but redact their corresponding values (unless whitelisted/safe) ? this would allow preserving the keys so we would know what other keys are there

This is what OCP must-gather does: tells you how many bytes. But until we have a use-case, I'd rather not.

miguelsorianod: The use case I was thinking was debugging. We would be able to confirm that at least the expected keys are there

PR #6232, opened 2026-07-22
feat: Take hosted cluster backup after cmk key rotation

adalrsjr1:
https://redhat.atlassian.net/browse/ARO-23292

### What

1. Every backup (scheduled and on-demand) is stamped with the active KMS key **fingerprint** (SHA-256 of `vault/key/version`, matching HyperShift's `FingerprintAzureKMSKey`) instead of the raw key version.
2. A new **Key Rotation Backup controller** watches for completed KMS key rotations and creates an on-demand Velero `Backup` under the new fingerprint as soon as re-encryption finishes.
3. On-demand backups are **retained and listable** via the admin API after creation:
   - ApplyDesire is **purged** (document hard-deleted from Cosmos) once the backup succeeds, stopping reconciliation without deleting the Velero `Backup`.
   - ReadDesire is **retained** until Velero's TTL GCs the backup, so the admin API can keep listing it.
   - Durable per-fingerprint completion state (`ServiceProviderCluster.Status.KeyRotationBackupFingerprint`) records that a backup was taken, preventing re-creation loops on later resyncs.
4. Admin API `GET .../backups` lists on-demand backups (name, phase, fingerprint, timestamps).
5. Supporting changes: HyperShift API bump (isolated commit) to expose `SecretEncryptionStatus`; `cosmos-data-flow.md` updated with desire lifecycle details; a pre-existing `syncDeletion` bug fixed (missing `ReadDesire` no longer blocks cleanup).

### Why

After a KMS key rotation, HyperShift re-encrypts etcd under the new key; backups taken with the old key become unrestorable once it's disabled. This closes the gap by taking a backup the moment re-encryption completes. The fingerprint lets operators/tooling tell which key a backup is recoverable with.

**Design critique addressed:** Prior approach treated the on-demand backup as permanent desired state (ServerSideApply forever), causing a delete-and-recreate loop once Velero GC'd the TTL-expired backup. The fix introduces durable per-fingerprint completion state and proper desire lifecycle: ApplyDesire is purged (architectural sanctioned non-destructive stop), ReadDesire is retained for observability, and a SPC status field gates re-creation. This satisfies the reviewer's structural concerns about one-shot backups and edge-case handling.

### Testing

- **Unit tests**:
  - Key rotation controller: `rotationComplete` matrix, `SyncOnce` gates & guards, happy path, idempotency, crash recovery, stale-desire cleanup (both purged and retained), fingerprint recording order, `syncDeletion` purging on-demand desires when the cluster is being deleted.
  - Schedule controller: conditional fingerprint annotation, `syncDeletion` missing-ReadDesire fix (verified to fail without it).
  - Admin `GET .../backups` handler: listing, filtering, name normalization.
  - Desire helpers: `purgeApplyDesire` (doc deletion vs. convert-to-Delete), retention during ReadDesire phase.
- [Integration tests](https://github.com/Azure/ARO-HCP/tree/main/test-integration): none — the flow depends on management-cluster Velero/HyperShift behavior not modeled by the emulator harness; covered by unit + E2E.
- [E2E tests](https://github.com/Azure/ARO-HCP/blob/main/test/e2e/README.md): `kms_key_rotation.go` — creates a KMS cluster, verifies scheduled backups, rotates the key (twice), and after each rotation polls the admin endpoint for an on-demand backup with the new fingerprint. Admin-API assertions are gated behind `framework.IsDevelopmentEnvironment()` since the admin API is only reachable that way in dev environments; the core rotation assertions still run in all environments.

- **Desire lifecycle:** Deleting an ApplyDesire document from Cosmos is the [sanctioned mechanism](https://github.com/Azure/ARO-HCP/blob/main/internal/api/kubeapplierapi/types_apply_desire.go#L48-L49) to stop reconciliation without deleting the target — no effect on the applied kube object. ReadDesire deletion is gated on both `KubeContent==nil` (observed deleted) and `Successful==True` (confirmed by kube-applier), closing ambiguity windows.
- **`syncDeletion` on cluster deletion:** on-demand ApplyDesire/ReadDesire are purged directly via `purgeApplyDesire` (not converted to `Type=Delete`), since they target Backup CRs with no in-flight check — converting to Delete could race a still-running Velero backup and destroy a valid restore point. The Backup CR itself is left to expire at its own TTL.
- **Zombie-backup cleanup:** each sync purges (hard-deletes the Cosmos document) both stale prior-rotation on-demand ApplyDesires and the current rotation's ApplyDesire once its backup succeeds. Purging is what stops kube-applier from re-applying the desire while leaving the already-taken Velero Backup untouched to expire at its own TTL, for both current and superseded rotations.
- **Durable completion state:** `ServiceProviderCluster.Status.KeyRotationBackupFingerprint` is recorded strictly before the ApplyDesire is purged, providing the per-fingerprint completion gate that prevents re-creation even after desires are cleaned up.
- **Out of scope:** N-1 backup retention/pruning (follow-up PR); two pre-existing cleanliness items flagged in review (unused field, `deleteApplyDesire` duplication) left for a dedicated cleanup.

### PR Checklist

- [x] Linked to relevant ticket/issue ([ARO-23292](https://redhat.atlassian.net/browse/ARO-23292))
- [ ] Screenshots included (if graph/UI/metrics changes) — N/A
- [x] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide) — pending final merge
- [x] Commit history is clean (rebased/squashed, fixups pending autosquash)

If E2E tests are included:

- [x] E2E demonstrates the fix: after each of the two key rotations, the test polls the admin API until an on-demand backup with the new key's fingerprint appears — this fails if the on-demand backup is never created, or if its ApplyDesire/ReadDesire are torn down before the admin API can observe it.


backend/pkg/controllers/cluster/backups/schedule_controller.go
```diff
@@ -225,6 +230,18 @@ func (c *backupScheduleSyncer) syncDeletion(ctx context.Context, key controlleru
 		}
 		readDesire, err := c.readDesireLister.GetForCluster(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, applyDesire.ResourceID.Name)
 		if err != nil {
+			if cosmosstorageutils.IsNotFoundError(err) {
```
tony-schndr: I think this might be old implementation?  This controller creates all read desires first, then creates apply desires which makes this crash proof.

If for some reason the read desire is missing and we return an error and let alerting notify us cluster deletion is taking too long.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,337 @@
...
+		if err := readDesireCRUD.Delete(ctx, name); err != nil && !cosmosstorageutils.IsNotFoundError(err) {
+			return false, utils.TrackError(fmt.Errorf("failed to delete stale ReadDesire %s: %w", name, err))
+		}
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func keyRotationBackupName(clusterID, fingerprint string) string {
```
tony-schndr:
```suggestion
func keyRotationBackupName(hostedClusterNamespace, kmsKeyFingerprint string) string {
```

admin/server/handlers/hcp/backups.go
```diff
@@ -226,3 +227,95 @@ func (h *HCPPatchBackupScheduleHandler) ServeHTTP(writer http.ResponseWriter, re
...
+// OnDemandBackupResponse is the JSON response for on-demand backup list endpoints.
+type OnDemandBackupResponse struct {
+	Backups []OnDemandBackupDetail `json:"backups"`
+}
+
+// OnDemandBackupDetail holds per-backup status from the Velero Backup ReadDesire.
+type OnDemandBackupDetail struct {
+	Name           string `json:"name"`
+	Phase          string `json:"phase,omitempty"`
+	Fingerprint    string `json:"fingerprint,omitempty"`
```
tony-schndr:
```suggestion
	KMSKeyFingerprint    string `json:"kmsKeyFingerprint,omitempty"`
```

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,337 @@
...
+		key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, desiredAD); requeue || err != nil {
+		return err
+	}
+
+	// Clean up desires from previous key rotations. Each rotation produces a new
+	// fingerprint-named desire pair; without this the stale ApplyDesires stay
+	// Type=ServerSideApply and kube-applier re-creates their Velero Backups after
+	// TTL expiry (zombie backups). Converting them to Delete stops the resurrection
+	// and removes the backing Velero Backup on the management cluster.
+	if requeue, err := c.deleteStaleOnDemandDesires(ctx, key, applyDesireCRUD, desiredAD); requeue || err != nil {
```
tony-schndr:
```suggestion
	if requeue, err := c.deleteStaleOnDemandApplyDesires(ctx, key, applyDesireCRUD, desiredAD); requeue || err != nil {
```

internal/backup/backup.go
```diff
@@ -26,12 +28,15 @@ import (
...
-	BackupScheduleDesireNamePrefix = "backupschedule-"
-	DesireTagKeySchedule           = "backupschedule"
-	VeleroGroup                    = "velero.io"
-	VeleroVersion                  = "v1"
-	VeleroScheduleResource         = "schedules"
-	VeleroNamespace                = "velero"
+	BackupScheduleDesireNamePrefix  = "backupschedule-"
+	OndemandBackupDesireNamePrefix  = "ondemandbackup-"
+	DesireTagKeySchedule            = "backupschedule"
+	DesireTagKeyOndemandKeyRotation = "ondemandbackup-keyrotation"
```
tony-schndr:
```suggestion
	DesireTagKeyOndemandBackup = "ondemandbackup"
```

I would keep this key generic.  Unless I've missed something, nothing I've read says its handled differently than a normal on demand backup.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,337 @@
...
+	if requeue, err := c.deleteStaleOnDemandReadDesires(ctx, key, readDesireCRUD); requeue || err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// deleteStaleOnDemandDesires drives every on-demand key-rotation ApplyDesire that
+// is not the current desired one through the Apply→Delete→Wait→Purge lifecycle.
+func (c *keyRotationBackupSyncer) deleteStaleOnDemandDesires(
```
tony-schndr: Once an apply desire is successfully created we can delete it, I don't see any reason to keep them around.

Also a rotation / key will likely outlast the backup TTL and will be recreated once velero deletes the schedule.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,337 @@
...
+		if requeue || err != nil {
+			return requeue, err
+		}
+	}
+
+	return false, nil
+}
+
+// deleteStaleOnDemandReadDesires removes on-demand key-rotation ReadDesires whose
+// companion ApplyDesire is already gone.
```
tony-schndr: Considering my comment on https://github.com/Azure/ARO-HCP/pull/6232/changes#r3797291334 , this would then change to deleting a read desire when the fingerprint is old or the kube content is nil (velero cleaned up on ttl)

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,337 @@
...
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to build key rotation backup desire: %w", err))
+	}
+
+	if requeue, err := ensureReadDesireFromApplyDesire(ctx, c.readDesireLister, readDesireCRUD,
+		key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, desiredAD); requeue || err != nil {
+		return err
+	}
+
+	if requeue, err := ensureApplyDesire(ctx, c.applyDesireLister, applyDesireCRUD,
```
tony-schndr: As a consequence of https://github.com/Azure/ARO-HCP/pull/6232/changes#r3797291334, there needs to be a gate on some status field in the service provider cluster that indicates if a backup was successful for a key rotation.

tony-schndr (commented):
Hope my comments make sense.  Generally, I think the controller needs to be restructured a bit to account for edge cases.

AI summary of my review for clarity:
1. The one-shot backup is modeled as standing desired state. The ApplyDesire stays ServerSideApply forever (until the next rotation), so when Velero GCs the backup at TTL, kube-applier just re-creates it
  2. No record that the backup succeeded. Every resync re-ensures the desire, so deleting it just recreates it.
  3. rotationComplete can't help — it stays true for the whole life of the key, so it permits the loop rather than stopping it.
  4. The fix needs durable, per-fingerprint completion state (e.g. a ServiceProviderCluster.Status field), since the backup's completion signal is lost once the desires are cleaned up.
Net: a one-shot backup is treated as permanent desired state with no completion tracking → periodic recreation and delete-and-recreate loops.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
adalrsjr1: Reply for this and to next comments:

Added `ServiceProviderCluster.Status.KeyRotationBackupFingerprint` as durable per-fingerprint completion state, written once the backup succeeds and strictly before the `ApplyDesire` is purged (hard-deleted from Cosmos). `SyncOnce` now checks this field before re-ensuring the desire, so once a fingerprint is recorded the loop stops re-creating it.

The `ApplyDesire` is no longer left as standing ServerSideApply state indefinitely. It's **purged** (document deleted, stopping reconciliation) as soon as it succeeds. The companion `ReadDesire` is **retained** so the admin API can keep listing the backup until Velero's TTL expires.

adalrsjr1:
failed due to slow uninstall
/retest

adalrsjr1:
failed to provision swift
/retest

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,425 @@
...
+// This includes: at least one rotation occurred, no pending key change, and the most recent
+// re-encryption finished. It also returns true after the initial AESCBC-to-KMS migration,
+// creating a known-good restore point under the first KMS key.
+func rotationComplete(hc *hyperv1beta1.HostedCluster) bool {
+	status := hc.Status.SecretEncryption
+	if len(status.History) == 0 {
+		return false
+	}
+	target := status.TargetKey.Azure.KeyVersion
+	if target != "" && target != status.ActiveKey.Azure.KeyVersion {
```
ahitacat: If I recall correctly, once the rotation succeeds the targetKey is clear.

> HCCO (new controller): Derives the current rotation phase from observable state on each reconciliation — not from a stored phase field. Detects key changes by comparing hcp.Spec.SecretEncryption against hcp.Status.SecretEncryption.ActiveKey. When a rotation is needed, sets targetKey, creates a history entry, waits for KAS convergence between stages, creates StorageVersionMigration CRs when conditions for Migrating are met, and on completion updates ActiveKey, clears TargetKey, and sets EtcdDataEncryptionUpToDate=True. Records the derived phase in history[0].state for observability, but does not use it as input for decisions.

https://github.com/openshift/enhancements/blob/master/enhancements/hypershift/etcd-data-reencryption-on-key-rotation.md#proposal

So this controller never say that the rotation has been complete.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,425 @@
...
+		return utils.TrackError(fmt.Errorf("get ReadDesire CRUD: %w", err))
+	}
+
+	activeKey := hostedCluster.Status.SecretEncryption.ActiveKey.Azure
+	kmsKeyFingerprint := backup.AzureKMSKeyFingerprint(activeKey.KeyVaultName, activeKey.KeyName, activeKey.KeyVersion)
+	resourceID := cachedCluster.ResourceID.String()
+	hostedClusterNamespace := cachedServiceProviderCluster.Status.HostedClusterNamespace
+	controlPlaneNamespace := cachedServiceProviderCluster.Status.ControlPlaneNamespace
+
+	backupName := keyRotationBackupName(hostedClusterNamespace, kmsKeyFingerprint)
```
ahitacat: This can be an inline

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,425 @@
...
+	currentDesireName := keyRotationDesireName(backupName)
```
ahitacat: This can be an inline

test/e2e/kms_key_rotation_backup.go
ahitacat, line 1: Why we don't reuse the kms-key-rotation test that already exists? 
This feature cannot work without the kms-key-rotation, if the test fail we can check what is the point of failure.

adalrsjr1: reusing the kms-key-rotation e2e test as you suggested. Added to blocks to take an on-demand backup with the new fingerprint when the kms-rotation completes.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
adalrsjr1: Fixed. Now it checks `TargetKey.Azure.KeyVersion == ""`. We also keep `History[0].State == Completed` as safety guard, since both are written in the same atomic status patch, so it fails safe (skip the backup rather than snapshot mid-rotation) if that atomicity is ever violated. Verified against the current controller (reencryption.go) and the enhancement doc. @ahitacat PTAL.

test/e2e/kms_key_rotation_backup.go
```diff
@@ -0,0 +1,217 @@
...
+							"backupName", b.Name,
+							"phase", b.Phase,
+							"fingerprint", b.KMSKeyFingerprint)
+						return true, nil
+					}
+				}
+				return false, nil
+			}, framework.BackupWaitTimeout, framework.BackupWaitInterval).Should(BeTrue(),
+				"on-demand backup with the new key fingerprint should be created after rotation")
+
```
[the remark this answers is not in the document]
adalrsjr1: description updated.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,426 @@
...
+				return c.recordKeyRotationBackupFingerprint(ctx, key, cachedServiceProviderCluster, kmsKeyFingerprint)
+			}
+			reason = "backup completed successfully; Velero Backup retained until its TTL"
+		}
+
+		logger.Info("purging on-demand backup ApplyDesire", "desire", applyDesire.ResourceID.Name, "reason", reason)
+		// Purge the Cosmos document directly, leaving the Velero Backup to expire at its TTL.
+		if err := purgeApplyDesire(ctx, *applyDesire, applyDesireCRUD); err != nil {
+			return false, err
+		}
```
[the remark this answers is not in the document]
adalrsjr1: description updated.

test/e2e/kms_key_rotation.go
```diff
@@ -114,6 +116,23 @@ var _ = Describe("Customer", func() {
 			err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig)
 			Expect(err).NotTo(HaveOccurred(), "failed to verify HCP cluster viability for update")
 
+			hcpResourceID := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenshift/hcpOpenShiftClusters/%s",
+				metadataapi.Must(tc.SubscriptionID(ctx)), *resourceGroup.Name, clusterName)
+
+			By("creating admin API HTTP client")
+			httpClient, adminAPIAddress, err := tc.NewAdminAPIHTTPClient(ctx)
```
machi1990: The admin API can only be accessed this way in dev environments; 
the verification that follows afterwards relying on the admin api client needs to be executed only if the env is dev.
otherwise, this test will fail in higher envs

machi1990: @adalrsjr1 @ahitacat

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,426 @@
...
+
+	currentDesireName := keyRotationDesireName(keyRotationBackupName(hostedClusterNamespace, kmsKeyFingerprint))
+
+	// Once this rotation's on-demand backup has already completed successfully,
+	// there is nothing left to create: skip straight to the cleanup sweep below.
+	if cachedServiceProviderCluster.Status.KeyRotationBackupFingerprint != kmsKeyFingerprint {
+		// On-demand key-rotation backups use the first (shortest-lived) schedule's TTL.
+		// The backup only needs to survive until the next scheduled backup captures
+		// the post-rotation state.
+		ttl := c.backupConfig.Schedules()[0].TTL
```
machi1990, line 200: What happens if we've no scheduled backups configured?

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,426 @@
...
+		desiredAD, err := buildOnDemandBackupApplyDesire(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, managementClusterResourceID, veleroBackup)
+		if err != nil {
+			return utils.TrackError(fmt.Errorf("failed to build key rotation backup desire: %w", err))
+		}
+
+		// Build the paired ReadDesire, then create-or-update both through the
+		// shared kubeapplierhelpers ensure helpers. The ReadDesire is ensured
+		// first so the on-demand backup's status is observed as soon as its
+		// ApplyDesire lands.
+		desiredRD, err := buildReadDesireFromApplyDesire(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, desiredAD)
```
machi1990:
```suggestion
		desiredReadDesire, err := buildReadDesireFromApplyDesire(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, desiredAD)
```

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,426 @@
...
+		// On-demand key-rotation backups use the first (shortest-lived) schedule's TTL.
+		// The backup only needs to survive until the next scheduled backup captures
+		// the post-rotation state.
+		ttl := c.backupConfig.Schedules()[0].TTL
+		veleroBackup := backup.NewBackup(keyRotationBackupName(hostedClusterNamespace, kmsKeyFingerprint), resourceID, kmsKeyFingerprint, ttl, hostedClusterNamespace, controlPlaneNamespace)
+
+		// No explicit in-flight check: Velero queues backups natively, ensuring
+		// on-demand backups run after scheduled backups and provide a valid backup
+		// as soon as possible, avoiding longer waits for the next scheduled backup.
+		desiredAD, err := buildOnDemandBackupApplyDesire(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, managementClusterResourceID, veleroBackup)
```
machi1990:
```suggestion
		desiredApplyDesire, err := buildOnDemandBackupApplyDesire(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, managementClusterResourceID, veleroBackup)
```

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,426 @@
...
+	if kubeApplierInformers != nil {
+		if err := controller.QueueForInformers(5*time.Minute, applyDesireInformer); err != nil {
+			panic(err)
+		}
+	}
+
+	return controller
+}
+
+func (c *keyRotationBackupSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
```
machi1990, line 126: When a cluster is being deleted; don't we need to clear gracefully remove the ApplyDesire that were written on demand and before valero could wipe them?

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
machi1990: let's type the whole name for clarity and not use acronym?

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
adalrsjr1: Thanks for double check. The [`ToBackendOptions()`](https://github.com/Azure/ARO-HCP/blob/c81c1b63d6ae0dbe81179e1a1b64cae62a4aebb3/backend/cmd/root.go#L466) in `root.go` enforces a backup config will always be created and so [`Schedules()[0]`](https://github.com/Azure/ARO-HCP/blob/c81c1b63d6ae0dbe81179e1a1b64cae62a4aebb3/backend/pkg/controllers/cluster/backups/config.go#L35) will never be empty or nil -- cannot panic.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
adalrsjr1: Fixed by adding a `syncDeletion` path, dispatched the same way `schedule_controller.go` does via `needsDeletionWork()`. One difference: it uses `purgeApplyDesire` instead of `EnsureApplyDesireRemoved`, since on-demand desires point at Backup CRs (not Schedules) and converting to `Type=Delete` could race an in-flight Velero backup removing the Velero backup object. Purging just drops the Cosmos doc and lets the Backup expire at its own TTL. Added a unit test for it.

test/e2e/kms_key_rotation.go
adalrsjr1: Good catch! the other e2e `backup_controller.go` is labelled as `DevelopmentOnly` which was overlooked on this new e2e verification.

I gated the backups verification with `IsDevelopmentEnvironment()`. Therefore, the `kms_key_rotation` tests doesn't fail in higher environments and still can exercise the on_demand backups in dev env.

@ahitacat let me know if you are ok with this approach or prefer I have a dedicated test for the backup on key rotation.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
machi1990: > Purging just drops the Cosmos doc and lets the Backup expire at its own TTL. Added a unit test for it.

Won't it prevent or prolong cluster deletion until it expires?

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
machi1990: Do we want in this case to error out in the controller if it's not the case?

adalrsjr1: In short this error check will be a dead code.

`Schedules()`'s switch only has two cases (testing 1 entry, default 3, which also catches the zero value), so every `BackupCadenceProfile` value returns non-empty. `root.go` also validates the cadence flag at startup, rejecting anything but production/testing. No reachable branch to error out on.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
adalrsjr1: Velero's Backup CRs don't carry a k8s finalizer, so the namespace's own teardown will just GC the un-finalized Backup CR along with everything else, nothing waits on it or its TTL. 

If deletion races an in-flight backup, Velero's own ItemOperationTimeout (15m, already set on our spec) force-cancels a stuck DataUpload.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
machi1990: The controller, a component on itself can validate the invariant to avoid panicking.
For example https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go#L445 returns an error if parsing fails instead of doing a `Must(semver.Parse(...))` assuming that the input is always valid. 
So even if the branch will essentially be a deadcode, should the future arrive and invalidate that, the controller won't panic but rather error out with a clear message that we can trace back.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
machi1990: Thanks for the explanation.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
adalrsjr1: Fair point. I adde a checker and returns a tracked error as in the example you shared.

test/e2e/kms_key_rotation.go
```diff
@@ -114,6 +117,30 @@ var _ = Describe("Customer", func() {
 			err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig)
 			Expect(err).NotTo(HaveOccurred(), "failed to verify HCP cluster viability for update")
 
+			hcpResourceID := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenshift/hcpOpenShiftClusters/%s",
+				metadataapi.Must(tc.SubscriptionID(ctx)), *resourceGroup.Name, clusterName)
```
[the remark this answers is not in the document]
adalrsjr1: false positive

adalrsjr1:
failing due to infra issues

machi1990 (commented):
/lgtm

Will leave the final lgtm to @ahitacat and @tony-schndr

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,490 @@
...
+// a conservative fail-safe: if those fields are ever patched separately, we skip the
+// backup instead of snapshotting mid-rotation. This also covers an existing AESCBC
+// cluster's spec-driven migration to KMS after the first KMS key is active; clusters
+// created directly with KMS never populate History and therefore never trigger this path.
+func rotationComplete(hc *hyperv1beta1.HostedCluster) bool {
+	status := hc.Status.SecretEncryption
+	if len(status.History) == 0 {
+		return false
+	}
+	return status.TargetKey.Azure.KeyVersion == "" &&
```
ahitacat, line 65: This will be true once it is completed, and until the next key rotation is present. Shouldn't is also needs to check if the latest backups? So if there is one that has been completed with the same fingerprint it is no need to create a new one?

The problem is that every time this controller runs against a cluster, the key rotation will match this rotationComplete logic after it has rotated once they kms key. Meaning, every time this will be true and a new backup will be created in consequence.

adalrsjr1: thx for review. This gate you suggested for checking the fingerprint already happens at line 191 (`cachedServiceProviderCluster.Status.KeyRotationBackupFingerprint != kmsKeyFingerprint`), checked after `rotationComplete` passes.

Once a backup completes for the active key, the fingerprint is recorded at line 370 and creation is skipped on every later reconcile until the key rotates again.

ahitacat: Thanks for the clarification, is easy to get lost in long changes.

ahitacat:
/lgtm
Waiting for final lgtm from @tony-schndr

geoberle:
the current dataflow seems to be 
* consume readdesire HostedCluster to spot changes in the used key
* use applydesires to annotate the backup

i think the mgmt-agent would be a better place to handle this, because everything can be managed directly on the mgmt cluster. HostedCluster resources lives there + the backups resources as well.

i understand the urgency to annotate the backups so i'm not going to block on this. but consider a followup action

backend/pkg/controllers/cluster/backups/schedule_controller.go
```diff
@@ -162,11 +162,16 @@ func (c *backupScheduleSyncer) SyncOnce(ctx context.Context, key controllerutils
 		return utils.TrackError(fmt.Errorf("get ReadDesire CRUD: %w", err))
 	}
 
+	kmsKeyFingerprint := ""
+	if cm := cachedCluster.CustomerProperties.Etcd.DataEncryption.CustomerManaged; cm != nil && cm.Kms != nil && cm.Kms.ActiveKey.Version != "" {
```
tony-schndr, line 169: There is a race here between the kmsKeyFingerprint being updated on the schedule, a backup being executed on the hour, and the rotation/reencryption not being completed.

The annotation on a scheduled backup can advance to a new fingerprint while the actual on-disk etcd encryption is still mid-rotation (or hasn't started). If the schedule's cron fires during that window, the backup captures whatever encryption state exists at that instant but is stamped with the desired fingerprint, which may not match the data's real encryption key.

tony-schndr: If my understanding above is correct, then we don't want to patch this field until we know that encryption has completed.

adalrsjr1: Fix: the fingerprint now comes from the live `HostedCluster.Status.SecretEncryption.ActiveKey.Azure`, same as `key_rotation_controller.go` already does. `CustomerProperties` is only used now to check "is CMK enabled at all."

I kept this independent of rotationComplete() on purpose, the schedule should always tag whatever key is actually active, even mid-rotation, not just once rotation finishes.

Also updated the existing rotation test and added a new one covering the in-flight-rotation case.

tony-schndr: This makes sense, thanks for addressing this.

admin/server/handlers/hcp/backups.go
```diff
@@ -226,3 +227,95 @@ func (h *HCPPatchBackupScheduleHandler) ServeHTTP(writer http.ResponseWriter, re
...
+			}
+			detail.Phase = string(veleroBackup.Status.Phase)
+			detail.KMSKeyFingerprint = veleroBackup.Annotations[controllerutils.HcpClusterKmsKeyFingerprintAnnotation]
+			if veleroBackup.Status.StartTimestamp != nil {
+				detail.StartTime = veleroBackup.Status.StartTimestamp.Time.UTC().Format(time.RFC3339)
+			}
+			if veleroBackup.Status.CompletionTimestamp != nil {
+				detail.CompletionTime = veleroBackup.Status.CompletionTimestamp.Time.UTC().Format(time.RFC3339)
+			}
+		}
```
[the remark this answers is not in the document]
adalrsjr1: In practice KubeContent.Raw is written by kube-applier serializing a real Velero Backup object it observed, so malformed JSON is unlikely under normal operation. This is a defensive-hardening suggestion, not a live bug.

backend/pkg/controllers/cluster/backups/key_rotation_controller.go
```diff
@@ -0,0 +1,501 @@
...
+		return utils.TrackError(fmt.Errorf("get ApplyDesire CRUD: %w", err))
+	}
+	readDesireCRUD, err := kubeApplierClient.ReadDesiresForCluster(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("get ReadDesire CRUD: %w", err))
+	}
+
+	activeKey := hostedCluster.Status.SecretEncryption.ActiveKey.Azure
+	kmsKeyFingerprint := backup.AzureKMSKeyFingerprint(activeKey.KeyVaultName, activeKey.KeyName, activeKey.KeyVersion)
+	resourceID := cachedCluster.ResourceID.String()
```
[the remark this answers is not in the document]
adalrsjr1: a false positive for a real bug (the bad state it describes is already precluded by rotationComplete)

PR #6240, opened 2026-07-23
test(e2e): fix z-stream upgrade tests

DrakeSeifert:
NOTE: This PR is awaiting the completion of the following:
- https://github.com/Azure/ARO-HCP/pull/6538

UPDATE: Above PR has completed.

Jira:
https://redhat.atlassian.net/browse/ARO-27283

### What

- made the z-stream tests more robust by updating logic to determine the initial install version and the target version (the latest z-stream version the cluster should upgrade to). This required writing a new method `GetShortestUpgradePath` to reach out to Cincinnati to get the graph of all valid upgrade paths for the current minor version `x.y`, and next minor version `x.y+1`. The data returned from Cincinnati was then used to:
  1. calculate the target version by determining the shortest path to `x.y+1.zLatest` from valid z candidates in `x.y`. (Note that if `x.y` is the latest available minor version, `x.y.zLatest` is automatically made the target version.)
  2. calculate the install version by determining the shortest path to the target version from the remaining valid z candidates within `x.y`. (If no remaining valid z candidates exist in `x.y`, the target version is made the install version and the target version is set to `""`, indicating that no upgrade path exists. Currently, this edge case causes the test to skip since there is no z-stream upgrade to test.)
- cleaned up and simplified z-stream tests:
  - removed the `baseInstallVersion` test data parameter as the install version is now determined automatically. The test should now automatically account for broken upgrade paths, so there is no need to manually specify an initial installation version in order to skip known broken versions. If the e2e test fails due to installing a broken version, that means we caught a regression in Cincinnati that could affect our customers.
  - skip the test early if no z-stream upgrade path exists, before provisioning resources.
  - added comments, `By()` annotations, and reordered lines for clarity.
  - end tests early with `gomega.StopTrying()` in order to stop the `Eventually()` loop from repeatedly looping over unrecoverable errors. This should save us ~45min per test when those errors are encountered.

### Why

- z-stream bug was causing test runs to fail.
- e2e z-stream test control flow was difficult to follow.

### Testing

e2e test validation here: https://github.com/Azure/ARO-HCP/pull/6241

The above PR filters the e2e tests down to only the modified tests:
image

z-stream e2e tests passed: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6241/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2080312215386198016
image

Note that the e2e test `Service Provider should upgrade the control plane z-stream automatically on behalf of the customer for 4.23` is currently being skipped on main, and is not a byproduct of this PR.

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/e2e/control_plane_automated_z_stream_upgrade.go
+++ b/test/e2e/control_plane_automated_z_stream_upgrade.go
@@ -34,7 +34,7 @@ import (
 var _ = Describe("Service Provider", func() {
 	DescribeTable("should upgrade the control plane z-stream automatically on behalf of the customer",
 		labels.MIContainers(1),
-		func(ctx context.Context, minorVersion string, baseInstallVersion string) {
+		func(ctx context.Context, minorVersion string) {
 			const (
 				customerNetworkSecurityGroupName = "customer-nsg-zstream-"
 				customerVnetName                 = "customer-vnet-zstream-"
@@ -42,18 +42,24 @@ var _ = Describe("Service Provider", func() {
 				customerClusterNamePrefix        = "cluster-zstream-"
 			)
 
+			// ARRANGE
+
 			tc := framework.NewTestContext()
 
-			if len(baseInstallVersion) == 0 {
-				baseInstallVersion = minorVersion // set it to minor so that we defaul to .0 as the patch version
-			}
-			installVersion, hasUpgradePath, err := framework.GetInstallVersionForZStreamUpgrade(ctx, "candidate", baseInstallVersion)
+			By("checking if z-stream upgrade path exists")
+			installVersion, targetVersion, err := framework.GetInstallVersionForZStreamUpgrade(ctx, "candidate", minorVersion)
 			if err != nil {
 				if cincinnati.IsCincinnatiVersionNotFoundError(err) {
-					Skip(fmt.Sprintf("Cincinnati returned version not found for configured id %s (minor %s)", baseInstallVersion, minorVersion))
+					Skip(fmt.Sprintf("Cincinnati returned version not found for configured id %s", minorVersion))
 				}
 				Expect(err).NotTo(HaveOccurred(), "failed to get install version for z-stream upgrade of %s", minorVersion)
 			}
+
+			if targetVersion == "" {
+				Skip(fmt.Sprintf("the z-stream version %s is latest and has no z-stream upgrade path available on the candidate channel", installVersion))
+			}
+
+			By("setting up test prerequisites")
 			if tc.UsePooledIdentities() {
 				err := tc.AssignIdentityContainers(ctx, 1, framework.IdentityContainerAssignmentRetryInterval)
 				Expect(err).NotTo(HaveOccurred(), "failed to assign pooled identity containers")
@@ -62,24 +68,20 @@ var _ = Describe("Service Provider", func() {
 			versionLabel := strings.ReplaceAll(minorVersion, ".", "-") // e.g. "4.20" -> "4-20"
 			suffix := rand.String(6)
 			clusterName := customerClusterNamePrefix + versionLabel + "-" + suffix
-			clusterParams := framework.NewDefaultClusterParams20240610()
-			clusterParams.ClusterName = clusterName
-			clusterParams.OpenshiftVersionId = installVersion
-
-			// We use the candidate channel to potentially catch early z-stream upgrades.
-			// issues before they reach stable.
-			By("using the candidate channel")
-			clusterParams.ChannelGroup = "candidate"
 
 			By("creating resource group")
 			resourceGroup, err := tc.NewResourceGroup(ctx, "rg-zstream-upgrade-"+versionLabel, tc.Location())
 			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for z-stream upgrade of %s", minorVersion)
 
-			managedResourceGroupName := framework.SuffixName(*resourceGroup.Name+"-zstream-"+suffix, "-managed", 64)
-			clusterParams.ManagedResourceGroupName = managedResourceGroupName
-
 			By("creating customer resources")
-			clusterParams, err = tc.CreateClusterCustomerResources20240610(ctx,
+			clusterParams := framework.NewDefaultClusterParams20240610()
+			clusterParams.ManagedResourceGroupName = framework.SuffixName(*resourceGroup.Name+"-zstream-"+suffix, "-managed", 64)
+			clusterParams.ClusterName = clusterName
+			clusterParams.OpenshiftVersionId = installVersion
+			clusterParams.ChannelGroup = "candidate" // use the candidate channel to potentially catch early z-stream upgrade issues before they reach stable.
+
+			clusterParams, err = tc.CreateClusterCustomerResources20240610(
+				ctx,
 				resourceGroup,
 				clusterParams,
 				map[string]interface{}{
@@ -96,8 +98,10 @@ var _ = Describe("Service Provider", func() {
 			// 4.22 control plane provisioning has been consistently slower and frequently hits the default timeout.
 			// Bump the create+wait budget to reduce flaky timeouts for this minor.
 			if minorVersion == "4.22" {
-				clusterCreationTimeout = 35 * time.Minute
+				clusterCreationTimeout += 15 * time.Minute
 			}
+
+			// ACT
 
 			By(fmt.Sprintf("creating the HCP cluster with version '%s' on candidate channel", installVersion))
 			err = tc.CreateHCPClusterFromParam20240610(
@@ -107,6 +111,9 @@ var _ = Describe("Service Provider", func() {
 				clusterParams,
 				clusterCreationTimeout,
 			)
+
+			// ASSERT
+
 			Expect(err).NotTo(HaveOccurred(), "failed to create HCP cluster %q with version %s on candidate channel", clusterName, installVersion)
 
 			By("verifying the cluster is viable")
@@ -121,25 +128,17 @@ var _ = Describe("Service Provider", func() {
 			err = verifiers.VerifyHCPCluster(ctx, adminRESTConfig)
 			Expect(err).NotTo(HaveOccurred(), "failed to verify HCP cluster %q is viable", clusterName)
 
-			if !hasUpgradePath {
-				By("skipping z-stream upgrade verification: no upgrade path (cluster installed at latest)")
-				return
-			}
-
 			By("verifying that only a z-stream upgrade was performed")
 			Eventually(func() error {
-				return verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyHostedControlPlaneZStreamUpgradeOnly(installVersion))
+				return verifiers.VerifyHCPCluster(ctx, adminRESTConfig, verifiers.VerifyHostedControlPlaneZStreamUpgradeOnly(installVersion, targetVersion))
 			}, framework.HCPClusterVersionUpgradeTimeout, 2*time.Minute).Should(Succeed())
 			GinkgoLogr.Info("z-stream upgrade verification passed", "installVersion", installVersion)
 		},
 
-		// for 4.19, if we start with 4.19.0, the version that has an upgrade path to 4.19.latest is
-		// 4.19.3 but cluster install on this version fails with KMS authentication problem.
-		// For all the other minor versions, we can start with 4.y.0 and install the latest version in the candidate channel.
-		Entry("for 4.19", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.19", "4.19.25"),
-		Entry("for 4.20", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.20", ""),
-		Entry("for 4.21", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.21", ""),
-		Entry("for 4.22", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, labels.AllowRetry, "4.22", ""), // owner: @raelga, tracking: AROSLSRE-1319. Known-issue test, retriable during EV2 gating. Remove this label when the issue is fixed.
-		Entry("for 4.23", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.23", ""),
+		Entry("for 4.19", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.19"),
+		Entry("for 4.20", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.20"),
+		Entry("for 4.21", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.21"),
+		Entry("for 4.22", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, labels.AllowRetry, "4.22"), // owner: @raelga, tracking: AROSLSRE-1319. Known-issue test, retriable during EV2 gating. Remove this label when the issue is fixed.
+		Entry("for 4.23", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.23"),
 	)
 })
--- a/test/e2e/control_plane_automated_z_stream_upgrade.go
+++ b/test/e2e/control_plane_automated_z_stream_upgrade.go
@@ -34,7 +34,8 @@ import (
 var _ = Describe("Service Provider", func() {
 	DescribeTable("should upgrade the control plane z-stream automatically on behalf of the customer",
 		labels.MIContainers(1),
-		func(ctx context.Context, minorVersion string, baseInstallVersion string) {
+		func(ctx context.Context, minorVersion string, installVersion string) {
+
 			const (
 				customerNetworkSecurityGroupName = "customer-nsg-zstream-"
 				customerVnetName                 = "customer-vnet-zstream-"
@@ -42,18 +43,24 @@ var _ = Describe("Service Provider", func() {
 				customerClusterNamePrefix        = "cluster-zstream-"
 			)
 
+			// ARRANGE
+
 			tc := framework.NewTestContext()
 
-			if len(baseInstallVersion) == 0 {
-				baseInstallVersion = minorVersion // set it to minor so that we defaul to .0 as the patch version
-			}
-			installVersion, hasUpgradePath, err := framework.GetInstallVersionForZStreamUpgrade(ctx, "candidate", baseInstallVersion)
+			By("checking if z-stream upgrade path exists")
+			hasUpgradePath, err := framework.HasZStreamUpgradePath(ctx, "candidate", installVersion)
 			if err != nil {
 				if cincinnati.IsCincinnatiVersionNotFoundError(err) {
-					Skip(fmt.Sprintf("Cincinnati returned version not found for configured id %s (minor %s)", baseInstallVersion, minorVersion))
+					Skip(fmt.Sprintf("Cincinnati returned version not found for configured id %s (minor %s)", installVersion, minorVersion))
 				}
 				Expect(err).NotTo(HaveOccurred(), "failed to get install version for z-stream upgrade of %s", minorVersion)
 			}
+
+			if !hasUpgradePath {
+				Skip(fmt.Sprintf("the z-stream version %s is latest and has no z-stream upgrade path available on the candidate channel", installVersion))
+			}
+
+			By("setting up test prerequisites")
 			if tc.UsePooledIdentities() {
 				err := tc.AssignIdentityContainers(ctx, 1, framework.IdentityContainerAssignmentRetryInterval)
 				Expect(err).NotTo(HaveOccurred(), "failed to assign pooled identity containers")
@@ -62,24 +69,20 @@ var _ = Describe("Service Provider", func() {
 			versionLabel := strings.ReplaceAll(minorVersion, ".", "-") // e.g. "4.20" -> "4-20"
 			suffix := rand.String(6)
 			clusterName := customerClusterNamePrefix + versionLabel + "-" + suffix
-			clusterParams := framework.NewDefaultClusterParams20240610()
-			clusterParams.ClusterName = clusterName
-			clusterParams.OpenshiftVersionId = installVersion
-
-			// We use the candidate channel to potentially catch early z-stream upgrades.
-			// issues before they reach stable.
-			By("using the candidate channel")
-			clusterParams.ChannelGroup = "candidate"
 
 			By("creating resource group")
 			resourceGroup, err := tc.NewResourceGroup(ctx, "rg-zstream-upgrade-"+versionLabel, tc.Location())
 			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for z-stream upgrade of %s", minorVersion)
 
-			managedResourceGroupName := framework.SuffixName(*resourceGroup.Name+"-zstream-"+suffix, "-managed", 64)
-			clusterParams.ManagedResourceGroupName = managedResourceGroupName
-
 			By("creating customer resources")
-			clusterParams, err = tc.CreateClusterCustomerResources20240610(ctx,
+			clusterParams := framework.NewDefaultClusterParams20240610()
+			clusterParams.ManagedResourceGroupName = framework.SuffixName(*resourceGroup.Name+"-zstream-"+suffix, "-managed", 64)
+			clusterParams.ClusterName = clusterName
+			clusterParams.OpenshiftVersionId = installVersion
+			clusterParams.ChannelGroup = "candidate" // use the candidate channel to potentially catch early z-stream upgrade issues before they reach stable.
+
+			clusterParams, err = tc.CreateClusterCustomerResources20240610(
+				ctx,
 				resourceGroup,
 				clusterParams,
 				map[string]interface{}{
@@ -96,8 +99,10 @@ var _ = Describe("Service Provider", func() {
 			// 4.22 control plane provisioning has been consistently slower and frequently hits the default timeout.
 			// Bump the create+wait budget to reduce flaky timeouts for this minor.
 			if minorVersion == "4.22" {
-				clusterCreationTimeout = 35 * time.Minute
+				clusterCreationTimeout += 15 * time.Minute
 			}
+
+			// ACT
 
 			By(fmt.Sprintf("creating the HCP cluster with version '%s' on candidate channel", installVersion))
 			err = tc.CreateHCPClusterFromParam20240610(
@@ -107,6 +112,9 @@ var _ = Describe("Service Provider", func() {
 				clusterParams,
 				clusterCreationTimeout,
 			)
+
+			// ASSERT
+
 			Expect(err).NotTo(HaveOccurred(), "failed to create HCP cluster %q with version %s on candidate channel", clusterName, installVersion)
 
 			By("verifying the cluster is viable")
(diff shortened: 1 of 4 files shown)


test/e2e/control_plane_automated_z_stream_upgrade.go
```diff
@@ -61,24 +67,19 @@ var _ = Describe("Service Provider", func() {
 			versionLabel := strings.ReplaceAll(minorVersion, ".", "-") // e.g. "4.20" -> "4-20"
 			suffix := rand.String(6)
 			clusterName := customerClusterNamePrefix + versionLabel + "-" + suffix
-			clusterParams := framework.NewDefaultClusterParams20240610()
```
DrakeSeifert, line 65: These params were reordered for clarity.

test/e2e/control_plane_automated_z_stream_upgrade.go
```diff
@@ -33,26 +33,32 @@ import (
...
+			// ARRANGE
+
 			tc := framework.NewTestContext()
 
-			if len(baseInstallVersion) == 0 {
-				baseInstallVersion = minorVersion // set it to minor so that we defaul to .0 as the patch version
-			}
-			installVersion, hasUpgradePath, err := framework.GetInstallVersionForZStreamUpgrade(ctx, "candidate", baseInstallVersion)
+			By("checking if z-stream upgrade path exists")
+			hasUpgradePath, err := framework.HasZStreamUpgradePath(ctx, "candidate", installVersion)
```
DrakeSeifert: The new `HasZStreamUpgradePath` function in `version_helper.go` is called here. Note that we no longer return an `installVersion`--that is now specified as a test case parameter. All we care about now is if a z-stream upgrade path exists. If so, install an HCP cluster at the `installVersion`.

test/e2e/control_plane_automated_z_stream_upgrade.go
```diff
@@ -95,9 +96,11 @@ var _ = Describe("Service Provider", func() {
 			// 4.22 control plane provisioning has been consistently slower and frequently hits the default timeout.
 			// Bump the create+wait budget to reduce flaky timeouts for this minor.
 			if minorVersion == "4.22" {
-				clusterCreationTimeout = 35 * time.Minute
+				clusterCreationTimeout += 15 * time.Minute
```
DrakeSeifert, line 101: Append time instead of hard setting a new value in case `framework.ClusterCreationTimeout` ever changes.

test/e2e/nodepool_version_upgrade.go
```diff
@@ -58,11 +58,15 @@ var _ = Describe("Customer", func() {
...
-				if cincinnati.IsCincinnatiVersionNotFoundError(err) {
-					Skip(fmt.Sprintf("Cincinnati returned version not found for target minor %s on channel %s", targetMinor, channelGroup))
+				nodePoolInitialVersion = targetMinor + ".0"
+				hasUpgradePath, err = framework.HasZStreamUpgradePath(ctx, channelGroup, targetMinor)
+				if err != nil {
+					if cincinnati.IsCincinnatiVersionNotFoundError(err) {
+						Skip(fmt.Sprintf("Cincinnati returned version not found for target minor %s on channel %s", targetMinor, channelGroup))
+					}
+					Expect(err).NotTo(HaveOccurred(), "failed to determine z-stream install version for minor %s", targetMinor)
 				}
```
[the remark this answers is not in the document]
DrakeSeifert: fixed

test/e2e/control_plane_automated_z_stream_upgrade.go
```diff
@@ -61,24 +67,19 @@ var _ = Describe("Service Provider", func() {
...
-			clusterParams.OpenshiftVersionId = installVersion
-
-			// We use the candidate channel to potentially catch early z-stream upgrades.
-			// issues before they reach stable.
-			By("using the candidate channel")
-			clusterParams.ChannelGroup = "candidate"
 
-			By("creating resource group")
 			resourceGroup, err := tc.NewResourceGroup(ctx, "rg-zstream-upgrade-"+versionLabel, tc.Location())
 			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for z-stream upgrade of %s", minorVersion)
```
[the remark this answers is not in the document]
DrakeSeifert: fixed

test/e2e/control_plane_automated_z_stream_upgrade.go
```diff
@@ -120,25 +127,18 @@ var _ = Describe("Service Provider", func() {
...
-		// 4.19.3 but cluster install on this version fails with KMS authentication problem.
-		// For all the other minor versions, we can start with 4.y.0 and install the latest version in the candidate channel.
-		Entry("for 4.19", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.19", "4.19.25"),
-		Entry("for 4.20", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.20", ""),
-		Entry("for 4.21", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.21", ""),
-		Entry("for 4.22", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.22", ""),
-		Entry("for 4.23", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.23", ""),
+		// upgrading from 4.19.0 to 4.19.3 fails with a KMS authentication problem, so it needs to be skipped.
+		Entry("for 4.19", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.19", "4.19.4"),
+		Entry("for 4.20", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.20", "4.20.0"),
```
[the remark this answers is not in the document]
DrakeSeifert: fixed

test/e2e/control_plane_automated_z_stream_upgrade.go
```diff
@@ -61,24 +67,20 @@ var _ = Describe("Service Provider", func() {
...
-			managedResourceGroupName := framework.SuffixName(*resourceGroup.Name+"-zstream-"+suffix, "-managed", 64)
-			clusterParams.ManagedResourceGroupName = managedResourceGroupName
-
 			By("creating customer resources")
-			clusterParams, err = tc.CreateClusterCustomerResources20240610(ctx,
+			clusterParams := framework.NewDefaultClusterParams20240610()
+			clusterParams.ManagedResourceGroupName = framework.SuffixName(*resourceGroup.Name+"-zstream-"+suffix, "-managed", 64)
+			clusterParams.ClusterName = clusterName
+			clusterParams.OpenshiftVersionId = installVersion
+			clusterParams.ChannelGroup = "candidate" // use the candidate channel to potentially catch early z-stream upgrades issues before they reach stable.
```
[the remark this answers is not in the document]
DrakeSeifert: fixed

DrakeSeifert:
Hi @deads2k and @miquelsi, this PR is now ready for review 🙂

test/e2e/control_plane_automated_z_stream_upgrade.go
machi1990: > All we care about now is if a z-stream upgrade path exists. If so, install an HCP cluster at the installVersion.

This comment https://github.com/DrakeSeifert/ARO-HCP/blob/67b41969242ae917ac5d5a22c726981dcafb3d8f/test/e2e/control_plane_automated_z_stream_upgrade.go#L80 was also a critical part of this test design;

We want to have some sort of guarantee that an upgrade path from any 4.y.z level to 4.y.z+1 as the new z+1 is released doesn't result onto a broken upgrade.

Concretely; with this change, 

```
[v0,v1,v2...vLatest-1]
```

And a new vLatest is released to give you 
```
[v0,v1,v2...vLatest-1,vLatest]
```

The test will always be exercising  the upgrade from v0 -> vLatest and not we lose the `vLatest-1 -> vLatest` upgrade.

Which is a desired property of the z-stream upgrade. At any given time, clusters will be on the latest z that has an upgrade path to the next minor; so when a new version is available with an upgrade to the next minor, we'll upgrade from the previous release and not the "some hardcoded or initial version within that x.y release line"

machi1990 (changes requested):
/hold 

Left a comment that needs thinking through

test/e2e/control_plane_automated_z_stream_upgrade.go
DrakeSeifert: Hey @machi1990, I appreciate the clarity here. Certainly, it appears the logic behind this was more complicated than I initially gave it credit for.

There are some aspects of the upgrade logic that I still do not understand; you say that "clusters will be on the latest z that has an upgrade path the the next minor", my question is: what situations would have to occur for that not to be the case, and how do we handle it? Would we introduce a patch to fix like this?:

`x.y.z`: Initial version. Customers here (due to not upgrading early, etc) will upgrade to `x.y.z+2` because `x.y.z+1` is known to be broken.
`x.y.z+1`: Broken version. Customers who unfortunately have upgraded here will upgrade to `x.y.z+2`. There is some engineering effort on our side to make that possible (do we roll them back to `x.y.z` first?)
`x.y.z+2`: Patched, stable version.

Also, why do we prioritize the path to the next y-stream, if what we really care about is the path to the latest _stable version_ (regardless of x/y/z value)? For example, if the customer is already on the latest y version, but is on a broken z version and needs to update, the y version is irrelevant. We want an upgrade path to the latest _stable version_ in all scenarios. And of course, if the upgrade path to the latest stable version includes a y update, we would stop one version before the y update so that only the z version is modified.

If I better understood this process, I think I could help make this test more useful and more stable.

machi1990: @DrakeSeifert In ARO-HCP, the customer manages X.Y part. The service maintain the z part via the https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go controller; z-stream upgrade is an automated process and not customer initiated. 

A bit of context:

1. During install; this gets exercised https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go#L270 and the latest version within the desired minor will be selected and it'll be used during install via https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go controller.

For upgrades (post install) there are two cases
2.a) Either it is a y-stream customer initiated upgrades in which case everything from https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go#L343 gets exercised.  
2.b) Either it is a z-stream automated upgrade (not customer initiated but service level initiated) https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/upgradecontrollers/control_plane_desired_version_controller.go#L340

Caveat: In the future, there will be a third scenario of rolling back; tracked under https://redhat.atlassian.net/browse/ARO-22685

Now, once a new upgrade version has been selected, this controller https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/upgradecontrollers/trigger_control_plane_upgrade_controller.go triggers the upgrade. 

Now with that context;

Going back to some of your questions

> Also, why do we prioritize the path to the next y-stream, if what we really care about is the path to the latest stable version (regardless of x/y/z value)? 

When picking a z-stream (we managed it not the customer), we want to ensure that the cluster upgradability path to the next minor (customer controlled) isn't broken resulting us in locking out the customer for some time from upgrading their cluster to their desired x.y+1 minor.

>  you say that "clusters will be on the latest z that has an upgrade path the the next minor", 

Yes; see the linked logic of the desired version selection.


> my question is: what situations would have to occur for that not to be the case, and how do we handle it? 

A bug in the design of the controller that selects the z level of a cluster or in the future once we've sre initiated rollbacks https://redhat.atlassian.net/browse/ARO-22685 to rollback a cluster to a certain level because the automated upgrade was deemed problematic for that cluster. 

> Would we introduce a patch to fix like this?:
> x.y.z: Initial version. Customers here (due to not upgrading early, etc) will upgrade to x.y.z+2 because x.y.z+1 is known to be broken.
> x.y.z+1: Broken version. Customers who unfortunately have upgraded here will upgrade to x.y.z+2. There is some engineering effort on our side to make that possible (do we roll them back to x.y.z first?)
> x.y.z+2: Patched, stable version.

Notice in the context I gave above how a normal customer doesn't get to chose the z-level: only possible for some specific subscription that have an AFEC flag - internal subscriptions.

But regardless of the path that get us to a x.y.z+1 that's broken the is often; add that version as a conditional edge in cincinatti graph so that the algorithm doesn't to consider it as a viable candidate.

If there is a newer x.y.z+2, all cluster in x.y.z+1 will be upgraded to it automatically if they meet the selection criteria. There a some special scenarios here that x.y.z+2 doesn't have an upgrade  to the available next minor, in this case x.y.z+2 won't be selected, so a rollforward won't be a fix but a roll back which is where https://redhat.atlassian.net/browse/ARO-22685 fits in.

With the premise of at any time `T` all clusters will be in latest z for a given minor; it's of paramount importance that the e2e tries to exercise the upgrade from the previous available level to the latest and not from a dotZero version like what's proposed in the PR. That will allow to catch any upgrade issues early from the versions we care about.

Cases:

1. If we've `[v0,v1,v2...vLatest-1,vLatest]`, and next minor is available and vLatest has path to the next minor
Then install should've be any version in [vLatest-1...v2,v1,v0] (notice the order) that has an upgrade path to vLatest. 
2.  If we've `[v0,v1,v2...vLatest-1,vLatest]`, and next minor is available and vLatest doesn't have the path to the next minor but `vLatest-1` does, then install should've be any version in [vLatest-2...v2,v1,v0] (notice the order) that has an upgrade path to vLatest-1. 
3. If we've `[v0,v1,v2...vLatest-1,vLatest]`, and next minor isn't available then install should've be any version in [vLatest-1...v2,v1,v0] (notice the order) that has an upgrade path to vLatest.

test/util/framework/version_helper.go
```diff
@@ -118,27 +120,59 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
...
+	for i := targetIdx + 1; i < len(candidates); i++ {
+		path, err := GetShortestUpgradePath(ctx, channelGroup, candidates[i].String(), target.String())
 		if err != nil {
+			if errors.Is(err, ErrNoUpgradePath) {
+				continue
+			}
 			return "", false, err
 		}
-		if len(upgradeTargets) > 0 {
-			return candidates[i+1].String(), true, nil
```
DrakeSeifert, line 138: Note there is a bug here: We are assuming there is an upgrade path from `candidates[i+1]` to `candidates[i]`. My code in "Phase 2" accounts for this.

test/e2e/control_plane_automated_z_stream_upgrade.go
```diff
@@ -121,25 +129,17 @@ var _ = Describe("Service Provider", func() {
...
 
-		// for 4.19, if we start with 4.19.0, the version that has an upgrade path to 4.19.latest is
-		// 4.19.3 but cluster install on this version fails with KMS authentication problem.
-		// For all the other minor versions, we can start with 4.y.0 and install the latest version in the candidate channel.
-		Entry("for 4.19", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.19", "4.19.25"),
-		Entry("for 4.20", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.20", ""),
-		Entry("for 4.21", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.21", ""),
-		Entry("for 4.22", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.22", ""),
-		Entry("for 4.23", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.23", ""),
+		Entry("for 4.19", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.19"),
```
DrakeSeifert, line 138: @machi1990 please take a close look at this line. I do not believe this workaround is needed any longer. `GetInstallVersionForZStreamUpgrade` now calculates the latest z-stream candidate that has a valid upgrade path to the next minor version. If Cincinnati's graph is correct, versions like 4.19.3 (with known broken upgrade paths) will not have edges leading to them, so the function will skip them. If Cincinnati does include edges to a broken version then the test should fail, which is the expected outcome because it means customers could land on that broken version too.

test/util/framework/version_helper.go
```diff
@@ -118,27 +120,59 @@ func GetInstallVersionForZStreamUpgrade(ctx context.Context, channelGroup string
...
+				return "", false, err
+			}
+			if len(path) > 0 {
+				targetIdx = i
+				break
+			}
+			if i == len(candidates)-1 {
+				return "", false, fmt.Errorf("no candidates in %s contain an upgrade path to %s", configuredVersionID, latestInNextMinor)
+			}
+		}
```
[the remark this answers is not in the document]
DrakeSeifert: fixed

test/e2e/control_plane_automated_z_stream_upgrade.go
DrakeSeifert: @machi1990 Thanks for the detailed response! I think I have a much better understanding of how versioning works now. I have updated the logic in version_helper.go to determine the initial install version to more closely align with what you are asking.

test/util/verifiers/hosted_control_plane_version.go
```diff
@@ -18,77 +18,147 @@ import (
...
 		if historyVersion.Major != initialSemver.Major || historyVersion.Minor != initialSemver.Minor {
-			return fmt.Errorf("version %q in clusterversion history has different major.minor than initial %q (expected %d.%d.x)",
-				historyVersion.String(), v.initialVersion, initialSemver.Major, initialSemver.Minor)
+			return gomega.StopTrying(
+				fmt.Sprintf("clusterversion history contains at least one upgrade that altered the major or minor version. Expected: %d.%d.z. Actual: %s",
+					initialSemver.Major,
+					initialSemver.Minor,
+					history.Version,
+				),
+			).Wrap(err)
```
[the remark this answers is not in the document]
DrakeSeifert: fixed

test/util/verifiers/hosted_control_plane_version.go
```diff
@@ -18,77 +18,145 @@ import (
...
+	if clusterVersion.Status.History[0].Version != v.targetVersion {
+		return fmt.Errorf("cluster did not upgrade to the correct target. Expected: %s. Actual: %s",
+			v.targetVersion,
+			clusterVersion.Status.History[0].Version,
+		)
+	}
 
-	var sawUpgrade bool
-	for _, history := range clusterVersion.Status.History {
+	// note that clusterVersion.Status.History has the latest version first and the install version last
```
machi1990: > 	// note that clusterVersion.Status.History has the latest version first and the install version last

not necessarily as you can downgrade a version too; so it doesn't mean that the latest will be first. It means that the most the first in the slice are the ones that are recent applied regardless of how fresh the version is in the release rank

test/util/verifiers/hosted_control_plane_version.go
```diff
@@ -18,77 +18,145 @@ import (
...
+
+	if clusterVersion.Status.History[0].Version != v.targetVersion {
+		return fmt.Errorf("cluster did not upgrade to the correct target. Expected: %s. Actual: %s",
+			v.targetVersion,
+			clusterVersion.Status.History[0].Version,
+		)
+	}
 
-	var sawUpgrade bool
-	for _, history := range clusterVersion.Status.History {
```
machi1990, line 64: What's the motive behind these changes?

DrakeSeifert: Hi @machi1990, everything here is simply meant to improve the test. We now fail fast from unrecoverable errors, have more clear error logs, and instead of checking that each version in the history array upgraded relative to the _install_ version, we now check that each version in the history upgraded relative to _the previous version in the history array_, ensuring no downgrades occurred mid-update. This checks for edge cases that the previous implementation wouldn't catch, such as:
`4.20.0` -> `4.20.2` -> `4.20.1`

We no longer have to account for `sawUpgrade` explicitly because this behavior is now implicitly accounted for when asserting that each version in history upgraded relative to the last.

test/util/verifiers/hosted_control_plane_version.go
```diff
@@ -18,77 +18,145 @@ import (
...
 
-	ginkgo.GinkgoLogr.Info("Retrieved openshift cluster version history",
-		"history", framework.SummarizeClusterVersionHistory(clusterVersion.Status.History))
+	ginkgo.GinkgoLogr.Info(
+		"Retrieved openshift cluster version history",
+		"history",
+		framework.SummarizeClusterVersionHistory(clusterVersion.Status.History),
+	)
+
+	if len(clusterVersion.Status.History) < 1 {
```
DrakeSeifert: @machi1990 I just read [your comment in external-wg-aro-hcp](https://redhat-external.slack.com/archives/C075PHEFZKQ/p1786006955885019). I may have ran into some bad timing here, because I expected there to be 2 items (the initial install version and the target version), but during testing only saw the single version in history (the target version only). I'm assuming then that the check here should instead be `len(clusterVersion.Status.History) < 2`?

test/util/verifiers/hosted_control_plane_version.go
DrakeSeifert: Would this be sufficient?
> // note that clusterVersion.Status.History should have the latest version first and the install version last

test/util/verifiers/hosted_control_plane_version.go
machi1990: As we are testing upgrade, then the .status.history should've 2 versions for this check to suceed or said otherwise, the install version should also be in the list.

Indeed this improvement landed in an unfortunate timing and I am hoping we can get tracking on the thread I started on how to correct test upgrades, otherwise as mentioned in the thread, "we are testing install succeeds to latest" which isn't the intent of this test.

test/util/verifiers/hosted_control_plane_version.go
machi1990: Okay thanks

test/util/verifiers/hosted_control_plane_version.go
DrakeSeifert: fixed

test/util/verifiers/hosted_control_plane_version.go
DrakeSeifert: Okay, seeing as the e2e test already wasn't checking this, I've decided to remove this check so that this PR is not blocked. I've created a separate branch locally with this change so that we can easily merge it once the bug is fixed.

DrakeSeifert:
@machi1990 This PR is ready to merge, can you please remove the hold and approve?

test/util/verifiers/hosted_control_plane_version.go
machi1990, line 1: As written right now, this doesn't check that an actual upgrade has been performed;
i.e from initialVersion -> targetVersion.
We need to fix that (see the discussion I started and linked you at in slack about how that can be fixed).

test/util/verifiers/hosted_control_plane_version.go
```diff
@@ -18,77 +18,139 @@ import (
...
 	return nil
 }
 
-// VerifyHostedControlPlaneZStreamUpgradeOnly returns a verifier that the HCP control plane has
-// performed only a z-stream upgrade from the initial version: at least one entry in
-// ClusterVersion status.history is greater than initialVersion, and every entry has the same
-// major.minor as initialVersion.
-func VerifyHostedControlPlaneZStreamUpgradeOnly(initialVersion string) HostedClusterVerifier {
-	return verifyHostedControlPlaneZStreamUpgradeOnly{initialVersion: initialVersion}
+func updateHistoryToString(updateHistory []configv1.UpdateHistory) string {
```
machi1990: Why are you re-inventing the wheel while you've `framework.SummarizeClusterVersionHistory()` util?

test/util/verifiers/hosted_control_plane_version.go
```diff
@@ -18,77 +18,139 @@ import (
...
+			prevHistoryVersion, err := semver.ParseTolerant(clusterVersion.Status.History[i-1].Version)
+			if err != nil {
+				return gomega.StopTrying(
+					fmt.Sprintf("error parsing clusterVersion.Status.History[i-1].Version: %q",
+						clusterVersion.Status.History[i-1].Version,
+					),
+				).Wrap(err)
+			}
+
+			// versions at earlier indices must be larger than versions at later indices
```
machi1990: This comment is wrong; a cluster can be rolledback to the previous version to the versions in earlier indices can technically be smaller than those in later indices.

test/util/verifiers/hosted_control_plane_version.go
DrakeSeifert: Agreed, this needs fixing. I had originally tried to guard this with a length check:
```go
if len(clusterVersion.Status.History) < 2 { ... }
```
but pulled it because the install-version bug you brought up in the Slack thread was (rightly) tripping it, and I didn't want to block this PR on that.

The tightest fix for what you're describing is to add a check that `History[len-1] == initialVersion`. That directly asserts that the upgrade occurred and makes any check on the length of the history array redundant. David's [approach in #6538]((https://github.com/Azure/ARO-HCP/pull/6538/changes#diff-feddfe5e6df70b2d177e137da4a018b10feb7db882d6d8a8713f4d93b03caa0eR95:~:text=//%20The%20automated%20z,%7D)) proves an upgrade happened, but on its own it wouldn't confirm we started from initialVersion specifically, so I'd lean toward verifying the initial and target versions explicitly, or combining the two approaches.

It seems at this point that his PR takes priority over mine. I'm juggling a few things at the moment but I'll get to the rest of David's PR as soon as I can.

test/util/verifiers/hosted_control_plane_version.go
DrakeSeifert: Good call; that slipped my mind since this code is still new to me. I'll switch to `SummarizeClusterVersionHistory()` and drop `updateHistoryToString()`.

test/util/verifiers/hosted_control_plane_version.go
DrakeSeifert: My intent was to document what we expect to hold if this test passes, not to describe what the history array can technically contain. I'll reword it to make that distinction clear.

The test setup uses the Cincinnati upgrade paths to place the cluster on a stable install version with a direct upgrade path to the target version. Given that setup, a rollback would be unexpected behavior, and if one occurred I'd argue the test should fail.

I'll update the comment to something like:
```go
// Given this test's setup, versions at earlier indices should be larger than versions at later indices. A rollback here is unexpected behavior.
```

test/util/verifiers/hosted_control_plane_version.go
DrakeSeifert: fixed (though this means that this PR is blocked until the install version bug is fixed)

test/util/verifiers/hosted_control_plane_version.go
DrakeSeifert: fixed

test/util/verifiers/hosted_control_plane_version.go
DrakeSeifert: fixed

test/util/verifiers/hosted_control_plane_version.go
```diff
@@ -33,62 +34,114 @@ import (
...
+			// Given this test's setup, versions at earlier indices should be larger than versions at later indices. A rollback here is unexpected behavior.
+			if prevHistoryVersion.LTE(historyVersion) {
+				return gomega.StopTrying(
+					fmt.Sprintf(
+						"downgrade unexpected: cluster went from version %q and unexpectedly downgraded to version %q. clusterVersion history: %v. Error when version at index %d downgraded to version at index %d",
+						history.Version,
+						prevHistoryVersion.String(),
+						framework.SummarizeClusterVersionHistory(clusterVersion.Status.History),
+						i,
+						i-1,
```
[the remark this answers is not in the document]
DrakeSeifert: fixed

DrakeSeifert:
Update: I expect some intense merge conflicts I will have to resolve for this PR. I do not have the bandwidth to focus on this right this second, but hoping to get to it early next week.

DrakeSeifert:
PR #6538 fundamentally changed the design for the upgrade process. Closing this PR.

PR #6246, opened 2026-07-23
feat(hcpctl): capture client-side azure.log in snapshots

roivaz:
## Why

Each e2e test's `azure.log` — the client-side azcore request/response/retry log the test framework writes per test to `${ARTIFACT_DIR}/<test>/azure.log` — records correlation IDs, client request IDs, retries, and LRO polling for every Azure SDK call the test makes. That client-side view was decisive in [AROSLSRE-1569](https://redhat.atlassian.net/browse/AROSLSRE-1569), where it reconstructed a 503→retry→409 `DeploymentActive` race. Until now `hcpctl snapshot` never captured it, so analysts had to fetch it by hand.

Jira: https://redhat.atlassian.net/browse/AROSLSRE-1569 (motivating RCA) / [AROSLSRE-1571](https://redhat.atlassian.net/browse/AROSLSRE-1571) (this enhancement)

## What

- `hcpctl snapshot from-prow-job` now downloads the per-test `azure.log` from the Prow job's GCS artifacts (same per-test prefix as node console logs), writes it to `azure_sdk_log/azure.log`, and indexes it in `manifest.json` under `azure_log` (with an `artifact_url`). Absence is non-fatal — not every test enables Azure SDK logging.
- `hcpctl snapshot analyze` is made aware of the artifact via the agent's data-directory layout (`pkg/agent/prompts/system.md`), so the client-side Azure SDK timeline is available for automated root-cause analysis (retries, racing/duplicate/conflicting deployments).
- Added unit tests for the writer and directory-layout entry.

Deliberately scoped: this makes the log **available** to analysis; it does not add a new structured proof-item type (like `node_console_log`), which would be a much larger cross-cutting change.

## Testing

- `go build ./...`, `go vet`, `gofmt -l` clean
- `go test ./pkg/snapshot/... ./cmd/snapshot/...` passes (new `azure_log_test.go`)

### DIFF
--- a/tooling/hcpctl/pkg/agent/prompts/system.md
+++ b/tooling/hcpctl/pkg/agent/prompts/system.md
@@ -210,6 +210,16 @@ resource. Review the manifest's `directory_layout` for the full structure. Key l
   `ignitionNotReached`, or `OSProvisioningTimedOut`, **always check these first** —
   they show the node's view of boot (ignition TLS errors, kubelet failures, CNI
   issues) that the orchestrator-side Kusto queries cannot reveal.
+- `azure_sdk_log/azure.log` — the client's own record of every Azure SDK call the
+  test makes, across all resource providers (ARO-HCP RP, Managed Identity, Key
+  Vault, Network, etc.): each request, response, error, retry, and LRO poll, with
+  correlation IDs and timing. Present when `manifest.azure_log` is set. It is the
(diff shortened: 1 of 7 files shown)


raelga (approved):
/lgtm
/approve

stevekuznetsov (approved):

PR #6247, opened 2026-07-23
fix(test): force-delete orphaned managed RGs in reaper (AROSLSRE-1591)

raelga:
Jira: [AROSLSRE-1591](https://redhat.atlassian.net/browse/AROSLSRE-1591)

### What

Make the standard e2e resource-group cleanup workflow force-delete an *orphaned* managed resource group instead of waiting for it forever. `cleanupResourceGroup` (`test/util/framework/per_test_framework.go`) already deletes HCP clusters via the RP and then waits up to 10 minutes for any managed RG to disappear. If that wait times out but the HCP clusters are already gone, the remaining managed RGs are now deleted directly (with `force`, mirroring the `no-rp` workflow) and cleanup proceeds to delete the parent RG.

### Why

When an HCP cluster is deleted but its managed RG lingers (its `managedBy` still references the now-deleted cluster), nothing ever deletes that managed RG. The 10-minute wait always times out and the reaper returns an error, so the parent customer RG is **never** deleted — both the parent and the managed RG leak forever.

This is live on INT (`64f0619f`): six overdue `idms-*` RGs (`deleteAfter` = 2026-03-21, ~4 months overdue) are stuck exactly this way, each with an orphaned `*-managed` RG still holding a DNS zone, load balancer, 10 public IPs and a storage account.

```text
DeleteAllHCPClusters            -> none (cluster already gone)
findManagedResourceGroups       -> idms-*-managed (managedBy -> deleted cluster)
waitForManagedResourceGroupsDeletion(10m) -> never deleted -> timeout -> error
=> parent RG never deleted (deadlock, repeats every reaper run)
```

### Testing

`go build ./...` and `go vet ./util/framework/...` pass in `test/`. The in-progress-deletion path is preserved: managed RGs that are genuinely mid-delete are still waited on first; only RGs that remain after the timeout (clusters already gone) are force-deleted.

The force-delete loop reuses the exact pattern already present in `cleanupResourceGroupNoRP` (`DeleteResourceGroup(..., force=true, ...)` guarded by `isIgnorableResourceGroupCleanupError`). Out of scope: the separate "stuck cluster resource still present" failure mode (INT `arm64-vm-cluster-*`, `cni-cilium-*`) where the RP delete itself hangs.

### PR Checklist

- [x] Changes build and vet cleanly
- [x] Linked Jira issue
- [x] Preserves existing in-progress-deletion behavior

### DIFF
--- a/test/util/framework/per_test_framework.go
+++ b/test/util/framework/per_test_framework.go
@@ -523,9 +523,10 @@ func (tc *perItOrDescribeTestContext) waitForManagedResourceGroupsDeletion(ctx c
 }
 
 // cleanupResourceGroup is the standard resourcegroup cleanup.  It attempts to
-// 1. delete all HCP clusters via the RP and wait for success
-// 2. wait for any managed resource groups to be deleted
-// 3. delete the resource group
+//  1. delete all HCP clusters via the RP and wait for success
+//  2. wait for any managed resource groups to be deleted, force deleting any that
+//     remain orphaned after the HCP clusters are already gone
+//  3. delete the resource group
 func (tc *perItOrDescribeTestContext) cleanupResourceGroup(ctx context.Context, resourceGroupName string, timeout time.Duration) error {
 	startTime := time.Now()
 	defer func() {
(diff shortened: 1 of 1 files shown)


raelga:
Pushed `e936742` (was `fa12296`): gofmt-only fixup. `ci/prow/lint` (gci) and `ci/prow/verify` (goimports/verify-deepcopy) both failed on `test/util/framework/per_test_framework.go:526` — editing the numbered-list doc comment on `cleanupResourceGroup` triggered gofmt's list-item reformatting (`// 1.` -> `//  1.` with aligned continuation). Ran `gofmt -w` on the file; no logic change. `gofmt -l` is now clean and the package builds.

raelga:
Closing — this approach is invalid for INT/stage/prod.

The fix force-deletes the orphaned `*-managed` resource group (mirroring the `no-rp` workflow), but in RP-backed environments that RG carries an RP-created, **system-protected deny assignment** that the e2e/SRE identity cannot bypass — even with PIM Owner.

Evidence (INT `64f0619f`, `idms-2n6xb2-managed`):

- Deny assignment `1849b7ca-36bc-5694-a2fa-24f70add0858`, `isSystemProtected: true`, scope = the managed RG, denies:
  - `Microsoft.Resources/subscriptions/resourceGroups/delete`
  - `Microsoft.Resources/subscriptions/resourceGroups/write`
  - `Microsoft.Resources/deployments/{delete,write}`
- `excludePrincipals` is **only** the RP / first-party service principals (`c9f3ba47…`, `b3b9d226…`, `7f66f342…`, `95d3daa0…`, `3d07d28d…`, `b0902de0…`). The reaper identity is not among them, so `DeleteResourceGroup(managedRG, force=true)` returns an authorization/deny error.
- System-protected deny assignments can't be removed by users; they only clear when the owning HCP cluster is deleted through the RP.

The `no-rp` force-delete only works in personal/dev envs where there is no RP (hence no deny assignment). Copying it into the standard (RP-backed) workflow doesn't hold.

Corrected root cause: these are **orphaned managed RGs** — the HCP cluster ARM resource was removed but its managed RG + deny assignment were left behind, so only the RP's own SPs can reclaim them. The parent customer RG (`idms-2n6xb2`) has zero deny assignments and is deletable, but the costly resources (LB, 10 public IPs, storage, DNS zone) live in the locked managed RG.

Tracking in [AROSLSRE-1591](https://redhat.atlassian.net/browse/AROSLSRE-1591); the real remediation is RP/HyperShift-side orphan cleanup, not a test-harness change.

PR #6252, opened 2026-07-23
new credential API

deads2k:


api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster.tsp
```diff
@@ -23,11 +25,23 @@ interface HcpOpenShiftClusters {
   listBySubscription is ArmListBySubscription<HcpOpenShiftCluster>;
 
   /** Request a temporary admin kubeconfig for the cluster */
-  requestAdminCredential is ArmResourceActionAsync<
+  @removed(Versions.v2026_06_30_preview)
+  @renamedFrom(Versions.v2026_06_30_preview, "requestAdminCredential")
+  noRequestModelRequestAdminCredential is ArmResourceActionAsync<
```
bennerv, line 30: I think we can utilize the `@sharedRoute` decorator on the new action to allow keeping the same request path, but having validation that requires the body in a newer API version.  

Unless we want to remove support for not passing a CSR outright.

deads2k: > I think we can utilize the `@sharedRoute` decorator on the new action to allow keeping the same request path, but having validation that requires the body in a newer API version.
> 
> Unless we want to remove support for not passing a CSR outright.

We will eventually. We'll have some overlap, but latency and security targets are at risk with the existing API.

backend/pkg/controllers/systemadmincredentialcontrollers/desire_deletion.go
```diff
@@ -0,0 +1,140 @@
...
+		return false, utils.TrackError(fmt.Errorf("get ApplyDesire %s: %w", desireName, err))
+	}
+
+	// If the desire is still a ServerSideApply, flip it to a Delete so the
+	// kube-applier tears down the applied object. TargetItem already names what
+	// to delete; the ServerSideApply payload is cleared.
+	if applyDesire.Spec.Type != kubeapplier.ApplyDesireTypeDelete {
+		applyDesire.Spec.Type = kubeapplier.ApplyDesireTypeDelete
+		applyDesire.Spec.ServerSideApply = nil
+		if _, err := applyCRUD.Replace(ctx, applyDesire, nil); err != nil && !database.IsNotFoundError(err) {
```
tony-schndr, line 126: I think `applyDesire.Status.Conditions = nil` is also needed here to distinguish between a "successful apply" from a "successful delete".

1. Apply succeeds → Successful=True written by kube-applier
2. Backend flips Type to Delete, writes back with stale Successful=True
3. If kube-applier hasn't reconciled yet, next backend sync sees Type=Delete + Successful=True → deletes the desire document
4. Kube object is never actually deleted from the management cluster

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster.tsp
bennerv: It may be useful to continue supporting the method that doesn't pass a CSR in the older API versions to not introduce a breaking change to clients that exist today (specifically capz).  Then we can deprecate it at the API version we introduce it at.

PR #6253, opened 2026-07-23
fix(frontend): return 429 instead of 500 for transaction precondition failures

deads2k:
When a Cosmos DB transactional batch step fails with 412 Precondition Failed (etag mismatch), the error was falling through to the default case in writeError() and being reported as a 500 Internal Server Error. This changes the mapping to return 429 Too Many Requests with Retry-After: 1s so callers know the operation is retryable.

Also exports TransactionStepError so callers can errors.As into it, includes CosmosDBTransactionStepDetails in the error message for debuggability, and logs the failing step's ResourceBody at error level before returning.

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -51,8 +51,8 @@ func isResponseError(err error, statusCode int) bool {
 	if errors.As(err, &responseError) && responseError.StatusCode == statusCode {
 		return true
 	}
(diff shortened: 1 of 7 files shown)


stevekuznetsov (approved):
/lgtm
/approve

miguelsorianod:
lgtm

PR #6263, opened 2026-07-24
feat: Add operation-level metrics to kube-applier

garrickdabbs:
## Summary

Implements operation-level metrics for kube-applier desires (apply, delete, resync), labeled by the **GroupVersionResource of the target Kubernetes object** the applier acts on. This gives outcome-level visibility as kube-applier replaces Maestro for resource management.

The applier is agnostic to the ARM ownership hierarchy (cluster/nodepool); reconciliation outcomes diverge by GVK, not by ARM parent. RBAC, CRD availability, and admission-webhook results are all properties of the target GVR, so that is the dimension that makes the metrics actionable and lets outcomes aggregate by kind across clusters.

### Metrics
- `kube-applier/pkg/app/desire_operation_metrics.go`
  - Two Prometheus metrics:
    - `kube_applier_desire_operations_total{operation, group, resource, status, reason}` — Counter
    - `kube_applier_desire_operation_last_timestamp_seconds{operation, group, resource}` — Gauge
  - `group`/`resource` are read from `desire.Spec.TargetItem` (the target's GVR). Core-group objects render as `group=""`. This replaces the earlier `resource_type` (ARM parent) label.
  - Operation types:
    - **apply**: ServerSideApply
    - **delete**: ApplyDesire with `Type=Delete`
    - **resync**: ReadDesire
  - `status`/`reason` come from the desire's `Successful` condition (`True`/`False`/`Unknown`, with reasons such as `NoErrors`, `PreCheckFailed`, `KubeAPIError`).
- `kube-applier/pkg/app/desire_metrics.go`
  - The `desireCollector` records operation metrics for each desire on every collection cycle.

### Grafana dashboard
- `observability/grafana-dashboards/kube-applier/kube-applier.json` — "Kube-Applier Operations", 3 panels:
  - Desire operation rate by type (apply/delete/resync), broken down by target GVR
  - Desire operation error rate
  - Operations by target resource (GVR) and status
- `observability/observability.yaml` — registers the **Kube Applier** dashboard folder so the dashboard is provisioned.

## Screenshots

Screenshot From 2026-08-13 11-40-43

## Non-breaking changes

The new metrics use distinct names and label sets from the existing `kube_applier_desires{type, condition}` gauge, so the only alert using existing metrics (`KubeApplierDesiresMetricNotPresent`) is unaffected.

## Testing

- [x] Builds: `cd kube-applier && go build ./...`
- [x] Unit tests: `cd kube-applier && go test ./...`
- [x] Verified live in a personal-dev environment: real GVR series flow from the management-cluster leader (e.g. `hostedclusters`, `controlplanecomponents`, `nodepools` under `group="hypershift.openshift.io"`, and core `secrets`/`configmaps` with `group=""`), including a `status="False"`/`reason="PreCheckFailed"` error series rendered by the dashboard.

## Related

Jira: [ARO-26754](https://redhat.atlassian.net/browse/ARO-26754)


garrickdabbs:
✅ Fixed in c76cf0094

You're correct - resync operations were being overcounted. The collector runs every 30s and was recording a new operation for every stable ReadDesire even when `InstanceVersion` hadn't changed.

**Change**: `recordReadDesireOperation()` now checks `lastProcessedGeneration` before recording. Only counts actual state changes from Cosmos, not every collection tick.

**Before**: Stable ReadDesire → +1 resync every 30s indefinitely
**After**: Stable ReadDesire → +1 resync only when `InstanceVersion` changes

garrickdabbs:
✅ **All GitHub Copilot comments addressed**

### Comment 1: ReadDesire overcounting (line 116)
**Fixed in**: c76cf0094  
**Issue**: Resync operations recorded every 30s even when InstanceVersion unchanged  
**Solution**: Check `lastProcessedGeneration` before recording

### Comment 2: ApplyDesire overcounting (line 91)  
**Fixed in**: aa986a9b8  
**Issue**: Create/update operations recorded every 30s even when InstanceVersion unchanged  
**Solution**: Check `lastProcessedGeneration` before recording, matching ReadDesire logic


---

Both overcounting issues had the same root cause: the collector runs every 30s and was recording operations on every tick regardless of whether the Cosmos document changed. Now operations are only counted when `InstanceVersion` changes, providing accurate metrics.

kube-applier/pkg/app/desire_metrics.go
```diff
@@ -64,17 +66,35 @@ func (c *desireCollector) Run(ctx context.Context) {
 func (c *desireCollector) collect() {
 	counts := initCounts()
 
+	// Track current desires for cleanup
+	currentDesires := make(map[string]bool)
```
stevekuznetsov: use k8s util sets

kube-applier/pkg/app/desire_operation_metrics.go
```diff
@@ -0,0 +1,206 @@
...
+
+// Operation type constants for metrics labels
+const (
+	OperationCreate = "create"
+	OperationUpdate = "update"
+	OperationDelete = "delete"
+	OperationResync = "resync"
+)
+
+// Resource type constants for metrics labels
```
stevekuznetsov: seems like something we can derive from the object's GVK dynamically

kube-applier/pkg/app/desire_operation_metrics.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-HCP/internal/api/kubeapplier"
+)
+
+// Operation type constants for metrics labels
+const (
```
stevekuznetsov, line 27: these don't match k8s API operations

kube-applier/pkg/app/desire_operation_metrics.go
```diff
@@ -0,0 +1,206 @@
...
+				Help: "Unix timestamp of the last completed operation, by operation type and resource type",
+			},
+			[]string{"operation", "resource_type"},
+		),
+		lastProcessedGeneration: make(map[string]int64),
+	}
+}
+
+// recordApplyDesireOperation records metrics for an ApplyDesire operation.
+// It determines the operation type (create/update/delete) based on the desire's
```
stevekuznetsov: we intentionally do not distinguish between these in the api, so why do we want it in the metric?

kube-applier/pkg/app/desire_operation_metrics.go
```diff
@@ -0,0 +1,206 @@
...
+	}).Inc()
+
+	// Record last operation timestamp
+	m.operationLastTimestamp.With(prometheus.Labels{
+		"operation":     operation,
+		"resource_type": resourceType,
+	}).SetToCurrentTime()
+}
+
+// extractResourceType parses the Cosmos resourceID to determine if this is a
```
stevekuznetsov: the azure sdk has a method for this

kube-applier/pkg/app/desire_operation_metrics.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// ServerSideApply: distinguish create vs update based on whether we've seen this desire before
+	if _, seen := lastProcessed[desire.ResourceID.String()]; seen {
+		return OperationUpdate
+	}
+
+	return OperationCreate
+}
+
+// cleanupStaleEntries removes entries from lastProcessedGeneration for desires
```
stevekuznetsov: you should just blow away the entire metric on relist since you have the state of the world

kube-applier/pkg/app/desire_operation_metrics.go
```diff
@@ -0,0 +1,206 @@
...
+	defer m.mu.Unlock()
+
+	for resourceIDStr := range m.lastProcessedGeneration {
+		if !currentDesires[resourceIDStr] {
+			delete(m.lastProcessedGeneration, resourceIDStr)
+		}
+	}
+}
+
+// isDesireSuccessful checks if the "Successful" condition is True.
```
stevekuznetsov: prefer to just expose the condition as a label

kube-applier/pkg/app/desire_operation_metrics.go
garrickdabbs: Could you clarify how to derive cluster vs nodepool from the object's GVK? I'm using the ARM ResourceID path to determine if the desire belongs to a cluster or  nodepool resource. I think the k8s object GVK would tell us what's being applied (Deployment, ConfigMap, etc.) but not which cluster/nodepool owns it. Or,  am I misunderstanding the suggestion?

kube-applier/pkg/app/desire_metrics.go
garrickdabbs: Changed to use sets.Set[string] from k8s.io/apimachinery/pkg/util/sets

kube-applier/pkg/app/desire_operation_metrics.go
garrickdabbs: Changed operation types to match actual desire operations: apply, delete, resync. Removed artificial create vs update distinction since the k8s API operation is always ServerSideApply.

kube-applier/pkg/app/desire_operation_metrics.go
garrickdabbs: Removed operation distinction logic. Now directly using the desire type: apply for ServerSideApply, delete for Delete, resync for ReadDesire.

kube-applier/pkg/app/desire_operation_metrics.go
garrickdabbs: Changed to use azcorearm.ParseResourceID() and ResourceType methods to properly parse the ARM resource path instead of string matching.

kube-applier/pkg/app/desire_operation_metrics.go
garrickdabbs: Removed cleanupStaleEntries. Now the metrics are recorded on every collection cycle directly from the current state of desires, without tracking state changes between cycles.

kube-applier/pkg/app/desire_operation_metrics.go
garrickdabbs: Changed to expose the Successful condition status and reason as metric labels instead of using a successful=true/false label. The metric now has status (True/False/Unknown) and reason labels for better observability.

kube-applier/pkg/app/desire_operation_metrics.go
garrickdabbs: hi @stevekuznetsov , does that make sense?

garrickdabbs: It reads the desire's Azure ARM ResourceID and parses it. The resource type is encoded in the ID's path, so cluster vs nodepool falls out of the parsed ResourceType.

garrickdabbs: Swapped resource_type for group/resource (the target's GVR) in the latest commit, so outcomes now break down by the kind we actually act on and aggregate  across clusters.

kube-applier/pkg/app/desire_operation_metrics.go
```diff
@@ -0,0 +1,113 @@
...
+package app
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-HCP/internal/api/kubeapplier"
+)
```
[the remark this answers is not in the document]
garrickdabbs: False positive — the import path `github.com/Azure/ARO-HCP/internal/api/kubeapplier` does exist; there is no `internal/api/kubeapplierapi` package in this repo. The module compiles cleanly (`cd kube-applier && go build ./...`), and CI builds confirm the same. Resolving.

stevekuznetsov (approved):
/lgtm
/approve

PR #6265, opened 2026-07-24
fix: make ExpiringWatcher send cancellation-safe

deads2k:
Guard the channel send in ExpiringWatcher with a select on ctx.Done() to prevent goroutine leaks when the watcher's context is cancelled before the expiration timer fires.


stevekuznetsov (approved):
/lgtm
/approve

two possible follow-ups:
- don't `time.After()` in the chan recv so we don't leak timers
- log `ctx.Err()` to be a good citizen for people who add `CancelCause()`

deads2k:
added the followup here
/label lgtm

stevekuznetsov (commented):
/lgtm

PR #6269, opened 2026-07-24
feat: add ClusterDenyAssignment controller for Azure deny assignments

deads2k:
Add a backend controller that creates Azure deny assignments on the managed resource group before cluster-service create and deletes them during cluster deletion. The controller tracks state via PendingDenyAssignments and DenyAssignments on ServiceProviderClusterStatus.

The deny assignment permissions catalog is ported from clusters-service, covering 19 resource-provider-specific deny assignments that enforce least-privilege access for OpenShift operator managed identities.

Key behaviors:
- Gates ClusterClusterServiceCreate on all deny assignments being created
- Blocks ServiceProviderCluster deletion while deny assignments remain
- Uses ensureDenyAssignment (read-compare-update) for idempotency
- Handles already-deleted assignments gracefully (404 → no-op)

/hold

cluster service needs fixes before we can do this.


backend/pkg/controllers/denyassignments/deny_assignment_definitions.go
```diff
@@ -0,0 +1,234 @@
...
+
+func allDenyAssignmentReferences(cluster *api.HCPOpenShiftCluster) ([]api.DenyAssignmentReference, error) {
+	defs := denyAssignmentDefinitions(cluster)
+	clusterARMResourceID := strings.ToLower(cluster.ID.String())
+	subscriptionID := cluster.ID.SubscriptionID
+	managedResourceGroup := cluster.CustomerProperties.Platform.ManagedResourceGroup
+
+	denyAssignmentReferences := make([]api.DenyAssignmentReference, 0, len(defs))
+	for _, d := range defs {
+		daUUID := generateDenyAssignmentUUID(clusterARMResourceID, d.denyAssignmentType)
```
deads2k: Critically, this name is "wrong" for matching the cluster-service expectation because cluster-service uses its own (generated?) ID to build this hash.  Cluster-service also appears to not track which ones are created in postgres.

Possible solutions
1. cluster-service stores the deny assignments it created, so we can mirror old records and provide the actual deny assignments we create during cluster-service create.  An example schema exists in serviceprovidercluster
2. cluster-service allows the create call to specify the ID.  I bet the API fights this super hard, but it might be worth it to have an extra field we can provide that specifies this during create.  Even a property value.

backend/pkg/controllers/cluster/creation/cluster_cluster_service_create_controller.go
```diff
@@ -169,6 +177,24 @@ func (c *clusterClusterServiceCreateSyncer) createPreconditionDesiredVersionReso
...
+
+	if len(serviceProviderCluster.Status.PendingDenyAssignments) == 0 && len(serviceProviderCluster.Status.DenyAssignments) > 0 {
+		return true, nil
+	}
+	pendingTypes := make([]string, 0, len(serviceProviderCluster.Status.PendingDenyAssignments))
+	for _, denyAssignmentReference := range serviceProviderCluster.Status.PendingDenyAssignments {
+		pendingTypes = append(pendingTypes, denyAssignmentReference.DenyAssignmentType)
+	}
+	logger.Info("Deny assignments not yet created, waiting for ClusterDenyAssignment controller",
+		"pendingDenyAssignmentTypes", pendingTypes)
```
JakobGray, line 207: How useful is it to log the types we're waiting on? Could we also just give a number?

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller_test.go
```diff
@@ -135,6 +135,7 @@ func TestClusterClusterServiceCreate_SyncOnce(t *testing.T) {
 			dbCluster:   newTestCluster(),
 			existingServiceProviderCluster: newTestSPC(func(spc *api.ServiceProviderCluster) {
 				spc.Spec.ControlPlaneVersion.DesiredVersion = desiredVersion
+				spc.Status.DenyAssignments = []api.DenyAssignmentReference{{DenyAssignmentType: "resources-deny-assignment", DenyAssignmentResourceID: api.Must(azcorearm.ParseResourceID("/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testManagedResourceGroup/providers/Microsoft.Authorization/denyAssignments/00000000-0000-0000-0000-000000000001"))}}
```
JakobGray: I suggest adding a new test where desiredVersion is set but deny assignments are still pending

backend/pkg/controllers/cluster/denyassignments/deny_assignment_definitions.go
```diff
@@ -0,0 +1,234 @@
...
+package denyassignments
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+)
+
+const (
```
bennerv, line 25: we're moving to a common single definition now that we have support for more excluded principals (Massimo can confirm if it's rolled out completely or not).

We should update this too only leverage a single deny assignment instead of multiple and hold the PR on massimo's confirmation it's rolled out to all regions we're deployed in.  

PR should just lay down a single deny assignment which looks like: https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/pkg/azure/denyassignmentcreator/deny_assignment_creator.go#L245-L303.  We're missing a few permissions from there, so we'll see the following additional permissions shortly: 
```
 		"Microsoft.Network/dnszones/CAA/write",
		"Microsoft.Network/dnszones/TXT/write",
		"Microsoft.Network/dnszones/TXT/delete",
		"Microsoft.Compute/virtualMachines/retrieveBootDiagnosticsData/action",
```

and the removal of `Microsoft.Authorization/roleAssignments/read` since read permissions are already excluded.

backend/pkg/controllers/cluster/denyassignments/deny_assignment_permissions.go
```diff
@@ -0,0 +1,324 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package denyassignments
+
+func resourcesActions() []string {
```
JakobGray, line 17: Are deny assignments ever updated? It seems that once they are created they are set and the controller takes no further action.

backend/pkg/controllers/cluster/denyassignments/deny_assignment_definitions.go
bennerv: If instead we want 1:1 parity between CS and what's here, we need to introduce a deny assignment similar to what is mentioned above ^ since that one doesn't appear in this PR.  

Today CS lays down all the ones mentioned + the one above if the excluded principal limit is large enough.

backend/pkg/controllers/denyassignments/deny_assignment_definitions.go
miguelsorianod: > Critically, this name is "wrong" for matching the cluster-service expectation because cluster-service uses its own (generated?) ID to build this hash. Cluster-service also appears to not track which ones are created in postgres.

About this:

CS generates the denyassignment IDs by generating a UUIDv5 with a namespace that CS defines for them.

The format of the UUIDv5 in that case is:
namespace UUID: `f75040b8-d8aa-4311-bda6-ba8af06db258`
input string: `<ocm_cs_cluster_id>-<denyassignmentsuffix>` where <denyassignmentsuffix> is a  suffix unique per denyassignment type. For example, "compute-deny-assignment". <ocm_cs_cluster_id> as you said is a generated id by CS associated to the cluster.

The <ocm_cs_cluster_id> ID is a random KSUID (https://github.com/segmentio/ksuid) in base32 with 32 lower characters whose alphabet we restrict to `0123456789abcdefghijklmnopqrstuv`.

miguelsorianod: > cluster-service allows the create call to specify the ID. I bet the API fights this super hard, but it might be worth it to have an extra field we can provide that specifies this during create. Even a property value.

About this, by "ID" do you refer to the denyassignment ID or to the CS Cluster ID? If it's the latter:

As of now now we prevent the ability to provide the CS cluster id to CS during cluster creation. We could however relax that. In that case we could move 1:1 the generation logic of that ID in the RP, storing it in Cosmos and passing it to CS during creation.

I think that would indeed make other migrations easier, as we have N places in CS where the identifier of a resource is derived or contains the name of the CS Cluster ID.

The remaining concern at that point would collisions of the ID in CS side. If we keep 1:1 generation of the CSID in the RP they are unlikely because of KSUID being used. Additionally on CS side we have the ID of the cluster as the primary key of the Clusters table at db level so if the worst happened the creation would fail because of DB conflict. We could add an additional db query check during creation but it would add more load to the DB and it's already covered by the primary key.

At that point we also need to think how we would transition and what to do with existing entries at Cosmos level.

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,529 @@
...
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if cluster.ServiceProviderProperties.DeletionTimestamp != nil {
+		return c.syncDeletion(ctx, key, cluster)
+	}
+
+	if cluster.ServiceProviderProperties.ClusterServiceID != nil &&
```
miguelsorianod: Is this part needed? wouldn't c.creationNeedsWork cover it? the only difference is the intermediate database.GetOrCreateServiceProviderCluster call

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,529 @@
...
+
+func (c *clusterDenyAssignmentSyncer) creationNeedsWork(cluster *api.HCPOpenShiftCluster, serviceProviderCluster *api.ServiceProviderCluster) bool {
+	if cluster.ServiceProviderProperties.DeletionTimestamp != nil {
+		return false
+	}
+	if cluster.ServiceProviderProperties.ClusterServiceID != nil &&
+		len(cluster.ServiceProviderProperties.ClusterServiceID.String()) > 0 {
+		return false
+	}
+	if len(serviceProviderCluster.Status.PendingDenyAssignments) == 0 && len(serviceProviderCluster.Status.DenyAssignments) > 0 {
```
miguelsorianod: This means that once created if for some reason they are deleted or the definition changed we would not reconcile back. Is that intentional?

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,529 @@
...
+) error {
+	poller, err := client.BeginDeleteByID(ctx, resourceID.String(), denyAssignmentAzureAPIVersion, nil)
+	if isResourceNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("BeginDeleteByID failed: %w", err))
+	}
+
+	_, err = poller.PollUntilDone(ctx, nil)
```
miguelsorianod, line 505: Do we want to block here or somehow signal "reenqueue and retry after a while" ?

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,529 @@
...
+			"IsSystemProtected": true,
+		},
+	}
+
+	poller, err := genericResourcesClient.BeginCreateOrUpdateByID(ctx, resourceID.String(), denyAssignmentAzureAPIVersion, resource, nil)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("BeginCreateOrUpdateByID failed: %w", err))
+	}
+
+	_, err = poller.PollUntilDone(ctx, nil)
```
miguelsorianod, line 408: as a note, if this ends up in a failed terminal state we would be erroring and retrying indefinitely

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,529 @@
...
+		logger.Info("Ensured deny assignment", "denyAssignmentType", denyAssignmentReference.DenyAssignmentType, "resourceID", denyAssignmentReference.DenyAssignmentResourceID.String())
+		ensured = append(ensured, denyAssignmentReference)
+	}
+
+	if len(ensured) > 0 {
+		for _, denyAssignmentReference := range ensured {
+			serviceProviderCluster.Status.DenyAssignments = append(serviceProviderCluster.Status.DenyAssignments, denyAssignmentReference)
+			serviceProviderCluster.Status.PendingDenyAssignments = removeDenyAssignmentRef(serviceProviderCluster.Status.PendingDenyAssignments, denyAssignmentReference.DenyAssignmentType)
+		}
+		_, err = serviceProviderClusterCRUD.Replace(ctx, serviceProviderCluster, nil)
```
miguelsorianod: an alternative is to deepcopy and do semantic deepequal like in other controllers, to only replace when they differ

backend/pkg/controllers/cluster/denyassignments/deny_assignment_definitions.go
machi1990: > the one above if the excluded principal limit is large enough.

+1; we should create the complete one as well to align with CS

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
machi1990, line 1: The controller also needs to delete deny assignments that should no longer be there;

This is for the long run where we are able to lay down a single deny assignment that contains everything and drop the sharded ones. 

See https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/pkg/azure/denyassignmentcreator/deny_assignment_creator.go#L204 for how CS lays it down

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
machi1990: +1 we need to handle full reconciliation:
- the deny assignment can be updated or a definition deleted.
We need to act in accordance to update or delete the definition.

CS already handles for it because there have been use cases where we needed to update the deny assignment or delete it.

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,529 @@
...
+func (c *clusterDenyAssignmentSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	cluster, err := c.clusterLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if cluster.ServiceProviderProperties.DeletionTimestamp != nil {
```
machi1990, line 98: There is no need to issue an arm request to delete a deny assingement when the cluster is being deleted.
That's because the MRG is going to be deleted and the corresponding deny assignment will be deleted in cascade.

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,529 @@
...
+		time.Minute,
+		syncer,
+	)
+}
+
+func (c *clusterDenyAssignmentSyncer) deletionNeedsWork(serviceProviderCluster *api.ServiceProviderCluster) bool {
+	return len(serviceProviderCluster.Status.DenyAssignments) > 0 || len(serviceProviderCluster.Status.PendingDenyAssignments) > 0
+}
+
+func (c *clusterDenyAssignmentSyncer) creationNeedsWork(cluster *api.HCPOpenShiftCluster, serviceProviderCluster *api.ServiceProviderCluster) bool {
```
machi1990: There is no point in doing the work if:
- The MRG doesn't exist
- or the principals ids collected here https://github.com/deads2k/ARO-HCP/blob/4ddb416b6d93b38a61890b264e90492d3db75551/backend/pkg/controllers/denyassignments/deny_assignment_controller.go#L467 are not there

Let's add at least the second part in the `creationNeedsWork`  check.

machi1990: The controller can only be run in environment where deny assignment can be created i.e stage & prod. Lower envs, this controller needs to be disabled

backend/pkg/app/backend.go
```diff
@@ -869,6 +876,7 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
 				go dispatchRequestCredentialController.Run(ctx, 20)
 				go dispatchRevokeCredentialsController.Run(ctx, 20)
 				go clusterClusterServiceCreateController.Run(ctx, 20)
+				go clusterDenyAssignmentController.Run(ctx, 20)
```
machi1990: See https://github.com/Azure/ARO-HCP/pull/6269#discussion_r3656385168

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,529 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: My understanding is that deny assignments can only be created in environments where a real FPA exists. On environments where it doesn't exist (like dev, integration) they can't be created because only a real FPA can create them. This would impact the execution of this controller (as well as preconditions . What signals to backend whether a real FPA is being used or not is the CLI flag `--insecure-ignore-user-azure-managed-identities-that-need-managed-identities-dataplane-available-and-use-mock` (backend/cmd/root.go).

I don't recall the details on whether attempting to create a denyassignment fails with an error, fails silently or simply creates the resource with no effect. If we are interested on this detail we should find it out.

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,529 @@
...
+		ClusterDenyAssignmentControllerName,
+		resourcesDBClient,
+		backendInformers,
+		nil,
+		time.Minute,
+		syncer,
+	)
+}
+
+func (c *clusterDenyAssignmentSyncer) deletionNeedsWork(serviceProviderCluster *api.ServiceProviderCluster) bool {
```
miguelsorianod: Here I think we should not proceed with deletion if CSID is still set:
```
if cluster.ServiceProviderProperties.ClusterServiceID != nil && len(cluster.ServiceProviderProperties.ClusterServiceID.String()) > 0 {
	return false
}
```

CS has an ordered and coordinated delete process. For example, we don't delete the managed resource group until the deny assignments within it have been deleted. Or another example being we don't delete the managed resource group until the cluster has been fully deleted from the management cluster side, because there are resources within the managed resourcegroup that were created as part of the creation of the cluster in the management cluster side

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
miguelsorianod: I give detail about that here: https://github.com/Azure/ARO-HCP/pull/6269#discussion_r3656664383

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
deads2k: working on this now.

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
deads2k: This exceeds my azure knowledge.  Are these fast or so slow?

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
deads2k: > as a note, if this ends up in a failed terminal state we would be erroring and retrying indefinitely

I bet it's an azure client gap on my part again. What is supposed to happen?

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
deads2k: updating for this now.

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
deads2k: > * or the principals ids collected here https://github.com/deads2k/ARO-HCP/blob/4ddb416b6d93b38a61890b264e90492d3db75551/backend/pkg/controllers/denyassignments/deny_assignment_controller.go#L467 are not there
> 
> 
> Let's add at least the second part in the `creationNeedsWork` check.

It's a little weird to fanout and I bet it ages poorly as the list of definitions changes.  You sure or shall we not add to pending when there isn't an ID for it?

machi1990: During early cluster creation phases, it is expected for the principalID to not be there up until when they are reconciled;
If we don't gate the controller for their existence, when collecting them we'll return an error https://github.com/deads2k/ARO-HCP/blob/4ddb416b6d93b38a61890b264e90492d3db75551/backend/pkg/controllers/denyassignments/deny_assignment_controller.go#L467 causing undesired retries due to https://github.com/deads2k/ARO-HCP/blob/4ddb416b6d93b38a61890b264e90492d3db75551/backend/pkg/controllers/denyassignments/deny_assignment_controller.go#L189 

 > or shall we not add to pending when there isn't an ID for it?
 
 Is the idea that we sync but only add the deny assignment to the pending list and not return an error?  If so, that can work too.

deads2k: Ok, you've phrased it as, "don't have principleIDs for deny assignment", but do you actually mean: 

`cluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ControlPlaneOperators` and ``cluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.DataPlaneOperators` and `cluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ServiceManagedIdentity` have values?

deads2k:
working on testing, but I think I got the other feedback fixed

backend/pkg/controllers/cluster/denyassignments/deny_assignment_definitions.go
ziccardi: > we're moving to a common single definition now that we have support for more excluded principals (Massimo can confirm if it's rolled out completely or not).

Just checked. The ExcludedPrincipal Limit has been increased across all regions (see [this kusto query](https://dataexplorer.azure.com/clusters/hcp-prod-uk.uksouth/databases/HostedControlPlaneLogs?query=H4sIAAAAAAAAA72UsWoDMQyG9zyFuSUXaJw2YyFDoN06FPoEOlu9M%2FHJh2T3kpCHr65k6Nal2KP0ow%2FJP3%2BhkMjMIQ%2BSCjs8CLuV0edikYzcNkPOkzzvdoObthMnvy0nW06qzoM9qSbZOZBPs1jC3GyshwwdCLbNB%2FJXcPiWetG6ViO26%2Ftc%2BdVcbx7%2BQEKxoD2GGABBcjVwx7ZjuIZYd18H1gGBB4e0bF0PPGz3VtQOV%2BQI5ClxxbWx2Bm1VzhNWI0ayN7PrKgA1bhF7OLlIvuKSFeTKbnXtFBH%2FXderG5mHpDRxNQblyhDIDHNC9LlKBJ6GvVDf1LNjElleYBsnh7N61nHefTvHMiFCaIYxwgZvZHiHIp8lhgvjc7Hc0byhrFfAvJglly8qWocgcMVlVootxvTXe6ab6K1YCNKBQAA) )

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,594 @@
...
+	return identityResourceIDs, nil
+}
+
+func resolvePrincipalIDs(cluster *api.HCPOpenShiftCluster, identityResourceIDs []*azcorearm.ResourceID) ([]string, error) {
+	if cluster.Identity == nil {
+		return nil, fmt.Errorf("cluster has no identity configuration")
+	}
+
+	lookup := make(map[string]string, len(cluster.Identity.UserAssignedIdentities))
+	for resourceID, identity := range cluster.Identity.UserAssignedIdentities {
```
machi1990, line 582: The `cluster.Identity.UserAssignedIdentities` field only contain the control plane identities + the service managed identities; the data plane identities will be missing;

We need a way to track their principal ids somewhere in the RP so that we can use them;

or alternatively, retrieve them via a call to ARM using the credentials of the ServiceManagedIdentity (that's how CS does it https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/pkg/clusterprovisioner/acm/aro/fetch_and_persist_data_plane_managed_identities_client_ids_and_principal_ids_step.go)

backend/pkg/controllers/denyassignments/deny_assignment_controller.go
machi1990: Sorry for the confusion in the phrasing.

Concretely, I meant that we should only trigger the creation of the deny assignment when we know that the principal IDs of the operator + smi have been retrieved so that we are able to resolve the principalIDs of each operator identity (data plane and control plane) + the service manage identity in this piece of code https://github.com/deads2k/ARO-HCP/blob/41353414c6de01f1704ed83529dee8bb40c5caec/backend/pkg/controllers/denyassignments/deny_assignment_controller.go#L549 

This line https://github.com/deads2k/ARO-HCP/blob/41353414c6de01f1704ed83529dee8bb40c5caec/backend/pkg/controllers/denyassignments/deny_assignment_controller.go#L565 will return an error each all the time even early on during provisioning phases where the principalIDs have not been retrieved from Azure yet.

In practice it is okay, since eventually they'll be reconciled, but the error is likely to trigger the `BackendControllerRetryHotLoop` when a cluster is being created in Stage/Prod envs where we can create deny assignments

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
miguelsorianod: It depends on the resource you are dealing with.
I haven't played much with DenyAssignments so I don't know how fast/slow they become ready (Massimo can have the answer to that if needed), but some resources can take a considerable amount of time to get in the desired state. Think for example loadbalancers, virtual machines, ... . And we will have some of those.

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
miguelsorianod: When you create a resource in azure and it ends up in a terminal state it never moves from that state unless something modifies it. Think for example creating a VM. The VM can end up being in terminal state "ready" or in terminal state "failed". Once it's failed it will remain in failed state as it's a terminal state. Unless you for example delete it (which will change its state) or do some update on it that makes it change state, however I am not sure that all resources support recovering from "failed" :thinking:

backend/pkg/azure/azuremockclient/mock_clients.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: We can use mockgen like it's done for other azure clients

deads2k: > We can use mockgen like it's done for other azure clients

came out with ugly test results for pagers.

backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go
```diff
@@ -0,0 +1,594 @@
...
+		return err
+	}
+
+	if len(replacement.Status.AzureResources.DenyAssignments.PendingAzureResources) == 0 {
+		replacement.Status.AzureResources.DenyAssignments.EarliestRecheckTime = c.recheckTime()
+	} else {
+		// we're unlikely to reach here with PendingAzureResource, but if we do, running again is probably a good idea.
+		replacement.Status.AzureResources.DenyAssignments.EarliestRecheckTime = nil
+	}
+	_, _, err = replaceServiceProviderClusterIfChanged(ctx, serviceProviderClusterCRUD, serviceProviderCluster, replacement, nil)
```
miguelsorianod, line 262: Here if we end up returning an error we will have set the earliestretry already: https://github.com/Azure/ARO-HCP/pull/6269/changes#diff-f060ec6fb0ba2b018e4b0f10978336bed126fe6ffaf9285e1461ba52930110a4R244
shouldn't be earliestretry unset in that case?

deads2k:
replaced by https://github.com/Azure/ARO-HCP/pull/6680

PR #6270, opened 2026-07-24
[WIP] Add new DeploymentWatcher to the management cluster agent

rogbas:
Jira: [ARO-28661](https://redhat.atlassian.net/browse/ARO-28661)

**Summary**

  Adds a DeploymentWatcher to mgmt-agent that logs apps/v1 Deployment events with snapshotType=kubernetes, routing them to the kubernetesResourceSnapshots Kusto table via existing
  Fluent Bit config. This makes Deployment objects (e.g. HyperShift Operator, Control Plane Operator) available in Kusto for diagnostics. No Fluent Bit or schema changes needed.

  This is the first step toward incorporating additional management cluster data into CI artifacts and snapshot analysis — future PRs will query Kusto for these artifacts and attach
  them to e2e test runs.

**Test plan**

  - [x] Unit tests pass
  - [x] go vet and build pass
  - [ ] Verify Deployment events appear in kubernetesResourceSnapshots after deploying to dev

### DIFF
--- a/mgmt-agent/cmd/options.go
+++ b/mgmt-agent/cmd/options.go
@@ -99,6 +99,7 @@ type completedControllerOptions struct {
 	ksmCtrl                  *ksmhcp.KSMHCPController
 	resourceWatcher          *controller.ResourceWatcher
 	podWatcher               *controller.PodWatcher
+	deploymentWatcher        *controller.DeploymentWatcher
 	configMapWatcher         *controller.ConfigMapWatcher
 	kubeInformers            kubeinformers.SharedInformerFactory
 	ksmKubeInformers         kubeinformers.SharedInformerFactory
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
>Currently, key management cluster resources like HyperShift Operator and Control Plane Operator Deployments are not captured in Kusto, making it difficult to diagnose provisioning failures that involve these components. When filing upstream bugs, we often lack the deployment state and pod logs needed to provide actionable context.

/hold

We have the entire state of every `Pod`, as well as all Kubernetes `Events`, _and_ all pod logs - what are we missing specifically and in which cases would we have needed the `Deployment`? We should avoid trying to replicate every single possible object into Kusto, especially ones that change as often as a `Deployment`.

rogbas:
> We have the entire state of every `Pod`, as well as all Kubernetes `Events`, _and_ all pod logs - what are we missing specifically and in which cases would we have needed the `Deployment`? 


Fair point.  The main goal is to improve troubleshooting for issues related to HO, CPO, and other components running on management clusters, especially when interfacing with upstream teams.
  
Having deployments will fill some troubleshooting gaps:

- Operator version: Image tag/digest lives on the Deployment spec. If the operator is crash-looping or scaled to 0, there may be no pods to pull this from (primary short term use case).
- Rollout state: Replica counts and Progressing/Available conditions show whether the operator was healthy at the time of failure, while Pods only tell you about individual containers.
- Replica failures: The ReplicaFailure condition captures why pods can't be created (quota, webhook rejection), which may not surface as Pod objects at all.

rogbas:
> We should avoid trying to replicate every single possible object into Kusto, especially ones that change as often as a `Deployment`.

To address the volume concern, I just added namespace filtering to only watch `hypershift` and `ocm-*` namespaces.

rogbas:
To further reduce the volume being captured into Kusto, also made changes to this PR filtering only updates, specifically when replica counts, condition statuses, observed generation, or spec template are impacted. Ignoring condition timestamp churn, resourceVersion bumps, and managedFields changes.

PR #6271, opened 2026-07-24
fix: push v20260630preview timebomb deadlines to 2026-08-07

bennerv:
## Summary
- Extends the v20260630preview API deployment timebomb deadlines from 2026-07-31 to 2026-08-07 in two E2E tests (`cluster_create_private_ingress` and `kms_key_rotation`) to match the updated API rollout timeline.

## Test plan
- [ ] Verify CI passes with updated deadlines
- [ ] Confirm alignment with `cluster_fips_mode.go` which already uses the 2026-08-07 deadline

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6281, opened 2026-07-27
chore(config): bump clusters-service image digest

raelga:
## What

Bumps the pinned `clustersService` image digest from `sha256:1ca0a2b3…` (2026-07-22) to the current `latest` `sha256:55bcc0e8…` (2026-07-27), plus the regenerated dev rendered configs.

## Why

The old digest has been garbage-collected from `quay.io/app-sre/aro-hcp-clusters-service`. Every `e2e-parallel` run mirrors that image in the `aro-hcp-provision-environment` pre-step and dies fleet-wide with:

```
Error response from registry: failed to resolve sha256:1ca0a2b3…:
quay.io/app-sre/aro-hcp-clusters-service@sha256:1ca0a2b3…: not found
```

Confirmed on multiple open PRs (#6277, #6272, #6271) with the identical signature, so it is not a flake. The automated digest-updater PR (#6272) did not pick CS up. The new digest has the same source commit (`ed6a71e9`); it is just a fresh rebuild republished under `latest`.

## How

```
AZURE_TOKEN_CREDENTIALS=dev ./image-updater update --config config.yaml --tags --groups cs
make -C config materialize
```

## Testing

- `make -C config materialize` regenerated the 6 dev rendered configs; helmtest passed.
- Diff is the CS digest line only (config.yaml + 6 rendered files).

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1032,7 +1032,7 @@ defaults:
     image:
       registry: quay.io
       repository: app-sre/aro-hcp-clusters-service
-      digest: sha256:1ca0a2b39872d87b89505fc1ed549110b4cf09a7c79b9dd6351f0d8efab2d5e4 # ed6a71e9b466739c9edd6ba69b2c35ee9587a6bc (2026-07-22 08:06)
+      digest: sha256:55bcc0e8e5f1e455e18369f757fa750984f653ea17bac5e9b2ca6f402934165c # ed6a71e9b466739c9edd6ba69b2c35ee9587a6bc (2026-07-27 08:21)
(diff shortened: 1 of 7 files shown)


davidffrench:
Approved for a force merge since all pre-merge checks are failing due to the missing CS image.

raelga:
Provision step already mirrored and deployed the new digest `55bcc0e8` with no `not found` and no errors, so the fix is doing its job. `integration` is green, `e2e-parallel` is the only job left.

It has lgtm + approved now, so tide will merge it on its own once e2e goes green, we don't actually need the force merge. Giving e2e ~15 more min. If it stalls I'll override `e2e-parallel` instead of force merging.

PR #6282, opened 2026-07-27
feat: add Managed Resource Group creation and deletion controller

miguelsorianod:
Add ClusterManagedResourceGroupLifecycle to create and delete the cluster's managed resource group in Azure, tracking progress on ServiceProviderCluster via PendingManagedResourceGroupResourceID and ManagedResourceGroupResourceID.

Gate ServiceProviderCluster cleanup on MRG deletion completing.

The Managed Resource Group (MRG) is an Azure Resource Group associated to the ARO-HCP Cluster that is created and it is used by the ARO-HCP service (RP/CS/Hypershift/OCP, ...) to put Azure resources associated to the cluster within it.
In environments that are in the AME tenant, after the MRG is created a next step is that we create an Azure DenyAssignment that prevents the end-users from (easily) modifying/deleting the Managed Resource Group nor its contents within it directly. In lower environments Microsoft's DenyAssignment integration is not available so no prevention occurs.

Other relevant notes:
* MRG names (and RG names in general) are unique within a Subscription (Azure does not allow creating two with the same name within a Subscription), even across azure regions
* A MRG can be used by one and only one ARO-HCP Cluster
* Once the ARO-HCP Cluster is deleted, one of the deprovisioning steps is the deletion of the MRG, which only occurs after other deprovisioning steps are performed
* The MRG must not previously exist
* MRG must be located in the same Azure location as the cluster's location

### DIFF
--- a/backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
+++ b/backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
@@ -1,0 +1,471 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clusterprovisioningcontrollers
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/utils/ptr"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+
+	azureclient "github.com/Azure/ARO-HCP/backend/pkg/azure/client"
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/backend/pkg/informers"
+	"github.com/Azure/ARO-HCP/backend/pkg/listers"
+	"github.com/Azure/ARO-HCP/internal/api"
+	controllerutil "github.com/Azure/ARO-HCP/internal/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/ocm"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+const (
+	// ClusterManagedResourceGroupLifecycleControllerName is the name of the controller that ensures that the
+	// Cluster's Managed Resource Group is reconciled/deleted.
+	ClusterManagedResourceGroupLifecycleControllerName = "ClusterManagedResourceGroupLifecycle"
+)
+
+// clusterManagedResourceGroupLifecycleSyncer is a Cluster syncer that ensures that the
+// Cluster's Managed Resource Group is provisioned.
+type clusterManagedResourceGroupLifecycleSyncer struct {
+	cooldownChecker controllerutil.CooldownChecker
+
+	resourcesDBClient            database.ResourcesDBClient
+	subscriptionLister           listers.SubscriptionLister
+	clusterLister                listers.ClusterLister
+	serviceProviderClusterLister listers.ServiceProviderClusterLister
+
+	azureFPAClientBuilder azureclient.FirstPartyApplicationClientBuilder
+}
+
+var _ controllerutils.ClusterSyncer = (*clusterManagedResourceGroupLifecycleSyncer)(nil)
+
+func NewClusterManagedResourceGroupLifecycleController(
+	resourcesDBClient database.ResourcesDBClient,
+	clustersServiceClient ocm.ClusterServiceClientSpec,
+	activeOperationLister listers.ActiveOperationLister,
+	backendInformers informers.BackendInformers,
+	azureFPAClientBuilder azureclient.FirstPartyApplicationClientBuilder,
+) controllerutils.Controller {
+	_, clusterLister := backendInformers.Clusters()
+	_, subscriptionLister := backendInformers.Subscriptions()
+	_, serviceProviderClusterLister := backendInformers.ServiceProviderClusters()
+
+	syncer := &clusterManagedResourceGroupLifecycleSyncer{
+		cooldownChecker:              controllerutils.DefaultActiveOperationPrioritizingCooldown(activeOperationLister),
+		resourcesDBClient:            resourcesDBClient,
+		subscriptionLister:           subscriptionLister,
+		clusterLister:                clusterLister,
+		serviceProviderClusterLister: serviceProviderClusterLister,
+		azureFPAClientBuilder:        azureFPAClientBuilder,
+	}
+
+	return controllerutils.NewClusterWatchingController(
+		ClusterManagedResourceGroupLifecycleControllerName,
+		resourcesDBClient,
+		backendInformers,
+		nil,
+		time.Minute,
+		syncer,
+	)
+}
+
+func (c *clusterManagedResourceGroupLifecycleSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	cluster, err := c.clusterLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if database.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	if cluster.ServiceProviderProperties.DeletionTimestamp != nil {
+		err := c.syncDeletion(ctx, cluster)
+		if err != nil {
+			return utils.TrackError(err)
+		}
+		return nil
+	}
+
+	// TODO decide if this check is needed or not here. Within syncCreation we also end up performing the check as of now.
+	// if cluster.ServiceProviderProperties.ClusterServiceID != nil &&
+	// 	len(cluster.ServiceProviderProperties.ClusterServiceID.String()) > 0 {
+	// 	return nil
+	// }
+
+	err = c.syncCreation(ctx, cluster)
+	if err != nil {
+		return utils.TrackError(err)
+	}
+
+	return nil
+}
+
+func (c *clusterManagedResourceGroupLifecycleSyncer) creationNeedsWork(cluster *api.HCPOpenShiftCluster, serviceProviderCluster *api.ServiceProviderCluster) bool {
+	if cluster.ServiceProviderProperties.DeletionTimestamp != nil {
+		return false
+	}
+
+	if cluster.ServiceProviderProperties.ClusterServiceID != nil && len(cluster.ServiceProviderProperties.ClusterServiceID.String()) > 0 {
+		return false
+	}
+
+	if serviceProviderCluster.Status.AzureResources.ManagedResourceGroup.PendingAzureResource == nil && serviceProviderCluster.Status.AzureResources.ManagedResourceGroup.AzureResource != nil {
+		return false
+	}
+
+	return true
+}
+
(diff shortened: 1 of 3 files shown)


backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
```diff
@@ -0,0 +1,454 @@
...
+		return nil
+	}
+
+	// TODO decide if this check is needed or not here. Within syncCreation we also end up performing the check as of now.
+	// if cluster.ServiceProviderProperties.ClusterServiceID != nil &&
+	// 	len(cluster.ServiceProviderProperties.ClusterServiceID.String()) > 0 {
+	// 	return nil
+	// }
+
+	err = c.syncCreation(ctx, cluster)
```
machi1990, line 114: before this controller can be merged, we'll need to enhance this CS https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/cmd/clusters-service/service/azure/azure_managed_resource_group_existence_validation.go#L48 check to take onto consideration that the MRG can be pre-created by the RP; the inflight check needs to check the managed_by field as well and if it matches the same cluster, it shouldn't error

miguelsorianod: I already mentioned/documented this in https://github.com/Azure/ARO-HCP/pull/6282/changes#diff-a3862feb2ae0fd160652133d081cbe396e03c92476ec1b65fb1be80901cead9aR331

miguelsorianod: > the inflight check needs to check the managed_by field as well and if it matches the same cluster, it shouldn't error

the provisionstep in CS side already checks that, which then errors and retries indefinitely if they don't match

machi1990: >>the inflight check needs to check the managed_by field as well and if it matches the same cluster, it shouldn't error

> the provisionstep in CS side already checks that, which then errors and retries indefinitely if they don't match

Yes to that, but specifically this comment 

> the inflight check needs to check the managed_by field as well and if it matches the same cluster, it shouldn't error

was about  https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/cmd/clusters-service/service/azure/azure_managed_resource_group_existence_validation.go#L48 needs to  enhanced in a way that it doesn't blindly fails the inflight check; but only fail if the managed_by field is different. 
Or if we want to go rogue; remove that validation all together. But I prefer the enhancement.

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+	if err != nil && !azureclient.IsResourceGroupNotFoundErr(err) {
+		return utils.TrackError(fmt.Errorf("failed to get managed resource group: %w", err))
+	}
+
+	// TODO should we error or just log and consider success
+	if getResp.ManagedBy == nil || *getResp.ManagedBy != cluster.ID.String() {
+		logger.Info("managed resource group is not managed by the cluster. Skipping deletion", "managed_by", ptr.Deref(getResp.ManagedBy, "<null>"))
+		return nil
+	}
```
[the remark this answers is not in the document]
miguelsorianod: This is not correct. `getResp` **embeds** `ResourceGroup` so referencing `getResp.ManagedBy` is correct

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
```diff
@@ -0,0 +1,465 @@
...
+	logger.Info("delete request sent to Azure")
+
+	// TODO: do we want to poll until done or do we want to implement something that allows earlies requeue time?
+	logger.Info("polling until deletion ends up in terminal state")
+	_, err = poller.PollUntilDone(ctx, nil)
+	if err != nil {
+		// An error is returned here also if the LRO ended up in a terminal state failed or canceled.
+		return utils.TrackError(fmt.Errorf("failed to delete managed resource group: %w", err))
+	}
```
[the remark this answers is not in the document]
miguelsorianod: We are aware of this. It's written as a TODO comment to decide how do we want to deal with this.

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
```diff
@@ -0,0 +1,465 @@
...
+// including clusters in different regions. CS enforces the no reuse across clusters within the same subscription and region by leveraging its
+// database. However, it does not enforce preventing cross-region reuse within the same subscription because the database is regional. What do we want to
+// do about this? As a note about resource groups: a resource group is a regional resource, and Azure enforces that no two resource group
+// can have the same name if they belong to the same subscription, even across regions. Note that we also have the ManagedBy attribute of the ManagedResourceGroup which
+// includes the cluster's full resource id.
+func (c *clusterManagedResourceGroupLifecycleSyncer) ensureManagedResourceGroup(ctx context.Context, cluster *api.HCPOpenShiftCluster, mrgResourceID *azcorearm.ResourceID, resourceGroupsClient azureclient.ResourceGroupsClient) error {
+	logger := utils.LoggerFromContext(ctx)
+	logger = logger.WithValues("managed_resource_group_resource_id", mrgResourceID.String())
+	ctx = utils.ContextWithLogger(ctx, logger)
```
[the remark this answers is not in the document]
miguelsorianod: We are aware. I am going to add unit tests once the design is clear and agreed on

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
miguelsorianod: The code comment already mentions the inflight needs to be changed to allow pre-existing MRG. If it exists and it's a different managedby then cs provisionstep will fail. We are not marking the cluster as failed on CS side when it's owned by a different cluster. No need to change the inflight check to consider that, we don't want to change the current behavior around that (retry indefinitely)

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+
+	if serviceProviderCluster.Status.ManagedResourceGroupResourceID == nil && serviceProviderCluster.Status.PendingManagedResourceGroupResourceID == nil {
+		return false
+	}
+
+	return true
+}
+
+func (c *clusterManagedResourceGroupLifecycleSyncer) syncCreation(ctx context.Context, cluster *api.HCPOpenShiftCluster) error {
```
miguelsorianod, line 162: With the current design this only supports creation and not really updates as it's gated on the serviceprovidercluster properties being set.

As of now the attributes that we manage in the managedresourcegroup are not mutable so that's not an issue but that doesn't mean in the future that could change (for example if we want to end up setting some tags and we have existing ones)

TODO discuss what do we want to do.

bennerv: if rbac for 1p application falls off for some reason, we'll want to support update.  This isn't super relevant to our scenario right now as we don't support tenant migration, but just something to consider:
https://stackoverflow.microsoft.com/a/245391/62320

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
```diff
@@ -0,0 +1,465 @@
...
+	logger = logger.WithValues("managed_resource_group_resource_id", mrgResourceID.String())
+	ctx = utils.ContextWithLogger(ctx, logger)
+
+	// TODO if it exists, do we consider the work done or do we want to somehow
+	// check if existing differs from desired and apply changes? The implications
+	// of attempting changes are:
+	// * Some attributes in Azure are READ-ONLY an cannot be changed after
+	//   creation
+	// * If there are changes that we do not have in desired we would "undo" them
+	getResp, err := resourceGroupsClient.Get(ctx, mrgResourceID.Name, nil)
```
bennerv, line 363: If the controller falls over in the middle of the `Create` call to the managed resource group, we should be able to recover from it. 

What I would do is: 
- Get the existing resource group -> it exists
- Validate managedBy field is what we expect on it 
- Validate Location is what we expect on it
- CreateOrUpdate() managed resource group with the field information, this can't hurt us since it's idempotent, maintaining any tags the customer set if the resource group already exists.

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
```diff
@@ -0,0 +1,465 @@
...
+
+	// Azure does not allow updating the ManagedBy attribute from unset to set. If that is attempted the API
+	// returns an error (HTTP 400 with error code ResourceGroupManagedByMismatch) and
+	// message "The managed by property of the resource group cannot be changed from its current value '<current_value>'."
+	// We error if the existing resource group is not a managed resource group. This is, if the ManagedBy attribute is not set.
+	if existingManagedResourceGroup.ManagedBy == nil {
+		return utils.TrackError(fmt.Errorf("existing managed resource group is not a managed resource group (ManagedBy attribute is not set)"))
+	}
+
+	if desiredManagedResourceGroup.ManagedBy == nil {
```
bennerv, line 428: How would this ever be nil?  Wouldn't this be a failure in the `buildDesiredManagedResourceGroup` function?

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
```diff
@@ -0,0 +1,465 @@
...
+		return utils.TrackError(fmt.Errorf("existing managed resource group ProvisioningState attribute is not Succeeded"))
+	}
+
+	return nil
+}
+
+func (c *clusterManagedResourceGroupLifecycleSyncer) buildDesiredManagedResourceGroup(cluster *api.HCPOpenShiftCluster, desiredManagedResourceGroupName string) *armresources.ResourceGroup {
+	managedBy := cluster.ID.String()
+	return &armresources.ResourceGroup{
+		// TODO do we set the Name for clarity too? According to the Go data type it is
```
bennerv, line 463: typically read-only fields are ignored on persist.  I think it's fine to ignore it since we're passing it elsewhere.

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
miguelsorianod: It could be nil if an end-user creates a RG that happens to have the same rg name as the desired MRG

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
miguelsorianod: The controller already contains the logic for that, but it doesn't enable updates for now as mentioned in https://github.com/Azure/ARO-HCP/pull/6282#discussion_r3658164394. To discuss the specifics on how to enable updates in that thread (what do we gate etc...)

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
miguelsorianod: The usefulness of setting the ID here too is that you don't need to pass around both the mrg id and the resourcegroup around. You can just pass the desired mrg which contains it

backend/pkg/controllers/clusterprovisioningcontrollers/managed_resource_group_provisioning_controller.go
machi1990: Sorry, when I said fail in my mind was `that inflight returning an error`

raelga:
Live data point in case it helps prioritize this: checked the subscription directly and found 9 of the 12 real managed-resource-group-linked clusters currently have a deleted parent `HCPOpenShiftCluster` ARM resource (404) while their MRG is still present, 1 to 11 days later, holding ~95 stranded role assignments combined. That matches the gap this PR closes, `ObserveManagedResourceGroup` (#6278) watches for MRG deletion but never drives it (`// TODO: begin deletion of the managed resource group`), so a cluster that Cluster Service never finishes tearing down on its own leaves the MRG (and its RBAC) stuck indefinitely.

No urgency pressure intended, just flagging that this is an active, measurable leak in case it's useful context for when you get back to the rebase.

PR #6285, opened 2026-07-27
Add fields to represent Azure Resources related to a cluster

deads2k:
This API structure provides a common way we can guide our controllers to guide fault tolerant behavior for creation and deletion.  It also provides a field to promote consideration of when rechecking is appropriate.


internal/api/types_serviceprovider_cluster.go
```diff
@@ -197,6 +197,56 @@ type ServiceProviderClusterStatus struct {
...
+	EarliestRecheckTime *metav1.Time `json:"earliestRecheckTime,omitempty"`
+}
+
+// AzureReference tracks a single Azure resource through its creation lifecycle.
+// PendingAzureResources holds a resource ID that has been requested but not yet confirmed;
+// AzureResources holds a resource ID that has been confirmed to exist.
+type AzureReference struct {
+	// PendingAzureResources is the resource ID that has been requested but
+	// not yet confirmed to exist in Azure.
+	PendingAzureResources *azcorearm.ResourceID `json:"pendingAzureResources,omitempty"`
```
miguelsorianod: Is there some reason PendingAzureResources and AzureResources are plural here?

deads2k: > Is there some reason PendingAzureResources and AzureResources are plural here?

Oversight.  Will fix and fix the fuzz testing.  Other thoughts?

internal/api/types_serviceprovider_cluster.go
```diff
@@ -197,6 +197,56 @@ type ServiceProviderClusterStatus struct {
...
+	// AzureResources is the resource ID that has been confirmed to exist in Azure.
+	AzureResources *azcorearm.ResourceID `json:"azureResources,omitempty"`
+	// EarliestRecheckTime is the earliest time at which the controller should
+	// re-check the pending resources. Nil means recheck immediately.
+	// This allows for controllers to avoid repeatedly hitting an Azure API to recheck that the desired state is true.
+	// Controllers should set this field with substantial jitter: without another concern, jitter of 50% is considered normal
+	// so that any storms are quickly dissipated.
+	// Additionally, long recheck times are recommended for resources outside of their active phases. Order of at least
+	// six hours is, with durations up to 24 hours considered normal.
+	EarliestRecheckTime *metav1.Time `json:"earliestRecheckTime,omitempty"`
```
miguelsorianod, line 249: How do we deal with timezones here? how is it supposed to be stored, retrieved and interpreted?

internal/api/types_cluster.go
```diff
@@ -87,6 +87,8 @@ type HCPOpenShiftClusterCustomerProperties struct {
 type HCPOpenShiftClusterServiceProviderProperties struct {
 	// Written by: Frontend PUT/PATCH/DELETE Cluster, OperationClusterCreate, OperationClusterUpdate, OperationClusterDelete
 	ProvisioningState arm.ProvisioningState `json:"provisioningState,omitempty"`
+	// PendingClusterServiceID will be written for our future transition.
+	PendingClusterServiceID *InternalID `json:"pendingClusterServiceID,omitempty"`
```
miguelsorianod, line 91: I see it's an InternalID, which means the full CS href instead of just the generated name part. Writing this down here to make sure we understand that we need to calculate and store the full part

internal/api/types_serviceprovider_cluster.go
deads2k: > How do we deal with timezones here? how is it supposed to be stored, retrieved and interpreted?

metav1.Time serializes with timezone information, but at its core it's an instant.  I don't understand the concern.

internal/api/types_cluster.go
deads2k: I'm open to a change.  Which is actualy used to calculate?

deads2k: Actually, maybe a bias for consistency is better.

miguelsorianod: Yes, I agree. I am fine with keeping it as InternalID. It just means we need to put the HREF there, as with `ClusterServiceID`.

internal/api/types_serviceprovider_cluster.go
miguelsorianod: The question was more on how we serialize it. Should we do always `.UTC()` before? should that be documened?

When comparing the times I understand timezones shouldn't matter as when using Before, After, Equal, ... the internal instants are leveraged.

internal/api/types_serviceprovider_cluster.go
```diff
@@ -197,6 +197,56 @@ type ServiceProviderClusterStatus struct {
...
+	// AzureResource is the resource ID that has been confirmed to exist in Azure.
+	AzureResource *azcorearm.ResourceID `json:"azureResource,omitempty"`
+	// EarliestRecheckTime is the earliest time at which the controller should
+	// re-check the pending resources. Nil means recheck immediately.
+	// This allows for controllers to avoid repeatedly hitting an Azure API to recheck that the desired state is true.
+	// Controllers should set this field with substantial jitter: without another concern, jitter of 50% is considered normal
+	// so that any storms are quickly dissipated.
+	// Additionally, long recheck times are recommended for resources outside of their active phases. Order of at least
+	// six hours is, with durations up to 24 hours considered normal.
+	EarliestRecheckTime *metav1.Time `json:"earliestRecheckTime,omitempty"`
```
miguelsorianod, line 249: Is there some reason we favor metav1.Time over time.Time?

internal/api/types_serviceprovider_cluster.go
deads2k: We'll serialize it however metav1.Time does.  I'm pretty sure it's rfc339 down to milli or micro.

internal/api/types_serviceprovider_cluster.go
deads2k: > Is there some reason we favor metav1.Time over time.Time?

there were serialization edges about 12 years ago. I don't recall exaxctly

internal/api/testing/fuzz.go
```diff
@@ -0,0 +1,309 @@
...
+		b.WriteString(sanitizeFuzzName(genName()))
+	}
+
+	return api.Must(azcorearm.ParseResourceID(b.String()))
+}
+
+// FuzzInternalID constructs a lexically valid InternalID for a Cluster Service
+// cluster resource with a random name component.
+func FuzzInternalID(genName func() string) api.InternalID {
+	return api.Must(api.NewInternalID("/api/clusters_mgmt/v1/clusters/" + sanitizeFuzzName(genName())))
```
miguelsorianod, line 74: I don't think it impacts as it's unit tests but the real api endpoints that we use against CS use `/api/aro_hcp/v1alpha1/clusters/'` instead of `/api/clusters_mgmt/v1/clusters/`

internal/api/v20240610preview/conversion_fuzz_test.go
```diff
@@ -26,140 +26,59 @@ import (
...
-			// InstanceVersion and PartitionKey are purely storage-layer state;
-			// they do not round-trip through the external API.
-			j.InstanceVersion = 0
-			j.PartitionKey = ""
-		},
+	fuzzer := apitesting.FuzzerFor(append(apitesting.CommonRoundTripFuzzFuncs(),
+		// ImageDigestMirrors, Ingress, and CryptoRestrictions do not exist in v20240610preview.
 		func(j *api.HCPOpenShiftClusterCustomerProperties, c randfill.Continue) {
 			c.FillNoCustom(j)
-			// ImageDigestMirrors is a v20251223preview field that does not exist in v20240610preview.
```
miguelsorianod, line 56: Should these comments be kept for historical context?

miguelsorianod: The same comment applies to other code comments removals I observe

deads2k:
deferring the comments.  Unsticking our tomorrows.

/label lgtm

PR #6287, opened 2026-07-27
tooling/hcpctl: expose analysis failures for observability

stevekuznetsov:
We need to know post-hoc what failure(s) our agents are running into, so we can update the system to be smarter.

### DIFF
--- a/tooling/hcpctl/pkg/agent/analyze.go
+++ b/tooling/hcpctl/pkg/agent/analyze.go
@@ -204,18 +204,18 @@ func ValidateDraftLoop(
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
E2E flaked on a PodNotReady alert. This is crucially missing data for the analysis tooling, right now we are seeing 1h+ delays in analysis publication without it. Otherwise, this PR just adds logging to an ancillary tool, which is very safe. Going to force-merge to get this upstreamed to sdp-pipelines and deployed to the release dashboard. Tide only has a batch with this PR and another, which just failed, so I'm not kicking others out of the queue.

PR #6295, opened 2026-07-27
docs(ev2): add per-region rollout lock design (AROSLSRE-1567)

raelga:
[AROSLSRE-1567](https://redhat.atlassian.net/browse/AROSLSRE-1567)

### What

Adds a design doc, `docs/ev2-region-lock.md`, for a **per-region EV2 rollout lock** for production. It lets several prod rollouts run at once while guaranteeing that, in any region, an older commit always deploys before a newer one. The doc covers the ordering gate, the shell-step + `autoRestart` release mechanism, the per-region blob lease, the new global lock-store infrastructure, and the evolution path to step-level locking. Linked from `docs/README.md`.

### Why

Read-only Fridays and baking windows stretch a single prod rollout across days, so the current "wait for the previous rollout to finish" rule serializes prod to about one rollout per week. We want multiple daily prod rollouts completing unattended, but without a guard a newer commit could overtake an older one in a region. This design is the per-region guard that makes concurrent rollouts safe, and it records the decisions and discarded alternatives from the EV2 locking investigation so the implementation (in sdp-pipelines) has an agreed reference.

### Testing

Docs-only change. Validated all 8 embedded Mermaid diagrams parse with the Mermaid parser. No code or pipeline behavior changes.

The mechanism (gate logic, lease client, EV2 step injection) is implemented in sdp-pipelines where the EV2 rollout tooling lives; the only in-repo artifact besides this doc will be the lock-store storage account in `global-infra.bicep`, tracked separately. This PR is the design only.

### PR Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/docs/ev2-region-lock.md
+++ b/docs/ev2-region-lock.md
@@ -1,24 +1,47 @@
-# Per-region EV2 rollout lock
+# Per-region rollout ordering for incremental prod rollouts
 
-This document describes the design of a **per-region rollout lock** for ARO HCP
-production deployments. The lock lets several prod rollouts run at once while
-guaranteeing that, in any given region, an older commit always deploys before a
-newer one. No newer commit can overtake an older one in a region.
+This document describes how we let several ARO HCP production rollouts run at the
+same time while guaranteeing that, in any given region, an older change always
+deploys before a newer one. A newer change can never overtake an older one in a
+region.
 
-This is a design document. The mechanism is implemented in the
+The goal is not a new bespoke gate bolted onto rollouts. The goal is to make our
+existing **incremental rollouts** safe to run in prod, continuously and
+unattended, several times a day, without an SRE babysitting each one to completion.
+
+This is a design document. The ordering mechanism is implemented in the
 [sdp-pipelines](https://dev.azure.com/msazure/AzureRedHatOpenShift/_git/sdp-pipelines?path=/tooling)
-EV2 tooling and the lock-store infrastructure lives in this repository's
-[global service group](#lock-store-infrastructure). See
-[EV2 Deployment](ev2-deployment.md) for the base rollout flow this builds on.
+EV2 tooling, next to the incremental-rollout generator it builds on. See
+[EV2 Deployment](ev2-deployment.md) for the base rollout flow.
+
+## Contents
+
+- [Problem](#problem)
+- [Goal](#goal)
+- [Background: incremental rollouts and the missing signal](#background-incremental-rollouts-and-the-missing-signal)
+- [Design overview](#design-overview)
+- [Ordering key: commit lineage](#ordering-key-commit-lineage)
+- [The gate: a single immediate predecessor](#the-gate-a-single-immediate-predecessor)
+- [Failure and cancellation semantics](#failure-and-cancellation-semantics)
+- [Recovery: failed rollouts and hotfixes](#recovery-failed-rollouts-and-hotfixes)
+- [Intent and completion store](#intent-and-completion-store)
+- [Mutual exclusion: a TTL and fenced lease](#mutual-exclusion-a-ttl-and-fenced-lease)
+- [Release mechanism: bounded retries, no central config](#release-mechanism-bounded-retries-no-central-config)
+- [EV2 APIs and guarantees](#ev2-apis-and-guarantees)
+- [Environments](#environments)
+- [Where each piece lives](#where-each-piece-lives)
+- [Possible future: step-level ordering](#possible-future-step-level-ordering)
+- [Discarded alternatives](#discarded-alternatives)
+- [Open items](#open-items)
 
 ## Problem
 
-Prod rollouts are long. A single rollout walks every prod region in sequence,
-and each region can be held for a long time by validation baking. Two operational
+Prod rollouts are long. A single rollout walks every prod region in sequence, and
+each region can be held for a long time by validation baking. Two operational
 constraints stretch this out further:
 
-- **Read-only Fridays.** There is no ARO SL SRE on-call over the weekend, so prod
-  rollouts are frozen from Friday to Monday. A rollout that starts mid-week can sit
+- **Read-only weekends.** There is no ARO SL SRE on-call over the weekend, so prod
+  rollouts are frozen Friday to Monday. A rollout that starts mid-week can sit
   half-done across the weekend.
 - **Baking windows.** Once per-region baking lands, a region is intentionally held
   for hours to days before the rollout advances.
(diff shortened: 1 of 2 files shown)


docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+rollout to fully finish before starting the next one. With the constraints above,
+that serializes prod to roughly one rollout per week, which is too slow.
+
+We want to **start a new prod rollout while the previous one is still in flight**,
+possibly several per day, and let them all complete **unattended**. The risk this
+introduces is **ordering**: without a guard, rollout `N+1` could reach region R and
+deploy commit `N+1` there before rollout `N` has deployed commit `N` to R, leaving
+that region on a newer commit than an earlier one shipped. That must never happen.
+
+The failure we are preventing, in one picture:
```
stevekuznetsov, line 59: Also, critically, we build a lot of tooling around incremental rollouts, and we can't use them in prod because the notion of "did we successfully roll out before" does not apply. We see a lot of downsides to always doing a full rollout, namely time and Azure resources that do not like being disturbed often

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+    Note over R: R now on N+1
+    N->>R: deploy commit N
+    Note over R: R rolled BACK to N ❌ out of order
+```
+
+## Goal
+
+- A **per-region ordering lock** for prod: rollout `N+1` must not begin deploying a
+  region until every older in-flight rollout has finished that same region (or was
+  cancelled).
```
stevekuznetsov: cancellation would invalidate the future ones if they are incremental fwiw

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+  region, unattended.**
+- Ordering is an **absolute invariant.** If it cannot be honored, the region step
+  **fails and exits the pipeline** rather than deploying out of order.
+
+## Non-goals
+
+- **Not a pipeline-level lock.** Azure DevOps already serializes and skips pipelines.
+  This is a finer, per-region guard so region-by-region progress cannot overtake
+  itself.
+- **Not for int/stg.** Those environments are effectively single-region, so one
```
stevekuznetsov: the design seems to apply cleanly to those environments, though - why not use it?

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+  rollout at a time is enough and there is no ordering to enforce. **Prod only.**
+- **Cancelled rollouts never block.** A cancelled older rollout is not going to
```
stevekuznetsov: this is in direct tension with point 1.

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+  wait trigger (`manual` needs a human, `incidentResolution` needs an ICM). But the
+  setting it waits on has to be published through EV2's configuration service, which
+  is external to the rollout and gated behind internal EV2 docs.
+- **Health checks** (`restHealthCheck`, `mdmHealthCheck`) are post-deploy monitors,
+  capped under 24h. They can gate advancing *past* a region, not *starting* one.
+
+So cross-rollout, per-region mutual exclusion has to be built on top of EV2, using
+EV2's own rollout state as the source of truth.
+
+## Design overview
```
stevekuznetsov, line 119: I would recommend creating rollouts that presume regularity - eg, they know the last rollout that was scheduled, and assume it (and all its predecessors) will be successful. Then, generate the incremental diff between N-1 and N, schedule N. Fail if, at runtime, N finds that N-m...N-1 have finished and not succeeded in region $foo

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+flowchart TD
+    subgraph rollout_np1["Rollout N+1 (region R)"]
+        gate["Ordering gate<br/>region-lock check"]
+        lease["Acquire per-region<br/>blob lease"]
+        deploy["Deploy region R"]
+        release["Release lease"]
+    end
+    status["EV2 rollout state<br/>(RegionStatuses)"] -->|older rollouts<br/>landed R?| gate
+    gate -->|"yes: all older landed / cancelled"| lease
+    gate -->|"no, and an older is still going"| wait["Wait, re-check later"]
```
stevekuznetsov: if you want to support "an older is still going" you also need a blob mechanism to declare your future intent to roll out

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+EV2's own rollout state as the source of truth.
+
+## Design overview
+
+Three parts, layered on the base EV2 flow:
+
+1. An **ordering gate** injected at the head of every prod region. It blocks the
+   region until all older in-flight rollouts have landed that region, then lets the
+   region deploy. This is where ordering is enforced.
+2. A **per-region mutex** (a blob lease) taken around the actual deploy, so two
```
stevekuznetsov: can we put specific links to what API(s) you want to use and what their semantics are? what guarantees do we get?

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+    lease --> deploy --> release
+    store[("Lock store<br/>(global storage account)")] --- lease
+```
+
+## The ordering gate
+
+The gate is the brain. For a rollout deploying region R, it answers one question:
+**may I deploy R now?**
+
+Answer *yes* only when, for region R, **every rollout older than me is either done
```
stevekuznetsov: this is insufficiently precise - we need to know _which_ rollouts we're waiting on. what is "older"?

what happens when rollout N-2 fails, N-2 and N were in flight, they failed because their pre-requisite failed. so we do a fix and now we expect to go from N-4 -> N+1 with a new commit. how does N know to wait on N-2...N-1 while N+1 knows it doesn't need to concern itself with N having failed?

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+with R or cancelled.** Otherwise wait. If I have somehow already passed an older
+rollout that has not finished R, that is an invariant violation and the step fails.
+
+### Ordering key
+
+Rollouts are ordered by `Umbrella.BuildID`, the ADO build id, which is monotonic
```
stevekuznetsov: An errant process or mistaken SRE that triggers a rollout from an old commit on the prod branch would have a temporally-more-recent release with a logically-less-recent content. I don't think ADO build ID is what we want here

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+```
+
+The gate reads live rollout state through the existing `RegionStatuses()` primitive
+in the sdp-pipelines EV2 tooling (EV2 REST today; the same shape can be sourced from
+Kusto rollout step-metadata later). No new EV2 API is required.
+
+### Why this yields strict order for N concurrent rollouts
+
+Ordering falls out of the gate, not the lock. Each rollout's blocking set contains
+its **immediate predecessor**, so the fan-in of N concurrent rollouts chains into
```
stevekuznetsov: your `CanProceed` refers to rollout_s_, but here you specify one independent predecessor. you need to choose one ;)

notably, the approach here with one immediate predecessor is IMO the robust and correct mechanism, and it naturally lends itself to incremental diffs.

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+- The only native way to hold with *no process* for days is `wait` +
+  `configExists`, but publishing the release signal it waits on needs EV2's config
+  service, which is gated. See [Open items](#open-items).
+
+So the release mechanism does **not** depend on `configExists`. Instead the gate is
+a short, leading shell step that **polls and re-runs**:
+
+- The gate step runs `region-lock check`. If the predecessor has landed R it exits
+  `0` and the region deploys.
+- If it is not yet R's turn, the step exits non-zero. EV2 `autoRestart` re-runs it
```
stevekuznetsov: check the internal Ev2 codebase to make sure they don't have sneaky caps on these, I would not put it past them

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+  later: `waitDurationAfterFailure` is the poll interval, `maxRestartAttempts` is
+  set high (the schema has no cap), and `skipSucceeded` means only the gate re-runs,
+  not completed work. Between attempts there is **no process**, so the hold spans
+  hours to days for free.
+- Read-only Fridays already freeze all rollouts over the weekend (a `noRollout`
+  window), so the gate's *active* wait is realistically under a day; `autoRestart`
+  covers the tail.
+
+`wait` + `configExists` stays as a nicer future swap (a process-less native hold
```
stevekuznetsov: we're not going to want to have some other external process mucking with central config state (think of the races, the blast radius, etc) - there's a reason we don't use the ev2 central config server

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+```
+
+## Mutual exclusion: per-region blob lease
+
+The gate orders the *start* of a region. A short **per-region blob lease** protects
+the hand-off window so two rollouts can never deploy the same region at the same
+instant:
+
+- Blob path per region: `<serviceGroup>/<env>/<region>`.
+- The wrapped deploy step (short, within `maxExecutionTime`) acquires the lease,
```
stevekuznetsov: think through how we make sure that a step that acquires the lease, then crashes (or has the ACI underneath it fail, etc) does not deadlock the system

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+
+`check` is read-only and safe to run anywhere. `acquire`/`release` operate on the
+lock store.
+
+## Where each piece lives
+
+| Piece | Repository |
+| --- | --- |
+| This design doc | ARO HCP (`docs/`) |
+| Lock-store storage account + role (bicep) | ARO HCP (`global-infra.bicep`) |
```
stevekuznetsov: could also put into sdp-pipelines, dangle some step off the global using topology merging

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+| Gate logic (`region-lock check`) | sdp-pipelines (`tooling`) |
+| Blob-lease client (`acquire`/`release`) | sdp-pipelines (`tooling`) |
+| EV2 shell-step injection + `autoRestart` wiring | sdp-pipelines (pipeline generation) |
+
+## Evolution to step-level locking
```
stevekuznetsov: not very sure we want this

raelga:
/test integration

Retesting. This PR only adds `docs/ev2-region-lock.md` and a `docs/README.md` entry, so it cannot affect the integration suite.

Job [2082456037352280064](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6295/pull-ci-Azure-ARO-HCP-main-integration/2082456037352280064) never ran a test. It failed on the build cluster being unable to schedule the test pod:

```
Reporting job state 'failed' with reason
'executing_graph:step_failed:executing_multi_stage_test:pod_pending'

0/114 nodes are available: 13 node(s) didn't match Pod's node affinity/selector,
14 Insufficient memory, 2 node(s) were unschedulable, 40 node(s) had untolerated
taint(s), 59 Insufficient cpu.
```

That is CI build-cluster capacity, not the diff. The run is also from 2026-07-29, so it predates the current base by some way.

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,427 @@
...
+    else lease held by another rollout
+        L-->>N1: busy -> back off, retry
+    end
+```
+
+## Lock store infrastructure
+
+There is no existing global store to reuse — the global service group today deploys
+only a global managed identity, a global Key Vault, DNS, and an encryption key. The
+lock needs a small new resource:
```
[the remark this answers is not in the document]
raelga: @copilot That was a fair point and it is already fixed in the current head, the thread is just outdated. The doc now spells the scope out instead of implying the global service group is only MSI, key vault, DNS and the encryption key:

> For context on the global service group this would attach to: it deploys the global infrastructure, reconciles the global **Grafana** instance and its roles, and deploys the **SVC and OCP ACRs**, along with the shared MSI, key vault, DNS, and encryption key. It is not limited to MSI, key vault, DNS, and the encryption key.

That matches `dev-infrastructure/global-pipeline.yaml`, which under `serviceGroup: Microsoft.Azure.ARO.HCP.Global` carries the `infra`, `acrs`, `grafana-reconcile` and `grafana-roles` steps among others.

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,481 @@
...
+    gate -->|write own intent| store
+    gate -->|acquire before mutate| lease
+    deploy -->|on success write completion| store
+    classDef s fill:#eef,stroke:#88a
+    class store,lease s
+```
+
+Three pieces make this work:
+
+1. An **ordering key** derived from commit lineage, so "older" and "newer" are well
```
stevekuznetsov, line 146: If we have 2 (eg a well-known previous to check) do we need a total order?

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,481 @@
...
+   defined and cannot be faked by triggering a rollout from an old commit.
+2. A **single immediate predecessor** per rollout, so the gate has exactly one
+   thing to check per region and it lines up with the incremental diff.
+3. A **store** of intent and completion records that the gate reads and writes,
```
stevekuznetsov, line 150: If all we're checking is the previous, do we need to store more than the last one?

stevekuznetsov: I guess we do, if we show up to region R and see N-3 running, to know that we need to keep waiting.

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,481 @@
...
+
+The single rule that keeps this simple:
+
+> **Only unfinished lower-lineage rollouts block. A finished rollout, whether it
+> succeeded, failed, or was cancelled, never blocks anyone.**
+
+From that rule:
+
+- A **successful** predecessor lets its successor proceed in R.
+- A **finished-failed** predecessor is done, so it stops blocking. But because it
```
stevekuznetsov, line 237: Note that a rollout N that arrived at region R while N-3 was running needs to know that - if it happens to observe a state where N-3 finished but N-2 had not yet started, that the state it saw still means it needs to wait.

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,481 @@
...
+  did not succeed in R, its successor's incremental base for R is stale, so the
+  successor **fails fast** in R. The region simply stays on the last good change
+  until a fix arrives.
+- A **cancelled** rollout is finished, so it also stops blocking. Cancelling a
```
stevekuznetsov, line 241: A cancelled rollout would be exposed in the store how? Seems like a rollout that is:
- cancelled before starting R appears invisible to the store for R
- cancelled during a rollout to R appears as running with an expired lease
- cancelled after finishing R appears as a successful rollout to R

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,481 @@
...
+
+This is why cancellation "does not block" but still cannot be ignored: it unblocks
+the ordering immediately, and it forces a base recompute for anything generated on
+top of it.
+
+## Recovery: failed rollouts and hotfixes
+
+This is the case Steve asked us to design for: `N` fails, and we ship a hotfix.
+
+The foundation is already in the incremental generator: `previousDeployment()`
```
stevekuznetsov, line 254: The issue here is a heterogeneous step graph would be difficult to encode - if a step were necessary in R but not R' due to different incremental bases, we would need to generate some execution constraints, etc, to build an Ev2 step graph that did what we wanted. This seems tough. Maybe we just run a complete rollout in this case and allow it to clobber any prior record in regions.

docs/ev2-region-lock.md
```diff
@@ -0,0 +1,481 @@
...
+## Intent and completion store
+
+To support "an older rollout is still going, wait for it," a successor needs to
+know that an earlier rollout **exists and intends to reach this region**, before it
+has produced any result. A pure "did it succeed" record is not enough, because a
+predecessor that has not started R yet has no success record and no failure record.
+So the store holds two kinds of record, keyed by `(serviceGroup, environment,
+region)`:
+
+- **Intent.** "Rollout for change C intends to deploy region R." Written by the
```
stevekuznetsov, line 306: If our stores are region-scoped, how do we intend to achieve this? How does a regional outage in our store infra affect the rest of the rollout?

PR #6296, opened 2026-07-28
ci(provision): centralize CI provisioning scripts into ARO-HCP

gmfrasca:
[ARO-28195](https://issues.redhat.com/browse/ARO-28195)
Supercedes #5866

### What

Extracts CI provisioning logic from `openshift/release` step-registry scripts into shared scripts under `hack/ci/`, preparing for the step-registry wrappers to become thin shims that `exec` into these scripts.

New scripts:
- `az-login.sh` — shared Azure SP login (sourced by all others)
- `resolve-acr-images.sh` — ACR polling, git history walk-back, and digest resolution
- `provision-from-main.sh` — upgrade baseline: checks out `PULL_BASE_SHA`, resolves ACR images, runs provisioning
- `provision-hypershift.sh` — hypershift PR testing: fetches main, resolves ACR images, runs provisioning

Updated:
- `provision-environment.sh` — parameterized JUnit output and completion marker names, added MGMT sizing overrides, delegates Azure login to `az-login.sh`

### Why

The three provisioning step-registry scripts (`provision-environment`, `provision-from-main`, `hypershift-deploy`) had significant duplicated logic across `openshift/release`. Centralizing into ARO-HCP makes the logic easier to test, review, and evolve without cross-repo coordination for every change.

The companion `openshift/release` PR will reduce those step-registry scripts to ~17-line wrappers.

### Testing

These scripts are additive and not yet consumed until the companion `openshift/release` PR lands. The existing `provision-environment.sh` changes are backward-compatible with current callers.

- No unit/integration/E2E tests — these are CI pipeline scripts validated by rehearsal runs on the `openshift/release` PR.
- shellcheck passes on all five scripts.

- This PR must merge **before** the companion `openshift/release` PR, which rewrites the step-registry `commands.sh` files to `exec` into these scripts.
- Individual commits are not independently runnable (e.g., `az-login.sh` is created after scripts that source it), but the final tree state is correct.

### PR Checklist

- [x] Self-reviewed the diff


hack/ci/resolve-acr-images.sh
```diff
@@ -0,0 +1,109 @@
...
+    exit 1
+  fi
+  TARGET_SHA="${IMAGE_SHA}"
+fi
+
+# Resolve each image tag to its digest from ACR.
+resolve_digest() {
+  local repo=$1 tag=$2
+  local digest
+  digest=$(az acr manifest show-metadata -r "${ACR_NAME}" -n "${repo}:${tag}" --query 'digest' -o tsv)
```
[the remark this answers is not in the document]
gmfrasca: This was a pre-existing bug, but the fix is simple enough, implemented

hack/ci/resolve-acr-images.sh
```diff
@@ -0,0 +1,109 @@
...
+  fi
+  echo "Attempt ${attempt}/${MAX_POLL}: not yet available, retrying in ${POLL_INTERVAL}s ..."
+  sleep ${POLL_INTERVAL}
+done
+
+if [[ "${FOUND_HEAD}" != "true" ]]; then
+  echo "Images for ${TARGET_SHA} not found after polling. Walking back through history ..."
+  MAX_WALK=20
+  IMAGE_SHA=""
+  for sha in $(git log "${TARGET_SHA}" --format='%h' --abbrev=7 --skip=1 -n ${MAX_WALK}); do
```
roivaz: The current live `aro-hcp-hypershift-deploy-commands.sh` deliberately avoids relying on local git history for `main` — it resolves the tip via `git ls-remote` and walks back history via the GitHub REST API (`curl api.github.com/.../commits?sha=main`) instead of `git log`. `provision-hypershift.sh` only does a `git fetch ... main` before sourcing this script, and CI checkouts here are commonly shallow/scoped to the PR's own tree, so this local `git log "${TARGET_SHA}" --skip=1` walk-back could silently come up empty if the fetch doesn't bring in full ancestry. Worth a comment noting this assumption (full, non-shallow fetch of `main`'s history) so it gets scrutinized/validated in the rehearsal run, since the walk-back path isn't exercised by the happy-path case.

hack/ci/provision-environment.sh
```diff
@@ -123,6 +115,22 @@ else
   echo "No MSI mock SP lease provided, skipping mock SP overrides"
 fi
 
+# Temporary MGMT cluster sizing overrides for single-wave E2E parallelism.
+# These will be removed once the matching config.yaml defaults land in ARO-HCP.
+# Scale up when identity containers are leased (E2E runs), set minCount=1
+# for leased healthcheck workflows, skip entirely when no lease is held
+# (e.g. hypershift-deploy) to preserve the base config sizing.
+if [[ -n "${LEASED_MSI_CONTAINERS:-}" ]]; then
```
roivaz: This whole block doesn't match what's actually live today. The only MGMT sizing override that currently exists upstream is the single check in `aro-hcp-provision-environment-commands.sh` (https://github.com/openshift/release/blob/a322c1c804c2a2da31d34918de8b2c8a5e9e0956/ci-operator/step-registry/aro-hcp/provision/environment/aro-hcp-provision-environment-commands.sh#L137-L143):
```
if [[ -z "${LEASED_MSI_CONTAINERS:-}" ]]; then
  yq -i ".clouds.dev.environments.${DEPLOY_ENV}.defaults.mgmt.aks.userAgentPool.minCount = 1" "${OVERRIDE_CONFIG_FILE}"
fi
```
The `minCount=7`/`infraAgentPool.vmSize=Standard_D8ds_v6` override and the `elif LEASED_MSI_MOCK_SP` branch appear to be stale config that survived in some of the other scripts (e.g. `provision-from-main-commands.sh`) but aren't the current source of truth. Please drop this block down to just the single `LEASED_MSI_CONTAINERS`-gated `minCount=1` override to match what's actually live, rather than centralizing the stale variant.

hack/ci/resolve-acr-images.sh
```diff
@@ -0,0 +1,109 @@
...
+#   ACR_CONFIG_FILE — path to rendered config YAML (for ACR name + repo paths)
+#   TARGET_SHA      — 7-char git commit SHA to look up in ACR
+#
+# Callers must ensure TARGET_SHA is reachable in local git history
+# (via checkout or fetch) so the walk-back fallback can use git log.
+#
+# Exports: BACKEND_IMAGE, FRONTEND_IMAGE, ADMIN_API_IMAGE, SESSIONGATE_IMAGE,
+#          FLEET_IMAGE, MGMT_AGENT_IMAGE, KUBE_APPLIER_IMAGE
+#
+# hcpRecovery is not pushed to ACR by images-push, so it is excluded.
```
roivaz, line 12: `aro-hcp-images-push-commands.sh` does push a `customExporter`/`exporter` image to ACR by commit SHA (alongside backend/frontend/etc.), so it should probably be resolved here too — this script only excludes `hcpRecovery` (correctly, since that one truly isn't pushed). This gap in the exporter/hcpRecovery distinction was already caught and fixed in the upstream `openshift/release` scripts recently, so it looks like this PR was written against an earlier snapshot. Worth syncing this script with the current upstream image list before merging.

hack/ci/provision-hypershift.sh
```diff
@@ -0,0 +1,33 @@
...
+git fetch https://github.com/Azure/ARO-HCP.git main
+export TARGET_SHA
+TARGET_SHA=$(git rev-parse --short=7 FETCH_HEAD)
+echo "ARO-HCP main HEAD: ${TARGET_SHA}"
+
+# Resolve service images from ACR
+# shellcheck source=hack/ci/resolve-acr-images.sh
+source "$(dirname "$0")/resolve-acr-images.sh"
+
+exec "$(dirname "$0")/provision-environment.sh"
```
roivaz, line 32: Wouldn't it make more sense that this lives in the hypershift provision step instead?

hack/ci/resolve-acr-images.sh
```diff
@@ -0,0 +1,109 @@
...
+echo "  mgmt-agent:   ${MGMT_AGENT_DIGEST}"
+echo "  kube-applier: ${KUBE_APPLIER_DIGEST}"
+
+export BACKEND_IMAGE="${ACR_URL}/${BACKEND_REPO}@${BACKEND_DIGEST}"
+export FRONTEND_IMAGE="${ACR_URL}/${FRONTEND_REPO}@${FRONTEND_DIGEST}"
+export ADMIN_API_IMAGE="${ACR_URL}/${ADMIN_API_REPO}@${ADMIN_API_DIGEST}"
+export SESSIONGATE_IMAGE="${ACR_URL}/${SESSIONGATE_REPO}@${SESSIONGATE_DIGEST}"
+export FLEET_IMAGE="${ACR_URL}/${FLEET_REPO}@${FLEET_DIGEST}"
+export MGMT_AGENT_IMAGE="${ACR_URL}/${MGMT_AGENT_REPO}@${MGMT_AGENT_DIGEST}"
+export KUBE_APPLIER_IMAGE="${ACR_URL}/${KUBE_APPLIER_REPO}@${KUBE_APPLIER_DIGEST}"
```
roivaz, line 109: A follow up to create a small go CLI out of this is probably worth it - better control flow, retries, etc.

hack/ci/provision-environment.sh
patriksuba: Can we extract the overriding into a `build-config-override.sh` script? It would be called by `aro-hcp-test-local-upgrade-commands.sh`, as mentioned in https://github.com/openshift/release/pull/80778.

`aro-hcp-test-local-upgrade-commands.sh` covers:
- parsing optional `*_IMAGE `pullspecs
- write image overrides via `IMAGE_MAP`
- `USE_OC_LOGIN_REGISTRIES`
- MSI / AKS sizing (callers can skip via env if unused)
- merge `${SHARED_DIR}/hypershift-image-overrides.yaml` when present

gmfrasca: Done - Extracted into `hack/ci/build-config-override.sh`, and now `provision-environment.sh` (and other scripts) can source it to reuse the build config override functionality

hack/ci/resolve-acr-images.sh
gmfrasca: Yep, synced. Added `EXPORTER_REPO`/`EXPORTER_DIGEST`/`EXPORTER_IMAGE` to `resolve-acr-images.sh` and `EXPORTER` to the `IMAGE_MAP` in `build-config-override.sh`.

hack/ci/provision-environment.sh
gmfrasca: ah good catch - this was sourced from the `from-main` provision commands, and I suppose only the `provision-environment` script was updated with the source-of-truth MGMT sizing override we want to use (and actually demonstrates one of the main motivations/benefits of this change).  updated to use current MGMT sizing override from `provision-environment`

hack/ci/resolve-acr-images.sh
gmfrasca: Since shallow git trees could be a concern, I think actually going the opposite direction and leveraging the GitHub REST API in this ACR resolver script, rather than `git log` and simply documenting a potential shortcoming, may be the better/safer path.  Updated to do so in a new commit so we can easily revert if there are other considerations we should make.

hack/ci/provision-hypershift.sh
gmfrasca: As in the step-registry script? The intended pattern is for each step-registry wrapper to be as light as possible (just handle slot envs then exec the appropriate provision script) and do the real work in ARO-HCP, where we have more control and flexibility.

hack/ci/resolve-acr-images.sh
gmfrasca: agreed, filed [ARO-28748](https://redhat.atlassian.net/browse/ARO-28748) for that

raelga (approved):
/lgtm
/approve

PR #6298, opened 2026-07-28
docs(mgmt-agent): add node-health controller design doc (AROSLSRE-1588)

raelga:
[AROSLSRE-1588](https://redhat.atlassian.net/browse/AROSLSRE-1588)

### What

Adds a design doc, `docs/controllers/node-health.md`, for the **node-health controller**: a management-cluster controller that detects "Ready but broken" nodes from kubelet Events and labels them so a separate mitigation controller can drive them out of service. It documents the problem (SWIFT v2 delegated-NIC nodes that stay `Ready` while the kubelet cannot create pod sandboxes), the detection approach (hard-coded, modular Go detectors that share a common toolkit, driven by shared informers with an Event indexer keyed by node, computed as a pure function of node state), the load-bearing discriminator (zero successful pod-sandbox creation across the window, which separates a hard-wedge from a harmless flap), the floor and dwell measured per pod from durable Pod state, the safety guards (`enabled` and a Node-Ready precondition), and the scope split: this controller only detects and labels/unlabels, while every disruptive action (cordon, taint, evict, delete) belongs to a separate mitigation controller left as future work. It also records the deliberate decision to keep detection thresholds hard-coded in each detector's Go file rather than in a ConfigMap. Linked from `docs/README.md`.

### Why

These wedged nodes are invisible to every normal health signal: the node stays `Ready`, the `MultitenantPodNetworkConfig`/`NodeNetworkConfig` CRDs stay populated, and no NodeCondition flips, yet every pod scheduled there fails with `FailedCreatePodSandBox`. Today an SRE has to notice the pattern and manually cordon and drain the node. This design records the detection approach and a reversible rollout (label in non-prod, then in prod), with all disruptive mitigation owned by a separate controller, so the implementation has an agreed reference and each step can be validated in production before the next is enabled.

### Testing

Docs-only change. All embedded Mermaid diagrams were validated with the Mermaid parser. No code or pipeline behavior changes.

This is the design only. The controller (detection plus label/unlabel) is implemented in a separate PR, [#6284](https://github.com/Azure/ARO-HCP/pull/6284); this doc is split out so the design can be reviewed on its own.

### PR Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/docs/controllers/node-health.md
+++ b/docs/controllers/node-health.md
@@ -118,7 +118,7 @@ lookup. Success is read from Pod state rather than Events (see Reading the signa
 flowchart TD
     E[Kubelet Events<br/>FailedCreatePodSandBox] --> IDX[Event informer<br/>Indexer keyed by node]
     N[Node informer] --> R[reconcile Node<br/>level-driven]
-    P[Pod informer<br/>PodReadyToStartContainers status + transition] --> R
+    P[Pod informer<br/>PodReadyToStartContainers status + transition<br/>container start times] --> R
     IDX --> R
     CM[ConfigMap informer<br/>operational switches] --> C[Config<br/>atomic pointer]
     C --> R
@@ -330,7 +330,8 @@ labels a node; only a continuous zero-success span does.
 ### Corroborating signals (optional)
 
 Version 1 needs only the central signal: `FailedCreatePodSandBox` Events for
-failures and the `PodReadyToStartContainers` Pod condition for successes.
+failures and the `PodReadyToStartContainers` Pod condition (plus container start
+times on finished pods) for successes.
 If a later revision gains node or CNS access, two signals strongly confirm a
 hard-wedge before any disruptive action: CNS reporting
 `SecondaryInterfacesExist: false` across the window, and the delegated MAC being
@@ -548,7 +549,8 @@ recorder patches an existing Event when it aggregates repeats), and
 `get`/`list`/`watch` on its own ConfigMap. Pods are
 read-only, used to read each failing pod's `PodReadyToStartContainers` condition
 (its status and `lastTransitionTime`, which supply the dwell and the success
-signal); the controller never mutates a Pod. It needs no access to Secrets or any
+signal) and, on finished pods, the container start times that carry the same
+success signal; the controller never mutates a Pod. It needs no access to Secrets or any
 guest-cluster resource, which bounds the blast radius of the ServiceAccount to
 node scheduling metadata.
 
--- a/docs/controllers/node-health.md
+++ b/docs/controllers/node-health.md
@@ -334,8 +334,11 @@ ConfigMap, plus a `Ready` precondition on the reconcile. The detection threshold
 the SWIFT-v2 node scoping, and the dwell are deliberately *not* here: they are part
 of the detection logic, not operational guards.
 
-- **`enabled`**: a hard off switch. When false the controller still runs its
-  informers but records no state, enqueues nothing, and takes no action.
+- **`enabled`**: a hard off switch. When false the controller enqueues nothing,
+  reconciles nothing, and takes no action on any node. The pod informer handlers
+  still write to the in-memory success map, but nothing reads it while disabled
+  and it is discarded on the disabled to enabled transition, so a disabled
+  controller carries no state forward.
 - **Node-Ready precondition**: a `NotReady` node is skipped and left to node
   lifecycle. The controller exists for the nodes that stay `Ready` while broken.
 
@@ -477,7 +480,7 @@ The controller adds no CRD, webhook, or finalizer, so there is no admission
 latency to account for. Operationally:
 
 - **Fail-safe.** If the controller is down, no node is labeled and
-  nothing is un-done; the pre-existing manual SRE procedure still applies. A
+  nothing is undone; the pre-existing manual SRE procedure still applies. A
   misfire is bounded by the SWIFT-v2 detection scope, and is fully reversible.
 - **Alerting.** Suggested alert: `nodehealth_wedged_nodes > 0` for longer than a
   chosen window, per enabled environment.
--- a/docs/controllers/node-health.md
+++ b/docs/controllers/node-health.md
@@ -358,7 +358,20 @@ field manager `mgmt-agent-node-health`:
 - label `node-health.aro-hcp.azure.com/status=wedged`
 - annotation `.../detector`: which detector fired
 - annotation `.../reason`: a short human-readable summary (failure/success counts)
+- annotation `.../signature`: the detector signature that classified most of the
+  failing pods, naming which failure mode within the detector's family the node
+  is showing
 - annotation `.../observed-at`: RFC3339 timestamp of the first label in this episode
+
+The single label is the whole selectable surface. Everything descriptive stays an
+annotation on purpose: anything selectable becomes a contract, and the contract
+here is deliberately one bit, "this node is wedged". The signature is triage
+detail so an operator does not have to go read Events that may already have been
+collected, and it is deliberately not a decision input. Mitigation keys on the
+detector, since one detector is one fault with one remedy; branching on the
+signature would require the mitigator to know detector internals. The same
+signature is reported as a label on the detection counter, where the set is
+bounded by the hard-coded signature list.
 
 The label and annotations are removed in two cases: when recovery is confirmed
 (the detector no longer fires and pods are starting again), and when no detector
@@ -434,8 +447,10 @@ rolled out with the detector.
 The controller registers Prometheus metrics on the `mgmt-agent` `/metrics`
 endpoint (subsystem `nodehealth`):
 
-- `nodehealth_detections_total{detector}`: counts wedged detections (label
-  transitions).
+- `nodehealth_detections_total{detector,signature}`: counts wedged detections
+  (label transitions), broken down by the failure mode within the detector's
+  family. Both label sets are bounded by the hard-coded detector and signature
+  lists.
 - `nodehealth_label_actions_total{action,result}`: label/unlabel actions by
   result.
 - `nodehealth_wedged_nodes`: gauge of nodes currently carrying the wedged label
@@ -489,8 +504,9 @@ latency to account for. Operationally:
 - **Alerting.** Suggested alert: `nodehealth_wedged_nodes > 0` for longer than a
   chosen window, per enabled environment.
 - **Inspect a flagged node.** `kubectl describe node <node>` shows the
-  `node-health.aro-hcp.azure.com/status=wedged` label, the `detector`/`reason`/
-  `observed-at` annotations, and the `NodeHealth*` Events.
+  `node-health.aro-hcp.azure.com/status=wedged` label, the
+  `detector`/`reason`/`signature`/`observed-at` annotations, and the
+  `NodeHealth*` Events.
 - **Disable quickly.** Set `enabled: false` in the ConfigMap to stop all action;
   the change is hot-reloaded and takes effect within one resync. A live edit is
   break-glass, not a durable setting: the Helm values in git own the ConfigMap's
--- a/docs/controllers/node-health.md
+++ b/docs/controllers/node-health.md
@@ -1,0 +1,667 @@
+# Node health
+
+This document describes the design of the **node-health controller**, a management-cluster
+controller that detects "Ready but broken" nodes from kubelet Events and
+**labels** them so they can be acted on. A separate mitigation controller, out of
+scope here, acts on the labeled nodes.
+
+This is a design document. The controller is part of the `mgmt-agent` component,
+in package `mgmt-agent/pkg/controller/nodehealth`, and ships as part of the
+`mgmt-agent` deployment. Detection and labeling are non-disruptive, so the
+controller can run in production and be confirmed before any separate mitigation
+controller acts on its labels. See [Monitoring](../monitoring.md) for how the metrics it emits
+are collected.
+
+Tracking: [AROSLSRE-1588](https://redhat.atlassian.net/browse/AROSLSRE-1588).
+
+## Problem
+
+SWIFT v2 delegated-NIC nodes on the AKS management clusters can wedge in a state
+that is invisible to every normal health signal. The node stays `Ready`, its
+`MultitenantPodNetworkConfig` and `NodeNetworkConfig` CRDs stay fully populated
+and healthy, and no NodeCondition flips. Yet the kubelet cannot create pod
+sandboxes on it: every pod scheduled there fails with a sustained
+`FailedCreatePodSandBox` signature (`route ... no such network interface`,
+`network is unreachable`, `mtpnc is not ready`, or `dhcp discover ... timed out`),
+and no pod ever starts.
+
+Because the node looks `Ready`, the scheduler keeps placing pods on it and they
+keep failing. Today an SRE has to notice the pattern, correlate the kubelet
+Events, and manually cordon and drain the node. That is slow, it is on-call toil,
+and it only happens after workloads have already been stuck on the bad node for a
+while.
+
+The failure we are reacting to, in one picture:
+
+```mermaid
+sequenceDiagram
+    autonumber
+    participant S as Scheduler
+    participant N as Wedged node (Ready=True)
+    participant K as Kubelet
+    Note over N: SWIFT NIC torn down, node still Ready
+    S->>N: schedule pod
+    N->>K: create sandbox
+    K-->>N: FailedCreatePodSandBox (route / mtpnc / dhcp)
+    Note over N: pod never starts, scheduler retries here
+    S->>N: schedule next pod
+    N->>K: create sandbox
+    K-->>N: FailedCreatePodSandBox again
+```
+
+## Goal
+
+- **Detect** wedged management-cluster nodes automatically from central kubelet
+  Events, with no per-node agent.
+- Detection is **hard-coded and modular**: detectors are Go units that share
+  common primitives (event-signature match, sustained-storm floor, and the
+  load-bearing zero-success-continuity check), each a pure function of the node's
+  observed state and the controller's recorded per-node success history, and
+  exhaustively unit-tested. Adding a failure mode is a new tested detector in
+  code, not a ConfigMap edit.
+- **Label** a detected node (and annotate why), and unlabel it when it recovers.
+  Labeling is non-disruptive and reversible; a separate mitigation controller
+  consumes the label and owns any disruptive action.
+- Labeling is **guarded**: an off switch, live via a ConfigMap. Detection is
+  intrinsically scoped (it only considers SWIFT-v2 nodes) and requires a sustained
+  storm, so it does not label the wrong node in the first place.
+
+## Non-goals
+
+- **Not a mitigator.** This controller never cordons, evicts, drains, deletes, or
+  taints a node or a pod. Mitigating labeled nodes is a separate, asynchronous
+  controller (see [Evolution](#evolution-a-separate-mitigation-controller)).
+- **Not a NotReady-node handler.** A node that has gone `NotReady` is already
+  handled by normal node lifecycle and the cloud provider; the controller explicitly
+  skips it. The controller exists precisely for the nodes that stay `Ready` while
+  broken.
+- **Not a per-node daemon.** Version 1 draws its only signal from central kubelet
+  Events read through the API server. There is no host agent, no `ip link`
+  probing, no CRD inspection.
+- **Not a config-driven matching engine.** SWIFT NIC teardown is the first
+  detector, hard-coded and tested. New fault families are added as new detectors
+  that reuse the shared detection primitives, in code, not as configuration.
+- **Not a blind node-rotator.** Every action is preceded by a durable label, an
+  Event, and a metric, so a wedge is always *surfaced* (as a new detector and a
+  bug for a signature we do not yet cover) rather than silently recycled. The goal
+  is to remove the manual toil of mitigating known faults, not to mask new ones by
+  rotating nodes without a diagnosis.
+
+## Background: why Events, and why label-only
+
+The wedge is invisible to CRDs and NodeConditions, so the control-plane objects
+cannot be the signal. The one reliable cluster-side symptom is a sustained stream
+of kubelet `FailedCreatePodSandBox` Events with a specific reason string and zero
+successful container starts on the node. Those Events are already in etcd and
+readable through the API server, so a controller can watch them centrally without
+touching the host.
+
+Keeping mitigation out of this controller is a deliberate rollout-safety choice.
+Detection and labeling are observable and completely non-disruptive, so this
+controller runs in production with zero blast radius while confidence is built that
+it identifies exactly the right nodes and nothing else. Any disruptive action lives
+in a separate mitigation controller that consumes the label and is rolled out on
+its own schedule.
+
+## Design overview
+
+The controller runs inside `mgmt-agent`. It watches Nodes, Pods, and kubelet
+Events through shared informers, plus a small ConfigMap for its operational
+switches. On any watch event it enqueues the affected Node and reconciles it
+level-driven, computing the decision as a pure function of what the informers
+currently hold for that node: its object, the Events indexed against it, and the
+state of its Pods. An `Indexer` on the Event informer, keyed by the node a pod is
+scheduled to, returns a node's recent `FailedCreatePodSandBox` Events in a single
+lookup. Success is read from Pod state rather than Events (see Reading the signal).
+
+```mermaid
+flowchart TD
+    E[Kubelet Events<br/>FailedCreatePodSandBox] --> IDX[Event informer<br/>Indexer keyed by node]
+    N[Node informer] --> R[reconcile Node<br/>level-driven]
+    P[Pod informer<br/>PodReadyToStartContainers status + transition] --> R
+    IDX --> R
+    CM[ConfigMap informer<br/>operational switches] --> C[Config<br/>atomic pointer]
+    C --> R
+    R --> D{"detect(node, events, pods)<br/>pure function"}
+    D -- wedged --> L[Label wedged]
+    D -- healthy, but labeled --> U[Unlabel<br/>on confirmed recovery]
+    L --> M[(Metrics + Events)]
+    U --> M
+```
+
+Reconciliation is level-driven. The signal that pushes a node over the threshold
+is a kubelet Event, and the controller watches those Events, so the API server
+delivers the very Event that crosses the line and the node is reconciled then.
+The informers' ordinary resync, together with the Node and Pod watches, is enough
+to re-evaluate recovery.
+
+A periodic sweep re-enqueues candidates so a firing that becomes true purely
+because the dwell elapsed is still reconciled without a new Event. The sweep is
+the union of two selectors rather than every node in the cluster: the SWIFT-v2
+nodes, which are the only detection candidates, and the nodes carrying the wedged
+label, so a stale label is swept and retired even if the node stopped being a
+candidate while the controller was down.
+
+## Detection
+
+### Two failure classes, one error string
+
+The SWIFT NIC fault presents as two mechanisms that emit the **identical** error
+string (`route ip+net: no such network interface` from the azure-vnet CNI), so
+the string alone is never a trigger. They differ only in whether the node can
+still start pods:
+
+- **Flap** (self-heals, do not act): the accelnet VF is still present on the VM;
+  the MANA NIC transiently presents the slave secondary NIC to the CNI, which
+  races and fails. Route errors come in **bounded bursts with clean gaps**, DHCP
+  discover still succeeds on retry, and **pods keep starting between the
+  failures**. The node recovers with no operator action.
+- **Hard-wedge** (only recovers on node delete): the VF is **gone** from the VM
+  (CRP/Compute detachment). Route errors are a **continuous storm with no gaps**,
+  DHCP discover never succeeds, and **no pod sandbox is ever created** until the
+  node is deleted.
+
+The single reliable discriminator, confirmed across four on-node log collections
+and ICM 832382845, is **successful pod-sandbox creation on the node**: a wedged
+node has zero successes for the whole failure span; a flapping node keeps
+succeeding in the same hours it fails. Everything the detector does is built
+around that signal, which is why the zero-successful-sandbox rule (below) is the load-bearing
+condition, not the error count.
+
+### Detectors
+
+A detector is a small, hard-coded Go unit that decides, for one fault family,
+whether a node is wedged. Every detector implements a common `Detector` interface
+(`Applies`, `Evaluate`, `Fires`, plus `Name`/`Reason` for surfacing), and the pure
+`decide` core iterates a code registry of detectors without depending on any
+concrete type. This keeps detection modular: the engine, the shared evaluation
+primitives, and each fault family live in separate files, so adding a family means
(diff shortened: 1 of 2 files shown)


docs/node-healer.md
```diff
@@ -0,0 +1,365 @@
+# Node healer
+
+This document describes the design of the **node healer**, a management-cluster
```
janboll: Why is this called healer? It's only labeling and perhaps also cordoning. 
I think wording matters a lot here, cause healer suggest it repairs or does something to change the node's health state. 
My suggestion would be: node-fencer

Other claude ideas:
- node-quarantine
- wedge-detector
- node-sentinel

janboll: oh wait, did you name it like that, cause you want to add node deletion later on?

gmfrasca: Even if this eventually adds node deletion/resolution functionality later, I think we should rename the overall controller.  Right now, the public metadata (labels/metrics like `node-healer.aro-hcp.azure.com/status`) would give false impressions that nodes are being auto-healed out-of-the-box, and could discourage manual intervention actions to actually address the issue.  Naming the controller something more generic would avoid that, and then node-healer could be a specific subcomponent that could be implemented later.  

Maybe something like `node-health` or `node-health-check` so that the labels and metrics would be `node-health.aro-hcp.azure.com/status=wedged` and `nodehealth_wedged_nodes`, respectively

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+    S->>N: schedule next pod
+    N->>K: create sandbox
+    K-->>N: FailedCreatePodSandBox again
+```
+
+## Goal
+
+- **Detect** wedged management-cluster nodes automatically from central kubelet
+  Events, with no per-node agent.
+- Detection is **declarative and generic**: failure modes are expressed as
```
stevekuznetsov: I would strongly not recommend doing this. It is just as easy to deploy a new version of the code as it is to roll a new version of config, and trying to model complex and difficult detection patterns in YAML with some parsing engine is a level of complexity we just don't need.

Just hard-code exactly the behavior we need to detect and test it well.

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+    T --> R
+    C --> R
+    R --> V{Detector fires?<br/>threshold + dwell}
+    V -- yes --> L[Phase 1: Label wedged<br/>Phase 2: also cordon]
+    V -- no, but labelled --> U[Unlabel / uncordon<br/>on confirmed recovery]
+    L --> M[(Metrics + Events)]
+    U --> M
+```
+
+Reconciliation is level-triggered. A background resync re-enqueues every
```
stevekuznetsov: This seems wrong. If you're watching `Event`s and your input that leads to labelling is `Event`s, you will be guaranteed to see the last `Event` that pushes a node over the threshold, and you do not need such a 30s relist.

docs/node-healer.md
```diff
@@ -0,0 +1,365 @@
...
+    failures: 20        # this many matching Events
+    window: 10m         # within this rolling window
+    requireZeroSuccess: true  # and zero successful container starts in the window
+```
+
+`matchAny` entries are regexes; an Event fires the detector if its `reason`
+matches and any regex matches its message. The detectors list is ordered and the
+first detector to fire wins.
+
+The default threshold (20 failures in 10m with zero successful starts) is chosen
```
stevekuznetsov: I don't know if this is possible - check the AKS clusters, I believe `Event` objects have a very short TTL (5min?). Your level-driven code will be required to work within that boundary. Perhaps you can join the recent view of `Event`s with the `Pod` state - how long has it been in `ContainerCreating`?

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+The default threshold (20 failures in 10m with zero successful starts) is chosen
+to be an order of magnitude above normal transient churn, not a tuned value: a
+healthy node starts containers continuously, so `requireZeroSuccess` alone
+already excludes most flapping, and the count/window just require the wedge to be
+sustained rather than a single bad Event. The numbers are ConfigMap-tunable per
+environment.
+
+### Tracker
+
+The `Tracker` keeps, per node, a sliding window of detector-failure timestamps
```
stevekuznetsov: This is simply an `Indexer` on the informer.

stevekuznetsov: Note - it *has* to be an `Indexer` on an informer, since the only cache you are allowed to keep is the one that can be re-created from the `LIST` on pod restart. Any cache you keep with data that cannot be recreated leads your controller to fail to do the right thing on restart.

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+and node-level success timestamps, bounded by a retention. Success is the kubelet
+`Started` Event (a container actually started on the node). A detector `Fires`
```
stevekuznetsov: No such thing as `Fires`. When you get an `Event` watch event, enqueue the `Node` that the pod is on, and process that node in a level-driven way from the current state you have in your informers.

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+pods.
+
+## Safety guards
+
+Every action passes through the same guard rails, all live-tunable via the
+ConfigMap:
+
+- **`enabled`**: a hard off switch. When false the controller still runs its
+  informers but records no state, enqueues nothing, and takes no action.
+- **`dryRun`**: the controller detects, emits Events and metrics, and logs the
```
stevekuznetsov: I don't think we need to be this careful. Labels are safe.

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+  label it *would* apply, without mutating any node. This is the recommended
+  initial rollout mode.
+- **`allowedNodeNamePrefixes`**: an allow-list of node-name prefixes (e.g.
```
stevekuznetsov: I don't see the value of this

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+  `aks-userswft`). A node not matching any prefix is never acted on. When
+  `enabled`, the config is rejected unless it lists at least one non-empty prefix,
+  so an empty allow-list can never silently match every node.
+- **`dwellBeforeAction`**: a node must have been failing for at least this long
```
stevekuznetsov: This is not a safety thing, this is core logic for your detection

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+  before any action; it filters out brief blips.
+- **Node-Ready guard**: a `NotReady` node is skipped and left to node lifecycle.
```
stevekuznetsov: This is not a safety thing, this is core logic. You might even not want this, if node auto-remediation from AKS is slow and you know the node is bad, you can act faster

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+- annotation `.../detector`: which detector fired
+- annotation `.../reason`: a short human-readable summary (failure/success counts)
+- annotation `.../observed-at`: RFC3339 timestamp of the first label in this episode
+
+When the node stops matching (the detector no longer fires and recovery is
+confirmed by the tracker), the label and annotations are removed. Labelling is
+idempotent and honours `dryRun`. Nothing else happens in Phase 1: the label is
+purely a signal for a human, an alert, or a later phase.
+
+A restart-safety subtlety: right after a controller restart the tracker has no
```
stevekuznetsov: Right, this is exactly why you don't build your own cache. Just use the informer.

docs/node-healer.md
```diff
@@ -0,0 +1,365 @@
...
+  (0 while disabled).
+- `nodehealer_dry_run`: 1 in dry-run, 0 otherwise.
+
+It also emits Kubernetes Events on the affected node (`NodeHealerWouldLabel` in
+dry-run, and the corresponding real-action Events), so the decision is visible in
+`kubectl describe node` as well as in metrics.
+
+## Test plan
+
+- **Unit** (`mgmt-agent/pkg/controller/nodehealer`, run in the `mgmt-agent` unit
```
stevekuznetsov: 1. write your core reconcile logic as a pure function of input state (node, events) to output state (do you label or not - the applyconfig)
2. write exhaustive tests for this exactly

see hypershift pki operator for prior art

docs/controllers/node-health.md
```diff
@@ -0,0 +1,365 @@
...
+
+The controller's ServiceAccount needs only: `get`/`list`/`watch` on Nodes and
+Events, `patch` on Nodes (to set/remove the label, annotations, and in Phase 2
+`spec.unschedulable`), and `get`/`list`/`watch` on its own ConfigMap. It needs no
+access to pods, secrets, or any guest-cluster resource, which bounds the blast
+radius of the ServiceAccount to node scheduling metadata.
+
+## Evolution: a separate drainer
+
+Draining is intentionally left to a **separate, asynchronous controller**, not
```
stevekuznetsov: s/Draining/Draining and deleting/

docs/node-healer.md
raelga: Good point, and I agree the name over-promises for what Phase 1 actually does (detect, label, cordon). I've renamed it to `node-health`, so the label is now `node-health.aro-hcp.azure.com/status=wedged` and the metrics are `nodehealth_*`.

To janboll's question: yes, node deletion is coming, but as a separate async controller (the drainer), not folded into this one. That split is deliberate. This controller only marks and fences a node, it never changes the node's health, and the actual remediation/replacement lives in the drainer. So the "heal" name is better reserved for that future remediation piece, which is exactly gmfrasca's point about node-healer being a later subcomponent.

I also moved the doc to `docs/controllers/node-health.md`, a new folder for per-controller design docs. The controller code in #6284 will follow with the same rename (package `nodehealth`, label and metric constants).

docs/controllers/node-health.md
raelga: Agreed, dropping the config-driven detection engine. Detection is now hard-coded and tested. I kept it modular though: detectors are a small library of Go units sharing common primitives (the event-signature match, the sustained-storm floor, and the load-bearing zero-success-continuity check), each a pure function of node state and unit-tested. `swift-vf-teardown` is the first one, and a new fault family is a new tested detector rather than a ConfigMap edit. I updated the goal, the detectors and configuration sections, and flipped the discarded-alternatives entry to match.

docs/controllers/node-health.md
raelga: Agreed, dropped it. The signal that pushes a node over the threshold is a kubelet Event and we watch those, so the API server delivers the very Event that crosses the line and we reconcile then. The doc now says reconciliation is level-driven and relies on the informers' ordinary resync plus the Node/Pod watches to re-evaluate recovery, no 30s relist.

docs/controllers/node-health.md
raelga: Right. Replaced the custom sliding-window cache with an `Indexer` on the Event informer keyed by the node a pod is scheduled to, so a node's recent `FailedCreatePodSandBox`/`Started` Events come back in a single lookup. No hand-rolled cache.

docs/controllers/node-health.md
raelga: Removed the `Fires` concept. On an Event watch event we enqueue the `Node` the pod is on and reconcile it level-driven from current informer state (its object, the indexed Events, its Pods). The decision is a pure function of that state.

docs/controllers/node-health.md
raelga: Exactly, dropped the custom tracker. On restart the controller rebuilds from the `LIST` of Nodes/Pods/retained Events and recomputes each node's decision from scratch, so the only state is what the informers already hold and can rebuild.

docs/node-healer.md
raelga: Checked the AKS side and verified it live on INT (`int-uksouth-mgmt-1`). Event retention is about 1 hour, not ~5 min: it is the upstream kube-apiserver `--event-ttl` default, which AKS keeps and does not expose as a configurable flag. On the live cluster every Event had a last-seen within the past hour, so the ~1h window comfortably covers our 30-60 min decision window. Kubernetes also aggregates a recurring `FailedCreatePodSandBox` into one Event with `firstTimestamp`/`lastTimestamp`/`count`, and during a continuous storm that object keeps getting updated (I saw a recurring event hold its `firstTimestamp` with a count in the thousands), so the storm age is readable directly.

That said, your `ContainerCreating` idea is in: the controller also reads duration from `Pod` state (how long the oldest failing pod has been stuck in `ContainerCreating`), which is durable and independent of Event GC. Doc updated to spell this out.

docs/controllers/node-health.md
raelga: Agreed, and this is now the core of the design. The discriminator between a self-healing flap and a hard wedge is whether any pod sandbox succeeds on the node during the storm: flap keeps starting pods (S>0), a wedge has zero successful sandbox creations across the window (S==0). That zero-success-over-a-window check lives in the detector, not in a safety guard. The detector reads it from indexed Pod state plus the node's Events, so it survives Event GC.

docs/controllers/node-health.md
raelga: Moved the dwell into detection as well: it is the minimum continuous span of failures-with-zero-success we require before classifying a node as wedged, which is what separates a real wedge from a transient flap burst.

One correction on the premise: we are not racing AKS node auto-repair here. AKS auto-repair only fires on a node that goes `NotReady` (or reports no status for 10 minutes), and it waits 5 minutes before acting. Our wedged nodes stay `Ready` the whole time, so AKS never remediates them. The dwell exists purely for wedge-vs-flap discrimination, not to beat AKS to the node.

docs/node-healer.md
raelga: Adopted this. The decision is a pure function of node state: given the node, its indexed Events, its Pods, and an injected clock, it returns the desired health-label state. No API-server access inside it, so it is exhaustively table-tested with fixtures and a fake clock, asserting the output with cmp.Diff. All the I/O (list/apply) sits in the outer sync layer.

This mirrors the hypershift control-plane-pki-operator pattern: an outer `sync*(ctx)` owns the reads and the server-side apply, and an inner `process*(ctx, inputs, now func() time.Time)` is a pure function over injected getters that returns an ApplyConfiguration. The tests call the inner function directly with fake getters and a fake clock. The Test plan section now describes this split.

docs/controllers/node-health.md
raelga: Agreed. Phase 1 labelling is non-disruptive, so it runs for real with no dry-run. I dropped the blanket "every action is guarded by dry-run" framing. `dryRun` now guards only the Phase 2 cordon, which is the disruptive step. The Configuration, Actions, Observability and Rollout sections were all updated to reflect that.

docs/controllers/node-health.md
raelga: Dropped the name allow-list. It was a fragile proxy for "this node can hit the SWIFT-v2 fault", over-inclusive and under-inclusive at the same time.

The scoping moved into the detector as a label predicate: `swift-vf-teardown` only evaluates nodes carrying `kubernetes.azure.com/podnetwork-swiftv2-enabled=true`. That is the stable AKS-managed signal for SWIFT-v2 capability. On the INT mgmt cluster the `userswft*` pools carry it (with `secondary-nic-count=7`), the system pool carries it with count 1, and the infra pools do not carry it at all, so a non-SWIFT node is never a candidate. Node-applicability is a shared detector primitive now, not a config-level allow-list.

docs/controllers/node-health.md
raelga: Reworded, but kept it generic rather than pinning it to "draining and deleting". The separate controller may drain, delete, taint, or otherwise recycle the node depending on the mitigation policy, so the doc now frames it as a "separate mitigation controller" and calls out the specific mitigation action (drain, delete, taint, or otherwise recycle) as out of scope here. The current manual runbook (cordon, drain with --disable-eviction, delete) stays in the section as today's example of what it will automate.

docs/controllers/node-health.md
```diff
@@ -0,0 +1,475 @@
...
+- **Extend coverage.** Add or adjust a detector (signature regex plus threshold)
+  in the ConfigMap; no redeploy required.
+
+### RBAC
+
+The controller's ServiceAccount needs only: `get`/`list`/`watch` on Nodes and
+Events, `patch` on Nodes (to set/remove the label, annotations, and in Phase 2
+`spec.unschedulable`), and `get`/`list`/`watch` on its own ConfigMap. It needs no
+access to pods, secrets, or any guest-cluster resource, which bounds the blast
+radius of the ServiceAccount to node scheduling metadata.
```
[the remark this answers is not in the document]
raelga: @copilot Good catch, that was a real contradiction. The design does rely on the Pod informer for the `ContainerCreating` dwell (the durable, GC-independent duration signal), so RBAC now grants `get`/`list`/`watch` on Pods alongside Nodes and Events. I kept it read-only and spelled that out: the controller never mutates a Pod, and it still needs no Secrets or guest-cluster access, so the blast radius stays bounded to node scheduling metadata.

docs/node-healer.md
```diff
@@ -0,0 +1,295 @@
...
+
+This document describes the design of the **node healer**, a management-cluster
+controller that detects "Ready but broken" nodes from kubelet Events and drives
+them out of service in phases: first by **labelling** them, then by marking them
+**unschedulable**. A separate drainer, out of scope here, later consumes those
+nodes.
+
+This is a design document. The controller is implemented in this repository under
+[`mgmt-agent/pkg/controller/nodehealer`](../mgmt-agent/pkg/controller/nodehealer)
+and ships as part of the `mgmt-agent` deployment. It is delivered in phases so we
```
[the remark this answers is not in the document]
raelga: @copilot Fixed. This PR is design-only (the controller lands in a separate implementation PR), and the intro both asserted "is implemented in this repository" and linked a `mgmt-agent/pkg/controller/nodehealth` path that does not exist in this branch. I dropped the "in this repository" claim and the broken link; the intro now names the target package as inline code and keeps the phased-delivery framing. The doc was also renamed to `docs/controllers/node-health.md`, so this thread is against the old path.

docs/node-healer.md
```diff
@@ -0,0 +1,295 @@
...
+    }
+    Wedged --> Healthy: recovery confirmed<br/>unlabel (+ uncordon)
+```
+
+## Configuration
+
+The full configuration is delivered as a ConfigMap the controller watches, so
+detectors, thresholds, and safety knobs are tuned without a redeploy. The
+ConfigMap is hot-reloaded: on update the controller recompiles and revalidates the
+config and swaps it atomically; on delete it reverts to a safe disabled default.
```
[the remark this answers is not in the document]
raelga: @copilot The concrete problem here (a claim of already-landed code with a broken link) was in the intro and is fixed: it no longer says "implemented in this repository" and no longer links a non-existent path. The Configuration section is written in the same timeless design voice as the rest of the doc (present-tense architecture), and the `mgmt-agent/values.yaml` it links to does exist; the `enabled`/`dryRun` switches it describes are the design intent that ships with the implementation. The doc was also renamed to `docs/controllers/node-health.md`, so this thread is against the old path.

raelga:
@copilot thanks for the review. Pushed bdd87f2ad addressing the two low-confidence notes:

- **README "declarative detectors"**: fixed. Changed the index blurb to "hard-coded detectors" so it matches the design doc, which states detectors are hard-coded Go units and deliberately rejects a config-driven/declarative engine.
- **Label key vs PR #6284**: no change needed. This doc already uses `node-health.aro-hcp.azure.com/status=wedged` everywhere, and PR #6284's description was reworked to the same `node-health` key. There is no `node-healer` label anywhere, so the two are consistent.

raelga:
Updated the design after the ARO Architecture Office Hours on 2026-07-29 (thanks @deads2k).

The main change is in the **Evolution: a separate mitigation controller** section. The automated mitigation now describes **`NoSchedule` taint + evict only the non-running pods, letting the existing cluster autoscaler supply capacity**, instead of the manual cordon/drain/delete runbook with a custom scale-up. Rationale from the review:

- Taint with `NoSchedule` (never `NoExecute`) so healthy running pods stay put and nothing new lands on the wedged node.
- Evict (not delete) only the pods that cannot start; they reschedule onto a healthy node immediately.
- Let the cluster autoscaler add capacity from the resulting pod pressure, so we do not reinvent scale-up or node-type selection.
- Drain/delete is demoted to a reserved big gun for a confirmed hard-wedge (VF gone, only recovers on delete).

Also in this push:

- Guardrails recorded: leave pod tolerations alone and rely on `NoSchedule`; disruption budget of one node at a time with a cap on concurrent taint/evict.
- New non-goal: this is not a blind node-rotator. Every action is surfaced as a label, Event, and metric, and an uncovered signature becomes a new detector and a bug, not a silent recycle.
- Two Discarded-alternatives entries: cordon vs `NoSchedule` taint for Phase 2, and the old cordon/drain/delete mitigation shape.

Scope note: this is all in the mitigation-controller design, which stays a separate controller and a separate follow-up. **Phase 2 of the detector in this PR remains a cordon** (`spec.unschedulable`), which is just the non-evicting half; the taint-and-evict half belongs to the mitigation controller. The Phase 1 detector code in [#6284](https://github.com/Azure/ARO-HCP/pull/6284) is unaffected.

@deads2k does this capture the taint-and-evict + autoscaler direction correctly?

raelga:
Reworked the design from the review:

- **Success signal** reads `PodReadyToStartContainers` transitioning to `True` in-window (keyed on `lastTransitionTime`), not kubelet `Started` Events, which also fire on container restarts and host-network pods. Absent condition = indeterminate, not zero successes. It is beta/default from k8s 1.29, GA 1.31, so presence on the AKS mgmt clusters is verified before enablement.
- **Failure signal** is a continuous active span, not a windowed sum of `Event.Count` (a recurring `FailedCreatePodSandBox` aggregates into one Event whose `count` grows over its whole lifetime, so summing overcounts).
- **Dwell** counts only pods that both match a `FailedCreatePodSandBox` Event and are still sandbox-not-ready, held continuously. Pods in `ContainerCreating` for unrelated reasons (image pull, volume, scheduling) no longer count.
- **Detection + label only, cordon removed.** A cordoned node gets no new pods, so the fresh-success signal can never re-appear on it, which left the old uncordon-on-recovery path blind. All disruptive action (cordon, taint, evict, delete, uncordon) moves to the separate mitigation controller that consumes the `status=wedged` label, matching the taint-and-evict + autoscaler-replacement direction.
- **Dropped `dryRun`** (it gated nothing here) from the doc and from [ARO-HCP#6284](https://github.com/Azure/ARO-HCP/pull/6284): the field, the `nodehealth_dry_run` metric, and the `values.yaml` value. Only `enabled` remains.

Tracking: [AROSLSRE-1588](https://redhat.atlassian.net/browse/AROSLSRE-1588).

raelga:
@copilot I went through the suppressed comments on this review. One was right, one was not.

The RBAC one was a real defect and I have fixed it. The section said Pods are read to get the `ContainerCreating` dwell, which contradicts the signal section: there is no `ContainerCreating` timestamp to read, and the dwell comes from the `PodReadyToStartContainers` condition's `lastTransitionTime`. It now says that. The architecture diagram had the same stale wording in the Pod informer node, so I fixed that too. I left the passage that explains why `ContainerCreating` is not the dwell, since that one is correct and is the reason the rest of the design is written the way it is.

I did not change the GA version. The comment says `PodReadyToStartContainers` went GA in 1.36 and that 1.37 is inaccurate. Checking KEP-3085, the graduation is alpha in 1.28, beta in 1.29, GA in 1.37, so the doc is correct as written. This matters for the sentence it supports: on the 1.35 management clusters the condition is beta-but-default-on, which is exactly why the design treats an absent condition as indeterminate rather than as a zero.

Landed in `docs(controllers): correct the pod data the controller reads`.

raelga:
@copilot Thanks, two of these were real and I fixed them. One I checked and left alone.

The `docs/README.md` line was a genuine leftover. It described a phased label then unschedulable flow, which stopped being the design when this was reworked to detection only. Making a node unschedulable is mitigation, and mitigation is explicitly a separate controller and out of scope here, so the index was advertising behaviour this design does not have. It now says detection and labeling only.

The spelling was inconsistent, so I settled the prose on the majority spelling.

On the Event reason, american spelling is the convention for this package. The file it lives in is `labeler.go`, so the reason strings were the inconsistent part. They are now `NodeHealthLabeled` and `NodeHealthUnlabeled` in both the implementation and here. Nothing consumes them, since the controller is not enabled in any environment, so there is no alert or runbook keying off the old strings.

On the intro paragraph, I left it as is. The design doc is the design of record rather than a status page, and the package location it names is where the controller lives once the implementation lands. Rewording it to future tense would age badly the moment that merges, and the doc would then be wrong in the other direction.

raelga:
@copilot Suppressed comment reviewed and applied. "un-done" is now "undone".

I also corrected the off-switch bullet in the same push. The `enabled: false` description now matches the implementation:

> **`enabled`**: a hard off switch. When false the controller enqueues nothing, reconciles nothing, and takes no action on any node. The pod informer handlers still write to the in-memory success map, but nothing reads it while disabled and it is discarded on the disabled to enabled transition, so a disabled controller carries no state forward.

That came out of the matching review on the implementation PR [Azure/ARO-HCP#6284](https://github.com/Azure/ARO-HCP/pull/6284), where the same claim lived in the `Config.Enabled` doc comment.

raelga:
@copilot Both applied. I checked the convention before changing it: every other top-level heading under `docs/` is title case, and the index entries either side of this one are too, so `Node health` was the outlier in both places.

The document heading is now `# Node Health` and the index link text is `Node Health`. The anchor is unaffected, since the slug lowercases either way, and nothing links to the heading anchor.

raelga:
I went back over the review comments on this doc and re-checked the design against each one, since the doc has been reworked twice since then (mitigation split out, then detection-only).

Two had drifted back in presentation, and I fixed both:

- The `Detector` interface still had a `Fires` method. The architecture is the level-driven one that was asked for, the reconcile processes the node from current informer state and the method is a pure predicate over evidence already gathered, but the name invited the original objection again. It is now `MeetsThreshold`, in the code, the package README and this doc.
- The `Ready` precondition was still listed under "Safety guards" even though the review called it core logic. It has moved into Detection, where the doc explains it as part of the detection premise: the fault this controller exists for is the one that leaves a node `Ready` while it cannot start pods, so a `NotReady` node is skipped and left to node lifecycle. `enabled` is now the only safety guard.

The rest still hold. Detection is hard-coded Go with a single boolean of config, the reconcile is driven by the kubelet Event that crosses the threshold rather than a short relist, Event data is joined with Pod state (the dwell comes from the `PodReadyToStartContainers` transition time, not from counting Events, which is what makes the short Event TTL survivable), state comes from informer indexers, and the decision core is a pure function with table tests plus a replay of the captured production evidence.

The periodic sweep that remains is not a relist for detection. It exists because a threshold that becomes true purely because the dwell elapsed produces no new Event, and because a label left on a node that stopped being a candidate while the controller was down still needs retiring. It is scoped to the union of the SWIFT-v2 nodes and the labeled nodes rather than every node.

raelga:
@copilot Correct, the RBAC summary was wrong, and it was wrong in more than the one way you flagged. Fixed in [7bdaa5b](https://github.com/Azure/ARO-HCP/pull/6298/commits/7bdaa5b5b).

The controller does create Events, so read-only verbs on Events were wrong. It also needs `get` on Nodes, which the summary left out: the labeler re-reads the node live before a write, so a resync tick and a watch event racing on the same stale cache view do not both fire the counter and the Event. And the recorder patches an existing Event when it aggregates repeats, so `patch` belongs there too.

The paragraph now matches the ClusterRole in the chart exactly: `get`/`list`/`watch`/`patch` on Nodes, `list`/`watch` on Pods, `list`/`watch`/`create`/`patch` on Events, and `get`/`list`/`watch` on its own ConfigMap.

raelga:
@copilot I checked this one and the doc is right, so I am leaving it as it is.

Upstream `pkg/features/kube_features.go` pins the lifecycle of `PodReadyToStartContainersCondition` as alpha in 1.28, beta and on by default in 1.29, and GA in 1.37:

```go
PodReadyToStartContainersCondition: {
	{Version: version.MustParse("1.28"), Default: false, PreRelease: featuregate.Alpha},
	{Version: version.MustParse("1.29"), Default: true, PreRelease: featuregate.Beta},
	{Version: version.MustParse("1.37"), Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // GA in 1.37, remove in 1.40
},
```

I also confirmed it on a live mgmt cluster rather than relying on the source alone. On the 1.35 AKS management cluster the apiserver reports `kubernetes_feature_enabled{name="PodReadyToStartContainersCondition",stage="BETA"} 1`, which matches what the doc says: beta but default-on at the version we actually run, and verified present before enablement.

raelga:
@copilot One of the two suppressed comments is right, one is not.

**Events inconsistency: fixed.** The observability section named only `NodeHealthLabeled` while the RBAC section named both, so the two disagreed. The controller emits both, so observability now says `NodeHealthLabeled` on label and `NodeHealthUnlabeled` when the label is cleared on recovery.

**The `enabled` switch wording: leaving as is.** This document describes the final design, not the state of any one branch, which is a deliberate convention for these docs. The switch is real and lands in `mgmt-agent/values.yaml` with the implementation in [PR #6284](https://github.com/Azure/ARO-HCP/pull/6284), defaulted to `false`. Hedging the sentence here would make the document describe a transient intermediate state instead of the design, and it would need reverting the moment the implementation merges. The link target exists on this branch, so nothing is broken in the meantime.

raelga:
@copilot This is the second time this claim comes up, and it is still not right. The document matches upstream, so I am leaving it as is.

From `pkg/features/kube_features.go` on kubernetes/kubernetes master:

```go
PodReadyToStartContainersCondition: {
	{Version: version.MustParse("1.28"), Default: false, PreRelease: featuregate.Alpha},
	{Version: version.MustParse("1.29"), Default: true, PreRelease: featuregate.Beta},
	{Version: version.MustParse("1.37"), Default: true, PreRelease: featuregate.GA, LockToDefault: true}, // GA in 1.37, remove in 1.40
},
```

Beta and `Default: true` from 1.29, GA at 1.37. There is no 1.36 milestone for this gate. The GA version is not a guess either, it is the upstream declaration with the removal release next to it, which is exactly the kind of detail a reader of this section needs when the controller depends on the condition being present.

The live evidence agrees. The 1.35 management cluster reports `kubernetes_feature_enabled{name="PodReadyToStartContainersCondition",stage="BETA"} 1`, which is the beta-but-default-on state the document describes, and it is why the document says the condition is verified present before enablement rather than assumed.

raelga:
Pushed `c1d0694bd`: updated the success section to match what actually ships.

The controller no longer keeps an in-memory `successAt` map. Success is read from the same Pod list the detector already gets, so nothing survives a restart and the warm-up window is gone. That is the rule @stevekuznetsov set on this doc, the only cache allowed is one the `LIST` can rebuild.

Code change is in [#6378](https://github.com/Azure/ARO-HCP/pull/6378). I updated the doc here rather than after the fact so this PR describes the shipped design.

Sections touched: the success/state discussion, the restart paragraph, and the mitigation guardrail wording. No diagram changes.

raelga:
Pushed `e28294652` on top: three spots still described success as coming from the `PodReadyToStartContainers` condition alone. Success also comes off container start times on finished pods, so the informer box in the architecture diagram, the RBAC paragraph, and the corroborating-signals summary now say so. No behaviour claims changed, just the doc catching up with what the code reads.

PR #6300, opened 2026-07-28
feat: add controller that calculates Cluster Data Plane Identities extra information

miguelsorianod:
We add a controller that retrieves the Client ID and Principal ID associated to the Data Plane operators identities associated to the ARO-HCP Cluster.

We leverage the Service Managed Identity associated to the ARO-HCP Cluster to retrieve the Data Plane operators identities information. We use Azure Go SDK's UserAssignedIdentities API to retrieve it. This is a different method than what's done for MSI based identities where the Managed Identities Data Plane service is used instead (https://github.com/Azure/ARO-HCP/pull/3820). This is because for the control plane identities, on the environments where the managed identities data plane service is not available, we use the mi mock identity instead, which includes its clientid+principalid instead of the ones associated to the identities passed in the cluster payload. By using the mock managed identities data plane client we retrieve that transparently. We do that also because that identity/information is the one that needs to be used by the control plane operators themselves on the control plane side.


backend/pkg/controllers/fetch_data_plane_operators_managed_identities_info.go
```diff
@@ -0,0 +1,200 @@
...
+	return controller
+}
+
+func (c *fetchDataPlaneOperatorsManagedIdentitiesInfoSyncer) needsWork(cluster *api.HCPOpenShiftCluster) bool {
+	if cluster.ServiceProviderProperties.DeletionTimestamp != nil {
+		return false
+	}
+
+	return true
+}
```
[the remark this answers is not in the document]
miguelsorianod: There will always be dataplane operators configured. This does not apply

backend/pkg/controllers/fetch_data_plane_operators_managed_identities_info.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	replacement := existingServiceProviderCluster.DeepCopy()
+	replacement.Status.DataPlaneOperatorsManagedIdentities = make(map[string]*api.ServiceProviderClusterDataPlaneOperatorManagedIdentity)
+
+	smiResourceID := existingCluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ServiceManagedIdentity
+	userAssignedIdentitiesClient, err := c.smiClientBuilder.UserAssignedIdentitiesClient(ctx, existingCluster.ServiceProviderProperties.ManagedIdentitiesDataPlaneIdentityURL, smiResourceID, existingCluster.ID.SubscriptionID)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get User Assigned Identities Client: %w", err))
+	}
```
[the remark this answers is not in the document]
miguelsorianod: This should never be nil as it's a mandatory identity

backend/pkg/controllers/fetch_data_plane_operators_managed_identities_info.go
```diff
@@ -0,0 +1,200 @@
...
+
+	var syncErrors []error
+	// TODO as of now if there's any error processing the identity, if it's missing properties or any of the expected info we remove the identity
+	// from the SPC. This means that if for some reason the SPC had an entry with the info and the API stopped returning it we are deleting it.
+	// Alternatives could be:
+	// - Keep what's in the SPC for that entry in that case
+	// - Write what we can. This makes us having the PrincipalID and ClientID attributes as pointers instead.
+	for _, dataPlaneOperatorIdentityToSync := range identitiesToSync {
+		currentMI, err := userAssignedIdentitiesClient.Get(ctx, dataPlaneOperatorIdentityToSync.ResourceID.ResourceGroupName, dataPlaneOperatorIdentityToSync.ResourceID.Name, nil)
+		if err != nil {
```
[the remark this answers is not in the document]
miguelsorianod: We should always have a resourceid as it comes from validated api

backend/pkg/controllers/cluster/identity/fetch_data_plane_operators_managed_identities_info.go
```diff
@@ -0,0 +1,200 @@
...
+	return true
+}
+
+func (c *fetchDataPlaneOperatorsManagedIdentitiesInfoSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	existingCluster, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
+	if database.IsNotFoundError(err) {
+		return nil // cluster doesn't exist, no work to do
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
```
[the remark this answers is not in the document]
miguelsorianod: Waiting until design is agreed upon

backend/pkg/controllers/cluster/identity/fetch_data_plane_operators_managed_identities_info.go
```diff
@@ -0,0 +1,200 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: TODO decide file location

machi1990: This can go in https://github.com/miguelsorianod/ARO-HCP/blob/5227da6dd31b94b095ea39560598ea3d0f5e14dc/backend/pkg/controllers/cluster/properties/

internal/api/types_serviceprovider_cluster.go
```diff
@@ -201,6 +201,12 @@ type ServiceProviderClusterStatus struct {
 	// AzureResources tracks the lifecycle of Azure resources associated with
 	// the cluster, including deny assignments and the managed resource group.
 	AzureResources AzureResources `json:"azureResources,omitempty"`
+
+	// DataPlaneOperatorsManagedIdentities is a map of data plane operator managed identities.
```
deads2k: What is a data plane operator?  Provide a few examples.  Everyone seems to have a different definition of what these are.

internal/api/types_serviceprovider_cluster.go
```diff
@@ -249,6 +255,13 @@ type AzureReference struct {
 	EarliestRecheckTime *metav1.Time `json:"earliestRecheckTime,omitempty"`
 }
 
+type ServiceProviderClusterDataPlaneOperatorManagedIdentity struct {
+	OperatorName string                `json:"operatorName"`
+	ResourceID   *azcorearm.ResourceID `json:"resourceID,omitempty"`
+	ClientID     string                `json:"clientID,omitempty"`
+	PrincipalID  string                `json:"principalID,omitempty"`
```
deads2k: document what these are and whether they are input or output.

backend/pkg/controllers/cluster/identity/fetch_data_plane_operators_managed_identities_info.go
```diff
@@ -0,0 +1,196 @@
...
+
+	if !c.needsWork(existingCluster) {
+		return nil
+	}
+
+	// TODO unclear if we should put the data plane operators managed identities info in the ServiceProviderCluster resource or in
+	// the HCPCluster resource. For now we put it in the ServiceProviderCluster resource.
+	// Maybe on the Cluster's ServiceProviderProperties because it would ensure that we can leverage ETag to ensure we calculated from
+	// the content of the Cluster?
+	existingServiceProviderCluster, err := database.GetOrCreateServiceProviderCluster(ctx, c.resourcesDBClient, key.GetResourceID())
```
deads2k: just read from the cache.

backend/pkg/controllers/fetch_data_plane_operators_managed_identities_info.go
```diff
@@ -0,0 +1,196 @@
...
+		nil,
+		1*time.Minute,
+		syncer,
+	)
+
+	return controller
+}
+
+func (c *fetchDataPlaneOperatorsManagedIdentitiesInfoSyncer) needsWork(cluster *api.HCPOpenShiftCluster) bool {
+	return cluster.ServiceProviderProperties.DeletionTimestamp == nil
```
deads2k: how often do we really need to check this?  `EarliestRecheckTime *metav1.Time ` seems like a good fit here.

backend/pkg/controllers/cluster/identity/fetch_data_plane_operators_managed_identities_info.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get User Assigned Identities Client: %w", err))
+	}
+
+	var syncErrors []error
+	// TODO as of now if there's any error processing the identity, if it's missing properties or any of the expected info we remove the identity
+	// from the SPC. This means that if for some reason the SPC had an entry with the info and the API stopped returning it we are deleting it.
+	// Alternatives could be:
+	// - Keep what's in the SPC for that entry in that case
+	// - Write what we can. This makes us having the PrincipalID and ClientID attributes as pointers instead.
```
deads2k: I'm willing to start anywhere, but for the alternatives it's probably best to decide which outcome is likely to result in the best secure behavior possible in most cases.

backend/pkg/controllers/cluster/identity/fetch_data_plane_operators_managed_identities_info.go
deads2k: a controllers/cluster/identity directory seems logical in the new structure.

deads2k (changes requested):
needs earliest recheck time.  otherwise some doc needs updating and its  a reasonable start.

backend/pkg/controllers/cluster/identity/fetch_data_plane_operators_managed_identities_info.go
miguelsorianod: :+1: . Waiting to do the change until we decide whether we need to store it in ServiceProviderCluster or in Cluster's ServiceProviderProperties

internal/api/types_serviceprovider_cluster.go
miguelsorianod: I updated the doc in general. Let me know if it's now more clear

internal/api/types_serviceprovider_cluster.go
miguelsorianod: updated doc

backend/pkg/controllers/fetch_data_plane_operators_managed_identities_info.go
miguelsorianod: Added an EarliestRecheckTime. Converted the original map to a struct that contains the map + earliestrechecktime (the recheck is at the overall map level)

backend/pkg/controllers/cluster/identity/fetch_data_plane_operators_managed_identities_info.go
miguelsorianod: It depends on what do we understand by "best secure behavior":

What do we want to do when we find unexpected missing parts of the expected result or issues processing the result: we can keep what's before to not "unset" if there was something, or simply consider it as not existing, and at what granularity?

miguelsorianod:
> a controllers/cluster/identity directory seems logical in the new structure.
moved there :+1:

backend/pkg/controllers/cluster/identity/fetch_data_plane_operators_managed_identities_info.go
miguelsorianod: Changed to there

backend/pkg/controllers/cluster/identity/fetch_data_plane_operators_managed_identities_info.go
miguelsorianod: For now what I've done is:

* We only perform a Cosmos replace if we've been able to successfully call the Azure API for **all** identities.
* When the previous point occurs, we copy the value returned by the Azure API as-is for the ClientID and PrincipalID. At the Go type level they are string pointers so we are storing them in Cosmos as string pointers too
* For an identity Azure API call result, if the result is that the identity is not found then we still persist it in Cosmos but we set it with the ClientID and PrincipalID as nil.

This behavior is similar to what's done for the MSI-based identities. In that case, the MIDataplane service always returns a response even if the managed identity does not exist. When it does not exist the result is returned and the ClientID and PrincipalID that it returns is nil.
The all-or-nothing behavior is also similar to what's done for the MSI-based identities. For those, to save API calls we request all the information in a single call. For this PR, the UserAssignedIdentities client does not allow batching the request so we need to perform N calls, but we imitate the all-or-nothing approach of the MSI-based identities controller.

miguelsorianod:
Closing this in favor of https://github.com/Azure/ARO-HCP/pull/6592 which builds on top of this one with some additional refining changes

PR #6301, opened 2026-07-28
feat: add controller that retrieves information about MSI based identities

miguelsorianod:
We add a controller that retrieves the Client ID and Principal ID associated to the following identities associated to an ARO-HCP Cluster:
* The Control Plane operators identities
* The Service Managed Identity

We leverage Microsoft's Managed Identities Data Plane service to retrieve the information. When the service is not available (outside of AME tenants) the fake managed identities data plane client is leveraged which returns the information associated to the mock msi identity for all requests/responses to it. We do not directly use Azure Go SDK's UserAssignedIdentities client because otherwise we would return the information of clientid+principalid of the passed identities in the payload instead of the actual clientid+principalid that ends up being used in the management cluster side.


backend/pkg/controllers/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/Azure/ARO-HCP/backend/pkg/controllers/controllerutils"
+	"github.com/Azure/ARO-HCP/backend/pkg/informers"
+	"github.com/Azure/ARO-HCP/backend/pkg/listers"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+const fetchMSIIdentitiesInfoControllerName = "FetchMSIIdentitiesInfo"
```
miguelsorianod: Alternative name: FetchMIDataplaneBasedIdentitiesInfo

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: TODO decide where to place this file

machi1990: What do you think of https://github.com/miguelsorianod/ARO-HCP/tree/6778e3492763e31429168696ae9995407cfa98a8/backend/pkg/controllers/clusterpropertiescontroller package for placement of this controller and the one in https://github.com/Azure/ARO-HCP/pull/6300 ?

backend/pkg/controllers/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+//  2. Calls the Managed Identities Data Plane (or the fake client implementation
+//     in environments where the real Managed Identities Data Plane service
+//     is not available) once with the set of identities.
+//  3. Matches each returned credential by ResourceID (case-insensitive.
+//     ARM IDs are case-insensitive and response order is not assumed)
+//     and sets ClientID and PrincipalID when the dataplane returns
+//     non-empty values.
+//  4. Replaces the HCPCluster document only when the identity map changed.
+//
+// It does not stop early when ClientID/PrincipalID are already set, so
```
machi1990: 👍

backend/pkg/controllers/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+		// MI dataplane may return a different casing (and order) than Cosmos.
+		// Keep writing under the Cosmos key so we do not duplicate entries.
+		_, replacementIdentity, ok := c.findUserAssignedIdentityByResourceID(replacement.Identity.UserAssignedIdentities, credentialResourceID)
+		if !ok {
+			syncErrors = append(syncErrors, utils.TrackError(fmt.Errorf("unexpected Managed Identities Data Plane Credential %s Resource ID is not found in the cluster's identities", credentialResourceID)))
+			continue
+		}
+
+		// TODO should we check if existingCluster/replacementIdentity.Identity is nil and initialize it?  or are we guaranteed that after Frontend stores to cosmos
+		// that section is not nil?
```
machi1990: https://github.com/miguelsorianod/ARO-HCP/blob/6778e3492763e31429168696ae9995407cfa98a8/frontend/pkg/frontend/cluster.go#L390 guarantees that it's never nil

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+const fetchMSIIdentitiesInfoControllerName = "FetchMSIIdentitiesInfo"
+
+// fetchMSIIdentitiesInfoSyncer fetches ClientID and PrincipalID for the
+// cluster's MSI-based user-assigned managed identities and writes them onto
+// HCPOpenShiftCluster.Identity.UserAssignedIdentities in Cosmos.
```
machi1990: https://github.com/Azure/ARO-HCP/pull/6300 for DP identities stores the field in ServiceProviderCluster: https://github.com/miguelsorianod/ARO-HCP/blob/61e2bf3a399569767391971651346b9440559eb3/backend/pkg/controllers/fetch_data_plane_operators_managed_identities_info.go#L38 

Meaning that consumer of these info e.g https://github.com/Azure/ARO-HCP/pull/6269 has to read these info from two objects, this adds congnitive load; can we consider store both CP, DP, SMI extra info in one cosmo object i.e ServiceProviderCluster in this case?

And when presenting the info to the API for the .Identity.UserAssignedIdentities we read from the extra info from the `ServiceProviderCluster`

Thoughts @miguelsorianod @deads2k ?

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
machi1990: Edit: with the re-arrangement of the controllers, the new pkg name is https://github.com/miguelsorianod/ARO-HCP/blob/5227da6dd31b94b095ea39560598ea3d0f5e14dc/backend/pkg/controllers/cluster/properties/

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
deads2k: The decision to store on `Cluster` object is based on "is the information needed to reply to the user when reading from ARM".  If the answer is yes, then the field goes on the `Cluster` object.  If the answer is no, then the field goes on the `ServiceProviderCluster` object.

backend/pkg/controllers/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+		nil,
+		1*time.Minute,
+		syncer,
+	)
+
+	return controller
+}
+
+func (c *fetchMSIIdentitiesInfoSyncer) needsWork(existingCluster *api.HCPOpenShiftCluster) bool {
+	return existingCluster.ServiceProviderProperties.DeletionTimestamp == nil
```
deads2k: needs EarliestRecheckTime. We have enough of these, I'm willing to consider a `ControllerToEarliestRecheckTime map[string]*metav1.Time` on ServiceProviderCluster.

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+	var identitiesToSyncResourceIDStrs []string
+	for identityResourceIDStr := range existingCluster.Identity.UserAssignedIdentities {
+		identitiesToSyncResourceIDStrs = append(identitiesToSyncResourceIDStrs, identityResourceIDStr)
+	}
+
+	// On environments where the real Managed Identities Data Plane service is not available, a
+	// fake implementation of the Managed Identities Data Plane client is used, which always returns the same information and
+	// same set of credentials for all requests, independently on what identity is request. The returned information is
+	// the information associated to the "MI Mock" identity.
+	fpaMIDataplaneClient, err := c.fpaMIdataplaneClientBuilder.ManagedIdentitiesDataplane(existingCluster.ServiceProviderProperties.ManagedIdentitiesDataPlaneIdentityURL)
```
deads2k, line 263: require this in needswork.

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	// We get all the Managed Identities information in a single Managed Identities Data Plane Credentials to minimize
+	// calls to the Managed Identities Data Plane Service.
+	fpaMIDataplaneCredentialsRequest := dataplane.UserAssignedIdentitiesRequest{IdentityIDs: identitiesToSyncResourceIDStrs}
+	fpaMIDataplaneCredentials, err := fpaMIDataplaneClient.GetUserAssignedIdentitiesCredentials(ctx, fpaMIDataplaneCredentialsRequest)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get Managed Identities Data Plane Credentials: %w", err))
+	}
+	if len(fpaMIDataplaneCredentials.ExplicitIdentities) == 0 {
```
deads2k: seems like the check below covers this and is more clear.

backend/pkg/controllers/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+	if len(fpaMIDataplaneCredentials.ExplicitIdentities) != len(identitiesToSyncResourceIDStrs) {
+		return utils.TrackError(fmt.Errorf("unexpected number of Managed Identities Data Plane Credentials. Expected: %d, Received: %d", len(identitiesToSyncResourceIDStrs), len(fpaMIDataplaneCredentials.ExplicitIdentities)))
+	}
+
+	// TODO at some point we will also have to implement logic that retrieves the initial set of credentials for the
+	// control plane operators managed identities and for the service managed identity and store it in the Managed
+	// Identities Key Vault (a Management Cluster scoped resource). Do we want to do it here at the same time because
+	// we are already calling the Managed Identities Data Plane Service and getting credentials here? As relevant context,
+	// these set of initial credentials should be stored in the Managed Identities Key Vault before creating the HostedCluster
+	// and those credentials have a limited lifespan (unknown which without investigating further).
```
deads2k: these look like separate concerns to me

1. list of identities and information
2. credentials for those identities

and we'd want separate consistency check frequency, error handling, and retries.

backend/pkg/controllers/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+
+	return errors.Join(syncErrors...)
+}
+
+// findUserAssignedIdentityByResourceID looks up an identity using a
+// case-insensitive ResourceID match. ARM resource IDs are case-insensitive
+// and the MI dataplane may return a different casing than Cosmos.
+// It returns the Cosmos map key (preserving stored casing), the matching
+// identity value (or nil if not found), and whether a match was found.
+func (c *fetchMSIIdentitiesInfoSyncer) findUserAssignedIdentityByResourceID(identities map[string]*arm.UserAssignedIdentity, resourceIDStr string) (string, *arm.UserAssignedIdentity, bool) {
```
deads2k: super ugly. Can we normalize into lowercase in a future step?

backend/pkg/controllers/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,234 @@
...
+			syncErrors = append(syncErrors, utils.TrackError(fmt.Errorf("unexpected Managed Identities Data Plane Credential %s Resource ID is not found in the cluster's identities", credentialResourceID)))
+			continue
+		}
+
+		// TODO should we check if existingCluster/replacementIdentity.Identity is nil and initialize it?  or are we guaranteed that after Frontend stores to cosmos
+		// that section is not nil?
+		// TODO as of now if the returned information from the MIDataplane has nil/empty ClientID/PrincipalID we don't set it in the replacement. Do
+		// we want to follow that approach or 1:1 set what's returned from the MIDataplane? That means that if for some reason it's set and the MIDataplane
+		// stops setting it we would be unsetting it too.
+		if fpaMIDataplaneCredential.ClientID != nil && len(*fpaMIDataplaneCredential.ClientID) > 0 {
```
deads2k: code would be a lot easier to read if you checked this and the principalID at the top of the for loop and errored and continued.  Is it really valuable to only write half of it?  Seems like it would just fail in a different spot.

When there is no known value, why leave old data versus clearing the data?  You chose the opposite path in your other identity controller.  I'm inclined to be consistent.

deads2k (changes requested):

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
miguelsorianod: This has been moved to pkg/controllers/cluster/identity

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,271 @@
...
+	// the information associated to the "MI Mock" identity.
+	fpaMIDataplaneClient, err := c.fpaMIdataplaneClientBuilder.ManagedIdentitiesDataplane(existingCluster.ServiceProviderProperties.ManagedIdentitiesDataPlaneIdentityURL)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get Managed Identities Data Plane Client: %w", err))
+	}
+
+	// We get all the Managed Identities information in a single Managed Identities Data Plane Credentials request to minimize
+	// calls to the Managed Identities Data Plane Service.
+	fpaMIDataplaneCredentialsRequest := dataplane.UserAssignedIdentitiesRequest{IdentityIDs: identitiesToSyncResourceIDStrs}
+	fpaMIDataplaneCredentials, err := fpaMIDataplaneClient.GetUserAssignedIdentitiesCredentials(ctx, fpaMIDataplaneCredentialsRequest)
```
machi1990, line 271: When we resync here; what happens to the credentials already stored in KV will they continue to work?

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,271 @@
...
+				ptr.Deref(fpaMIDataplaneCredential.ObjectID, ""),
+			))
+		}
+		credentialResourceID := *fpaMIDataplaneCredential.ResourceID
+
+		// Match case-insensitively: ARM resource IDs are case-insensitive and the
+		// MI dataplane may return a different casing than Cosmos, as well as different order than
+		// how it's been requested.
+		// We do not store the Resource ID lowercased because the resource id ends up exposed to the end-user in the Cluster payload API response
+		// in the `identity` section and we want to preserve the casing as received from the original request.
```
deads2k: > we want to preserve the casing as received from the original request
Is this required.  The code will be a  lot cleaner if you don't do this.

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,271 @@
...
+		return utils.TrackError(fmt.Errorf("failed to get Managed Identities Data Plane Credentials: %w", err))
+	}
+
+	if len(fpaMIDataplaneCredentials.ExplicitIdentities) != len(identitiesToSyncResourceIDStrs) {
+		return utils.TrackError(fmt.Errorf("unexpected number of Managed Identities Data Plane Credentials. Expected: %d, Received: %d", len(identitiesToSyncResourceIDStrs), len(fpaMIDataplaneCredentials.ExplicitIdentities)))
+	}
+
+	replacement := existingCluster.DeepCopy()
+
+	for idx, fpaMIDataplaneCredential := range fpaMIDataplaneCredentials.ExplicitIdentities {
```
deads2k, line 284: this loop doesn't guarantee that missing elements are cleared in the existing map.  Write it so it does.

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
miguelsorianod: This information is provided in the `identity` section, and the resource ids  are provided by the end-user. My understanding is that we need to preserve the casing because of ARM requirements. cc @bennerv can you confirm this or point to where it's documented, if anywhere?

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,271 @@
...
+
+	earliestRecheckTime := existingCluster.ServiceProviderProperties.MSIIdentitiesEarliestRecheckTime
+	if earliestRecheckTime != nil && c.clock.Now().Before(earliestRecheckTime.Time) {
+		return false
+	}
+
+	return true
+}
+
+// TODO do we actually want to implement continuous syncing of the identities as of now? Changing this over time
```
miguelsorianod: This is very relevant to the current implementation in this PR.
This means for example that if someone were to recreate an identity with the same resource id, this controller would reconcile it and other controllers that depend on this would reconcile the changes, which would be send downstream where changing the identities is not supported (CS at least doesn't currently support it), with whatever effects would have attempting to change the data there (denied api responses, persisting the new values but not really passing it down the line ending up with inaccurate data they indicate is in use and so on)

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
miguelsorianod: The elements that we request to the MI DP are the elements that are originally in the `identity` section. That section and the resource ids within it(the keys of the map) are written by the user (or derived by the frontend from customerProperties.platform.operatorsAuthentication.userAssignedIdentities.{ControlPlaneOperators,ServiceManagedIdentity} which is controlled by the user. If those were to change it would be because of an user-initiated change in the frontend. This includes removals.
This means that we always request what's currently in the `identity` section to the MI DP service.

As a side note: we currently do not support changing the identities at the API level. This is related to the other question in https://github.com/Azure/ARO-HCP/pull/6301/changes#diff-a8acfaa5778ff377026d3035895fe20c0a629560b3f830a51c5e6ec6aebf88ffR146. Support for that still needs to be implemented. The change needed is not only relaxing the restriction at API level, but also doing the necessary changes that come as a consequence of that (issuing of new credentials, propagating the changes to different components (CS, Hypershift, OCP, ...), those components implementing the support themselves and so on)

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
miguelsorianod: What about ETag concerns? For the data plane operators identities (the other PR), based on that criteria we would put it the extra information associated to the resource ids in ServiceProviderCluster. However, at that point because the original resource ids come from the `Cluster` resource, if we have its associated extra information in the `ServiceProviderCluster`, we would have no guarantee that the information is consistent between the Cosmos resources. If we were to place them in the Cluster's ServiceProviderProperties in that case we would ensure that the data is consistent because the replace would fail if the resource ids have changed (which are set in .customerProperties.platform.operatorsAuthentication.userAssignedIdentities of the `Cluster` type).

Isn't in that case better to put it in the `Cluster` object, even when it's not exposed to the user?

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
machi1990: The credentials in KV should continue to work.

miguelsorianod: to complete:
* New credentials are generated each time
* Previous creds are not invalidated, they continue to be valid until they expire
* Each credentials are generated with an expiration of 90d since cert credential was generated.

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -219,6 +219,103 @@ type ServiceProviderClusterStatus struct {
...
+	// Controllers should set this field with substantial jitter: without another
+	// concern, jitter of 50% is considered normal so that any storms are quickly
+	// dissipated. Additionally, long recheck times are recommended for resources
+	// outside of their active phases. Order of at least six hours is, with
+	// durations up to 24 hours considered normal.
+	// Written by: FetchMSIIdentitiesInfo
+	EarliestRecheckTime *metav1.Time `json:"earliestRecheckTime,omitempty"`
+	// ControlPlaneOperatorsIdentities is a map containing extra information about
+	// the Managed Service Identity (MSI) based Azure User-Assigned Managed Identities
+	// for the cluster's control plane operators.
```
deads2k: where will these pods run?

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -219,6 +219,103 @@ type ServiceProviderClusterStatus struct {
...
+	// and ServiceManagedIdentity.
+	// This allows the controller to avoid repeatedly hitting an Azure API to
+	// recheck that the desired state is true.
+	// Controllers should set this field with substantial jitter: without another
+	// concern, jitter of 50% is considered normal so that any storms are quickly
+	// dissipated. Additionally, long recheck times are recommended for resources
+	// outside of their active phases. Order of at least six hours is, with
+	// durations up to 24 hours considered normal.
+	// Written by: FetchMSIIdentitiesInfo
+	EarliestRecheckTime *metav1.Time `json:"earliestRecheckTime,omitempty"`
```
deads2k, line 281: I still predict we'll want a shared map for this, but I'll allow it to start.

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -219,6 +219,103 @@ type ServiceProviderClusterStatus struct {
...
+	// ControlPlaneOperatorsIdentities is a map containing extra information about
+	// the Managed Service Identity (MSI) based Azure User-Assigned Managed Identities
+	// for the cluster's control plane operators.
+	// The key is the fully lowercased Azure Resource ID of the identity.
+	// Written by: FetchMSIIdentitiesInfo
+	ControlPlaneOperatorsIdentities map[string]*ServiceProviderClusterControlPlaneOperatorIdentity `json:"controlPlaneOperatorsIdentities,omitempty"`
+	// ServiceManagedIdentity holds resolved ClientID/PrincipalID for the cluster's
+	// service managed identity.
```
deads2k, line 291: useless description. What uses this later, where does it run, why is it special, what does it do?

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -219,6 +219,103 @@ type ServiceProviderClusterStatus struct {
...
+	// OperatorName is the name of the control plane operator associated to the Azure User Assigned Managed Identity.
+	// The set of recognized operator names by the service is located in internal/azure/cluster_scoped_identities_config.go.
+	// Its value comes from the Cluster's CustomerProperties.
+	OperatorName string `json:"operatorName,omitempty"`
+	// ResourceID is the Azure Resource ID of the Azure User Assigned Managed Identity that is associated to the corresponding
+	// control plane operator (keyed by OperatorName). Its value comes from the Cluster's CustomerProperties.
+	// The ControlPlaneOperatorsIdentities map key is the fully lowercased form
+	// of this ID used for lookups.
+	// Its value comes from the Cluster's CustomerProperties.
+	ResourceID *azcorearm.ResourceID `json:"resourceId,omitempty"`
```
deads2k, line 306:
```suggestion
	UserManagedIdentityResourceID *azcorearm.ResourceID `json:"resourceId,omitempty"`
```

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -219,6 +219,103 @@ type ServiceProviderClusterStatus struct {
...
+	// It may be be nil or empty.
+	PrincipalID *string `json:"principalId,omitempty"`
+}
+
+// ServiceProviderClusterServiceManagedIdentity is the resolved metadata for the
+// cluster's service managed identity.
+type ServiceProviderClusterServiceManagedIdentity struct {
+	// ResourceID is the Azure Resource ID of the Azure User Assigned Managed Identity that is associated to the cluster's Service Managed Identity.
+	// Its value comes from the Cluster's CustomerProperties. Its value comes from the Cluster's CustomerProperties.
+	ResourceID *azcorearm.ResourceID `json:"resourceId,omitempty"`
```
deads2k, line 322:
```suggestion
	UserManagedIdentityResourceID *azcorearm.ResourceID `json:"resourceId,omitempty"`
```

backend/pkg/controllers/cluster/identity/cluster_identity_sync.go
```diff
@@ -0,0 +1,204 @@
...
+	if cosmosstorageutils.IsNotFoundError(err) {
+		return nil // cluster doesn't exist, no work to do
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
+	}
+	// check if we need to do work again. Sometimes the live data is ahead of the cache and obviates the need to do any work
+	if !c.NeedsWork(ctx, existingCluster) {
+		return nil
+	}
```
deads2k, line 132: no reason for live gets

backend/pkg/controllers/cluster/identity/cluster_identity_sync.go
```diff
@@ -0,0 +1,204 @@
...
+
+	logger.Info("synced identity information from ServiceProviderCluster")
+	return nil
+}
+
+// updateIdentityUserAssignedIdentitiesFromSPC walks the existing Identity map and, for
+// each key, looks up the lowercased resource ID in SPC control-plane operator
+// identities or the service managed identity. When found, ClientID and
+// PrincipalID are updated in place. Keys missing from SPC are left as-is.
+func (c *clusterIdentitySyncer) updateIdentityUserAssignedIdentitiesFromSPC(
```
deads2k, line 172: don't abbreviate SPC

backend/pkg/controllers/cluster/identity/cluster_identity_sync.go
```diff
@@ -0,0 +1,204 @@
...
+			clientID = spcIdentity.ClientID
+			principalID = spcIdentity.PrincipalID
+			// If we found the identity in the SPC service managed identity, we use the ClientID and PrincipalID from the SPC.
+		} else if spcServiceManagedIdentity != nil && spcServiceManagedIdentity.ResourceID != nil &&
+			strings.ToLower(spcServiceManagedIdentity.ResourceID.String()) == lowerResourceIDStr {
+			clientID = spcServiceManagedIdentity.ClientID
+			principalID = spcServiceManagedIdentity.PrincipalID
+		} else { // otherwise, we leave the identity as-is
+			continue
+		}
```
deads2k, line 193: refactor to switch/case please

backend/pkg/controllers/cluster/identity/cluster_identity_sync.go
```diff
@@ -0,0 +1,204 @@
...
+	identityUserAssignedIdentities map[string]*coreapi.UserAssignedIdentity,
+	spcControlPlaneOperatorsIdentities map[string]*coreapi.ServiceProviderClusterControlPlaneOperatorIdentity,
+	spcServiceManagedIdentity *coreapi.ServiceProviderClusterServiceManagedIdentity,
+) {
+	for identityResourceIDStr := range identityUserAssignedIdentities {
+		lowerResourceIDStr := strings.ToLower(identityResourceIDStr)
+
+		var clientID, principalID *string
+
+		// If we found the identity in the SPC control plane operators identities, we use the ClientID and PrincipalID from the SPC.
```
deads2k, line 182: don't abbrevaite spc throughout

backend/pkg/controllers/cluster/identity/cluster_identity_sync.go
```diff
@@ -0,0 +1,204 @@
...
+		if spcIdentity, ok := spcControlPlaneOperatorsIdentities[lowerResourceIDStr]; ok && spcIdentity != nil {
+			clientID = spcIdentity.ClientID
+			principalID = spcIdentity.PrincipalID
+			// If we found the identity in the SPC service managed identity, we use the ClientID and PrincipalID from the SPC.
+		} else if spcServiceManagedIdentity != nil && spcServiceManagedIdentity.ResourceID != nil &&
+			strings.ToLower(spcServiceManagedIdentity.ResourceID.String()) == lowerResourceIDStr {
+			clientID = spcServiceManagedIdentity.ClientID
+			principalID = spcServiceManagedIdentity.PrincipalID
```
deads2k, line 190: so this is some kind of a default? Really, that's shocking.

backend/pkg/controllers/cluster/identity/cluster_identity_sync.go
```diff
@@ -0,0 +1,204 @@
...
+		} else { // otherwise, we leave the identity as-is
+			continue
```
deads2k, line 192: don't like this.  We should clear it.  There was a way to express empty before.  I don't recall if it was nil or zero-value reference, but we allowed it.  Leave the key, but clear the value.

backend/pkg/controllers/cluster/identity/cluster_identity_sync.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		// We initialize the identity if it does not exist, to avoid nil dereferencing
+		if identityUserAssignedIdentities[identityResourceIDStr] == nil {
+			identityUserAssignedIdentities[identityResourceIDStr] = &coreapi.UserAssignedIdentity{}
+		}
+		// We update the identity with the ClientID and PrincipalID values the SPC. Those values themselves
+		// maybe be nil or empty strings.
+		identityUserAssignedIdentities[identityResourceIDStr].ClientID = clientID
+		identityUserAssignedIdentities[identityResourceIDStr].PrincipalID = principalID
```
deads2k, line 202: why not just always do 
```
identityUserAssignedIdentities[identityResourceIDStr] = &coreapi.UserAssignedIdentity{
clientID: clientID, 
principlalID: principalID,
}
```


Also, inline into the switch/case above.

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,397 @@
...
+// the identities stored on SPC still match desiredIdentitiesToFetch; on
+// mismatch (or if future skip-prerequisites fail), it returns true immediately.
+// When identities match, it returns false while EarliestRecheckTime is in the
+// future, and true when EarliestRecheckTime is nil or already past. Callers
+// must skip needsWork entirely when the cluster is deleting.
+func (c *fetchMSIIdentitiesInfoSyncer) needsWork(existingSPC *coreapi.ServiceProviderCluster, desiredIdentitiesToFetch *msiBasedIdentitiesToFetch) bool {
+	// Only honor EarliestRecheckTime when the desired identity set still matches
+	// SPC. Any mismatch (or future "must work now" conditions added alongside
+	// this check) should fall through to return true and query the dataplane.
+	if c.desiredMSIResourceIDsMatchSPC(desiredIdentitiesToFetch, existingSPC) {
```
deads2k, line 179: don't abbreviate spc

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,397 @@
...
+		if !ok {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (c *fetchMSIIdentitiesInfoSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	existingCluster, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
```
deads2k, line 227: use lister. eventual consistency is fine.

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,397 @@
...
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get Cluster: %w", err))
+	}
+
+	if existingCluster.ServiceProviderProperties.DeletionTimestamp != nil {
+		return nil
+	}
+
+	spcCRUD := c.resourcesDBClient.ServiceProviderClusters(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	existingSPC, err := spcCRUD.Get(ctx, coreapi.ServiceProviderClusterResourceName)
```
deads2k, line 240: same, no reason for live get.

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,397 @@
...
+	earliestRecheckAt := metav1.NewTime(c.clock.Now().Add(wait.Jitter(
+		msiIdentitiesRecheckInterval,
+		msiIdentitiesRecheckJitter,
+	)))
+	replacement.Status.MSIManagedIdentities.EarliestRecheckTime = &earliestRecheckAt
+
+	controlPlaneOperatorsUnchanged := equality.Semantic.DeepEqual(replacement.Status.MSIManagedIdentities.ControlPlaneOperatorsIdentities, existingSPC.Status.MSIManagedIdentities.ControlPlaneOperatorsIdentities)
+	serviceManagedIdentityUnchanged := equality.Semantic.DeepEqual(replacement.Status.MSIManagedIdentities.ServiceManagedIdentity, existingSPC.Status.MSIManagedIdentities.ServiceManagedIdentity)
+	recheckUnchanged := equality.Semantic.DeepEqual(replacement.Status.MSIManagedIdentities.EarliestRecheckTime, existingSPC.Status.MSIManagedIdentities.EarliestRecheckTime)
+	if controlPlaneOperatorsUnchanged && serviceManagedIdentityUnchanged && recheckUnchanged {
```
deads2k, line 348: why not just do the easy deepequal on everything?

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,397 @@
...
+		return nil
+	}
+	if err != nil {
+		// Same as precondition failure: EarliestRecheckTime was not
+		// persisted, so needsWork will still see the previously persisted value.
+		return utils.TrackError(fmt.Errorf("failed to replace ServiceProviderCluster: %w", err))
+	}
+
+	return nil
+}
```
deads2k, line 365: acceptable controller behavior.  Better be right about the claims of the azure API.  Guess we'll find out.

deads2k (changes requested):

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -219,6 +219,103 @@ type ServiceProviderClusterStatus struct {
...
+}
+
+// ServiceProviderClusterControlPlaneOperatorIdentity is the resolved metadata for a
+// single Managed Service Identity (MSI) based Azure User-Assigned Managed Identity
+// bound to a control plane operator.
+type ServiceProviderClusterControlPlaneOperatorIdentity struct {
+	// OperatorName is the name of the control plane operator associated to the Azure User Assigned Managed Identity.
+	// The set of recognized operator names by the service is located in internal/azure/cluster_scoped_identities_config.go.
+	// Its value comes from the Cluster's CustomerProperties.
+	OperatorName string `json:"operatorName,omitempty"`
```
miguelsorianod: We cannot introduce the current design as is regarding OperatorName:

Although right now we enforce that each identity is used in one operator only, there are plans to relax that restriction. This would make this fall short.

The options at that point that I can think of are:
* Change this to OperatorNames []string. However, then further questions arise around what controller would be the responsible of updating them. The fetch msi identities info would have more responsibilities which would be also detecting changes on operatornames and have unecessary checks against Azure when only the operatornames change. Alternatives would be writing their own dedicated controllers but then we would have two controllers that potentially write the same parent entries
* Do not introduce OperatorName at all. In that case we will lose the ability to correlate the identity resource id to the operator name from the information in the SPC itself. It should be possible to iterater over the Cluster's properties and see what operator names match the resourceids, but that's interaction with the other type
* Key by operator name instead of resourceid at the SPC level. However, that has other downsides where `identity` doesn't contain the operatorName, as well as we would end up with duplicated resourceids in the spc across operators entries

backend/pkg/controllers/cluster/identity/fetch_msi_identities_info.go
```diff
@@ -0,0 +1,397 @@
...
+	// resolved. Combined with msiIdentitiesRecheckJitter via wait.Jitter.
+	msiIdentitiesRecheckInterval = 12 * time.Hour
+	msiIdentitiesRecheckJitter   = 0.5
+)
+
+// controlPlaneOperatorIdentityToFetch describes one control-plane operator MSI based
+// identity for which extra information should be fetched via the Managed Identities Data Plane.
+type controlPlaneOperatorIdentityToFetch struct {
+	// operatorName is the name of the control-plane operator this identity is bound to.
+	operatorName string
```
miguelsorianod: This has the same issue as described in https://github.com/Azure/ARO-HCP/pull/6301/changes#r3784928555

internal/api/coreapi/types_serviceprovider_cluster.go
miguelsorianod: Even in the case of the current PR where OperatorName is a single string, that's not good enough either because at some point the resourceid can be associated to a different operator

miguelsorianod: What we ended up deciding is to remove OperatorName. Which operators are associated to each resource id  is tracked on the Cluster document: Cluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ControlPlaneOperators.

and it will need to be reconstructed by combining both documents information

miguelsorianod:
Closing this in favor of https://github.com/Azure/ARO-HCP/pull/6589 which builds on top of this one with some additional refining changes

PR #6304, opened 2026-07-28
[wip] allow specifying a set of minimum versions

deads2k:
And provide an admin API to set them.

/hold  I haven't reviewed this yet.


backend/pkg/controllers/cluster/version/minimum_version_override.go
```diff
@@ -0,0 +1,98 @@
...
+// a version that satisfies the minimum requirements.
+//
+// Rules (from the field documentation):
+//  1. If a minimum exists for the selected version's minor and the selected version
+//     is already >= that minimum, keep the selected version.
+//  2. If a minimum exists for the selected version's minor but the selected version
+//     is < the minimum, return the minimum instead.
+//  3. If any minimum targets a minor AHEAD of the cluster's current minor (from
+//     activeVersions), return that minimum directly (forces y-stream upgrade).
+//     When multiple higher-minor minimums exist, use the lowest one (one step at a time).
```
machi1990, line 36: In case this happens; We'll change the customers' desired version from X+Y to X.Y+1. 
However, with the way that's done, this happens only in the ServiceProviderCluster object and not the HCPOpenShift object, is that intentional? @deads2k

backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go
```diff
@@ -209,6 +216,10 @@ func (c *controlPlaneDesiredVersionSyncer) SyncOnce(ctx context.Context, key con
 		return utils.TrackError(err)
 	}
 
+	// Apply SRE minimum version override.
+	minimumVersions := cachedServiceProviderCluster.Spec.ControlPlaneVersion.MinimumVersions
+	desiredVersion = applyMinimumVersionOverride(desiredVersion, activeVersions, minimumVersions)
```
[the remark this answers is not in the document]
machi1990: +1 this makes sense

admin/server/handlers/hcp/minimumversions.go
```diff
@@ -0,0 +1,111 @@
...
+		return err
+	}
+
+	existing, err := database.GetOrCreateServiceProviderCluster(request.Context(), h.resourcesDBClient, resourceID)
+	if err != nil {
+		return fmt.Errorf("failed to get ServiceProviderCluster: %w", err)
+	}
+
+	replacement := existing.DeepCopy()
+	replacement.Spec.ControlPlaneVersion.MinimumVersions = parsed
```
machi1990, line 72: We probably should validate that the minimum versions are compatible with the cluster 
e.g reject if there is a node pool in a minor higher than the minimums? (node pool vs control plane skew checks)

admin/server/handlers/minimumversions.go
```diff
@@ -0,0 +1,112 @@
...
+
+	iterator, err := h.resourcesDBClient.ResourcesGlobalListers().ServiceProviderClusters().List(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("failed to list ServiceProviderClusters: %w", err)
+	}
+
+	updatedCount := 0
+	for _, spc := range iterator.Items(ctx) {
+		replacement := spc.DeepCopy()
+		replacement.Spec.ControlPlaneVersion.MinimumVersions = parsed
```
machi1990, line 71: same comment as https://github.com/Azure/ARO-HCP/pull/6304/changes#r3681148568

admin/server/handlers/minimumversions.go
```diff
@@ -0,0 +1,112 @@
...
+
+type allClustersMinimumVersionsRequest struct {
+	MinimumVersions []string `json:"minimumVersions"`
+}
+
+type allClustersMinimumVersionsResponse struct {
+	UpdatedCount int `json:"updatedCount"`
+}
+
+func (h *AllClustersMinimumVersionsHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) error {
```
machi1990, line 49: Is the intention of this handler to solve the rollback of a z-level use case?

admin/server/server/admin.go
```diff
@@ -126,9 +126,15 @@ func NewAdminAPI(
...
 		hcpMiddleware.HandlerFunc(errorutils.ReportError(hcp.NewHCPDesiredControlPlaneSizeHandler(resourcesDBClient).ServeHTTP)),
 	)
+	middlewareMux.Handle(
+		middleware.V1HCPResourcePattern("POST", "/minimumversions"),
+		hcpMiddleware.HandlerFunc(errorutils.ReportError(hcp.NewHCPMinimumVersionsHandler(resourcesDBClient).ServeHTTP)),
+	)
 
 	// Non-HCP admin routes
 	middlewareMux.Handle("GET /admin/helloworld", handlers.HelloWorldHandler())
+	middlewareMux.Handle("POST /admin/v1/minimumversions",
```
machi1990, line 136: `/minimumversions`handler applies the minimum version to all HCPs, why is the endpoint placed in non-hcp?

Or should we name it like

```suggestion
	middlewareMux.Handle("POST /admin/v1/hcpsminimumversions",
```
?

PR #6308, opened 2026-07-28
feat: add UserFacingConditions to Cluster, NodePool, and ExternalAuth status

miguelsorianod:
Add a status.userFacingConditions field ([]metav1.Condition) on the internal Cluster, NodePool, and ExternalAuth documents.

These conditions are intended for ARM-facing status and must not include internal details. This also means that Type and Reason become part of the public API.	Addition of new conditions here should be done only when strictly necessary, sparingly and only done when there is a clear benefit to doing so. We expect the number of conditions at this level to be kept to a minimum.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/internal/api/types_cluster.go
+++ b/internal/api/types_cluster.go
@@ -51,6 +51,21 @@ type HCPOpenShiftClusterStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// UserFacingConditions is a list of conditions that tracks user-facing cluster
+	// conditions. Each Condition Type should be unique among all conditions.
+	// The conditions here are exposed to the ARM API. This means that UserFacingConditions
+	// must not contain any internal details. This also means the Type and Reason
+	// values become part of the public API.
+	// Addition of new conditions here should be done only when strictly necessary, sparingly and only done
+	// when there is a clear benefit to doing so. We expect the number of conditions at this
+	// level to be kept to a minimum.
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	UserFacingConditions []metav1.Condition `json:"userFacingConditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 }
 
 var _ arm.CosmosPersistable = &HCPOpenShiftCluster{}
--- a/internal/api/v20260630preview/external_auth_methods.go
+++ b/internal/api/v20260630preview/external_auth_methods.go
@@ -304,7 +304,7 @@ func (v version) NewHCPOpenShiftClusterExternalAuth(from *api.HCPOpenShiftCluste
 			SystemData: api.PtrOrNil(newSystemData(from.SystemData)),
 			Properties: &generated.ExternalAuthProperties{
 				ProvisioningState: api.PtrOrNil(generated.ExternalAuthProvisioningState(from.Properties.ProvisioningState)),
-				Status:            newResourceStatus(from.Status.Conditions),
+				Status:            api.PtrOrNil(newExternalAuthResourceStatus(&from.Status)),
 				Issuer:            api.PtrOrNil(newTokenIssuerProfile(&from.Properties.Issuer)),
 				Claim:             api.PtrOrNil(newExternalAuthClaimProfile(&from.Properties.Claim)),
 			},
@@ -320,4 +320,13 @@ func (v version) NewHCPOpenShiftClusterExternalAuth(from *api.HCPOpenShiftCluste
 		})
 	}
 	return out
+}
+
+func newExternalAuthResourceStatus(from *api.HCPOpenShiftClusterExternalAuthStatus) generated.ResourceStatus {
+	if from == nil {
+		return generated.ResourceStatus{}
+	}
+	return generated.ResourceStatus{
+		Conditions: newConditions(from.UserFacingConditions),
+	}
 }
--- a/internal/api/v20260630preview/hcpopenshiftclusters_methods.go
+++ b/internal/api/v20260630preview/hcpopenshiftclusters_methods.go
@@ -261,14 +261,35 @@ func newKmsKey(from *api.KmsKey) generated.KmsKey {
 }
 
 func newConditions(from []metav1.Condition) []*generated.Condition {
-	// TODO (bvesel): propose a method for how we want to translate our internal metav1.Conditions type to this external
-	// type.  Until then, we'll plumb it with empty data.
-	return []*generated.Condition{}
+	if from == nil {
+		return nil
+	}
+
+	out := make([]*generated.Condition, 0, len(from))
+	for i := range from {
+		c := from[i]
+		cond := &generated.Condition{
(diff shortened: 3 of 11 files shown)


internal/api/types_cluster.go
```diff
@@ -51,6 +51,21 @@ type HCPOpenShiftClusterStatus struct {
...
+	// values become part of the public API.
+	// Addition of new conditions here should be done only when strictly necessary, sparingly and only done
+	// when there is a clear benefit to doing so. We expect the number of conditions at this
+	// level to be kept to a minimum.
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	UserFacingConditions []metav1.Condition `json:"userFacingConditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
miguelsorianod, line 68: I considered naming it `userConditions` but maybe that would be a bit confusing with the concept of "user" of the cluster or something so I left it as `userFacingConditions`. I am open to changing this or other suggestions if we want

miguelsorianod, line 68: I also considered placing this in `.serviceProviderProperties`. Conceptually, this is evaluated state so to me having it in `.status` seems an appropriate place (although right now we have a mix where I consider we have some attributes in `.serviceProviderProperties` that conceptually might belong in `.status`).

I also considered placing this in the ServiceProviderCluster cosmos document. Because it's something that ends up being mapped to exposed ARM API I consider that the Cluster document is an appropriate place instead of ServiceProviderCluster

miguelsorianod, line 68: `[]metav1.Condition` instead of `[]*metav1.Condition`. This follows other condition lists that we have in our cosmos documents as well as how it's defined in K8s CRs

deads2k:
add the conversion code.

internal/api/v20260630preview/external_auth_methods.go
```diff
@@ -304,7 +304,7 @@ func (v version) NewHCPOpenShiftClusterExternalAuth(from *api.HCPOpenShiftCluste
 			SystemData: api.PtrOrNil(newSystemData(from.SystemData)),
 			Properties: &generated.ExternalAuthProperties{
 				ProvisioningState: api.PtrOrNil(generated.ExternalAuthProvisioningState(from.Properties.ProvisioningState)),
-				Status:            newResourceStatus(from.Status.Conditions),
+				Status:            api.PtrOrNil(newExternalAuthResourceStatus(&from.Status)),
```
miguelsorianod, line 307: Before this PR, the way it was implemented, we were always serializing both status and conditions:

```
               "status": {
                       "conditions": []
               },
```

even when there were no conditions (nil)

Right now with the changes in the PR we only serialize `conditions` when the internal userFacingConditons are not nil. Additionally, we only serialize status when the struct is not the zero-value. So for example if the internal userFacingConditions are nil then the result won't show `status` at all.

Is this what we want?

This is based on how the others are structured (api.PtrOrNil, ...)

notice also how I changed the `newResourceStatus`, which was receiving the internal userfacingconditions to one per type (newClusterResourceStatus, newNodePoolResourceStatus, newExternalAuthResourceStatus), because I now pass the whole internal status type and it's different per resource.

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
```diff
@@ -261,14 +261,35 @@ func newKmsKey(from *api.KmsKey) generated.KmsKey {
 }
 
 func newConditions(from []metav1.Condition) []*generated.Condition {
-	// TODO (bvesel): propose a method for how we want to translate our internal metav1.Conditions type to this external
-	// type.  Until then, we'll plumb it with empty data.
-	return []*generated.Condition{}
+	if from == nil {
```
miguelsorianod, line 264: Here if `from` is nil we return nil, but if `from` is empty we return empty

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
```diff
@@ -261,14 +261,35 @@ func newKmsKey(from *api.KmsKey) generated.KmsKey {
...
+	out := make([]*generated.Condition, 0, len(from))
+	for i := range from {
+		c := from[i]
+		cond := &generated.Condition{
+			Type:    api.Ptr(generated.ConditionType(c.Type)),
+			Status:  api.Ptr(generated.StatusType(c.Status)),
+			Reason:  api.Ptr(c.Reason),
+			Message: api.Ptr(c.Message),
+		}
+		if !c.LastTransitionTime.IsZero() {
```
miguelsorianod, line 277: Are we ok with this? this means that:
* If LastTransitionTime is nil we return LastTransitionTime as nil
* If LatTransitionTime is time.Time's zero-value (0001-01-01 00:00:00 +0000 UTC) we also return nil

internal/api/v20260630preview/hcpopenshiftclusters_methods.go
miguelsorianod: This follows the same behavior that we currently have for other slices processing

internal/api/types_cluster.go
```diff
@@ -51,6 +51,21 @@ type HCPOpenShiftClusterStatus struct {
...
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// UserFacingConditions is a list of conditions that tracks user-facing cluster
+	// conditions. Each Condition Type should be unique among all conditions.
+	// The conditions here are exposed to the ARM API. This means that UserFacingConditions
+	// must not contain any internal details. This also means the Type and Reason
+	// values become part of the public API.
+	// Addition of new conditions here should be done only when strictly necessary, sparingly and only done
+	// when there is a clear benefit to doing so. We expect the number of conditions at this
```
deads2k, line 61: explicitly indicate the only allowed one are
1. RequirementsValid
2. Degraded
3. Progressing

deads2k (approved):
conversion bits are a little weird, but I'll take it to start.

rhamitarora (approved):

raelga:
/hold as per Chai Bot recommendation until we review what's going on.

> pull-ci-Azure-ARO-HCP-main-e2e-parallel has failed 8 of 9 batch runs in the last 24h, blocking the merge queue.
> 
> Failing test: Customer should be able to rotate KMS key for a cluster with version >= 4.22
> Error: Timeout (18m) waiting for UpdateHCPCluster20260630 — cluster update with new KMS key never reaches Succeeded state (kms_key_rotation.go:159)
> 
> Likely culprit: [PR #6308](https://github.com/Azure/ARO-HCP/pull/6308) — feat: add UserFacingConditions to Cluster, NodePool, and ExternalAuth status (miguelsorianod)
> • Present in all 8 failing batches but was NOT in the sole successful batch
> • Modifies hcpopenshiftclusters_methods.go (+28/-7 lines) — the v20260630preview cluster update API path that the failing test calls
> 
> Contributing factor: [PR #6120](https://github.com/Azure/ARO-HCP/pull/6120) — Make keyManagementMode required for etcd encryption (mbarnes) — merged to main at the start of this loop, changes etcd encryption validation

PR #6309, opened 2026-07-28
Add ClusterPendingClusterServiceIDAssign controller and wire into CS create

deads2k:
Introduce a new controller that assigns a PendingClusterServiceID
(using ksuid-based /api/aro_hcp/v1alpha1/clusters/<uid>) on clusters
that have neither PendingClusterServiceID nor ClusterServiceID set.
    
The ClusterClusterServiceCreate controller now requires
PendingClusterServiceID to be present before it will act, and uses
its ClusterID() as the cluster-service instance ID instead of
generating a new UID inline.
    

only the last commit is new. The others are lgtm'd in https://github.com/Azure/ARO-HCP/pull/6285


backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
```diff
@@ -239,13 +240,19 @@ func (c *clusterClusterServiceCreateSyncer) createClusterServiceCluster(ctx cont
 	if err != nil {
 		return nil, utils.TrackError(fmt.Errorf("failed to build CS cluster: %w", err))
 	}
+	clusterServiceUID := cluster.ServiceProviderProperties.PendingClusterServiceID.ClusterID()
+	csClusterBuilder.ID(clusterServiceUID)
```
miguelsorianod: Can we move this to within `ocm.BuildCSCluster`, within the withImmutableAttribtues part?

internal/ocm/csuid.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2026 Microsoft Corporation
```
miguelsorianod, line 1: nitpick: it would be better to name the file csuid.go as well as the name of the function and variables:

NewCSClusterUID()
csClusterUIDAlphabet
csClusterUIDEncoding

In that way we restrict this to only the CS Cluster UID and we make sure that people don't accidentally reuse it for some other aspects (general UUID generation etc.)

backend/pkg/controllers/clustercreation/cluster_pending_cluster_service_id_assign_controller.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	uid := ocm.NewUID()
+	pendingID, err := api.NewInternalID(fmt.Sprintf("/api/aro_hcp/v1alpha1/clusters/%s", uid))
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to create PendingClusterServiceID: %w", err))
+	}
+
+	logger.Info("Assigning PendingClusterServiceID", "pendingClusterServiceID", pendingID.String())
+	cluster.ServiceProviderProperties.PendingClusterServiceID = &pendingID
```
miguelsorianod, line 94: I see a couple of issues here:
* For Clusters that already have `cluster.ServiceProviderProperties.ClusterServiceID` the PendingClusterServiceID attribute is never going to be set
* During the transition there can be a race where the cluster gets created in CS with CS-autogenerated UID, then the pending csidgenerator generates it and we end up with mismatched IDs and an orphan cluster: Ordering of events
  1. old code creates CS cluster with CS-generated UID, and no persist occurs
  2. new code rolls out and it generates PendingClusterServiceID
  3. new code creates cluster with the PendingClusterServiceID. The cluster in point 1 ends up orphan

backend/pkg/controllers/clustercreation/cluster_cluster_service_create_controller.go
miguelsorianod, line 237: We cannot merge this until CS changes that allow providing the CS Cluster ID externally have been rolled out to production

backend/pkg/controllers/clustercreation/cluster_pending_cluster_service_id_assign_controller.go
deads2k: all accepted.

internal/ocm/convert.go
```diff
@@ -561,6 +560,10 @@ func withImmutableAttributes(clusterBuilder *arohcpv1alpha1.ClusterBuilder, hcpC
 		NodesOutboundConnectivity(arohcpv1alpha1.NewAzureNodesOutboundConnectivity().
 			OutboundType(outboundType))
 
+	if hcpCluster.ServiceProviderProperties.PendingClusterServiceID != nil {
```
miguelsorianod: I don't think the check is needed. If it's nil it would be a bug and we should fail. Otherwise it will go through with autogenerated id and we want to prevent that

PR #6312, opened 2026-07-29
fix(e2e): correct client factory cache checks and lock misuse

aklymenk:
### What

Fixes three errors in the cached accessors on `perItOrDescribeTestContext` in `test/util/framework/per_test_framework.go`:

| Location | Before | After |
|---|---|---|
| `GetARMSubscriptionsClientFactory` | fast path gated on `clientFactory20240610` | gated on `armSubscriptionsClientFactory` |
| `getARMSubscriptionsClientFactoryUnlocked` | gated on `armResourcesClientFactory` | gated on `armSubscriptionsClientFactory` |
| `SubscriptionID` | `contextLock.Lock()` released with `RUnlock()` | `RLock()` / `RUnlock()` |

All three now follow the `RLock` → check → `RUnlock` → `Lock` → `*Unlocked()`
shape used by every other accessor in the file.

### Why

**Both subscriptions-factory accessors gate on the wrong field.** They check one
cache field but return and cache a different one, so either can return
`(nil, nil)` — leaving callers to nil-deref on `clientFactory.NewClient()`.

**`SubscriptionID` mixes read and write lock modes.** `sync.RWMutex` keeps the
reader count and the writer flag separate, so releasing a write lock with
`RUnlock()` is wrong twice over:

1. The writer flag is never cleared — `contextLock` stays exclusively held, and
   every subsequent accessor (`GetARMResourcesClientFactory`, `RecordTestStep`,
   `commitTimingMetadata`, …) blocks until Ginkgo's node timeout.
2. With zero active readers the counter goes negative and Go hits
   `fatal("sync: RUnlock of unlocked RWMutex")` in `rUnlockSlow`. `fatal()` is a
   runtime throw, **not** a panic — neither `recover()` nor the
   `utilruntime.HandleCrash` guards the framework relies on can catch it. The
   whole test binary dies, skipping cleanup and leaking Azure resource groups.

That branch is currently unreachable because the per-test `subscriptionID` field
is never assigned (only `perBinaryInvocationTestContext.subscriptionID` is), but
any future `tc.subscriptionID = …` — a natural-looking one-line change — arms it.

### Testing

No new tests. 

### PR Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/test/util/framework/per_test_framework.go
+++ b/test/util/framework/per_test_framework.go
@@ -949,7 +949,7 @@ func (tc *perItOrDescribeTestContext) Get20240610ClientFactoryOrDie(ctx context.
 
 func (tc *perItOrDescribeTestContext) GetARMSubscriptionsClientFactory() (*armsubscriptions.ClientFactory, error) {
 	tc.contextLock.RLock()
-	if tc.clientFactory20240610 != nil {
+	if tc.armSubscriptionsClientFactory != nil {
 		defer tc.contextLock.RUnlock()
 		return tc.armSubscriptionsClientFactory, nil
 	}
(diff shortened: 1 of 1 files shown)


mvacula02 (approved):
/lgtm

mgahagan73:
The fixes look correct and should make the codebase more robust. We don't have unit tests which would have caught this, but I won't hold this PR up for that. It's fine to add those in follow-up.

/approve

PR #6316, opened 2026-07-29
feat: surface Cluster and NodePool ServiceProvider validations as RequirementsValid

miguelsorianod:
Builds on top of https://github.com/Azure/ARO-HCP/pull/6308.

Add Cluster and NodePool controllers that roll ServiceProviderCluster
and ServiceProviderNodePool Status.Validations respectively up onto
their corresponding Status.UserFacingConditions as a single
RequirementsValid condition, which ends up exposed via the ARM API.

Aggregation rules:
- Type is always RequirementsValid.
- When all validations are True then Status is set to True with
  Reason=Valid
- When there's at least one validation that is False or Uknown then
  Status is set to False with Reason=Degraded. In that case, messages
  of both false or unknown validations are aggregated in the shared
  Degraded-style format

### DIFF
--- a/backend/pkg/controllers/statuscontrollers/requirements_valid.go
+++ b/backend/pkg/controllers/statuscontrollers/requirements_valid.go
@@ -1,0 +1,83 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package statuscontrollers
+
+import (
+	"sort"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// requirementsValidConditionType is the user-facing condition type that
+	// aggregates ServiceProvider* Status.Validations onto the parent resource's
+	// Status.UserFacingConditions.
+	requirementsValidConditionType = "RequirementsValid"
+
+	// requirementsValidConditionReasonValid is set when every known validation
+	// has Status=True (or there are no validations yet). Paired with
+	// Status=True and an empty Message.
+	requirementsValidConditionReasonValid = "Valid"
+	// requirementsValidConditionReasonDegraded is set when at least one
+	// validation has Status False or Unknown. Paired with Status=False and a
+	// Message that enumerates those non-True validations.
+	requirementsValidConditionReasonDegraded = "Degraded"
+)
+
+// aggregateRequirementsValidCondition builds the single RequirementsValid
+// condition that RequirementsValid aggregators write onto
+// status.userFacingConditions.
+//
+// Input is the ServiceProvider* Status.Validations slice.
+//
+// Result:
+//   - Type is always RequirementsValid.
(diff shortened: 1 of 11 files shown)


miguelsorianod:
/hold for now:

* We want to wait until we see in the e2e output in kusto that it's working as expected
* We also might not want to merge this until the redesign of the validation controllers that doesn't expose internal information is merged

backend/pkg/controllers/statuscontrollers/requirements_valid.go
```diff
@@ -0,0 +1,83 @@
...
+	sort.Slice(failed, func(i, j int) bool {
+		return failed[i].name < failed[j].name
+	})
+
+	if len(failed) == 0 {
+		return metav1.Condition{
+			Type:    requirementsValidConditionType,
+			Status:  metav1.ConditionTrue,
+			Reason:  requirementsValidConditionReasonValid,
+			Message: "",
```
miguelsorianod, line 73: When we determine that RequirementsValid is `True`, we do not set any message, and if the validations that succeeded had a message we don't surface it. Are we fine with this? alternatives:
* We set a message at this level anyway (which one?)
* We surface the messages of the succeeded validations (if any)
* A combination of both points

backend/pkg/controllers/statuscontrollers/requirements_valid.go
```diff
@@ -0,0 +1,83 @@
...
+//   - When every validation has Status=True (or the slice is empty):
+//     Status=True, Reason=Valid, Message empty.
+//   - When at least one validation has Status False or Unknown: Status=False,
+//     Reason=Degraded, Message lists only those non-True validations, sorted
+//     by Type and formatted by joinNamedMessages as "ValidationName: <line>"
+//     (one line per unique message line), joined with newlines.
+func aggregateRequirementsValidCondition(validations []metav1.Condition) metav1.Condition {
+	failed := make([]namedMessage, 0, len(validations))
+	for _, validation := range validations {
+		if validation.Status == metav1.ConditionTrue {
```
miguelsorianod, line 56: This means that we consider it failed if the condition was either False or Unknown

backend/pkg/controllers/statuscontrollers/union_condition.go
```diff
@@ -125,13 +125,34 @@ func UnionCondition(
...
+			message: src.Condition.Message,
+		})
+	}
+	return joinNamedMessages(named)
+}
+
+// namedMessage is a source name paired with a free-form message. Used by
+// joinNamedMessages so Degraded aggregation (controller name) and
+// RequirementsValid aggregation (validation type) share one formatter.
+type namedMessage struct {
```
miguelsorianod, line 141: In this PR I initially used `SourcedCondition` in the requirements valid controllers, but that type has the concept of ControllerName which doesn't really apply to that. Because of that, I refactored unionMessage, introducing the concept of namedMessage, which is independent of whether it's a condition of a controller or not, and then unionMessage builds that. That makes then possible to reuse the logic of joinNamedMessages in both unionMessage and the requirements valid controllers

backend/pkg/controllers/statuscontrollers/union_condition.go
```diff
@@ -125,13 +125,34 @@ func UnionCondition(
...
 	lines := []string{}
 	for _, src := range sources {
-		if len(src.Condition.Message) == 0 {
+		if len(src.message) == 0 {
 			continue
 		}
-		for _, line := range uniq(strings.Split(src.Condition.Message, "\n")) {
-			lines = append(lines, fmt.Sprintf("%s: %s", src.ControllerName, line))
+		for _, line := range uniq(strings.Split(src.message, "\n")) {
+			lines = append(lines, fmt.Sprintf("%s: %s", src.name, line))
```
deads2k, line 155: isn't this a different format than the "source" you used on operations? Keep them the same.

PR #6325, opened 2026-07-30
fix: retry Graph 400 on POST to cover eventual-consistency race

inbharajmani:
https://redhat.atlassian.net/browse/AROSLSRE-1655

## Why

e2e specs that call `NewAppRegistrationWithServicePrincipal` (Authorized CIDRs Connectivity, external auth) started failing shortly after [#6200](https://github.com/Azure/ARO-HCP/pull/6200) ("fix: move Graph API retry logic to transport-level middleware", Jul 28) — each failure lands ~1 s after `CreateServicePrincipal` is invoked, immediately after its cluster/node pool reached `Succeeded`.

Root cause: Microsoft Graph replication lag. Immediately after `POST /applications`, `POST /servicePrincipals` can hit a replica that has not yet seen the new app and returns:

- HTTP **400** `Request_BadRequest`
- Detail: `NoBackingApplicationObject` — *"The appId '…' of the service principal does not reference a valid application object."*

The old `CreateServicePrincipal` had an application-level `wait.PollUntilContextTimeout` loop that retried on *any* error for 2 minutes specifically to cover this. The refactor replaced it with `graphRetryHandler`, which only retries 429/503/504 and 404-on-POST. Graph surfaces this particular eventual-consistency race as **400**, not 404, so the new middleware doesn't retry it and the setup step fails immediately.

Failures: https://cihealth.tools.hcpsvc.osadev.cloud/run-log?q=failed+to+create+service+principal%3A+create+service+principal%3A+error+status+code+received+from+the+API

Corroborating source: [Microsoft Learn Q&A — MS Graph API failing to create service principal](https://learn.microsoft.com/en-us/answers/questions/5883620/ms-graph-api-failing-to-create-service-principal-f) confirms the exact HTTP 400 / `Request_BadRequest` / `NoBackingApplicationObject` response and recommends retrying until the app replicates.

## What

- `internal/graph/util/graph_retry_handler.go`: extend `isRetriableStatusCode` to also retry **400 on POST** — same shape as the existing 404-on-POST rule. GET/PUT/DELETE 400 remain fatal.
- `internal/graph/util/serviceprincipal.go`: wrap the `CreateServicePrincipal` error with `odataErrorWithDiagnostics` (mirroring `AddPassword`). The current failing test log only shows Kiota's fallback `"error status code received from the API"` because the SP path discards the OData status/code; this ensures future Graph failures surface HTTP status + OData code directly.
- `internal/graph/util/graph_retry_handler_test.go`: add test cases for 400-on-POST retry, and 400-on-GET/DELETE non-retry.

## Test plan
- [x] `go test ./internal/graph/util/...` — passes
- [ ] e2e run to confirm both `Authorized CIDRs Connectivity` and `external-auth` specs proceed past `NewAppRegistrationWithServicePrincipal`


mvacula02 (approved):
/lgtm

closes [ARO-27423](https://redhat.atlassian.net/browse/ARO-27423) and related #6096

raelga (approved):
/lgtm
/approve

inbharajmani:
/close
Superseded by https://github.com/Azure/ARO-HCP/pull/6348

PR #6336, opened 2026-07-30
add new admincredentials request with a CSR on it

deads2k:
This will let us fix latency and avoid hold customer key material in area where a read-only exposure can result is compromising a customer cluster.


api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
```diff
@@ -1307,6 +1307,15 @@
             "required": true,
             "type": "string",
             "pattern": "^[a-zA-Z]([-a-zA-Z0-9]{0,52}[a-zA-Z0-9])?$"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "description": "The content of the action request",
+            "required": false,
```
bennerv: We want to require the body on this api version, we'll want to update typespec to make the request body required instead of optional.

PR #6339, opened 2026-07-30
Add types for admin credentials

deads2k:
builds on https://github.com/Azure/ARO-HCP/pull/6336 to add the types we need in the last two commits.

this includes 

1. internal representations of the new request type
2. internal admin credentials and revocation types we'll use to build the controllers
3. additional nesting to FooDesire types for admin credentials and revocation
4. supporting crud, informers, and listers


internal/api/kubeapplier/registry.go
```diff
@@ -44,4 +44,14 @@ var (
 	ClusterScopedReadDesireResourceType = nestedResourceType(api.ClusterResourceTypeName, ReadDesireResourceTypeName)
 	// NodePoolScopedReadDesireResourceType is readDesires nested under a NodePool under a Cluster.
 	NodePoolScopedReadDesireResourceType = nestedResourceType(api.ClusterResourceTypeName, api.NodePoolResourceTypeName, ReadDesireResourceTypeName)
+
+	// CredentialRequestScopedApplyDesireResourceType is applyDesires nested under a SystemAdminCredentialRequest under a Cluster.
+	CredentialRequestScopedApplyDesireResourceType = nestedResourceType(api.ClusterResourceTypeName, api.SystemAdminCredentialRequestResourceTypeName, ApplyDesireResourceTypeName)
```
miguelsorianod: If we follow the naming conventions this should be named SystemAdminCredentialScopedApplyDesireResourceType

Same applies to the others:

SystemAdminCredentialRevocationScopedApplyDesireResourceType
SystemAdminCredentialRevocationScopedReadDesireResourceType

internal/api/types_admincredential.go
```diff
@@ -16,6 +16,10 @@ package api
 
 import "time"
 
+type HCPOpenShiftClusterAdminCredentialRequest struct {
```
miguelsorianod, line 19: Where is this going to plumb to? I see it defined but not referenced anywhere for now

internal/api/types_admincredential.go
```diff
@@ -16,6 +16,10 @@ package api
 
 import "time"
```
miguelsorianod, line 17: How are AdminCredential and SystemAdminCredential going to be related?

internal/api/types_operation.go
```diff
@@ -92,6 +92,24 @@ type Operation struct {
 	// This will be removed once all clusters whose deletion was triggered before the new approach is fully rolled out have been
 	// fully deleted in all ARO-HCP permanent environments, for all regions.
 	UsesNewClusterDeletionApproach bool `json:"usesNewClusterDeletionApproach"`
+
+	// Written by: Frontend POST requestAdminCredential, backend systemadmincredentialcontrollers
+	AdminCredentialsRequest *OperationAdminCredentialsRequest `json:"adminCredentialsRequest,omitempty"`
```
miguelsorianod: The `Operation` internal API type is shared by the operations of different resource types. It is not specific to admincredentials. Is this something temporary or permanent? If it's permanent, it doesn't seem to be the appropriate place to put it. 

What is the purpose of it?

internal/api/types_systemadmincredential.go
```diff
@@ -0,0 +1,94 @@
...
+// Lifecycle is tracked via metav1.Conditions rather than an explicit Phase enum
+// so that individual aspects of the request (issuance, revocation, cleanup) can
+// progress independently and callers can reason about each concern separately.
+//
+// +k8s:deepcopy-gen=true
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type SystemAdminCredentialRequest struct {
+	CosmosMetadata `json:"cosmosMetadata"`
+
+	Spec   SystemAdminCredentialRequestSpec   `json:"spec"`
```
miguelsorianod, line 33: Is there some reason we don't follow the CustomerProperties / ServiceProviderProperties organization like we do for other internal api types? how are we going to separate those?

internal/api/types_systemadmincredential.go
```diff
@@ -0,0 +1,94 @@
...
+	SignedCertificate string `json:"signedCertificate,omitempty"`
+	// Conditions tracks lifecycle state using standard metav1.Conditions.
+	// Known condition types:
+	//   - "Issued": True when the CSR has been signed and the cert is available.
+	//   - "Failed": True when the CSR was denied or issuance otherwise failed.
+	//   - "AwaitingRevocation": True when revocation has been requested but not yet confirmed.
+	//   - "Revoked": True when the credential has been revoked.
+	//   - "ContentDeleted": True when all MC-side objects have been cleaned up.
+	// Written by: SystemAdminCredentialIssuanceObserver (Issued, Failed)
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
miguelsorianod, line 77: We could also add `UserFacingConditions` now that we started to do it for the other resource types (cluster, nodepool, externalauth

internal/api/types_systemadmincredential.go
```diff
@@ -0,0 +1,94 @@
...
+	// RevokedAt is set when the Revoked condition transitions to True.
+	RevokedAt *metav1.Time `json:"revokedAt,omitempty"`
+	// DeleteTimestamp is set when deletion has been requested for this credential
+	// request. Controllers use this to drive teardown of associated kube-applier
+	// desires before removing the credential request document itself.
+	// Written by: SystemAdminCredentialRevocationMarkRequests
+	DeleteTimestamp *metav1.Time `json:"deleteTimestamp,omitempty"`
```
miguelsorianod: For the other resources  (cluster, nodepool, externalauth) we call it `DeletionTimestamp`

internal/api/types_systemadmincredential.go
```diff
@@ -0,0 +1,94 @@
...
+	ExpirationTimestamp metav1.Time `json:"expirationTimestamp"`
+	// OperationID is the ARM operation that created this credential request.
+	// Written by: SystemAdminCredentialDispatchRequestCredential
+	OperationID string `json:"operationID"`
+	// CertificateRequestPEM is the PEM-encoded PKCS#10 certificate signing
+	// request provided by the caller. The backend embeds this directly in
+	// the Kubernetes CertificateSigningRequest it creates on the management
+	// cluster. The corresponding private key is held only by the caller.
+	// Written by: SystemAdminCredentialDispatchRequestCredential
+	CertificateRequestPEM string `json:"certificateRequestPEM"`
```
miguelsorianod: Can we name it CertificateSigningRequestPEM? that stands for the CSR (Certificate Signing Request) acronym

internal/api/types_systemadmincredential.go
```diff
@@ -0,0 +1,94 @@
...
+	Spec   SystemAdminCredentialRequestSpec   `json:"spec"`
+	Status SystemAdminCredentialRequestStatus `json:"status"`
+}
+
+// SystemAdminCredentialRequestSpec contains the desired state of the credential request.
+// Written by: SystemAdminCredentialDispatchRequestCredential (all fields set at creation)
+type SystemAdminCredentialRequestSpec struct {
+	// Username is the K8s username embedded in the cert CN.
+	// Written by: SystemAdminCredentialDispatchRequestCredential
+	Username string `json:"username,omitempty"`
```
miguelsorianod, line 42: Do we expect potentially multiple names? CN + SAN
I am not familiar on details on how it works so asking just in case.

internal/api/types_systemadmincredential.go
```diff
@@ -0,0 +1,94 @@
...
+	// CreationTimestamp is when the credential request was created (server-set).
+	// The garbage collector deletes each request once it is older than the
+	// retention window, regardless of the request's status.
+	// Written by: SystemAdminCredentialDispatchRequestCredential
+	CreationTimestamp metav1.Time `json:"creationTimestamp"`
```
miguelsorianod, line 47: CreationTimestamp and ExpirationTimestamp are `metav1.Time` but DeleteTimestamp and RevokedTimestamp are `*metav1.Time`. Is that intentional? is it because they are always set at the frontend level synchronously in the initial call?

internal/api/types_systemadmincredential.go
```diff
@@ -0,0 +1,94 @@
...
+	// ExpirationTimestamp is when the cert ceases to be valid (server-set, now + 24h).
+	// Written by: SystemAdminCredentialDispatchRequestCredential
+	ExpirationTimestamp metav1.Time `json:"expirationTimestamp"`
```
miguelsorianod, line 50: Do we think there's a chance at some point we might end-up wanting some that don't have expirationtimestamp? I would guess that no as that would defeat the security posture, but asking because I am sure that at some point some end-user will ask support for this

internal/api/types_systemadmincredentialrevocation.go
```diff
@@ -0,0 +1,82 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SystemAdminCredentialRevocation represents a single revocation of all of a
```
miguelsorianod, line 21: I wonder if the behavior of "represents a single revocation of all cluster's system admin credentials" was because of simplicty and/or limits when it was on CS. Have we considered if we want and/or if it's possible to have individual revocations?

internal/api/types_systemadmincredential.go
```diff
@@ -0,0 +1,94 @@
...
+	//   - "Issued": True when the CSR has been signed and the cert is available.
+	//   - "Failed": True when the CSR was denied or issuance otherwise failed.
+	//   - "AwaitingRevocation": True when revocation has been requested but not yet confirmed.
+	//   - "Revoked": True when the credential has been revoked.
+	//   - "ContentDeleted": True when all MC-side objects have been cleaned up.
+	// Written by: SystemAdminCredentialIssuanceObserver (Issued, Failed)
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// RevokedAt is set when the Revoked condition transitions to True.
+	RevokedAt *metav1.Time `json:"revokedAt,omitempty"`
+	// DeleteTimestamp is set when deletion has been requested for this credential
```
miguelsorianod: Is it when deletion has been requested, or when revocation has been requested? in SystemAdminCredentialRevocation it's documented that once revocation occurs the SystemAdminCredentialRequests deletion timestamps are set.

internal/api/v20260901preview/admincredential_methods.go
```diff
@@ -15,9 +15,12 @@
 package v20260901preview
```
miguelsorianod, line 16: The new API version will have the new admincredentials approach, which should solve the latency issues.

What will we do then with the older API versions which will still rely on CS? that means that
* CS will need to be kept there
* Those will have the latency issues

Has a plan for the older versions been discussed?

internal/api/v20240610preview/admincredential_methods.go
```diff
@@ -30,3 +30,7 @@ func newHCPOpenShiftClusterAdminCredential(from *api.HCPOpenShiftClusterAdminCre
 func (v version) MarshalHCPOpenShiftClusterAdminCredential(from *api.HCPOpenShiftClusterAdminCredential) ([]byte, error) {
 	return arm.MarshalJSON(newHCPOpenShiftClusterAdminCredential(from))
 }
+
+func (v version) UnmarshalHCPOpenShiftClusterAdminCredentialRequest(_ []byte) (*api.HCPOpenShiftClusterAdminCredentialRequest, error) {
```
miguelsorianod, line 34: Is there some reason we return the struct initialized as empty vs `nil` ?

internal/api/v20260901preview/admincredential_methods.go
```diff
@@ -15,9 +15,12 @@
...
 
 import (
+	"encoding/json"
+
 	"github.com/Azure/ARO-HCP/internal/api"
 	"github.com/Azure/ARO-HCP/internal/api/arm"
 	"github.com/Azure/ARO-HCP/internal/api/v20260630preview/generated"
```
miguelsorianod: This imports both:

	"github.com/Azure/ARO-HCP/internal/api/v20260630preview/generated"

and

	generated0901 "github.com/Azure/ARO-HCP/internal/api/v20260901preview/generated"

shouldn't we just have 

    "github.com/Azure/ARO-HCP/internal/api/v20260901preview/generated"

?

the other _methods files only seem to refer to their own API version

internal/api/types_operation.go
miguelsorianod: Would this be AdminCredentialsRequest or SystemAdminCredentialsRequest? same question for revocation

internal/api/types_systemadmincredentialrevocation.go
```diff
@@ -0,0 +1,82 @@
...
+	Spec   SystemAdminCredentialRevocationSpec   `json:"spec"`
+	Status SystemAdminCredentialRevocationStatus `json:"status"`
+}
+
+// SystemAdminCredentialRevocationSpec contains the desired state of the revocation.
+// Written by: SystemAdminCredentialDispatchRevokeCredentials (all fields set at creation)
+type SystemAdminCredentialRevocationSpec struct {
+	// OperationID is the ARM operation that triggered the revocation.
+	// Written by: SystemAdminCredentialDispatchRevokeCredentials
+	OperationID string `json:"operationID"`
```
miguelsorianod, line 52: Can it be that this changes over time for an instance? for example,  a revocation is requested but the operation is marked as failed, so someone retries again

Same question applies to the SystemAdminCredentialRequestSpec type

internal/api/types_systemadmincredentialrevocation.go
```diff
@@ -0,0 +1,82 @@
...
+	// RevokeOpSuffix is the shortened operation ID used in CRR object names.
+	// Written by: SystemAdminCredentialDispatchRevokeCredentials
+	RevokeOpSuffix string `json:"revokeOpSuffix"`
```
miguelsorianod, line 55: What is the purpose and motivation of this? it mentions "is the shortened operation ID used in CRR object names" but why do we need that?

internal/api/types_systemadmincredentialrevocation.go
```diff
@@ -0,0 +1,82 @@
...
+	//     certificates have been revoked on the hosted cluster.
+	//   - "Complete": True when the whole revocation flow is done and this document
+	//     may be deleted.
+	// Written by: SystemAdminCredentialRevocationMarkRequests (CredentialsMarkedForDeletion), SystemAdminCredentialRevocationCompletion (CertificatesRevoked, Complete)
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// DeleteTimestamp is set once revocation is complete and the deletion
+	// controller should tear down the revocation's desires and remove this
+	// document.
+	// Written by: SystemAdminCredentialRevocationCompletion
+	DeleteTimestamp *metav1.Time `json:"deleteTimestamp,omitempty"`
```
miguelsorianod: Same comment as the other about `DeletionTimestamp`

internal/database/database.go
```diff
@@ -145,6 +145,14 @@ type ResourcesDBClient interface {
 
 	ServiceProviderNodePools(subscriptionID, resourceGroupName, clusterName, nodePoolName string) ResourceCRUD[api.ServiceProviderNodePool, *api.ServiceProviderNodePool]
 
+	// SystemAdminCredentialRequests retrieves a CRUD interface for managing SystemAdminCredentialRequest documents
+	// nested under a specific cluster.
+	SystemAdminCredentialRequests(subscriptionID, resourceGroupName, clusterName string) SystemAdminCredentialRequestsCRUD
```
miguelsorianod: Should this and SystemAdminCredentialRevocations belong within `HCPClusters` instead? for NodePools and ExternalAuths we have it there for example 🤔

internal/database/kube_applier_client.go
```diff
@@ -79,10 +79,18 @@ type KubeApplierDBClient interface {
 	ApplyDesiresForCluster(subscriptionID, resourceGroupName, clusterName string) (ResourceCRUD[kubeapplier.ApplyDesire, *kubeapplier.ApplyDesire], error)
 	// ApplyDesiresForNodePool returns a CRUD scoped to a nodepool parent.
 	ApplyDesiresForNodePool(subscriptionID, resourceGroupName, clusterName, nodePoolName string) (ResourceCRUD[kubeapplier.ApplyDesire, *kubeapplier.ApplyDesire], error)
+	// ApplyDesiresForCredentialRequest returns a CRUD scoped to a SystemAdminCredentialRequest parent.
+	ApplyDesiresForCredentialRequest(subscriptionID, resourceGroupName, clusterName, credentialRequestName string) (ResourceCRUD[kubeapplier.ApplyDesire, *kubeapplier.ApplyDesire], error)
```
miguelsorianod: Better rename to:

ApplyDesiresForSystemAdminCredentialRequest
ApplyDesiresForSystemAdminCredentialRevocation
ReadDesiresForSystemAdminCredentialRequest
ReadDesiresForSystemAdminCredentialRevocation

it's more verbose but consistent with how the type is named

internal/database/kube_applier_client.go
```diff
@@ -110,12 +118,16 @@ type KubeApplierListers interface {
 type KubeApplierApplyDesireCRUD interface {
 	ApplyDesiresForCluster(subscriptionID, resourceGroupName, clusterName string) (ResourceCRUD[kubeapplier.ApplyDesire, *kubeapplier.ApplyDesire], error)
 	ApplyDesiresForNodePool(subscriptionID, resourceGroupName, clusterName, nodePoolName string) (ResourceCRUD[kubeapplier.ApplyDesire, *kubeapplier.ApplyDesire], error)
+	ApplyDesiresForCredentialRequest(subscriptionID, resourceGroupName, clusterName, credentialRequestName string) (ResourceCRUD[kubeapplier.ApplyDesire, *kubeapplier.ApplyDesire], error)
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/6339/changes/c714087f4b954b18fa3079ce290f2c4004560150#r3689633799

internal/database/kube_applier_client.go
```diff
@@ -110,12 +118,16 @@ type KubeApplierListers interface {
...
+	ApplyDesiresForRevocation(subscriptionID, resourceGroupName, clusterName, revocationName string) (ResourceCRUD[kubeapplier.ApplyDesire, *kubeapplier.ApplyDesire], error)
 }
 
 // KubeApplierReadDesireCRUD is the ReadDesire peer of KubeApplierApplyDesireCRUD.
 type KubeApplierReadDesireCRUD interface {
 	ReadDesiresForCluster(subscriptionID, resourceGroupName, clusterName string) (ResourceCRUD[kubeapplier.ReadDesire, *kubeapplier.ReadDesire], error)
 	ReadDesiresForNodePool(subscriptionID, resourceGroupName, clusterName, nodePoolName string) (ResourceCRUD[kubeapplier.ReadDesire, *kubeapplier.ReadDesire], error)
+	ReadDesiresForCredentialRequest(subscriptionID, resourceGroupName, clusterName, credentialRequestName string) (ResourceCRUD[kubeapplier.ReadDesire, *kubeapplier.ReadDesire], error)
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/6339/changes/c714087f4b954b18fa3079ce290f2c4004560150#r3689633799

internal/database/listers/apply_desire_lister.go
```diff
@@ -38,6 +38,14 @@ type ApplyDesireLister interface {
 	// containing NodePool identity and the desire's name.
 	GetForNodePool(ctx context.Context, subscriptionID, resourceGroupName, clusterName, nodePoolName, name string) (*kubeapplier.ApplyDesire, error)
 
+	// GetForCredentialRequest fetches a single ApplyDesire nested under a
+	// SystemAdminCredentialRequest by the request's identity and the desire's name.
+	GetForCredentialRequest(ctx context.Context, subscriptionID, resourceGroupName, clusterName, credentialRequestName, name string) (*kubeapplier.ApplyDesire, error)
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/6339/changes/c714087f4b954b18fa3079ce290f2c4004560150#r3689633799

internal/database/listers/read_desire_lister.go
```diff
@@ -30,6 +30,8 @@ type ReadDesireLister interface {
 	List(ctx context.Context) ([]*kubeapplier.ReadDesire, error)
 	GetForCluster(ctx context.Context, subscriptionID, resourceGroupName, clusterName, name string) (*kubeapplier.ReadDesire, error)
 	GetForNodePool(ctx context.Context, subscriptionID, resourceGroupName, clusterName, nodePoolName, name string) (*kubeapplier.ReadDesire, error)
+	GetForCredentialRequest(ctx context.Context, subscriptionID, resourceGroupName, clusterName, credentialRequestName, name string) (*kubeapplier.ReadDesire, error)
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/6339/changes/c714087f4b954b18fa3079ce290f2c4004560150#r3689633799

internal/database/unionlisters/kubeapplier/union_desire_lister.go
```diff
@@ -46,6 +46,8 @@ type DesireLister[T any] interface {
 	List(ctx context.Context) ([]*T, error)
 	GetForCluster(ctx context.Context, subscriptionID, resourceGroupName, clusterName, name string) (*T, error)
 	GetForNodePool(ctx context.Context, subscriptionID, resourceGroupName, clusterName, nodePoolName, name string) (*T, error)
+	GetForCredentialRequest(ctx context.Context, subscriptionID, resourceGroupName, clusterName, credentialRequestName, name string) (*T, error)
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/6339/changes/c714087f4b954b18fa3079ce290f2c4004560150#r3689633799

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -1019,6 +1019,13 @@ scalar DiskEncryptionSetResourceId
  * =======================================
  */
 
+/** HCP cluster admin credential request body */
+@added(Versions.v2026_09_01_preview)
+model HcpOpenShiftClusterAdminCredentialRequest {
```
miguelsorianod, line 1024: Is the idea that at the ARM API level we won't introduce the new systemAdminCredential resource and we will keep adminCredential? will the kubeconfig still be retrieved as part of the operation result, and we won't have adminCredential and/or systemAdminCredential as a resource in itself at the ARM level?

internal/api/types_admincredential.go
deads2k: > Where is this going to plumb to? I see it defined but not referenced anywhere for now

We'll use this in the frontend to deserialize the user request body into an external type (specific for their version), then convert to an internal type (this one) for us to have a unified code path.

internal/api/types_admincredential.go
deads2k: > How are AdminCredential and SystemAdminCredential going to be related?

AdminCredential is used in frontend only to deserialize the content.  It will create an operation (as it does today) and the operation will carry the CSR through. There will be dispatch controller (using the pattern you established for create) that goes from operation to SystemAdminCredentialRequest, which is then used to power the controller workflow.

internal/api/types_operation.go
deads2k: > The `Operation` internal API type is shared by the operations of different resource types. It is not specific to admincredentials. Is this something temporary or permanent? If it's permanent, it doesn't seem to be the appropriate place to put it.
> 
> What is the purpose of it?

It will be permanent until we remove the cluster-service workflow. Then it is possible to refactor this to create a SystemAdminCredentialRequest in the frontend.  It's not a priority for me though.  It's location here is an artifact of how the existing code is built and the relative benefit of refactor at this stage (low).

internal/api/types_systemadmincredential.go
deads2k: > Is there some reason we don't follow the CustomerProperties / ServiceProviderProperties organization like we do for other internal api types? how are we going to separate those?

A user cannot modify or view this.  It's purely an internal detail.

internal/api/types_systemadmincredential.go
deads2k: >  We could also add UserFacingConditions now that we started to do it for the other resource types (cluster, nodepool, externalauth

This resource isn't userFacing, but I'm willing to do this as a future step to make the error cases function better.  A followup/future change though.

internal/api/types_systemadmincredential.go
deads2k: Doh.

internal/api/types_systemadmincredential.go
deads2k: > Do we expect potentially multiple names? CN + SAN
> I am not familiar on details on how it works so asking just in case.

No.  the frontend will actually restrict values here.

internal/api/types_systemadmincredential.go
deads2k: > CreationTimestamp and ExpirationTimestamp are `metav1.Time` but DeleteTimestamp and RevokedTimestamp are `*metav1.Time`. Is that intentional? is it because they are always set at the frontend level synchronously in the initial call?

Yes, it's because every instance should have a value there.

internal/api/types_systemadmincredential.go
deads2k: > Do we think there's a chance at some point we might end-up wanting some that don't have expirationtimestamp? I would guess that no as that would defeat the security posture, but asking because I am sure that at some point some end-user will ask support for this

we will never support this.

internal/api/types_systemadmincredentialrevocation.go
deads2k: > I wonder if the behavior of "represents a single revocation of all cluster's system admin credentials" was because of simplicty and/or limits when it was on CS. Have we considered if we want and/or if it's possible to have individual revocations?

HCP can't offer that.  We don't have CRLs in kube, so we revoke the entire signer.

internal/api/types_systemadmincredential.go
deads2k: > Is it when deletion has been requested, or when revocation has been requested? in SystemAdminCredentialRevocation it's documented that once revocation occurs the SystemAdminCredentialRequests deletion timestamps are set.

revocation marks the logical delete.

internal/api/v20260901preview/admincredential_methods.go
deads2k: > What will we do then with the older API versions which will still rely on CS? that means that
> 
>     * CS will need to be kept there
> 
>     * Those will have the latency issues
> 
> 
> Has a plan for the older versions been discussed?

Yes, we're going to break them at some point.

internal/api/v20240610preview/admincredential_methods.go
deads2k: > Is there some reason we return the struct initialized as empty vs `nil` ?

just what I'm used to

internal/api/types_systemadmincredentialrevocation.go
deads2k: >Can it be that this changes over time for an instance? for example, a revocation is requested but the operation is marked as failed, so someone retries again
>
>Same question applies to the SystemAdminCredentialRequestSpec type

not expecting it to change.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
deads2k: > Is the idea that at the ARM API level we won't introduce the new systemAdminCredential resource and we will keep adminCredential? will the kubeconfig still be retrieved as part of the operation result, and we won't have adminCredential and/or systemAdminCredential as a resource in itself at the ARM level?

Correct.

deads2k:
all comments addressed
/label lgtm

PR #6342, opened 2026-07-30
chore(owners): add SRE approvers for RP rules

mmazur:
https://redhat.atlassian.net/browse/AROSLSRE-1641

## Summary
- grants `sre-approvers` access to selected generated RP alerting and recording-rule Bicep outputs, plus RP metrics templates
- adds scoped `OWNERS` files for alert definitions in the selected service, documentation, and observability directories
- keeps SRE ownership limited to alert-related files rather than broad service or infrastructure directories
- **does not** grant access to other teams' alerts bicep files, meaning it's impossible for SRE to approve changes to alerts they don't own

### DIFF
--- a/admin/alerts/OWNERS
+++ b/admin/alerts/OWNERS
@@ -1,0 +1,2 @@
+approvers:
+  - sre-approvers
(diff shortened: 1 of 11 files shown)


mmazur:
/override e2e-parallel

This changes only OWNERS files, there is nothing to e2e test.

PR #6344, opened 2026-07-30
admin credential API with transition

deads2k:
Now ready to review.  It adds a new flow for getting credentials, while keeping the old working for a transition.  We merge this one so that dev PRs exercise the newer path, while int, stg, prod use the old until we can rollout the new API version.


davidffrench:
Is this PR still needed or can it be closed? It looks like the admin credential work was completed via the incremental approach in PR #6390 (merged Aug 12) and PR #6699 (remove old API fallback, merged Sep 1). This PR has had `do-not-merge/hold` since Aug 13.

PR #6346, opened 2026-07-30
fix: deduplicate UJNodePoolStuckOperation alert across metric replicas

ehvs:
https://redhat.atlassian.net/browse/ARO-28778

### What

Deduplicates the `UJNodePoolStuckOperation` node-pool SLO alert so a single wedged node pool produces exactly one alert/incident regardless of how many Prometheus replicas scrape the metric. Specifically:

- Wraps the stuck-operation expression in  `max by (cluster, environment, region, subscription_id, resource_id,
  resource_type, operation_type, phase) (...)`, dropping volatile scrape/replica labels (`pod` / `instance` / `prometheus_replica`) from the alert identity
- Adds a promtool unit test (Test 16) that feeds two identical stuck-operation series differing only by a volatile `instance` label and asserts a single alert fires.

### Why

The `max by (...)` aggregation in the `UJNodePoolStuckOperation` rule was introduced specifically to deduplicate alerts across volatile scrape/replica labels (`pod` / `instance` / `prometheus_replica`), so a single wedged node pool does not
fan out into multiple identical IcM incidents. The existing promtool tests only used single-series inputs and never exercised the multi-replica duplication case, leaving this behavior uncovered — a silent regression here would page on-call with duplicate incidents. This test locks in the dedup guarantee.

### Testing

- **Unit tests**: Added `Test 16` in `nodepool-slo-prometheusRule_test.yaml`. Verified it passes via the rule generator, which runs `promtool test rules` on the extracted rules:
  ```sh
  make -C tooling/prometheus-rules run-rp-services

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/generatedRPPrometheusAlertingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedRPPrometheusAlertingRules.bicep
@@ -526,7 +526,7 @@ resource arohcpNodepoolSloErrorAlerts 'Microsoft.AlertsManagement/prometheusRule
           summary: '{{ $labels.cluster }}: Node Pool operation for {{ $labels.resource_id }} stuck in {{ $labels.phase }} for over 2 hours'
           title: '{{ $labels.cluster }}: Node Pool operation for {{ $labels.resource_id }} stuck in {{ $labels.phase }} for over 2 hours'
         }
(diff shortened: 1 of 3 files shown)


tuxerrante:
Non-blocking: can we add `environment` and `region` to the new Test 16 input series and expected alert labels in `observability/alerts/nodepool-slo-prometheusRule_test.yaml`? The new `max by (...)` allowlist is meant to preserve routing labels as well as deduplicate replica noise, but the current fixture only proves that `instance` is dropped. As written, the test would still pass if `environment` or `region` were accidentally removed from the allowlist later.

SudoBrendan:
alert-tester results show a >10x decrease in alert firings, but the same number of _impacted_ resource/verb pairs. This appears to be the correct fix.

I'm confused though - why CorrelationID didn't resolve this already. We may need to look further into why we were seeing these duplicate ICMs for the same resource/verb pairs or if additional ICM Connector config could prevent this kind of thing in the future... the tickets don't lie though - multiple tickets for the same correlationIDs.

/approve
/lgtm

PR #6352, opened 2026-07-31
fix(ci): repair dev-ci privileged pipeline (pin mock-identity certs, shorten RG name, drop legacy cleanup)

roivaz:
This PR bundles the fixes required to make the `dev-ci` privileged grants pipeline (`make dev-ci-privileged-local-run`) safely runnable again after #5976. **All of them need to land for the pipeline to be usable.**

## 1. Pin mock-identity certificates instead of SNI (CI outage fix)

**Why.** #5976 switched the mock Entra apps from pinned-certificate auth (`keyCredentials`) to Subject Name and Issuer (SNI) trust (`trustedSubjectNameAndIssuers`). The mock certs are **self-signed** and Clusters Service authenticates in send-certificate-chain mode, so Entra validates the full chain and rejects the self-signed leaf with `AADSTS7000213: Invalid certificate chain`, with **no thumbprint fallback**. This broke mock-SP auth across **DEV and INT** and took CI down until SNI was removed and the pinned certs restored by hand. Re-running the pipeline as it stands on `main` re-breaks auth.

**What.**
- `modules/entra/app.bicep`: use separate conditional application resources for managed and unmanaged credentials. The Microsoft Graph Bicep provider rejects `keyCredentials: null` rather than omitting it, so the unmanaged path must use a resource declaration that does not contain `keyCredentials` at all; pinned certificates now survive redeployment, while callers that opt in can still reconcile credentials explicitly.
- `templates/mock-identity-apps.bicep`: stop setting `trustedSubjectNameAndIssuers` (an empty array now **clears stale SNI** on redeploy — self-healing) and no longer manage `keyCredentials`. Apps + SPs are still created; auth is reconciled by the following Shell step. Drop the now-unused `poolCertBaseDns` param.
- `tooling/entra-app-credentials`: use the Key Vault SDK plus Microsoft's official Graph SDK to manage the complete mock-certificate lifecycle. Normal reconciliation creates only missing self-signed certificates, preserves every existing certificate policy, skips matching app credentials, and verifies updates by thumbprint. Explicit disruptive rotation requires `--rotate` plus the exact `--confirm-disruptive-rotation=replace-key-vault-certificates` phrase, reuses the existing Key Vault policy, waits for a genuinely new current certificate, and then pins it.
- `dev-infrastructure/dev-ci/e2e-subscription-rbac-grants/pipeline.yaml`: build the CLI and run DEV and INT certificate reconciliation immediately after their corresponding app deployments. Each RBAC step depends on successful reconciliation, so the privileged flow is `apps → create missing certificates and pin current thumbprints → RBAC`. Increasing the configured MSI mock pool size creates and wires the missing indexed Azure resources automatically; decreasing it intentionally leaves higher indexes abandoned.
- Remove the Azure-writing mock app/certificate Make targets and the old `create-kv-cert.sh` implementation. The privileged pipeline is the only routine lifecycle entrypoint. Rotation remains an exceptional, documented direct CLI operation.

**Replacement safety.** The mock applications are dedicated to one pinned certificate, so the desired state is exactly one `AsymmetricX509Cert` / `Verify` credential. The CLI makes replacement of the complete `keyCredentials` collection explicit with the required `--replace-all` flag; this is not an implicit generic default. Rotation is intentionally disruptive for DEV CI and fails rather than creating a certificate when the requested existing certificate cannot be found.

**Shell-step identity.** The DEV privileged entrypoint is executed locally through templatize by an OWNERS-group member. Templatize does not impersonate `shellIdentity` for Shell steps; it executes them with the caller's copied Azure CLI credentials. (`shellIdentity` remains as an `unused-by-templatize` placeholder because the shared pipeline schema requires the field.) The caller therefore needs Key Vault certificate create/read and Graph app-write/ownership in addition to subscription Owner for the RBAC steps.

## 2. Shorten dev-ci mock-identities resourceGroup label

**Why.** The pipeline schema limits a resourceGroup's logical `name` to **12 chars**; the `mock-identities` label (15 chars) added in #5976 fails schema validation, so the pipeline can't load at all.

**What.** Rename the label `mock-identities` → `mock-idents` (11 chars) and update the dependency references. Only the pipeline's internal label changes; the deployed Azure resource group (`{{ .global.rg }}`) is unchanged.

## 3. Drop legacy mock-identity cleanup script

Remove `dev-infrastructure/scripts/delete-legacy-mock-identity-rbac.sh` (and its docs reference) — a one-time helper to purge legacy random-named role assignments before the declarative Bicep grants took over. That migration is complete.

## Testing

- Bicep build clean for `app.bicep`, `mock-identity-apps.bicep`, and the other `app.bicep` callers (geneva/ci-bot/mise).
- CLI: `go test -race ./...`, `go vet ./...`, and CGO-disabled builds pass.
- Unit tests cover create-if-missing, established certificate policy construction, policy-preserving rotation, asynchronous Key Vault operation polling/failure reporting, waiting for a new certificate version, explicit rotation confirmation, missing-certificate rotation rejection, duplicate/shared-certificate rotation rejection, unchanged credentials, explicit replacement, dry-run, aggregated failures, Graph retries, and thumbprint verification.
- Templatize entrypoint graph resolves both dependency chains as `mock-identity-apps → pin-mock-certs → mock-identity-rbac` (DEV and INT). Pipeline inspection resolves all app, vault, DNS, and pool variables to the expected values.
- Config materialization and Helm fixture tests pass with no generated drift.
- Read-only live reconciliation previously recognized all **23 DEV** and **3 INT** mock apps as already pinned to their current Key Vault certificates (`unchanged=26`, `failed=0`).
- Live-validated the pinning mechanism during the incident: removing SNI + restoring pinned `keyCredentials` on the DEV + INT mock apps drained `AADSTS7000213` and restored cluster provisioning.
- Actively acquired real Entra tokens as all 3 INT mock identities with x5c/send-certificate-chain enabled after restoration; all succeeded.
- Ran the mock-identity RBAC steps of the privileged pipeline against DEV + INT after the resource-group label rename; both loaded and reconciled cleanly.
- Per request, the new certificate creation and rotation paths were **not** executed against Azure; they are validated locally and left for maintainer canary testing.

The broad `pipeline validate --only-changed` command remains blocked by pre-existing restricted-template findings in existing range-based bicepparam templates. The targeted templatize graph/inspection path used above fully parses this pipeline, resolves its config references, and validates the new step schema/dependencies.


roivaz:
Superseeded by https://github.com/Azure/ARO-HCP/pull/6376

PR #6354, opened 2026-07-31
feat(observability): add empty RP HCP alert lane for future use

mmazur:
https://redhat.atlassian.net/browse/AROSLSRE-1641

Between this and #6342 SRE should be able to self service their current and future alerts.


janboll (commented):
/lgtm

PR #6355, opened 2026-07-31
test: add E2E coverage for deleting an HCP cluster with missing MIs

aklymenk:
https://issues.redhat.com/browse/ARO-23172

### What

Adds an E2E test that verifies a customer can delete a hosted ARO HCP cluster even when the cluster's managed identities were deleted first.

The test:

1. Creates a customer resource group, customer infrastructure (NSG, VNet, subnet, KeyVault) and **dedicated** managed identities.
2. Creates an HCP cluster and confirms it is viable.
3. Deletes **all** of the cluster's managed identities.
4. Deletes the cluster and asserts that both the cluster resource and its managed resource group are actually gone (HTTP 404).

It also adds a small framework helper, `DeleteUserAssignedIdentities`, because the Azure credentials on the test context are unexported and a spec in `package e2e` cannot construct an `armmsi` client itself.

### Why

Customers can delete the managed identities backing a running HCP cluster.

### PR Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/test/e2e/cluster_delete_missing_identities.go
+++ b/test/e2e/cluster_delete_missing_identities.go
@@ -1,0 +1,181 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("Customer", func() {
+	It("should be able to delete an HCP cluster whose managed identities were deleted first",
+		labels.RequireNothing,
+		labels.High,
+		labels.Positive,
+		labels.StageAndProdOnly,
+		// This test creates its own dedicated identities (useMsiPool=false) so it can
+		// safely delete them, and never leases from the shared identity container pool.
+		labels.MIContainers(0),
+		func(ctx context.Context) {
+			const customerClusterName = "missing-mi-hcp-cluster"
+			tc := framework.NewTestContext()
+
+			By("creating the customer resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "delete-mi-rg", tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create customer resource group")
+
+			clusterParams := framework.NewDefaultClusterParams20240610()
+			clusterParams.ClusterName = customerClusterName
+			managedResourceGroupName := framework.SuffixName(*resourceGroup.Name, "-managed", 64)
+			clusterParams.ManagedResourceGroupName = managedResourceGroupName
+
+			By("deploying customer infrastructure (NSG, VNet, subnet, KeyVault)")
+			customerInfraResult, err := tc.CreateBicepTemplateAndWait(ctx,
+				framework.WithTemplateFromFS(TestArtifactsFS, "test-artifacts/generated-test-artifacts/modules/customer-infra.json"),
+				framework.WithDeploymentName("customer-infra-"+customerClusterName),
+				framework.WithScope(framework.BicepDeploymentScopeResourceGroup),
+				framework.WithClusterResourceGroup(*resourceGroup.Name),
+				framework.WithParameters(map[string]interface{}{
+					"clusterName": customerClusterName,
+				}),
+				framework.WithTimeout(45*time.Minute),
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to deploy customer infrastructure")
+			clusterParams, err = framework.PopulateClusterParamsFromCustomerInfraDeployment20240610(clusterParams, customerInfraResult)
+			Expect(err).NotTo(HaveOccurred(), "failed to populate cluster params from customer infra deployment")
+
+			// This test deletes the cluster's managed identities, so it must own dedicated
+			// identities it can safely delete. Staging and Production CI runs with POOLED_IDENTITIES=true,
+			// where the standard helper would lease managed identities from a shared, reused
+			// pool; deleting those would corrupt the pool for other specs and consume Azure
+			// directory quota. We therefore always create dedicated identities in the customer
+			// resource group (useMsiPool=false), regardless of the pooled-identities setting,
+			// and never lease from the pool.
+			By("deploying dedicated managed identities in the customer resource group")
+			identities := framework.NewDefaultIdentitiesWithSuffix(customerClusterName)
+			managedIdentitiesResult, err := tc.CreateBicepTemplateAndWait(ctx,
+				framework.WithTemplateFromFS(TestArtifactsFS, "test-artifacts/generated-test-artifacts/modules/managed-identities.json"),
+				framework.WithDeploymentName(fmt.Sprintf("mi-%s-%s", customerClusterName, rand.String(6))),
+				framework.WithScope(framework.BicepDeploymentScopeSubscription),
+				framework.WithLocation(tc.Location()),
+				framework.WithParameters(map[string]interface{}{
+					"useMsiPool":               false,
+					"clusterResourceGroupName": *resourceGroup.Name,
+					"msiResourceGroupName":     *resourceGroup.Name,
+					"identities":               identities,
+					"rbacScope":                framework.RBACScopeResourceGroup,
+					"nsgName":                  clusterParams.NsgName,
+					"vnetName":                 clusterParams.VnetName,
+					"subnetName":               clusterParams.SubnetName,
+					"keyVaultName":             clusterParams.KeyVaultName,
+					"clusterName":              customerClusterName,
+				}),
+				framework.WithTimeout(45*time.Minute),
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to deploy dedicated managed identities")
+			clusterParams, err = framework.PopulateClusterParamsFromManagedIdentitiesDeployment20240610(clusterParams, managedIdentitiesResult)
+			Expect(err).NotTo(HaveOccurred(), "failed to populate cluster params from managed identities deployment")
+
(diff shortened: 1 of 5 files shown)


test/e2e/cluster_delete_missing_identities.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("Customer", func() {
+	It("should be able to delete an HCP cluster whose managed identities were deleted first",
+		labels.RequireNothing,
+		labels.High,
+		labels.Positive,
```
mbukatov, line 40: I would label this as a negative test case.

test/e2e/cluster_delete_missing_identities.go
```diff
@@ -0,0 +1,180 @@
...
+			// identities it can safely delete. Staging and Production CI runs with POOLED_IDENTITIES=true,
+			// where the standard helper would lease managed identities from a shared, reused
+			// pool; deleting those would corrupt the pool for other specs and consume Azure
+			// directory quota. We therefore always create dedicated identities in the customer
+			// resource group (useMsiPool=false), regardless of the pooled-identities setting,
+			// and never lease from the pool.
+			By("deploying dedicated managed identities in the customer resource group")
+			identities := framework.NewDefaultIdentitiesWithSuffix(customerClusterName)
+			managedIdentitiesResult, err := tc.CreateBicepTemplateAndWait(ctx,
+				framework.WithTemplateFromFS(TestArtifactsFS, "test-artifacts/generated-test-artifacts/modules/managed-identities.json"),
```
mbukatov, line 83: Do you really have to deploy MI module separately? Can't you just deploy the customer infra all together? I will have to recheck this.

test/e2e/cluster_delete_missing_identities.go
```diff
@@ -0,0 +1,180 @@
...
+				if errors.As(err, &respErr) {
+					if respErr.StatusCode == http.StatusNotFound {
+						return nil
+					}
+					logObservedDelta(fmt.Sprintf("HTTP %d (%s)", respErr.StatusCode, respErr.ErrorCode))
+				} else {
+					logObservedDelta(fmt.Sprintf("non-HTTP error: %v", err))
+				}
+				return fmt.Errorf("unexpected error getting managed resource group %q: %w", managedResourceGroupName, err)
+			}, 15*time.Minute, 30*time.Second).Should(Succeed(), "expected managed resource group %q to be deleted", managedResourceGroupName)
```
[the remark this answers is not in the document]
mgahagan73: Correct me if I'm wrong, but I don't think we have a standardized time for resource group deletion or if there is any published SLA we could reference. I'm ok with leaving the hard-coded timeout as-is, investigating at a later time and creating a follow-up PR. Same thing with resource creation.

test/e2e/cluster_delete_missing_identities.go
```diff
@@ -0,0 +1,180 @@
...
+			const customerClusterName = "missing-mi-hcp-cluster"
+			tc := framework.NewTestContext()
+
+			By("creating the customer resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "delete-mi-rg", tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create customer resource group")
+
+			clusterParams := framework.NewDefaultClusterParams20240610()
+			clusterParams.ClusterName = customerClusterName
+			managedResourceGroupName := framework.SuffixName(*resourceGroup.Name, "managed", 64)
```
mgahagan73: Change "managed" to "-managed" here so we keep naming consistent with our other managed RG names.

test/e2e/cluster_delete_missing_identities.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("Customer", func() {
+	It("should be able to delete an HCP cluster whose managed identities were deleted first",
```
machi1990, line 37: @aklymenk be aware of this bug https://redhat.atlassian.net/browse/ARO-28620 that prevents cluster deletion if the SMI is deleted. 
CS has a PR to fix it https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/398.

test/e2e/cluster_delete_missing_identities.go
aklymenk: I put "managed" instead of "-managed" here because `framework.SuffixName` returns this:

```go
name := fmt.Sprintf("%s-%s", base, suffix)
```

So it already puts `-` before suffix. When it is called with `-managed` suffix, returned value will end with `--managed`. Maybe it would be a good idea to have a separate PR that changes `-managed` to `managed` in all E2E tests, but for now I will change it to `-managed` to be consistent.`

test/e2e/cluster_delete_missing_identities.go
aklymenk: Good question — I looked into it. There is a combined template (`test/e2e-setup/bicep/infra-only.bicep`) that nests both modules (customer infra and managed identities) in a single deployment. Two things block it here today:

**1. It doesn't expose `useMsiPool`.**

`infra-only.bicep` doesn't pass that parameter through to the managed-identities module, so it inherits the module default `param useMsiPool bool = true` — i.e. pooled mode. This test *deletes* the identities it is given, so it must run with `useMsiPool=false` and dedicated identities. Using `infra-only.json` as-is would make it delete identities out of the shared pool, corrupting it for every other spec that later leases the same container (and burning AAD directory quota). Stage and Prod CI runs with `POOLED_IDENTITIES=true`, so this isn't hypothetical.

**2. It doesn't re-export the customer-infra outputs.**

Its only outputs are `userAssignedIdentitiesValue` and `identityValue`. This test creates the cluster through the SDK (`CreateHCPClusterFromParam20240610`), which needs `nsgID`, `vnetSubnetID`, `vnetIntegrationSubnetID`, `keyVaultName`, `etcdEncryptionKeyName` and `etcdEncryptionKeyVersion` — all consumed by `PopulateClusterParamsFromCustomerInfraDeployment20240610`.

**Worth noting:** the two-deployment split isn't something I introduced. The standard bundled helper, `CreateClusterCustomerResources20240610`, does exactly the same two sequential deployments internally — customer-infra first, then `DeployManagedIdentities` fed with `nsgName`/`vnetName`/`subnetName`/`keyVaultName` from the first deployment's outputs.

The reason I don't just call `CreateClusterCustomerResources20240610` is that its MI step routes through `DeployManagedIdentities`, which decides pooled-vs-dedicated from the `POOLED_IDENTITIES` env var and offers no way to force non-pooled.

**Happy to change it if you'd prefer** — the options I see:

- **(a)** Keep as-is (two explicit deployments, `useMsiPool=false` pinned at the call site).
- **(b)** Add a `useMsiPool` parameter and customer-infra passthrough outputs to `infra-only.bicep`, then use it here. Cleaner call site, but it modifies a shared setup template.
- **(c)** Add an option to `DeployManagedIdentities` (e.g. `WithDedicatedIdentities()`) to force non-pooled, and use `CreateClusterCustomerResources20240610`. Most reusable if other tests ever need destructive identity scenarios.

I went with (a) to keep the blast radius inside this test. Let me know which you'd like.

mbukatov: Actually, it seems that both modules/managed-identities.bicep and modules/customer-infra.bicep are used from the main bicep file, so these are on the same level.

So no need to change anything.

mbukatov: /me would have resolved this thread if there was a way to do it

test/e2e/cluster_delete_missing_identities.go
aklymenk: @machi1990 do you think this PR shouldn't be merged until [this PR](https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/398) is merged? this new E2E test has passed twice in both stage and prod.

machi1990: Yes. In the interest of having a stable test, I believe we shouldn't merge until CS has that card

mbukatov (commented):
/lgtm

test/e2e/cluster_delete_missing_identities.go
mbukatov: I'm little puzzled that we don't see this test case to fail even with [ARO-28620](https://redhat.atlassian.net/browse/ARO-28620) bug - as I hoped that this will show how the test handles error reporting. Is that really expected?

machi1990: In lower envs (e.g prow, CI, int) we are using a msi mock credentials in place of all controlplane and SMI credential, the test will always pass even if those MIs are deleted as the really cluster's MI  passed isn't used.
In higher envs (stage/prod), It really depends on how the deletion of the MI is propagated across systems vs the progression of the cluster deletion.

machi1990: @aklymenk the change has been rolled out to prod

machi1990: Might be worth trying out to see if it passes

machi1990: @aklymenk FYI @JameelB mentioned that the issue isn't actually fixed yet on the CS side regarding the SMI. There are discussion ongoing.

mbukatov:
I see that a rerun in prod as well as stage failed on `InternalServerError`:

```
cluster deletion did not complete before the deadline; [clusterServiceDeletion] ClusterService cluster /api/aro_hcp/v1alpha1/clusters/2s13oc0r60vrsni6sup6ck1pmqctmnep still exists (deletion dispatched at 2026-08-06T13:32:47Z); [clusterServiceStatus] ClusterService state is \"uninstalling\"; [descendantResources] remaining resources: 1 microsoft.redhatopenshift/hcpopenshiftclusters/serviceProviderClusters
```

I assume this is what @machi1990 is raising [above](https://github.com/Azure/ARO-HCP/pull/6355#discussion_r3730271296):

> [Andrii Klymenko (@aklymenk)](https://github.com/aklymenk) FYI [Jameel Briones (@JameelB)](https://github.com/JameelB) mentioned that the issue isn't actually fixed yet on the CS side regarding the SMI. There are discussion ongoing.

This unfortunately means that we should wait until the fix is finished.

/hold
/lgtm

test/e2e/cluster_delete_missing_identities.go
aklymenk: @machi1990 could you please reference this discussion? is [ARO-28620](https://redhat.atlassian.net/browse/ARO-28620) still used to track this issue?

machi1990: Yes, it is the same card.

mvacula02 (approved):

mvacula02:
/hold cancel

the fix is live in both STG and PROD as per [ARO-28620](https://redhat.atlassian.net/browse/ARO-28620)

mvacula02 (approved):
/lgtm

stevekuznetsov:
/hold

for justification on running this not in dev

mvacula02:
branch owner is no longer in the organisation, can't rebase 

 #6610 (clean cherry-pick of this PR) has been created instead and this PR should be closed

PR #6356, opened 2026-07-31
feat: add nodepool ephemeral OS disk and VM quota validations

oshrirao:
Jira issue :
https://redhat.atlassian.net/browse/ARO-9533
- https://redhat.atlassian.net/browse/ARO-9983
- https://redhat.atlassian.net/browse/ARO-9982

### What

Adds two backend async NodePool validations :

1. **Ephemeral OS disk** — when `osDisk.diskType` is `Ephemeral`, verifies the requested VM size(instance type) SKU advertises `EphemeralOSDiskSupported`.
2. **VM quota** — verifies the subscription has enough Compute vCPU quota in the node pool location for the requested scale (fixed `replicas`, or autoscaler `max`), checking both the VM **family** quota and total **regional** (`cores`) quota.

### Why

Customers can request ephemeral disks on unsupported SKUs, or sizes/replica counts that exceed Azure quota. Catching this asynchronously in the RP (via Resource SKUs + Usage APIs) surfaces clear validation.

### Testing

- [x] Unit tests: ephemeral validation (skip managed / pass / fail capability / SKU errors)
- [x] Unit tests: quota validation (fixed replicas, autoscaling max, family fail, regional fail, unlimited limit, missing usage/SKU)
- [x] Manual test: ephemeral pass (`D8s_v3`) and fail (`D4s_v5` / `D8s_v5`)
- [x] Manual test: quota pass (small scale / autoscaling) and fail (fixed replicas + autoscaling `max` over family remaining)

Depends on SKU cache work in `ARO-15594`(https://github.com/Azure/ARO-HCP/pull/6111)

### PR Checklist

- [x] Self-reviewed the diff


backend/pkg/controllers/validationcontrollers/validations/resource_sku_capabilities.go
```diff
@@ -0,0 +1,83 @@
...
+		if strings.EqualFold(*capability.Name, capabilityName) {
+			return capability.Value
+		}
+	}
+	return nil
+}
+
+// isCapabilityEphemeralOSDiskSupported reports whether the Resource SKU
+// advertises EphemeralOSDiskSupported with value "True" (case-insensitive).
+func isCapabilityEphemeralOSDiskSupported(sku *armcompute.ResourceSKU) (supported bool, found bool) {
```
miguelsorianod: Do not use named returns. Document appropriately instead

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation.go
```diff
@@ -0,0 +1,192 @@
...
+		return utils.TrackError(fmt.Errorf("failed to create usage client: %w", err))
+	}
+
+	familyUsage, regionalUsage, err := v.lookupFamilyAndRegionalVCPUUsages(ctx, usageClient, nodePool.Location, family)
+	if err != nil {
+		return err
+	}
+
+	var errs []error
+	if !isVCPUUsageSufficient(familyUsage, requiredVCPUs) {
```
miguelsorianod: Are isVCPUUsageSufficient and remainingVCPUsFromComputeUsage really needed? can't we calculate it inline storing it in variables? I am unsure about the value of those functions. Those are also package-level, but specify vcpus and take a general usage which is not guaranteeed to be a vcpu one, which seems contradictory

backend/pkg/utils/validationutils/azure_nodepool_vm_quota_validation.go
```diff
@@ -0,0 +1,192 @@
...
+	if regionalUsage == nil {
+		return nil, nil, utils.TrackError(fmt.Errorf("compute usage %q (total regional vCPUs) was not found in location %q", computeUsageNameTotalRegionalVCPUs, location))
+	}
+	return familyUsage, regionalUsage, nil
+}
+
+// isValidComputeUsage reports whether usage has the fields required for a vCPU
+// remaining calculation (non-empty Name.Value, CurrentValue, and Limit).
+func isValidComputeUsage(usage *armcompute.Usage) bool {
+	return usage != nil && usage.Name != nil && usage.Name.Value != nil && *usage.Name.Value != "" && usage.CurrentValue != nil && usage.Limit != nil
```
miguelsorianod, line 179: if are are absolutely sure that limit nil will never be returned (are we fully sure?), shouldn't we return an error instead of failing silently? https://github.com/shitaljante/ARO-HCP/pull/1#discussion_r3690167134

miguelsorianod: Take into account that the function is named `isValidComputeUsage and receives a general vCPU usage. However the function description describes vCPU. This seems contradictory.
Also the answer about nil limit will change depending on whether you are scoping this to general usages or only to vcpu ones

backend/pkg/utils/validationutils/azure_nodepool_vm_quota_validation.go
```diff
@@ -0,0 +1,192 @@
...
+//
+// Parameters:
+//   - usageClient: Compute Usage client for the customer subscription
+//   - location: Azure region of the node pool (Usage API is per-location)
+//   - family: Resource SKU family name (e.g. "standardDSv3Family"), matched
+//     case-insensitively against Usage.Name.Value
+//
+// Returns the family usage and regional ("cores") usage, or an error if listing
+// fails or either entry is missing from the Usage API response.
+func (v *AzureNodePoolVMQuotaValidation) lookupFamilyAndRegionalVCPUUsages(ctx context.Context, usageClient azureclient.UsageClient, location string, family string) (*armcompute.Usage, *armcompute.Usage, error) {
```
miguelsorianod, line 137: how sure about what you replied in https://github.com/shitaljante/ARO-HCP/pull/1#discussion_r3690160288? If you are absolutely sure then we are fine. If not, change to return an error if the unexpected happens

backend/pkg/utils/validationutils/azure_nodepool_vm_quota_validation.go
```diff
@@ -0,0 +1,192 @@
...
+	logger := utils.LoggerFromContext(ctx)
+	var familyUsage *armcompute.Usage
+	var regionalUsage *armcompute.Usage
+
+	pager := usageClient.NewListPager(location, nil)
```
miguelsorianod, line 142: About this: https://github.com/shitaljante/ARO-HCP/pull/1#discussion_r3690203675. That's a non small amount of results. Let's see how it behaves. If we detect issues it could be that we need to slow down the execution rate of this validation. This will be controllable once we have the validation controller redesign is completed

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_ephemeral_os_disk_validation_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+func TestAzureVMSizeSupportsEphemeralOSDiskValidation_Validate(t *testing.T) {
+	ctx := context.Background()
+	cluster := &api.HCPOpenShiftCluster{}
+	subscription := newTestSubscription()
+
+	tests := []struct {
+		name       string
+		nodePool   *api.HCPOpenShiftClusterNodePool
+		setupSKU   func(skuReader *cachedreader.MockVirtualMachineResourceSKUsCachedReader)
```
miguelsorianod: nitpick: rename the test arg parameter to setupMockVMSKUCachedReader

miguelsorianod: apply this to all other similar ones

backend/pkg/utils/validationutils/azure_nodepool_ephemeral_os_disk_validation_test.go
```diff
@@ -0,0 +1,192 @@
...
+			},
+		},
+		{
+			name:     "ephemeral OS disk succeeds when capability is true (case-insensitive)",
+			nodePool: newTestNodePool(t, api.OsDiskTypeEphemeral, testVMSize),
+			setupSKU: func(skuReader *cachedreader.MockVirtualMachineResourceSKUsCachedReader) {
+				skuReader.EXPECT().
+					GetVirtualMachineSKU(gomock.Any(), testTenantID, testSubscriptionID, testVMSize).
+					Return(makeTestVMResourceSKU(testVMSize, &armcompute.ResourceSKUCapabilities{
+						Name:  ptr.To("ephemeralosdisksupported"),
```
miguelsorianod, line 134: Isn't this computeResourceSKUCapabilityNameEphemeralOSDiskSupported?

oshrirao: For case-insensitive test case, it is added.
computeResourceSKUCapabilityNameEphemeralOSDiskSupported is  "EphemeralOSDiskSupported"

backend/pkg/utils/validationutils/azure_nodepool_vm_quota_validation.go
oshrirao: Okay works.

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
```diff
@@ -0,0 +1,505 @@
...
+func newQuotaTestNodePool(t *testing.T, replicas int32, autoScaling *api.NodePoolAutoScaling) *api.HCPOpenShiftClusterNodePool {
+	t.Helper()
+	np := newTestNodePool(t, api.OsDiskTypeManaged, testVMSize)
+	np.Location = testLocation
+	np.Properties.Replicas = replicas
+	np.Properties.AutoScaling = autoScaling
+	return np
+}
+
+func usageListPager(usages []*armcompute.Usage, fetchErr error) *runtime.Pager[armcompute.UsageClientListResponse] {
```
miguelsorianod: this is a package-level function. It's intended for testing but it has a general name. Change to be clear about its testing purpose. newTestUsageListPager

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
```diff
@@ -0,0 +1,505 @@
...
+			if fetchErr != nil {
+				return armcompute.UsageClientListResponse{}, fetchErr
+			}
+			idx++
+			return pages[idx], nil
+		},
+	})
+}
+
+func expectGetVirtualMachineSKU(skuReader *cachedreader.MockVirtualMachineResourceSKUsCachedReader, sku *armcompute.ResourceSKU, err error) {
```
miguelsorianod: don't put expectGetVirtualMachineSKU and expectUsageList  as package-level variables if they are only used in one test. Put them inside the TestAzureNodePoolVMQuotaValidation_Validate test

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+func TestAzureNodePoolVMQuotaValidation_Validate(t *testing.T) {
+	ctx := utils.ContextWithLogger(context.Background(), logr.Discard())
+	cluster := &api.HCPOpenShiftCluster{}
+	subscription := newTestSubscription()
+
+	tests := []struct {
+		name       string
+		nodePool   *api.HCPOpenShiftClusterNodePool
+		setupMocks func(ctrl *gomock.Controller, skuReader *cachedreader.MockVirtualMachineResourceSKUsCachedReader, fpaBuilder *azureclient.MockFirstPartyApplicationClientBuilder)
```
miguelsorianod: separate each one into its own setupMock function. This includes the sku client, which I see is being created in expectUsageList which it shouldn't

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
miguelsorianod: I am not sure of the value of expectGetVirtualMachineSKU. You are almost repeating the same than just setting the desired expection in each place. Inline directly

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
miguelsorianod: Maybe there are some that require passing together because one being set with the other (maybe fpa builder returning a specific mock). In that case they can be together

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
```diff
@@ -0,0 +1,505 @@
...
+}
+
+func TestAzureNodePoolVMQuotaValidation_lookupFamilyAndRegionalVCPUUsages(t *testing.T) {
+	ctx := utils.ContextWithLogger(context.Background(), logr.Discard())
+	validation := &AzureNodePoolVMQuotaValidation{}
+
+	tests := []struct {
+		name             string
+		usages           []*armcompute.Usage
+		fetchErr         error
```
miguelsorianod: don't have fetchErr. have a setup of usage mock, similar to the other test. In this specific case it's not very important as what it tests is small and unlikely to change over time, but get used to do it that way. The reason is that for tests that test broader logic, you start with fetchErr, and when you want to mock another part of it you add another bool (or string, or int, ...) and then when a new need appear another, and then another and another... and you end up with 70 different setup booleans/strings/int which make the test body and logic very hard to maintain and understand over time. What you really want conceptually is to provide the state of the mock you want

backend/pkg/utils/validationutils/azure_nodepool_vm_quota_validation.go
oshrirao: Yes i am sure that there won't be duplicate in the same region. Quotas are predefined by Azure for specific resource types and families

miguelsorianod: :+1: all good then

backend/pkg/utils/validationutils/azure_nodepool_ephemeral_os_disk_validation_test.go
miguelsorianod: ah, I see. :+1: . Resolving

backend/pkg/controllers/validationcontrollers/validations/resource_sku_capabilities.go
oshrirao: Incorporated the suggested changes. Documented appropriately.

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation.go
oshrirao: Incorporated the suggested changes. Made it inline

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_ephemeral_os_disk_validation_test.go
oshrirao: Incorporated the suggested changes.

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
oshrirao: Incorporated the suggested changes.

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
oshrirao: Incorporated the suggested changes.

oshrirao: Incorporated the suggested changes. inline now.

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
oshrirao: Seperated them as two setupMocks. Incorporated the suggested changes.

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
oshrirao: Incorporated the suggested changes.

backend/pkg/utils/validationutils/azure_nodepool_vm_quota_validation.go
oshrirao: Yes, we have confirmed that limit = nil will never be returned. I also updated the description of isValidComputeUsage to remove the contradiction between the function name and its description.

backend/pkg/controllers/validationcontrollers/validations/azure_nodepool_vm_quota_validation_test.go
```diff
@@ -0,0 +1,552 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			usageClient := azureclient.NewMockUsageClient(ctrl)
+			tt.setupMockUsageClient(usageClient)
+
+			familyUsage, regionalUsage, err := validation.lookupFamilyAndRegionalVCPUUsages(ctx, usageClient, testLocation, testVMFamily)
```
miguelsorianod: using `testLocation` and `testVMFamily` is incorrect here. Those should be test arguments

oshrirao: Added them as test arguments. Thanks.

miguelsorianod:
/hold cancel

This is ready to merge. I verified in the previous CI run before the rebase that the validation result is persisted as succeeded for both validations.

PR #6363, opened 2026-07-31
docs: add runbook for SWIFT v2 delegated-NIC wedged node mitigation

avollmer-redhat:
## Why

Management-cluster nodes on SWIFT v2 pools can enter a permanent `FailedCreatePodSandBox` state due to an Azure platform issue (Cloudnet/RNC) where the delegated FrontendNIC virtual function is torn down after CNI attach (ICM 832382845). This has now recurred in **three regions** — prod uksouth, prod australiaeast, and prod centralindia — each time requiring manual JIT intervention with no existing runbook to follow.

The centralindia occurrence ([AROSLSRE-1717](https://redhat.atlassian.net/browse/AROSLSRE-1717)) exhibits a chronic node-hopping pattern where the wedged-node role migrates across VMSS instances over 2+ weeks, with DNC backend HTTP 500s and PodNetwork status-patch conflicts indicating regional Swift v2 control-plane degradation. This confirms the issue is not a one-off and strengthens the case for a documented mitigation procedure.

## What

Adds `docs/ops/mitigate-swiftv2-wedged-node.md` covering:

- **Diagnosis**: Kusto queries to identify wedged nodes and distinguish hard-wedge from transient flap (Rael's flap-vs-wedge decision rule: 0 successes for 30-60 min = wedge; interleaved successes = flap, do not recycle)
- **Pre-checks**: AKS node auto-repair status, `recreate-system-pool` self-healer applicability for system pool nodes
- **Mitigation**: ACN log collection (Steve Kuznetsov's script), cordon, VMSS instance deletion
- **Post-mitigation**: Kusto verification query, cluster-create recovery validation
- **Node healer design takeaways**: Criteria for future automated remediation
- **Appendix**: Full ACN log-collection script

## References

- ICM 832382845 (Azure-owned root cause)
- [AROSLSRE-1524](https://redhat.atlassian.net/browse/AROSLSRE-1524) (prod/uksouth)
- [AROSLSRE-1563](https://redhat.atlassian.net/browse/AROSLSRE-1563) / [AROSLSRE-1585](https://redhat.atlassian.net/browse/AROSLSRE-1585) (prod/australiaeast)
- [AROSLSRE-1717](https://redhat.atlassian.net/browse/AROSLSRE-1717) (prod/centralindia — chronic, node-hopping variant)
- AROSLSRE-1642/1643 (flap vs. wedge investigation)


avollmer-redhat:
Closing per team discussion — this mitigation is being automated in the node healer, and the procedures are already documented in the AROSLSRE JIRAs.

avollmer-redhat:
PR became relevant again as there are ongoing wedged nodes with the same pattern.

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,335 @@
+# Mitigation: SWIFT v2 Delegated-NIC Wedged Node
+
+## Overview
+
+Azure SWIFT v2 management-cluster nodes can enter a permanent failure state where every pod scheduled to the node fails with `FailedCreatePodSandBox`. The root cause is an Azure platform issue (Cloudnet/RNC) where the delegated FrontendNIC virtual function (VF) is torn down and re-added at the platform level just after CNI attach, leaving the node's network stack broken. The CNI errors are the symptom, not the cause.
```
raelga: Two facts worth putting right here, because they stop people digging in the wrong place.

```suggestion
Azure SWIFT v2 management-cluster nodes can enter a failure state where every pod scheduled to the node fails with `FailedCreatePodSandBox`. The root cause is an Azure platform issue (Cloudnet/RNC) where the delegated FrontendNIC virtual function (VF) is torn down and re-added at the platform level just after CNI attach, leaving the node's network stack broken. The CNI errors are the symptom, not the cause.

Two things to know before you start digging:

- **CNS is not the culprit, so do not spend time there.** On the cases we have analysed, `azure-cns` assigned and released the pod's IP cleanly on every single retry with `ReturnCode:Success`, and the `MultitenantPodNetworkConfig` resolved a secondary IP within about a second of the pod being scheduled. IP pool exhaustion is ruled out too. The fault is in the `azure-vnet` CNI plugin's endpoint-creation step, which consumes the CNS response and programs the interface into the pod netns.
- **`azure-vnet` has no logs in Kusto.** It runs as a host binary invoked by containerd, not as a pod, so nothing it emits is ingested into ServiceLogs. That is exactly why the log-collection script in the appendix exists: if Cloudnet/RNC want evidence, it has to be pulled off the node before you recycle it.
```

raelga (commented):
Read through this since it overlaps with the node-health work. The structure is right and the flap-vs-wedge decision gate is the correct shape for this runbook, so this is worth having. Left suggestions inline.

The one I would treat as must-fix before merge: **all three KQL blocks query a `ContainerEvent` table that does not exist in our Kusto.** The right table is `ServiceLogs.kubernetesEvents` with lowercase columns. As written, every diagnosis step in the runbook errors out with table-not-found, which is the worst possible failure mode for something you open during an incident.

The rest are freshness items: the node healer is no longer a "future" thing, the isolation symptom does not hold up against AROSLSRE-1717, centralindia is missing from the occurrence list, and node deletion has not proven durable so that deserves a caveat.

Happy to take any of these as follow-ups if you would rather land the doc first, apart from the Kusto tables.

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,335 @@
...
+
+The node will not self-heal. The standing mitigation is to **cordon and delete the wedged VMSS instance** so pods reschedule onto healthy nodes and AKS replaces the instance.
+
+## Tracking
+
+- **ICM**: 832382845 (Azure-owned, Cloudnet/RNC root cause)
+- **Past occurrences**:
+  - `prod uksouth` — node `aks-userswft2-40171262-vmss000004`, ~18h of continuous failure, ~2,584 events (AROSLSRE-1524)
+  - `prod australiaeast` — node `aks-userswft2-40474110-vmss00000f`, ~3 days of continuous failure, ~2,419 events (AROSLSRE-1563 / AROSLSRE-1585)
```
raelga, line 19: Missing the most recent one.

```suggestion
  - `prod australiaeast` — node `aks-userswft2-40474110-vmss00000f`, ~3 days of continuous failure, ~2,419 events (AROSLSRE-1563 / AROSLSRE-1585)
  - `prod centralindia`, node `aks-userswft3-38105468-vmss000005`, ongoing since 2026-08-01, ~1,074 events, took out three consecutive prod e2e runs on 2026-08-04 (AROSLSRE-1717)
```

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,335 @@
...
+## Symptoms
+
+All of these will be present on the affected node:
+
+| Signal | Detail |
+|---|---|
+| **Event** | `FailedCreatePodSandBox` at ~15s retry cadence, thousands of occurrences |
+| **Error messages** | `route ip+net: no such network interface`, `mtpnc is not ready`, `dhcp discover timed out` |
+| **Blast radius** | 100% of pods scheduled to the node fail; hosted-cluster router replicas are especially impacted → `KASLoadBalancerNotReachable` → cluster-create timeouts |
+| **Isolation** | Only one node is affected; other nodes in the same VMSS pool are healthy |
```
raelga: This one I would soften. It reads as a discriminator, and if someone uses it to rule the issue out they will get it wrong.

In AROSLSRE-1717 five HCP namespaces hit the identical CNI errors inside the same six-minute window, across more than one node, and four of them recovered. So concentration on a single node is the common case, not a reliable test.

```suggestion
| **Isolation** | Failures usually concentrate on one node, but a short cluster-wide burst across several nodes is possible (see AROSLSRE-1717). Confirm with per-node event counts rather than assuming a single node |
```

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,335 @@
...
+
+### 1. Identify the wedged node via Kusto (hcp-prod-{region})
+
+```kql
+// Look for sustained FailedCreatePodSandBox on a single node
+ContainerEvent
+| where Reason == "FailedCreatePodSandBox"
+| where Message has "no such network interface" or Message has "mtpnc is not ready" or Message has "dhcp discover timed out"
+| summarize EventCount = count(), FirstSeen = min(TimeGenerated), LastSeen = max(TimeGenerated) by NodeName
+| order by EventCount desc
```
raelga: `ContainerEvent` does not exist in our Kusto. Kubernetes events land in `ServiceLogs.kubernetesEvents`, and the columns are lowercase (`timestamp`, `host`, `reason`, `message`), see [docs/ai/kusto-debugging.md](https://github.com/Azure/ARO-HCP/blob/main/docs/ai/kusto-debugging.md). As written this query fails with table-not-found, which is the worst thing that can happen in a runbook someone opens mid-incident.

Same problem in the other two KQL blocks below.

```suggestion
// Look for sustained FailedCreatePodSandBox, per node.
// Run against the region's cluster, database ServiceLogs.
kubernetesEvents
| where timestamp > ago(7d)
| where reason == "FailedCreatePodSandBox"
| where message has "no such network interface"
     or message has "mtpnc is not ready"
     or message has "dhcp discover"
| summarize events=count(), nsAffected=dcount(eventNamespace), podsAffected=dcount(objectName),
            firstSeen=min(timestamp), lastSeen=max(timestamp)
  by cluster, host
| order by events desc
```

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,335 @@
...
+                             or Message has "dhcp discover timed out")),
+    SuccessCount = countif(Reason in ("Created", "Started")),
+    FirstFail = minif(TimeGenerated, Reason == "FailedCreatePodSandBox"),
+    LastFail = maxif(TimeGenerated, Reason == "FailedCreatePodSandBox"),
+    LastSuccess = maxif(TimeGenerated, Reason in ("Created", "Started"))
+    by NodeName
+| extend FailDuration = LastFail - FirstFail,
+         TimeSinceLastSuccess = LastFail - LastSuccess
+| project NodeName, FailCount, SuccessCount, FailDuration,
+          TimeSinceLastSuccess, FirstFail, LastFail, LastSuccess
```
raelga, line 97: Same table and column fix here. I also swapped `NodeName` for `host` and dropped the `let` for the cluster scope, since you almost always want to pin `cluster` too.

```suggestion
let suspect_node = "<node-name>";
let suspect_cluster = "<mgmt-cluster-name>";
let lookback = 2h;
kubernetesEvents
| where timestamp > ago(lookback)
| where cluster == suspect_cluster and host == suspect_node
| where reason in ("FailedCreatePodSandBox", "Created", "Started")
| summarize
    FailCount = countif(reason == "FailedCreatePodSandBox"
                        and (message has "no such network interface"
                             or message has "mtpnc is not ready"
                             or message has "dhcp discover")),
    SuccessCount = countif(reason in ("Created", "Started")),
    FirstFail = minif(timestamp, reason == "FailedCreatePodSandBox"),
    LastFail = maxif(timestamp, reason == "FailedCreatePodSandBox"),
    LastSuccess = maxif(timestamp, reason in ("Created", "Started"))
    by host
| extend FailDuration = LastFail - FirstFail,
         TimeSinceLastSuccess = LastFail - LastSuccess
| project host, FailCount, SuccessCount, FailDuration,
          TimeSinceLastSuccess, FirstFail, LastFail, LastSuccess
```

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,335 @@
...
+// Run against hcp-prod-{region} after node deletion.
+// Expect 0 results for the deleted node, and no new nodes exhibiting the same pattern.
+ContainerEvent
+| where TimeGenerated > ago(30m)
+| where Reason == "FailedCreatePodSandBox"
+| where Message has "no such network interface"
+     or Message has "mtpnc is not ready"
+     or Message has "dhcp discover timed out"
+| summarize EventCount = count(), LastSeen = max(TimeGenerated) by NodeName
+| order by EventCount desc
```
raelga: Same fix for the post-mitigation check.

```suggestion
// Run against the region's cluster, database ServiceLogs, after node deletion.
// Expect 0 results for the deleted node, and no new nodes showing the same pattern.
kubernetesEvents
| where timestamp > ago(30m)
| where reason == "FailedCreatePodSandBox"
| where message has "no such network interface"
     or message has "mtpnc is not ready"
     or message has "dhcp discover"
| summarize events=count(), lastSeen=max(timestamp) by cluster, host
| order by events desc
```

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,335 @@
...
+  --instance-ids <instance-id>
+```
+
+Alternatively, if you only have `kubectl` access:
+
+```bash
+kubectl delete node <node-name>
+```
+
+> **Note**: Deleting the node object from Kubernetes does not delete the underlying VMSS instance. For a full cleanup, use `az vmss delete-instances`. The AKS cluster autoscaler or node-pool reconciler will provision a replacement.
```
raelga: Worth adding a durability caveat here. AROSLSRE-1717 shows the faulty-node role migrating across VMSS instances over weeks, with clean handoffs (one instance stops failing the same minute the next one starts). Deleting the instance fixes the incident in front of you but has not held.

```suggestion
> **Note**: Deleting the node object from Kubernetes does not delete the underlying VMSS instance. For a full cleanup, use `az vmss delete-instances`. The AKS cluster autoscaler or node-pool reconciler will provision a replacement.

> **This mitigation is not durable.** The fault has been observed moving to a different instance in the same pool, sometimes within minutes of the previous one being recycled (AROSLSRE-1717). Recycling clears the immediate incident, it does not fix the underlying platform issue. Update ICM 832382845 on every recurrence so Azure has the pattern.
```

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,335 @@
...
+| Node pool | `userswft2` / `userswft3` (SWIFT v2 pools) | Any pool |
+
+## Appendix: Node Healer Design Takeaways
+
+From the investigation of the first observed self-heal (AROSLSRE-1642/1643), these criteria distinguish actionable wedges from transient flaps. Any future automated node healer should implement these rules:
+
+1. **Key on concurrent success on the node** -- do not wait passively for self-heal.
+2. **0 successful sandbox creations for >= 30-60 min of continuous route errors** → wedge → cordon/drain/delete.
+3. **Any successes interleaved with the failures** → flap → alert only, do not recycle.
+4. Both conditions are queryable in `hcp-prod-{region}` ServiceLogs.
```
raelga: This is no longer hypothetical. The healer shipped in [#6284](https://github.com/Azure/ARO-HCP/pull/6284) and is being turned on by default in [#6394](https://github.com/Azure/ARO-HCP/pull/6394) (AROSLSRE-1588), and it implements exactly these rules.

That also makes diagnosis cheaper than hand-writing Kusto, so it is worth pointing at from step 1 as well.

````suggestion
## Appendix: Automated Detection (node-health controller)

The criteria below came out of the first observed self-heal (AROSLSRE-1642/1643) and are now implemented by the `node-health` controller in mgmt-agent (AROSLSRE-1588):

1. **Key on concurrent success on the node**, do not wait passively for self-heal.
2. **0 successful sandbox creations for >= 30-60 min of continuous route errors** → wedge → cordon/drain/delete.
3. **Any successes interleaved with the failures** → flap → alert only, do not recycle.

Where it has rolled out, a wedged node carries:

- label `node-health.aro-hcp.azure.com/status=wedged`
- annotations `node-health.aro-hcp.azure.com/detector`, `/reason`, `/signature`, `/observed-at`
- metrics `nodehealth_node_wedged`, `nodehealth_detections_total`, `nodehealth_label_actions_total`

So the quickest first pass at diagnosis is:

```bash
kubectl get nodes -l node-health.aro-hcp.azure.com/status=wedged
kubectl get node <node-name> -o jsonpath='{.metadata.annotations}' | jq
```

The controller only labels and annotates, it does not cordon, drain or delete. The mitigation steps in this runbook are still manual. Fall back to the Kusto queries above if the label is not present, for example on a cluster that has not picked up the controller yet.
````

File: docs/ops/mitigate-swiftv2-wedged-node.md [hunk 9]
```diff
@@ -0,0 +1,375 @@
...
+
+This prevents new pods from being scheduled to the node while you verify and delete.
+
+### Step 3: Delete the VMSS instance
+
+```bash
+# Get the VMSS name and instance ID from the node's provider ID
+PROVIDER_ID=$(kubectl get node <node-name> -o jsonpath='{.spec.providerID}')
+echo "${PROVIDER_ID}"
+# Format: azure:///subscriptions/.../virtualMachineScaleSets/<vmss-name>/virtualMachines/<instance-id>
```
gmfrasca: Let's fully expand this format as it has useful metadata (such as node-resource-group), which is useful for the step immediately after

```suggestion
# Format: azure:///subscriptions/<subscription-id>/resourceGroups/<node-resource-group>/providers/Microsoft.Compute/virtualMachineScaleSets//virtualMachineScaleSets/<vmss-name>/virtualMachines/<instance-id>
```

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,375 @@
...
+### 6. Confirm router impact (if cluster-creates are timing out)
+
+```bash
+# Check if router replicas are landing on the wedged node
+kubectl get pods -A -o wide | grep router | grep <node-name>
+```
+
+## Prerequisites
+
+- JIT access to the affected management-cluster AKS (include this runbook link and ICM 832382845 in the justification)
```
gmfrasca, line 187: what level of access for the JIT?

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,375 @@
...
+  --instance-ids <instance-id>
+```
+
+Alternatively, if you only have `kubectl` access:
+
+```bash
+kubectl delete node <node-name>
+```
+
+> **Note**: Deleting the node object from Kubernetes does not delete the underlying VMSS instance. For a full cleanup, use `az vmss delete-instances`. The AKS cluster autoscaler or node-pool reconciler will provision a replacement.
```
hbhushan3: >Deleting the node object from Kubernetes does not delete 

We saw that deleting the node deletes the vmss instance on Tuesday.

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,375 @@
...
+
+### Step 2: Cordon the node
+
+```bash
+kubectl cordon <node-name>
+```
+
+This prevents new pods from being scheduled to the node while you verify and delete.
+
+### Step 3: Delete the VMSS instance
```
hbhushan3: Deleting the node, deletes the vmss instance.

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,375 @@
...
+
+## Mitigation Procedure
+
+### Step 1: (Optional) Collect ACN logs for evidence preservation
+
+Run the log-collection script **before** deleting the node, in case the Azure networking team (Cloudnet/RNC) needs on-node state for root-cause analysis.
+
+```bash
+# Target only the wedged node
+NODES_FILTER='<vmss-instance-id>' ./collect-swiftv2-node-logs.sh
```
hbhushan3: Step 1 invokes ./collect-swiftv2-node-logs.sh but it only exists inline in the appendix. Commit it under docs/ops/ or hack so it's runnable

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,377 @@
...
+If Kusto shows recent failures but kubectl events are empty, schedule a test pod directly on the suspect node to check whether the NIC is currently functional:
+
+```bash
+kubectl run nic-test --image=mcr.microsoft.com/cbl-mariner/base/core:2.0 \
+    --overrides='{"spec":{"nodeName":"<node-name>"}}' \
+    --command -- sleep 300
+kubectl get pod nic-test -w
+```
+
+- If the pod reaches `Running`: the NIC is not actively wedged right now. The fault is intermittent and will likely recur under scheduling pressure (e.g. during e2e runs). Proceed with the flap-vs-wedge query (Step 2) to decide whether to delete.
```
raelga: This is the one I would change.

The error we actually see is `SecondaryEndpointClient Error: route ip+net: no such network interface`. `SecondaryEndpointClient` is the delegated-NIC path in azure-vnet, so it only runs for pods that get an MTPNC. `kubectl run nic-test` in whatever namespace the operator happens to be in has no `PodNetworkInstance` and no delegated NIC, so it takes the ordinary endpoint path and never reaches the code that is broken. On the uksouth node I captured, every failing pod was in an `ocm-arohcpprod-*` namespace.

So a `Running` nic-test does not tell you the node is healthy, it tells you the primary CNI path works. Right now the step reads it as "the NIC is not actively wedged", and someone using that to rule the issue out will leave a genuinely wedged node in the pool.

I would keep the pod as a positive signal only (if it fails you are definitely wedged) and make the pass case explicitly inconclusive:

```suggestion
- If the pod reaches `Running`: this is inconclusive, not a clean bill of health. The probe pod has no delegated NIC, so it exercises the ordinary CNI path and not the `SecondaryEndpointClient` path that fails here. Decide from the flap-vs-wedge query (Step 2), not from this pod.
```

raelga (commented):
Read the whole thing again against the current head.

Everything I raised earlier is in, and the Kusto queries, the flap-vs-wedge rule, the durability caveat and the extracted script all look right to me now. I also checked the node-health appendix against the code: the label, the four annotations and the three metric names all match `mgmt-agent/pkg/controller/nodehealth/consts.go` and `metrics.go`, so that section is accurate. One thing you could add there is `nodehealth_wedged_nodes`, the fleet-wide count gauge, since that is the one you would alert on rather than the per-node series.

As far as I can tell all 16 open threads are addressed in this head, including the `kubectl delete node` behaviour, so they just need resolving.

Four things left. The first one is the only one I would want fixed before this merges, because it can send someone the wrong way.

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,377 @@
...
+
+## Symptoms
+
+All of these will be present on the affected node:
+
+| Signal | Detail |
+|---|---|
+| **Event** | `FailedCreatePodSandBox` at ~15s retry cadence, thousands of occurrences |
+| **Error messages** | `route ip+net: no such network interface`, `mtpnc is not ready`, `dhcp discover timed out` |
+| **Blast radius** | 100% of pods scheduled to the node fail; hosted-cluster router replicas are especially impacted → `KASLoadBalancerNotReachable` → cluster-create timeouts |
```
raelga: Same point as the test-pod step, and it is the reason that step looks safe. Going by the `SecondaryEndpointClient` signature and the pods I captured on the uksouth node, what fails is the pods that need a delegated NIC, which in practice is the hosted-control-plane workload. I would not claim 100% of pods, because that is exactly what makes a plain test pod look like a valid probe.

```suggestion
| **Blast radius** | Every pod needing a delegated NIC fails on that node, which in practice is the whole hosted-control-plane workload scheduled there; hosted-cluster router replicas are especially impacted → `KASLoadBalancerNotReachable` → cluster-create timeouts |
```

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,377 @@
...
+# Target only the wedged node (run from this directory: docs/ops/)
+NODES_FILTER='<vmss-instance-id>' ./collect-swiftv2-node-logs.sh
+```
+
+The script is read-only: it creates temporary busybox debug pods, pulls `/host/var/log/azure-vnet*` and `/host/var/run/azure-cns` state, tars them, and deletes only the debug pods it created.
+
+### Step 2: Cordon the node
+
+```bash
+kubectl cordon <node-name>
```
raelga, line 217: Worth a line here about what Step 3 then does to the pods still on the node.

Cordon only stops new scheduling. `kubectl delete node` makes the pod garbage collector force-delete everything bound to that node, and that path does not go through the Eviction API, so PDBs are not consulted and there is no graceful termination. On a mgmt cluster those are customer control-plane pods, and the ones that started before the wedge began are still serving fine.

Either put a drain between cordon and delete:

```bash
kubectl drain <node-name> --ignore-daemonsets --delete-emptydir-data --timeout=5m
```

or say explicitly that we skip it and why. I do not mind which one, but as written the reader cannot tell it was a decision.

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,377 @@
...
+
+## Mitigation Procedure
+
+### Step 1: (Optional) Collect ACN logs for evidence preservation
+
+Run the log-collection script **before** deleting the node, in case the Azure networking team (Cloudnet/RNC) needs on-node state for root-cause analysis. The script lives alongside this runbook at [`collect-swiftv2-node-logs.sh`](collect-swiftv2-node-logs.sh).
+
+```bash
+# Target only the wedged node (run from this directory: docs/ops/)
+NODES_FILTER='<vmss-instance-id>' ./collect-swiftv2-node-logs.sh
```
raelga: `NODES_FILTER` is used as a `grep -E` pattern against node names, and the script's own header shows `NODES_FILTER='vmss000004'`. An instance ID is a bare number, so as a regex it matches any node with that digit anywhere in its name and you collect from more nodes than you meant.

```suggestion
NODES_FILTER='<node-name>' ./collect-swiftv2-node-logs.sh
```

docs/ops/collect-swiftv2-node-logs.sh
```diff
@@ -0,0 +1,84 @@
...
+# Optional: set NODES_FILTER to a grep pattern to target specific node(s), e.g.
+#   NODES_FILTER='vmss000004' ./collect-swiftv2-node-logs.sh
+# Default (unset) collects from ALL nodes.
+
+set -uo pipefail
+set -x
+
+OUTPUT_DIR="$(mktemp -d)"
+TARBALL="node-logs-$(date +%Y%m%dT%H%M%S).tar.gz"
+IMAGE="busybox"
```
raelga: Two problems with a bare `busybox`. It floats, so it resolves to whatever `latest` happens to be that day, and it pulls from Docker Hub, where anonymous pulls are rate limited and are often blocked outright from a prod mgmt cluster. That turns evidence collection into a coin flip at the exact moment you need it to work.

The runbook already uses an MCR image for the test pod, so I would use the same one here and stay off Docker Hub. It has `tar`, which is all `kubectl cp` needs.

```suggestion
IMAGE="mcr.microsoft.com/cbl-mariner/base/core:2.0"
```

avollmer-redhat (commented):
Addressing remaining review comments from @raelga's Aug 5 review. Will push fixes for:
1. Blast radius wording (delegated NIC pods, not 100%)
2. Test pod `Running` result is inconclusive, not clean bill of health
3. Add explicit drain step between cordon and delete
4. NODES_FILTER should use node-name, not vmss-instance-id
5. Script image should be MCR, not Docker Hub busybox

Also resolving comments from @hbhushan3 (script already split out as `collect-swiftv2-node-logs.sh`) and @gmfrasca (providerID already expanded).

docs/ops/mitigate-swiftv2-wedged-node.md
```diff
@@ -0,0 +1,377 @@
...
+
+```bash
+# Check if router replicas are landing on the wedged node
+kubectl get pods -A -o wide | grep router | grep <node-name>
+```
+
+## Prerequisites
+
+- JIT access to the affected management-cluster AKS (include this runbook link and ICM 832382845 in the justification)
+- `kubectl` configured against the management AKS
```
avollmer-redhat, line 188: Done in commit b3493f27 — the providerID format line now reads:
```
azure:///subscriptions/<subscription-id>/resourceGroups/<node-resource-group>/providers/Microsoft.Compute/virtualMachineScaleSets/<vmss-name>/virtualMachines/<instance-id>
```
This shows all the metadata needed for the `az vmss delete-instances` step that follows.

hack/collect-swiftv2-node-logs.sh
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
```
avollmer-redhat, line 1: Done — the script was split out as a standalone file in commit b3493f27. The runbook Step 1 now references it as `[collect-swiftv2-node-logs.sh](collect-swiftv2-node-logs.sh)` with a `run from this directory: docs/ops/` note.

docs/ops/mitigate-swiftv2-wedged-node.md
avollmer-redhat: Good catch — will fix in next push. Changing to:
> Every pod needing a delegated NIC fails on that node, which in practice is the whole hosted-control-plane workload scheduled there; hosted-cluster router replicas are especially impacted → `KASLoadBalancerNotReachable` → cluster-create timeouts

docs/ops/mitigate-swiftv2-wedged-node.md
avollmer-redhat: Agreed — the probe pod has no delegated NIC so it exercises the ordinary CNI path, not the `SecondaryEndpointClient` path. Will update the Running case to say this is inconclusive rather than a clean bill of health.

docs/ops/mitigate-swiftv2-wedged-node.md
avollmer-redhat, line 217: Good call — will add an explicit `kubectl drain` step between cordon and delete, with a note that skipping drain means PDBs are not consulted and existing healthy pods get force-deleted.

docs/ops/mitigate-swiftv2-wedged-node.md
avollmer-redhat: Will fix — changing to `NODES_FILTER='<node-name>'` since the filter is used as a `grep -E` pattern against node names.

docs/ops/collect-swiftv2-node-logs.sh
avollmer-redhat: Agreed — switching to `mcr.microsoft.com/cbl-mariner/base/core:2.0` to match the test-pod image and avoid Docker Hub rate limits on prod clusters.

raelga:
Checked the new head, all five are in and they read well. The drain step in particular is better than what I suggested, the note about most pods already being stuck on a wedged node is a good addition.

One leftover from adding it: the decision rule still says the action is "cordon + delete" in two places, line 101 and the node-health criteria in the appendix, so it now understates the procedure by one step. Worth making both say cordon, drain, delete so someone skimming the rule does not skip the drain you just added.

Not blocking, happy either way.

roivaz:
@avollmer-redhat couple of things:
- Could we move the script out of `docs/`? `dev-infrastructure/scripts` would be a better place for it.
- Consider creating a follow up to add the relevant logs to our log ingestion so they are available in Kusto rather than having to JIT into the clusters to get them. Feel free to drop this suggestion if the effort is not worth for the frequency at which we are required to retrieve these logs.

avollmer-redhat (commented):
Self-review (cannot approve own PR):

Reviewed HEAD (244c6027). All review threads from raelga, gmfrasca, and hbhushan3 have been addressed:

- Kusto table names fixed (`kubernetesEvents` with lowercase columns)
- centralindia (AROSLSRE-1717/1723) added to occurrences
- Blast radius corrected ("every pod needing a delegated NIC" not "100%")
- Test pod `Running` result marked inconclusive (no delegated NIC = ordinary CNI path)
- Explicit drain step added between cordon and delete
- `NODES_FILTER` uses node-name not vmss-instance-id
- Script image switched from Docker Hub `busybox` to MCR `mcr.microsoft.com/cbl-mariner/base/core:2.0`
- Durability caveat added (fault migrates across instances)
- JIT Role Selection table added with 3 methods
- Script split out as standalone `collect-swiftv2-node-logs.sh`
- Node-health controller appendix updated (shipped, no longer hypothetical)

Non-blocking follow-up: consider adding the `nodehealth_wedged_nodes` gauge metric to the automated detection appendix (per Rael's last comment thread) as a future PR.

Ready for /lgtm from a reviewer.

gmfrasca (commented):
/lgtm

janboll (commented):
/lgtm
/approve

raelga (approved):
/unhold
/lgtm
/approve

PR #6375, opened 2026-08-03
feat: add admin API endpoint to set per-nodepool minimum version

ahitacat:
### What

- Add `MinimumVersion` field to `ServiceProviderNodePoolSpecVersion` — an SRE-controlled semver floor per node pool, nil when unset
- New admin API endpoint `POST .../nodepools/{nodepoolName}/minimumversion` to set or clear the minimum version
- Update `nodePoolVersionSyncer` to select `max(customerDesiredVersion, sreMinimumVersion)` as the target version before validation
- Handler tests use callback-based table pattern (`setup`/`validate`); controller tests consolidate SRE minimum version cases into the existing `_IntentFailed` table

### Why

SRE needs to enforce a version floor on individual node pools, ensuring they don't remain on versions below a required threshold. When set, the minimum version overrides the customer's desired version if it is higher; when cleared, the customer's version is used as before.

### Testing

- Unit tests for the admin handler (callback-based table-driven, covering validation errors, set, overwrite, and clear)
- Controller tests extended with SRE minimum version cases in the `_IntentFailed` table test

**Why is this WIP?**
Once we add the SRE selection of a minimumVersion, there will be nodepool updated. This nodepool will be represented in the RP/CosmosDB with a different version that they really have. We need to agree on a correct solution to reflect this reality.

**Usage example:**

Set a minimum version:
```bash
curl -X POST \
  "https://<admin-api>/admin/v1/hcp/subscriptions/<subscription-id>/resourcegroups/<resource-group>/providers/microsoft.redhatopenshift/hcpopenshiftclusters/<cluster-name>/nodepools/<nodepool-name>/minimumversion" \
  -H "Content-Type: application/json" \
  -H "X-Ms-Client-Principal-Name: user@example.com" \
  -d '{"minimumVersion": "4.21.0"}'
```

Response:
```json
{"minimumVersion": "4.21.0"}
```
### PR Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/admin/server/handlers/hcp/minimumnodepoolversion.go
+++ b/admin/server/handlers/hcp/minimumnodepoolversion.go
@@ -1,0 +1,104 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hcp
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/blang/semver/v4"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	"github.com/Azure/ARO-HCP/internal/database"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// HCPMinimumNodePoolVersionHandler sets ServiceProviderNodePoolSpecVersion
+// MinimumVersion on the per-nodepool ServiceProviderNodePool.
+// It intentionally writes only that one field — every other Spec/Status value
+// is left as-is — so SRE callers can adjust the version of a particular nodepool
+// without touching anything else on the document.
+type HCPMinimumNodePoolVersionHandler struct {
+	resourcesDBClient database.ResourcesDBClient
+}
+
+func NewHCPMinimumNodePoolVersionHandler(resourcesDBClient database.ResourcesDBClient) *HCPMinimumNodePoolVersionHandler {
+	return &HCPMinimumNodePoolVersionHandler{resourcesDBClient: resourcesDBClient}
+}
+
+// minimumNodePoolVersionRequest is the wire shape for the request body. The
+// MinimumVersion field is a pointer-string so callers can distinguish "set to value"
+// (non-nil) from "clear the version" (nil/absent). An explicit empty string is
+// rejected.
+type minimumNodePoolVersionRequest struct {
+	MinimumVersion *string `json:"minimumVersion"`
+}
+
+func (h *HCPMinimumNodePoolVersionHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) error {
+	clusterResourceID, err := utils.ResourceIDFromContext(request.Context())
+	if err != nil {
+		return arm.NewCloudError(http.StatusBadRequest, arm.CloudErrorCodeInvalidRequestContent, "", "invalid resource identifier in request")
+	}
+
+	nodepoolName := request.PathValue("nodepoolName")
+	if nodepoolName == "" {
+		return arm.NewCloudError(http.StatusBadRequest, arm.CloudErrorCodeInvalidRequestContent, "", "nodepoolName parameter is required")
+	}
+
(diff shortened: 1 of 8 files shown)


admin/server/handlers/hcp/minimumnodepoolversion.go
```diff
@@ -0,0 +1,104 @@
...
+		versionPtr = &version
+	}
+
+	existing, err := database.GetOrCreateServiceProviderNodePool(request.Context(), h.resourcesDBClient, resourceID)
+	if err != nil {
+		return fmt.Errorf("failed to get ServiceProviderNodePool: %w", err)
+	}
+
+	replacement := existing.DeepCopy()
+	replacement.Spec.NodePoolVersion.MinimumVersion = versionPtr
```
machi1990, line 95: We need to use this minimum version for validation in the frontend so that all versions update to this node pool are higher than the minimum version

machi1990, line 95: Is this handler missing a validation that minimum version x.y isn't higher than clusters x.y?

backend/pkg/controllers/nodepool/version/nodepool_version_controller.go
```diff
@@ -135,13 +138,19 @@ func (c *nodePoolVersionSyncer) NeedsWork(nodePool *api.HCPOpenShiftClusterNodeP
 		return true
 	}
 
+	if serviceProviderNodePool.Spec.NodePoolVersion.MinimumVersion != nil && serviceProviderNodePool.Spec.NodePoolVersion.MinimumVersion.GT(*serviceProviderNodePool.Spec.NodePoolVersion.DesiredVersion) {
```
machi1990, line 141: Why greater than? 
We allow downgrade too so I believe the check is != 
i.e if minimum version set and != SPNP desired, then work should be done

backend/pkg/controllers/nodepool/version/nodepool_version_controller.go
```diff
@@ -107,10 +107,13 @@ func NewNodePoolVersionController(
...
 // NeedsWork reports whether this controller has anything to do for the given
 // NodePool. The work this controller does is persist the customer's desired
-// version onto the ServiceProviderNodePool, which is needed when:
+// version or SRE desired version onto the ServiceProviderNodePool, which is
+// needed when:
 //   - the NodePool's customer-visible Properties.Version.ID is set, and
 //   - the ServiceProviderNodePool's Spec.NodePoolVersion.DesiredVersion does
-//     not already equal that value (otherwise nothing would change).
+//     not already equal that value (otherwise nothing would change). Or
+//   - the ServiceProviderNodePool's Spec.NodePoolVersion.MinimumVersion
```
machi1990, line 115: See https://github.com/Azure/ARO-HCP/pull/6375/changes#r3703650882

backend/pkg/controllers/nodepool/version/nodepool_version_controller.go
machi1990: Ah okay, I see, we only need to perform the work if we need to set the minimum version, otherwise there is no point if we are at a higher version.

Ignore my comment above

backend/pkg/controllers/nodepool/version/nodepool_version_controller.go
```diff
@@ -196,6 +205,14 @@ func (c *nodePoolVersionSyncer) SyncOnce(ctx context.Context, key controllerutil
 		return utils.TrackError(err)
 	}
 
+	targetVersion := customerDesiredVersion
+
+	sreDesiredVersion := cachedServiceProviderNodePool.Spec.NodePoolVersion.MinimumVersion
+
+	if sreDesiredVersion != nil && sreDesiredVersion.GT(customerDesiredVersion) {
+		targetVersion = *sreDesiredVersion
+	}
```
machi1990, line 214: When we use the minimum version, it's critical that we surface to the end use somehow, perhaps when we've the actual version reported, this could be our mechanism;
- until then I believe, the frontend validation should be enforced to reject any version updates that are lower than minimum SRE enforced versions; this is on top of the actual version validation we've there

backend/pkg/controllers/nodepool/version/nodepool_version_controller.go
machi1990: ignore

admin/server/handlers/hcp/minimumnodepoolversion.go
ahitacat: Makes sense, I have opened this question let's see what is the agreement.

ahitacat: That validation is performed in the upgradecontroller. If it is higher the desiredVersion will not be set to this version, logging an error. SRE can change the minimumVersion to the correct one. 

I considered that the validation that the upgradecontroller have is enough, this will be not used that much by SREs and I trust on SRE picking the correct version.

machi1990: The admin handler has everything for the correct implementation of the validation to be done rather than leaving it off to a human for manual verification

ahitacat:
Decided not to implement this https://github.com/openshift-online/aro-enhancements/pull/9

PR #6385, opened 2026-08-03
fix(hcp): add automatedRetry for region rollout ARM propagation-lag failures

raelga:
## Why

The westus3 stg region rollout ([AROSLSRE-1689](https://redhat.atlassian.net/browse/AROSLSRE-1689)) hit two transient failures in the same pipeline run, both caused by an ARM deploy running too soon after a dependency it needs was just created/registered. Neither was a real config or capacity problem, both went away on a plain retry, but we're about to bring up several more new regions and don't want every rollout to need a manual retry.

## What

- `kustoDataConnection`/`alertEventsDataConnection` in `region-pipeline.yaml` had no `automatedRetry` at all, so an Internal Server Error right after the geography pipeline creates the target Kusto cluster/database (a separate service group with no cross-pipeline `dependsOn`) failed the whole rollout. The observed failure was `error.code=BadInput` with that message text, already covered by the `InternalServerError`/`Internal server error`/`500 Internal Server Error` patterns without needing a separate `BadInput` matcher (which would be too broad and risk masking real config mistakes). Added `automatedRetry` to both steps ([AROSLSRE-1707](https://redhat.atlassian.net/browse/AROSLSRE-1707)).
- `regional/infra`'s `automatedRetry` only matched `"Internal error"`, missing the `InternalServerError`/`500 Internal Server Error` variants seen right after `rpRegistration` completes for a brand new subscription. Widened the error match list ([AROSLSRE-1708](https://redhat.atlassian.net/browse/AROSLSRE-1708)).

Not changing dependency ordering or adding cross-pipeline waits, just widening retry coverage to match the transient error signatures we actually saw.

[AROSLSRE-1706](https://redhat.atlassian.net/browse/AROSLSRE-1706)

### DIFF
--- a/dev-infrastructure/region-pipeline.yaml
+++ b/dev-infrastructure/region-pipeline.yaml
@@ -71,6 +71,20 @@ resourceGroups:
     dependsOn:
     - resourceGroup: regional
       step: rpRegistration
+    # On a first-time region rollout this deploys shortly after the geography
+    # pipeline creates the Kusto cluster/database it targets (they are separate
+    # service groups with no cross-pipeline dependsOn), so it can race ahead of
+    # ARM RP propagation and fail with a transient Internal Server Error (the
+    # observed failure was error.code=BadInput with that message text, already
+    # covered by the patterns below without a separate BadInput matcher).
+    # See AROSLSRE-1707.
+    automatedRetry:
+      errorContainsAny:
+      - "InternalServerError"
+      - "Internal server error"
+      - "500 Internal Server Error"
+      maximumRetryCount: 3
+      durationBetweenRetries: 1m
     variables:
     - name: auditLogsEventHubId
       input:
@@ -85,6 +99,14 @@ resourceGroups:
     dependsOn:
     - resourceGroup: regional
       step: rpRegistration
+    # Same transient-failure risk as kustoDataConnection above. See AROSLSRE-1707.
+    automatedRetry:
+      errorContainsAny:
+      - "InternalServerError"
+      - "Internal server error"
+      - "500 Internal Server Error"
+      maximumRetryCount: 3
+      durationBetweenRetries: 1m
     variables:
     - name: alertEventsEventHubId
       input:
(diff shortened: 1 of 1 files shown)


ashishmax31 (approved):
/lgtm

raelga:
This change does not affect to E2E.

/retest

raelga:
/test e2e-parallel

Retesting. The failure is the known nodepool osDisk flake, not something this PR can cause.

Job `pull-ci-Azure-ARO-HCP-main-e2e-parallel/2084725847880634368` failed 1 of 85 specs:

```
Customer should be able to create an HCP cluster and custom node pool osDisk size
ERROR CODE: InternalServerError
```

That is the exact spec tracked by [AROSLSRE-1721](https://redhat.atlassian.net/browse/AROSLSRE-1721), and it is the same `InternalServerError` family currently hitting the fleet: at the time of writing, e2e-parallel is also red on [#6414](https://github.com/Azure/ARO-HCP/pull/6414), [#6408](https://github.com/Azure/ARO-HCP/pull/6408) and [#6407](https://github.com/Azure/ARO-HCP/pull/6407). [#6407](https://github.com/Azure/ARO-HCP/pull/6407) is the useful control there, it only edits `dev-infrastructure/openshift-ci/Dockerfile` and still fails the same way.

This PR only changes `dev-infrastructure/region-pipeline.yaml`, an EV2 rollout config that is not read by the e2e job at all, so there is no path from this diff to a guest cluster creation error.

raelga:
/test e2e-parallel

Retesting. This is a fleet-wide CI episode tonight, not something these diffs cause.

Across the open PRs right now e2e-parallel is `pass=4 fail=6` (#6419, #6416, #6414, #6409, #6408, #6407 all red). Every failure carries the same cluster-create signature:

```
ERROR CODE: InternalServerError
[clusterServiceClusterStatus] <no_message>; [hypershiftHostedCluster] hosted cluster has no installed version
```

The victim specs differ per run, which is what a shared-infra problem looks like rather than a real regression in any one diff. The clearest control is [#6385](https://github.com/Azure/ARO-HCP/pull/6385), which only edits `dev-infrastructure/region-pipeline.yaml`, an EV2 rollout config the e2e job never reads, and it fails with the identical signature.

It has also been getting worse through the evening, from 1 of 85 specs a few hours ago to 3 and 4 of 85 now. I am raising this with the team separately rather than relying on retests alone.

PR #6390, opened 2026-08-03
mergable admin creds change with fallback to old version

deads2k:
like https://github.com/Azure/ARO-HCP/pull/6344 with a live fallback that could mask issues with our endpoint.  But we'll be merging this one.


kube-applier/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_kube_applier.yaml
```diff
@@ -78,6 +78,18 @@ rules:
   - update
   - patch
   - delete
+- apiGroups:
+  - certificates.hypershift.openshift.io
+  resources:
+  - certificatesigningrequestapprovals
```
miguelsorianod, line 84: Are there other permissions needed around revocation?

frontend/asynchronous-operations.md
```diff
@@ -179,7 +179,7 @@ The "payload" section of an asynchronous operation document includes all the inf
 
 <a name="operation-document-request-field"></a>
 
-3. The `request` field captures the nature of the operation. Valid values are "Create", "Update", "Delete", "RequestCredential", and "RevokeCredentials".
+3. The `request` field captures the nature of the operation. Valid values are "Create", "Update", "Delete", "SystemAdminCredentialRequest", and "SystemAdminCredentialRevocation".
```
miguelsorianod: Is this correct  for the `request` field?  Internally we do have these new values but this seems to be external?

frontend/pkg/frontend/frontend.go
```diff
@@ -335,6 +344,30 @@ func (f *Frontend) ArmResourceActionRequestAdminCredential(writer http.ResponseW
 		return utils.TrackError(err)
 	}
 
+	var certificateRequest string
+	if apiVersion.GE(api.APIVersionV20260901Preview) {
```
miguelsorianod: Here it would help readers explain that since this version we use the new API approach where we receive the CSR, and in that case we store it in the Cosmos operation. Otherwise we don't. And that's what allows us to differentiate at the operation level whether it's using the new admincredential api or the old one

frontend/pkg/frontend/frontend.go
```diff
@@ -335,6 +344,30 @@ func (f *Frontend) ArmResourceActionRequestAdminCredential(writer http.ResponseW
...
+		if err != nil {
+			return utils.TrackError(err)
+		}
+
+		var errs field.ErrorList
+		if credentialRequest == nil {
+			errs = append(errs, field.Required(field.NewPath("certificateRequest"), ""))
+		} else if credentialRequest.CertificateRequest == "" {
+			errs = append(errs, field.Required(field.NewPath("certificateRequest"), ""))
+		}
```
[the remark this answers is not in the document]
miguelsorianod: This applies. The attribute name we expose at the ARM API level is `certificateSigningRequest` so this should be changed to reflect that.

In our internal api type we serializing it in Cosmos as `certificateRequest` (internal/api/types_admincredential.go). Would it be possible to change that so we also rename the Go attribute name as well as its json serialization to CertificateSigningRequest (json certificateSigningRequest)?

at that point also renaming the local variable here from certificateRequest to certificateSigningRequest too.

frontend/pkg/frontend/frontend.go
```diff
@@ -335,6 +344,30 @@ func (f *Frontend) ArmResourceActionRequestAdminCredential(writer http.ResponseW
...
 	}
 
+	var certificateRequest string
+	if apiVersion.GE(api.APIVersionV20260901Preview) {
+		body, err := BodyFromContext(ctx)
+		if err != nil {
+			return utils.TrackError(err)
+		}
+
+		credentialRequest, err := versionedInterface.UnmarshalHCPOpenShiftClusterAdminCredentialRequest(body)
```
miguelsorianod, line 361: Is there some reason we don't have AdminCredentialRequest as a `VersionedCreatableResource`  similar to other resources?

frontend/pkg/frontend/frontend.go
```diff
@@ -1004,6 +1042,16 @@ func (f *Frontend) OperationResult(writer http.ResponseWriter, request *http.Req
 	var responseBody []byte
 
 	switch {
+	case operation.SystemAdminCredentialRequest != nil:
```
miguelsorianod, line 1059: A comment might help the reader here where we determine we use the new admincredential api if the operation has it set and otherwise we fallback to checking the breakglass kind for operation internalid kind

frontend/pkg/frontend/frontend.go
```diff
@@ -1056,6 +1104,55 @@ func (f *Frontend) OperationResult(writer http.ResponseWriter, request *http.Req
 	return nil
 }
 
+// assembleAdminCredentialFromCosmos looks up the SystemAdminCredentialRequest Cosmos
+// document pointed to by Operation.InternalID and assembles a kubeconfig from
```
miguelsorianod: It mentions "pointed to by Operation.InternalID", but we are not setting that. This code comment should be updated to reflect the reality of the method

frontend/pkg/frontend/frontend.go
```diff
@@ -1075,6 +1123,55 @@ func (f *Frontend) OperationResult(writer http.ResponseWriter, request *http.Req
...
+// its signed certificate, private key, and the serving CA bundle from the
+// ServiceProviderCluster.
```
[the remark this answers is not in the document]
miguelsorianod: This applies. Similarly to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722257940

internal/database/cosmosstoragetesting/corecosmosstoragetesting/mock_init.go
```diff
@@ -193,10 +199,56 @@ func (m *MockResourcesDBClient) addController(ctx context.Context, controller *a
 		controllerCRUD := m.HCPClusters(resourceID.SubscriptionID, resourceID.ResourceGroupName).ExternalAuth(clusterName).Controllers(externalAuthName)
 		_, err := controllerCRUD.Create(ctx, controller, nil)
 		return err
+	case armhelpers.ResourceTypeEqual(parentType, api.SystemAdminCredentialRequestResourceType):
+		if resourceID.Parent.Parent == nil {
+			return fmt.Errorf("credential request controller is missing grandparent cluster ID")
```
miguelsorianod: nitpick: credential request -> system admin credential request

internal/database/cosmosstoragetesting/corecosmosstoragetesting/mock_init.go
```diff
@@ -193,10 +199,56 @@ func (m *MockResourcesDBClient) addController(ctx context.Context, controller *a
...
+		}
+		clusterName := resourceID.Parent.Parent.Name
+		credentialName := resourceID.Parent.Name
+		controllerCRUD := m.HCPClusters(resourceID.SubscriptionID, resourceID.ResourceGroupName).SystemAdminCredentialRequests(clusterName).Controllers(credentialName)
+		_, err := controllerCRUD.Create(ctx, controller, nil)
+		return err
+	case armhelpers.ResourceTypeEqual(parentType, api.SystemAdminCredentialRevocationResourceType):
+		if resourceID.Parent.Parent == nil {
+			return fmt.Errorf("credential revocation controller is missing grandparent cluster ID")
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722406790

internal/database/cosmosstoragetesting/corecosmosstoragetesting/mock_init.go
```diff
@@ -193,10 +199,56 @@ func (m *MockResourcesDBClient) addController(ctx context.Context, controller *a
...
+		_, err := controllerCRUD.Create(ctx, controller, nil)
+		return err
 	}
 	return fmt.Errorf("unsupported parent resource type: %s", parentType)
 }
 
+func (m *MockResourcesDBClient) addSystemAdminCredentialRequest(ctx context.Context, cred *api.SystemAdminCredentialRequest) error {
+	resourceID := cred.GetResourceID()
+	if resourceID == nil {
+		return fmt.Errorf("credential request is missing resource ID")
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722406790

internal/database/cosmosstoragetesting/corecosmosstoragetesting/mock_init.go
```diff
@@ -193,10 +199,56 @@ func (m *MockResourcesDBClient) addController(ctx context.Context, controller *a
...
+	}
+	if resourceID.Parent == nil {
+		return fmt.Errorf("credential request is missing parent cluster ID")
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722406790

internal/database/cosmosstoragetesting/corecosmosstoragetesting/mock_init.go
```diff
@@ -193,10 +199,56 @@ func (m *MockResourcesDBClient) addController(ctx context.Context, controller *a
...
+	clusterName := resourceID.Parent.Name
+	credCRUD := m.HCPClusters(resourceID.SubscriptionID, resourceID.ResourceGroupName).SystemAdminCredentialRequests(clusterName)
+	_, err := credCRUD.Create(ctx, cred, nil)
+	return err
+}
+
+func (m *MockResourcesDBClient) addSystemAdminCredentialRevocation(ctx context.Context, revocation *api.SystemAdminCredentialRevocation) error {
+	resourceID := revocation.GetResourceID()
+	if resourceID == nil {
+		return fmt.Errorf("credential revocation is missing resource ID")
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722406790

internal/database/cosmosstoragetesting/corecosmosstoragetesting/mock_init.go
```diff
@@ -193,10 +199,56 @@ func (m *MockResourcesDBClient) addController(ctx context.Context, controller *a
...
+	}
+	if resourceID.Parent == nil {
+		return fmt.Errorf("credential revocation is missing parent cluster ID")
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722406790

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+// JSON-marshal the object without depending on sigs.k8s.io/controller-runtime.
+type KubeObject interface {
+	metav1.Object
+	runtime.Object
+}
+
+const (
+	// ownerAnnotationKey is the annotation applied to every k8s object we land
+	// on a management cluster via ApplyDesire.
+	ownerAnnotationKey = "aro-hcp.openshift.io/owner"
```
miguelsorianod, line 42: Side question: is aro-hcp.openshift.io the domain we can use in general for K8s labels and annotations?

miguelsorianod: >  is the annotation applied to every k8s object we land on a management cluster via ApplyDesire.

will we want then to promote it at some point at the kube-applier package level?

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// defaultExpirationSeconds is the CSR expiration (approximately 24 hours).
+	defaultExpirationSeconds = int32(86400)
+
+	// customerBreakGlassSignerSuffix is the HyperShift signer name suffix.
+	// This is HyperShift's contract, not ours to rename.
+	customerBreakGlassSignerSuffix = "customer-break-glass"
```
miguelsorianod, line 50: Are those specific values required by Hypershift? I was trying to understand this and if that's the case it would be great to link to some kind of reference (doc, code, ...)

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+// (e.g. "ocm-<env>-<csClusterID>").
+// csrPEM is the PEM-encoded PKCS#10 certificate request provided by the caller.
+func BuildCSR(owner *azcorearm.ResourceID, credName, hcpNamespace string, csrPEM []byte) *certificatesv1.CertificateSigningRequest {
+	requireOwner(owner)
+
+	signerName := fmt.Sprintf("hypershift.openshift.io/%s.%s", hcpNamespace, customerBreakGlassSignerSuffix)
+
+	return &certificatesv1.CertificateSigningRequest{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "certificates.k8s.io/v1",
```
miguelsorianod: nitpick: certificatesv1.SchemeGroupVersion.String() is available

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+			Kind:       "CertificateSigningRequest",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        fmt.Sprintf("system-admin-credential-%s", credName),
```
miguelsorianod, line 86: Are there K8s name length restrictions that apply at the K8s level and also at specifically CertificateSigningRequest level? do we want to enforce some error/panic checking here?

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+		ownerAnnotationKey: strings.ToLower(owner.String()),
+	}
+}
+
+// BuildCSR builds a CertificateSigningRequest for a system admin credential.
+// The hcpNamespace is the HyperShift HCP namespace on the management cluster
+// (e.g. "ocm-<env>-<csClusterID>").
+// csrPEM is the PEM-encoded PKCS#10 certificate request provided by the caller.
+func BuildCSR(owner *azcorearm.ResourceID, credName, hcpNamespace string, csrPEM []byte) *certificatesv1.CertificateSigningRequest {
+	requireOwner(owner)
```
miguelsorianod, line 76: Side question: how do we decide what is a panic and what's an error? sometimes it's not clear what's the best approach. The criteria I often follow is: if it shouldn't occur then panicking is a possibility, but that counterbalances with not wanting to crash the process if that unexpectedly occurs so that plays against it at the same time

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+// BuildCSRApproval builds a CertificateSigningRequestApproval for a system admin credential.
+func BuildCSRApproval(owner *azcorearm.ResourceID, credName, hcpNamespace string) *certificatesv1alpha1.CertificateSigningRequestApproval {
+	requireOwner(owner)
+	return &certificatesv1alpha1.CertificateSigningRequestApproval{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "certificates.hypershift.openshift.io/v1alpha1",
+			Kind:       "CertificateSigningRequestApproval",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        fmt.Sprintf("system-admin-credential-%s", credName),
```
miguelsorianod, line 110: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722672083

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+		},
+	}
+}
+
+// BuildCSRApproval builds a CertificateSigningRequestApproval for a system admin credential.
+func BuildCSRApproval(owner *azcorearm.ResourceID, credName, hcpNamespace string) *certificatesv1alpha1.CertificateSigningRequestApproval {
+	requireOwner(owner)
+	return &certificatesv1alpha1.CertificateSigningRequestApproval{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "certificates.hypershift.openshift.io/v1alpha1",
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722534251

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+// BuildRevocationRequest builds a CertificateRevocationRequest that revokes all
+// customer-break-glass certificates for the cluster.
+func BuildRevocationRequest(owner *azcorearm.ResourceID, revokeOpSuffix, hcpNamespace string) *certificatesv1alpha1.CertificateRevocationRequest {
+	requireOwner(owner)
+	return &certificatesv1alpha1.CertificateRevocationRequest{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "certificates.hypershift.openshift.io/v1alpha1",
```
miguelsorianod: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722695860

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+			Kind:       "CertificateRevocationRequest",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        fmt.Sprintf("system-admin-credential-revocation-%s", revokeOpSuffix),
```
miguelsorianod, line 127: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722689542

internal/systemadmincredential/builders.go
```diff
@@ -0,0 +1,134 @@
...
+			Namespace:   hcpNamespace,
```
miguelsorianod, line 128: Similar to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3722689542 but for namespace

internal/systemadmincredential/keypair.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+const (
+	// rsaKeyBits is the RSA key size used for system admin credentials.
+	rsaKeyBits = 4096
+)
+
+// GenerateKeypair generates an RSA keypair and returns PEM-encoded public and
+// private key bytes.
+func GenerateKeypair() (publicPEM, privatePEM []byte, err error) {
```
miguelsorianod: In general we don't use named returns. They also make the code sometimes hard to follow and can cause subtle bugs. In this case because the function is short it's not very problematic but could we change this to not rely on them and document instead?

test/util/framework/helpers_v20260901preview.go
```diff
@@ -758,3 +765,92 @@ func (tc *perItOrDescribeTestContext) get20260901ClientFactoryUnlocked(ctx conte
...
+	adminCredentialRequestPoller, err := hcpClient.BeginRequestAdminCredential(
+		ctx,
+		resourceGroupName,
+		hcpClusterName,
+		hcpsdk20260901preview.HcpOpenShiftClusterAdminCredentialRequest{
+			CertificateSigningRequest: to.Ptr(string(csrPEM)),
+		},
+		nil,
+	)
+	if err != nil {
```
miguelsorianod, line 811: How are we differentiating between an error because of something went wrong with the new endpoint vs an error because it's not implemented? this seems like it might fallback silently in some cases where it shouldn't

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -151,9 +151,9 @@ var _ = Describe("Authorized CIDRs", func() {
 				GinkgoWriter.Printf("Connection from unauthorized IP address failed as expected on error: %v\n", err)
```
miguelsorianod, line 152: 👍 to move to the new method. I see the old method is still being used in other files though:
```
test/e2e/admin_credential_lifecycle.go:291:			newAdminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
test/e2e/cluster_autoscaling.go:142:			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
test/e2e/cluster_create_private_kas.go:131:			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
test/e2e/control_plane_minor_upgrade.go:186:			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
test/e2e/oidc_issuer_workload_identity.go:233:			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
test/e2e/region_upgrade.go:121:			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
test/e2e/cluster_create_feature_aggregation.go:209:			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
```

Do we want to move them to the new approach too?

I understand we should still keep at least one test with the old approach

test-integration/backend/launch/datadump_test.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/Azure/ARO-HCP/internal/utils"
+	"github.com/Azure/ARO-HCP/test-integration/utils/databasemutationhelpers"
+	"github.com/Azure/ARO-HCP/test-integration/utils/integrationutils"
+)
+
+func TestDataDumpControllerWithClusterAndCredentialRequest(t *testing.T) {
+	defer integrationutils.VerifyNoNewGoLeaks(t)
+
+	integrationutils.WithAndWithoutCosmos(t, func(t *testing.T, withMock bool) {
+		t.Skip("slow test")
```
miguelsorianod, line 43: If we always skip it what's the purpose? for someone to run it locally when they want to exercise and see the log output?

test-integration/backend/launch/datadump_test.go
```diff
@@ -0,0 +1,212 @@
...
+			backendInformers.RunWithContext(ctx)
+			backendErrCh <- nil
+		}()
+		go func() {
+			dataDumpController.Run(ctx, 20)
+			backendErrCh <- nil
+		}()
+
+		clusterResourceID := api.Must(azcorearm.ParseResourceID("/subscriptions/32350638-2403-4bc9-a36e-4922c8c99b52/resourceGroups/resourceGroupName/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/basic"))
+		frontendClientAccessor := databasemutationhelpers.NewVersionedHTTPTestAccessor(testInfo.FrontendURL, "2024-06-10-preview")
```
miguelsorianod, line 99: Is the usage of the oldest version intentional?

backend/pkg/utils/controllerutils/system_admin_credential_revocation_watching_controller.go
```diff
@@ -0,0 +1,111 @@
...
+	syncer SystemAdminCredentialRevocationSyncer,
+) Controller {
+
+	systemAdminCredentialRevocationSyncer := &systemAdminCredentialRevocationWatchingController{
+		name:              name,
+		resourcesDBClient: resourcesDBClient,
+		syncer:            syncer,
+	}
+	controller := newGenericWatchingController(name, api.SystemAdminCredentialRevocationResourceType, systemAdminCredentialRevocationSyncer)
+
```
miguelsorianod, line 67: Is there some reason here we don't want to trigger on readdesire?

backend/pkg/kubeapplierhelpers/system_admin_credentials.go
```diff
@@ -0,0 +1,117 @@
...
+//   - the ReadDesire has not been created yet (NotFound),
+//   - the ReadDesire exists but the kube-applier has not yet observed
+//     the target (Status.KubeContent is nil or empty).
+//
+// Returns a non-nil error only for hard failures: a non-NotFound lister
+// error, or unmarshal failure.
+func GetCachedCertificateRevocationRequestForSystemAdminCredentialRevocation(
+	ctx context.Context,
+	readDesireLister dblisters.ReadDesireLister,
+	subscriptionName, resourceGroupName, clusterName, revocationName, revokeOpSuffix string,
```
miguelsorianod: Is `revokeOpSuffix` intentionally not being used?

backend/pkg/kubeapplierhelpers/desire_helpers.go
```diff
@@ -0,0 +1,285 @@
...
+		},
+	}
+
+	// Consult the lister first: if an ApplyDesire already exists with the desired
+	// content there is nothing to do, and we skip the Cosmos write.
+	existing, err := parent.getApplyDesire(ctx, applyDesireLister, subscriptionID, resourceGroupName, hcpClusterName, desireName)
+	if err != nil && !cosmosstorageutils.IsNotFoundError(err) {
+		return utils.TrackError(fmt.Errorf("get ApplyDesire %s from lister: %w", desireName, err))
+	}
+	if existing != nil && applyDesireSpecEqual(existing.Spec, desire.Spec) {
```
miguelsorianod: Here we are comparing if the spec is equal. If it's equal we return. If it's not equal we continue and attempt to create, but if it exists we don't update. Is that intentional? if that's the case, I am unsure about what's the value of this comparison, as if they differ and it exists the Create call will always fail

backend/pkg/kubeapplierhelpers/desire_helpers.go
```diff
@@ -0,0 +1,285 @@
...
+		return nil
+	}
+
+	if _, err := crud.Create(ctx, desire, nil); err != nil && !cosmosstorageutils.IsConflictError(err) {
+		return utils.TrackError(fmt.Errorf("create ApplyDesire %s: %w", desireName, err))
+	}
+	logger.Info("created ApplyDesire", "desire", desireName, "targetResource", target.Resource, "targetName", target.Name)
```
miguelsorianod: At the log level this won't allow us to differentiate between "this has really been created now in the DB" and "it already existed so all good".
Can we log different text so we can determine when it's one situation or the other?

backend/pkg/kubeapplierhelpers/desire_helpers.go
```diff
@@ -0,0 +1,285 @@
...
+		},
+	}
+
+	// Consult the lister first: if a ReadDesire already exists with the desired
+	// content there is nothing to do, and we skip the Cosmos write.
+	existing, err := parent.getReadDesire(ctx, readDesireLister, subscriptionID, resourceGroupName, hcpClusterName, desireName)
+	if err != nil && !cosmosstorageutils.IsNotFoundError(err) {
+		return utils.TrackError(fmt.Errorf("get ReadDesire %s from lister: %w", desireName, err))
+	}
+	if existing != nil && readDesireSpecEqual(existing.Spec, desire.Spec) {
```
miguelsorianod: Similar comment to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3734748271

backend/pkg/kubeapplierhelpers/desire_helpers.go
```diff
@@ -0,0 +1,285 @@
...
+		return nil
+	}
+
+	if _, err := crud.Create(ctx, desire, nil); err != nil && !cosmosstorageutils.IsConflictError(err) {
+		return utils.TrackError(fmt.Errorf("create ReadDesire %s: %w", desireName, err))
+	}
+	logger.Info("created ReadDesire", "desire", desireName, "targetResource", target.Resource, "targetName", target.Name)
```
miguelsorianod: Similar comment to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3734753580

backend/pkg/kubeapplierhelpers/desire_helpers.go
```diff
@@ -0,0 +1,285 @@
...
+func readDesireSpecEqual(existing, desired kubeapplier.ReadDesireSpec) bool {
+	return controllerutil.ResourceIDsEqual(existing.ManagementCluster, desired.ManagementCluster) &&
+		existing.TargetItem == desired.TargetItem
+}
+
+// targetRefForKubeObject builds a kube-applier ResourceReference for a typed
+// Kubernetes object by deriving the resource name from its kind.
+func TargetRefForKubeObject(obj systemadmincredential.KubeObject) kubeapplier.ResourceReference {
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	resource := strings.ToLower(gvk.Kind) + "s"
```
miguelsorianod: Out of curiosity: does this hold true for all of them?

backend/pkg/kubeapplierhelpers/desire_deletion.go
```diff
@@ -0,0 +1,144 @@
...
+// waiting for — one entry per desire that has not finished deleting yet. An empty
+// slice means teardown is complete. Callers join the reasons into a single wait
+// message. This helper is shared by the cluster-deletion-cleanup,
+// post-issuance-cleanup, and revocation-deletion controllers.
+func DeleteDesires(
+	ctx context.Context,
+	kubeApplierClient kubeappliercosmosstorage.KubeApplierDBClient,
+	parent DesireParent,
+	subscriptionID, resourceGroupName, hcpClusterName string,
+	matches func(desireName string) bool,
```
miguelsorianod: Why is this a predicate and not the desireName? is it because we might want to match in a different way to pure equality, like for example prefix or something?

backend/pkg/kubeapplierhelpers/desire_deletion.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, utils.TrackError(fmt.Errorf("list ApplyDesires: %w", err))
+	}
+	for _, desire := range applyIter.Items(ctx) {
+		desireName := desire.ResourceID.Name
+		if !matches(desireName) {
+			continue
+		}
+		removed, err := removeApplyDesireForDeletion(ctx, desireName, applyCRUD)
+		if err != nil {
+			return nil, err
```
miguelsorianod: Do we want to return here or accumulate errors?

backend/pkg/kubeapplierhelpers/desire_deletion.go
```diff
@@ -0,0 +1,144 @@
...
+		if !removed {
+			waitingFor = append(waitingFor, fmt.Sprintf("ApplyDesire %q", desireName))
+		}
+	}
+	if err := applyIter.GetError(); err != nil {
+		return nil, utils.TrackError(fmt.Errorf("iterate ApplyDesires: %w", err))
+	}
+
+	// Step 2: delete each matching ReadDesire directly.
+	readIter, err := readCRUD.List(ctx, nil)
```
miguelsorianod, line 84: Is it intentional that we delete the readdesires even when the applydesires haven't been deleted yet?

backend/pkg/kubeapplierhelpers/desire_deletion.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return nil, utils.TrackError(fmt.Errorf("list ReadDesires: %w", err))
+	}
+	for _, desire := range readIter.Items(ctx) {
+		desireName := desire.ResourceID.Name
+		if !matches(desireName) {
+			continue
+		}
+		if err := readCRUD.Delete(ctx, strings.ToLower(desireName)); err != nil && !cosmosstorageutils.IsNotFoundError(err) {
+			return nil, utils.TrackError(fmt.Errorf("delete ReadDesire %s: %w", desireName, err))
```
miguelsorianod: Similar comment to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3734868379

backend/pkg/kubeapplierhelpers/desire_deletion.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	// If the desire is still a ServerSideApply, flip it to a Delete so the
+	// kube-applier tears down the applied object. TargetItem already names what
+	// to delete; the ServerSideApply payload is cleared.
+	if applyDesire.Spec.Type != kubeapplier.ApplyDesireTypeDelete {
+		applyDesire.Spec.Type = kubeapplier.ApplyDesireTypeDelete
+		applyDesire.Spec.ServerSideApply = nil
+		// clearing conditions so that we can be certain that a success means we deleted.
+		// TODO, different conditions.
```
miguelsorianod, line 130: what does this mean?

backend/pkg/kubeapplierhelpers/desire_deletion.go
```diff
@@ -0,0 +1,144 @@
...
+		applyDesire.Status.Conditions = nil
+		if _, err := applyCRUD.Replace(ctx, applyDesire, nil); err != nil && !cosmosstorageutils.IsNotFoundError(err) {
```
miguelsorianod, line 132: That means that if the applydesire is changed from apply to delete we will always have at least two calls needed to DeleteDesires right?

backend/pkg/kubeapplierhelpers/desire_deletion.go
```diff
@@ -0,0 +1,144 @@
...
+			return false, utils.TrackError(fmt.Errorf("convert ApplyDesire %s to Delete: %w", desireName, err))
+		}
+		return false, nil
+	}
+
+	// The desire is a Delete — remove the document once the delete has succeeded.
+	for _, cond := range applyDesire.Status.Conditions {
+		if cond.Type == kubeapplier.ConditionTypeSuccessful && cond.Status == "True" {
```
miguelsorianod: How do we know that successful == true means it's been deleted? is it because we unset the conditions above? because otherwise it could be because of it being applied instead of deleted

backend/pkg/controllers/cluster/credentialrequest/helpers.go
```diff
@@ -0,0 +1,28 @@
...
+
+package credentialrequest
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+
+	"github.com/Azure/ARO-HCP/internal/api"
+)
+
+func IsCredentialRequestPending(cred *api.SystemAdminCredentialRequest) bool {
```
miguelsorianod: What does pending mean in this context? is it the lack of the specified conditions being true? is there some reason we don't have a condition to explicitly signal that?

Is ContentDeleted condition type intentionally omitted?

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
```diff
@@ -0,0 +1,235 @@
...
+	if cluster.ServiceProviderProperties.ClusterServiceID == nil {
+		return false
+	}
+	return credentialrequest.IsCredentialRequestPending(cred)
+}
+
+func (c *desiresCreator) SyncOnce(ctx context.Context, key controllerutils.SystemAdminCredentialRequestKey) error {
+	logger := utils.LoggerFromContext(ctx)
+
+	existingCluster, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
```
miguelsorianod: Is there some reason we use Cosmos client instead of lister for the cluster? is it because getting outdated information could trigger some undesired action that we want to avoid?

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
```diff
@@ -0,0 +1,235 @@
...
+	clusterResourceID := key.GetClusterResourceID()
+
+	if err := c.ensureDesires(ctx, key, cred, key.CredentialName, controlPlaneNamespace, clusterResourceID, mcResourceID, kubeApplierClient); err != nil {
+		return err
+	}
+
+	logger.Info("ensured desires for credential", "credential", key.CredentialName)
+	return nil
+}
+
```
miguelsorianod, line 168: Side question: With the new directory organization where creation and deletion is separate, how would we introduce controllers that deal with creation and deletion in the same place, which we might want for some controllers? unifying those can be desirable in some cases and reduces the chances of racing between create and delete

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("get ApplyDesire CRUD: %w", err))
+	}
+	readCRUD, err := kubeApplierClient.ReadDesiresForSystemAdminCredentialRequest(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, credName)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("get ReadDesire CRUD: %w", err))
+	}
+
+	// 1. CSR ApplyDesire
+	csrDesireName := "systemadmincredentialcsr"
```
miguelsorianod, line 190: Similar to what we have for ReadDesireNameForSystemAdminCredentialRequestCSR. Would it be worth having ApplyDesireNameForSystemAdminCredentialRequestCSR in the kubeapplier package?

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
```diff
@@ -0,0 +1,235 @@
...
+	csrObj := systemadmincredential.BuildCSR(owner, credName, controlPlaneNamespace, []byte(cred.Spec.CertificateSigningRequestPEM))
+	if err := kubeapplierhelpers.EnsureApplyDesire(ctx, applyCRUD, c.applyDesireLister, parent,
+		key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName,
+		csrDesireName, mcResourceID, csrTarget(csrObj), csrObj); err != nil {
+		return err
+	}
+
+	// 2. CSRApproval ApplyDesire
+	csrApprovalDesireName := "systemadmincredentialcsrapproval"
```
miguelsorianod, line 199: Similar comment to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3735144222

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
```diff
@@ -0,0 +1,235 @@
...
+
+// NewDesiresCreatorController returns a CredentialRequestWatchingController that
+// creates the per-credential ApplyDesires (CSR, CSRApproval, 2 RBAC bundles) and
+// ReadDesire (CSR) for individual SystemAdminCredentialRequest documents that are
+// pending.
+//
+// The controller also fires on ReadDesire changes and consults the ApplyDesire /
+// ReadDesire listers before writing so it skips the create entirely when a desire
+// already exists with the desired content.
+func NewDesiresCreatorController(
```
miguelsorianod, line 60: Do I understand correctly that this controller only takes care of firing the creation of the resources, but it doesn't make sure that the content actually ends up with the desired state on the management cluster side?

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
```diff
@@ -0,0 +1,235 @@
...
+		return err
+	}
+
+	// 3. CSR ReadDesire
+	csrReadDesireName := kubeapplierhelpers.ReadDesireNameForSystemAdminCredentialRequestCSR()
+	csrReadTarget := kubeapplier.ResourceReference{
+		Group:    "certificates.k8s.io",
+		Version:  "v1",
+		Resource: "certificatesigningrequests",
+		Name:     fmt.Sprintf("system-admin-credential-%s", credName),
```
miguelsorianod, line 220: Would using csrApprovalObj.Name be more appropriate here instead?

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
```diff
@@ -0,0 +1,235 @@
...
+	serviceProviderCluster, err := c.serviceProviderClusterLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if cosmosstorageutils.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get ServiceProviderCluster: %w", err))
+	}
+	mcResourceID := serviceProviderCluster.Status.ManagementClusterResourceID
+	if mcResourceID == nil {
+		return nil
```
miguelsorianod, line 143: Should we log here too?

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
```diff
@@ -0,0 +1,235 @@
...
+
+	controlPlaneNamespace := serviceProviderCluster.Status.ControlPlaneNamespace
+	if len(controlPlaneNamespace) == 0 {
+		logger.Info("waiting for ServiceProviderCluster.Status.ControlPlaneNamespace before creating desires")
+		return nil
+	}
+
+	kubeApplierClient := c.kubeApplierDBClients.For(ctx, mcResourceID)
+	if kubeApplierClient == nil {
+		return nil
```
miguelsorianod, line 155: Should we log here too?

backend/pkg/controllers/cluster/credentialrevocation/creation/revocation_desires.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil
+	}
+
+	kubeApplierClient := c.kubeApplierDBClients.For(ctx, mcResourceID)
+	if kubeApplierClient == nil {
+		logger.Info("waiting for kube-applier client for management cluster", "managementCluster", mcResourceID.String())
+		return nil
+	}
+
+	suffix := revocation.Spec.RevokeOpSuffix
```
miguelsorianod, line 138: In the other PR I asked about the meaning of RevokeOpSuffix. Can you explain it so I can understand?

In the internal api description it mentions:
```
	// RevokeOpSuffix is the shortened operation ID used in CRR object names.
```

but I would like to understand why

backend/pkg/controllers/cluster/credentialrevocation/creation/revocation_desires.go
```diff
@@ -0,0 +1,190 @@
...
+	// 1. CRR ApplyDesire.
+	crrObj := systemadmincredential.BuildRevocationRequest(owner, suffix, controlPlaneNamespace)
+	crrTarget := kubeapplier.ResourceReference{
+		Group:     "certificates.hypershift.openshift.io",
+		Version:   "v1alpha1",
+		Resource:  "certificaterevocationrequests",
+		Namespace: controlPlaneNamespace,
+		Name:      crrObj.Name,
+	}
+	crrDesireName := "systemadmincredentialrevocation"
```
miguelsorianod, line 179: Similar to kubeapplier.ReadDesireNameForSystemAdminCredentialRequestRevocation() should we define a ApplyDesireNameForSystemAdminCredentialRequestRevocation() in that package?

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
miguelsorianod, line 60: Here it mentions RBAC bundles in the code comments, but I can't see it. Is the doc outdated, or is it missing?

backend/pkg/controllers/cluster/credentialrevocation/creation/revocation_desires.go
```diff
@@ -0,0 +1,190 @@
...
+	serviceProviderClusterLister corelisters.ServiceProviderClusterLister
+	applyDesireLister            dblisters.ApplyDesireLister
+	readDesireLister             dblisters.ReadDesireLister
+}
+
+var _ controllerutils.SystemAdminCredentialRevocationSyncer = (*revocationDesires)(nil)
+
+// NewRevocationDesiresController returns a RevocationWatchingController that
+// manages the CertificateRevocationRequest (CRR) desires used to revoke a
+// cluster's already-issued certificates. It creates the RBAC, CRR ApplyDesire,
```
miguelsorianod: Here it mentions RBAC in the code comments, but I can't see it. Is the doc outdated, or is it missing?

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
miguelsorianod: I see other controller check the readdesire, but how will we know if there's an issue at the ApplyDesire level?

backend/pkg/controllers/cluster/credentialrequest/creation/issuance_observer.go
```diff
@@ -0,0 +1,173 @@
...
+	replacement := cred.DeepCopy()
+	meta.SetStatusCondition(&replacement.Status.Conditions, metav1.Condition{
+		Type:    api.SystemAdminCredentialRequestConditionIssued,
+		Status:  metav1.ConditionTrue,
+		Reason:  "Issued",
+		Message: "CSR has been signed",
+	})
+	replacement.Status.SignedCertificate = signedCert
+
+	_, err = credCRUD.Replace(ctx, replacement, nil)
```
miguelsorianod, line 143: Here if we get a preconditionfailed error it will trigger a controller error which will contribute to the controller error metrics. Should we check it to avoid returning an error?

backend/pkg/controllers/cluster/credentialrequest/creation/issuance_observer.go
```diff
@@ -0,0 +1,173 @@
...
+) error {
+	replacement := cred.DeepCopy()
+	meta.SetStatusCondition(&replacement.Status.Conditions, metav1.Condition{
+		Type:    api.SystemAdminCredentialRequestConditionFailed,
+		Status:  metav1.ConditionTrue,
+		Reason:  "CSRFailed",
+		Message: message,
+	})
+
+	_, err := credCRUD.Replace(ctx, replacement, nil)
```
miguelsorianod, line 169: Similar comment to https://github.com/Azure/ARO-HCP/pull/6390/changes#r3735262249

backend/pkg/controllers/cluster/credentialrequest/operations/dispatch_request_credential.go
```diff
@@ -0,0 +1,218 @@
...
+		return false
+	}
+	return true
+}
+
+func (c *dispatchRequestCredential) SynchronizeOperation(ctx context.Context, key controllerutils.OperationKey) error {
+	logger := utils.LoggerFromContext(ctx)
+	logger.Info("checking operation")
+
+	operation, err := c.resourcesDBClient.Operations(key.SubscriptionID).Get(ctx, key.OperationName)
```
miguelsorianod, line 99: Should this be an operation lister?

backend/pkg/controllers/cluster/credentialrequest/operations/dispatch_request_credential.go
```diff
@@ -0,0 +1,218 @@
...
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get active operation: %w", err)
+	}
+	if !c.ShouldProcess(ctx, operation) {
+		return nil
+	}
+
+	cluster, err := c.resourcesDBClient.HCPClusters(operation.ExternalID.SubscriptionID, operation.ExternalID.ResourceGroupName).Get(ctx, operation.ExternalID.Name)
```
miguelsorianod: is the cluster client being cosmos instead of lister intentional?

backend/pkg/controllers/cluster/credentialrequest/operations/dispatch_request_credential.go
```diff
@@ -0,0 +1,218 @@
...
+
+	// Cancel the operation if a revocation is in progress.
+	if len(cluster.ServiceProviderProperties.RevokeCredentialsOperationID) > 0 {
+		logger.Info("revocation in progress, canceling operation",
+			"revoke_credentials_operation_id", cluster.ServiceProviderProperties.RevokeCredentialsOperationID)
+
+		replacement := operation.DeepCopy()
+		apihelpers.CancelOperation(replacement, c.clock.Now())
+
+		_, err = c.resourcesDBClient.Operations(key.SubscriptionID).Replace(ctx, replacement, nil)
```
miguelsorianod, line 123: We should check for precondition failed to return gracefully

backend/pkg/controllers/cluster/credentialrequest/operations/dispatch_request_credential.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	for _, cred := range iter.Items(ctx) {
+		if cred.Spec.OperationID == operationIDStr {
+			replacement := operation.DeepCopy()
+			if replacement.SystemAdminCredentialRequest == nil {
+				replacement.SystemAdminCredentialRequest = &api.OperationSystemAdminCredentialRequest{}
+			}
+			replacement.SystemAdminCredentialRequest.SystemAdminCredentialRequestResourceID = cred.ResourceID
+			_, err = c.resourcesDBClient.Operations(key.SubscriptionID).Replace(ctx, replacement, nil)
+			if err != nil {
```
miguelsorianod, line 156: Same comment as https://github.com/Azure/ARO-HCP/pull/6390/changes#r3735291276

backend/pkg/controllers/cluster/credentialrequest/operations/dispatch_request_credential.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := iter.GetError(); err != nil {
+		return utils.TrackError(fmt.Errorf("failed to iterate SystemAdminCredentialRequests: %w", err))
+	}
+
+	if operation.SystemAdminCredentialRequest == nil || operation.SystemAdminCredentialRequest.CertificateSigningRequest == "" {
+		return fmt.Errorf("operation %s has no CertificateSigningRequest", operation.OperationID.Name)
+	}
+
+	// Generate a credential name: first 16 hex chars of a new UUID.
+	credName := strings.ReplaceAll(uuid.New().String(), "-", "")[:16]
```
miguelsorianod, line 173: Is there some reason we do this? I guess it might be because of K8s length restrictions? if so, it would be good to document it.

This also theorically increases the chances of collisions as we take less bits than the full UUID. That should still be fine given that if there's a collision the Cosmos create fall will fail because of a duplicated entry

backend/pkg/controllers/cluster/credentialrequest/operations/dispatch_request_credential.go
```diff
@@ -0,0 +1,218 @@
...
+
+	// Determine the username from the operation's client identity.
+	username := operation.ClientID
+	if username == "" {
+		username = "system-admin"
```
miguelsorianod: Does that mean in that case we are asking for K8s system-admin credentials, and if ClientID is set then we ask for the specific user instead of system-admin? or is the username purely descriptive?

backend/pkg/controllers/cluster/credentialrequest/operations/dispatch_request_credential.go
```diff
@@ -0,0 +1,218 @@
...
+
+	replacement := operation.DeepCopy()
+	if replacement.SystemAdminCredentialRequest == nil {
+		replacement.SystemAdminCredentialRequest = &api.OperationSystemAdminCredentialRequest{}
+	}
+	replacement.SystemAdminCredentialRequest.SystemAdminCredentialRequestResourceID = credResourceID
+
+	_, err = c.resourcesDBClient.Operations(key.SubscriptionID).Replace(ctx, replacement, nil)
+	if err != nil {
+		return utils.TrackError(err)
```
miguelsorianod, line 217: Same as https://github.com/Azure/ARO-HCP/pull/6390/changes#r3735291276

backend/pkg/controllers/cluster/credentialrequest/operations/dispatch_request_credential.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	logger.Info("dispatched SystemAdminCredential",
+		"credential_name", credName,
```
miguelsorianod: nitpick: we accept relatively long lines in the project. This can be in a single one

backend/pkg/controllers/cluster/credentialrequest/operations/operation_request_credential.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/Azure/ARO-HCP/backend/pkg/utils/controllerutils"
+	operationbase "github.com/Azure/ARO-HCP/backend/pkg/utils/operationutils"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/api/arm"
+	"github.com/Azure/ARO-HCP/internal/database/cosmosstorage/corecosmosstorage"
+	"github.com/Azure/ARO-HCP/internal/database/cosmosstorage/cosmosstorageutils"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+type operationRequestCredentialPoll struct {
```
miguelsorianod, line 36: I see we have two controllers that deal with operation for the same concept. Is there some reason we split it? For the other resources we only have a single operation controller and all the other controllers are independent of the operation.

backend/pkg/controllers/cluster/credentialrequest/operations/operation_request_credential.go
```diff
@@ -0,0 +1,151 @@
...
+	var notifyFn operationbase.PostAsyncNotificationFunc
+	if c.notificationClient != nil {
+		client := c.notificationClient
+		notifyFn = func(ctx context.Context, op *api.Operation) error {
+			return operationbase.PostAsyncNotification(ctx, client, op)
+		}
+	}
+	err = operationbase.UpdateOperationStatus(ctx, c.clock, c.resourcesDBClient, oldOperation, newOperationStatus, newOperationError, notifyFn)
+	if err != nil {
+		return utils.TrackError(err)
```
miguelsorianod, line 149: We should not error if we get cosmos preconditionfailed

backend/pkg/controllers/cluster/credentialrequest/operations/operation_request_credential.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	if !c.ShouldProcess(ctx, oldOperation) {
+		return nil
+	}
+
+	credResourceID := oldOperation.SystemAdminCredentialRequest.SystemAdminCredentialRequestResourceID
+	credName := credResourceID.Name
+
+	// Look up the SystemAdminCredentialRequest doc.
+	cred, err := c.resourcesDBClient.HCPClusters(oldOperation.ExternalID.SubscriptionID, oldOperation.ExternalID.ResourceGroupName).SystemAdminCredentialRequests(
```
miguelsorianod, line 107: What happens if the SystemAdminCredentialRequest has the deletion marker?

backend/pkg/controllers/cluster/credentialrevocation/operations/dispatch_revoke_credentials.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if operation.Request != api.OperationRequestSystemAdminCredentialRevocation {
+		return false
+	}
+	if operation.Status != arm.ProvisioningStateAccepted {
+		return false
+	}
+	return true
+}
+
```
miguelsorianod, line 90: Similar comments here as the corresponding controller of request credential

backend/pkg/controllers/cluster/credentialrevocation/creation/revocation_mark_requests.go
```diff
@@ -0,0 +1,130 @@
...
+	for _, cred := range iter.Items(ctx) {
+		if cred.Status.DeletionTimestamp != nil {
+			continue
+		}
+		replacement := cred.DeepCopy()
+		now := metav1.NewTime(c.clock.Now())
+		replacement.Status.DeletionTimestamp = &now
+		// This controller is keyed on the revocation, not the credential request,
+		// so a failed credential write is NOT retriggered by our informer. Return
+		// an error (including on precondition failure) so the workqueue retries.
```
miguelsorianod, line 100: I guess this is good enough for now but if PreconditionFaileds at some point become frequent enough that it ends up crossing the threshold that triggers alerting we will need to change this to enqueue explicitly or something from here

backend/pkg/controllers/cluster/credentialrevocation/deletion/revocation_deletion.go
```diff
@@ -0,0 +1,135 @@
...
+var _ controllerutils.SystemAdminCredentialRevocationSyncer = (*revocationDeletion)(nil)
+
+// NewRevocationDeletionController returns a RevocationWatchingController that runs
+// once a revocation has been marked for deletion (Status.DeletionTimestamp set). It
+// tears down the revocation's desires (CRR ApplyDesire/ReadDesire and the CRR
+// RBAC) via the shared deleteDesires helper and, when they are all gone, deletes
+// the SystemAdminCredentialRevocation document. The RevokeCredentials operation
+// completes once the document no longer exists.
+func NewRevocationDeletionController(
+	activeOperationLister corelisters.ActiveOperationLister,
```
miguelsorianod: This variable seems to be unused

backend/pkg/controllers/cluster/credentialrequest/deletion/cluster_deletion_cleanup.go
```diff
@@ -0,0 +1,146 @@
...
+//  1. Flips every credential ApplyDesire to Type=Delete so the kube-applier tears
+//     down the applied object.
+//  2. Waits for those Delete desires to report success, then removes them.
+//  3. Deletes the credential ReadDesires.
+//  4. Deletes the SystemAdminCredentialRequest document itself.
+//
+// The desire teardown is performed by the shared deleteDesires helper, which is
+// also used by the post-issuance-cleanup and revocation-deletion controllers.
+func NewClusterDeletionCleanupController(
+	activeOperationLister corelisters.ActiveOperationLister,
```
miguelsorianod: This variable seems to be unused

backend/pkg/controllers/cluster/credentialrequest/deletion/cluster_deletion_cleanup_test.go
```diff
@@ -0,0 +1,95 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/utils/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/database/cosmosstoragetesting/corecosmosstoragetesting"
+	"github.com/Azure/ARO-HCP/internal/database/listertesting/corelistertesting"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+func TestClusterDeletionCleanup_SyncOnce(t *testing.T) {
```
miguelsorianod, line 34: This seems to need a bit more coverage

backend/pkg/controllers/cluster/credentialrequest/deletion/post_issuance_cleanup.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+var _ controllerutils.SystemAdminCredentialRequestSyncer = (*postIssuanceCleanup)(nil)
+
+// NewPostIssuanceCleanupController returns a CredentialRequestWatchingController
+// that eagerly tears down per-credential CSR/CSRApproval/RBAC ApplyDesires and
+// ReadDesires once an individual credential reaches Issued or Failed condition,
+// freeing MC resources. The teardown reuses the shared deleteDesires helper.
+func NewPostIssuanceCleanupController(
+	activeOperationLister corelisters.ActiveOperationLister,
```
miguelsorianod: This variable seems to be unused

backend/pkg/controllers/cluster/credentialrequest/deletion/post_issuance_cleanup.go
```diff
@@ -0,0 +1,139 @@
...
+	serviceProviderCluster, err := c.serviceProviderClusterLister.Get(ctx, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if cosmosstorageutils.IsNotFoundError(err) {
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(err)
+	}
+	mcResourceID := serviceProviderCluster.Status.ManagementClusterResourceID
+	if mcResourceID == nil {
+		return nil
```
miguelsorianod, line 111: Should we add logging here?

backend/pkg/controllers/cluster/credentialrequest/deletion/post_issuance_cleanup.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	kubeApplierClient := c.kubeApplierDBClients.For(ctx, mcResourceID)
+	if kubeApplierClient == nil {
+		return nil
```
miguelsorianod, line 117: Should we add logging here?

backend/pkg/controllers/cluster/credentialrequest/deletion/post_issuance_cleanup.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	logger.Info("post-issuance cleanup complete", "credential", credName)
+	return nil
+}
+
+// isCredentialDesire returns true if the desire name contains the credential
+// name as a suffix component (e.g. "systemAdminCredentialCSR-<credName>").
+func isCredentialDesire(desireName, credName string) bool {
```
miguelsorianod: This can be a method of the controller so it doesn't occupy the package namespace as it's only used here

backend/pkg/controllers/cluster/credentialrequest/deletion/post_issuance_cleanup_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/utils/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/database/cosmosstoragetesting/corecosmosstoragetesting"
+	"github.com/Azure/ARO-HCP/internal/database/listertesting/corelistertesting"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+func TestPostIssuanceCleanup_SyncOnce(t *testing.T) {
```
miguelsorianod, line 31: This seems to be missing a bit of coverage

backend/pkg/controllers/cluster/credentialrequest/deletion/revoked_gc.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+type revokedGC struct {
+	clock             utilsclock.PassiveClock
+	resourcesDBClient corecosmosstorage.ResourcesDBClient
+}
+
+var _ controllerutils.SystemAdminCredentialRequestSyncer = (*revokedGC)(nil)
+
+// NewRevokedGCController returns a CredentialRequestWatchingController that
```
miguelsorianod, line 44: Is the addition of this so it covers potential race condition between creation and deletion?

backend/pkg/controllers/cluster/credentialrequest/deletion/revoked_gc.go
```diff
@@ -0,0 +1,98 @@
...
+// deletes every SystemAdminCredentialRequest document 48 hours after it was
+// created, regardless of the request's status.
+func NewRevokedGCController(
+	clock utilsclock.PassiveClock,
+	activeOperationLister corelisters.ActiveOperationLister,
```
miguelsorianod: This variable seems to be unused

backend/pkg/controllers/cluster/credentialrevocation/deletion/revocation_completion_test.go
```diff
@@ -0,0 +1,99 @@
...
+	clocktesting "k8s.io/utils/clock/testing"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/utils/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/api"
+	"github.com/Azure/ARO-HCP/internal/database/cosmosstoragetesting/corecosmosstoragetesting"
+	"github.com/Azure/ARO-HCP/internal/database/listertesting/kubeapplierlistertesting"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+func TestRevocationCompletion_SyncOnce(t *testing.T) {
```
miguelsorianod, line 35: This could have a bit more coverage

backend/pkg/controllers/cluster/credentialrevocation/deletion/revocation_deletion.go
```diff
@@ -0,0 +1,135 @@
...
+		return utils.TrackError(fmt.Errorf("failed to delete SystemAdminCredentialRevocation: %w", err))
+	}
+	logger.Info("deleted revocation after tearing down its desires", "revokeOpSuffix", suffix)
+	return nil
+}
+
+// isRevocationDesire reports whether a desire name belongs to the revocation
+// identified by suffix. Every revocation desire (CRR ApplyDesire/ReadDesire and
+// the CRR RBAC ApplyDesires) is named with the revocation's unique suffix.
+func isRevocationDesire(desireName, suffix string) bool {
```
miguelsorianod: This can be a method of the controller as it's not used anywhere else in the package.

backend/pkg/controllers/cluster/credentialrevocation/deletion/revocation_deletion_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/go-logr/logr/testr"
+	"github.com/stretchr/testify/require"
+
+	"github.com/Azure/ARO-HCP/backend/pkg/utils/controllerutils"
+	"github.com/Azure/ARO-HCP/internal/database/cosmosstoragetesting/corecosmosstoragetesting"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+func TestRevocationDeletion_SyncOnce(t *testing.T) {
```
miguelsorianod, line 29: This needs a bit more coverage

kube-applier/zz_fixture_TestHelmTemplate_dev_westus3_mgmt_1_kube_applier.yaml
deads2k: Just this I think.

frontend/pkg/frontend/frontend.go
deads2k: > Is there some reason we don't have AdminCredentialRequest as a `VersionedCreatableResource` similar to other resources?

Haven't unwound that ball of mud yet.  tbh, I don't know what it does, but if we don't need it (and we don't), I'd avoid adding it.

internal/systemadmincredential/builders.go
deads2k: > will we want then to promote it at some point at the kube-applier package level?

probably.



> Side question: is aro-hcp.openshift.io the domain we can use in general for K8s labels and annotations?

yes

internal/systemadmincredential/builders.go
deads2k: > Are there K8s name length restrictions that apply at the K8s level and also at specifically CertificateSigningRequest level? do we want to enforce some error/panic checking here?

No. We'll let kube do the validation and spend our effort making sure we see the errors.

internal/systemadmincredential/builders.go
deads2k: > if it shouldn't occur then panicking is a possibility, b

Yes. Coding error can panic.

internal/systemadmincredential/builders.go
deads2k: So these links aren't working. What do you want?

internal/systemadmincredential/builders.go
deads2k: So these links aren't working. What do you want?

test/util/framework/helpers_v20260901preview.go
deads2k: We aren't.  It will silently fall back. That's why the proof PR doesn't have a fallback.

test/e2e/cluster_authorized_cidrs_connectivity.go
deads2k: We are keeping some using both to ensure both paths work.

test-integration/backend/launch/datadump_test.go
deads2k: Yes, and to ensure it is forced to keep compiling. I had weird problems adding this type.

test-integration/backend/launch/datadump_test.go
deads2k: Yes.  Had really weird errors.

backend/pkg/utils/controllerutils/system_admin_credential_revocation_watching_controller.go
deads2k: > Is there some reason here we don't want to trigger on readdesire?

Adding

backend/pkg/kubeapplierhelpers/desire_helpers.go
deads2k: > Out of curiosity: does this hold true for all of them?

No, only used for testing now.

backend/pkg/kubeapplierhelpers/desire_deletion.go
deads2k: > what does this mean?

We'll probably followup with conditions for SuccessfulApply and SuccessfulDelete

backend/pkg/kubeapplierhelpers/desire_deletion.go
deads2k: > That means that if the applydesire is changed from apply to delete we will always have at least two calls needed to DeleteDesires right?

Correct

backend/pkg/kubeapplierhelpers/desire_deletion.go
deads2k: > How do we know that successful == true means it's been deleted? is it because we unset the conditions above? because otherwise it could be because of it being applied instead of deleted

Correct.

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
deads2k: > Side question: With the new directory organization where creation and deletion is separate, how would we introduce controllers that deal with creation and deletion in the same place, which we might want for some controllers? unifying those can be desirable in some cases and reduces the chances of racing between create and delete

So far I'm liking the division here.  I'm open to others.

backend/pkg/controllers/cluster/credentialrequest/creation/desires_creator.go
deads2k: > I see other controller check the readdesire, but how will we know if there's an issue at the ApplyDesire level?

We won't do anything specific with it.  We could later consdier aggregating into a condition for parent resources.

backend/pkg/controllers/cluster/credentialrevocation/creation/revocation_desires.go
deads2k: > In the other PR I asked about the meaning of RevokeOpSuffix. Can you explain it so I can understand?
> 
> In the internal api description it mentions:
> 
> ```
> 	// RevokeOpSuffix is the shortened operation ID used in CRR object names.
> ```
> 
> but I would like to understand why

We need non-conflict kube resource names, this gives a way to do that, while staying within naming limits and restrictions.  We could use generateName, but if we did, we'd have to track the apply in order to do the ReadDesire.

backend/pkg/controllers/cluster/credentialrevocation/creation/revocation_desires.go
deads2k: > Similar to kubeapplier.ReadDesireNameForSystemAdminCredentialRequestRevocation() should we define a ApplyDesireNameForSystemAdminCredentialRequestRevocation() in that package?

Below threshold of concern for this PR.  Feel free to refactor later.

backend/pkg/controllers/cluster/credentialrequest/operations/operation_request_credential.go
deads2k: > I see we have two controllers that deal with operation for the same concept. Is there some reason we split it? For the other resources we only have a single operation controller and all the other controllers are independent of the operation.

one is legacy and one is new

backend/pkg/controllers/cluster/credentialrequest/operations/dispatch_request_credential.go
deads2k: Adding validation to frontend of CN and organizations and removing this field.

deads2k: It's cruft from a previous iteration of the designs

backend/pkg/controllers/cluster/credentialrequest/operations/operation_request_credential.go
deads2k: > What happens if the SystemAdminCredentialRequest has the deletion marker?

deletion markers make the GET return a notfound.

deads2k:
good review.  All comments addressed. Moving ahead to unblock review and get soak

/label ltm

PR #6399, opened 2026-08-04
Schema changes for submitting Node health to geneva health systems using Geneva Health SDK 

venkateshsredhat:
What ?
submitting Node health to geneva health systems using Geneva Health SDK we need these schema changes .

Why?
We want to enable Resource Health Check for Node related metrics , currently these node metrics reach our HCP AMW but we want them to reach Geneva Health Systems to enable onboarding to Resource Health .


stevekuznetsov (approved):
/lgtm
/approve

PR #6403, opened 2026-08-04
feat(api): new API version 2026-10-03-preview that introduces `KeyEncryptionKeyURL`

cadenmarchese:
https://redhat.atlassian.net/browse/ARO-27778

### What

This is the first of 3 separate PRs / releases to adopt `KeyEncryptionKeyURL` API field and deprecate the old KMS `activeKey` values. The roadmap is as follows:

Add new API version v20261003preview for mHSM KMS changes (https://github.com/Azure/ARO-HCP/pull/6403)
a. Use full Key URL in the API rather than separate KMS fields
b. Backfill older cluster cosmosdb records in cosmos_migration.go
c. Continue reading old KMS fields for now

pkg/internal prefers KeyEncryptionKeyURL for KMS (https://github.com/Azure/ARO-HCP/pull/6863)
a. Export ParseKeyEncryptionKeyURL
b. Build activeKeyBuilder with key name and vault name
c. Safe to apply without a completed cosmosdb migration

Stop writing activeKey in all API versions (https://github.com/Azure/ARO-HCP/pull/6864)
a. Validate KeyEncryptionKeyURL rather than KmsKey
b. Drop ActiveKey from pkg/internal conversion

Old structure:
```
					"kms": {
						"activeKey": {
							"name": "vc-encryption-key",
							"version": "2024-12-01-preview"
						},
						"vaultName": "vc-key-vault",
						"visibility": "Public"
					}
```
New structure:
```
					"kms": {
						"keyEncryptionKeyUrl": "https://vc-key-vault.vault.azure.net/keys/vc-encryption-key/2024-12-01-preview",
						"visibility": "Public"
					}
```

### Why
- The API needs to expose the full Key URL for Azure Policy, so that we can create an alias allowing users to audit keyvault type, name, key and version using the same string. Using the full Key URL is the official recommendation (see https://eng.ms/docs/products/arm/api_contracts/resource-provider-contract/v10/common-api-contracts#customer-managed-key-encryption, and I have also confirmed this with contacts at MSFT).
- The keyvault URLs in other clouds (such as FairFax) is different. Using the URL instead of a few disconnected fields that point to the key will save the RP additional effort determining what cloud the keyvault is in.
- The existing KMS object contains nested API fields that are redundant once Key URL is added, so we can clean those up.

### Future
- This API change unblocks mHSM support and eventually also streamlines support for BYOK in sovereign clouds. It also will streamline the onboarding of azure policy aliases.

### Testing

- Changes to fuzz tests

- While this API change(s) originated from the mHSM feature, I'm proposing that we fully decouple the API side from the mHSM feature end-to-end since this is a Microsoft best practice that we should do regardless of our support for other key stores. I'm proposing that we move forward with this API change without mHSM code changes fully in place to de-risk both and ensure that we can deliver them.

### PR Checklist

- [x] Self-reviewed the diff


api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/examples/HcpOperatorIdentityRoleSets_List_MinimumSet_Gen.json
ahitacat, line 1: Are these `Minimum` files required?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
```diff
@@ -2781,66 +2772,17 @@
         }
       },
       "required": [
-        "vaultName",
-        "visibility",
-        "activeKey"
+        "visibility"
       ]
```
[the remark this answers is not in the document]
ahitacat: +1 to what copilot says. Is this something that is missing or we want to make this keyvault not required?

internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
```diff
@@ -236,28 +238,64 @@ func newCustomerManagedEncryptionProfile(from *api.CustomerManagedEncryptionProf
...
-		VaultName:  api.PtrOrNil(from.ActiveKey.VaultName),
+	out := generated.KmsEncryptionProfile{
 		Visibility: api.PtrOrNil(generated.KeyVaultVisibility(from.Visibility)),
 	}
+	if from.KeyEncryptionKeyURL != "" {
+		out.KeyEncryptionKeyURL = api.PtrOrNil(from.KeyEncryptionKeyURL)
+	} else if from.ActiveKey.Name != "" {
+		// Fallback for old clusters created before keyEncryptionKeyUrl was stored.
+		// Safe to assume vault.azure.net — mHSM did not exist before this field.
+		keyURL := buildKeyEncryptionKeyURL(from.ActiveKey.VaultName, from.ActiveKey.Name, from.ActiveKey.Version)
```
ahitacat, line 258: Additionally, when working on RP side, this old information will be mutated to URL, right? Shouldn't we rely on that instead?

internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
```diff
@@ -633,44 +671,53 @@ func normalizeAutoscaling(p *generated.ClusterAutoscalingProfile, out *api.Clust
...
+	if p.Kms != nil {
+		if p.Kms.KeyEncryptionKeyURL == nil || len(*p.Kms.KeyEncryptionKeyURL) == 0 {
+			errs = append(errs, field.Required(fldPath.Child("kms", "keyEncryptionKeyUrl"), "keyEncryptionKeyUrl is required when kms is specified"))
+		} else {
+			if out.Kms == nil {
+				out.Kms = &api.KmsEncryptionProfile{}
+			}
+			out.Kms.KeyEncryptionKeyURL = *p.Kms.KeyEncryptionKeyURL
+			out.Kms.Visibility = api.KeyVaultVisibility(api.Deref(p.Kms.Visibility))
+			vaultName, keyName, version, err := parseKeyEncryptionKeyURL(*p.Kms.KeyEncryptionKeyURL)
```
ahitacat, line 729: This seems some check that we usually do in the frontend. Why are we still populating vaultName, keyName and version?

test/util/framework/helpers_v20260901preview.go
```diff
@@ -505,12 +505,12 @@ func BuildHCPClusterFromParams20260901(
 					CustomerManaged: &hcpsdk20260901preview.CustomerManagedEncryptionProfile{
 						EncryptionType: to.Ptr(hcpsdk20260901preview.CustomerManagedEncryptionType(parameters.EncryptionType)),
 						Kms: &hcpsdk20260901preview.KmsEncryptionProfile{
-							VaultName:  to.Ptr(parameters.KeyVaultName),
 							Visibility: to.Ptr(hcpsdk20260901preview.KeyVaultVisibility(parameters.KeyVaultVisibility)),
-							ActiveKey: &hcpsdk20260901preview.KmsKey{
-								Name:    to.Ptr(parameters.EtcdEncryptionKeyName),
-								Version: to.Ptr(parameters.EtcdEncryptionKeyVersion),
-							},
+							KeyEncryptionKeyURL: to.Ptr(fmt.Sprintf("https://%s.vault.azure.net/keys/%s/%s",
```
ahitacat: Why no start passing the URL instead of the 3 parameters?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/examples/HcpOperatorIdentityRoleSets_List_MinimumSet_Gen.json
cadenmarchese: From what I can tell, yes, these examples are independent of my change and were automatically generated when I ran `oav generate-examples` in response to my API changes. It looks like they were just never committed to the repository when this API was scaffolded, so committing them should be correct.

internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
cadenmarchese: I think that since KeyEncryptionKeyURL is a required field, we would need to return it on a GET even if it's an old cluster and the cluster document in CosmsoDB only has the old values, so we need some kind of conversion in the API layer from the old values to the new.

We can make keyEncryptionKeyUrl optional in TypeSpec so we can omit it entirely on GET, but I think this goes against existing behavior where KMS is required, so assuming that we want to preserve that I think we'd need to keep this.

internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
cadenmarchese: OCM needs the values as they were before (I'm guessing that's why the initial API values were vaultName, keyName, version)

test/util/framework/helpers_v20260901preview.go
cadenmarchese: Maybe this is open to discussion, but I figured it would be easier to keep the scope of this change at the API only for now rather than change e2e behavior since e2e behavior only tests with standard keyvaults today. Once I add support for an mHSM keyvault and test for that, I'll need to change this, but I would argue it can be a followup.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -704,8 +705,17 @@ model KmsEncryptionProfile {
...
   @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   activeKey: KmsKey;
+
+  /** The versioned URL of the key used for ETCD data encryption, hosted in Azure Key Vault or Managed HSM.
+   * Format: https://{vault-name}.vault.azure.net/keys/{key-name}/{key-version}
+   * or https://{hsm-name}.managedhsm.azure.net/keys/{key-name}/{key-version}
+   */
+  @added(Versions.v2026_09_01_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  keyEncryptionKeyUrl: url;
```
mbarnes: This might be already decided, but if we have to break the API anyway I would have preferred a discriminated union for for the two key management types over a single all-in-one URL field that we hope will be future-proof.  Just personal preference.

cadenmarchese: I thought about that during the initial spike, but ended up proposing the URL because having a discriminated union (say keyVaultType of `managedHSM` or `keyVault`) wouldn't account for MSFT-side changes in either URL structure / domains, or the addition of new key vault types other than those two, which would require more API changes later.

Also, if you look at other Azure resources (like disk encryption sets for example), the Key URL is always what is used, and the policy team recommends it. I'm hoping that the full KeyURL brings us in better alignment with other ARM resources.

internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
ahitacat: Ok this makes sense, although my question wasn't very clear.

The idea is that in the RP we should have some mechanism that also adds the keyURL to the cluster representation. So, when getting a new cluster that doesn't have this keyURL this will be calculated. With that GETs from the new API version will show the keyURL even though the internal representation doesn't contain it.

ahitacat (commented):
@cadenmarchese I don't understand this design if we compare it to how it will be supported in hypershift https://github.com/openshift/hypershift/pull/9199

That change seems to only add a new field (`KeyVaultType`) to configure kms, the keyvault DNS is calculated inside Hypershift. 

Adding the keyVaultURL to the configuration will make us validate the basename to be the same as the ones Hypershift calculates and divided it in the 4 parameters hypershift requires. 

Why we not just add that field and create the hostedCluster according? Is there a different work planned to change Hypershift?

cadenmarchese:
@ahitacat I'm not as concerned about the shape of the HyperShift API with regards to keyvault type, since that is an internal component that customers don't interact with and can't create Azure policies around. I'd be opposed to making a customer-facing API change for the sake of aligning with HyperShift when there's already an established ARM convention in Azure around CMK that has been specifically recommended by policy folks (in my opinion, this should have been the API shape from the beginning). Moving to a key URL solves multiple problems at once:

1. Customers can now provide an mHSM to the install and the API will know that it isn't a standard keyvault
2. Customers can create policies that audit keyvault type, keyvault name, key and key version using the same single API field
3. Policies that audit keyEncryptionKeyUrl patterns can use the same rule across ARO-HCP and other Azure services
4. The full URL future-proofs us from changes in URL structure or keyvault domains in the future
5. The full URL would support additional keyvault types in the future without further API changes
6. Key rotation is just a single field update
7. We just have one field instead of 3 which is better / simpler

We could solve problem 1 with an additional field only, but not the rest. Ideally HyperShift follows us on this, but I'm not sure it's strictly required since we can handle the translation.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
cadenmarchese: Also discussed in https://github.com/Azure/ARO-HCP/pull/6403#issuecomment-5205036356

internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
cadenmarchese: > The idea is that in the RP we should have some mechanism that also adds the keyURL to the cluster representation.

Do you mean the response we give to the customer / ARM needs to contain the KeyEncryptionKeyURL, or the CosmosDB document for the cluster itself needs to be populated with KeyEncryptionKeyURL in RP if not present?

ahitacat: both, the new ARM API response should contain it (as it is required). 
And the CosmosDB representation will also require the URL, to know if it is a mHSM. And that URL contains the other 3 fields, so we can transition to not having those. This whole discussion is part of the RP work, just wanted to know that.

cadenmarchese: ok, thanks! I left a note in https://redhat.atlassian.net/browse/ARO-27779 that this will be required, then.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
miguelsorianod: Have we considered having both?

if we have the URL only, it means we will need to figure out from it which specific one it is. Can the URL change if it's not public cloud for example? we will need to support multiple clouds at some point so how are we going to approach it?

Additionally, could we have different update restrictions over time where for example one type we would support update or the other not or pace it differently in time?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -704,8 +705,17 @@ model KmsEncryptionProfile {
...
   /** The details of the active key. */
   @added(Versions.v2026_06_30_preview)
+  @removed(Versions.v2026_09_01_preview)
   @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   activeKey: KmsKey;
+
+  /** The versioned URL of the key used for ETCD data encryption, hosted in Azure Key Vault or Managed HSM.
+   * Format: https://{vault-name}.vault.azure.net/keys/{key-name}/{key-version}
+   * or https://{hsm-name}.managedhsm.azure.net/keys/{key-name}/{key-version}
+   */
```
miguelsorianod, line 711: Does hypershift already support the usage of hsm?
The Hypershift API currently doesn't provide the ability to pass the URL, just the vault name, key and version separately. That doesn't include the parent domain part.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -704,8 +705,17 @@ model KmsEncryptionProfile {
 
   /** The details of the active key. */
   @added(Versions.v2026_06_30_preview)
+  @removed(Versions.v2026_09_01_preview)
   @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
   activeKey: KmsKey;
+
```
miguelsorianod, line 708: This seems to break the API. Do we have a clear path and idea on how and when we are going to introduce this, what are we going to do with pre-existing resources and so on?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
```diff
@@ -2752,16 +2752,6 @@
...
-          "type": "string",
-          "description": "vaultName is the name of the keyvault that contains the secret.",
-          "minLength": 1,
-          "maxLength": 255,
-          "x-ms-mutability": [
-            "read",
-            "create"
-          ]
-        },
         "visibility": {
```
miguelsorianod, line 2765: Does visibility also apply in the case of HSM, as well as its supported values?

internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
```diff
@@ -236,28 +238,64 @@ func newCustomerManagedEncryptionProfile(from *api.CustomerManagedEncryptionProf
...
-		ActiveKey:  api.PtrOrNil(newKmsKey(&from.ActiveKey)),
-		VaultName:  api.PtrOrNil(from.ActiveKey.VaultName),
+	out := generated.KmsEncryptionProfile{
 		Visibility: api.PtrOrNil(generated.KeyVaultVisibility(from.Visibility)),
 	}
+	if from.KeyEncryptionKeyURL != "" {
+		out.KeyEncryptionKeyURL = api.PtrOrNil(from.KeyEncryptionKeyURL)
+	} else if from.ActiveKey.Name != "" {
+		// Fallback for old clusters created before keyEncryptionKeyUrl was stored.
+		// Safe to assume vault.azure.net — mHSM did not exist before this field.
```
miguelsorianod: at some point it won't be safe to assume vault.azure.net because we will support multiple clouds, I guess even on older api versions.

miguelsorianod:
> The API needs to expose the full Key URL for Azure Policy, so that we can create an alias allowing users to audit keyvault type, name, key and version using the same string.

Could you expand on this?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -704,8 +705,17 @@ model KmsEncryptionProfile {
...
+  /** The versioned URL of the key used for ETCD data encryption, hosted in Azure Key Vault or Managed HSM.
+   * Format: https://{vault-name}.vault.azure.net/keys/{key-name}/{key-version}
```
miguelsorianod: Is there a possibility that in the future we need to support other types of keyencryptionkeys that might be another kind of keyvault, or even a non keyvault service? I am asking this to evaluate whether discriminators would help here to allow us to expand

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mbarnes: > Also, if you look at other Azure resources (like disk encryption sets for example), the Key URL is always what is used, and the policy team recommends it. I'm hoping that the full KeyURL brings us in better alignment with other ARM resources.

Fair enough if that's the recommended format.  It does protect us from future MSFT-side format changes at the cost of a slightly harder-to-use API for the user.  But I suppose that could be mitigated by some user-friendly assistance in at least the CLI.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
cadenmarchese: Not yet, that is being handled in https://github.com/openshift/hypershift/pull/9199 as we speak. Regarding compat with HyperShift's API, I commented on that here: https://github.com/Azure/ARO-HCP/pull/6403#issuecomment-5205036356

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
cadenmarchese: If this PR merges, as of today:

- keyEncryptionKeyUrl is added to v20260901preview only
- Old clusters created with any older API version have activeKey.* values in CosmosDB, but no keyEncryptionKeyURL
- A v20260901preview GET on an older cluster would have an absent keyEncryptionKeyUrl field but visibility would still be returned
- An older API version GET/PUT/PATCH on any cluster would not change what's in CosmosDB, and likewise KeyEncryptionKeyURL is not wiped by older API version put/patch

Assuming v20260901preview is used with an existing cluster created with an old API version:

- The user would have to provide the KeyURL on the first put or patch via v20260901preview, that would store the URL in CosmosDB alongside the deprecated fields and from that point on v20260901preview GETs return the stored URL only.
  - Note - the user would have to provide the Key URL on first put/patch. We wouldn't assume it.
- Key URL remains invisible for older API versions.

Future (not in this PR):

- Once mHSM support lands in HyperShift and OCM via https://github.com/openshift/hypershift/pull/9199 + the HCP-RP changes, the URL hostname distinguishes vault type so we don't need further API changes
- A CosmosDB migration (suggested in this comment and recorded in JIRA https://github.com/Azure/ARO-HCP/pull/6403#discussion_r3719848421) could enrich cosmosdb with the KeyEncryptionKeyURL for all existing clusters by constructing it from ActiveKey fields with .vault.azure.net (since mHSM support didn't exist yet)
- Older preview API versions would eventually be deprecated, at which point the activeKey.* fields just become internal for clusters that had them

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json
cadenmarchese: Yes, it should apply to mHSM in the same way as a standard KV.

internal/azureapi/v20260901preview/hcpopenshiftclusters_methods.go
cadenmarchese: Agreed. We would only be able to assume vault.azure.net until such time as the Key URL is in the API and available to customers to use.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
cadenmarchese: It's possible that new key vault types and keys and key URL structures will need to be supported in the future. For example, mHSM is a newer thing. If we use the Key URL as intended, we would be able to accommodate those without an API change in the future (although we'd want to ensure this comment doesn't exist in the newer specs)

cadenmarchese: Also @miguelsorianod regarding your question about policy aliases here: https://github.com/Azure/ARO-HCP/pull/6403#issuecomment-5206994755

Azure policy evaluates resource properties using aliases, and a single keyEncryptionKeyUrl field like https://my-vault.vault.azure.net/keys/my-key/abc123 lets a policy rule pattern-match on all four pieces:

- Vault type: hostname suffix (.vault.azure.net vs .managedhsm.azure.net or even a new future type)
- Vault name: hostname prefix (my-vault)
- Key name: second path segment (my-key)
- Key version: third path segment (abc123)

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
cadenmarchese: @miguelsorianod I would have recommended keeping both existing KMS fields + new Key URL if this API was already stable and we wanted to avoid the breaking change due to that. We could technically probably just have both existing KMS fields and new KeyURL be "optional" from the API perspective but have the RP enforce that one or the other is provided. However, since these are preview APIs there should be no real issue from an Azure perspective as long as we commit to cleaning these preview APIs up before they are public for more than a year. We can avoid breaking change rules and remove the redundancy without too much friction if we do it now, but the longer we wait the harder it will be. I would argue that having both existing KMS fields + KeyURL is a cluttered design.

cadenmarchese: To answer the question about other clouds, in FairFax / sovereign clouds, the URL is different:

> https://<vault-name>.vault.usgovcloudapi.net

So with the URL approach we don't need RP-side logic to be aware of the cloud and populate the URL accordingly like we would have to do if we went with discriminators.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
ahitacat: So, just to clarify:

- New API will have the `keyEncryptionKeyUrl` external and the `activeKey` will just be internal. 
- Old API versions will only have the `keyEncryptionKeyUrl` internal, and only populated if it has been updated with the new API or we have some CosmosDB mechanism. The `activeKey` fields will be the external ones.

Is this correct? If so, can you update the description of this PR, this is kind of the design it can be good to have easy to access once this is merged and we want to come back in the future to check some decisions. 

Also with this we are having a duplication of information `keyEncryptionKeyUrl` and `activeKey`. Which one will be the source of truth?
I  am thinking about, these 2 scenarios:

1.  Existing cluster with and `activeKey`, that is updated with the new `keyEncryptionKeyUrl` that differs in the keyName and version from what it had previously.
2. Newly created cluster with a `keyEncryptionKeyUrl`, then is updated using the 20260630 version preview to a new `activeKey.Version`.

In both cases the information inside `keyEncryptionKeyUrl` and `activeKey` will differ.

cadenmarchese: Yes, good call, I've updated the PR description with a combination of our comments so that design is clear without reading the comment threads.

Regarding the duplication of info between` keyEncryptionKeyUrl` and `activeKey`, I guess not having those in sync is a concern for rotation. In my opinion we would need to take a one-way migration approach, and stop using `activeKey` in RP entirely as soon as `keyEncryptionKeyURL` exists in cosmos. We could combine that with some functionality in the frontend to always populate the old  `activeKey` fields with the provided `keyEncryptionKeyUrl` so that GETs using old API versions give accurate key info, but I think we'd also be saying that we don't support key rotation with old API versions since they don't contain `keyEncryptionKeyUrl`. What are your thoughts?

cadenmarchese: In my view, we can just say that rotation / key management in general post-install is only supported via `keyEncryptionKeyUrl` + the new API that contains that field, not the old values, which we will do our best to ensure are up to date but won't use. This will only be a concern up until we clean up these preview API versions.

ahitacat: That will break also the 20260630 API as it is the version that allows updated, I am not sure how that can be managed.
@mbarnes @miguelsorianod what are your thoughts about this?

cadenmarchese: @ahitacat I think we can avoid the cosmosdb migration + the duplication of data and backend breakage we're worried about here by taking a phased approach:

1. We make the API change first, where this newest API version writes both activeKey and KeyEncryptionKeyURL. Older API versions stay the same, only writing activeKey, but EnsureDefaults backfills all cluster docs with the KeyEncryptionKeyURL
2. Raise a separate PR later that switches the older API versions to also writing both activeKey and KeyEncryptionKeyURL to cosmos
3. After all of that has baked, we switch backend to use KeyEncryptionKeyURL exclusively
4. Finally, we raise a change to stop writing activeKey entirely

Currently, this PR satisfies 1, and the rest would be handled in subsequent PRs. Does this sound OK?

deads2k:
/hold we will not merge and then block our external API change queue on a change that requires downstream modification until that downstream modification is present and tested.  https://github.com/openshift/hypershift/pull/9199 appears to be a prereq.

Additionally, our internal representation must align onto a single representation.  It doesn't have to match external representations if we have a good reason to drift, but every internal record must be migrated to represent data the same way.  The usual manner for staging is
1. PR with, write both old and new, read old: allows downgrade to previous level
2. PR with, write old and new, read new: allows downgrade to previous level
3. PR with, remove old field, read new: allows downgrade to previous level

All three PRs to be opened before merging the first one.  There are no IOUs and "I opened a Jira" is not in any way an acceptable replacement for the actual code being written before we merge the first step.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
cadenmarchese: I was recommended this approach by David to introduce and then backfill: https://github.com/Azure/ARO-HCP/pull/6403#issuecomment-5316520594

Let me know if this makes sense.

internal/admission/admit_cluster.go
```diff
@@ -745,7 +745,18 @@ func admitClusterEtcdKmsKeyVersionChange(_ context.Context, admissionContext *Cl
...
+	// Extract the key version from KeyEncryptionKeyURL when available,
+	// falling back to ActiveKey.Version for backward compatibility.
+	newVersion := newObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version
+	oldVersion := oldObj.Etcd.DataEncryption.CustomerManaged.Kms.ActiveKey.Version
+	if newObj.Etcd.DataEncryption.CustomerManaged.Kms.KeyEncryptionKeyURL != "" {
+		_, _, newVersion, _ = coreapi.ParseKeyEncryptionKeyURL(newObj.Etcd.DataEncryption.CustomerManaged.Kms.KeyEncryptionKeyURL)
+	}
+	if oldObj.Etcd.DataEncryption.CustomerManaged.Kms.KeyEncryptionKeyURL != "" {
+		_, _, oldVersion, _ = coreapi.ParseKeyEncryptionKeyURL(oldObj.Etcd.DataEncryption.CustomerManaged.Kms.KeyEncryptionKeyURL)
+	}
```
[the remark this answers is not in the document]
cadenmarchese: No longer relevant to this PR - the pkg/internal changes will land in a separate PR.

internal/api/coreapi/kms_url.go
```diff
@@ -0,0 +1,54 @@
...
+		return "", "", "", fmt.Errorf("invalid keyEncryptionKeyUrl: host must end in .vault.azure.net or .managedhsm.azure.net, got %q", hostname)
+	}
+	if idx := strings.IndexByte(hostname, '.'); idx > 0 {
+		vaultName = hostname[:idx]
+	} else {
+		return "", "", "", fmt.Errorf("invalid keyEncryptionKeyUrl: cannot extract vault name from host %q", hostname)
+	}
+	if vaultName == "" {
+		return "", "", "", fmt.Errorf("invalid keyEncryptionKeyUrl: vault name is empty")
+	}
```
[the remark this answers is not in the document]
cadenmarchese: No longer relevant to this PR - the pkg/internal changes will land in a separate PR.

internal/ocm/cluster_update_dispatch_config.go
```diff
@@ -226,10 +226,15 @@ func clusterUpdateDispatchEtcdFromRP(etcd coreapi.EtcdProfile) clusterUpdateDisp
 
 	res := clusterUpdateDispatchConfigEtcd{}
 	if etcd.DataEncryption.CustomerManaged.EncryptionType == metadataapi.CustomerManagedEncryptionTypeKMS {
+		kms := etcd.DataEncryption.CustomerManaged.Kms
+		version := kms.ActiveKey.Version
+		if kms.KeyEncryptionKeyURL != "" {
+			_, _, version, _ = coreapi.ParseKeyEncryptionKeyURL(kms.KeyEncryptionKeyURL)
+		}
```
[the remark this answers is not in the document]
cadenmarchese: No longer relevant to this PR - the pkg/internal changes will land in a separate PR.

backend/pkg/controllers/cosmosmigration/cosmos_migration.go
```diff
@@ -565,9 +565,25 @@ func migrateManagementClusterContents(ctx context.Context, logger logr.Logger, m
...
+// exist before this field was introduced.
+func migrateClusterKMS(cluster *coreapi.HCPOpenShiftCluster) {
+	kms := cluster.CustomerProperties.Etcd.DataEncryption.CustomerManaged
+	if kms == nil || kms.Kms == nil {
+		return
+	}
+	if kms.Kms.KeyEncryptionKeyURL == "" && kms.Kms.ActiveKey.Name != "" {
+		kms.Kms.KeyEncryptionKeyURL = fmt.Sprintf("https://%s.vault.azure.net/keys/%s/%s",
+			kms.Kms.ActiveKey.VaultName, kms.Kms.ActiveKey.Name, kms.Kms.ActiveKey.Version)
+	}
```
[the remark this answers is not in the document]
cadenmarchese: post KeyEncryptionKeyURL (including future mHSM and sovereign cloud support) it won't be supported to use an older API version for those keyvault types. The cluster KMS will only be migrated in cosmos if it was created with an older API version that only supports the standard vault.azure.net keyvault.

internal/azureapi/v20261003preview/hcpopenshiftclusters_methods.go
```diff
@@ -0,0 +1,804 @@
...
+	}
+	return out
+}
+
+func buildKeyEncryptionKeyURL(vaultName, keyName, version string) string {
+	base := fmt.Sprintf("https://%s.vault.azure.net/keys/%s", vaultName, keyName)
+	if version != "" {
+		return base + "/" + version
+	}
+	return base
```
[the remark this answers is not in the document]
cadenmarchese: See https://github.com/Azure/ARO-HCP/pull/6403#discussion_r3936347776

ahitacat:
@cadenmarchese Can you split this PR into 3? One for the new API scaffold, the API change with the kmskeyURL and the RP change to manage the braking change. This can be to prove everything works as it is an API change. But this PR is taking a magnitude that makes it hard to review it properly.

cadenmarchese:
@ahitacat thanks for your review, I've split this PR into 3 separate commits to decouple the API scaffold from the API change and make this easier to review in parts. If possible, I want to avoid further separating the change out into more PRs since the API version I'm introducing is for this change specifically and is only required for it.

@deads2k thanks for your review, I've split this change according to your guidance:

1. This PR - write both old and new, read old: allows downgrade to previous level
2. https://github.com/Azure/ARO-HCP/pull/6863 - write old and new, read new: allows downgrade to previous level
3. https://github.com/Azure/ARO-HCP/pull/6864 - remove old field, read new: allows downgrade to previous level

While this change does unblock support for additional key stores / vault types, the change isn't strictly 1:1 with mHSM or any other feature, and the idea is to get this API change in first and independently since it is the MSFT-recommended structure for BYOK and then build off of this for subsequent mHSM-related PRs.

ahitacat:
I miss the e2e test that checks the cluster and nodepool creation works for this version with the timebomb.

There will be some changes that are done in 20261001-preview that we should not forgot to update in this new version.

PR #6404, opened 2026-08-04
fix(test): add bicep-install target with retries to test build

janboll:
## Summary
- Adds a `bicep-install` Make target that runs `az bicep install` with up to 5 retries (10s backoff) to handle transient failures
- Makes `bicep-install` a dependency of the `build` target so the Bicep CLI is guaranteed to be available before any `az bicep build` runs

## Why
`az bicep build` fails if the Bicep CLI is not installed. In CI environments this can also fail transiently. Adding a retry loop makes the build more resilient.

## Test plan
- [ ] `make -C test build` succeeds with Bicep CLI not yet installed
- [ ] `make -C test build` succeeds with Bicep CLI already installed (no-op install)
- [ ] Retry logic triggers on transient `az bicep install` failure

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -6,6 +6,13 @@ all: build
 
 BICEP_OUTPUT_DIR=$(TEST_DIR)e2e/test-artifacts/generated-test-artifacts/
 
+bicep-install:
+	@for i in 1 2 3 4 5; do \
+		az bicep install && break || \
+		echo "az bicep install failed (attempt $$i/5), retrying in 10s..." && sleep 10; \
+	done
+.PHONY: bicep-install
+
 DEMO_BICEP_DIR=$(TEST_DIR)../demo/bicep
 DEMO_BICEP_SOURCES = $(shell find $(DEMO_BICEP_DIR) -name '*.bicep')
 DEMO_BICEP_OUTPUTS = $(addprefix $(BICEP_OUTPUT_DIR)standard-cluster-create/,$(addsuffix .json,$(basename $(notdir $(DEMO_BICEP_SOURCES)))))
@@ -37,7 +44,7 @@ PROW_JOB_EXECUTOR = $(TEST_DIR)prow-job-executor
 $(PROW_JOB_EXECUTOR): $(PROW_JOB_EXECUTOR_SOURCES) $(MAKEFILE_LIST) $(TEST_DIR)/go.mod $(TEST_DIR)/go.sum
 	cd $(TEST_DIR) && go build -o $@ ./cmd/prow-job-executor
 
-build: $(PROW_JOB_EXECUTOR) $(ARO_HCP_TESTS)
+build: bicep-install $(PROW_JOB_EXECUTOR) $(ARO_HCP_TESTS)
 .PHONY: build
 
 # Apply the slot-managed identity pool ARM deployment stack for a given environment.


janboll:
bicep should be there: https://github.com/Azure/ARO-HCP/blob/main/dev-infrastructure/openshift-ci/Dockerfile#L15
so there is something broken with the image setup

test/Makefile
```diff
@@ -6,6 +6,13 @@ all: build
 
 BICEP_OUTPUT_DIR=$(TEST_DIR)e2e/test-artifacts/generated-test-artifacts/
```
avollmer-redhat, line 8: **Question**: The CI image already installs bicep at build time (`dev-infrastructure/openshift-ci/Dockerfile:15`) and places it at `/usr/local/bin/bicep`. Running `az bicep install` again at runtime is a no-op at best, and at worst could conflict with the pre-installed binary or fail on permissions (CI runs as a non-root UID).

If `az bicep build` is failing transiently in CI, the root cause isn't a missing bicep CLI — what failure prompted this PR?

test/Makefile
```diff
@@ -6,6 +6,13 @@ all: build
...
 
+bicep-install:
+	@for i in 1 2 3 4 5; do \
+		az bicep install && break || \
```
avollmer-redhat, line 11: **Bug**: If all 5 attempts fail, the last command executed is `sleep 10`, which exits 0. Make treats the target as successful, and the real failure surfaces later as a confusing `az bicep build` error.

The repo's own `storage.sh` handles this correctly with an explicit `exit 1` after the loop:
```bash
for i in 1 2 3 4 5; do \
    az bicep install && break || \
    echo "..." && sleep 10; \
done; \
az bicep version >/dev/null 2>&1 || exit 1
```

test/Makefile
```diff
@@ -37,7 +44,7 @@ PROW_JOB_EXECUTOR = $(TEST_DIR)prow-job-executor
 $(PROW_JOB_EXECUTOR): $(PROW_JOB_EXECUTOR_SOURCES) $(MAKEFILE_LIST) $(TEST_DIR)/go.mod $(TEST_DIR)/go.sum
 	cd $(TEST_DIR) && go build -o $@ ./cmd/prow-job-executor
 
-build: $(PROW_JOB_EXECUTOR) $(ARO_HCP_TESTS)
+build: bicep-install $(PROW_JOB_EXECUTOR) $(ARO_HCP_TESTS)
 .PHONY: build
```
avollmer-redhat, line 49: **Bug**: `bicep-install` is a sibling prerequisite of `$(ARO_HCP_TESTS)` under `build`, but the actual `az bicep build` calls happen inside the pattern rules that `$(ARO_HCP_TESTS)` depends on. With `make -j`, those pattern rules can start concurrently with `bicep-install`.

The fix is to make the pattern rules depend on `bicep-install` directly, ideally as order-only prerequisites:
```makefile
$(BICEP_OUTPUT_DIR)standard-cluster-create/%.json: $(DEMO_BICEP_DIR)/%.bicep | bicep-install
$(BICEP_OUTPUT_DIR)%.json: $(TEST_BICEP_DIR)/%.bicep | bicep-install
```

avollmer-redhat (changes requested):
The CI image already bakes bicep into `/usr/local/bin` at build time (`dev-infrastructure/openshift-ci/Dockerfile:15`), so `az bicep install` at runtime is redundant in CI. Worth understanding the actual root cause before adding retry plumbing that may not address it.

If this is kept regardless, the two inline comments below are bugs that need fixing.

PR #6408, opened 2026-08-04
feat: add per-controller earliest recheck times on service provider cluster status

miguelsorianod:
Introduce ControllerEarliestRecheckTimes so Cosmos-backed controllers can persist when expensive recheck work may run again for a given cluster.

This gates SyncOnce after a reconcile is already enqueued. It does not replace changefeed delivery, informer resync, or ResyncPeriod.

### DIFF
--- a/internal/api/types_serviceprovider_cluster.go
+++ b/internal/api/types_serviceprovider_cluster.go
@@ -201,6 +201,24 @@ type ServiceProviderClusterStatus struct {
 	// AzureResources tracks the lifecycle of Azure resources associated with
 	// the cluster, including deny assignments and the managed resource group.
 	AzureResources AzureResources `json:"azureResources,omitempty"`
+
+	// ControllerEarliestRecheckTimes is a map of Cosmos controller names to their intended earliest recheck times.
+	// The key is the name of the controller. If a controller name is not present as a map key, it means recheck immediately.
+	// This gates work after a reconcile is already triggered. It does not replace or interact with changefeed delivery,
+	// informer resync, or controller ResyncPeriod: those still enqueue. This field decides whether SyncOnce should
+	// skip work until the stored time.
+	// The value is the earliest time at which the controller should re-check. Nil means recheck immediately.
+	// When interacting with this map, it can be nil. In that case, it means recheck immediately too. To add elements to it,
+	// it must be initialized first.
+	// This allows for controllers to avoid repeatedly doing the same work before a certain amount of time has passed.
+	// Controllers should set this field with substantial jitter: without another concern, jitter of 50% on top of the base
+	// recheck time is considered normal so that any storms are quickly dissipated. For example, if the base recheck time is 12 hours,
+	// the resulting range with 50% jitter would be between 12 and 18 hours.
+	// Additionally, long recheck times are recommended for controllers outside of their active phases. Order of at least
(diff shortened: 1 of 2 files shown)


internal/api/types_serviceprovider_cluster.go
```diff
@@ -201,6 +201,24 @@ type ServiceProviderClusterStatus struct {
...
+	// it must be initialized first.
+	// This allows for controllers to avoid repeatedly doing the same work before a certain amount of time has passed.
+	// Controllers should set this field with substantial jitter: without another concern, jitter of 50% on top of the base
+	// recheck time is considered normal so that any storms are quickly dissipated. For example, if the base recheck time is 12 hours,
+	// the resulting range with 50% jitter would be between 12 and 18 hours.
+	// Additionally, long recheck times are recommended for controllers outside of their active phases. Order of at least
+	// six hours is, with durations up to 24 hours considered normal.
+	// This gates work after a reconcile is triggered. It does not replace or interact with changefeed delivery,
+	// informer resync, or controller ResyncPeriod.
+	ControllerEarliestRecheckTimes map[string]*metav1.Time `json:"controllerEarliestRecheckTimes,omitempty"`
```
miguelsorianod, line 221: alternative name: ControllerScopedEarliestRecheckTimes

internal/api/types_serviceprovider_cluster.go
```diff
@@ -201,6 +201,24 @@ type ServiceProviderClusterStatus struct {
 	// AzureResources tracks the lifecycle of Azure resources associated with
 	// the cluster, including deny assignments and the managed resource group.
 	AzureResources AzureResources `json:"azureResources,omitempty"`
+
+	// ControllerEarliestRecheckTimes is a map of Cosmos controller names to their intended earliest recheck times.
```
miguelsorianod, line 205: Let me know if you see the doc well.

internal/api/types_serviceprovider_cluster.go
```diff
@@ -201,6 +201,24 @@ type ServiceProviderClusterStatus struct {
 	// AzureResources tracks the lifecycle of Azure resources associated with
 	// the cluster, including deny assignments and the managed resource group.
 	AzureResources AzureResources `json:"azureResources,omitempty"`
+
```
miguelsorianod, line 204: Placed in ServiceProviderCluster's .status (and not in spec, nor in the `Cluster` cosmos resource).

internal/api/types_serviceprovider_cluster.go
```diff
@@ -201,6 +201,24 @@ type ServiceProviderClusterStatus struct {
 	// AzureResources tracks the lifecycle of Azure resources associated with
 	// the cluster, including deny assignments and the managed resource group.
```
miguelsorianod, line 202: Once we decide if this goes in and the specific design for SPC I will replicate it for ServiceProviderNodePool

internal/api/types_serviceprovider_cluster.go
miguelsorianod, line 221: Should we initialize the map in `initialController` function? or leave it as empty? The consequence of leaving it as empty is that addition to it when is nil will panic. Which means controllers will need to always consider the map as potentially empty and initializing it if they want to add something

PR #6413, opened 2026-08-04
fix: remove OpenShift 4.19 install support and associated tests

deads2k:
4.19 is no longer a supported install version. This change:
- Makes 4.20 the minimum version for all clusters (removes the experimental-flag exception that allowed 4.19)
- Bumps cluster-service OCP version channel minVersions to 4.20
- Removes 4.19 from the operator versions to mirror
- Removes 4.19 from version-to-patch maps in OCM client and backend create operation
- Removes E2E tests that install 4.19 clusters (backlevel test, z-stream upgrade, nodepool downgrade to 4.19)
- Updates nodepool y-stream downgrade test from 4.21→4.19 to 4.22→4.20
- Removes integration test steps that create 4.19 resources
- Updates test fixture data from 4.19 to 4.20

Managed identity minOpenShiftVersion configs are intentionally left at 4.19 since existing clusters still need those identities.


config/rendered/dev/dev/westus3.yaml
```diff
@@ -475,7 +475,7 @@ imageSync:
       registry: arohcpsvcdev.azurecr.io
       repository: image-sync/oc-mirror
     jobNamePrefix: ""
-    operatorVersionsToMirror: 4.18,4.19,4.20,4.21,4.22
+    operatorVersionsToMirror: 4.18,4.20,4.21,4.22
```
bennerv: not sure why we're still mirroring 4.18 versions, want to remove them as well?

miguelsorianod: We still have two clusters in 4.19 that are pending deletion, currently stuck in the deletion process. A CS change has been just merged in the CS repository to attempt to unstuck that. Is the safest to wait until those are deleted? is the mirroring behavior "delete if it's not mirrored", or is it "stop mirroring but keep if it was already mirrored"?

cluster-service/deploy/templates/aro-hcp-ocp-versions-config.configmap.yaml
```diff
@@ -11,17 +11,17 @@ data:
     channelGroups:
       - url: https://api.openshift.com/api/upgrades_info/graph
         channelGroupName: stable
-        minVersion: 4.19.0
+        minVersion: 4.20.0
```
miguelsorianod, line 14: Can there be some issue with the two 4.19 pending existing clusters to delete (stuck in deleting, CS fix attempt merged in CS repo and pending to rollout) we remove 4.19 at the OCP versions config level on CS side? considering versions workers, DB entries...

cc @JameelB @ahitacat

config/rendered/dev/dev/westus3.yaml
bennerv: TL;DR: I think we're fine to remove the mirroring on these older images. 

It looks like the azure retention policy only applies to untagged manifests for images.  So since all the operator catalog images are pinned to the release tag (i.e. v4.19), then the latest one will still exist for existing clusters, but will never be updated after we remove them from mirroring (which imo is fine due to no longer supporting those versions). If there was an `az acr purge` task, then it would/could clean up images with tags, but there isn't one for the OCP ACR.  

Refs: 
- [Azure Image Purge policy ](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-retention-policy?tabs=azure-cli)
    - > Azure Container Registry gives you the option to set a retention policy for stored image manifests that don't have any associated tags. When you enable a retention policy, the registry automatically deletes untagged manifests after the number of days you set
- [ACR Purge jobs](https://github.com/Azure/ARO-HCP/blob/b91e1688ee434f53d60fba1043d741a6bd1719a8/dev-infrastructure/modules/acr/cache.bicep#L113-L152), showing that the [OCP ACR has no purge jobs configured](https://github.com/Azure/ARO-HCP/blob/main/dev-infrastructure/templates/global-acr.bicep).
    - the actual purge job appears to only be configured in the [svc ACR](https://github.com/Azure/ARO-HCP/blob/b91e1688ee434f53d60fba1043d741a6bd1719a8/dev-infrastructure/configurations/acr-svc.tmpl.bicepparam#L22-L45)without any references to the operator indices.  
- [OCP ACR Retention policy ](https://github.com/Azure/ARO-HCP/blob/b91e1688ee434f53d60fba1043d741a6bd1719a8/dev-infrastructure/modules/acr/acr.bicep#L54-L56)and it's [config](https://github.com/Azure/ARO-HCP/blob/b91e1688ee434f53d60fba1043d741a6bd1719a8/config/config.yaml#L101-L103):

cluster-service/deploy/templates/aro-hcp-ocp-versions-config.configmap.yaml
ahitacat: When 4.19.0 is removed in this config, CS set those version to disable, all 4.19.x. Meaning that no new clusters or nodepools can be created or updated to that version, but existing resources can be deleted. 

I think we are ok in the current situation.

JameelB, line 14: we're targeting a min version of 4.20.8, though we're selecting this now in the rp, it would be nice to disable versions below this on cs side.

config/config.yaml
```diff
@@ -1105,7 +1105,7 @@ defaults:
         repository: image-sync/oc-mirror
         digest: sha256:aa1a693a0d0d990392bc851153352a8d9180a1d9b1e501af22a13234af52a7d3 # cd55650 (2026-07-30 05:18)
       pullSecretName: ocmirror-pull-secret
-      operatorVersionsToMirror: "4.18,4.19,4.20,4.21,4.22"
+      operatorVersionsToMirror: "4.18,4.20,4.21,4.22"
```
JameelB: can we remove 4.18 from here as well?

internal/validation/validate_cluster.go
```diff
@@ -377,13 +377,7 @@ func validateVersionProfile(ctx context.Context, op operation.Operation, fldPath
 	if oldObj == nil || len(oldObj.ID) > 0 {
 		errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("id"), &newObj.ID, nil)...)
 
-		if !op.HasOption(api.FeatureExperimentalReleaseFeatures) {
```
JameelB: Is this pr focusing on cluster only? we also want to remove <4.20.8 installs for node pools. https://github.com/Azure/ARO-HCP/blob/main/internal/validation/validate_nodepools.go#L282-L288

test/e2e/cluster_version_backlevel.go
```diff
@@ -1,402 +0,0 @@
...
-	"github.com/Azure/ARO-HCP/test/util/verifiers"
-)
-
-var _ = Describe("Customer", func() {
-
-	backlevelVersions := []backlevelVersionSpec{
-		{
-			controlPlaneVersion: "4.19",
-			nodePoolVersions:    []string{"4.19.7"},
-			bicepModulesDir:     "test-artifacts/generated-test-artifacts/modules-4.19",
```
JameelB, line 45: The bicep templates could probably be removed as well? https://github.com/deads2k/ARO-HCP/tree/main/test/e2e-setup/bicep/modules-4.19

internal/validation/validate_cluster.go
deads2k: We don't care what version nodepools do.

config/config.yaml
deads2k: I don't know what it controls.  But if you want to remove it, do so in another PR.

cluster-service/deploy/templates/aro-hcp-ocp-versions-config.configmap.yaml
deads2k: Feel free to introduce a PR after this one. Don't keep growing a single PR that improves the state the world.

internal/validation/validate_cluster.go
JameelB: > We don't care what version nodepools do.

Do you mean we want to keep allowing installs of 4.19 as well for node pools? 

There was also a discussion awhile back around whether we wanted to also set a min version for node pools. At the time, an agreement was reached that we want to set this to 4.20.8. 

The main reason to move to 4.20.8 was due to a new [feature](https://github.com/openshift/hypershift/pull/6904) in 4.20 that allowed dynamic retrieval of the azure marketplace image based on the release which means we didn't need to have any hardcoded images in our repo. However, the release payload was only updated from [4.20.8](https://github.com/openshift/installer/blob/release-4.20/data/data/coreos/marketplace-rhcos.json) onwards (in previous versions these pointed to 4.19 images). 

If we don't set the min to 4.20.8, we'll have to either:
* keep the hardcoded azure marketplace images for versions for ocp versions <4.20.8 to keep the same behavior as we have now
* allow use of the dynamic retrieval feature for all 4.20 versions, which means node pools with 4.20.0 to 4.20.6 versions will use 4.19 azure marketplace images, but keep the hardcoded images for 4.19 (if we're keeping 4.19).

That decision was made quite awhile ago, i'm happy to reopen it if thoughts around this has changed.

JameelB: ah, i just remember this as well - we don't have the min version separated for cluster and node pools in CS 😞. If the minversion is changed in the cs config it'll disable the versions for both resource types. We'd allow it in RP but CS will return an error when you try to install np with it/upgrade a np to it.

config/config.yaml
JameelB: sounds good, will create a separate pr for this one 👍

cluster-service/deploy/templates/aro-hcp-ocp-versions-config.configmap.yaml
JameelB: sounds good, will create a separate pr for this one.

JameelB:
> /hold until some conversations around findings of some a cluster in 4.19 that is not deleting potentially still being tracked in our systems are clarified

@miguelsorianod we got confirmation that the sub is cancelled which means the resources belonging to it will remain until it gets automatically deleted after 90d. Considering that, do we still want to hold for this reason? According to @ahitacat and @bennerv's comments in the threads above, it shouldn't prevent a cluster from being deleted even if the version is disabled. However, do we want to keep the tests since a cluster with this version still remains? 

Other than that, it would be great to get clarification on https://github.com/Azure/ARO-HCP/pull/6413#discussion_r3720829587 before proceeding if possible 🙏

PR #6426, opened 2026-08-05
Upgrade to OCP 4.22 as default version in test framework.

miquelsi:
https://redhat.atlassian.net/browse/ARO-28872

### What

Upgrade to OCP 4.22 as default version in test framework.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/util/framework/deployment_params.go
+++ b/test/util/framework/deployment_params.go
@@ -43,7 +43,7 @@ const (
 
 	// Default OpenShift channel group and version for the E2E tests
 	DefaultOCPChannelGroup         = "candidate"
-	DefaultOCPVersionId            = "4.20"
+	DefaultOCPVersionId            = "4.22"
 	DefaultOCPNodePoolChannelGroup = "candidate"
 
 	DefaultPodCIDR      = "10.128.0.0/14"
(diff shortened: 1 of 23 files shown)


test/util/framework/deployment_params.go
```diff
@@ -38,7 +38,7 @@ const (
 
 	// Default OpenShift channel group and version for the E2E tests
 	DefaultOCPChannelGroup         = "candidate"
-	DefaultOCPVersionId            = "4.20"
+	DefaultOCPVersionId            = "4.22"
```
machi1990, line 46: Might be worth to update tests like https://github.com/miquelsi/ARO-HCP/blob/6b50c3468e09a3122dc0742801e466920e4b283f/test/e2e/kms_key_rotation.go#L59 to stop hardcoding 4.22 now that it is the default one, what do you think or you prefer the test be declarative of the version they want?

miquelsi: I would keep that one hardcoded, unless we know for sure that will be safe to upgrade to 4.23 (or 5.0 actually) when the time comes.

WDYT, @ziccardi?

machi1990: Yes it is safe. It is more like a minimum OCP version needed for those tests; Anything lower than that won't work

miquelsi:
Closing as we're updating to 4.21 instead. See https://redhat.atlassian.net/browse/ARO-28833

PR #6429, opened 2026-08-05
Improve VM helper to be able to stop commands.

miquelsi:
https://redhat.atlassian.net/browse/ARO-27425
### What

RunVMCommand now uses the managed VirtualMachineRunCommandsClient (BeginCreateOrUpdate / BeginDelete) instead of the legacy VirtualMachinesClient.BeginRunCommand action. Each invocation creates a named run-command resource, polls it to completion, reads results via GetByVirtualMachine with expand=instanceView, and deletes the resource afterward — including on timeout so a stuck command can be canceled and later commands are not blocked. The script source is set with Script only, which matches the managed Run Commands API (one of script / scriptUri / commandId), rather than the old RunShellScript + script action pattern.

### Why

This is required so e2e helpers can reliably stop hung VM commands and observe real exit codes, stdout, and stderr from instance view; the CreateOrUpdate LRO alone often omits that data. Because the helper now fails on non-zero exit codes, the authorized-CIDRs test’s APIServices and ClusterOperators checks append || true after grep -v ':True$', so a healthy empty filter (grep exit 1) is not treated as a VM command failure.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/e2e/cluster_authorized_cidrs_connectivity.go
+++ b/test/e2e/cluster_authorized_cidrs_connectivity.go
@@ -189,9 +189,11 @@ var _ = Describe("Authorized CIDRs", func() {
 				}, 5*time.Minute, 10*time.Second).Should(Succeed())
 
 				By("verifying aggregated API services from authorized VM")
-				// Only output unavailable services (filter out :True lines) to stay within 4KB VM output limit
+				// Filter :True lines on the VM to stay within the 4KB run-command output limit.
+				// Write kubectl output first so a kubectl failure fails the command. Isolate grep
+				// so only its rc=1 ("no matches", the healthy case) is treated as success.
 				apiServicesCmd := fmt.Sprintf(
-					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get apiservices -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' | grep -v ':True$'`,
+					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get apiservices -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' > /tmp/apiservices.status && { grep -v ':True$' /tmp/apiservices.status; rc=$?; [ $rc -le 1 ]; }`,
 					kubeconfigB64,
 				)
 
@@ -359,9 +361,11 @@ var _ = Describe("Authorized CIDRs", func() {
 				Expect(err).NotTo(HaveOccurred(), "failed to create console OAuth client secret for external auth via VM")
 
 				By("verifying all cluster operators are healthy from authorized VM")
-				// Only output unavailable operators (filter out :True lines) to stay within 4KB VM output limit
+				// Filter :True lines on the VM to stay within the 4KB run-command output limit.
+				// Write kubectl output first so a kubectl failure fails the command. Isolate grep
+				// so only its rc=1 ("no matches", the healthy case) is treated as success.
 				clusterOperatorsCmd := fmt.Sprintf(
-					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get clusteroperators -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' | grep -v ':True$'`,
+					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get clusteroperators -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' > /tmp/clusteroperators.status && { grep -v ':True$' /tmp/clusteroperators.status; rc=$?; [ $rc -le 1 ]; }`,
 					kubeconfigB64,
 				)
 
--- a/test/e2e/cluster_create_private_kas.go
+++ b/test/e2e/cluster_create_private_kas.go
@@ -126,8 +126,8 @@ var _ = Describe("Customer", func() {
 
 			By("verifying KAS is reachable from VM inside the VNet")
 			// Get admin credentials via ARM and override the server URL with
-			// the internal LB IP so kubectl on the VM connects through the
-			// private KAS endpoint.
+			// the internal LB's private IP so kubectl on the VM connects
+			// through the private KAS endpoint rather than the public route.
 			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
 				ctx,
 				tc.Get20240610ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient(),
@@ -142,6 +142,15 @@ var _ = Describe("Customer", func() {
 			GinkgoLogr.Info("Found private KAS internal LB", "ip", internalIP, "managedRG", clusterParams.ManagedResourceGroupName)
 
 			adminRESTConfig.Host = fmt.Sprintf("https://%s:443", internalIP)
+			// The admin kubeconfig's CAData does not correspond to the
+			// certificate served directly by the internal LB (it targets the
+			// cluster's public route/CA chain), and the LB IP is not in that
+			// certificate's SAN list either. This check only needs to prove
+			// network reachability of the private KAS endpoint, not validate
+			// the full TLS chain, so skip verification here — matching the
+			// insecureSkipVerify used by the other connectivity checks below.
+			adminRESTConfig.Insecure = true
+			adminRESTConfig.CAData = nil
 
 			kubeconfig, err := framework.GenerateKubeconfig(adminRESTConfig)
 			Expect(err).NotTo(HaveOccurred(), "failed to generate kubeconfig from admin REST config")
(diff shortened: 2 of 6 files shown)


test/util/framework/vm_helper.go
```diff
@@ -122,75 +122,153 @@ func RunVMCommand(ctx context.Context, tc interface {
...
+		return responseErr.StatusCode == http.StatusNotFound
+	}
+	return false
+}
+
+func toValue(value *string) string {
+	if value == nil {
+		return ""
+	}
+	return *value
```
mbukatov: Could you drop this helper and use [k8s.io/utils/ptr.Deref](https://pkg.go.dev/k8s.io/utils/ptr#Deref) instead?

test/util/framework/vm_helper.go
```diff
@@ -122,75 +122,153 @@ func RunVMCommand(ctx context.Context, tc interface {
...
+	return nil
+}
+
+func isNotFoundResponseError(err error) bool {
+	var responseErr *azcore.ResponseError
+	if errors.As(err, &responseErr) {
+		return responseErr.StatusCode == http.StatusNotFound
+	}
+	return false
+}
```
mbukatov: Could you drop this function, and use `isKeyVaultNotFound` from per_test_framework.go instead?

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -358,8 +360,10 @@ var _ = Describe("Authorized CIDRs", func() {
 
 				By("verifying all cluster operators are healthy from authorized VM")
 				// Only output unavailable operators (filter out :True lines) to stay within 4KB VM output limit
+				// grep exits 1 when every line is filtered out (the healthy case).
+				// Append || true so RunVMCommand's exit-code check does not treat that as failure.
```
mbukatov: I would rephrase this comment to make it more clear and readable, something like:

> we use ` || true` in the end of the command to not treat grep failure as an error (which is expected)

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -190,8 +190,10 @@ var _ = Describe("Authorized CIDRs", func() {
 
 				By("verifying aggregated API services from authorized VM")
 				// Only output unavailable services (filter out :True lines) to stay within 4KB VM output limit
+				// grep exits 1 when every line is filtered out (the healthy case).
+				// Append || true so RunVMCommand's exit-code check does not treat that as failure.
 				apiServicesCmd := fmt.Sprintf(
-					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get apiservices -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' | grep -v ':True$'`,
+					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get apiservices -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' | grep -v ':True$' || [ $? -eq 1 ]`,
```
mbukatov: Doesn't `|| [ $? -eq 1 ]` behave like `|| true` here?

Because it tried this construct out and this is what I see:

```
$ ls | grep je
je
martin@telesto ~/tmp
$ ls | grep neni
martin@telesto 1 ~/tmp
$ ls | grep je || [ $? -eq 1 ]
je
martin@telesto ~/tmp
$ ls | grep neni || [ $? -eq 1 ]
martin@telesto ~/tmp
```

By appending `|| [ $? -eq 1 ]`, both success and failure just retun 0 in the end. Exactly as `|| true`.

mbukatov:
It would have helped me if the Azure API function calls in the description were referenced, so that I don't have to look up [VirtualMachineRunCommandsClient](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/arm/compute#VirtualMachineRunCommandsClient). This starts to be obvious when you start to read code, but usually a PR description helps you to understand the code change, not vice versa.

test/util/framework/vm_helper.go
```diff
@@ -122,75 +122,153 @@ func RunVMCommand(ctx context.Context, tc interface {
...
+func deleteVirtualMachineRunCommandBestEffort(
+	runCommandsClient *armcompute.VirtualMachineRunCommandsClient,
+	resourceGroup string,
+	vmName string,
+	runCommandName string,
+) error {
+	deleteCtx, deleteCancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer deleteCancel()
+	return deleteVirtualMachineRunCommand(deleteCtx, runCommandsClient, resourceGroup, vmName, runCommandName)
+}
```
mbukatov, line 238: This function creates it's own background timeout context, so it can't be interrupted from the outside via higher context cancellation ... do we really want that? Do we even need this wrapper?

test/util/framework/vm_helper.go
mbukatov, line 115: This RunVMCommand function grows more than I would be comfortable with. Could we explore options how to write some unit tests for it? I'm thinking about changes like for example:

- don't pass tc into it, but pass subscriptionID and AzureCredential directly
- clientFactory from the outside (don't create it inside of the function)
- processing of instanceView could be moved into a separate function, and unit tested

mbukatov (commented):
I'm would like to make this code more robust, see comments below.

Also, have we tried to demonstrate that it works when it fails during E2E run?

test/util/framework/vm_helper.go
miquelsi: We're talking about cancelling a "delete VM command". If a higher context is cancelled, there is no harm in continuing with the command deletion.

test/e2e/cluster_create_private_kas.go
```diff
@@ -126,8 +126,8 @@ var _ = Describe("Customer", func() {
```
miquelsi, line 126: @zgalor can you review changes in this test, since you were the original author?

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -190,8 +190,10 @@ var _ = Describe("Authorized CIDRs", func() {
 
 				By("verifying aggregated API services from authorized VM")
 				// Only output unavailable services (filter out :True lines) to stay within 4KB VM output limit
+				// grep exits 1 when every line is filtered out (the healthy case).
+				// Append || true to not treat grep failure as an error (which is expected)
 				apiServicesCmd := fmt.Sprintf(
-					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get apiservices -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' | grep -v ':True$'`,
+					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get apiservices -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' | grep -v ':True$' || true`,
```
zgalor: `|| true` masks all failures (including kubectl errors, grep syntax errors — exit code 2), not just grep’s expected "no matches" exit code 1. This has been flagged across multiple review rounds. Note that @mbukatov also demonstrated that `|| [ $? -eq 1 ]` behaves identically to `|| true` in a pipeline, so that’s not a fix either.

Consider separating the concern — e.g. pipe through `cat` so grep’s exit code doesn’t propagate, or capture output and check emptiness instead of relying on grep’s exit code.

Same applies to the `clusterOperatorsCmd` on line 368.

test/util/framework/per_test_framework.go
```diff
@@ -829,10 +829,10 @@ func (tc *perItOrDescribeTestContext) purgeDeletedKeyVaultsInResourceGroup(ctx c
 	}
 }
 
-// isKeyVaultNotFound reports whether err is an Azure 404 response, which for a
+// IsNotFoundError reports whether err is an Azure 404 response, which for a
 // purge means the vault is already gone (already purged or soft-delete window
 // expired) and can be treated as a successful no-op.
```
zgalor: Now that this function is exported and used for VM run-command deletion too, the doc comment should be generalized — it still references Key Vault purge semantics specifically.

test/e2e/cluster_create_private_kas.go
zgalor: Changes in the test LGTM.
I added two additional review comments

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -359,9 +361,11 @@ var _ = Describe("Authorized CIDRs", func() {
 				Expect(err).NotTo(HaveOccurred(), "failed to create console OAuth client secret for external auth via VM")
 
 				By("verifying all cluster operators are healthy from authorized VM")
-				// Only output unavailable operators (filter out :True lines) to stay within 4KB VM output limit
+				// Filter :True lines on the VM to stay within the 4KB run-command output limit.
+				// Write kubectl output first so a kubectl failure fails the command; pipe grep
+				// through cat so grep's "no matches" (healthy) exit status does not.
 				clusterOperatorsCmd := fmt.Sprintf(
-					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get clusteroperators -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' | grep -v ':True$'`,
+					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get clusteroperators -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' > /tmp/clusteroperators.status && grep -v ':True$' /tmp/clusteroperators.status | cat`,
```
mvacula02: why not use the same `; rc=$?; [ $rc -le 1 ]` pattern used above in apiServicesCmd?

miquelsi: Sorry I forgot! Fixed now

mvacula02: thank you. Is the clusteroperators -> apiservices change on the line intentional 😅?

test/e2e/cluster_authorized_cidrs_connectivity.go
```diff
@@ -189,9 +189,11 @@ var _ = Describe("Authorized CIDRs", func() {
 				}, 5*time.Minute, 10*time.Second).Should(Succeed())
 
 				By("verifying aggregated API services from authorized VM")
-				// Only output unavailable services (filter out :True lines) to stay within 4KB VM output limit
+				// Filter :True lines on the VM to stay within the 4KB run-command output limit.
+				// Write kubectl output to a file first so a kubectl failure fails the command, then
+ 				// allow grep's rc=1 ("no matches", which is healthy) without failing the VM command.
 				apiServicesCmd := fmt.Sprintf(
-					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get apiservices -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' | grep -v ':True$'`,
+					`echo '%s' | base64 -d > /tmp/kubeconfig && kubectl --kubeconfig=/tmp/kubeconfig get apiservices -o jsonpath='{range .items[*]}{.metadata.name}:{.status.conditions[?(@.type=="Available")].status}{"\n"}{end}' > /tmp/apiservices.status && grep -v ':True$' /tmp/apiservices.status; rc=$?; [ $rc -le 1 ]`,
```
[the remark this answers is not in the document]
mvacula02: kubectl writes error messages to stderr, and output FromRunCommandInstanceView already fails when the Error field is non-empty. So kubectl would need to exit 1 without any stderr output to actually slip through

mvacula02 (approved):
LGTM, but let's wait for @mbukatov  or @zgalor  to approve as well before merging.

PR #6439, opened 2026-08-05
fix: ARO-29372: Enables continuous validation

patilsuraj767:
### What
Cluster and node pool validation controllers no longer stop after a Passed condition, and they no longer drop work in SyncOnce because of EarliestRetryAfter.

- Removed shouldProcess, which skipped validation once Status.Validations[<name>] was True.
- Removed the retryCooldownChecker.CanSync short-circuit from SyncOnce.
- Wired retryCooldownChecker through clusterWatchingController / nodePoolWatchingController CooldownChecker(), so it is only used by GenericWatchingController when changed == false (same Cosmos ETag / resync): https://github.com/Azure/ARO-HCP/blob/main/internal/controllerutils/generic_watching_controller.go#L247-L249
- ETag changes still enqueue immediately (changed=true skips CanSync), and SyncOnce always runs Validate(). Failed/Unknown still use EnqueueAfter. Passed still sets the 12h cooldown so unchanged 1-minute resyncs stay quiet.

**Tradeoff**: any Cluster/SPC (or NodePool) ETag change retriggers Validate(), including backend-only writes. Typical extra cost is one more Validate()


docs/cosmos-data-flow.md
```diff
@@ -1010,14 +1010,14 @@ No Cosmos writes. Posts `NodePoolUpgradePolicy` to Cluster Service.
...
-- SyncOnce also checks `DeletionTimestamp == nil` on the resource
+**Gate:**
+- SyncOnce checks `DeletionTimestamp == nil` on the resource
+- The validation always re-runs regardless of the previously stored condition
 
 | | Object | Fields |
 |---|--------|--------|
-| Read | `ServiceProviderCluster` | <ul><li>`Status.Validations[<name>]` (shouldProcess: condition must not be True)</li></ul> |
-| Read | `ServiceProviderNodePool` | <ul><li>`Status.Validations[<name>]` (shouldProcess: condition must not be True)</li></ul> |
+| Read | `ServiceProviderCluster` | <ul><li>`Status.Validations[<name>]` (used to compute consecutive-Unknown suppression, not to gate whether validation runs)</li></ul> |
```
miguelsorianod: Is this AI-generate change correct? it at the least doesn't seem very well explained

mbukatov:
Could you switch this to a draft state? I assume you are still working on it, and don't ask for a review yet. So this way it would be easier to filter out when checking PRs for a review.

patilsuraj767:
Added debug logs and verified the Kusto logs:
1. For clusters with a Passed outcome, the controller continues to run the validation as expected, respecting the cooldown checker gate.
2. There are no rate-limit issues specifically with CheckAccess. The token bucket is also working as expected.

PR #6456, opened 2026-08-06
CNTRLPLANE-4009: feat(test): add HyperShift presubmit e2e test and nightly version support

bryan-cox:
## Summary

- Adds `HypershiftPresubmit` label and `hypershift-presubmit/parallel` test suite for running ARO-HCP e2e tests triggered by openshift/hypershift PRs
- Adds a version-agnostic e2e test (`cluster_create_hypershift_presubmit.go`) that creates a cluster with CPO override, verifies viability, creates a nodepool, and runs a web app verification
- Adds `ARO_HCP_OPENSHIFT_VERSION_ID` env var to override the default OCP minor version (4.20) used for nightly release stream resolution — hypershift main targets OCP 5.0, so the presubmit CI needs to query `5.0.0-0.nightly-multi` instead of `4.20.0-0.nightly-multi`
- Excludes `HypershiftPresubmit` tests from integration/stage/prod suites — these tests require `CPO_IMAGE_OVERRIDE` and nightly channel group env vars only set in the presubmit job
- Makes the e2e suite configurable via `ARO_HCP_E2E_SUITE` env var (defaults to `rp-api-compat-all/parallel`) so the presubmit job can run `hypershift-presubmit/parallel`

## Why

[CNTRLPLANE-4009](https://redhat.atlassian.net/browse/CNTRLPLANE-4009) / [CNTRLPLANE-3980](https://redhat.atlassian.net/browse/CNTRLPLANE-3980)

We need a CI job on openshift/hypershift PRs that tests HO and CPO from the PR against an ARO-HCP environment. CPO is not backwards compatible — it needs a matching nightly release image from the same OCP development stream. The approach uses the existing nightly channel group support in Cluster Service and the test framework (`ARO_HCP_OPENSHIFT_CHANNEL_GROUP=nightly`), with the CPO override flowing through the existing ARM API experimental tag pipeline.

Companion PR: https://github.com/openshift/release/pull/83108 (CI job and step registry changes)

## Test plan

- [ ] Verify `go build ./util/framework/` compiles
- [ ] Verify `go test ./util/framework/` passes
- [ ] Verify the new test file builds with `go build -tags E2Etests ./e2e/...`
- [ ] Verify `TestMainListSuitesForEachSuite` passes — hypershift presubmit test only in `hypershift-presubmit/parallel`, not in integration/stage/prod
- [ ] Verify `make e2e-local/run` still defaults to `rp-api-compat-all/parallel`

### DIFF
--- a/test/cmd/aro-hcp-tests/main.go
+++ b/test/cmd/aro-hcp-tests/main.go
@@ -425,7 +425,7 @@ func setupCli() *cobra.Command {
 
 	// The tests that a suite is composed of can be filtered by CEL expressions. By
 	// default, the qualifiers only apply to tests from this extension.
-	integrationQuery := fmt.Sprintf(`labels.exists(l, l=="%s") && !labels.exists(l, l=="%s") && !labels.exists(l, l=="%s")`, labels.RequireNothing[0], labels.DevelopmentOnly[0], labels.StageAndProdOnly[0])
+	integrationQuery := fmt.Sprintf(`labels.exists(l, l=="%s") && !labels.exists(l, l=="%s") && !labels.exists(l, l=="%s") && !labels.exists(l, l=="%s")`, labels.RequireNothing[0], labels.DevelopmentOnly[0], labels.StageAndProdOnly[0], labels.HypershiftPresubmit[0])
 	integrationTestTimeout := 150 * time.Minute
 	ext.AddSuite(e.Suite{
 		Name: "integration/parallel",
@@ -451,7 +451,7 @@ func setupCli() *cobra.Command {
 		ResourcePools: miPools,
 	})
 
-	stageQuery := fmt.Sprintf(`labels.exists(l, l=="%s") && !labels.exists(l, l=="%s") && !labels.exists(l, l=="%s")`, labels.RequireNothing[0], labels.IntegrationOnly[0], labels.DevelopmentOnly[0])
+	stageQuery := fmt.Sprintf(`labels.exists(l, l=="%s") && !labels.exists(l, l=="%s") && !labels.exists(l, l=="%s") && !labels.exists(l, l=="%s")`, labels.RequireNothing[0], labels.IntegrationOnly[0], labels.DevelopmentOnly[0], labels.HypershiftPresubmit[0])
 	stageTestTimeout := 150 * time.Minute
 	ext.AddSuite(e.Suite{
 		Name: "stage/parallel",
@@ -476,7 +476,7 @@ func setupCli() *cobra.Command {
 		ResourcePools: miPools,
 	})
 
-	prodQuery := fmt.Sprintf(`labels.exists(l, l=="%s") && !labels.exists(l, l=="%s") && !labels.exists(l, l=="%s")`, labels.RequireNothing[0], labels.IntegrationOnly[0], labels.DevelopmentOnly[0])
+	prodQuery := fmt.Sprintf(`labels.exists(l, l=="%s") && !labels.exists(l, l=="%s") && !labels.exists(l, l=="%s") && !labels.exists(l, l=="%s")`, labels.RequireNothing[0], labels.IntegrationOnly[0], labels.DevelopmentOnly[0], labels.HypershiftPresubmit[0])
 	prodTestTimeout := 150 * time.Minute
 	ext.AddSuite(e.Suite{
 		Name: "prod/parallel",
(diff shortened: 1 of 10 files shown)


test/cmd/aro-hcp-tests/main.go
```diff
@@ -324,6 +324,16 @@ func setupCli() *cobra.Command {
 		ResourcePools: miPools,
 	})
 
+	hypershiftPresubmitQuery := fmt.Sprintf(`labels.exists(l, l=="%s")`, labels.HypershiftPresubmit[0])
+	hypershiftPresubmitTimeout := 150 * time.Minute
+	ext.AddSuite(e.Suite{
+		Name:          "hypershift-presubmit/parallel",
+		Qualifiers:    []string{hypershiftPresubmitQuery},
```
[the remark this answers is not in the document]
bryan-cox: Not making this change. Unlike the broad environment-based suites (integration, stage, prod) where slow tests can slip in via wide label criteria, this suite selects by an explicit opt-in label. Whoever tags a test with `HypershiftPresubmit` is deliberately choosing to include it in presubmit — if they tag a slow test, that's a conscious decision. Adding an empty `/slow` suite for a hypothetical scenario just adds noise.

---
*AI-assisted response via Claude Code*

test/util/framework/deployment_params.go
```diff
@@ -65,6 +65,9 @@ func resolveDefaultControlPlaneVersion() (string, error) {
 		version := os.Getenv("ARO_HCP_OPENSHIFT_CONTROLPLANE_VERSION")
 		if len(version) == 0 {
 			version = DefaultOCPVersionId
+			if v := os.Getenv("ARO_HCP_OPENSHIFT_VERSION_ID"); v != "" {
```
bennerv, line 68: Is `ARO_HCP-OPENSHIFT_CONTROLPLANE_VERSION` not sufficient?

test/e2e/cluster_create_hypershift_presubmit.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("ARO-HCP HyperShift Presubmit", func() {
+	It("should create a cluster and nodepool to completion",
+		labels.HypershiftPresubmit,
+		labels.RequireNothing,
+		labels.Critical,
+		labels.Positive,
+		labels.MIContainers(1),
```
bennerv, line 38: I think we need the local RP label here as well, so that it will support local RP instances.

test/e2e/cluster_create_hypershift_presubmit.go
```diff
@@ -0,0 +1,117 @@
...
+					"customerVnetName":       "customer-vnet-hs-pre-" + suffix,
+					"customerVnetSubnetName": "customer-vnet-subnet-hs-pre-" + suffix,
+				},
+				TestArtifactsFS,
+				framework.RBACScopeResourceGroup,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create customer resources for cluster %q", clusterName)
+
+			By(fmt.Sprintf("creating the HCP cluster with version '%s' on %s channel", clusterParams.OpenshiftVersionId, clusterParams.ChannelGroup))
+			err = tc.CreateHCPClusterFromParam20251223(
```
bennerv, line 74: Are we missing the CPO override tag? `TagClusterCPOImageOverride`

test/e2e/cluster_create_hypershift_presubmit.go
bennerv: If we don't want localRP instances, then the CPO override test is sufficient I think.  Need some validation.

test/util/framework/deployment_params.go
bryan-cox: Good question. `ARO_HCP_OPENSHIFT_CONTROLPLANE_VERSION` is meant for setting an *exact* version string (e.g. `4.20.7`) that bypasses channel resolution entirely. The new `ARO_HCP_OPENSHIFT_VERSION_ID` sets just the *minor* version seed (e.g. `5.0`) that feeds into `GetLatestInstallVersion()` for nightly channel resolution. In the companion openshift/release PR (#83108), the CI step sets `ARO_HCP_OPENSHIFT_VERSION_ID=5.0` with `ARO_HCP_OPENSHIFT_CHANNEL_GROUP=nightly` — the test framework then resolves `5.0.0-0.nightly-...` at runtime from the cluster service's install-version API.

We can't use `ARO_HCP_OPENSHIFT_CONTROLPLANE_VERSION` for this because it would skip resolution and `5.0` isn't a valid install version.

---
*AI-assisted response via Claude Code*

test/e2e/cluster_create_hypershift_presubmit.go
bryan-cox: Done. Added `labels.AroRpApiCompatible` so the test will also run in local RP suites.

---
*AI-assisted response via Claude Code*

test/e2e/cluster_create_hypershift_presubmit.go
bryan-cox: The CPO override is already applied — \`NewDefaultClusterParams20251223()\` calls \`applyCPOImageOverride(params.Tags)\` at \`helpers_v20251223preview.go:120\`, which reads \`CPO_IMAGE_OVERRIDE\` from the environment and sets \`TagClusterCPOImageOverride\` on the cluster tags. In CI, the aro-hcp-hypershift-images-push step writes the env var to \`SHARED_DIR/hypershift-cpo-override.env\` and the test-local step sources it before running the tests (see openshift/release PR #83108).

---
*AI-assisted response via Claude Code*

test/e2e/cluster_create_hypershift_presubmit.go
bryan-cox: Added the \`AroRpApiCompatible\` label so it supports both local RP and deployed RP instances. The CPO override is handled automatically via \`applyCPOImageOverride()\` in the default cluster params (see reply above).

---
*AI-assisted response via Claude Code*

PR #6463, opened 2026-08-06
fix: treat OCM4001 inflight check failure as non-terminal during cluster create

cssjr:
## Summary

- When Cluster Service (CS) reports `ClusterStateError` with error code `OCM4001` (inflight MI permission checks failed) during a **CREATE** operation, the backend now keeps the operation in `Provisioning` state instead of marking it as terminal `Failed`
- This allows the operation controller to continue polling CS, giving time for delayed role assignments to propagate through Azure RBAC
- The 60-minute `CreateOperationCompletionDeadline` provides a hard backstop if CS never recovers
- Scoped to CREATE only — UPDATE and DELETE operations continue to treat OCM4001 as terminal `Failed`

## Root Cause ([ARO-28755](https://redhat.atlassian.net/browse/ARO-28755))

The E2E test `cluster_delayed_role_assignments.go` fails ~7% of the time because of a timing race: when Managed Identity role assignments are deployed after cluster creation begins, CS's inflight MI permission checks may exhaust their retry window before the role assignments propagate through Azure RBAC. Once CS reports `ClusterStateError`, the backend unconditionally mapped this to terminal `ProvisioningStateFailed`, preventing any recovery even after role assignments were in place.

## Test plan

- [x] Unit tests: 3 new test cases in `TestConvertClusterStatus` covering OCM4001 during CREATE (stays Provisioning), OCM4001 during UPDATE (stays Failed), and non-OCM4001 during CREATE (stays Failed)
- [x] All existing backend tests pass with no regressions
- [ ] CI E2E `prod/parallel` suite validates the delayed role assignment test passes

Resolves: https://redhat.atlassian.net/browse/ARO-28755

### DIFF
--- a/backend/pkg/utils/operationutils/utils.go
+++ b/backend/pkg/utils/operationutils/utils.go
@@ -457,6 +457,12 @@ func ConvertClusterStatus(ctx context.Context, clusterServiceClient ocm.ClusterS
 			if err != nil {
 				return newOperationStatus, opError, err
 			}
+			// During CREATE, keep the operation non-terminal so the controller
+			// continues polling CS. CS may recover once role assignments propagate.
+			// The CreateOperationCompletionDeadline (60 min) provides a backstop.
+			if operation.Request == cosmosstorageutils.OperationRequestCreate {
+				newOperationStatus = arm.ProvisioningStateProvisioning
+			}
 		default:
 			opError = &arm.CloudErrorBody{Code: code, Message: message}
 		}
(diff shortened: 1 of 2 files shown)


miguelsorianod:
/hold

> When Cluster Service (CS) reports ClusterStateError with error code OCM4001 (inflight MI permission checks failed) during a CREATE operation, the backend now keeps the operation in Provisioning state instead of marking it as terminal Failed

This statement seems it might not be correct, and therefore the described issue is not correct and the code changes would not change any behavior. In ARO-HCP, CS permanently retries without marking the inflight as failed.

Discussion around this continues through other channels...

PR #6475, opened 2026-08-07
test: remove expired APIService verifier retry deadline (AROSLSRE-1443)

raelga:
Jira: [AROSLSRE-1443](https://redhat.atlassian.net/browse/AROSLSRE-1443)

### What

Removes the expired dead-man's-switch in `VerifyAllAPIServicesAvailable` (`test/util/verifiers/apiservice.go`) that hard-fails a test as soon as the aggregated `APIService` poll takes more than one attempt, once the calendar passes Jan 15 2026. Also drops the now-unused `attempts` counter.

### Why

The guard was added in https://github.com/Azure/ARO-HCP/pull/3650 as a temporary forcing function, betting HyperShift would fix its aggregated-API readiness contract by Jan 15 2026. That deadline passed 7 months ago and HyperShift still doesn't wait for aggregated APIs before marking a HostedCluster ready, so this guard now converts transient control-plane slowness into a hard test failure on `main`, acting as a fleet-wide flake amplifier on any slow/degraded-infra day. The bounded 5 minute exponential backoff retry loop already handles the intended tolerance on its own, so removing the deadline check restores that behavior without papering over anything new.

### Testing

`go build ./util/verifiers/...` and `go vet ./util/verifiers/...` pass in `test/`. No test behavior changes for fast/healthy clusters (first-poll success); slow-but-healthy clusters now pass within the existing retry window instead of hard-failing.

Pure deletion, no new logic added. Confirmed the `attempts` variable was otherwise unused after removing the deadline check.

### DIFF
--- a/test/util/verifiers/apiservice.go
+++ b/test/util/verifiers/apiservice.go
@@ -46,7 +46,6 @@ func (v verifyAllAPIServicesAvailableImpl) Verify(ctx context.Context, adminREST
 	retryCtx, cancel := context.WithTimeoutCause(ctx, 5*time.Minute, errors.New("API service retry timeout"))
 	defer cancel()
 
-	attempts := 0
 	var lastErr error
 	verifyErr := wait.ExponentialBackoffWithContext(retryCtx, wait.Backoff{
 		Duration: 800 * time.Millisecond,
@@ -55,8 +54,6 @@ func (v verifyAllAPIServicesAvailableImpl) Verify(ctx context.Context, adminREST
 		Steps:    10,
 		Cap:      20 * time.Second,
 	}, func(ctx context.Context) (done bool, err error) {
-		attempts++
-
 		allAPIServices, err := apiserviceClient.APIServices().List(ctx, metav1.ListOptions{})
 		if err != nil {
 			return false, fmt.Errorf("failed to list all APIServices: %w", err)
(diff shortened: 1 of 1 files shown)


test/util/verifiers/apiservice.go
```diff
@@ -99,15 +96,6 @@ func (v verifyAllAPIServicesAvailableImpl) Verify(ctx context.Context, adminREST
...
 	})
 
-	deadline, err := time.Parse(time.RFC3339, "2026-01-15T00:00:00Z")
-	if err != nil {
-		return fmt.Errorf("failed to parse deadline: %w", err)
-	}
-
-	if attempts > 1 && time.Now().After(deadline) {
-		return errors.New("the APIService verifier is not allowed to re-try after Jan 15th, 2026 - HyperShift should only post ready status once aggregated APIs are ready then")
-	}
```
mbukatov, line 109: Dropping this if clause (when we used more attempts after the deadline) will make us miss cases where we used more retries instead of failing outright.

~Shouldn't we remove the whole retry logic as well then? Or should we just drop the deadline logic and keep the attempts counter?~

The reasoning behind this change states that:

> HyperShift still doesn't wait for aggregated APIs before marking a HostedCluster ready

So it boils down into a decision whether we still consider this a bug to be fixed or not.

We should ask @stevekuznetsov and @geoberle about that.

If we agree that this won't be fixed, it would be good to reference the hypershift bug/issue/task.

mgahagan73:
If we are no longer expecting this to be one-shot, should we log the number of attempts we took as well as the total time it took? This would give us 2 metrics we can track that would be useful in debugging future issues.

janboll:
Potential fix was merged, let's see: https://github.com/openshift/hypershift/pull/7434

test/util/verifiers/apiservice.go
stevekuznetsov: >So it boils down into a decision whether we still consider this a bug to be fixed or not.

Yes, 100% correct. If HyperShift truly does not wait for aggregated APIs, that is a bug we need to report against them.

PR #6477, opened 2026-08-07
fix: drop apiserver_request_total watch/connect verb series from SRE metric set

janboll:
## Summary

- Drop `apiserver_request_total` timeseries with `verb=WATCH|CONNECT` from the SRE metric set
- These verbs account for ~29% of raw `apiserver_request_total` timeseries (~76,700 out of 261,732 across 44 HCPs) but are never consumed by any recording rule or alert — all upstream rules explicitly filter to `verb=~"LIST|GET|POST|PUT|PATCH|DELETE"`
- Fix verb regex case: Kubernetes exports verb labels in uppercase, so the original `(watch|connect)` regex never matched — changed to `(WATCH|CONNECT)`
- Add an E2E test that parses the configmap template, extracts all drop rules, and queries Azure Monitor Workspace to verify each dropped series is absent (using `up{job=...}` as a positive control to confirm metrics are flowing for all six components)

## Test plan

- [x] Verify helm fixture files are consistent (`cd config && make materialize`)
- [x] Lint passes (`make lint`)
- [x] Test compiles (`go vet -tags E2Etests ./e2e/...`)
- [ ] Confirm no recording rule or alert references `verb="watch"` or `verb="connect"` for `apiserver_request_total`
- [ ] After rollout, validate timeseries count reduction in Grafana
- [ ] Run E2E test in dev: `make -C test -f E2ELocal.mk e2e-local/run-test TEST_NAME="SRE Metrics Set"`

References: https://redhat.atlassian.net/browse/AROSLSRE-1550


janboll:
/hold
need to think about this again, simply discarding by label can be confusing

PR #6480, opened 2026-08-07
fix(e2e): add pre-flight API version availability checks to preview API tests

gmfrasca:
### What

Adds `tc.IsHCPAPIVersionAvailable()` to the E2E test framework and uses it in all five preview API tests to skip early when required ARM resource types don't yet support the target API version. Also retains existing `isAPINotDeployedError` guards after cluster/nodepool creation as defense-in-depth.

### Why

The FIPS test was failing in INT because `v20260630preview` was partially deployed -- `hcpOpenShiftClusters` accepted the API version but `locations/hcpOperationStatuses` did not. The SDK polls operation status using the same API version, and ARM rejects it with `NoRegisteredProviderFound`. The existing pre-flight check only validated `hcpOpenShiftClusters`, so the test proceeded to create resources before failing mid-create.

### Testing

- No new test cases; this fixes skip logic in existing E2E tests
- Verified resource type names against `az provider show --namespace Microsoft.RedHatOpenShift`
- `go vet` and lint pass

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

- [x] E2E tests follow Principles of Good E2E Test Case Design

### DIFF
--- a/test/e2e/cluster_fips_mode.go
+++ b/test/e2e/cluster_fips_mode.go
@@ -49,33 +49,13 @@ var _ = Describe("FIPS Mode Support", func() {
 				tc := framework.NewTestContext()
 
 				By("checking API version availability")
-				if !framework.IsDevelopmentEnvironment() {
-					resourcesFactory, err := tc.GetARMResourcesClientFactory(ctx)
-					Expect(err).NotTo(HaveOccurred(), "failed to get ARM resources client factory")
-
-					providersClient := resourcesFactory.NewProvidersClient()
-					provider, err := providersClient.Get(ctx, "Microsoft.RedHatOpenShift", nil)
-					Expect(err).NotTo(HaveOccurred(), "failed to get Microsoft.RedHatOpenShift resource provider")
-
-					available := false
-					for _, rt := range provider.ResourceTypes {
-						if rt.ResourceType == nil || !strings.EqualFold(*rt.ResourceType, "hcpOpenShiftClusters") {
-							continue
-						}
-						for _, v := range rt.APIVersions {
-							if v != nil && strings.EqualFold(*v, apiVersion) {
-								available = true
-								break
-							}
-						}
+				apiAvailable, err := tc.IsHCPAPIVersionAvailable(ctx, apiVersion)
+				Expect(err).NotTo(HaveOccurred(), "failed to check API version availability")
+				if !apiAvailable {
+					if time.Now().After(framework.V20260630PreviewDeploymentDeadline) {
+						Fail(fmt.Sprintf("API version %s should be fully available by %s", apiVersion, framework.V20260630PreviewDeploymentDeadline.Format(time.RFC3339)))
 					}
-					if !available {
-						if time.Now().After(framework.V20260630PreviewDeploymentDeadline) {
-							Fail(fmt.Sprintf("API version %s should be available for Microsoft.RedHatOpenShift/hcpOpenShiftClusters by %s", apiVersion, framework.V20260630PreviewDeploymentDeadline.Format(time.RFC3339)))
-						}
-						Skip(fmt.Sprintf("API version %s is not available for Microsoft.RedHatOpenShift/hcpOpenShiftClusters in this environment", apiVersion))
-					}
-					GinkgoLogr.Info("API version available", "version", apiVersion)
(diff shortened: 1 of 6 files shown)


bennerv:
Can we put a helper function in the test helpers which validates for a given api version, that all resource types we need are required?  

That should fix the issue, and as well as futureproof us when we include it.

bennerv:
FYI, API release should be completed in ~4 hrs for us to use.

gmfrasca:
> Can we put a helper function in the test helpers which validates for a given api version, that all resource types we need are required?
> 
> That should fix the issue, and as well as futureproof us when we include it.

Done - extracted `framework.IsAPIVersionAvailable(provider, apiVersion)` into `hcp_helper.go`. It checks `hcpOpenShiftClusters`, `hcpOperationStatuses`, and `hcpOperationResults` and returns the first missing resource type. The FIPS test uses it now, and the other `v2026*` tests can adopt it as a follow-up.

test/util/framework/hcp_helper.go
```diff
@@ -286,3 +288,38 @@ func HasNodeTaint(nodes []corev1.Node, key, value string, effect corev1.TaintEff
...
 }
+
+// requiredResourceTypesForAPIVersion lists the ARM resource types that must all
+// support a given API version for async operations (create/delete/update) to work
+// end-to-end. The generated SDK uses the same api-version query parameter for both
+// the resource and its operation status polling endpoint.
+var requiredResourceTypesForAPIVersion = []string{
+	"hcpOpenShiftClusters",
+	"hcpOperationStatuses",
+	"hcpOperationResults",
```
bennerv: Can we confirm these resource types are correct?

```
az provider show --namespace microsoft.redhatopenshift  | jq -r .resourceTypes[].resourceType
OpenShiftClusters
locations/openshiftversions
locations/platformworkloadidentityrolesets
locations
locations/operationresults
locations/operationsstatus
operations
HCPOpenShiftClusters
HCPOpenShiftClusters/NodePools
HCPOpenShiftClusters/ExternalAuths
locations/hcpOpenShiftVersions
locations/hcpOperatorIdentityRoleSets
locations/hcpOperationResults
locations/hcpOperationStatuses
```

test/util/framework/hcp_helper.go
```diff
@@ -286,3 +288,38 @@ func HasNodeTaint(nodes []corev1.Node, key, value string, effect corev1.TaintEff
...
+}
+
+// IsAPIVersionAvailable checks whether apiVersion is registered for all resource
+// types required by the ARO-HCP SDK (clusters, operation statuses, and operation
+// results). It returns ("", true) when all types support the version, or
+// (missingType, false) identifying the first type that does not.
+func IsAPIVersionAvailable(provider armresources.Provider, apiVersion string) (string, bool) {
```
bennerv: can we use this function everywhere we validate API versions?  And change the signature to look something like:

```
func (tc *perItOrDescribeTestContext) IsHCPAPIVersionAvailable(ctx context.Context, apiVersion string) (bool, error) {
```

bennerv: Can we also update all tests that use preview API versions with timebombs to use this method instead of what they're doing now?

test/util/framework/hcp_helper.go
gmfrasca: Good catch - updated to `locations/hcpOperationStatuses` and `locations/hcpOperationResults` to match the ARM manifest. Skipping the other types (`NodePools`, `ExternalAuths`, etc.) since they're child/list-only resources that deploy alongside `hcpOpenShiftClusters`.

test/util/framework/hcp_helper.go
gmfrasca: Yep, added in a second commit -- all five preview API tests now use `tc.IsHCPAPIVersionAvailable()`

bennerv:
/approve

needs a `make lint` otherwise lgtm

test/e2e/cluster_fips_mode.go
```diff
@@ -117,6 +97,12 @@ var _ = Describe("FIPS Mode Support", func() {
 					clusterResource,
 					framework.ClusterCreationTimeout,
 				)
+				if isAPINotDeployedError(err) {
```
bennerv, line 100: nit: probably don't need this anymore but it's fine.

PR #6486, opened 2026-08-10
Revert "feat(kusto): add entity-groups pipeline step and retain deployer Admin"

janboll:
Reverts Azure/ARO-HCP#6294


mmazur:
The original change is causing int deployment failures. Force merging the revert to unblock deployments.
More info: https://redhat.atlassian.net/browse/AROSLSRE-1783

PR #6057, opened 2026-07-13
refactor: replace error-based validation with structured ValidationResult

deads2k:
Validation controllers now return *ValidationResult (Passed/Failed/Unknown)
instead of error, giving each outcome explicit retry timing, user-facing
messages, and service-provider diagnostics.
    
Key changes:
- Validate() returns *ValidationResult with Outcome, Failed/Unknown details,
  and EarliestRetryAfter
- SettableCooldownChecker enforces per-key retry timing inside SyncOnce
- AfterEnqueuer exposes workqueue AddAfter for explicit requeue scheduling
- Unknown results preserve the previous stored condition for up to 10
  consecutive attempts before overwriting
- DeepEqual guard skips Cosmos Replace when the condition hasn't changed
- DefaultResult and BuildCondition extracted as shared helpers

PR #6487, opened 2026-08-10
ARO-28729 No preflight validation that Key Vault is accessible for et…

YuriyNik:
[Link to Jira: https://issues.redhat.com/browse/ARO-28729]

### What

Add AzureClusterKeyVaultAccessibilityValidation backend validation controller that validates the customer's Key Vault is accessible to the cluster's KMS operator managed identity. It:

  - Resolves the kms operator identity from the cluster's ControlPlaneOperators map and authenticates as that identity via the Managed Identities Data Plane — the same credential path etcd itself uses for encrypt/decrypt
  operations.
  - Calls GetKey on the customer's Key Vault to confirm the key is reachable.
  - No-ops for clusters not using customer-managed KMS encryption.
  - Reports KeyVaultNotAccessible on failure with an actionable error, AsExpected on success.

  Also:
  - Adds KeyVaultKeysClient interface and KeyVaultKeysClient() method to ServiceManagedIdentityClientBuilder for data-plane access to customer-owned Key Vaults.
  - Refactors credentialsForServiceManagedIdentity into a generic credentialForIdentity that works for any identity the cluster has been granted (SMI or cluster operator identity like "kms").

### Why

  There is currently no preflight check that the etcd KMS Key Vault is reachable. Misconfigurations surface late and opaquely rather than being caught early with a clear, actionable error.

  [Link to Jira: https://issues.redhat.com/browse/ARO-28729]

### Testing

justification.

- Unit tests

  This PR targets the current plain-`error`-based validation interface, so it's
  self-contained and ready to merge independently. I also have a version
  adapted to the new structured `ValidationResult`/`Outcome` type from #6057,
  kept on a separate local branch — not included here since it depends on
  types not yet in `main`. I'll follow up with that once #6057 merges.

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
```diff
@@ -0,0 +1,118 @@
...
+		azureFPAClientBuilder: azureFPAClientBuilder,
+	}
+}
+
+func (a *AzureClusterKeyVaultAccessibilityValidation) Name() string {
+	return "AzureClusterKeyVaultAccessibilityValidation"
+}
+
+func (a *AzureClusterKeyVaultAccessibilityValidation) Validate(
+	ctx context.Context, clusterSubscription *arm.Subscription, cluster *api.HCPOpenShiftCluster,
```
miguelsorianod: nitpick: We accept considerably long lines in the project. Put in the same line

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
```diff
@@ -0,0 +1,118 @@
...
+) error {
+	kmsProfile := customerManagedKmsProfile(cluster)
+	if kmsProfile == nil {
+		return nil
+	}
+
+	vaultsClient, err := a.azureFPAClientBuilder.KeyVaultVaultsClient(
+		*clusterSubscription.Properties.TenantId,
```
miguelsorianod: Same as https://github.com/Azure/ARO-HCP/pull/6487/changes#r3749960151

backend/pkg/azure/client/fpa_client_builder.go
```diff
@@ -103,6 +105,15 @@ func (b *firstPartyApplicationClientBuilder) UsageClient(tenantID string, subscr
 	return armcompute.NewUsageClient(subscriptionID, creds, b.options)
 }
 
+func (b *firstPartyApplicationClientBuilder) KeyVaultVaultsClient(tenantID string, subscriptionID string) (KeyVaultVaultsClient, error) {
```
miguelsorianod: This is not correct. The FPA does not have access to resources created by the end-user (this is, resources outside of the cluster's managed resource group)

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
```diff
@@ -0,0 +1,118 @@
...
+		cluster.ID.SubscriptionID,
+	)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get key vault client: %w", err))
+	}
+
+	resp, err := vaultsClient.Get(ctx, cluster.ID.ResourceGroupName, kmsProfile.ActiveKey.VaultName, nil)
```
miguelsorianod: The resourcegroup name part is not correct: As of now, there's no enforcement that requires the key vault created by the user lives in the same RG than the Cluster's RG.

What makes this even worse: They do not provide the RG name. Only the vault name. However, this requires the resourcegroup name. Starting to require that information would require a breaking change in the API, which makes this non trivial.

An alternative would be to list all the key vaults in the subcription and then filter client-side by the URL. However that is considerably costly as we don't control the number of them, as well as permissions would need to be granted to list (as well as read)

backend/pkg/azure/client/fpa_client_builder.go
miguelsorianod: This also means that whatever identity is used, it will have to have the permissions to do what it needs against the resource. Which means that some new requirements for the end-users regarding permissions addition might be needed, which would require altering pre-existing resources / permissions too

nimrodshn:
cc @ahitacat

nimrodshn:
@YuriyNik As a general comment can we change the PR description to _describe the changes introduce rather than the issue that its trying to solve_ eg : **"Add AzureClusterKeyVaultAccessibilityValidation controller to validate clients KeyVault"**.

WDYT?

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
YuriyNik: ok, fixed

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+func (a *AzureClusterKeyVaultAccessibilityValidation) Name() string {
+	return "AzureClusterKeyVaultAccessibilityValidation"
+}
+
+func (a *AzureClusterKeyVaultAccessibilityValidation) Validate(ctx context.Context, _ *coreapi.Subscription, cluster *coreapi.HCPOpenShiftCluster) ValidationResult {
+	if cluster.CustomerProperties.Etcd.DataEncryption.KeyManagementMode != metadataapi.EtcdDataEncryptionKeyManagementModeTypeCustomerManaged {
+		return PassedValidation(coreapi.ControllerConditionReasonAsExpected, "As expected", "Cluster does not use customer-managed KMS encryption.")
```
miguelsorianod: This should be Skipped and not passed

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
```diff
@@ -0,0 +1,108 @@
...
+		return UnknownValidation(
+			"InternalError",
+			"Unable to verify Key Vault accessibility.",
+			"customer-managed key management mode is set but customer-managed encryption profile is nil",
+			ControllerReportingPolicyTypeError,
+		)
+	}
+
+	if cm.EncryptionType != metadataapi.CustomerManagedEncryptionTypeKMS {
+		return PassedValidation(coreapi.ControllerConditionReasonAsExpected, "As expected", "Cluster does not use KMS encryption.")
```
miguelsorianod: This should be Skipped and not Passed

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
```diff
@@ -0,0 +1,108 @@
...
+			ControllerReportingPolicyTypeError,
+		)
+	}
+
+	kmsProfile := cm.Kms
+
+	kmsIdentityResourceID := cluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ControlPlaneOperators[string(internalazure.ClusterOperatorIdentifierKMS)]
+	if kmsIdentityResourceID == nil {
+		internalAndUserMsg := fmt.Sprintf("cluster has no %q operator identity configured", internalazure.ClusterOperatorIdentifierKMS)
+		return FailedValidation("KmsIdentityNotConfigured", internalAndUserMsg, internalAndUserMsg)
```
miguelsorianod: This seems incorrect. kmsIdentityResourceID should never be nil if the etcd dataencryption has been configured as customer managed with kms encryption

miguelsorianod: Additionally, doublecheck if the RP Frontend is validating that. If not, that means we miss the validation there and it should be added. On CS side this validation should be there already (you can also doublecheck)

backend/pkg/azure/client/fpa_client_builder.go
YuriyNik: The implementation no longer uses the FPA.
  It now authenticates as the cluster's KMS operator UAMI (ControlPlaneOperators["kms"]) through a dedicated ClusterOperatorIdentityClientBuilder.KeyVaultKeysClient, and calls the Key Vault data plane (azkeys.Client.GetKey) to verify access.
  This is the same identity that etcd itself uses for encrypt/decrypt, and the customer already grants it the Key Vault Crypto User role as part of standard cluster setup — so there are no new permission requirements for end-users.
  This addresses both points:
  - No FPA access to customer resources — we use the cluster's own KMS identity.
  - No new permission requirements — that identity already has the needed RBAC.

backend/pkg/azure/client/smi_client_builder.go
```diff
@@ -63,42 +70,57 @@ func (b *serviceManagedIdentityClientBuilder) BuilderType() ServiceManagedIdenti
...
+}
+
+func (b *serviceManagedIdentityClientBuilder) KeyVaultKeysClient(ctx context.Context, clusterIdentityURL string, identityResourceID *azcorearm.ResourceID, vaultName string) (KeyVaultKeysClient, error) {
+	creds, err := b.credentialForIdentity(ctx, clusterIdentityURL, identityResourceID)
+	if err != nil {
+		return nil, err
 	}
 
-	return dataplane.GetCredential(b.azCoreARMClientOptions.ClientOptions, resp.ExplicitIdentities[0])
+	vaultURL := fmt.Sprintf("https://%s.vault.azure.net/", vaultName)
```
YuriyNik: The KeyVaultKeysClient currently hardcodes the .vault.azure.net suffix, which is Azure Public-only. Sovereign clouds use different suffixes (.vault.usgovcloudapi.net for Gov, .vault.azure.cn for China).

  Since ARO-HCP currently only deploys to AzurePublicCloud, is it acceptable to keep this as-is for now (with a TODO/follow-up issue for sovereign cloud support), or should I add a keyVaultDNSSuffix property to the cloud
  environment config in this PR?

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
YuriyNik: The new implementation uses the Key Vault data plane API (azkeys.Client.GetKey) instead of the ARM API.
The data plane is addressed by vault DNS name only (https://{vaultName}.vault.azure.net/) — no resource group needed. This also aligns with how etcd itself accesses the vault at runtime.

PS: probably we need to add url as a parameter of configuration not a hardcoded one - i added separate comment about this.

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
YuriyNik: Checked internal/admission/ — didn't find validation requiring KMS identity when CustomerManaged+KMS is selected. This check catches that gap. Should I add RP validation in this PR or as follow-up?

backend/pkg/azure/client/smi_client_builder.go
```diff
@@ -50,6 +51,12 @@ type ServiceManagedIdentityClientBuilder interface {
 	UserAssignedIdentitiesClient(ctx context.Context, clusterIdentityURL string, smiResourceID *azcorearm.ResourceID, subscriptionID string) (UserAssignedIdentitiesClient, error)
 	// SubnetsClient returns a new Subnet client.
 	SubnetsClient(ctx context.Context, clusterIdentityURL string, smiResourceID *azcorearm.ResourceID, subscriptionID string) (SubnetsClient, error)
+	// KeyVaultKeysClient returns a new Key Vault Keys data plane client,
+	// authenticated as the given customer-granted managed identity (e.g. a
+	// cluster operator identity such as "kms") rather than as the SMI. This
+	// is the only credential path that has RBAC access to a customer-owned
+	// Key Vault; the FPA and SMI do not.
+	KeyVaultKeysClient(ctx context.Context, clusterIdentityURL string, identityResourceID *azcorearm.ResourceID, vaultName string) (KeyVaultKeysClient, error)
```
miguelsorianod: No this design is not correct. SMIClientBuilder is **exclusively** used to authenticate as the SMI identity.
Its purpose is to allow you to obtain clients that are authenticated as the SMI identity exclusively.

This is explicitly described in the type documentation:
```
// ServiceManagedIdentityClientBuilder offers the ability to create Azure clients
// authenticating as the Cluster's Service Managed Identity, which is
// a cluster-scoped identity.
```

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return UnknownValidation(
+			"InternalError",
+			"Unable to verify Key Vault accessibility.",
+			fmt.Sprintf("failed to get key vault client: %s", err),
+			ControllerReportingPolicyTypeError,
+		)
+	}
+
+	if _, err := keysClient.GetKey(ctx, kmsProfile.ActiveKey.Name, kmsProfile.ActiveKey.Version, nil); err != nil {
```
miguelsorianod, line 95: This is not really verifying whether keyvault is accessible or not. It's verifying whether it can get the key. You can have the keyvault completely accessible but that this returns an error because the key does not exist

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
miguelsorianod: > Checked internal/admission/

That's not the only place where validation is performed in the frontend

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
YuriyNik: ok, fixed - azure_cluster_keyvault_accessibility_validation.go:54 — non-customer-managed → SkippedValidation("NotApplicable", ...)

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
YuriyNik: ok, its fixed -   - azure_cluster_keyvault_accessibility_validation.go:64 — non-KMS encryption type → SkippedValidation("NotApplicable", ...)

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
YuriyNik: I added the missing validation to the RP frontend: validateKmsIdentityRequirement in internal/validation/validate_cluster.go (wired into ValidateCluster). When customer-managed KMS encryption is configured but the kms operator identity is absent, it now returns a field.Required error — so a create/update request can no longer reach the backend in that state. (The earlier check was only in internal/admission/, which isn't the validation path you were referring to.)
  Given that guarantee, kmsIdentityResourceID should indeed never be nil here. So instead of a user-facing FailedValidation, the backend now treats a nil as an internal inconsistency and returns UnknownValidation
  (InternalError) rather than blaming the user. I kept the guard (rather than dropping it) so a nil can't panic in the credential path downstream.

backend/pkg/azure/client/smi_client_builder.go
YuriyNik: Agreed — KeyVaultKeysClient didn't belong on ServiceManagedIdentityClientBuilder, which is SMI-exclusive by design.
  I moved it to a new dedicated interface, ClusterOperatorIdentityClientBuilder (in operator_identity_client_builder.go), whose purpose is to build clients authenticated as a cluster operator identity (here, the kms UAMI).
  ServiceManagedIdentityClientBuilder no longer exposes any Key Vault method and remains SMI-only, consistent with its type documentation.

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
YuriyNik: Thanks for flagging this. Here's how it works now:
  The check uses the Key Vault data plane (GetKey) via the cluster's KMS operator identity, and the outcome is classified by the error returned:
  - 401/403 → KeyVaultAccessDenied (vault not accessible/authorized for the identity)
  - 404 → KmsKeyNotFound (vault is accessible, but the key doesn't exist)
  - anything else → Unknown (transient/unclassified)
  So the "vault reachable but key missing" case you described is reported distinctly as KmsKeyNotFound, not as an accessibility failure. GetKey is used intentionally, since it exercises the same operation etcd performs at
  runtime.

miguelsorianod: If we focus on the reported issue, they mention that it could happen that the actual visibility configuration changes. This PR and validation is not attempting to check that, which requires the ARM control plane API part. It's attempting to instead perform some kind of API request at the dataplane level to try to do some sort of connectivity verification. It's also doing that because currently the key vault resource id is not available as I mentioned in https://github.com/Azure/ARO-HCP/pull/6487#discussion_r3750045147, so this is trying to workaround the fact that we don't have the resourcegroup information by using key vault's data plane api and relying on the KMS identity itself (which also comes with additional calls to the MIDataplane service as a consequence) to try to do something at that level.

With that, we end up with some sort of validation that doesn't really validate what was intended originally. Even with that approach, the validation doesn't validate what it currently claims to do (according to the current code doc "Validates that the Azure Key Vault used for customer-managed etcd KMS encryption is reachable by the cluster's KMS operator managed identity -- the same identity")

With the current approach, getting a 404 key not found is treated as the validation failed, which is incorrect. That the key doesn't exist it doesn't mean there's no connectivity, it means that the key doesn't exist.

With the current approach, getting a 401/403 doesn't mean that the keyvault is not accessible at connection level, it means that the identity can't authenticate or doesn't have permissions at the data plane level, that doesn't mean there isn't connectivity.

Aside from that, even when the current approach ends with successful validation, there's no guarantee that there will be connectivity from where it's needed: The validation which runs in backend is occurring from the service cluster, but the actual interaction runs  from within the management cluster, so even if there's connectivity from backend level nothing guarantees that the actual workload running in the management cluster side will have connectivity.

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
```diff
@@ -0,0 +1,144 @@
...
+		return a.classifyGetKeyError(err, kmsProfile.ActiveKey.VaultName, kmsProfile.ActiveKey.Name)
+	}
+
+	return PassedValidation(coreapi.ControllerConditionReasonAsExpected, "As expected", "KMS key is accessible to the cluster's KMS operator identity.")
+}
+
+// classifyGetKeyError examines the error from GetKey and returns an appropriate
+// ValidationResult based on the error type:
+//   - 401/403: Permission denied - the KMS identity lacks access to the Key Vault
```
miguelsorianod, line 104: > //   - 401/403: Permission denied - the KMS identity lacks access to the Key Vault

This claim is not correct as mentioned in https://github.com/Azure/ARO-HCP/pull/6487/changes#r3796054243.

That you get that doesn't mean that the identity doesn't have access to the key vault. It means that it can't authenticate, or doesn't have permissions to perform a very specific action (get against a specific key)

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
```diff
@@ -0,0 +1,144 @@
...
+//   - 404: Key not found - the specified key does not exist in the vault
```
miguelsorianod, line 105: > //   - 404: Key not found - the specified key does not exist in the vault

This statement is accurate, but this validation is not about validating whether the key exists or not. Having that as a validation outcome of failure is conceptually wrong. This is explained in detail here https://github.com/Azure/ARO-HCP/pull/6487/changes#r3796054243

backend/pkg/utils/validationutils/azure_cluster_keyvault_accessibility_validation.go
miguelsorianod: What I am observing from here is that we are trying to implement something without having really done the work of understanding the issue, what is that we **really** want to check and whether the check is correct given that validations run in another workload than where the actual workloads interact with the keyvault which means that the network configuration, topology, source is different.

The order of events that occurred are:
1. We first tried to implement the PR by checking the Key Vault's configuration using the Key Vault's control plane API
2. We then noticed that we can't do 1. because for that the key vault's azure resource id is needed and we don't currently have it because it's not passed as part of the cluster's creation payload. Because of this, we tried to workaround it and we changed the original focus from configuration checking to interaction at the data plane level taking the keys endpoint and the provided key to try to do something

Before continuing any further implementation:

Understand what we want to check, and whether doing it provides any value. The original Jira seemed to be about configuration of a specific setting and not actual connectivity. N different settings can impact connectivity. If for some reason we determine we want to check connectivity instead of the config (why? and also, what is that we define as "connectivity") then understand what would be an accurate way to check that (as mentioned the actual workloads don't run in the same cluster than backend), if possible at all, and that it doesn't mix concerns (connectivity vs key existence etc...). Notice also that for example by focusing on "connectivity" you don't really detect configuration drift changes around the key vault visibility, and when focusing on configuration drift changes you don't focus on connectivity (connectivity can change not only based on the configuration of the keyvault visibility). A potential solution could even require changing the API if we determine that it's something that we want to start requiring (with its potential implications like possibly API breaking changes and so on). Don't necessarily discard something because some information not being currently there.

PR #6492, opened 2026-08-10
test: integration tests for unmarshalling errors propagated as http 400

mzazrivec:
https://redhat.atlassian.net/browse/ARO-28842

### What

This pull request changes the frontend behavior in situations when a PUT request payload contains invalid properties. Before, the frontend would catch an error during json unmarshalling and return http/500. With this fix in place, we're returning http/400.

EDIT: With https://github.com/Azure/ARO-HCP/pull/6498 being merged, this PR still adds the integration tests for the described scenario.

### Why

http/400 is more appropriate in these situations.

### Testing

The PR contains integration tests for the described problem.

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test-integration/frontend/artifacts/FrontendCRUD/NodePool/update-invalid-request-content/08-httpReplace-nodepool/expected-error.txt
+++ b/test-integration/frontend/artifacts/FrontendCRUD/NodePool/update-invalid-request-content/08-httpReplace-nodepool/expected-error.txt
@@ -1,0 +1,4 @@
+{
+    "code": "InvalidRequestContent",
+    "message": "The request content was invalid and could not be deserialized: \"unmarshalling type *generated.NodePool: struct field Properties: unmarshalling type *generated.NodePoolProperties: json: cannot unmarshal string into Go value of type map[string]json.RawMessage\""
+  }
--- a/test-integration/frontend/artifacts/FrontendCRUD/Cluster/invalid-request-content/00-httpCreate-subscription/00-key.json
+++ b/test-integration/frontend/artifacts/FrontendCRUD/Cluster/invalid-request-content/00-httpCreate-subscription/00-key.json
@@ -1,0 +1,4 @@
+{
+	"id": "2f2d4ba3-1ee6-57b6-bde9-b61d09eaf989",
+	"resourceID": "/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7"
+}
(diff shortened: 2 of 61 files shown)


stevekuznetsov:
dupe of https://github.com/Azure/ARO-HCP/pull/6498

test-integration/frontend/artifacts/FrontendCRUD/NodePool/update-invalid-request-content/08-httpReplace-nodepool/expected-error.txt
```diff
@@ -0,0 +1,3 @@
+{
+    "code": "InvalidRequestContent",
+    "message": "The request content was invalid and could not be deserialized:
```
stevekuznetsov: Why is this weird half-JSON?

mzazrivec: Content of `expected-error.txt` is used as a [substring to match against](https://github.com/Azure/ARO-HCP/blob/main/test-integration/utils/databasemutationhelpers/step_httpcreate.go#L84), not as a JSON document to parse.

In this particular case the the point is to catch the `InvalidRequestContent` error code, not the deserializer's variable tail.

stevekuznetsov: Don't we fully bound the inputs? Why not just be specific and put the entire error here?

stevekuznetsov: Looking at the rest of the repo FWIW we are fully explicit for the entire error object that is returned in all the test fixtures I reviewed. I don't know specifically why we allow a substring match as opposed to forcing a reproducible and fully explicit error, but we should always aim to make our tests as precise and explicit as possible, please.

mzazrivec: Yes, some of the error expects contain full error jsons, but most of them contain partial error, that looks like a json, some contain just a plain error strings (no json structure) and some even contain plain unfinished error strings ([here](https://github.com/Azure/ARO-HCP/blob/main/test-integration/frontend/artifacts/FrontendCRUD/ExternalAuth/duplicate-client-components/04-httpCreate-externalauth/expected-error.txt) for example).

Anyway, I added the trailing error msgs to my error expects with a trailing `"` and `}` which is what I think you meant.

PR #6498, opened 2026-08-10
frontend: return unmarshal errors as 4xx to users

stevekuznetsov:
When a user passes an invalid JSON payload to our API endpoints, we were currently wrapping the unmarshalling errors and backstopping them with 500s, while in reality these are client-side errors.


frontend/pkg/frontend/cluster.go
```diff
@@ -214,7 +214,7 @@ func decodeDesiredClusterCreate(ctx context.Context, azureLocation string, reque
 
 	externalClusterFromRequest := versionedInterface.NewHCPOpenShiftCluster(nil)
 	if err := json.Unmarshal(body, &externalClusterFromRequest); err != nil {
-		return nil, utils.TrackError(err)
+		return nil, coreapi.NewInvalidRequestContentError(err)
```
deads2k: still wrap in `utils.TrackError`.  Knowing exactly which one is helpful.

stevekuznetsov: done

PR #6501, opened 2026-08-10
docs: add SOP for manually triggering an E2E gate run

hbhushan3:
https://issues.redhat.com/browse/AROSLSRE-1785

### What

Adds `docs/sops/manual-e2e-gate-run.md`, documenting how to manually trigger the same Prow E2E job EV2 uses for `regionalGating` — either to clear a flaky rollout gate in INT/STG, or to run the suite against a deployed environment from a branch.

Linked from `docs/README.md`, `docs/ci/README.md`, `docs/ci/e2e-testing.md`, and `docs/ci/ev2-integration.md`.

### Why

INT and STG rollouts have **no per-step retry** on the gating step (the `automatedRetry` block in `test/e2e-pipeline.yaml` only takes effect in PROD — see [AROSLSRE-1764](https://redhat.atlassian.net/browse/AROSLSRE-1764)). When the gate fails on a flake, the options are to retry the entire rollout, wait for another incremental rollout, or re-run the gate job manually with the rollout version associated. The third is by far the fastest, and it was completely undocumented.

The `make -C test/ {int,stage,prod}-e2e` targets are the mechanism EV2 itself uses, but they appear nowhere under `docs/`. Working the procedure out from source during a real uksouth INT gate failure took about an hour, including ~22 minutes lost to a silent authentication retry loop that produced no output at all.

Everything that cost time is now written down:

- Mapping an EV2 rollout to its Prow run via the `ev2.rollout/*` annotations, and reconstructing `zz_injected_EV2RolloutVersion` from them — required so the manual run is associated with the rollout rather than orphaned.
- The mandatory `AZURE_TOKEN_CREDENTIALS=dev`, which the make target does not set.
- The per-environment `AZURE_TENANT_ID` (INT is the Red Hat tenant, STG/PROD are Microsoft).
- Failure modes: wrong-tenant login surfacing as `AADSTS50020` and then silently backing off for up to ~63 minutes, expired `prow-token` returning an HTML login page, Gangway 429s, stale `vendor/modules.txt`, and the `tail` buffering trap that hides the Prow job URL for the whole run.

### Testing

Documentation only — no code changes, so no unit, integration, or E2E tests apply.

The procedure itself was validated end to end while retrying the failed uksouth INT gate for rollout build `175911391` (ARO-HCP `06acec5302a0`): the documented commands produced [a passing run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/branch-ci-Azure-ARO-HCP-main-e2e-integration-e2e-parallel/2086845890059833344) of the same job on the same commit that had [failed under EV2](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/branch-ci-Azure-ARO-HCP-main-e2e-integration-e2e-parallel/2086798198998831104).

Two things worth a second opinion:

1. **`docs/ci/e2e-testing.md` currently states that manual testing against INT/STG/PROD is not allowed.** I read that as a prohibition on manually creating clusters via `az`/the portal rather than on triggering the gate job, and said so in the SOP, but I have not edited that page's claim. If the intent was broader, this SOP needs rewording.
2. **The token vault.** `test/Makefile` reads the Prow token from `arohcpdev-global` (via `config/rendered/dev/dev/westus3.yaml`) regardless of target environment. The SOP states that as-is, but I could not verify how a Microsoft-tenant login reaches a `dev` vault for STG/PROD runs.

Related: [AROSLSRE-1786](https://redhat.atlassian.net/browse/AROSLSRE-1786) tracks manual runs carrying no `ev2.rollout` provenance when the rollout version is omitted.

### DIFF
--- a/docs/sops/manual-e2e-gate-run.md
+++ b/docs/sops/manual-e2e-gate-run.md
@@ -1,0 +1,191 @@
+# Manually Trigger an E2E Gate Run
+
+This SOP describes how to manually trigger the same Prow E2E job that EV2 uses for regional gating. See [When To Use This Procedure](#when-to-use-this-procedure).
+
+## Background
+
+Every EV2 rollout gates regional promotion on the `regionalGating` validation step defined in [`test/e2e-pipeline.yaml`](../../test/e2e-pipeline.yaml). The step triggers an environment-specific Prow job through the Gangway API, pinned via `--base-sha` to the exact ARO-HCP commit being rolled out. See [CI EV2 Integration](../ci/ev2-integration.md) for the full wiring.
+
+The `make -C test/ {int,stage,prod}-e2e` targets invoke the same `prow-job-executor` with the same arguments, which is what makes them a faithful reproduction of the gate.
+
+## When To Use This Procedure
+
+| Situation | Environment | What to do |
+|---|---|---|
+| Failed rollout gate | PROD | Use the retry button in the [EV2 portal](https://ra.ev2portal.azure.net). Do **not** use this SOP. |
+| Failed rollout gate | INT, STG | No per-step retry exists. Use this SOP — steps 1, 2 (retry variant), 3. See the alternatives below. |
+| Validating E2E test-code changes, no PR open | INT, STG, PROD | Use this SOP — step 2 (branch variant) and step 3. |
+| Validating E2E test-code changes, PR already open | INT, STG, PROD | Prefer the `/test` comment triggers in [Running E2E Tests In CI](../ci/e2e-testing.md); they need no local Azure credentials. |
+
+The prohibition on manual testing in INT/STG/PROD described in that guide concerns manually creating clusters via `az` or the portal, not triggering the gate job.
+
+### Alternatives For A Failed INT Or STG Gate
+
+The EV2 portal offers no retry control on the gating step in these environments, so a flaky failure leaves three choices:
+
+1. **Retry the whole rollout.** Correct but slow — it re-executes everything, not just the gate.
+2. **Trigger another incremental rollout.** Also slow, and only makes sense if there is a subsequent revision to roll out anyway.
+3. **Run the gate job manually with the rollout version associated** — this SOP. Fastest way to clear a flake, and the reason the `zz_injected_EV2RolloutVersion` association in [step 2](#retrying-a-failed-int-or-stg-gate) is mandatory rather than cosmetic: it is what ties the run back to the rollout being unblocked.
+
+The `automatedRetry` block on the `regionalGating` step currently only takes effect in PROD. STG and INT set `useExclusiveLocks` in `stages.yaml`, and the sdp-pipelines generator ignores an explicit `automatedRetry` whenever the stage fails fast, so those stages fall back to the default MSI/connection-failure-only retry. Tracked as a generator fix in AROSLSRE-1764.
+
+> [!IMPORTANT]
+> This does not re-execute the EV2 step itself — it produces a passing gate run associated with the rollout. If the rollout still does not progress, fall back to option 1 or 2.
+
+### Constraints When Testing Your Own Changes
+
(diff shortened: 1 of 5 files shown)


docs/sops/manual-e2e-gate-run.md
```diff
@@ -0,0 +1,193 @@
...
+
+1. **Retry the whole rollout.** Correct but slow — it re-executes everything, not just the gate.
+2. **Trigger another incremental rollout.** Also slow, and only makes sense if there is a subsequent revision to roll out anyway.
+3. **Run the gate job manually with the rollout version associated** — this SOP. Fastest way to clear a flake, and the reason the `zz_injected_EV2RolloutVersion` association in [step 2](#retrying-a-failed-int-or-stg-gate) is mandatory rather than cosmetic: it is what ties the run back to the rollout being unblocked.
+
+The `automatedRetry` block on the `regionalGating` step currently only takes effect in PROD. STG and INT set `useExclusiveLocks` in `stages.yaml`, and the sdp-pipelines generator ignores an explicit `automatedRetry` whenever the stage fails fast, so those stages fall back to the default MSI/connection-failure-only retry. Tracked as a generator fix in AROSLSRE-1764.
+
+> [!IMPORTANT]
+> This does not re-execute the EV2 step itself — it produces a passing gate run associated with the rollout. If the rollout still does not progress, fall back to option 1 or 2.
```
roivaz, line 33: I don't expect the rollout to progress without re-executing the EV2 step or skipping it. The procedure described in this SOP runs a gating e2e outside of EV2, so there's no way the failed EV2 gating step knows that a retry succeeded. Can we clarify this instead of speculating? This is the whole point to having this SOP so the outcome should be known and documented.

docs/sops/manual-e2e-gate-run.md
```diff
@@ -0,0 +1,193 @@
...
+> Always revert the filter before merging. Leaving it in place silently skips tests in CI.
+
+## Environment Reference
+
+Every command below needs the values on one row of this table.
+
+| Environment | Make target | `AZURE_TENANT_ID` (tenant) | Prow job |
+|---|---|---|---|
+| INT | `int-e2e` | `64dc69e4-d083-49fc-9569-ebece1dd1408` (Red Hat) | `branch-ci-Azure-ARO-HCP-main-e2e-integration-e2e-parallel` |
+| STG | `stage-e2e` | `72f988bf-86f1-41af-91ab-2d7cd011db47` (Microsoft) | `branch-ci-Azure-ARO-HCP-main-e2e-stage-e2e-parallel` |
```
roivaz, line 54: This tenant mapping is incorrect for manual triggers. `test/Makefile` unconditionally resolves `PROW_TOKEN_KEYVAULT` and `PROW_TOKEN_SECRET` from `config/rendered/dev/dev/westus3.yaml`; `ENVIRONMENT=stg|prod` only selects the target Prow job and test environment. Authentication is therefore against the shared DEV Key Vault, so STG and PROD triggers should also use the Red Hat tenant (`64dc69e4-d083-49fc-9569-ebece1dd1408`).

Please separate the target environment from the token-retrieval tenant: remove the per-environment tenant values (or set all three to Red Hat), update the prerequisite to say that `az login` must access `arohcpdev-global`, and update the STG/PROD command examples accordingly. The current Microsoft-tenant guidance will fail unless that identity independently has access to the DEV vault.

avollmer-redhat (approved):
Excellent operational doc, validated against a real incident (uksouth INT gate failure, build 175911391).

Key value-adds:
- `zz_injected_EV2RolloutVersion` reconstruction from `ev2.rollout/*` annotations — non-obvious and previously undocumented
- Troubleshooting table with the AADSTS50020 silent backoff trap (~63 min dead time) and `tail` buffering gotcha — high-value operational knowledge
- Environment reference table with per-env tenant IDs
- `docs/ci/e2e-testing.md` clarification (NOTE distinguishing manual cluster creation from gate triggering) resolves the ambiguity well

Cross-links to `docs/README.md`, `docs/ci/README.md`, and `docs/ci/ev2-integration.md` are correct and well-placed.

PR #6503, opened 2026-08-11
feat: gate kube-applier startup probe on CosmosDB reachability

ashishmax31:
kube-applier had only a readiness probe, and it checked the leader-election lease — nothing checked Cosmos. During ITN-2026-00171 every Cosmos query returned 403 while the pods stayed Ready, so the broken version rolled out and stage E2E cluster creation timed out.

Add /readyz, which runs one real data-plane query (single ApplyDesire page) against this pod's container, and wire it to a new startupProbe so a pod never goes Ready while Cosmos is unreachable. Add a livenessProbe on /healthz. Liveness and readiness deliberately stay off Cosmos: a Cosmos blip must not restart-loop every replica or drop the pods out of the metrics endpoints mid-incident.

Jira: [ARO-27566](https://redhat.atlassian.net/browse/ARO-27566)

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/kube-applier/deploy/templates/deployment.yaml
+++ b/kube-applier/deploy/templates/deployment.yaml
@@ -58,6 +58,19 @@ spec:
         - name: metrics
           containerPort: 8081
           protocol: TCP
+        startupProbe:
+          httpGet:
+            path: /readyz
+            port: healthz
+          periodSeconds: 10
+          timeoutSeconds: 5
+          failureThreshold: 30
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: healthz
+          periodSeconds: 20
+          failureThreshold: 3
         readinessProbe:
           httpGet:
             path: /healthz
--- a/kube-applier/pkg/app/kube_applier.go
+++ b/kube-applier/pkg/app/kube_applier.go
@@ -24,6 +24,7 @@ import (
 
 	_ "k8s.io/component-base/metrics/prometheus/clientgo"
 
+	"github.com/go-logr/logr"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -33,6 +34,7 @@ import (
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/component-base/metrics/legacyregistry"
 
+	"github.com/Azure/ARO-HCP/internal/database/cosmosstorage/cosmosstorageutils"
 	"github.com/Azure/ARO-HCP/internal/database/informers/kubeapplierinformers"
 	sharedleaderelection "github.com/Azure/ARO-HCP/internal/leaderelection"
 	"github.com/Azure/ARO-HCP/internal/utils"
@@ -40,6 +42,10 @@ import (
 	"github.com/Azure/ARO-HCP/kube-applier/pkg/controllers/apply_desire"
 	"github.com/Azure/ARO-HCP/kube-applier/pkg/controllers/read_desire_manager"
 )
+
+// cosmosProbeTimeout bounds the Cosmos query behind /readyz so a hung query
+// cannot hold the handler open after the kubelet has given up on the probe.
+const cosmosProbeTimeout = 5 * time.Second
 
 // Run is the binary's main loop. It serves /healthz and /metrics, then runs
 // the controllers under a leader-election lease. Run returns when ctx is
@@ -63,21 +69,7 @@ func (o *Options) Run(ctx context.Context) error {
 	)
 
 	if o.HealthzServerListenAddress != "" {
-		healthGauge := promauto.With(o.metricsRegisterer()).NewGauge(prometheus.GaugeOpts{
-			Name: "kube_applier_health", Help: "kube_applier_health is 1 when healthy",
-		})
-		mux := http.NewServeMux()
-		mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
-			if err := electionChecker.Check(r); err != nil {
-				logger.Error(err, "readiness probe failed")
-				http.Error(w, "lease not renewed", http.StatusServiceUnavailable)
-				healthGauge.Set(0)
-				return
-			}
-			w.WriteHeader(http.StatusOK)
-			healthGauge.Set(1)
-		})
-		server := &http.Server{Addr: o.HealthzServerListenAddress, Handler: mux}
+		server := &http.Server{Addr: o.HealthzServerListenAddress, Handler: o.newHealthzMux(logger, electionChecker)}
 		wg.Add(1)
 		go func() {
 			defer cancel(fmt.Errorf("healthz server exited"))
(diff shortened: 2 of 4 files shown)


inbharajmani:
## Review finding: `HealthzAdaptor` lease check is a no-op

The `/healthz` endpoint calls `electionChecker.Check(r)`, which is intended to detect a deadlocked leader that stopped renewing its lease. However, the check **always returns healthy** because the `HealthzAdaptor` is never connected to the `LeaderElector`.

### Root cause

`kube-applier/pkg/app/kube_applier.go:193-198` uses `NewLeaderElector` + `Run()`:

```go
le, err := leaderelection.NewLeaderElector(leaderElectionConfig)
if err != nil {
    return err
}
le.Run(ctx)
```

The `WatchDog: electionChecker` is set in the config (line 187), but neither `NewLeaderElector` nor `Run()` calls `SetLeaderElection(le)` on the adaptor — only `RunOrDie()` does that (upstream `k8s.io/client-go@v0.35.3/tools/leaderelection/leaderelection.go:232-233`).

Since `SetLeaderElection` is never called, the adaptor's internal `le` pointer stays `nil`. In `HealthzAdaptor.Check()` (upstream `healthzadaptor.go:45-52`):

```go
func (l *HealthzAdaptor) Check(req *http.Request) error {
    l.pointerLock.Lock()
    defer l.pointerLock.Unlock()
    if l.le == nil {
        return nil       // ← always hits
    }
    return l.le.Check(l.timeout)
}
```

This means `/healthz` unconditionally returns 200 for every replica, leader or not.

### Scope

This is a pre-existing issue, not introduced by this PR. The same pattern exists in:
- `backend/pkg/app/backend.go:1147-1152`
- `fleet/pkg/manager/manager.go:292-296`

Confirmed: `grep -rn 'SetLeaderElection' --include='*.go'` returns zero hits across the entire repo.

### Why it was never noticed

Before this PR, these three services only had a **readinessProbe** on `/healthz` — no livenessProbe. Even if the check worked and a non-leader failed it, readiness failure just drops the pod from Service endpoints (fine for leader-elected workloads). And since the check always returned 200 anyway, there was no observable symptom.

### Impact on this PR

The new **livenessProbe** on `/healthz` is safe (it won't restart-loop non-leader replicas), but it also provides no value — it can't detect a deadlocked leader because the lease check never runs. The **startupProbe** on `/readyz` (Cosmos reachability) is the valuable part of this PR and works correctly.

### Suggested fix

Add the missing wiring after `NewLeaderElector` returns:

```go
le, err := leaderelection.NewLeaderElector(leaderElectionConfig)
if err != nil {
    return err
}
electionChecker.SetLeaderElection(le)   // ← add this line
le.Run(ctx)
```

This is safe for non-leader replicas — upstream `LeaderElector.Check()` explicitly returns `nil` for non-leaders (`leaderelection.go:508-510`: `if !le.IsLeader() { return nil }`). Once wired, it would only fail if the leader has not renewed the lease for longer than `LeaseDuration + timeout` (137s + 20s = 157s), catching the deadlocked-leader scenario.

This fix could be done in this PR or as a follow-up across all three affected services (kube-applier, backend, fleet).

ashishmax31:
Thi

> ## Review finding: `HealthzAdaptor` lease check is a no-op
> The `/healthz` endpoint calls `electionChecker.Check(r)`, which is intended to detect a deadlocked leader that stopped renewing its lease. However, the check **always returns healthy** because the `HealthzAdaptor` is never connected to the `LeaderElector`.
> 
> ### Root cause
> `kube-applier/pkg/app/kube_applier.go:193-198` uses `NewLeaderElector` + `Run()`:
> 
> ```go
> le, err := leaderelection.NewLeaderElector(leaderElectionConfig)
> if err != nil {
>     return err
> }
> le.Run(ctx)
> ```
> 
> The `WatchDog: electionChecker` is set in the config (line 187), but neither `NewLeaderElector` nor `Run()` calls `SetLeaderElection(le)` on the adaptor — only `RunOrDie()` does that (upstream `k8s.io/client-go@v0.35.3/tools/leaderelection/leaderelection.go:232-233`).
> 
> Since `SetLeaderElection` is never called, the adaptor's internal `le` pointer stays `nil`. In `HealthzAdaptor.Check()` (upstream `healthzadaptor.go:45-52`):
> 
> ```go
> func (l *HealthzAdaptor) Check(req *http.Request) error {
>     l.pointerLock.Lock()
>     defer l.pointerLock.Unlock()
>     if l.le == nil {
>         return nil       // ← always hits
>     }
>     return l.le.Check(l.timeout)
> }
> ```
> 
> This means `/healthz` unconditionally returns 200 for every replica, leader or not.
> 
> ### Scope
> This is a pre-existing issue, not introduced by this PR. The same pattern exists in:
> 
> * `backend/pkg/app/backend.go:1147-1152`
> * `fleet/pkg/manager/manager.go:292-296`
> 
> Confirmed: `grep -rn 'SetLeaderElection' --include='*.go'` returns zero hits across the entire repo.
> 
> ### Why it was never noticed
> Before this PR, these three services only had a **readinessProbe** on `/healthz` — no livenessProbe. Even if the check worked and a non-leader failed it, readiness failure just drops the pod from Service endpoints (fine for leader-elected workloads). And since the check always returned 200 anyway, there was no observable symptom.
> 
> ### Impact on this PR
> The new **livenessProbe** on `/healthz` is safe (it won't restart-loop non-leader replicas), but it also provides no value — it can't detect a deadlocked leader because the lease check never runs. The **startupProbe** on `/readyz` (Cosmos reachability) is the valuable part of this PR and works correctly.
> 
> ### Suggested fix
> Add the missing wiring after `NewLeaderElector` returns:
> 
> ```go
> le, err := leaderelection.NewLeaderElector(leaderElectionConfig)
> if err != nil {
>     return err
> }
> electionChecker.SetLeaderElection(le)   // ← add this line
> le.Run(ctx)
> ```
> 
> This is safe for non-leader replicas — upstream `LeaderElector.Check()` explicitly returns `nil` for non-leaders (`leaderelection.go:508-510`: `if !le.IsLeader() { return nil }`). Once wired, it would only fail if the leader has not renewed the lease for longer than `LeaseDuration + timeout` (137s + 20s = 157s), catching the deadlocked-leader scenario.
> 
> This fix could be done in this PR or as a follow-up across all three affected services (kube-applier, backend, fleet).

This is looks like an existing issue not introduced in this PR.

raelga (approved):
/lgtm
/approve

avollmer-redhat (commented):
/lgtm

Code is clean, probe separation is correct, tests pass, and the incident motivation (ITN-2026-00171) is well-documented.

One concern worth noting (non-blocking): the `livenessProbe` on `/healthz` will restart-loop the non-leader replica after the 20s `LeaderHealthzAdaptor` timeout expires, since only the leader gets `SetHealthy` called via the watchdog. The existing `readinessProbe` already had this endpoint so the standby was already not-ready, but the new liveness probe adds restart-looping on top. If that turns out noisy, a trivial `/livez` that always returns 200 would fix it. Fine to address as follow-up if needed.

The Copilot comment about `failingIterator(l)` being an invalid conversion is wrong — both `failingLister` and `failingIterator` have identical underlying structure (`struct { err error }`), so the conversion is valid Go. CI test-unit confirms it compiles and passes.

PR #6516, opened 2026-08-12
observability: add Ingress Availability and Latency SLI/SLO definitions (ARO-22170)

ArrisLee:
https://redhat.atlassian.net/browse/ARO-22170

### What

Adds SLI recording rules for the Ingress customer user journey ([ARO-22167](https://redhat.atlassian.net/browse/ARO-22167)), covering both **Availability** and **Latency**.

**Recording rules** (`ingress-availability-slo-recordingRule.yaml`):

Availability (`arohcp_ingress_availability_slo_recording_rules`):
- `availability:ingress_canary:ratio` — Availability SLI: fraction of synthetic canary route observations reporting the ingress route as reachable, aggregated per HCP cluster via `sum/count by (_id, cluster, region)`. Uses ratio rather than `min` so a single failed observation in a redundant setup does not report the whole cluster as unavailable.
- `errors:ingress_canary:error_rate` — error rate derived from availability (`1 - ratio`), intended for burn-rate alert expressions ([ARO-25969](https://redhat.atlassian.net/browse/ARO-25969)).

Latency (`arohcp_ingress_latency_slo_recording_rules`):
- `latency:ingress_canary:ratio` — Latency SLI: fraction of synthetic canary route checks completing within 200ms, from the `ingress_canary_check_duration` histogram, per `(_id, cluster, region)`.
- `errors:ingress_canary_latency:error_rate` — error rate derived from latency (`1 - ratio`), for burn-rate alerts ([ARO-25969](https://redhat.atlassian.net/browse/ARO-25969)).

SLO for both: **99.5%** (error budget 0.5%). Registered in `recording-rules-hcps.yaml` (HCP AMW layer).

Alert rules ([ARO-25969](https://redhat.atlassian.net/browse/ARO-25969)) and the dashboard ([ARO-22171](https://redhat.atlassian.net/browse/ARO-22171)) are intentionally delivered in separate PRs; this PR establishes the SLI/SLO foundation only.

### Why

[ARO-22167](https://redhat.atlassian.net/browse/ARO-22167) requires SLI/SLO definitions as the foundation for dashboards, alerting, and TSGs. This PR resolves the impediment flagged in [ARO-22170](https://redhat.atlassian.net/browse/ARO-22170).

**SLI scope (important for downstream TSGs):** `ingress_canary_route_reachable` and `ingress_canary_check_duration` are **end-to-end synthetic route checks** performed by `ingress-operator` against the cluster's canary route. A breach means the ingress data path is not reachable (or is slow) end to end — the failing hop could be DNS, the load balancer, ingress routing, or router pods. It is not a direct router-pod-availability signal, and `host` series are not validated to map 1:1 to pods, so the ratio is expressed as "fraction of synthetic canary route observations that succeed".

Metrics validated against `Managed_Prometheus_hcps-*`:
- `ingress_canary_route_reachable`: `_id`, `cluster`, `region` labels confirmed (INT 2026-07-29; label set re-confirmed on prod 2026-08-14), 26 HCP clusters visible, emitted by `ingress-operator` (management cluster, authoritative source). `subscription_id` / `resource_id` are not present on this metric.
- `ingress_canary_check_duration_bucket`: confirmed on prod (2026-08-17) as a histogram with buckets 25/50/100/200/400/800/1600/+Inf ms, same label set.

### Testing

Promtool unit tests (`ingress-availability-slo-recordingRule_test.yaml`):
- Availability: all reachable → ratio 1; half unreachable → 0.5; all unreachable → 0; multi-cluster isolation (each `_id` independent); no-data → empty result set.
- Latency: all checks fast → ratio 1; half slow → 0.5; no-data → empty result set.

- Metrics are HCP-layer (emitted by `ingress-operator` on the management cluster), hence registered in `recording-rules-hcps.yaml` rather than `recording-rules-services.yaml`.
- `availability:` / `latency:` introduce new recording-rule namespaces alongside the existing `errors:` convention. This is intentional: the ratios represent good events / total events, and the derived `errors:*:error_rate` rules follow the existing `errors:` convention that alert expressions consume.
- No-data behavior matches the kube-apiserver SLI pattern (empty result when all series disappear); a companion `absent()` liveness alert is planned in [ARO-25969](https://redhat.atlassian.net/browse/ARO-25969).


observability/recording-rules-hcps.yaml
```diff
@@ -3,5 +3,6 @@ prometheusRules:
   - alerts/HCPkasRecord-prometheusRule-KSM.yaml
   - alerts/HCPkasRecord-prometheusRule-apiserver-requests.yaml
   - alerts/HCPkasRecord-prometheusRule-latency.yaml
+  - recording-rules/ingress-availability-slo-recordingRule.yaml
```
swiencki, line 7: Please run `make -C observability recording-rules` and commit the resulting `generatedHCPRecordingRules.bicep`, otherwise these rules will not deploy.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
```diff
@@ -0,0 +1,52 @@
...
+    # Used by burn-rate alert expressions.
+    - record: errors:ingress_canary:error_rate
+      expr: |
+        1 - availability:ingress_canary:ratio
+    # Errors SLI: ingress controller degraded state per HCP cluster.
+    # Value is 0 (healthy) or 1 (degraded). Reflects the ingress controller's
+    # own assessment — a complementary signal to the canary-based availability SLI.
+    - record: errors:ingress_controller:degraded
+      expr: |
+        max by (_id) (ingress_controller_conditions{condition="degraded"})
```
swiencki: Please reconcile this lowercase `degraded` matcher with the upstream `Degraded` label and make the test use the confirmed production casing.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
```diff
@@ -0,0 +1,52 @@
...
+  # A follow-up story is needed to route data plane metrics into AMW.
+  - name: arohcp_ingress_availability_slo_recording_rules
+    interval: 1m
+    rules:
+    # Availability SLI: fraction of canary pods reporting ingress as reachable,
+    # aggregated per HCP cluster.
+    # Ratio format per ADR: good events / total events
+    # Uses sum/count (not min) so that a single pod failure in a redundant setup
+    # does not report the entire cluster as unavailable.
+    - record: availability:ingress_canary:ratio
```
swiencki, line 54: This is an instantaneous ratio of currently scraped hosts, not the stated 99.5% over 30 days SLI, and it returns no series when all canary metrics disappear.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
```diff
@@ -0,0 +1,52 @@
...
+      expr: |
+        sum by (_id) (ingress_canary_route_reachable)
```
swiencki: Please confirm that the production metric carries `cluster`, `subscription_id`, and `resource_id`, then preserve them alongside `_id` so the recording rule retains regional, customer, and exact HCP identity.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
```diff
@@ -0,0 +1,52 @@
...
+        count by (_id) (ingress_canary_route_reachable)
+    # Error rate derived from the availability SLI.
+    # Used by burn-rate alert expressions.
+    - record: errors:ingress_canary:error_rate
+      expr: |
+        1 - availability:ingress_canary:ratio
+    # Errors SLI: ingress controller degraded state per HCP cluster.
+    # Value is 0 (healthy) or 1 (degraded). Reflects the ingress controller's
+    # own assessment — a complementary signal to the canary-based availability SLI.
+    - record: errors:ingress_controller:degraded
```
swiencki: This binary controller condition is component health, not an Errors SLI with a denominator, measurement window, and target.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
```diff
@@ -0,0 +1,52 @@
...
+  #
+  # SLO: 99.5% ingress canary availability
+  # Error Budget: 0.5% (1 - 0.995)
+  #
+  # Validated on 2026-07-29 against INT AMW (Managed_Prometheus_hcps-uksouth):
+  # - ingress_canary_route_reachable: _id label confirmed, 26 HCP clusters visible,
+  #   emitted by ingress-operator (control plane, authoritative source)
+  # - ingress_controller_conditions: _id label confirmed, condition=available/degraded
+  #
+  # Note: haproxy metrics (Errors, Latency, Traffic, Saturation) are not present
```
swiencki: `ingress_canary_check_duration` is a histogram exposed in production as `_bucket`, `_count`, and `_sum`, so please remove the claim that Latency lacks telemetry and evaluate a threshold-based SLI from those series.

observability/recording-rules/ingress-availability-slo-recordingRule_test.yaml
```diff
@@ -0,0 +1,124 @@
...
+    eval_time: 5m
+    exp_samples:
+    - labels: 'errors:ingress_canary:error_rate{_id="ccc"}'
+      value: 1
+# ========================================
+# Test 4: Controller degraded = 1
+# ========================================
+- interval: 1m
+  input_series:
+  - series: 'ingress_controller_conditions{_id="ddd", condition="degraded"}'
```
swiencki: Please use the exporter's real `name` and `Degraded` labels and add a case where all canary series are missing.

observability/recording-rules/ingress-availability-slo-recordingRule_test.yaml
```diff
@@ -0,0 +1,124 @@
...
+    values: "1+0x10"
+  - series: 'ingress_canary_route_reachable{_id="ggg", host="node2.apps.aro.example.io"}'
+    values: "0+0x10"
+  - series: 'ingress_canary_route_reachable{_id="ggg", host="node3.apps.aro.example.io"}'
+    values: "0+0x10"
+  promql_expr_test:
+  - expr: availability:ingress_canary:ratio
+    eval_time: 5m
+    exp_samples:
+    - labels: 'availability:ingress_canary:ratio{_id="fff"}'
```
swiencki: Please include the confirmed `cluster`, `subscription_id`, and `resource_id` labels in the test input and expected output.

ArrisLee: Confirmed. Production label set for ingress_canary_route_reachable: _id, cluster, container, endpoint, environment, host, instance, job, namespace, pod, prometheus, prometheus_replica, region. subscription_id and resource_id are not present on this metric — ingress-operator does not emit them. Updated by() to by (_id, cluster) to preserve per-HCP identity and regional management cluster context.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
```diff
@@ -0,0 +1,79 @@
...
+    interval: 1m
+    rules:
+    # Availability SLI: fraction of canary pods reporting ingress as reachable,
+    # aggregated per HCP cluster.
+    # Ratio format per ADR: good events / total events
+    # Uses sum/count (not min) so that a single pod failure in a redundant setup
+    # does not report the entire cluster as unavailable.
+    - record: availability:ingress_canary:ratio
+      expr: |
+        sum by (_id, cluster) (ingress_canary_route_reachable)
```
swiencki: The source metric is documented as carrying region, but this aggregation keeps only _id and cluster, so downstream alerts and dashboards lose the region dimension needed to filter HCP-specific ingress issues; include region in the grouping.

observability/recording-rules-hcps.yaml
```diff
@@ -4,5 +4,6 @@ prometheusRules:
   - alerts/HCPkasRecord-prometheusRule-apiserver-requests.yaml
   - alerts/HCPkasRecord-prometheusRule-latency.yaml
   - alerts/UserJourneyKubeApiserverAvailabilityRecord-prometheusRule-KSM.yaml
+  - recording-rules/ingress-availability-slo-recordingRule.yaml
```
swiencki, line 7: This registers only the ingress SLO recording rules, but the PR does not add companion MWMBR alerts or a dashboard, so the new SLO will be recorded without any paging or visualization path.

dev-infrastructure/modules/metrics/rules/generatedHCPRecordingRules.bicep
```diff
@@ -241,3 +241,47 @@ resource userjourneyKubeapiserverAvailabilityRecordingRules 'Microsoft.AlertsMan
...
+  location: location
+  properties: {
+    scopes: [
+      azureMonitoring
+    ]
+    enabled: true
+    interval: 'PT1M'
+    rules: [
+      {
+        record: 'availability:ingress_canary:ratio'
```
pepedocs, line 256: This measures end to end ingress availability not just router pods. There are points where the problem could be due to DNS or something else - we need to clarify the scope here because the downstream TSGs will be based on this.

pepedocs (changes requested):
It is important that we are precise about the SLI semantics here because this recording rule will drive future alerts, dashboards, and TSGs. ingress_canary_route_reachable is an end-to-end synthetic route check, not inherently a pod-availability metric.

The current comments and tests describe individual metric series as canary pods and imply the ratio represents tolerance of ingress-pod failures, but the series are only differentiated by host and this PR does not establish that they map 1:1 to ingress or canary pods. Could we describe this as the fraction of synthetic canary route observations that succeed, unless that topology mapping has been validated and documented?

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
ArrisLee: doublechecked it on prod, still shows "degraded"

dev-infrastructure/modules/metrics/rules/generatedHCPRecordingRules.bicep
ArrisLee: Thanks, agreed — this is exactly the precision we need since the TSGs depend on it. Fixed in 897125988:

- Reframed the rule comments and the PR description to describe the SLI as "the fraction of synthetic canary route observations that succeed", explicitly noting a breach means the ingress data path is not reachable **end to end** (DNS / load balancer / ingress routing / router pods), and that downstream TSGs must triage which hop failed.
- Removed all "canary pod" / "healthy pods" wording from comments and tests, since `host` series are **not** validated as 1:1 to pods.
- Tests now differentiate observations by `instance` and use synthetic route hostnames to reflect the route-check semantics.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
ArrisLee: Good catch — added `region` to the grouping in both `availability:ingress_canary:ratio` and `latency:ingress_canary:ratio` (`by (_id, cluster, region)`), and to the test input/expected output. `region` is same-source as `cluster`, so it introduces no incorrect series splitting. Fixed in 897125988.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
```diff
@@ -0,0 +1,79 @@
...
+    # Used by burn-rate alert expressions (ARO-25969).
+    - record: errors:ingress_canary:error_rate
+      expr: |
+        1 - availability:ingress_canary:ratio
+  # ========================================
+  # Ingress Latency SLO Recording Rules
+  # ========================================
+  # SLO: 99.5% of canary checks complete within 200ms
+  # Error Budget: 0.5% (1 - 0.995)
+  #
```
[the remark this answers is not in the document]
ArrisLee: Updated the PR title and description to include the Latency SLI — it's now an intentional part of this PR (option 2). Confirmed `ingress_canary_check_duration` exists in prod as a histogram (buckets 25/50/100/200/400/800/1600/+Inf ms), so the latency SLI is derived from real telemetry rather than being a gap.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
```diff
@@ -0,0 +1,79 @@
...
+    - record: availability:ingress_canary:ratio
+      expr: |
+        sum by (_id, cluster) (ingress_canary_route_reachable)
+        /
+        count by (_id, cluster) (ingress_canary_route_reachable)
+    # Error rate derived from the availability SLI.
+    # Used by burn-rate alert expressions (ARO-25969).
+    - record: errors:ingress_canary:error_rate
+      expr: |
+        1 - availability:ingress_canary:ratio
```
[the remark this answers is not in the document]
ArrisLee: The PR description was stale — removed the `errors:ingress_controller:degraded` reference. That controller-degraded rule was dropped earlier in review (it's binary component health, not an errors-SLI with a denominator/window/target), so the description now matches the actually-delivered rules.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
ArrisLee: Intentional split, following the existing SLI convention in this repo. This PR delivers the **instantaneous ratio** recording rules (same shape as `cluster-provision-slo` / `nodepool-slo` / `access-cluster-slo`). The 30-day error-budget evaluation and the time-window `sli_sum_5m` / `sli_count_5m` rules (kube-apiserver pattern) belong with the burn-rate alerts in **ARO-25969**, where they're consumed.

The no-series-on-disappearance behavior is deliberate and matches the kube-apiserver SLI; a companion `absent()` liveness alert is planned in ARO-25969 to page when canary metrics vanish entirely.

observability/recording-rules-hcps.yaml
ArrisLee: By design — this PR is scoped to the SLI/SLO foundation only. The companion MWMBR alerts land in **ARO-25969** and the dashboard in **ARO-22171**. Splitting keeps each PR reviewable and lets the recording rules bake before alerts page on them.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
```diff
@@ -0,0 +1,95 @@
...
+  # Production label set for ingress_canary_route_reachable (confirmed 2026-08-14):
+  #   _id, cluster, container, endpoint, environment, host, instance, job,
+  #   namespace, pod, prometheus, prometheus_replica, region
+  # Note: subscription_id and resource_id are NOT present on this metric.
+  # by(_id, cluster, region) preserves per-HCP identity, the management cluster,
+  # and the region dimension used by downstream alerts and dashboards.
+  #
+  # No-data behavior: when all canary series disappear, sum/count returns empty
+  # (no series), consistent with kube-apiserver SLI pattern. A separate absent()
+  # alert should detect canary liveness (follow-up in ARO-25969).
```
swiencki, line 45: These SLI recording rules look solid, but with the burn-rate and canary-absence alerts deferred to ARO-25969 and no dashboard in this PR, please sequence the Ingress runbook so it does not route on-call to alerts that are not implemented yet.

tkesharw-RH:
Looks good to me. I reviewed the changes and tested the recording rules locally with the pinned promtool version. The rule tests and Go tests passed, and regenerating the Bicep file produced no changes.

observability/recording-rules/ingress-availability-slo-recordingRule.yaml
ArrisLee: Good point on sequencing. Added a comment to the rule file (c14975225) clarifying:

- Burn-rate (MWMBR) alerts and the canary-absence alert are delivered in **ARO-25969**
- The Ingress runbook and TSG will be published **alongside** those alerts, not ahead of them
- Until ARO-25969 merges, these recording rules produce no paging path — they are a foundation only

So on-call will not be routed to any unimplemented alerts.

dev-infrastructure/modules/metrics/rules/generatedHCPRecordingRules.bicep
```diff
@@ -1,3 +1,4 @@
+
```
swiencki, line 1: Running `make -C observability recording-rules` on this branch removes this leading blank line while leaving the rest of the generated file unchanged, so please regenerate and commit to drop it and keep the bicep in sync with the generator.

ArrisLee: Fixed in 571559aeb — ran `az bicep format` on the file to drop the stray leading blank line, so it now matches the other generated rule files and is in sync with the generator output (verified: `diff` against `main` shows only the added ingress SLI/latency rule groups, zero removed lines).

Note for anyone regenerating locally: `make -C observability recording-rules` currently fails partway through on a pre-existing floating-point rounding mismatch in the **frontend** SLO test (`sli:frontend_http:latency_p99/p95`, unrelated to this PR), which aborts before the `fmt-devinfra` step that strips the blank line. Running `az bicep format` directly (what that step does) is the equivalent fix here. Also just merged latest `main`.

PR #6519, opened 2026-08-12
Upgrade to OCP 4.21 as default version in test framework

miquelsi:
https://redhat.atlassian.net/browse/ARO-29377

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/e2e/cluster_version_backlevel.go
+++ b/test/e2e/cluster_version_backlevel.go
@@ -1,0 +1,405 @@
+// Copyright 2025 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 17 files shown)


test/e2e/cluster_version_backlevel.go
```diff
@@ -0,0 +1,405 @@
...
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+var _ = Describe("Customer", func() {
+
+	backlevelVersions := []backlevelVersionSpec{
+		{
+			controlPlaneVersion: "4.20",
+			nodePoolVersions:    []string{"4.20.32"},
+			bicepModulesDir:     "test-artifacts/generated-test-artifacts/modules-4.20",
```
mbukatov, line 45: Do we assume we will have multiple back level versions there?

If yes, I see one issue with the ARO HCP API version, which will eventually prevent this to work. I assume that we should use API level which matches what was used at the time of given OCP version (eg. here, we use  20240610), and so it can happen that if we have 2 different versions there, we will need to use different API calls. The API is partially freezed in the bicep modules for 4.20 as well.

mbukatov (commented):
/lgtm

providing comment about one of the idea in the code we don't use yet

mvacula02 (approved):
/lgtm

test/e2e/cluster_version_backlevel.go
miquelsi: For now we only test one backlevel version.

PR #6523, opened 2026-08-12
Adding kasGoMemLimit for a small size clusters

mukrishn:
### What

kasGoMemLimit sets a soft cap on KAS memory during traffic surges. Without it, KAS on an HCP can grow without bound and trigger OOM kills on worker nodes. 

### Why

This is missing in one of the ClusterSize criteria 0-60, setting this limit equivalent to medium size cluster for now.


mukrishn:
@geoberle and @tony-schndr PTAL, GOMEMLIMIT is missing for small size clusters.

stevekuznetsov (approved):
/lgtm
/approve

PR #6526, opened 2026-08-12
fix: block Private KAS cluster creation without Swift networking

zgalor:
## Summary

- Private KAS (`api.visibility=Private`) requires Swift networking (`vnetIntegrationSubnetId` must be set)
- Adds cross-field validation in the shared validator that rejects Private KAS clusters without `vnetIntegrationSubnetId` on create
- For API versions `v20251223preview` and later, `vnetIntegrationSubnetId` is already mandatory during conversion, so this validation only has practical effect for `v20240610preview`

## Why

Currently it's possible to create a Private KAS cluster via the `v20240610preview` API without Swift networking (`vnetIntegrationSubnetId` not set). This cluster would fail at provisioning time since Private KAS requires Swift. This PR catches the misconfiguration early at the API validation layer.

## Jira

[ARO-28990](https://issues.redhat.com/browse/ARO-28990)

## Test plan

- [x] Unit tests added for both cases:
  - Private KAS + Swift (allowed)
  - Private KAS + non-Swift (blocked)
- [ ] Existing validation tests pass (`go test ./validation/...`)

### DIFF
--- a/internal/validation/validate_cluster.go
+++ b/internal/validation/validate_cluster.go
@@ -88,12 +88,37 @@ func ValidateCluster(ctx context.Context, op operation.Operation, newCluster, ol
 	// there several resourceIDs that must be verified with respect to this ID.  This is the only level of validation with access to both
 	errs = append(errs, validateResourceIDsAgainstClusterID(ctx, op, newCluster, oldCluster)...)
 
+	// Private KAS requires Swift networking (vnetIntegrationSubnetId must be set).
+	// For API versions v20251223preview and later, vnetIntegrationSubnetId is already
+	// enforced as required during conversion, so this check only has practical effect
+	// for v20240610preview.
+	errs = append(errs, validatePrivateKASRequiresSwift(ctx, op, newCluster)...)
+
 	// there are pieces of clusterProperties that are dependent upon values in .identity
 	errs = append(errs, validateOperatorAuthenticationAgainstIdentities(ctx, op, newCluster, oldCluster)...)
 
 	RewriteValidationFieldPaths(errs, validationPathMapper)
 
 	return errs
+}
+
+func validatePrivateKASRequiresSwift(_ context.Context, op operation.Operation, newCluster *coreapi.HCPOpenShiftCluster) field.ErrorList {
+	if op.Type != operation.Create {
+		return nil
+	}
+	if newCluster.CustomerProperties.API.Visibility != metadataapi.VisibilityPrivate {
+		return nil
+	}
+	vnetSubnet := newCluster.CustomerProperties.Platform.VnetIntegrationSubnetID
+	if vnetSubnet == nil || vnetSubnet.String() == "" {
+		return field.ErrorList{
+			field.Required(
+				field.NewPath("customerProperties", "platform", "vnetIntegrationSubnetId"),
+				"required when customerProperties.api.visibility is Private",
+			),
+		}
+	}
+	return nil
 }
 
 func validateOperatorAuthenticationAgainstIdentities(ctx context.Context, op operation.Operation, newCluster, _ *coreapi.HCPOpenShiftCluster) field.ErrorList {
--- a/internal/validation/validate_cluster.go
+++ b/internal/validation/validate_cluster.go
@@ -88,11 +88,11 @@ func ValidateCluster(ctx context.Context, op operation.Operation, newCluster, ol
 	// there several resourceIDs that must be verified with respect to this ID.  This is the only level of validation with access to both
 	errs = append(errs, validateResourceIDsAgainstClusterID(ctx, op, newCluster, oldCluster)...)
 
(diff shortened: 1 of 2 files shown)


zgalor:
@miguelsorianod PTAL

internal/validation/validate_cluster.go
```diff
@@ -96,6 +102,25 @@ func ValidateCluster(ctx context.Context, op operation.Operation, newCluster, ol
 	return errs
 }
 
+func validatePrivateKASRequiresSwift(_ context.Context, op operation.Operation, newCluster *coreapi.HCPOpenShiftCluster) field.ErrorList {
+	if op.Type != operation.Create {
```
miguelsorianod: This should be independent on whether it's create or not. newCluster needs to hold the invariant no matter if we are creating or updating.

internal/validation/validate_cluster.go
```diff
@@ -96,6 +102,25 @@ func ValidateCluster(ctx context.Context, op operation.Operation, newCluster, ol
 	return errs
 }
 
+func validatePrivateKASRequiresSwift(_ context.Context, op operation.Operation, newCluster *coreapi.HCPOpenShiftCluster) field.ErrorList {
```
miguelsorianod: Define errs := field.ErrorList{} and return that throughout the method, appending to it when necessary

internal/validation/validate_cluster.go
```diff
@@ -96,6 +102,25 @@ func ValidateCluster(ctx context.Context, op operation.Operation, newCluster, ol
...
+	if op.Type != operation.Create {
+		return nil
+	}
+	if newCluster.CustomerProperties.API.Visibility != metadataapi.VisibilityPrivate {
+		return nil
+	}
+	vnetSubnet := newCluster.CustomerProperties.Platform.VnetIntegrationSubnetID
+	if vnetSubnet == nil || vnetSubnet.String() == "" {
```
miguelsorianod: There's no need to check for empty string on vnetSubnet. That would be an invalid azcorearm.ResourceID.

internal/validation/validate_cluster.go
```diff
@@ -96,6 +102,25 @@ func ValidateCluster(ctx context.Context, op operation.Operation, newCluster, ol
 	return errs
 }
 
+func validatePrivateKASRequiresSwift(_ context.Context, op operation.Operation, newCluster *coreapi.HCPOpenShiftCluster) field.ErrorList {
+	if op.Type != operation.Create {
+		return nil
+	}
+	if newCluster.CustomerProperties.API.Visibility != metadataapi.VisibilityPrivate {
```
miguelsorianod: This can be changed to:

if newCluster.CustomerProperties.API.Visibility == private && vnetintegrationsubnet == nil
  errs = append(.... field.Required....)
}

return errs

We avoid early returns throughout the validation code

internal/validation/validate_cluster.go
miguelsorianod: Instead of Swift in the method name, name it VNetIntegrationSubnetID. We don't have any reference to swift over here

miguelsorianod: define oldCluster as parameter too, even when not used. This is consistent with how we do it in validateOperatorAuthenticationAgainstIdentities and it signals readers that both need to be considered in general

internal/validation/validate_cluster.go
zgalor: Both fields are create-only, but I can remove the condition

miguelsorianod: > Both fields are create-only

That doesn't matter. When you perform an update, the validation logic is executed anyway. We should remove exceptions depending on operation when possible. Here what matters is that the desired state of the cluster (newCluster) is the one we want, independently on whether it's create or update

internal/validation/validate_cluster.go
```diff
@@ -96,6 +102,20 @@ func ValidateCluster(ctx context.Context, op operation.Operation, newCluster, ol
...
+	if newCluster.CustomerProperties.API.Visibility == metadataapi.VisibilityPrivate &&
+		newCluster.CustomerProperties.Platform.VnetIntegrationSubnetID == nil {
+		errs = append(errs, field.Required(
+			field.NewPath("customerProperties", "platform", "vnetIntegrationSubnetId"),
+			"required when customerProperties.api.visibility is Private",
+		))
+	}
+
+	return errs
+}
```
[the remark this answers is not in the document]
miguelsorianod: > any existing Private KAS clusters created without vnetIntegrationSubnetId (the case this PR is fixing)

Even though this could happen at the RP level, we have the validation in CS, so attempting to create a non-swift cluster with private KAS would end up with a non provisioned cluster.

Regarding update, both visibility and vnetintegrationsubnetid are immutable at the RP level, so the potential concern at update time doesn't exist

PR #6533, opened 2026-08-13
test: calculate latest z-stream for specific OCP version

miquelsi:
### What

This change adds support for ARO_HCP_OPENSHIFT_LATEST_Z_STREAM. When it is set to true together with ARO_HCP_OPENSHIFT_CONTROLPLANE_VERSION, the E2E framework no longer uses the configured value as a fixed install version. Instead, resolveDefaultControlPlaneVersion parses the configured major.minor (or full semver), looks up the latest z-stream for that minor in the active channel group, and uses that resolved version for default cluster params. Without the flag, an explicitly set control plane version continues to be used as-is.

### Why

This is necessary because CI and local runs often need to target a specific OpenShift minor (for example 4.20) while still installing the newest available patch in that line. Pinning only ARO_HCP_OPENSHIFT_CONTROLPLANE_VERSION=4.20 previously left the version unresolved when the env var was set, so tests could install an incomplete or outdated identifier, or required operators to hard-code a full z-stream and keep updating it. The new flag lets callers choose the minor and automatically track the latest z-stream, which reduces maintenance and keeps E2E coverage on current releases.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -75,6 +75,17 @@ $ export ARO_HCP_OPENSHIFT_CONTROLPLANE_VERSION=4.20
 $ export ARO_HCP_OPENSHIFT_NODEPOOL_VERSION=4.20.15
 ```
 
+When `ARO_HCP_OPENSHIFT_CONTROLPLANE_VERSION` is set, you can also set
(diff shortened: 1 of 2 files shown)


mvacula02 (approved):
/lgtm

PR #6535, opened 2026-08-13
feat(mgmt-agent): add capacity reporting controller

geoberle:
### What

Add a capacity reporting controller that periodically collects node allocatable resources, pod resource requests and usage (via Metrics API), and HostedControlPlane counts. It writes a CapacityReport custom resource via Server-Side Apply so fleet-level tooling can observe per-cluster capacity without direct management-cluster access.

```yaml
apiVersion: mgmtagent.aro-hcp.azure.com/v1alpha1
kind: CapacityReport
metadata:
  name: cluster
status:
  hostedControlPlanes:
    notReady: 0
    ready: 1
  lastReportedAt: "2026-08-13T16:07:22Z"
  nodes:
  - allocatable:
      aro.openshift.io/swift-nic: "3"
      cpu: 11580m
      memory: 36258028Ki
    notReady: 0
    ready: 3
    sku: Standard_D4s_v3
  requested:
    aro.openshift.io/swift-nic: "3"
    cpu: 3375m
    memory: 9093Mi
  usage:
    aro.openshift.io/swift-nic: "3"
    cpu: 988m
    memory: 7781336Ki
  conditions:
  - lastTransitionTime: "2026-08-13T13:41:02Z"
    message: ""
    observedGeneration: 0
    reason: DataCollected
    status: "True"
    type: ReportCurrent
```

Includes:
- CapacityReport CRD (v1alpha1) with generated clientset, informers, listers
- Controller with 30s sync interval and 25s data collection timeout
- Prometheus metrics: sync_duration_seconds histogram, sync_errors_total counter
- Alert: MgmtAgentCapacityReportingSyncFailing (warning, fires after sustained failures)

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


mgmt-agent/pkg/controller/capacityreporting/controller.go
```diff
@@ -0,0 +1,462 @@
...
+	for sku, data := range bysku {
+		result = append(result, capacityreportv1alpha1.NodeSKUCapacity{
+			SKU:         sku,
+			Ready:       data.ready,
+			NotReady:    data.notReady,
+			Allocatable: data.allocatable,
+		})
+	}
+	sort.Slice(result, func(i, j int) bool {
+		return result[i].SKU < result[j].SKU
```
stevekuznetsov: strings.compare?

mgmt-agent/pkg/controller/capacityreporting/controller.go
```diff
@@ -0,0 +1,462 @@
...
+	total := corev1.ResourceList{}
+	for _, pod := range pods {
+		if !strings.HasPrefix(pod.Namespace, ocmNamespacePrefix) {
+			continue
+		}
+		if pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed {
+			continue
+		}
+		for _, container := range pod.Spec.Containers {
+			addResourceList(total, filterTrackedResources(container.Resources.Requests))
```
stevekuznetsov: why bother filtering?

mgmt-agent/pkg/controller/capacityreporting/controller.go
```diff
@@ -0,0 +1,462 @@
...
+
+// aggregatePodRequests sums resource requests across containers of
+// non-terminal pods in ocm-* namespaces.
+func aggregatePodRequests(pods []*corev1.Pod) corev1.ResourceList {
+	total := corev1.ResourceList{}
+	for _, pod := range pods {
+		if !strings.HasPrefix(pod.Namespace, ocmNamespacePrefix) {
+			continue
+		}
+		if pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed {
```
stevekuznetsov: There are other sneaky phases IIRC, so probably better to allow-list the includes than playing whack-a-mole with excludes?

mgmt-agent/pkg/controller/capacityreporting/controller.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	return total
+}
+
+// aggregatePodMetrics sums actual resource usage from PodMetrics for pods in
+// ocm-* namespaces.
+func aggregatePodMetrics(metrics []metricsv1beta1.PodMetrics) corev1.ResourceList {
+	total := corev1.ResourceList{}
+	for _, podMetric := range metrics {
+		if !strings.HasPrefix(podMetric.Namespace, ocmNamespacePrefix) {
```
stevekuznetsov: you should be able to get the authoritative set of namespaces from where we find `HostedCluster` objects, so I wouldn't do this - I don't know if our silly `ocm-` prefix continues to exist once open-cluster-management is removed from the project

stevekuznetsov: s/`HostedCluster`/`HostedControlPlane`/g

mgmt-agent/pkg/controller/capacityreporting/controller.go
```diff
@@ -0,0 +1,462 @@
...
+		if isHCPAvailable(hcp) {
+			count.Ready++
+		} else {
+			count.NotReady++
+		}
+	}
+	return count
+}
+
+// computeAverageHCPResourceUsage divides total usage by the number of ready HCPs.
```
stevekuznetsov: if we expose the total and the count, why do we want to pre-compute this as opposed to letting the client do what they need after their readesire shows them the info?

mgmt-agent/pkg/controller/capacityreporting/controller.go
```diff
@@ -0,0 +1,462 @@
...
+	result := corev1.ResourceList{}
+	for resourceName, quantity := range usage {
+		switch resourceName {
+		case corev1.ResourceCPU:
+			result[resourceName] = *resource.NewMilliQuantity(quantity.MilliValue()/int64(readyCount), resource.DecimalSI)
+		case corev1.ResourceMemory:
+			result[resourceName] = *resource.NewQuantity(quantity.Value()/int64(readyCount), resource.BinarySI)
+		}
+	}
+	// every HCP needs exactly `swiftNICsPerHCP` NICs.
```
stevekuznetsov: maybe safer to backfill with the const if and only if the pods are missing the data, to let us keep the source of truth otherwise?

mgmt-agent/pkg/controller/capacityreporting/controller.go
geoberle: filtering now for running and pending

mgmt-agent/pkg/controller/capacityreporting/controller.go
geoberle: removed the avg

PR #6538, opened 2026-08-13
Use new upgrade scheme

deads2k:
built on top of https://github.com/Azure/ARO-HCP/pull/6513


backend/pkg/controllers/controlplaneversion/cincinnati.go
```diff
@@ -0,0 +1,139 @@
...
+			Version: node.Version,
+			Image:   node.Payload,
+		})
+	}
+
+	slices.SortFunc(releases, func(a, b configv1.Release) int {
+		vA := semver.MustParse(a.Version)
+		vB := semver.MustParse(b.Version)
+		return -vA.Compare(vB)
+	})
```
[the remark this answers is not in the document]
machi1990: can't happen

internal/admission/admit_cluster.go
```diff
@@ -87,9 +87,101 @@ func MutateCluster(ctx context.Context, admissionContext *ClusterAdmissionContex
...
+	if admissionContext.OriginalCluster != nil {
+		tags = admissionContext.OriginalCluster.Tags
+	}
+	tagPresent := hasTag(tags, metadataapi.TagClusterControlPlaneExactVersion)
+	tagValue := lookupTag(tags, metadataapi.TagClusterControlPlaneExactVersion)
+
+	versionID := newObj.CustomerProperties.Version.ID
+	// version.id carries an exact pin only when it includes a patch component
+	// (e.g. "4.17.3"); a bare "<major>.<minor>" does not.
+	versionIDHasPatch := len(versionID) > 0 && len(strings.SplitN(versionID, ".", 3)) == 3
```
machi1990: versionIDHasPatch will be false if a nightly version or a prerelease version e.g *"5.0.0-0.nightly-multi-2026-07-09-124132"* or `5.0.0-ec.6` are given

Alternative could be to user semver.Parse(versionID), if it errors, then it wasn't a complete version. If it didn't then it was a full x.y.z version pin.

machi1990: A bit similar to what's there in line 169

backend/pkg/controllers/cluster/version/control_plane_version_controller.go
```diff
@@ -0,0 +1,517 @@
...
+
+// NeedsWork reports whether this controller applies to the given cluster. It acts on any cluster
+// whose customer specified both a desired minor version and a channel group: it seeds the initial
+// desired control plane version when none is set yet, and advances it (z-stream/y-stream) once one
+// has been seeded.
+func (c *controlPlaneVersionSyncer) NeedsWork(cluster *coreapi.HCPOpenShiftCluster, serviceProviderCluster *coreapi.ServiceProviderCluster) bool {
+	if serviceProviderCluster == nil {
+		return false
+	}
+	if len(cluster.CustomerProperties.Version.ID) == 0 {
```
machi1990: version.id is required https://github.com/deads2k/ARO-HCP/blob/57744b8526c5ecc952405156caaed7d8e50acea9/internal/validation/validate_cluster.go#L386 we don't need this

backend/pkg/controllers/cluster/version/control_plane_version_controller.go
```diff
@@ -0,0 +1,517 @@
...
+		return false
+	}
+	if len(cluster.CustomerProperties.Version.ChannelGroup) == 0 {
```
machi1990: version.channelGroup is required https://github.com/deads2k/ARO-HCP/blob/57744b8526c5ecc952405156caaed7d8e50acea9/internal/validation/validate_cluster.go#L413 we can drop this check

backend/pkg/controllers/cluster/version/control_plane_version_controller.go
```diff
@@ -0,0 +1,517 @@
...
+
+	if desiredMinorVersion.GT(actualLatestMinorVersion) {
+		if err := validation.OpenshiftVersionAtMostOneMinorSkew(actualLatestMinorVersion.String(), desiredMinorVersion.String()); err != nil {
+			return utils.TrackError(err)
+		}
+		clusterNodePools, ready, err := c.listClusterAdmissionNodePools(ctx, key.GetResourceID())
+		if err != nil {
+			return utils.TrackError(err)
+		}
+		if !ready {
```
machi1990: This skip no longer holds it's meaning; previous it used to meant that, do not compute desired version. Now it means; the minor version change is valid and such https://github.com/deads2k/ARO-HCP/blob/57744b8526c5ecc952405156caaed7d8e50acea9/backend/pkg/controllers/cluster/version/control_plane_version_controller.go#L224 won't error and we'll continue with the version resolution logic.

We've two options;
- Either return an error 
- or return a bool indicating that we want a requeue

backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go
```diff
@@ -0,0 +1,517 @@
...
+// stableChannelGroup is the OpenShift channel group of generally-available stable releases.
+const stableChannelGroup = "stable"
+
+// defaultControlPlaneVersionOffset is the offset from the tip of a channel used when selecting a
+// desired control plane version. Offset 0 selects the most recent (tip) release in the channel.
+const defaultControlPlaneVersionOffset uint = 0
+
+// GetZStreamOffset returns the offset from the tip of a channel to use when selecting a z-stream
+// control plane version (offset 0 = the most recent release). The stable channel group uses the
+// penultimate release (offset 1) to avoid selecting a brand-new stable z-stream immediately; all
```
machi1990, line 420: > The stable channel group uses the
// penultimate release (offset 1) to avoid selecting a brand-new stable z-stream immediately; all

What's the impact of this choice in regards to CVEs fixes available only in latest releases and previous ones? I can think of; it means that we'll at least wait 2 weeks to get the CVE fix when the latest becomes the previous last? Is that correct and intentional?

test/e2e/control_plane_automated_z_stream_upgrade.go
```diff
@@ -133,13 +154,10 @@ var _ = Describe("Service Provider", func() {
...
 
-		// for 4.19, if we start with 4.19.0, the version that has an upgrade path to 4.19.latest is
-		// 4.19.3 but cluster install on this version fails with KMS authentication problem.
-		// For all the other minor versions, we can start with 4.y.0 and install the latest version in the candidate channel.
-		Entry("for 4.19", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.19", "4.19.25"),
-		Entry("for 4.20", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.20", ""),
-		Entry("for 4.21", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.21", ""),
-		Entry("for 4.22", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, labels.AllowRetry, "4.22", ""), // owner: @raelga, tracking: AROSLSRE-1319. Known-issue test, retriable during EV2 gating. Remove this label when the issue is fixed.
-		Entry("for 4.23", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.23", ""),
+		Entry("for 4.19", labels.RequireNothing, labels.Critical, labels.Positive, labels.AroRpApiCompatible, "4.19"),
```
machi1990, line 157: np: We can remove the 4.19 entry and add 5.0 all the way to 5.12 entry;

test/e2e/control_plane_automated_z_stream_upgrade.go
machi1990, line 152: let's enhance https://github.com/deads2k/ARO-HCP/blob/main/test/util/verifiers/hosted_control_plane_version.go#L44 verifier to also check that there are at least 2 unique versions;
- installVersion (the one we pinned the cluster to)
- the latest version within the same minor (the one we automatically upgraded to)

backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go
```diff
@@ -0,0 +1,517 @@
...
+
+// selectControlPlaneVersion resolves the desired control plane version for a minor by querying the
+// OpenShift update service (Cincinnati graph API) for that minor's channel and selecting the release
+// at the given offset from the tip (offset 0 = most recent release). It bridges
+// controlplaneversion.SelectControlPlaneVersion, which returns a configv1.Release, into the
+// *semver.Version used throughout this controller.
+//
+// This must not be called for the nightly channel group: the graph API does not serve nightly
+// builds. See the nightly channel guard in SyncOnce.
+func selectControlPlaneVersion(ctx context.Context, roundTripper controlplaneversion.RoundTrip, channelGroup string, minorVersion semver.Version, offset uint) (*semver.Version, error) {
```
machi1990, line 437: love the simplicity brought by this change! Awesome work @deads2k @wking

machi1990 (changes requested):
Love the direction we are taking with this, thanks for the improvements. Left some comments for clarification and enhancement.

Also CI isn't happy with things like
```
# github.com/Azure/ARO-HCP/test/util/framework
util/framework/deployment_params.go:94:21: undefined: GetLatestInstallVersion
make: Leaving directory '/opt/app-root/src/github.com/Azure/ARO-HCP/test'
```
Once comments addressed and CI fixed, I'll take another look and tag it

backend/pkg/controllers/controlplaneversion/recency_based_control_plane_version_selection.go
machi1990, line 1: aro.go for an name initially caught my eye for a name; 
I can think of a 
```
recency_based_control_plane_version_selection.go
```
as an alternative

backend/pkg/controllers/controlplaneversion/recency_based_control_plane_version_selection_test.go
machi1990, line 1: similar comment as https://github.com/Azure/ARO-HCP/pull/6538/changes#r3783667622

backend/pkg/controllers/controlplaneversion/recency_based_control_plane_version_selection.go
```diff
@@ -0,0 +1,44 @@
...
+		return nil, err
+	}
+	if len(releases) == 0 {
+		return nil, fmt.Errorf("no releases found in %s", updateService)
+	}
+	if int(offset) >= len(releases) {
+		return nil, fmt.Errorf("%d releases found in %s, which is not enough for the requested %d offset", len(releases), updateService, offset)
+	}
+	return &releases[offset], nil
+}
```
machi1990, line 56: In this method; do we need to add filtering of `releases` to only those matching the desired major.minor?

e.g to avoid picking a version from an older release line e.g minor-1 

For example 

```
 curl https://api.openshift.com/api/upgrades_info/graph\?channel\=stable-4.21\&arch\=multi | jq '.nodes' 
```

returns 4.20 versions too; which I believe we'll be better off not returning here? 

```
{
    "metadata": {
      "io.openshift.upgrades.graph.previous.remove_regex": "4[.]19[.].*|^4[.]19[.]([0-9]|1[0-7])[+].*$|.*",
      "io.openshift.upgrades.graph.release.channels": "candidate-4.20,eus-4.20,fast-4.20,stable-4.20,candidate-4.21,fast-4.21,stable-4.21,candidate-4.22,eus-4.22,fast-4.22,stable-4.22",
      "io.openshift.upgrades.graph.release.manifestref": "sha256:1f1db6ace88613a5253824d9b8933232b27f39a8d13f43d150f6c379ec4933e9",
      "release.openshift.io/architecture": "multi",
      "url": "https://access.redhat.com/errata/RHSA-2025:22257"
    },
    "payload": "quay.io/openshift-release-dev/ocp-release@sha256:1f1db6ace88613a5253824d9b8933232b27f39a8d13f43d150f6c379ec4933e9",
    "version": "4.20.6"
  },
```

What do you think? Or the intent if for the filtering to be done in the caller - similar to https://github.com/deads2k/ARO-HCP/blob/57744b8526c5ecc952405156caaed7d8e50acea9/backend/pkg/controllers/cluster/version/control_plane_version_controller.go#L247 

If so, I think we'll need another enhancement in the controller to prevent things like; an offset being large enough we find ourselves picking a 4.y-1 version for 4.y install. https://github.com/deads2k/ARO-HCP/blob/57744b8526c5ecc952405156caaed7d8e50acea9/backend/pkg/controllers/cluster/version/control_plane_version_controller.go#L255
i.e the picked version must be in the target minor

backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go
```diff
@@ -0,0 +1,517 @@
...
+		return clearIntentFailed(ctx, c.resourcesDBClient, key, controlPlaneDesiredVersionControllerName)
+	case previousDesiredVersion != nil && desiredVersion.LE(*previousDesiredVersion):
+		// Newly resolved version is not greater than the stored desired; leave it unchanged.
+		return clearIntentFailed(ctx, c.resourcesDBClient, key, controlPlaneDesiredVersionControllerName)
+	default:
+		logger.Info("Selected desired version", "desiredVersion", desiredVersion, "previousDesiredVersion", previousDesiredVersion)
+	}
+
+	replacement := cachedServiceProviderCluster.DeepCopy()
+	replacement.Spec.ControlPlaneVersion.DesiredVersion = ptr.To(*desiredVersion)
```
machi1990, line 247: See https://github.com/Azure/ARO-HCP/pull/6538/changes#r3783748889 about considering to return an error if the picked version isn't in the same target major.minor release line

backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go
deads2k: > it means that we'll at least wait 2 weeks to get the CVE fix when the latest becomes the previous last? Is that correct and intentional?

Yes to both.  Someone choosing stable is choosing to stay behind. We have an option of forcing the fleet regardless (separate and future change), but for automatic it'll be one behind.

test/e2e/control_plane_automated_z_stream_upgrade.go
deads2k: > np: We can remove the 4.19 entry and add 5.0 all the way to 5.12 entry;

pain to regenerate. maybe later.

deads2k:
/retest

Looks like ACM didn't see an API URL in a cluster claim and not in ManagedClusterClientConfigs, but HCP thinks the cluster already finished installing?  And it's not at all obvious why/how the version change factors in

deads2k:
Holy crap that worked.  I'll re-nerf the 4.22 to 5.0 upgrade and we can proceed.

machi1990:
> Holy crap that worked. 

🎉 nice!

> I'll re-nerf the 4.22 to 5.0 upgrade and we can proceed.

This failure is expected; CS needs to be fixed @YuriyNik has the work on his rader

See the previous skip 

>> Skip("CS does not support major upgrade yet; skipping until https://redhat.atlassian.net/browse/ARO-25230 is resolved")

test/util/framework/deployment_params.go
```diff
@@ -62,48 +65,79 @@ var (
...
+			defaultCPVersionErr = fmt.Errorf("failed getting controlPlaneVersion: %w", err)
+			return
 		}
-		defaultCPVersion = version
+		if release == nil {
+			defaultCPVersionErr = fmt.Errorf("no release found for controlPlaneVersion: %w", err)
+			return
+		}
+
+		defaultCPVersion = resultingControlPlaneMinorVersion
```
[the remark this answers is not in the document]
deads2k: it's doing that it should do.

backend/pkg/controllers/controlplaneversion/cincinnati.go
```diff
@@ -0,0 +1,139 @@
...
+
```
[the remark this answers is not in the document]
deads2k: we won't break every OCP install int eh world.

machi1990:
/lgtm

🚢

PR #6541, opened 2026-08-13
e2e: actually attempt to use the cluster autoscaler

bennerv:
### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/e2e/nodepool_autoscaling_trigger.go
+++ b/test/e2e/nodepool_autoscaling_trigger.go
@@ -1,0 +1,242 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+)
+
+var _ = Describe("Customer", func() {
+	BeforeEach(func() {
+		// do nothing. per test initialization usually ages better than shared.
+	})
+
+	It("should trigger nodepool autoscaling when pods exhaust single-node resources",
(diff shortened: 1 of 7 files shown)


test/e2e/nodepool_autoscaling_trigger.go
```diff
@@ -0,0 +1,242 @@
...
+		labels.RequireNothing,
+		labels.Medium,
+		labels.Positive,
+		labels.AroRpApiCompatible,
+		labels.MIContainers(1),
+		func(ctx context.Context) {
+			const (
+				customerClusterName        = "autoscale-trigger"
+				customerNodePoolName       = "stress-pool"
+				autoscalingMin       int32 = 1
```
patriksuba, line 53: **Heads up: the "starts with 1 node" assertion is racing against the in-cluster image registry.**

`openshift-image-registry` runs 2 replicas with pod anti-affinity across nodes. On a fresh `min=1` nodepool, those 2 registry pods can't co-schedule on the single starting node, so the in-cluster autoscaler will often scale `stress-pool` from 1→2 on its own, *before* the test's stress load is ever deployed — independent of anything this test is doing. When that race loses, the "verifying the nodepool starts with 1 node" check (and the later "scale to 2 nodes" check, which is trivially already true) fail even though every ARO-HCP service behaved correctly.

Timeline from a repro:
- nodepool created with autoscaling min=1, max=2
- CS creates the MachineDeployment with replicas=1
- ~5 min later, the autoscaler finds the registry's Pending pod and bumps stress-pool 1→2
- second node comes up before the test's stress Deployment is even created
- "starts with 1 node" assertion sees 2 and fails

A couple of ways to make this deterministic instead of racy — happy to go with whichever fits best, or something else:

1. **Bump the baseline (`min=2, max=3`) instead of `min=1, max=2`.** With 2 nodes provisioned up front, the registry's 2 anti-affine replicas land one-per-node with no autoscaler involvement, removing the race entirely. But the stress load needs to scale with it: 2 nodes at 8 vCPU each can already hold up to 4 pods of 3 CPU (2/node), which is exactly today's `stressReplicas := 4` — so that would no longer reliably force a Pending pod. Suggest bumping `stressReplicas` to 6 (3 nodes × 2 pods/node), preserving the same "N replicas guarantees Pending regardless of other overhead" property the test currently relies on, just shifted by one node.

2. **Disable the image registry for this test** (`clusterImageRegistry.state: Disabled`, same pattern as `image_registry_cluster_create.go`). Removes the confound entirely rather than out-guessing its footprint, and keeps `min=1, max=2` / `stressReplicas=4` as-is. Downside: the test no longer exercises a "normal" cluster shape (registry present), so it diverges slightly from what a real customer nodepool looks like.

test/e2e/nodepool_autoscaling_trigger.go
```diff
@@ -0,0 +1,242 @@
...
+			Expect(npResp.Properties.AutoScaling.Max).To(Equal(to.Ptr(autoscalingMax)),
+				"expected autoscaling max to be %d", autoscalingMax)
+
+			By("verifying the nodepool starts with 1 node")
+			nodes, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred(), "failed to list nodes after nodepool creation")
+			poolNodes, err := framework.SelectNodesBelongingToNodePool(nodes.Items, customerNodePoolName)
+			Expect(err).NotTo(HaveOccurred(), "failed to select nodes for nodepool %s", customerNodePoolName)
+			Expect(len(poolNodes)).To(Equal(int(autoscalingMin)),
+				"expected nodepool %s to start with %d node(s)", customerNodePoolName, autoscalingMin)
```
patriksuba, line 157: Could you please add actual count? Optionally you can use verifiers for that either VerifyNodePoolReadyAndSchedulableNodeCount or VerifyNodeCount and VerifyNodesReady.

PR #6515, opened 2026-08-11
api: scaffold new API version 2026-10-01-preview

twolff-gh:
## Summary
- Scaffold the `2026-10-01-preview` API version as an identical copy of `2026-09-01-preview` with no feature changes, establishing the entry point for incremental feature additions in follow-up PRs
- Add an E2E cluster create test with a `2026-11-01` timebomb: it skips while the API version is not yet available/deployed, and hard-fails once the deadline passes if the API is still missing (mirrors the existing `2026-09-01-preview` test's gating)

## Test plan
- [x] `make lint` — 0 issues
- [x] `make verify` — passes
- [x] `make tidy` / `make fmt` — clean
- [x] `go test ./internal/azureapi/v20261001preview/...` — pass
- [x] `go test ./internal/api/coreapi/ -run TestTypeSpecDefaultsConsistency` — pass
- [x] `make -C api validate-examples` / `lint-openapi` — pass
- [x] `go build -tags E2Etests ./test/...` — compiles
- [x] `go test -tags E2Etests ./test/cmd/aro-hcp-tests/` — pass (suite-listing fixtures regenerated for the new spec)
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6547, opened 2026-08-14
feat(api): expose activeVersions in v20261001preview ARM API

ahitacat:
https://redhat.atlassian.net/browse/ARO-28915 
https://redhat.atlassian.net/browse/ARO-28916

### What

Expose the active control plane and node pool versions to customers via `properties.status.activeVersions` in the `v20261001preview` ARM API. Add the `ActiveVersion` model to TypeSpec, regenerate the OpenAPI spec and SDK. Populated the activeVersions `HCPOpenshiftCluster/HCPOpenshiftNodePool` at the same time as it was populated in the `ServiceProviderCluster/ServiceProviderNodePool`

Differentiated activeVersion for external and internal representation of the resource. Renaming `HCPOpenShiftClusterActiveVersion` to `ServiceProviderClusterActiveVersion`. Nodepool were also renamed.

### Why

Customers currently have no ARM API visibility into which versions are actively running on their cluster control plane and node pools. This is especially important during upgrades,  where multiple versions may coexist. Surfacing activeVersions in the status lets customers monitor upgrade progress and confirm completion through the standard ARM API.

### Testing

- Unit tests: existing conversion and deepcopy tests cover the new fields.
- Integration tests: GET cluster and node pool responses include activeVersions when SP documents have version data.
- E2E tests: upgrade tests should be updated to assert Properties.Status.ActiveVersions contains the target version after upgrade (follow-up PR once the new API is rolled out up to prod).

- Used v20260901preview of the API, it should be done in the latest WIP API version (need to be checked). Confirmed that the version should be v20261001preview, Rebased all the previous work into that version.
- Found a bug during verifying this test in hypershift. It causes that during nodepool upgrades the activeVersions are the same as the targetVersion not having a meaningful value. https://redhat.atlassian.net/browse/OCPBUGS-109668
- I have considered fill the activeVersions during Get/List methods by getting the information from the ServiceProvider Cosmo Documents. However, I discarded it because it add some extra calls to the CosmosDB during these methods. The current approach follows the same as the conditions.

### PR Checklist

- [x] Self-reviewed the diff


api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -1087,13 +1087,27 @@ model ExternalAuthProperties {
...
-@added(Versions.v2026_06_30_preview)
 model ResourceStatus {
   /** The conditions on the resource */
+  @added(Versions.v2026_06_30_preview)
   @visibility(Lifecycle.Read)
   @identifiers(#["type"])
   @maxItems(10)
   conditions?: Condition[];
+
+  /** The activeVersions of a resource */
```
deads2k: Add details for the doc. Indicate that it is a slice because during an upgrade more than one version will be active at the same time and both are important for making decisions about what versions of nodepools are a good idea.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -1087,13 +1087,27 @@ model ExternalAuthProperties {
...
+  @added(Versions.v2026_09_01_preview)
+  @visibility(Lifecycle.Read)
+  @identifiers(#[])
+  activeVersions?: ActiveVersion[];
+}
+
+/** ActiveVersion represents a version of a resource. */
+@added(Versions.v2026_09_01_preview)
+model ActiveVersion {
+  /** The version of the resource */
```
deads2k: indicate that this will be major.minor, same as the input of the field.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -1087,13 +1087,27 @@ model ExternalAuthProperties {
 }
 
 /** ResourceStatus represents the observed status of the resource. */
-@added(Versions.v2026_06_30_preview)
 model ResourceStatus {
```
deads2k, line 1105: huh, is this shared between nodepools and cluster?  probably want different ones at this point.

internal/api/coreapi/types_cluster.go
```diff
@@ -69,6 +69,12 @@ type HCPOpenShiftClusterStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	UserFacingConditions []metav1.Condition `json:"userFacingConditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// ActiveVersions are the active versions of the cluster control plane, ordered newest first.
+	// Written by: ControlPlaneActiveVersions
+	// +optional
+	// +listType=slice
+	ActiveVersions []HCPClusterActiveVersion `json:"activeVersions,omitempty"`
```
deads2k, line 79: Good choice of location.  In external types (ones that projects outside of yours may depend on), never share types like this across cluster,nodepool,serviceprovidercluster,etc: if you do so, when you need new fields you end up breaking everyone's compilation by adding new types that you didn't need to break.  But since this purely internal, we can do so.

ahitacat:
Manual tested this. Cluster view will show under status the versions:

```
"status": {
            "activeVersions": [
                {
                    "version": "4.22"
                }
            ],
            "conditions": [
                {
                    "lastTransitionTime": "2026-08-14T10:26:44Z",
                    "message": "",
                    "reason": "Valid",
                    "status": "True",
                    "type": "RequirementsValid"
                }
            ]
  },
```

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
ahitacat: Addressed, added some details on the expected values.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
ahitacat: done

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
ahitacat: Created different types for Clusters and Nodepool. Leaving ResourceStatus for any other resource like ExternalAuth.

internal/api/coreapi/types_cluster.go
ahitacat: Avoided to share internal types to external types. I have renamed the internal types to be ServiceProvider<Resource>ActiveVersions, and HCP<Resource>ActiveVersion are now the external types.

This way we can decouple it, so in the future we can add different information for the external type (user-facing object) and the internal type, the one the service use to do the logic and validation.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -140,7 +140,14 @@ model HcpOpenShiftClusterProperties {
   /** Status of the cluster resource */
   @visibility(Lifecycle.Read)
   @added(Versions.v2026_06_30_preview)
-  status?: ResourceStatus;
+  @renamedFrom(Versions.v2026_09_01_preview, "status")
```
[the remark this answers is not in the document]
mbarnes: I think Copilot is getting confused here.

The field was added in 2026-06-30-preview, but then changed to a cluster-specific status model in 2026-09-01-preview, which is backward-compatible with `ResourceStatus`.

Looks fine to me from a TypeSpec and Swagger standpoint.  The change in model names might introduce a breaking change in SDKs, but I'm not seeing a way around that.  We're moving from a shared model to a not-shared model.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -1090,10 +1104,61 @@ model ExternalAuthProperties {
...
+
+/** ClusterResourceStatus represents the observed status of the cluster resource. */
+@added(Versions.v2026_09_01_preview)
+model ClusterResourceStatus {
+  /** The conditions on the resource */
+  @added(Versions.v2026_09_01_preview)
   @visibility(Lifecycle.Read)
   @identifiers(#["type"])
   @maxItems(10)
   conditions?: Condition[];
```
mbarnes: Optional, but another way to do this would be to "[spread](https://typespec.io/docs/language-basics/models/#spread)" `ResourceStatus` inside `ClusterResourceStatus`.  Same for `NodePoolResourceStatus`.

Only potential issue is the `@maxItems` would not be independently adjustable for clusters and node pools. But this being a _read-only_ field, I don't think we even need `@maxItems` here.  These decorators are meant to limit _input_.  There's no need to restrict ourselves to an output limit that we might have to break in the future.

I'd say either drop `@maxItems` here and in `NodePoolResourceStatus` or, if you choose to do the spread, redefine the field in `ResourceStatus` without `@maxItems` (use versioning decorators).

```suggestion
  ...ResourceStatus
```

mbarnes (changes requested):
Just looked at the TypeSpec for a first pass.  Asked for a small change.

I'll continue reviewing the rest but just wanted to give some quick feedback.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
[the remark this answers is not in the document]
ahitacat: Yes, I was following the same pattern as we used in the past to change the visibility.

backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
```diff
@@ -211,3 +242,12 @@ func getHostedClusterDesiredVersionChannels(hostedCluster *hsv1beta1.HostedClust
...
 }
+
+func hcpClusterActiveVersionFromServiceProviderActiveVersions(activeVersions []coreapi.ServiceProviderClusterActiveVersion) []coreapi.HCPClusterActiveVersion {
+	var versions []coreapi.HCPClusterActiveVersion
+	for _, activeVersion := range activeVersions {
+		parsedVersion := semver.Version{Major: activeVersion.Version.Major, Minor: activeVersion.Version.Minor}
+		versions = append(versions, coreapi.HCPClusterActiveVersion{Version: &parsedVersion})
+	}
+	return versions
+}
```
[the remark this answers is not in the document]
ahitacat: Verson cannot be nil at this point, this method is used once the activeVersions.Version are parsed and populated .

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
ahitacat: maxItems were there so I didn't touch it. I just copy pasted it. I will go with your approach of use the spread and remove the maxItems from the ResourceStatus

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
mbarnes: > Leaving ResourceStatus for any other resource like ExternalAuth.

For the sake of SDK stability, it wouldn't hurt to define `ExternalAuthResourceStatus` now and spread `ResourceStatus` into it.  Not a blocking request though.

mbarnes (approved):
Informal LGTM for the moment.  I'm not very familiar with the RP backend anymore so I only gave that a cursory look.  Someone else should review the backend in more detail.  But I can tag this if you need me to.

Mild suggestion for future PRs is consider splitting things like type renames into separate commits, just to make reviews a little easier.  If I see that commits are well structured I like to review one at a time.  Easier to digest that way.

backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
```diff
@@ -129,22 +130,52 @@ func (c *controlPlaneActiveVersionSyncer) SyncOnce(ctx context.Context, key cont
...
+	if err := c.writeActiveVersionsToCluster(ctx, key, newActiveVersions); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *controlPlaneActiveVersionSyncer) writeActiveVersionsToCluster(ctx context.Context, key controllerutils.HCPClusterKey, activeVersions []coreapi.ServiceProviderClusterActiveVersion) error {
+	clusterCRUD := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName)
+	freshCluster, err := clusterCRUD.Get(ctx, key.HCPClusterName)
```
machi1990: Why do we need to get a fresh cluster again? 
Can't we use https://github.com/ahitacat/ARO-HCP/blob/7f28c64e8c5933668c108968a1c970ccfbba9aef/backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go#L83
```
existingCluster, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
```
if it's outdated, replace will fail and it'll be requeued

backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
machi1990, line 244: What does this return if activeVersions = [{version: 4.22.10, state=Partial}, {version: 4.22, state=Completed}]?

Will it be `[4.22, 4.22]`?

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
machi1990: @deads2k @ahitacat Why not returning the full X.Y.Z so that the user sees the full resolved version for their X.Y input? This caters for both cases where the user input was a full semver version e.g nightly channel or X.Y; And it gives the user information that isn't current visible for them. 

For now, this is purely information as they don't need it for anything but it answers the question of; what's the latest full version my cluster is running on.

backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
machi1990: I left a comment asking if we'll be better off returning the full X.Y.Z version

backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
```diff
@@ -129,22 +130,52 @@ func (c *controlPlaneActiveVersionSyncer) SyncOnce(ctx context.Context, key cont
...
+		serviceProviderClustersCosmosClient := c.resourcesDBClient.ServiceProviderClusters(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+		_, err = serviceProviderClustersCosmosClient.Replace(ctx, replacement, nil)
+		if err != nil {
+			return utils.TrackError(fmt.Errorf("failed to replace ServiceProviderCluster: %w", err))
+		}
+	}
+
+	// Write active versions onto the customer-facing cluster document so the
+	// frontend can serve them without an extra Cosmos read.
+	if err := c.writeActiveVersionsToCluster(ctx, key, newActiveVersions); err != nil {
```
machi1990: I think we can inline the logic here, it reads straight forward

backend/pkg/controllers/nodepool/version/nodepool_active_version_controller.go
```diff
@@ -135,22 +135,52 @@ func (c *nodePoolActiveVersionSyncer) SyncOnce(ctx context.Context, key controll
...
-		"newActiveVersions", newActiveVersions)
-	replacement := cachedServiceProviderNodePool.DeepCopy()
-	replacement.Status.NodePoolVersion.ActiveVersions = newActiveVersions
-	_, err = c.resourcesDBClient.ServiceProviderNodePools(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, key.HCPNodePoolName).Replace(ctx, replacement, nil)
+	return nil
+}
+
+func (c *nodePoolActiveVersionSyncer) writeActiveVersionsToNodePool(ctx context.Context, key controllerutils.HCPNodePoolKey, activeVersions []coreapi.ServiceProviderNodePoolActiveVersion) error {
+	nodePoolCRUD := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).NodePools(key.HCPClusterName)
+	freshNodePool, err := nodePoolCRUD.Get(ctx, key.HCPNodePoolName)
```
machi1990: Read from the Lister, no need to read directly from cosmos.

backend/pkg/controllers/nodepool/version/nodepool_active_version_controller.go
```diff
@@ -135,22 +135,52 @@ func (c *nodePoolActiveVersionSyncer) SyncOnce(ctx context.Context, key controll
...
-	logger.Info("Active versions changed",
-		"oldActiveVersions", cachedServiceProviderNodePool.Status.NodePoolVersion.ActiveVersions,
-		"newActiveVersions", newActiveVersions)
-	replacement := cachedServiceProviderNodePool.DeepCopy()
-	replacement.Status.NodePoolVersion.ActiveVersions = newActiveVersions
-	_, err = c.resourcesDBClient.ServiceProviderNodePools(key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, key.HCPNodePoolName).Replace(ctx, replacement, nil)
+	return nil
+}
+
+func (c *nodePoolActiveVersionSyncer) writeActiveVersionsToNodePool(ctx context.Context, key controllerutils.HCPNodePoolKey, activeVersions []coreapi.ServiceProviderNodePoolActiveVersion) error {
```
machi1990: np: the logic seems straight forward and can be inlined.

backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
ahitacat: Yes, not the good approach. I need to change this, to only show 1 equal version.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
deads2k: The experience we're catering for on the control plane is y-stream guarantee only.  We don't want to expand our visible API surface area broader, because it ends up making it our effective guarantee.

ahitacat:
/hold until new API has been created
We have decided the API version where this will be launched. Waiting for this PR to be merged
https://github.com/Azure/ARO-HCP/pull/6515

backend/pkg/controllers/nodepool/version/nodepool_active_version_controller.go
ahitacat: I added the Lister for the HCPNodePools, as it wasn't used by this controller.

machi1990 (changes requested):
The implementation mostly looks good.
Requesting e2e to be added once the sdk is available so that we can add assertion for it.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
ahitacat: @mbarnes added the `ExternalAuthResourceStatus` to typespec. Also split the typespce+generated filed for the API and the backend change in 2 different commits so it is easy to review.

backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
ahitacat: Done, is now using the existingCluster.

test/e2e/cluster_create_nodepool_active_versions.go
```diff
@@ -112,6 +112,31 @@ var _ = Describe("Customer", func() {
 			Expect(err).NotTo(HaveOccurred(), "failed to create node pool %q for v20261001preview cluster %q",
 				customerNodePoolName, customerClusterName)
 
+			By("verifying active versions are returned for the cluster")
+			clusterResp, err := tc.Get20261001ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient().Get(ctx, *resourceGroup.Name, customerClusterName, nil)
+			Expect(err).NotTo(HaveOccurred(), "failed to GET cluster %q", customerClusterName)
+			Expect(clusterResp.Properties).NotTo(BeNil(), "cluster %q response Properties was nil", customerClusterName)
```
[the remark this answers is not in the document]
ahitacat: Updated PR description.

ahitacat:
/hold cancel
The changes have been moved to the correct API. It can be finally review and merged.

internal/api/coreapi/types_cluster.go
```diff
@@ -69,6 +71,17 @@ type HCPOpenShiftClusterStatus struct {
...
+}
+
+type HCPClusterActiveVersion struct {
+	// Version is the user-facing version in x.y format (e.g., "4.20")
+	Version *semver.Version `json:"version,omitempty"`
```
machi1990: When this is serialised; doesn't a semver.Version{Major: 4, Minor: 20} give something like 
```
"version": "4.20.0"
```

notice the .0 as the patch version, is that intended? @ahitacat 

Can the e2e verify that we get x.y only?

machi1990: We probably might want to keep it as a string ; something like what we do for the version.id
```
type VersionProfile struct {
	ID           string `json:"id,omitempty"`
	ChannelGroup string `json:"channelGroup,omitempty"`
}
```

backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
```diff
@@ -211,3 +230,18 @@ func getHostedClusterDesiredVersionChannels(hostedCluster *hsv1beta1.HostedClust
 	}
 	return hostedCluster.Status.Version.Desired.Channels
 }
+
+func hcpClusterActiveVersionFromServiceProviderActiveVersions(activeVersions []coreapi.ServiceProviderClusterActiveVersion) []coreapi.HCPClusterActiveVersion {
+	var versions []coreapi.HCPClusterActiveVersion
+	var lastSeen string
+	for _, activeVersion := range activeVersions {
+		version := fmt.Sprintf("%d.%d", activeVersion.Version.Major, activeVersion.Version.Minor)
+		if lastSeen == version {
```
machi1990: Will this fall short if in some hypothetical future we allow customers to downgrade their major.minor?

machi1990: Why not, just get all unique major.minor ?

i.e instead of checking uniqueness assuming that we won't allow downgrading a major.minor.

You've a Set structs, that tracks what's been collected and the check here becomes `has this version already been collected; if yes, then bail out and go to the next one`

internal/api/coreapi/types_cluster.go
ahitacat: ok, that makes sense. I though that being a version could add some value, but honestly this is purely meant to show the version number. I don't think we will gain too much using a version compare with a string.

backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
ahitacat: Ok I get your point. I will revert this as what was proposed before. I think a proper discussion can happen when working on rollbacks.

test/e2e/cluster_create_v20261001preview.go
```diff
@@ -112,6 +112,31 @@ var _ = Describe("Customer", func() {
...
+				*resourceGroup.Name, customerClusterName, customerNodePoolName,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to GET node pool %q", customerNodePoolName)
+			Expect(nodePool.Properties).NotTo(BeNil(), "node pool %q response Properties was nil", customerNodePoolName)
+			Expect(nodePool.Properties.Status).NotTo(BeNil(), "node pool %q response Properties.Status was nil", customerNodePoolName)
+			Expect(nodePool.Properties.Status.ActiveVersions).NotTo(BeEmpty(), "node pool %q should have at least one active version", customerNodePoolName)
+			for _, v := range nodePool.Properties.Status.ActiveVersions {
+				Expect(v.Version).NotTo(BeNil(), "node pool %q active version entry should have a non-nil Version", customerNodePoolName)
+				Expect(*v.Version).NotTo(BeEmpty(), "node pool %q active version should not be empty", customerNodePoolName)
+			}
```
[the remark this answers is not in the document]
ahitacat: Creation of cluster/nodepool is done using the method `CreateNodePoolFromParam20261001`, `CreateHCPClusterFromParam20261001`, and passes the timeout. Meaning this method will wait until the resource is created or it reaches the timeout to complete.

The assertions run immediately, and if it is successful the resource should has been created. There is no need to wait more.

test/e2e/cluster_create_nodepool_active_versions.go
```diff
@@ -112,6 +112,31 @@ var _ = Describe("Customer", func() {
...
+			Expect(clusterResp.Properties.Status).NotTo(BeNil(), "cluster %q response Properties.Status was nil", customerClusterName)
+			Expect(clusterResp.Properties.Status.ActiveVersions).NotTo(BeEmpty(), "cluster %q should have at least one active version", customerClusterName)
+			for _, v := range clusterResp.Properties.Status.ActiveVersions {
+				Expect(v.Version).NotTo(BeNil(), "cluster %q active version entry should have a non-nil Version", customerClusterName)
+				Expect(*v.Version).NotTo(BeEmpty(), "cluster %q active version should not be empty", customerClusterName)
+			}
```
machi1990, line 128: let's further verify the exact value of the version

test/e2e/cluster_create_v20261001preview.go
machi1990: let's verify the exact value of the version

machi1990 (changes requested):
It's looking good, I am ready to tag once the latest requested changes have been addressed. But you'll also need someone to double check the api changes; @mbarnes maybe?

test/e2e/cluster_create_nodepool_active_versions.go
machi1990, line 1: can we rename the file so that it's concrete what it is testing? 

Similar to the test name in https://github.com/ahitacat/ARO-HCP/blob/f5e7684b1196e053ef4607530152247b1cf1e0fe/test/e2e/cluster_create_v20261001preview.go#L33 

So that the use case reads `Customer should be able to retrieve cluster and nodepool active versions`

ahitacat: done

test/sdk/v20260901preview/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp/models.go
```diff
@@ -1056,6 +1066,9 @@ type Resource struct {
 
 // ResourceStatus represents the observed status of the resource.
 type ResourceStatus struct {
+	// READ-ONLY; The activeVersions of a resource
+	ActiveVersions []*ActiveVersion
```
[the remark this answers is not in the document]
ahitacat: fixed

internal/api/coreapi/types_cluster.go
```diff
@@ -69,6 +69,17 @@ type HCPOpenShiftClusterStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	UserFacingConditions []metav1.Condition `json:"userFacingConditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// ActiveVersions are the active versions of the cluster control plane, ordered newest first.
+	// Written by: ControlPlaneActiveVersions
+	// +optional
+	// +listType=slice
+	ActiveVersions []HCPClusterActiveVersion `json:"activeVersions,omitempty"`
```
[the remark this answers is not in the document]
ahitacat: fixed, specified as set

internal/api/coreapi/types_nodepool.go
```diff
@@ -70,6 +70,17 @@ type HCPOpenShiftClusterNodePoolStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	UserFacingConditions []metav1.Condition `json:"userFacingConditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// ActiveVersions are the active versions of the node pool, ordered newest first.
+	// Written by: NodePoolActiveVersions
+	// +optional
+	// +listType=slice
```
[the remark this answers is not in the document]
ahitacat: fixed, specified as set

test/testdata/zz_fixture_TestMainListSuitesForEachSuite_stage_parallelstage_parallel.txt
```diff
@@ -11,12 +11,11 @@ Customer should create a cluster with private ingress using v20260630preview and
 Customer should create a cluster with private KAS and verify API server is only reachable from VNet while ingress remains public
 Create HCPOpenShiftCluster with Private KeyVault should create a cluster with private keyvault using v20251223preview API
 Customer should create a cluster using v20260901preview API and verify cluster health
-Customer should create a cluster using v20261001preview API and verify cluster health
 ARO HCP Service should successfully provision cluster when MI role assignments are added after cluster creation
 Customer should be able to create an HCP cluster then delete it by deleting the customer resource group
 Customer should be able to delete an HCP cluster whose managed identities were deleted first
 FIPS Mode Support with v2026-06-30 API should create an HCP cluster with FIPS mode enabled via cryptoRestrictions property
-Customer should be able to issue the default ingress serving certificate from Let's Encrypt using cert-manager with ACME DNS-01 challenges
+Customer should be able to retrieve cluster and nodepool status active versions
```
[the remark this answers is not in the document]
machi1990: It sure does have a look at https://github.com/ahitacat/ARO-HCP/blob/fe68151239dcb062e711b8d37b3e04faa9c58e44/test/e2e/cluster_nodepool_active_versions.go

machi1990 (approved):
/lgtm
/hold 

As I would like @deads2k or @miguelsorianod to have a look at the API side of things.

backend/pkg/controllers/nodepool/version/nodepool_active_version_controller.go
```diff
@@ -135,23 +138,45 @@ func (c *nodePoolActiveVersionSyncer) SyncOnce(ctx context.Context, key controll
...
 		return nil
 	}
 	if err != nil {
-		return utils.TrackError(fmt.Errorf("failed to replace ServiceProviderNodePool: %w", err))
+		return utils.TrackError(fmt.Errorf("failed to get NodePool: %w", err))
 	}
+	newHCPActiveVersions := hcpNodePoolActiveVersionFromServiceProviderActiveVersions(newActiveVersions)
+	if internalcontrollerutils.NeedsUpdate(cachedNodePool.Status.ActiveVersions, newHCPActiveVersions) {
+		nodePoolReplacement := cachedNodePool.DeepCopy()
+		nodePoolReplacement.Status.ActiveVersions = newHCPActiveVersions
```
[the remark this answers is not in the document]
ahitacat: follow up work

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -1097,12 +1118,60 @@ model ExternalAuthProperties {
...
 }
 
+/** ClusterResourceStatus represents the observed status of the cluster resource. */
+@added(Versions.v2026_10_01_preview)
+model ClusterResourceStatus {
+  /** The conditions on the resource */
+  ...ResourceStatus;
+
+  /** The activeVersions of a resource. This field is an array because during upgrades there can be more than one active version. */
```
deads2k, line 1134: This is often non-obvious, so we should explain this commonly happens during upgrades and that ordering in the array has no meaning.

api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/hcpCluster-models.tsp
```diff
@@ -1097,12 +1118,60 @@ model ExternalAuthProperties {
...
+  @visibility(Lifecycle.Read)
+  version: string;
+}
+
+/** NodePoolResourceStatus represents the observed status of the nodepool resource. */
+@added(Versions.v2026_10_01_preview)
+model NodePoolResourceStatus {
+  ...ResourceStatus;
+
+  /** The activeVersions of a resource. This field is an array because during upgrades there can be more than one active version. */
```
deads2k, line 1154: Indicate this often happens during upgrades and with nodepools some level of skew is extremely common.

deads2k (commented):
docs need improvement.  willing to have as a followup.

deads2k:
/approve

leaving lgtm with @machi1990

mbarnes:
Informal LGTM, mostly for the API specifications.  Leaving formal LGTM for the backend logic to @machi1990.

machi1990 (approved):
/hold cancel
/lgtm 

@ahitacat please followup with https://github.com/Azure/ARO-HCP/pull/6547#pullrequestreview-5178378535 in a separate PR

PR #6554, opened 2026-08-14
observability: Frontend ARM RP SLO recording rules (ARO-28705)

shubhadapaithankar:
## Summary
- Adds Frontend (ARM RP) user-journey SLO recording rules for [ARO-28705](https://redhat.atlassian.net/browse/ARO-28705).
- Covers ADR baseline SLIs:
  - Availability: good/total ratio (non-5xx / total)
  - Errors: 5xx / total
  - Latency: `histogram_quantile` p99 / p95 **in seconds** (`sli:frontend_http:latency_p99:rate5m`, `_p95:`) — not a ≤1s bucket share
  - Traffic: request rate (RPS)
  - Ready: kube Deployment available/spec (`sli:frontend:ready:ratio5m`; probes hit `/healthz`)
  - Saturation: Frontend pod CPU vs request, memory vs limit (kube-state-metrics requests/limits deduped with `max by (cluster, namespace, pod, container)`)
- Excludes ARM async poll routes `hcpoperationresults` and `hcpoperationstatuses` (`route!~".*hcpoperation(results|statuses).*"`). Dedups HA scrapes with `max without (prometheus_replica)`.
- Zero ARM traffic: HTTP availability/errors/latency/traffic emit **no sample** (avoid NaN). Idle-cluster promtool case covers that. Distinguish a Frontend outage with the ready ratio (0 vs absent).
- Regenerates `generatedRecordingRules.bicep`. Istio/Cosmos saturation deferred until owned series are confirmed.
- Alerts ([ARO-28707](https://redhat.atlassian.net/browse/ARO-28707) / #6557) and dashboards ([ARO-28706](https://redhat.atlassian.net/browse/ARO-28706) / #6555) are out of scope for this PR.

## Test plan
- [ ] `prometheus-rules` + `promtool test rules` for `frontend-slo-recordingRule_test.yaml` (including idle-but-healthy cluster)
- [x] `az bicep format` on `generatedRecordingRules.bicep` (CI `observability-prometheus` / `bicep-lint` green)
- [ ] Confirm recorded series appear after deploy on INT svc Prometheus
- [ ] Spot-check against Grafana Frontend SLO dashboard (#6555)

## Related
- Epic: [ARO-28703](https://redhat.atlassian.net/browse/ARO-28703) · Story: [ARO-28705](https://redhat.atlassian.net/browse/ARO-28705)
- Companion docs: Frontend UJ runbook / TSG (Azure-Documents-Common)
- Merge order: **#6554 → #6555 → #6557**

### DIFF
--- a/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
@@ -100,7 +100,7 @@ resource arohcpFrontendSloRecordingRules 'Microsoft.AlertsManagement/prometheusR
     rules: [
       {
         record: 'sli:frontend_http:availability:rate5m'
-        expression: '(sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{code!~"5..",route!~".*hcpoperation(results|statuses).*"}[5m]))) / sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m])))) and on (cluster) (sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m]))) > 0)'
+        expression: '((sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{code!~"5..",route!~".*hcpoperation(results|statuses).*"}[5m]))) or 0 * sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m])))) / sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m])))) and on (cluster) (sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m]))) > 0)'
       }
       {
         record: 'sli:frontend_http:availability:rate_avg_30d'
--- a/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
@@ -103,8 +103,8 @@ resource arohcpFrontendSloRecordingRules 'Microsoft.AlertsManagement/prometheusR
         expression: '((sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{code!~"5..",route!~".*hcpoperation(results|statuses).*"}[5m]))) or 0 * sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m])))) / sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m])))) and on (cluster) (sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m]))) > 0)'
       }
       {
-        record: 'sli:frontend_http:availability:rate_avg_30d'
-        expression: 'avg_over_time(sli:frontend_http:availability:rate5m[30d:5m])'
+        record: 'sli:frontend_http:good:rate5m'
+        expression: '(sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{code!~"5..",route!~".*hcpoperation(results|statuses).*"}[5m]))) or 0 * sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m]))))'
       }
       {
         record: 'errors:frontend_http:error_rate:rate5m'
@@ -121,6 +121,10 @@ resource arohcpFrontendSloRecordingRules 'Microsoft.AlertsManagement/prometheusR
       {
         record: 'traffic:frontend_http:request_rate:rate5m'
         expression: 'sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m])))'
+      }
+      {
+        record: 'sli:frontend_http:availability:rate_avg_30d'
+        expression: '(avg_over_time(sli:frontend_http:good:rate5m[30d:5m]) / avg_over_time(traffic:frontend_http:request_rate:rate5m[30d:5m])) and avg_over_time(traffic:frontend_http:request_rate:rate5m[30d:5m]) > 0'
       }
       {
         record: 'sli:frontend:ready:ratio5m'
--- a/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
@@ -128,11 +128,11 @@ resource arohcpFrontendSloRecordingRules 'Microsoft.AlertsManagement/prometheusR
       }
       {
         record: 'sli:frontend:saturation_cpu:ratio5m'
-        expression: '(sum by (cluster) (max without (prometheus_replica) (rate(container_cpu_usage_seconds_total{container="aro-hcp-frontend",namespace="aro-hcp"}[5m]))) / sum by (cluster) (max without (prometheus_replica) (kube_pod_container_resource_requests{container="aro-hcp-frontend",namespace="aro-hcp",resource="cpu"}))) and on (cluster) (sum by (cluster) (max without (prometheus_replica) (kube_pod_container_resource_requests{container="aro-hcp-frontend",namespace="aro-hcp",resource="cpu"})) > 0)'
+        expression: '(sum by (cluster) (max without (prometheus_replica) (rate(container_cpu_usage_seconds_total{container="aro-hcp-frontend",namespace="aro-hcp"}[5m]))) / sum by (cluster) (max by (cluster, namespace, pod, container) (kube_pod_container_resource_requests{container="aro-hcp-frontend",job="kube-state-metrics",namespace="aro-hcp",resource="cpu"}))) and on (cluster) (sum by (cluster) (max by (cluster, namespace, pod, container) (kube_pod_container_resource_requests{container="aro-hcp-frontend",job="kube-state-metrics",namespace="aro-hcp",resource="cpu"})) > 0)'
       }
       {
         record: 'sli:frontend:saturation_memory:ratio5m'
-        expression: '(sum by (cluster) (max without (prometheus_replica) (container_memory_working_set_bytes{container="aro-hcp-frontend",namespace="aro-hcp"})) / sum by (cluster) (max without (prometheus_replica) (kube_pod_container_resource_limits{container="aro-hcp-frontend",namespace="aro-hcp",resource="memory"}))) and on (cluster) (sum by (cluster) (max without (prometheus_replica) (kube_pod_container_resource_limits{container="aro-hcp-frontend",namespace="aro-hcp",resource="memory"})) > 0)'
+        expression: '(sum by (cluster) (max without (prometheus_replica) (container_memory_working_set_bytes{container="aro-hcp-frontend",namespace="aro-hcp"})) / sum by (cluster) (max by (cluster, namespace, pod, container) (kube_pod_container_resource_limits{container="aro-hcp-frontend",job="kube-state-metrics",namespace="aro-hcp",resource="memory"}))) and on (cluster) (sum by (cluster) (max by (cluster, namespace, pod, container) (kube_pod_container_resource_limits{container="aro-hcp-frontend",job="kube-state-metrics",namespace="aro-hcp",resource="memory"})) > 0)'
       }
     ]
   }
--- a/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
+++ b/dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
@@ -1,3 +1,4 @@
+
 param azureMonitoring string
 
 param location string = resourceGroup().location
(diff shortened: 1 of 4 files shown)


shubhadapaithankar:
Addressed Copilot review on tip:
- Latency bucket match: `le=~"^1(\\.0)?$"` (dashboard pattern); test fixture uses `le="1"`
- Zero-traffic guards: `and on (cluster) (... > 0)` on availability / errors / latency / saturation ratios
- Regenerated `generatedRecordingRules.bicep` via the recording-rules generator (`make -C observability recording-rules` path). This PR is SLIs only — `make alerts` / burn-rate alerts are ARO-28707.

`ci/prow/images` failure looked like an infra registry flake (network unreachable on exporter image); retriggered with `/test images`.

observability/recording-rules/frontend-slo-recordingRule.yaml
```diff
@@ -0,0 +1,228 @@
...
+    rules:
+    # Availability SLI: non-5xx / total (ratio in [0,1]).
+    - record: sli:frontend_http:availability:rate5m
+      expr: |
+        (
+          sum by (cluster) (
+            max without (prometheus_replica) (
+              rate(frontend_http_requests_total{
+                code!~"5..",
+                route!="/subscriptions/{subscriptionid}/providers/microsoft.redhatopenshift/locations/{location}/hcpoperationresults/{operationid}"
```
swiencki: This filter excludes the hcpoperationresults poll route but not its sibling hcpoperationstatuses route (a GET registered at frontend/pkg/frontend/routes.go:448 and lowercased by MuxPattern), which ARM polls at least as often, so fast successful status polls dilute these SLIs and skew the thresholds the RP-lane alerts rely on.

observability/recording-rules/frontend-slo-recordingRule.yaml
```diff
@@ -0,0 +1,228 @@
...
+        )
+    # Latency SLI: requests completing within 1s / total.
+    # Atomic single expression; le matches "1" or "1.0".
+    - record: sli:frontend_http:latency:rate5m
+      expr: |
+        (
+          sum by (cluster) (
+            max without (prometheus_replica) (
+              rate(frontend_http_requests_duration_seconds_bucket{
+                le=~"^1(\\.0)?$",
```
swiencki: Minor: the leading ^ and trailing $ in this le=~ matcher are redundant because PromQL already fully anchors =~ matches.

observability/recording-rules/frontend-slo-recordingRule.yaml
shubhadapaithankar: Fixed in `e53ab5cd3`: SLIs now use `route!~.*hcpoperation(results|statuses).*` so both ARM async poll routes are excluded (status polls were diluting the denominator). Regenerated `generatedRecordingRules.bicep` via prometheus-rules.

observability/recording-rules/frontend-slo-recordingRule.yaml
shubhadapaithankar: Fixed in `e53ab5cd3`: simplified to `le=~1(\\.0)?` (PromQL `=~` is already fully anchored).

observability/recording-rules/frontend-slo-recordingRule.yaml
```diff
@@ -0,0 +1,234 @@
...
+    interval: 1m
+    rules:
+    # Availability SLI: non-5xx / total (ratio in [0,1]).
+    - record: sli:frontend_http:availability:rate5m
+      expr: |
+        (
+          sum by (cluster) (
+            max without (prometheus_replica) (
+              rate(frontend_http_requests_total{
+                code!~"5..",
```
swiencki: Make the availability numerator zero-fill against the total request series, because `code!~"5.."` disappears during all-5xx traffic and the SLI will stop emitting instead of recording 0.

shubhadapaithankar: Fixed in `1b281d7ad`: availability numerator now zero-fills with `or 0 * total` (same pattern as error_rate), so all-5xx traffic records availability=0 instead of dropping the series. Added a promtool case for that.

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
```diff
@@ -87,3 +87,45 @@ resource arohcpUserJourneyClusterUpgradeRecordingRules 'Microsoft.AlertsManageme
...
+  properties: {
+    scopes: [
+      azureMonitoring
+    ]
+    enabled: true
+    interval: 'PT1M'
+    rules: [
+      {
+        record: 'sli:frontend_http:availability:rate5m'
+        expression: '((sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{code!~"5..",route!~".*hcpoperation(results|statuses).*"}[5m]))) or 0 * sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m])))) / sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m])))) and on (cluster) (sum by (cluster) (max without (prometheus_replica) (rate(frontend_http_requests_total{route!~".*hcpoperation(results|statuses).*"}[5m]))) > 0)'
```
SudoBrendan, line 103: As written here, complete outages of FE (urgent) look identical to traffic droughts (normal) - they both measure as absent (no sample).

Ideally we work towards a probe of some kind, alerting on /healthz or /readyz functioning. There's nothing else we can use? If not - we'll want to hack around with `up`/`absent` or something in the alert definition.

observability/recording-rules/frontend-slo-recordingRule.yaml
```diff
@@ -0,0 +1,245 @@
...
+          sum by (cluster) (
+            max without (prometheus_replica) (
+              rate(frontend_http_requests_total{
+                route!~".*hcpoperation(results|statuses).*"
+              }[5m])
+            )
+          )
+          > 0
+        )
+    - record: sli:frontend_http:latency:rate5m
```
SudoBrendan: This threshold-ratio SLI (requests within 1s / total) treats a 1.01s response and a 50s response as equivalently bad, since both simply exceed the 1s threshold. The only signal is what fraction of requests missed the mark, not how badly.

Per the ADR, we should use `histogram_quantile` instead for latency. A p99/p95 recording rule would give us:

- The same burn-rate signal you already have: `p99 > 1s` is equivalent to "more than 1% of requests exceeded 1s"
- Severity escalation: we can distinguish `p99 = 1.2s` (barely missing) from `p99 = 15s` (something is very wrong) and alert at different thresholds (e.g. 1s, 5s)
- ADR alignment: the alerting ADR naming convention uses `LatencyP95`/`LatencyP99`, which maps directly to `histogram_quantile(0.95, ...)` and `histogram_quantile(0.99, ...)`

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
shubhadapaithankar: Addressed in 513e12cf1.

/healthz exists, but it is the kube probe path -- MISE/POP means we cannot blackbox it through ARM. HTTP SLIs still go absent with zero ARM traffic (outage looks like a drought).

Added sli:frontend:ready:ratio5m (available/spec on aro-hcp-frontend). That emits 0 on a full replica outage instead of disappearing. Alerts in #6557: userJourneyFrontendReady pages on ready==0; traffic-drop only fires if pods are still Ready.

observability/recording-rules/frontend-slo-recordingRule.yaml
shubhadapaithankar: Addressed in 513e12cf1 (alerts/dashboard follow in #6557 / #6555).

Replaced the 1s-bucket share with histogram_quantile recording rules:
- sli:frontend_http:latency_p99:rate5m
- sli:frontend_http:latency_p95:rate5m

Alerts: fast pages on p99 > 5s (severe), medium on p99 > 1s (SLO miss), so 1.2s vs 15s are distinct. Same poll-route exclude as before.

observability/recording-rules/frontend-slo-recordingRule_test.yaml
```diff
@@ -0,0 +1,159 @@
...
+  - series: 'kube_deployment_status_replicas_available{cluster="svc-down", namespace="aro-hcp", deployment="aro-hcp-frontend", prometheus_replica="a"}'
+    values: "0+0x15"
+  - series: 'kube_deployment_spec_replicas{cluster="svc-down", namespace="aro-hcp", deployment="aro-hcp-frontend", prometheus_replica="a"}'
+    values: "2+0x15"
+  promql_expr_test:
+  - expr: sli:frontend:ready:ratio5m{cluster="svc-down"}
+    eval_time: 5m
+    exp_samples:
+    - labels: '{__name__="sli:frontend:ready:ratio5m", cluster="svc-down"}'
+      value: 0
```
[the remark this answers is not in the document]
shubhadapaithankar: Addressed in `970f69dda`. Added an idle-but-healthy promtool case (`svc-idle`): kube ready=1, no ARM HTTP series. Availability / errors / latency p99+p95 / traffic emit no sample; `sli:frontend:ready:ratio5m` stays 1. Also ran prometheus-rules generate + `az bicep format` on `generatedRecordingRules.bicep` (drops the generator leading blank) so CI `make alerts` / bicep-lint stay clean.

observability/recording-rules/frontend-slo-recordingRule.yaml
```diff
@@ -0,0 +1,304 @@
...
+    prometheus: k8s
+    role: recording-rules
+  name: frontend-slo-recording-rules
+  namespace: monitoring
+spec:
+  groups:
+  # ========================================
+  # Frontend (ARM RP) User Journey SLO Recording Rules
+  # ========================================
+  # Baseline SLIs per ARO HCP Alerting Recommendation ADR (good / total ratios)
```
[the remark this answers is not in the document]
shubhadapaithankar: Addressed in `970f69dda`. Header now states availability/errors are ratios and latency is `histogram_quantile` p99/p95 seconds (not a <=1s share). PR description updated to match the quantile-based SLI (no `le=~` 1s-bucket rule).

dev-infrastructure/modules/metrics/rules/generatedRecordingRules.bicep
shubhadapaithankar: Please re-review when you have a chance.

Addressed in `513e12cf1` (ready SLI + `userJourneyFrontendReady` on #6557) with follow-ups `970f69dda` (idle-cluster promtool case: HTTP SLIs absent, ready=1) and `dba7d1724` (kube-state-metrics saturation dedup).

`/healthz` is kube-probe only (MISE/POP). HTTP SLIs stay absent on zero ARM traffic; `sli:frontend:ready:ratio5m == 0` is the outage signal vs a traffic drought.

observability/recording-rules/frontend-slo-recordingRule.yaml
shubhadapaithankar: Please re-review when you have a chance.

Addressed in `513e12cf1`. Latency is now `histogram_quantile` p99/p95 seconds (`sli:frontend_http:latency_p99:rate5m` / `_p95:`), not a 1s-bucket share. Fast alert pages on p99 > 5s; medium on p99 > 1s so 1.2s vs 15s are distinct. Dashboard (#6555) and alerts (#6557) follow the same recording rules.

observability/recording-rules/frontend-slo-recordingRule.yaml
```diff
@@ -0,0 +1,314 @@
...
+          sum by (cluster) (
+            max without (prometheus_replica) (
+              rate(frontend_http_requests_total{
+                route!~".*hcpoperation(results|statuses).*"
+              }[5m])
+            )
+          )
+          > 0
+        )
+    - record: sli:frontend_http:availability:rate_avg_30d
```
SudoBrendan: Add some tests for this please.

observability/recording-rules/frontend-slo-recordingRule.yaml
```diff
@@ -0,0 +1,314 @@
...
+      # 30-day smoothed availability for SLO status panels (not for fast burn alerts).
+
+      expr: avg_over_time(sli:frontend_http:availability:rate5m[30d:5m])
```
SudoBrendan: I'm not sure what the implications are _actually_ to this or whether it should hold or not... but I at least want to point out that this math is funky because `availability` is calculated based on request traffic (which is variable - as opposed to if this was a probe going off at a standard rate e.g. every 15s). Put another way, if we are taking an `avg` of `rate([5m])`s - that doesn't account for traffic volume in the intervals.

This means this recording rule can misrepresent "availability" if we get scenarios like:

Minutes 1-5: 10,000 requests, 100% "good"
Minutes 6-10: 1 request, 0% "good"

if we avg this - we get "50% available", with 1 request failing out of 10,001, which we probably would rather have represented as 99.99% available if all requests are equal.

The opposite can also be true - 

Minutes 1-5: 10,000 requests, 0% "good"
Minutes 6-10: 1 request, 100% "good"

Also results in "50% available", when it's actually 0.01% available.

observability/recording-rules/frontend-slo-recordingRule.yaml
shubhadapaithankar: Added. `frontend-slo-recordingRule_test.yaml` now covers `good:rate5m` / `rate_avg_30d` on the constant-ratio, all-5xx, and idle cases, plus a volume-skew fixture matching your 10k-good then 1-fail example: `avg_over_time` of the 5m ratio is 50%, the recording rule is 10000/10001 (`4e67bc77d`).

observability/recording-rules/frontend-slo-recordingRule.yaml
shubhadapaithankar: Agreed that extra `rate1h`/`rate6h` *ratio* windows would not fix this — still avg-of-ratios, just maybe less bursty if 6h traffic is smoother.

The out (same idea as KAS `request_total` / `request_5xx` volume series) is to record good RPS and total RPS, then 30d availability = `avg_over_time(good) / avg_over_time(total)` over equal 5m steps, which equals sum(good requests)/sum(total requests). That is now `sli:frontend_http:availability:rate_avg_30d`. The 5m availability *ratio* stays for fast burn — it is already volume-weighted inside each 5m window.

The test in this commit is your 10k-vs-1 case. MWMBR alerts in #6557 still avg the 5m ratio over 1h/6h/3d; we can switch those to the same good/total split as a follow-up if you want that to hold there too. (`4e67bc77d`)

SudoBrendan (approved):
/approve /lgtm

PR #6556, opened 2026-08-14
revert: enable customer-managed key for nodePool OS disk encryption

mgahagan73:
## Summary

Reverts #5682 ("feat: enable customer-managed key for nodePool OS disk encryption").

This is a straight `git revert -m 1` of the merge commit. It builds clean and all affected unit tests pass; no other code references the removed symbols.

## Why

The new `nodepool_osdisk_encryption.go` e2e test **leaks real Azure resources on every single run, regardless of whether the test passes or fails.** Checked across 8 runs spanning presubmit, 3 passing INT runs, 2 Stage runs, and 2 Prod runs — every one shows the identical cleanup failure:

* An orphaned Azure role assignment on the customer Key Vault scope (`Key Vault Crypto Service Encryption User`) — cleanup gets `403 AuthorizationFailed` because the test identity lacks `Microsoft.Authorization/roleAssignments/delete` on that scope.
* An orphaned Azure role assignment on the DiskEncryptionSet scope (`Reader`) — same `403 AuthorizationFailed` on delete.
* A soft-deleted, unpurgeable Key Vault (`cust-kv-*`) — cleanup gets `400 MethodNotAllowed` ("Operation 'DeletedVaultPurge' is not allowed") because the vault is created with purge protection enabled, and it remains in the subscription for its full soft-delete retention window.

This isn't correlated with the test outcome — the leak happens whether the test passes (confirmed on the PR's own presubmit run and on 3 passing INT runs) or fails (confirmed on Stage and Prod). It happens because the test's cleanup path logs these errors instead of failing on them (tracked separately in [ARO-29067](https://redhat.atlassian.net/browse/ARO-29067)) — but until that's fixed, this test guarantees an orphaned role assignment plus an unpurgeable Key Vault on every single invocation, across every environment this test runs in (INT presubmit, INT/Stage/Prod periodic and rollout, ocp-nightly).

Separately, on the two real Prod runs observed, the underlying customer-managed-key encryption itself doesn't appear to take effect on the actual VM OS disks (`VerifyVMOSDiskCustomerEncryption` found disks still on `EncryptionAtRestWithPlatformKey`), even though it passes in Dev/CI and INT. That's a secondary concern and not the primary reason for this revert, but it means the feature isn't yet proven end-to-end in the environments most exposed to the leak.

Recommend re-landing once:
1. The cleanup path fails the test on these errors instead of swallowing them ([ARO-29067](https://redhat.atlassian.net/browse/ARO-29067)), and
2. The test identity/Key Vault purge-protection setup is fixed so cleanup actually succeeds.

## Test plan

- [x] `go build ./...` in `internal` and `test` modules
- [x] `go test ./ocm/...` (internal) — pass
- [x] `go test ./util/...` (test) — pass
- [x] `go test -run TestMainListSuitesForEachSuite ./...` (test/cmd/aro-hcp-tests) — pass, fixtures consistent
- [x] Verified no remaining references to removed symbols (`CreateDiskEncryptionSet`, `VerifyVMOSDiskCustomerEncryption`, etc.)


cadenmarchese:
It looks like the 2 role assignments were unable to be deleted by the cleaner because the resource group (scope) had been deleted in a previous step. The fix would be to clean the role assignments before cleaning the resource group.

1. "no left behind managed resource groups found\" \"resourceGroup\"=\"des-encrypt-5btwpbgw2vw7\"\n\"ts\"=\"2026-08-14 18:53:41.356623\" \"level\"=0 \"msg\"=\**"deleting resource group\" \"resourceGroup\"=\"des-encrypt-5btwpbgw2vw7**\"\n\"ts\"=\"2026-08-14 18:55:08.256425\" \"level\"=0 \"msg\"

1. does not have authorization to perform action 'Microsoft.Authorization/roleAssignments/delete' over scope '/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/**des-encrypt-5btwpbgw2vw7**/providers/Microsoft.KeyVault/vaults/cust-kv-s7ionugrizfvo/providers/Microsoft.Authorization/roleAssignments/24acdcf7-77c0-5eb5-8141-3560c8dc20fb' or the scope is invalid

As for the soft deleted keyvault, it's expected that it can't be purged, but the issue here is that the cleanup is attempting to purge it anyway. The fix would be to exclude this specific keyvault from the cleanup.

I can either handle this in a followup PR on Monday, or if the orphaned role assignment leak is an issue if we leave it over the weekend, feel free to proceed here.

cadenmarchese:
I've opened the followup PR to fix this issue: https://github.com/Azure/ARO-HCP/pull/6570

Prefer we close this PR in favor of 6570 since it avoids us needing to revert user impacting features.

mvacula02 (commented):
can we restrict the test to dev and integration instead of a complete revert?

janboll:
## Force Merge Justification

**What:** Reverts PR #5682 — removes the nodepool OS disk encryption E2E test, its helpers, and the `EncryptionSetID` conversion/wiring across 4 API versions.

**Why:** The test is blocking the Stage rollout. Stage E2E fails because the test identity lacks DES permissions. Additionally, the test leaks Azure resources (orphaned role assignments, unpurgeable Key Vaults) on every run in every environment, and the feature itself doesn't work in Prod (disks still show platform-managed encryption).

**Risk:** Low. Clean `git revert -m 1`, builds clean, all unit tests pass, no remaining references to removed symbols. No customer-facing functionality is lost — the feature wasn't working in Stage/Prod. Can be re-landed once cleanup and permission issues are fixed.

mgahagan73:
@janboll  How do you feel about closing this one and then merging https://github.com/Azure/ARO-HCP/pull/6570 after adding labels to the e2e test so it only runs in dev and stage?

janboll:
@mgahagan73 We want to release to prod soon to mitigate p0 defects. hence we need to move this fast. We can not risk waiting for potential fixes. Also I do not see changes to the MI which would grant the missing read in #6570. Hence, the revert. This is our usual pracitce, something breaks a release, we revert it and proceed. Otherwise we can not keep our cadence

PR #6578, opened 2026-08-18
docs: refresh hosted cluster backup documentation

tony-schndr:
### What

Refreshes `docs/backups.md` so it matches the current backup implementation:

- Renames the pause/state model from `BackupState`/`Paused` to `BackupScheduleState`/`Disabled` (`Enabled` or `Disabled`), matching the field on the `ServiceProviderCluster` document (`spc.Spec.BackupScheduleState`).
- Corrects config keys: cadence is selected via `backend.backupCadenceProfile` (was `backend.backupCadence`) and global pause via `backend.backupScheduleState` (was `backend.backupPaused`).
- Rewrites the **Cluster Deletion** section: teardown is driven by the backup schedule controller's Apply→Delete→Wait→Purge lifecycle, and the generic cluster child resources cleanup controller explicitly skips `backupschedule-`-prefixed desires.
- Clarifies pause semantics: the Velero Schedule pause flag is `globalState == Disabled || clusterState == Disabled`.
- Updates the Admin API PATCH state values (`Enabled`/`Disabled`) and the incident-response walkthrough accordingly.
- Removes `pod` from the list of backed-up standard workload resources.

### Why

The documentation had drifted from the implementation after the backup controller work landed. The field names, config keys, and the deletion/teardown flow described in the doc no longer reflected the code, which is misleading for SREs following it during incident response.

### Testing

Docs-only change — no code paths affected, so no unit/integration/E2E tests apply.

### DIFF
--- a/docs/backups.md
+++ b/docs/backups.md
@@ -55,7 +55,7 @@ flowchart LR
 2. kube-applier reads the ApplyDesires and applies the Velero Schedule resources to the appropriate management cluster in the `velero` namespace.
 3. Velero executes backups according to each schedule's cadence, uploading backup data to Azure Blob Storage.
 4. kube-applier reads the Velero Schedule status and writes it back into ReadDesire status in Cosmos DB.
-5. The admin API reads ReadDesire statuses to serve per-schedule backup time and phase. The ServiceProviderCluster record stores only the backup schedule enabled/paused state.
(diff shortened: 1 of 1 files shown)


avollmer-redhat (approved):
Verified all renamed fields, config keys, and deletion lifecycle claims against the implementation — everything checks out. Good cleanup, thanks Tony.

PR #6600, opened 2026-08-19
fix: filter E2E GPU SKU selection against RP allowlist

geoberle:
## Summary
- Add RP allowlist filtering to VM SKU selection so non-allowlisted GPU SKUs (e.g. A10) are excluded from both preferred and fallback paths
- Export `EnabledNodePoolAzureVMSizes` for cross-package use by the E2E test framework
- Add `IgnoreRPAllowlist` opt-out for selectors (Jumpbox) that intentionally use non-allowlisted sizes

Fixes the westus3 failure where all T4 preferred entries were restricted and fallback selected Standard_NC16ads_A10_v4, causing 400 InvalidRequestContent after ~16 min of cluster creation.

[ARO-29173](https://issues.redhat.com/browse/ARO-29173)

## Test plan
- [ ] `go test ./test/util/framework/ -count=1` passes
- [ ] New table test cases cover: allowlist filtering on fallback, IgnoreRPAllowlist opt-out, all-filtered-yields-error

### DIFF
--- a/internal/validation/enabled_nodepool_vm_sizes.go
+++ b/internal/validation/enabled_nodepool_vm_sizes.go
@@ -16,14 +16,14 @@ package validation
 
 import "k8s.io/apimachinery/pkg/util/sets"
 
-// enabledNodePoolAzureVMSizes returns the set of VM sizes that are enabled in the ARO-HCP service for node pools.
+// EnabledNodePoolAzureVMSizes returns the set of VM sizes that are enabled in the ARO-HCP service for node pools.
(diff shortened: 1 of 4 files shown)


machi1990 (approved):
/lgtm

mbukatov (commented):
/lgtm

This will compromise the nature of E2E test here (since we use internal knowledge of the service in E2E test), esp. for periodic runs (since there, we don't make sure we run the same version of the E2E tests matching the service).

That said, here is a conflict with what this gpu test case is testing vs what we want to gate during a prod rollout, which is beyond this PR. Let's fix the rollout gating issue.

PR #6615, opened 2026-08-19
test: categorize CI alert firings by blast radius instead of a skip list

cssjr:
## Jira
Fixes: [ARO-28187](https://redhat.atlassian.net/browse/ARO-28187)

## Why
A batch of generic Kubernetes/Prometheus alerts (`KubePodNotReady`, `KubeDeploymentReplicasMismatch`, `KubeNodeNotReady`, etc.) fire during dev e2e provisioning and get converted to JUnit failures by `gather-observability`, blocking the merge queue. As of the latest ticket update these have worsened significantly: 530+ DEV runs affected since 2026-07-07, ~70% DEV failure rate, `KubePodNotReady [hcp]` alone spiking to 30.8% of runs in a week. The [AROSLSRE-1395](https://redhat.atlassian.net/browse/AROSLSRE-1395) fix did not resolve them.

An earlier PR ([#6581](https://github.com/Azure/ARO-HCP/pull/6581)) proposed adding these alerts to the known-issues skip list. David Eads suggested that:

> We made alerting fail in CI clusters to improve our service and our alerting, not so we could build a skip list. Instead of this, how about digging into these failures and starting to categorize them like [...] and then we discuss reasonable alerting thresholds for each category.

## What
Replaces the binary known-issues skip list with a blast-radius **category** model:

- Every alert firing is classified into a named category with David's tier (1-6), a CI policy (`fail` / `fail-over-threshold` / `warn` / `ignore`), and a reason. Config lives in `test/cmd/aro-hcp-tests/gather-observability/alert-categories/categories.yaml` (replaces `known-issues/knownIssues.yaml`); matching/classification in `categories.go` (replaces `known_issues.go`).
- The old known-issues entries are migrated verbatim as tier-0 `ignore` categories (their original reasons preserved) -- they're genuinely understood transient noise, not the [ARO-28187](https://redhat.atlassian.net/browse/ARO-28187) alerts.
- New categories map the [ARO-28187](https://redhat.atlassian.net/browse/ARO-28187) alerts to David's tiers using real ARO-HCP namespaces: frontend/backend/cluster-service/maestro/fleet on the svc cluster, kube-applier/hypershift/mgmt-agent on the mgmt cluster, and the Cluster Service `ocm-<env>-<clusterID>[-<domain>]` convention to split a hosted cluster's controlplane namespace (tier 4, customer-visible) from its hostedcluster namespace (tier 5, non-customer-visible).
- `junit.go`'s policy engine evaluates each category independently per alert rule -- a single rule (e.g. `KubePodNotReady [hcp]`) can span multiple categories in one run, and the test case fails if any of them calls for it, while non-failing categories are still reported for visibility instead of disappearing into a skip list.
- HTML dashboard and JSON summary get a category badge/filter and per-category counts.

## Open questions for Architecture Office Hours
See [`docs/ci/ARO-28187-alert-categorization.md`](docs/ci/ARO-28187-alert-categorization.md) for the full alert-to-tier mapping, default policy table, and explicit open questions (frontend/backend namespace disambiguation, node-alert tiering, tier-3 threshold values, tier-5 policy choice, catch-all coverage for the `infra` workspace). **Policies and thresholds in this PR are provisional pending that discussion** -- this is a draft for review ahead of Office Hours, not a request to merge as-is.

## Testing
- `go test ./cmd/aro-hcp-tests/gather-observability/...` -- includes a new `TestDefaultCategoriesConfig` that validates the actual production `categories.yaml` (not just ad hoc test fixtures) correctly discriminates all 6 tiers plus the tier-4/5 namespace-segment-count ordering, and a new HTML-template smoke test.
- `go vet` / `gofmt` / `golangci-lint` clean.
- No `config/ make materialize` needed -- `categories.yaml` is `//go:embed`-ed into the test binary, not part of the helm/config render system.


test/cmd/aro-hcp-tests/gather-observability/alert-categories/categories.yaml
```diff
@@ -0,0 +1,289 @@
...
+# knownIssues.yaml, removed by this change) and are NOT related to the
+# ARO-28187 cluster-degradation alerts. They are listed first because
+# several are label-scoped exceptions to the broader namespace-based rules
+# further down (e.g. specific pods inside an ocm-* namespace) and must be
+# claimed before those broader rules run.
+categories:
+
+- name: "expected-noise-backend-nodepool-upgrade-retry"
+  tier: 0
+  policy: ignore
```
cssjr, line 57: Should known issues that are being actively worked on should "warn" or "ignore"?

test/cmd/aro-hcp-tests/gather-observability/alert-categories/categories.yaml
```diff
@@ -0,0 +1,289 @@
...
+  reason: "being worked on in https://issues.redhat.com/browse/ARO-28029"
+  match:
+  - name: "BackendControllerRetryHotLoop"
+    labels:
+      name: "triggernodepoolupgrade"
+
+- name: "expected-noise-backend-cluster-create-retry"
+  tier: 0
```
cssjr, line 64: Should policy be per tier or per issue?  Currently, it looks like each individual issue can have its own policy, rather than one policy for the whole tier.

test/cmd/aro-hcp-tests/gather-observability/testdata/zz_fixture_TestAlertsToJUnit_main.xml
```diff
@@ -3,13 +3,13 @@
     <properties></properties>
     <testcase name="[aro-hcp-observability] [hcp] alert HCPClusterHealth does not fire" time="0"></testcase>
     <testcase name="[aro-hcp-observability] [svc] alert BackendControllerRetryHotLoop does not fire" time="2603">
-      <skipped message="known issue: Nodepool create controller retry hot loops are observed during e2e runs. Needs investigation."></skipped>
+      <skipped message="non-blocking: expected-noise-nodepool-create-retry (tier 0, ignore): Nodepool create controller retry hot loops are observed during e2e runs. Needs investigation."></skipped>
```
cssjr, line 6: Do we want to keep "known issue" or change to "non-blocking"?

cssjr (commented):
Have a few notes/questions.

test/cmd/aro-hcp-tests/gather-observability/alert-categories/categories.yaml
```diff
@@ -0,0 +1,272 @@
...
+  - name: "OrphanedMRGDetected"
+- name: "expected-noise-backend-async-operation-stuck"
+  tier: 0
+  policy: ignore
+  reason: "Newly added stuck-async-operation alerts (https://github.com/Azure/ARO-HCP/pull/5952) surface a pre-existing problem: Azure Monitor Workspace ingestion latency (metrics can take ~10min to ingest) makes async operations look stuck and latches onto E2E runs even when the tests pass. Observed blocking green E2E runs (per https://cihealth.tools.hcpsvc.osadev.cloud/run-log?q=BackendAsync): BackendAsyncOperationClusterDeleteStuck, BackendAsyncOperationExternalAuthDeleteStuck, BackendAsyncOperationClusterCreateStuck, BackendAsyncOperationNodePoolDeleteStuck, BackendAsyncOperationNodePoolCreateStuck, BackendAsyncOperationStuck. Owner @stevekuznetsov; durable fix is the AMW auto-scaler (https://github.com/Azure/ARO-HCP/pull/6036). Marked known so it stops blocking the merge queue until that lands."
+  match:
+  - name: "BackendAsyncOperation.*Stuck"
+- name: "expected-noise-klusterlet-pods"
+  tier: 0
+  policy: ignore
```
deads2k, line 101: There should never be a policy of "ignore". This is indicative of an alert expression that needs refinement to indicate when it actually needs to fire.

deads2k (changes requested):
1. how we are going to identify which alerts belong to which category in production
2. how we are going to set different thresholds for alerting in different categories
3. how we are going to set custom alert expression with similar/same names to better reflect when they need to fire.

stevekuznetsov:
I think I'm missing something here. Couple observations:

1. we wrote the alerts, converted them into jUnit and made them fail the job _with intention_
2. for many months, we did not have issues with these alerts
3. presumably, if they are firing, something has gone wrong and we should look into it

The previous pull request and this one are focused on massaging the downstream implications of a firing alert, instead of ... addressing why they are firing in the first place. I do not see any analysis of why we have seen an uptick or why we think the uptick is permanent and cannot be remediated. 

I could see two options for changes to the allow-list:

1. if we have someone working solely on root-causing the issue and we want to unblock the rest of the team in the meanwhile
2. if we root-cause the issue and determine that for `$reason` some extremely limited subset of the alert (see how the other entries in the list have label selectors?) is irrecoverable, so our alert there is still useful in prod but not useful signal for tests

If we determine that the alert is bad, we fix the alert. If we determine that the system or infra are bad, we fix the system or infra.

Changing the skip-list is the _last_ order of business and done as a last-ditch effort when all else fails.

test/cmd/aro-hcp-tests/gather-observability/alert-categories/categories.yaml
cssjr: I was wondering if these should be at least "warn"

cssjr:
Hi @stevekuznetsov thanks for the feedback.  My impression of the ARO bug and feedback on the earlier draft is that the purpose/scope of this is detection, alerting, and categorization to bring visibility to some of these items.

If the visibility is already there and the intent of the ARO is to fix the underlying root causes, I can pivot that direction.

cc: @deads2k

stevekuznetsov:
All alerts that fire (in dev or otherwise) are ingested into a central Kusto cluster, so if there is some need to track the incidence of alerts over time, build dashboards for visibility, etc, I would recommend an ADX page. 

That being said, I do not know that we have such a need (ad-hoc queries are certainly possible today).

PR #6618, opened 2026-08-20
Fix: Improve correctness and usability of Velero Grafana Dashboard

tmstff:
Improve correctness and usability of Velero Grafana Dashboard

https://redhat.atlassian.net/browse/ARO-29199

### What

add backup attempt count; fix Backup Success Rate; differentiate Validation failres (Failure, Parital Failure, Validation failure); unify display for Success Rates, Failures Rates and attempt counts for Backup Deletion, CSI Snapshots and Valume Snapshots; optimize layout; optimize display of data in Velero CR inventory; fix Backup Duration Heatmap;

Looks like this now:
image

Checkout here:
https://arohcp-prod-g5d9a9akashnb5gd.suk.grafana.azure.com/goto/ffwcsp59lk54wd?orgId=1

### Why

To have a correct and useful Dashboard to monitor Velero Backup

### Testing

Manual Tryout in int

### PR Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/observability/grafana-dashboards/velero/velero.json
+++ b/observability/grafana-dashboards/velero/velero.json
@@ -19,7 +19,6 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
-  "id": null,
   "links": [],
   "panels": [
     {
@@ -31,16 +30,16 @@
         "y": 0
       },
       "id": 101,
+      "panels": [],
       "title": "Backup Health Overview",
-      "type": "row",
-      "panels": []
+      "type": "row"
     },
     {
       "datasource": {
         "type": "prometheus",
         "uid": "${datasource}"
       },
-      "description": "Total number of successful backups",
+      "description": "Total number of backups - and number of backup attempts in the selected time range",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -61,7 +60,8 @@
             "mode": "absolute",
             "steps": [
               {
-                "color": "green"
+                "color": "green",
+                "value": 0
               }
             ]
           },
@@ -70,8 +70,8 @@
         "overrides": []
       },
       "gridPos": {
-        "h": 4,
-        "w": 4,
+        "h": 6,
+        "w": 8,
         "x": 0,
         "y": 1
       },
@@ -80,7 +80,8 @@
         "colorMode": "value",
         "graphMode": "area",
         "justifyMode": "auto",
-        "orientation": "auto",
+        "orientation": "vertical",
+        "percentChangeColorMode": "standard",
         "reduceOptions": {
           "calcs": [
             "lastNotNull"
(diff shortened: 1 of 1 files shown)


observability/grafana-dashboards/velero/velero.json
```diff
@@ -632,189 +789,194 @@
...
+          "legendFormat": "CSI Snapshot",
+          "range": false,
+          "refId": "B"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "expr": "",
```
hbhushan3: **Dead query target.** This `refId: C` target has an empty `expr`, so it contributes nothing to the "Success Rates" gauge (panel `109`) — looks like leftover state from editing in the UI. Can you drop this target?

observability/grafana-dashboards/velero/velero.json
```diff
@@ -632,189 +789,194 @@
...
+        "y": 11
+      },
+      "id": 312,
+      "options": {
+        "code": {
+          "language": "plaintext",
+          "showLineNumbers": false,
+          "showMiniMap": false
+        },
+        "content": "Schedules with >46 Hours Since Last Backup",
```
hbhushan3: **Threshold mismatch (46h vs 48h).** This label and the stat below it (panel `313`) count schedules at `> 46*3600`, but the adjacent "Hours Since Last Backup" table (panel `111`) colours at `24h` / `48h` and its description says *"Green < 24h, Orange < 48h, Red >= 48h"*.

So a schedule sitting at 47h gets counted here but still renders orange in the table next to it, which is confusing for the on-call. Can we settle on one number (48h, to match the table thresholds) and use it in both the query and the label text?

observability/grafana-dashboards/velero/velero.json
```diff
@@ -92,71 +93,85 @@
...
-      "title": "Backup Total",
+      "title": "Backups",
       "type": "stat"
     },
     {
       "datasource": {
         "type": "prometheus",
         "uid": "${datasource}"
       },
+      "description": "Total number of successful backups",
```
hbhushan3: **Copy-paste description.** This panel (`309`) is titled "Attempts" and shows backup deletion / CSI snapshot / volume snapshot *attempts*, but the description says "Total number of successful backups". The same string is also on panel `102` ("Backups"), where it's equally wrong — that one shows all-time total + attempts.

Since this PR is explicitly about dashboard correctness, worth fixing both descriptions.

observability/grafana-dashboards/velero/velero.json
```diff
@@ -850,4105 +1024,4506 @@
...
+      "pluginVersion": "12.4.6",
       "targets": [
         {
           "datasource": {
             "type": "prometheus",
             "uid": "${datasource}"
           },
           "editorMode": "code",
-          "expr": "ALERTS{alertname=~\"Velero.*\", alertstate=\"firing\", cluster=~\"$cluster\"}",
+          "expr": "  count(\r\n    (time() - max without(prometheus_replica) (velero_backup_last_successful_timestamp{cluster=~\"$cluster\", schedule=~\"$schedule\", schedule!=\"\"}))     > 46*3600\r\n  ) or vector(0)\r\n",
```
hbhushan3: **Windows line endings embedded in the PromQL string.** This `expr` contains literal `\r\n` sequences (`"  count(\r\n    (time() - ...`). Harmless to Prometheus, but it makes the query render oddly in the editor and adds noise to future diffs. Worth normalising to `\n` (or collapsing to a single line) while you're in here.

observability/grafana-dashboards/velero/velero.json
```diff
@@ -850,4105 +1024,4506 @@
...
+          "pluginVersion": "12.4.6",
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "${datasource}"
+              },
+              "editorMode": "code",
+              "exemplar": false,
+              "expr": "sum(max without(prometheus_replica) (velero_backup_duration_seconds_bucket{cluster=~\"$cluster\", schedule=~\"$schedule\",le!=\"+Inf\"})) by (le)",
```
hbhushan3: **Heatmap is summing a cumulative counter.** `velero_backup_duration_seconds_bucket` is a monotonically increasing histogram counter, and this query sums it raw with no `increase()` / `rate()`. Grafana will de-cumulate along the `le` axis, but not along time — so every time column shows the *all-time* duration distribution rather than the backups that happened in that bucket of time. The heatmap ends up monotonically getting denser and doesn't really answer "were backups slow recently?".

If the intent is "duration distribution over the window", this probably wants:

```promql
sum(max without(prometheus_replica) (increase(velero_backup_duration_seconds_bucket{cluster=~"$cluster", schedule=~"$schedule"}[$__rate_interval]))) by (le)
```

Separately, `le!="+Inf"` drops the overflow bucket, so any backup slower than the largest bucket disappears from the panel entirely rather than showing up as an outlier.

Non-blocking / happy to be wrong if it looked right in int — but worth a second look since "fix Backup Duration Heatmap" is one of the stated goals.

observability/grafana-dashboards/velero/velero.json
```diff
@@ -632,189 +789,194 @@
...
       "targets": [
         {
           "datasource": {
             "type": "prometheus",
             "uid": "${datasource}"
           },
+          "editorMode": "code",
           "exemplar": false,
-          "expr": "sum(max without(prometheus_replica) (velero_backup_deletion_success_total{cluster=~\"$cluster\", schedule=~\"$schedule\"})) / clamp_min(sum(max without(prometheus_replica) (velero_backup_deletion_attempt_total{cluster=~\"$cluster\", schedule=~\"$schedule\"})), 1)",
+          "expr": "sum(max without(prometheus_replica) (increase(velero_backup_deletion_success_total{cluster=~\"$cluster\", schedule=~\"$schedule\"}[$__range]))) / sum(max without(prometheus_replica) (increase(velero_backup_deletion_attempt_total{cluster=~\"$cluster\", schedule=~\"$schedule\"}[$__range])))",
```
hbhushan3: **Inconsistent divide-by-zero handling across the rate panels.** There are three different behaviours in this dashboard when the denominator (attempts) is zero:

1. `... / clamp_min(sum(...attempt_total...), 1)` — most of the failure-rate panels. Renders **0%**.
2. `... / sum(...attempt_total...)` — this line, plus CSI snapshot, volume snapshot and restore success rates. Renders **NaN / N/A**.
3. `... / (sum(...attempt_total...) > 0)` — the per-schedule success rates. Renders **No data**.

So on a quiet cluster the "Success Rates" gauge shows a mix of `0%`, `N/A` and empty depending on which sub-metric you look at, which reads like something is broken rather than "nothing happened". Could we pick one convention and apply it everywhere?

FWIW `clamp_min(x, 1)` has its own wrinkle: `increase()` extrapolates and can return fractional values, so with 1 attempt / 1 failure in the window you can get `0.7 / clamp_min(0.7, 1)` = 70% instead of 100%. Option 3 (`> 0`) is the most honest of the three.

Not a blocker for this PR — fine as a follow-up if you'd rather keep the diff scoped.

hbhushan3 (commented):
Went through this end-to-end — the dashboard is a clear improvement and I'm happy with the direction. Blast radius is limited to Grafana rendering, so nothing here is scary.

Six comments inline. **(1)–(4) are small cleanups** I'd like to see before I `/approve`; **(5)–(6) are questions/follow-ups** and I won't hold the PR on them:

1. Dead query target with an empty `expr` in the "Success Rates" gauge
2. 46h vs 48h threshold mismatch between the stale-backup count and the table next to it
3. Copy-paste `description` ("Total number of successful backups") on the "Backups" and "Attempts" panels
4. Literal `\r\n` inside a PromQL string
5. Backup Duration Heatmap sums a cumulative counter without `increase()`
6. Three different divide-by-zero conventions across the rate panels

Things I checked so nobody else has to re-do it:

- No duplicate panel IDs, and `uid: ozk-vlr-mon` is unchanged, so this updates the existing dashboard in place.
- `"id": null` removal and `"version": 1 → 42` are just Grafana export artifacts — provisioning ignores both.
- The baked-in `datasource.current` of `Managed_Prometheus_services-uksouth` looks alarming for multi-region, but ~28 other dashboards in this repo already do the same thing, so it's the existing convention here and not a regression from this PR.
- The panels missing a `cluster=~"$cluster"` filter (the `by (cluster)` overview panels) are **pre-existing** — identical in `main`.
- `schemaVersion` 41 → 42 / `pluginVersion` 11.6.9 → 12.4.6 is only a risk if some AMG instance lags int's Grafana version. AMG versions roll globally, so low risk — flagging only for awareness.
- The failing `ci/prow/e2e-parallel` is unrelated; a dashboard JSON can't affect e2e.

observability/grafana-dashboards/velero/velero.json
tmstff: yeah, that really made no sense - don't know how it ended up this way 🤷

observability/grafana-dashboards/velero/velero.json
tmstff: Replaced it with a time series "Backup Durations" - which seems to serve the same purpose and actually works as far as I can tell. The heatmap approach had several pitfalls and I gave up on finding a suitable solution for it after investing quite some time.

tmstff:
> Went through this end-to-end — the dashboard is a clear improvement and I'm happy with the direction. Blast radius is limited to Grafana rendering, so nothing here is scary.
> 
> Six comments inline. **(1)–(4) are small cleanups** I'd like to see before I `/approve`; **(5)–(6) are questions/follow-ups** and I won't hold the PR on them:
> 
> ...

@hbhushan3 thanks a lot for your detailed and very helpful review!

I took care of all the findings - please take a look again! I hope it's not too much of a bother that I already squashed the fix commits into a single one for this branch ...

raelga (approved):
/lgtm
/approve

raelga (approved):
/lgtm
/approve

PR #6621, opened 2026-08-20
feat(gather-observability): add CosmosDB standard-metrics panel

stevekuznetsov:
Add an Azure Monitor platform-metrics query source alongside the existing Prometheus path so panels can plot metrics that aren't scraped into the managed Prometheus workspaces. Use it for a new "CosmosDB Metrics" panel that charts NormalizedRUConsumption (Max) and AutoscaledRU (Max) on one plot for the RP Cosmos DB account.

QuerySpec gains source/resource/aggregation/metrics fields with per-source validation; azureMonitor results are converted into the existing PrometheusResult shape so chart rendering is reused unchanged. The Cosmos account resource ID is assembled from --subscription-id plus regionRG and frontend.cosmosDB.name in the rendered config.


deads2k:
sure, let's try it

/lgtm

PR #6624, opened 2026-08-20
feat: publish hcpctl linux/windows binaries via GoReleaser

mmazur:
## Summary
- hcpctl currently has no build/publish pipeline producing downloadable binaries — only local `make build` and ad-hoc Windows cross-compile Makefile targets.
- Adds `tooling/hcpctl/.goreleaser.yaml` building raw `linux/amd64`, `windows/amd64`, and `windows/arm64` binaries (no archive wrapping).
- Adds `.github/workflows/hcpctl-release.yml`, which runs hourly (plus manual dispatch) and publishes to a single rolling `hcpctl-nightly` GitHub Release (replaced/overwritten each run, since hcpctl isn't versioned). A change-detection step skips the tag move and build unless `tooling/hcpctl/` actually changed since the last successful build, so it isn't rebuilding/republishing every hour unconditionally.
- Verified end-to-end (tag push, build, GitHub Release creation/upload) by temporarily running the workflow against a personal fork before opening this PR.

### DIFF
--- a/.github/workflows/hcpctl-release.yml
+++ b/.github/workflows/hcpctl-release.yml
@@ -1,0 +1,58 @@
+name: "hcpctl Nightly Build"
+on:
+  schedule:
+  # Check hourly; the "skip if unchanged" step below keeps this a no-op
+  # unless tooling/hcpctl actually changed since the last successful build.
+  # The nightly tag is only moved after a successful GoReleaser publish, so
+  # a failed run doesn't mask itself as "up to date" for the next check.
+  - cron: '0 * * * *'
+  workflow_dispatch:
+permissions:
+  contents: write
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+        ref: main
+        fetch-depth: 0
+    - name: Check for changes since last nightly build
+      id: check
+      run: |
+        git fetch origin refs/tags/hcpctl-nightly:refs/tags/hcpctl-nightly || true
+        if git rev-parse -q --verify refs/tags/hcpctl-nightly >/dev/null && \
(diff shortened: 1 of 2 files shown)


mmazur:
Addressed the Copilot review comments:

- Fixed the release header wording (it's not necessarily "the latest commit that touched tooling/hcpctl", since it also runs hourly/on dispatch).
- Added a `concurrency` group so overlapping runs queue instead of racing on the shared `hcpctl-nightly` tag/release.
- `workflow_dispatch` now explicitly checks out `main` regardless of the ref it's dispatched from, so manual runs can't accidentally publish from an arbitrary branch.

Two comments I'm not changing, with reasoning:

- **`--skip=validate`**: this is required, not just a "mask issues" flag. `hcpctl` isn't versioned, so this uses a rolling non-semver tag (`hcpctl-nightly`) rather than real release tags. I verified locally that without `--skip=validate`, GoReleaser hard-fails with `failed to parse tag 'hcpctl-nightly' as semver: invalid semantic version` — so it's structurally necessary for this rolling-release approach, not optional.
- **`release.replace_existing_artifacts`**: this is a valid GoReleaser v2 config key (confirmed via `goreleaser jsonschema`), and it was present during a successful end-to-end test run I did against a personal fork (tag pushed, binaries built, release created with all assets). `release.mode: replace` recreates the release itself; `replace_existing_artifacts` additionally covers the case where an asset with the same filename already exists on that release.

mmazur:
Addressed a follow-up suppressed Copilot comment: the nightly tag was being force-moved *before* running GoReleaser, so a failed GoReleaser run would still leave the tag advanced, causing the next scheduled check to wrongly treat it as "up to date" and skip retrying.

Fixed by moving the tag-move step to *after* a successful GoReleaser run. While implementing this I initially tried removing the manual tag push entirely in favor of GoReleaser's `target_commitish`, but verified end-to-end on a personal fork that GitHub's release API only honors `target_commitish` when creating a brand new tag — once `hcpctl-nightly` exists, subsequent releases silently keep the tag pinned to its original commit. So the explicit `git tag -f && push --force` step is still required; it's now just ordered after the GoReleaser step instead of before.

Verified with two consecutive real runs against a fork: tag correctly re-pointed to the new commit both times, release content updated correctly. Test artifacts are visible at https://github.com/mmazur/ARO-HCP/releases/tag/hcpctl-nightly for anyone who wants to inspect them.

PR #6641, opened 2026-08-21
feat(backup/restore): add etcdctl snapshot backup/restore

tony-schndr:
### What
Adds velero pre/post hooks to take an etcdctl snapshot before the PVC is CSI snapshotted to capture etcd data for recovery.
Documents how to perform an etcd only restore on a hosted control plane using --mark-compacted and --bump-revision to invalidate caches, preserving the customers data plane.

### Why

A temporary solution to avoid tear down of resources that was required when only using CSI snapshots.  ARO-HCP plans to move to https://redhat.atlassian.net/browse/CNTRLPLANE-2676 once viable.

Original attempts at restoring by growing etcd members did not work, the stateful set that hypershift uses for etcd has a reset-member init container that replaces learners with voters.  Therefore, the restore procedure requires selecting the best etcd db (highest revision), copying that db to all PVCs, then restoring all etcd data directories with etcdctl snapshot restore.

The reason the db is copied to all PVCs is because there does not seem to be a straightforward method to upload the database file to storage account.  This requires a Managed Identity that and the backup path, of which velero already has but requires a custom plugin to pull it off.  This is what [CNTRLPLANE-2676](https://redhat.atlassian.net/browse/CNTRLPLANE-2676) will provide when ARO-HCP can move to it.

### Testing

Restored one hosted control plane following the documented procedure.  Verified backed up config map was restored, and all cluster operators were healthy.

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/docs/backups.md
+++ b/docs/backups.md
@@ -66,12 +66,56 @@ Captured resources include:
 - HyperShift resources: `hostedcluster`, `nodepool`, `hostedcontrolplane`
 - Cluster API resources: `cluster`, `machinedeployment`, `machineset`, `machine`, `clusterdeployment`
 - Azure-specific resources: `azurecluster`, `azuremachine`, `azuremachinetemplate`
-- Standard workload resources: `deployments`, `statefulsets`, `configmap`, `secrets`, `services`, `sa`, `role`, `rolebinding`, `secretproviderclass`, `route`
+- Standard workload resources: `deployments`, `statefulsets`, `pods`, `configmap`, `secrets`, `services`, `sa`, `role`, `rolebinding`, `secretproviderclass`, `route`
+  - `pods` must be included for the [etcd snapshot hooks](#etcd-snapshot-hooks) to fire — Velero only runs exec hooks against pods that are part of the backup.
 - Policy resources: `priorityclasses`, `pdb`
 - Storage resources: `pvc`, `pv`
 - Swift Networking resources: `multitenantpodnetworkconfigs`, `podnetworks`,
 
 Volume snapshots are enabled. Snapshot data is moved to the backup storage location so it is durable outside of the originating Azure region.
+
+### etcd snapshot hooks
+
+A raw block-level snapshot of etcd PVCs does not provide a method to reset informer/kubelet caches on the dataplane, therefore backup hooks
+are used to capture an etcd snapshot. Each backup carries a Velero exec hook (`etcd-snapshot`), scoped to the control-plane namespace and
+selecting pods labeled `app=etcd`, with both a pre-backup and a post-backup phase that run inside the `etcd` container.
+
+**Pre-backup hook** (`onError: Fail`, 10-minute timeout) — takes the logical snapshot and flushes it to the underlying block device:
+
+```sh
+ETCDCTL_API=3 etcdctl \
+  --endpoints=https://etcd-client.<control-plane-ns>.svc:2379 \
+  --cacert /etc/etcd/tls/etcd-ca/ca.crt \
+  --cert /etc/etcd/tls/client/etcd-client.crt \
+  --key /etc/etcd/tls/client/etcd-client.key \
+  snapshot save /var/lib/backup.db \
+  && sync -f /var/lib/backup.db
+```
+
+The trailing `sync -f` is required for restore correctness. `etcdctl snapshot save` writes to `<path>.part`, fsyncs that file, then atomically renames it to `<path>` — but it does **not** fsync the parent directory. Because the CSI snapshot captures block-device state, the rename (a directory metadata operation still sitting in page cache / the filesystem journal) can be missing from the snapshot, leaving only `backup.db.part` on the restored PVC. `sync -f <path>` issues `syncfs()` on the data PVC, committing the file data **and** the rename to the block device before Velero snapshots the volume. Velero runs the pre-hook to completion before snapshotting the pod's volumes, so the flush is guaranteed to land first.
+
+**Post-backup hook** (`onError: Continue`, 2-minute timeout) — deletes the snapshot to reclaim space on the data PVC once Velero has finished snapshotting the pod's volumes:
+
+```sh
+rm -f /var/lib/backup.db /var/lib/backup.db.part
+```
+
+The post-hook uses `onError: Continue` because the snapshot has already been captured by the time it runs; a failed cleanup should not fail the backup. (This replaces the previous behavior of leaving `/var/lib/backup.db` in place to be overwritten on the next run.)
+
+The pre-backup hook plays two independent roles:
+
+- **Where the snapshot data comes from** — the `--endpoints` value is the `etcd-client` ClusterIP Service, not a pod or the headless discovery service. `etcd-client` leaves `publishNotReadyAddresses` unset (false), so it routes only to Ready endpoints; the etcd `readinessProbe` (`readyz` on port 9980) reflects serving health, so the Service load-balances to a healthy member. `etcdctl snapshot save` opens a single gRPC connection, so the snapshot is pulled from exactly one healthy member — the readiness semantics do the "pick a healthy member" for us. The headless `etcd-discovery` service must **not** be used here: it sets `publishNotReadyAddresses: true` and could hand back a crash-looping member.
+- **Which pods Velero execs into** — the hook selects all etcd members via the `app=etcd` label. Velero runs the exec once per matched pod, so every running member writes a `backup.db` to its own PVC. This is deliberately independent of any single pod's health: the backup does not hinge on `etcd-0` being up. The snapshot *data* in every case still comes from a healthy member via the Service, so each member's `backup.db` is equally consistent.
+
+`/var/lib/backup.db` sits on the member's data PVC (mounted at `/var/lib`) but outside etcd's live data directory (`/var/lib/data`, set via `ETCD_DATA_DIR`), so live etcd ignores the file while the CSI snapshot of that PVC still captures it. Velero runs the pod pre-hook before snapshotting that pod's volumes, so the clean snapshot is present when the volume snapshot is taken. The post-hook removes the file after the volume snapshot completes, so it does not accumulate on the data PVC between runs.
+
+The pre-hook uses `onError: Fail`. If a member's exec fails, Velero records the error, skips that pod, and continues — the other members' snapshots and all CSI volume snapshots are still taken, and the backup is marked **PartiallyFailed** (it does *not* abort the whole backup). This makes a failing member visible rather than letting a total failure (e.g. a bad cert path on every member) silently pass as it would with `onError: Continue`.
+
+### Restoring from the etcd snapshot
+
+All etcd PVCs (`data-etcd-0/1/2`) are CSI-snapshotted, and each running member's snapshot contains its own `backup.db`. Because the pre-hook fires per-pod at slightly different instants, those files are individually consistent but usually captured at different revisions, so there is no single designated "primary" snapshot.
+
+The full restore procedure — selecting a backup, picking the freshest snapshot, and rebuilding a 3-member etcd cluster from it — is documented in [etcd-restore.md](etcd-restore.md).
 
 ## Components
 
--- a/docs/etcd-restore.md
+++ b/docs/etcd-restore.md
@@ -1,0 +1,644 @@
+# SOP — Restore a HyperShift (ARO-HCP) etcd cluster from a Velero backup
+
+**What this does:** rebuilds a Managed 3-member etcd cluster for a HostedControlPlane
+from a Velero backup, using one canonical snapshot restored identically
+into all three PVCs (so all three boot as ONE `state=new` cluster).
+
+**When to use:** etcd data loss / split-brain / a member restored single-node, and
+you need a clean HA etcd from a known-good backup.
+
+---
+
+## CAUTION
+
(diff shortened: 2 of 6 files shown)


internal/backup/backup_test.go
```diff
@@ -0,0 +1,89 @@
...
+	assert.Contains(t, cmd, "ETCDCTL_API=3")
+	assert.Contains(t, cmd, "snapshot save /var/lib/backup.db")
+	// The rename done by `snapshot save` must be flushed to the block device before
+	// the CSI snapshot, or the PVC restores with only backup.db.part (rename lost).
+	assert.Contains(t, cmd, "&& sync -f /var/lib/backup.db", "hook must fsync the data PVC after the snapshot save to avoid the .part race")
+	assert.Contains(t, cmd, "--cacert /etc/etcd/tls/etcd-ca/ca.crt")
+	assert.Contains(t, cmd, "--cert /etc/etcd/tls/client/etcd-client.crt")
+	assert.Contains(t, cmd, "--key /etc/etcd/tls/client/etcd-client.key")
+	// Snapshot is pulled from the client Service (healthy member), not localhost or a pod.
+	assert.Contains(t, cmd, "--endpoints=https://etcd-client.ocm-cp.svc:2379")
```
tmstff: Monk nitpick: have same parameter order for verification as in code under test (makes human verification easier ;-) ) - feel free to ignore this one :-)

```suggestion
	assert.Contains(t, cmd, "ETCDCTL_API=3")
	assert.Contains(t, cmd, "--endpoints=https://etcd-client.ocm-cp.svc:2379")
	assert.Contains(t, cmd, "--cacert /etc/etcd/tls/etcd-ca/ca.crt")
	assert.Contains(t, cmd, "--cert /etc/etcd/tls/client/etcd-client.crt")
	assert.Contains(t, cmd, "--key /etc/etcd/tls/client/etcd-client.key")
	// Snapshot is pulled from the client Service (healthy member), not localhost or a pod.
	assert.Contains(t, cmd, "snapshot save /var/lib/backup.db")
	// The rename done by `snapshot save` must be flushed to the block device before
	// the CSI snapshot, or the PVC restores with only backup.db.part (rename lost).
	assert.Contains(t, cmd, "&& sync -f /var/lib/backup.db", "hook must fsync the data PVC after the snapshot save to avoid the .part race")
```

internal/backup/backup_test.go
```diff
@@ -0,0 +1,89 @@
...
+	// snapshot is cut, so backup.db does not linger during normal operation.
+	require.Len(t, hook.PostHooks, 1, "expected exactly one post-hook to clean up backup.db")
+	post := hook.PostHooks[0].Exec
+	require.NotNil(t, post, "post-hook must be an exec hook")
+	assert.Equal(t, "etcd", post.Container)
+	// Cleanup must never fail an otherwise-good backup.
+	assert.Equal(t, velerov1api.HookErrorModeContinue, post.OnError, "post-hook cleanup must not fail the backup")
+	require.Len(t, post.Command, 3, "post-hook command should be a shell invocation")
+	assert.Contains(t, post.Command[2], "rm -f /var/lib/backup.db", "post-hook must remove the snapshot artifact from the live PVC")
+	assert.Contains(t, post.Command[2], "/var/lib/backup.db.part", "post-hook should also remove a stray .part from an interrupted save")
```
tmstff: not part of the implementation yet as it seems

docs/etcd-restore.md
```diff
@@ -0,0 +1,289 @@
...
+  kubectl --kubeconfig=$KC -n $HCP_NS exec etcd-holder-${i} -- etcdutl snapshot status /var/lib/backup.db -w json
+done
+
+# 5c. pull canonical local (EDIT the source index to the freshest), verify, push to the other two
+export CANON=<index>
+kubectl --kubeconfig=$KC -n $HCP_NS cp etcd-holder-${CANON}:/var/lib/backup.db /Users/tschneid/devel/tmp/canonical-backup.db
+shasum -a 256 /Users/tschneid/devel/tmp/canonical-backup.db
+for i in 0 1 2; do
+  [ "$i" = "$CANON" ] && continue
+  kubectl --kubeconfig=$KC -n $HCP_NS cp /Users/tschneid/devel/tmp/canonical-backup.db etcd-holder-${i}:/var/lib/backup.db
```
tmstff: to make it copy-pastable (in case this is supposed to stay for a while until the implementation for automating this is finished)
```suggestion
kubectl --kubeconfig=$KC -n $HCP_NS cp etcd-holder-${CANON}:/var/lib/backup.db /tmp/canonical-backup.db
shasum -a 256 /tmp/canonical-backup.db
for i in 0 1 2; do
  [ "$i" = "$CANON" ] && continue
  kubectl --kubeconfig=$KC -n $HCP_NS cp /tmp/canonical-backup.db etcd-holder-${i}:/var/lib/backup.db
```

docs/backups.md
```diff
@@ -66,13 +66,53 @@ Captured resources include:
...
+```sh
+ETCDCTL_API=3 etcdctl \
+  --endpoints=https://etcd-client.<control-plane-ns>.svc:2379 \
+  --cacert /etc/etcd/tls/etcd-ca/ca.crt \
+  --cert /etc/etcd/tls/client/etcd-client.crt \
+  --key /etc/etcd/tls/client/etcd-client.key \
+  snapshot save /var/lib/backup.db \
+  && sync -f /var/lib/backup.db
+```
+
```
tmstff, line 94:
```suggestion

Note: A PVC is mounted at /var/lib, so backup.db is stored on the PVC and therefore backed up by Velero.

```

docs/etcd-restore.md
```diff
@@ -0,0 +1,289 @@
...
+> **GATE:** all three logs must show the **same `cluster-id`** and the same three
+> `added member` IDs. That is what makes them one cluster on boot.
+
+---
+
+## 7. C-3 — Start the cluster
+
+```bash
+kubectl --kubeconfig=$KC -n $HCP_NS scale statefulset/etcd --replicas=3
+kubectl --kubeconfig=$KC -n $HCP_NS rollout status statefulset/etcd --timeout=10m
```
tmstff: don't we need to scale the kube-apiserver up as well again?

We scaled it down to 0 ealier: `kubectl --kubeconfig=$KC -n $HCP_NS scale deployment/kube-apiserver --replicas=0`

docs/etcd-restore.md
```diff
@@ -0,0 +1,289 @@
...
+  name: $RESTORE_NAME
+  namespace: velero
+spec:
+  backupName: $BACKUP
+  restorePVs: true
+  existingResourcePolicy: update
+  includedResources:
+  - pv
+  - pvc
+  - statefulsets
```
tmstff, line 312: just to be sure: Does this scale up the stateful sets again? if so, they might need to be scaled down again.

tmstff: AI says, that the stateful set might not even be required ("the stated reason ("PVCs need the StatefulSet to bind") is incorrect for Velero datamover — PVCs bind to VolumeSnapshotContent-provisioned PVs directly; the StatefulSet is not required."

Maybe worth checking out.

docs/etcd-restore.md
```diff
@@ -0,0 +1,289 @@
...
+        volumeMounts: [{ name: data, mountPath: /var/lib }]
+      volumes:
+      - { name: data, persistentVolumeClaim: { claimName: data-etcd-${i} } }
+EOF
+done
+
+for i in 0 1 2; do
+  kubectl --kubeconfig=$KC -n $HCP_NS wait --for=condition=complete job/etcd-restore-${i}-$TS --timeout=5m
+  kubectl --kubeconfig=$KC -n $HCP_NS logs job/etcd-restore-${i}-$TS | grep -Ei "cluster-id|added member|restored snapshot|bumping"
+  kubectl --kubeconfig=$KC -n $HCP_NS delete job/etcd-restore-${i}-$TS
```
tmstff: maybe do the deletion when everything is fine at the end, because we might want to look into the logs in case something goes wrong?

docs/etcd-restore.md
```diff
@@ -0,0 +1,289 @@
...
+  -p '{"spec":{"pausedUntil":"true"}}'
+kubectl --kubeconfig=$KC -n $HC_NS get hostedcluster/$HC -o jsonpath='pausedUntil={.spec.pausedUntil}{"\n"}'
+```
+
+## 2. Scale etcd and KAS to 0
+
+```bash
+kubectl --kubeconfig=$KC -n $HCP_NS scale statefulset/etcd --replicas=0
+kubectl --kubeconfig=$KC -n $HCP_NS scale deployment/kube-apiserver --replicas=0
+kubectl --kubeconfig=$KC -n $HCP_NS wait --for=delete pod -l app=etcd --timeout=5m
```
tmstff: Maybe wait for KAS pod deletion as well?

deads2k: > Maybe wait for KAS pod deletion as well?

won't matter once etcd is unavailabl.e

docs/etcd-restore.md
tony-schndr: This PVC is on Velero's datamover path, not the direct CSI snapshot restore path — it has the velero.io/dynamic-pv-restore selector label and no spec.dataSource. On the datamover path there's no VolumeSnapshotContent to bind to (data comes from object storage via DataDownload), so "PVCs bind to VolumeSnapshotContent-provisioned PVs directly" doesn't apply here.

The StorageClass (managed-csi-premiumv2) is WaitForFirstConsumer — confirmed by the volume.kubernetes.io/selected-node annotation on the PVC, which the scheduler only sets for WFC volumes. Velero's node-agent explicitly waits on that: WaitPVCConsumed (pkg/util/kube/pvc_pv.go:443) blocks until selected-node is populated, and that annotation only appears once a consumer pod is scheduled. For etcd, that consumer is the StatefulSet pod. No StatefulSet → no scheduled pod → no selected-node → the DataDownload stalls and the restore times out.

One way around this might be to update the storage class volumeBindingMode, feels like a bigger change for a temporary backup strategy.

> just to be sure: Does this scale up the stateful sets again? if so, they might need to be scaled down again.

It will scale up etcd and yes you do have to scale it back down to restore.  I'm going to add an annotation so only etcd stateful set is restored.

docs/etcd-restore.md
tony-schndr: Unpausing the HostedCluster will scale KAS back up on step 9.

internal/backup/backup_test.go
tony-schndr: Yeah, its old.  I'm not a fan of these AI slop unit tests any way, not much value IMO.  I'm removing them before opening up this PR.

docs/etcd-restore.md
```diff
@@ -0,0 +1,547 @@
...
+  kubectl  -n $HCP_NS logs job/etcd-restore-${i}-$TS | grep -Ei "cluster-id|added member|restored snapshot|bumping"
+done
+
+# Validate restore was successful prior to removing pod
+for i in 0 1 2; do
+  kubectl  -n $HCP_NS delete job/etcd-restore-${i}-$TS
+done
+```
+
+> **GATE:** all three logs must show the **same `cluster-id`** and the same three
```
adalrsjr1, line 483: you could include the expected output here, we have a valid example here: https://github.com/openshift/hypershift/pull/9048#issuecomment-5241105054

docs/etcd-restore.md
```diff
@@ -0,0 +1,547 @@
...
+
+PVCs are labeled `app=etcd` only (no per-member label), so restore all three.
+
+The etcd StatefulSet is restored because the `data-etcd-*` PersistentVolumeClaims use a
+WaitForFirstConsumer storage class: the PersistentVolume is not provisioned, and the
+volume snapshot not restored, until a pod mounts the claim. The StatefulSet pods are that
+first consumer, so the StatefulSet must be restored to trigger binding and restore.
+NOTE: Match labels does not work for the StatefulSet, there is nothing to match on.
+```bash
+export RESTORE_NAME=restore-$HC-$TS
```
adalrsjr1, line 297: it's worth recall the reader about these parameters, at this point, during a troubleshoting, someone could lost track that, e.g., TS was first set in the very beginning of this process.

docs/etcd-restore.md
```diff
@@ -0,0 +1,547 @@
...
+
+## 2. Identify a valid backup
+
+Identifying the proper backup is a two step process.
+1. Identify when the problem that requires recovery first occurred.
+2. Select a backup that was successfully taken just before the start of the issue.
+
+
+```bash
+# List available backups for this cluster
```
adalrsjr1, line 106: nit: it would nice to have the format of a backup name for reference.

velero/deploy/templates/velero-rbac.yaml
```diff
@@ -17,6 +17,11 @@ rules:
   - serviceaccounts
   - namespaces
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete", "deletecollection"]
+# pods/exec is used for velero backup pre/post hooks
+- apiGroups: [""]
+  resources:
+  - pods/exec
+  verbs: ["create"]
```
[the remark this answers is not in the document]
tony-schndr: Comment is legit but this doesn't buy us much IMO.  Velero needs access to the highest value namespaces on the management clusters, the control plane namespaces...

PR #6659, opened 2026-08-24
feat: set node pool min version in the frontend to 4.20.8 for all users

JameelB:
JIRA: https://redhat.atlassian.net/browse/ARO-29063

### What

Disables support for node pool install and upgrades using ocp versions under 4.20.8 for all users.

This follows on from https://github.com/Azure/ARO-HCP/pull/6609

### Why

4.20.8 was selected for the min version as versions below this were missing some features, mostly related to allowing dynamic retrieval of azure marketplace images for node pools. Moving away from these versions will allow us to remove any hardcoded marketplace images in the future.

### Testing

- Unit tests
   - updated to accommodate for the new min version
   - removed test cases under experimental features that validated successful install of a node pool using a version under 4.20.8.
   - the [invalid-version/05-httpcreate-node-lower-than-min-version](https://github.com/Azure/ARO-HCP/tree/main/test-integration/frontend/artifacts/FrontendCRUD/NodePool/invalid-version/05-httpCreate-node-lower-than-min-version) covers the test for ensuring this isn't allowed for all users.
  - All e2e tests against 4.19 have been removed in https://github.com/Azure/ARO-HCP/pull/6609

N/A

### PR Checklist

- [ ] ~~Screenshots included (if graph/UI/metrics changes)~~
- [x] Self-reviewed the diff
- [ ] ~~Draft PR used for WIP (if applicable)~~

If E2E tests are included:

- [ ] ~~E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md)~~
- [ ] ~~If new E2E use case is covered (via a new test or new check/verifier),
  proper error message and logs which communicates nature of the problem.~~


hlipsig:
The copilot feedback looks good. Address those points and then I think this is good.

internal/validation/validate_nodepools.go
```diff
@@ -281,12 +281,12 @@ func validateNodePoolVersionProfile(ctx context.Context, op operation.Operation,
...
 	if !op.HasOption(metadataapi.FeatureExperimentalReleaseFeatures) {
 		errs = append(errs, validate.Enum(ctx, op, fldPath.Child("channelGroup"), &newObj.ChannelGroup, safe.Field(oldObj, toNodePoolVersionProfileChannelGroup), metadataapi.AllowedChannelGroups, nil)...)
-		// without feature flag, only allow version 4.20.8 and above
-		errs = append(errs, VersionMustBeAtLeast(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID), "4.20.8")...)
 	} else {
 		errs = append(errs, validate.Enum(ctx, op, fldPath.Child("channelGroup"), &newObj.ChannelGroup, safe.Field(oldObj, toNodePoolVersionProfileChannelGroup), metadataapi.AllowedChannelGroupsWithExperimentalFlag, nil)...)
 	}
 
+	errs = append(errs, VersionMustBeAtLeast(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID), "4.20.8")...)
+
```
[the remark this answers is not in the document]
JameelB: resolved by combining both validations into one single validator

internal/validation/validate_nodepools_comprehensive_test.go
```diff
@@ -1809,6 +1809,7 @@ func TestValidateNodePoolVersionWithFeatureFlags(t *testing.T) {
 			opOptions: testNodePoolFeatureOptions(metadataapi.FeatureExperimentalReleaseFeatures),
 			expectErrors: []utils.ExpectedError{
 				{Message: "Invalid character(s) found in patch number", FieldPath: "properties.version.id"},
+				{Message: "Invalid character(s) found in patch number", FieldPath: "properties.version.id"},
 			},
```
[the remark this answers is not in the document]
JameelB: see https://github.com/Azure/ARO-HCP/pull/6659#discussion_r3861594440

hlipsig (approved):
LGTM

internal/validation/validate_nodepools.go
```diff
@@ -271,25 +271,35 @@ func validateNodePoolVersionProfile(ctx context.Context, op operation.Operation,
...
 
+// validateNodePoolVersionID validates the version ID format and checks it against the minimum version requirement.
+func validateNodePoolVersionID(ctx context.Context, op operation.Operation, fldPath *field.Path, newObj, oldObj *coreapi.NodePoolVersionProfile) field.ErrorList {
+	errs := field.ErrorList{}
+
+	errs = append(errs, OpenShiftWithOptionalPrerelease(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID))...)
+	if len(errs) == 0 {
+		// Only validate if version is valid, otherwise we'll get a duplicate invalid version error
+		errs = append(errs, VersionMustBeAtLeast(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID), "4.20.8")...)
+	}
```
machi1990:
```suggestion
		// Only validate if version is valid, otherwise we'll get a duplicate invalid version error
		errs = append(errs, VersionMustBeAtLeast(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID), "4.20.8")...)
	
```

Always validate, even if it's duplicate the user gets to fix the two errors at the same time and not get an error par API request.

See https://github.com/JameelB/ARO-HCP/blob/0c591ceb36eb708fe96d339fc42b2d290210c245/internal/validation/validate_cluster.go#L443 how its done for cluster

machi1990 (changes requested):

internal/validation/validate_nodepools.go
```diff
@@ -271,25 +271,32 @@ func validateNodePoolVersionProfile(ctx context.Context, op operation.Operation,
...
 }
 
+// validateNodePoolVersionID validates the version ID format and checks it against the minimum version requirement.
+func validateNodePoolVersionID(ctx context.Context, op operation.Operation, fldPath *field.Path, newObj, oldObj *coreapi.NodePoolVersionProfile) field.ErrorList {
+	errs := field.ErrorList{}
+
+	errs = append(errs, OpenShiftWithOptionalPrerelease(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID))...)
+	errs = append(errs, VersionMustBeAtLeast(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID), "4.20.8")...)
+	return errs
+}
```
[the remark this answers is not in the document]
machi1990: This is fine.
It's done so that when the errors are actually different, we'll return the errors in one API call that user made instead of the user having to fix an error at a time.

machi1990 (approved):
/lgtm

internal/validation/validate_nodepools.go
JameelB: this has been addressed, thanks!

internal/validation/validate_nodepools.go
```diff
@@ -271,25 +271,32 @@ func validateNodePoolVersionProfile(ctx context.Context, op operation.Operation,
...
 
 	return errs
 }
 
+// validateNodePoolVersionID validates the version ID format and checks it against the minimum version requirement.
+func validateNodePoolVersionID(ctx context.Context, op operation.Operation, fldPath *field.Path, newObj, oldObj *coreapi.NodePoolVersionProfile) field.ErrorList {
+	errs := field.ErrorList{}
+
+	errs = append(errs, OpenShiftWithOptionalPrerelease(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID))...)
+	errs = append(errs, VersionMustBeAtLeast(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID), "4.20.8")...)
```
miguelsorianod, line 296: Are all currently existing nodepools in **all** aro-hcp permanent environments >= 4.20.8? the validation will be executed for both create and **updates**. If there are pre-existing nodepools that are < 4.20.8 this would start to produce validation errors during updates, which means we cannot merge this if that's the case.

internal/validation/validate_nodepools.go
```diff
@@ -271,25 +271,32 @@ func validateNodePoolVersionProfile(ctx context.Context, op operation.Operation,
...
+		errs = append(errs, validateNodePoolVersionID(ctx, op, fldPath, newObj, oldObj)...)
 	}
 
 	//ChannelGroup string `json:"channelGroup,omitempty"`
 	errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("channelGroup"), &newObj.ChannelGroup, safe.Field(oldObj, toNodePoolVersionProfileChannelGroup))...)
 
 	if !op.HasOption(metadataapi.FeatureExperimentalReleaseFeatures) {
 		errs = append(errs, validate.Enum(ctx, op, fldPath.Child("channelGroup"), &newObj.ChannelGroup, safe.Field(oldObj, toNodePoolVersionProfileChannelGroup), metadataapi.AllowedChannelGroups, nil)...)
-		// without feature flag, only allow version 4.20.8 and above
-		errs = append(errs, VersionMustBeAtLeast(ctx, op, fldPath.Child("id"), &newObj.ID, safe.Field(oldObj, toNodePoolVersionProfileID), "4.20.8")...)
```
miguelsorianod, line 285: Do I understand correctly that before this PR we were already forcing >= 4.2.0.8 for most of the end-users? this validation was only applying to subscriptions that do **not** have the metadataapi.FeatureExperimentalReleaseFeatures option, which means it only applied to our(RH) subscriptions (unless we've allowed some end-user to register it). So this is about removing this small subset that was still allowing it. Asking because the PR title and description is written as if it's a general relaxing, but this seems to be about what I described in the beginning of this comment.

If you were/are aware of this and that's the intention, that seems correct to me. Just raising to make sure everyone has the same shared understanding.

miguelsorianod:
/lgtm
/approve

but I put a /hold until the comments are addressed. The hold should only be removed if it's safe to do so as described in the comments I left. If verifications haven't been made they must be made before taking any decision.

internal/validation/validate_nodepools.go
JameelB: Hey yes that's correct, we were only allowing installs of <4.20.8 for select subscriptions that has this afec flag enabled - it was mostly for testing 4.19 which has now been removed in a subsequent [PR](https://github.com/Azure/ARO-HCP/pull/6609).

internal/validation/validate_nodepools.go
JameelB: Yes, can confirm there are no longer any clusters or node pools using versions under 4.20.8 in all environments

JameelB:
/hold cancel

We were only allowing installs of <4.20.8 for select subscriptions that has this afec flag enabled - it was mostly for testing 4.19 which has now been removed in a subsequent PR https://github.com/Azure/ARO-HCP/pull/6609. As of today, there are no longer any node pools using versions under 4.20.8 in all environments

PR #6673, opened 2026-08-25
hcpctl: publish rolling release to hcpctl-latest tag

mmazur:
## Summary

Follow-up to #6671. Switches the hourly hcpctl build from the `hcpctl-nightly` tag/release to the already-present `hcpctl-latest` tag, and never moves the tag.

## Changes

- **GoReleaser** (`tooling/hcpctl/.goreleaser.yaml`): `GORELEASER_CURRENT_TAG` → `hcpctl-latest`; release `name_template` → `hcpctl (latest)`.
- **Tag is never moved**: `hcpctl-latest` already exists as a tag. GoReleaser publishes the release against that existing (frozen) tag — GitHub's release API only honors `target_commitish` when *creating* a tag, and the tag already exists, so the ref stays put. The GitHub Release page still resolves; binaries refresh each run via `mode: replace`.
- **Tracking** keeps using the `refs/markers/` mechanism from #6671, renamed `refs/markers/hcpctl-nightly-built` → `refs/markers/hcpctl-latest-built`. Refs outside `refs/tags/` are invisible to Prow clonerefs' `git fetch --tags`, so no clobber.
- **Rename**: workflow name `hcpctl Nightly Build` → `hcpctl Latest Build`; release display name `hcpctl (nightly)` → `hcpctl (latest)`.

## Notes

- First run after merge: the `hcpctl-latest-built` marker won't exist yet, so change-detection treats it as "changed" and does one build — correct.
- The old `hcpctl-nightly` tag/release is left in place (no longer updated). It can be cleaned up separately.

---
*AI-generated. Review for accuracy.*

### DIFF
--- a/.github/workflows/hcpctl-release.yml
+++ b/.github/workflows/hcpctl-release.yml
@@ -1,4 +1,4 @@
-name: "hcpctl Nightly Build"
+name: "hcpctl Latest Build"
 on:
   schedule:
   # Check hourly; the "skip if unchanged" step below keeps this a no-op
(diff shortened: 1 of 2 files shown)


raelga:
Approach looks right to me. `hcpctl-latest` already exists and is frozen, GoReleaser publishes the release against it without moving the ref, and tracking stays on the `refs/markers/` ref so clonerefs' `--tags` fetch never sees it. That kills the clobber for good going forward.

One thing we still need to do on top of this: delete the stale `hcpctl-nightly` tag. This PR stops future moves but leaves that tag sitting at ba73a406, so the clonerefs caches that are already stuck on an older value keep clobbering until it's gone. Safe to delete now, nothing recreates it anymore.

/lgtm
/approve

mmazur:
Force merge justification:
We need to move away from the nightly tag for ci to stop failing and this job keep recreating it.

PR #6642, opened 2026-08-21
test(e2e): timebomb 4.20->4.21 minor upgrade until Cincinnati publishes edges

inbharajmani:
## Summary

- Scoped `Skip` in `test/e2e/control_plane_minor_upgrade.go` for the `candidate`/`fast` **4.20 → 4.21** case, expiring **2026-08-31**.
- Nightly and other minor jumps are unaffected.

## Why

All of the dev CI is failing on the test "Customer should be able to successfully upgrade control plane minor version from 4.20 minor to 4.21 minor"

Upstream `openshift/cincinnati-graph-data` promoted **4.20.35** as a node in `candidate-4.21` on 2026-08-21 (commit [12b0b5a5](https://github.com/openshift/cincinnati-graph-data/commit/12b0b5a5)), but has not yet added outgoing `4.20.35 → 4.21.z` edges. Because the RP picks the channel tip (offset 0) for `candidate`/`fast` (`GetZStreamOffset` in `backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go`), every 4.20 → 4.21 run lands on 4.20.35 and is rejected by frontend admission (`internal/admission/admit_cluster.go:670-723`) with:

> `no upgrade path to update channel "candidate-4.21" is currently available for this cluster`

Confirmed via graph query:

```
curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.21&arch=amd64' \
  | jq -r '.nodes as $n | .edges[] | select($n[.[0]].version=="4.20.35") | $n[.[1]].version' \
  | grep '^4\.21' || echo "no 4.20.35 -> 4.21.x edge yet"
```

Runtime evidence (int-uksouth, 2026-08-21 04:11Z): `serviceproviderclusters` snapshot for the failing cluster shows `desiredVersionChannels = ["candidate-4.20"]`, `activeVersions = [{"version":"4.20.35","state":"Completed"}]`.

Jira: [AROSLSRE-1853](https://redhat.atlassian.net/browse/AROSLSRE-1853)

## Follow-up

Replace the timebomb with a preflight that queries the Cincinnati graph for an outgoing edge from the resolved install version to the target minor and `Skip`s if absent — removes the race for good and self-heals across future z-stream promotions.

## Test plan

- [ ] CI green on this branch
- [ ] Verify the 4.20 → 4.21 entry reports `SKIP` (not fail) in E2E output while the timebomb is active
- [ ] Verify the 4.21 → 4.22 entry still runs

PR #6683, opened 2026-08-25
Don't enforce upgrade edge for candidate or nightly

deads2k:
There may not be a path from a candidate build to the next release. That's ok, don't enforce one.

We may even be able to eliminate it for fast and stable, but we dont' have to do that now.

Reverts the short-term changes in https://github.com/Azure/ARO-HCP/pull/6642 and https://github.com/Azure/ARO-HCP/pull/6652

### DIFF
--- a/internal/admission/admit_cluster.go
+++ b/internal/admission/admit_cluster.go
@@ -685,6 +685,11 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
 	if len(newObj.ChannelGroup) == 0 {
 		return nil
 	}
+	// on nightlies and candidates, we don't wait for the target version to show up. We trust these people to
+	// know what they're doing.
+	if newObj.ChannelGroup == "nightly" || newObj.ChannelGroup == "candidate" {
+		return nil
+	}
 
 	// No channel data yet: fail open (see the doc comment above). A genuinely
 	// missing ServiceProviderCluster prefetch is still surfaced as an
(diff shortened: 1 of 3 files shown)


internal/admission/admit_cluster.go
```diff
@@ -685,6 +685,11 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
 	if len(newObj.ChannelGroup) == 0 {
 		return nil
 	}
+	// on nightlies and candidates, we don't wait for the target version to show up. We trust these people to
+	// know what they're doing.
+	if newObj.ChannelGroup == "nightly" || newObj.ChannelGroup == "candidate" {
```
machi1990, line 690: FYI; A change in CS is needed as well to stop considering upgrade edges for nightly / candidate. 
I'll prepare the change

machi1990: CS MR is up and queued for merge in 30 mins.

machi1990:
/lgtm
/hold 

Unhold when CS MR's lands in ARO-HCP.

deads2k:
https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/428 must get merged. Consequences of out of order prod delivery is a fail closed, so we don't have to wait once it can reliably work in dev.

machi1990:
The CS image is being bumped in https://github.com/Azure/ARO-HCP/pull/6745; good to unhold once that PR lands

mvacula02:
fixes https://redhat.atlassian.net/browse/ARO-29278

PR #6684, opened 2026-08-25
feat(kusto): record CI job outcomes so alerts can be attributed to runs

galchammat:
## Why

The dev Kusto has no record of CI run outcomes. Its telemetry is keyed by cluster, so an alert can be found but not attributed to the run that produced it, and no table in `ServiceLogs` names a job, build, or test result.

That makes CI outcomes the missing join key: any question relating a CI run to the telemetry it produced has to be answered by hand-correlating Prow with Kusto, one investigation at a time.

The first payoff is separating an alert that failed a job from one that merely fired alongside a failure that had already happened. Seeded with real data, that separation is already visible (last 2 days):

| alert | job failed | alerts | runs |
| --- | --- | --- | --- |
| KubePodNotReady | true | 2570 | 59 |
| CPUThrottlingHigh | **false** | 184 | 9 |
| ServiceMemoryTrend | **false** | 103 | 9 |

Closes AROSLSRE-1626.

## What

1. **`ciJobOutcomes` table** (`dev-infrastructure/modules/logs/kusto/tables/`) — one row per Prow run. Written directly rather than fed from Event Hub, so no raw staging table or update policy; the ingestion mapping is kept so a data connection stays possible later.
2. **A writer in the `tenant-quota` collector** — reads runs from Sippy and queues them for ingestion the same way the cluster log forwarder writes.
3. **A Kusto grant** for the collector's existing identity, made from the opstool pipeline — `Ingestor` to write rows and `Viewer` to read back which runs are already stored, so re-reading a window cannot duplicate them.

### Why `tenant-quota` rather than a new service

It is already the one thing polling Prow on a timer, already runs as a single instance on `opstool`, and already has an identity with a federated credential — so the only genuinely new thing is the grant. `aro-hcp-exporter` would have been a cheaper grant (its principal is already in `grant-access.bicep` as a reader) but is deployed **per service cluster**, so every dev cluster would write the same rows into a store with no uniqueness constraint.

### Details worth a reviewer's eye

- **Runs are re-read from a fixed window rather than resumed from a watermark.** A run appears in Sippy only once it has completed, and runs complete out of order — a fast provisioning failure is recorded while a longer run that started earlier is still going. Resuming after the newest stored start time would therefore skip that longer run permanently. Re-reading the window is safe because runs already stored are subtracted by build id, and it self-heals: a gap inside the window is filled by the next pass. Kusto has no uniqueness constraint, so that subtraction is the only thing preventing duplicates, and it is also why the identity needs `Viewer` alongside `Ingestor`.
- **Only presubmit runs get a `clusterToken`.** Other releases run against clusters that already exist, whose telemetry lives in that environment's own Kusto, so a token derived from their build id would join to nothing here while looking like it should.
- **Batch and per-PR runs are told apart by URL, not job name** — they share a job name, and Sippy reports a pull request link for both.
- **Sippy's timestamp is decoded from either form.** It reports RFC 3339 today but has reported Unix milliseconds; that field is the fetch cursor, so a decode failure is silent and total.

## CI/CD configuration changes

Per the contributing guide, calling these out explicitly:

- `dev-infrastructure/dev-ci/cluster/opstool-aks-pipeline.yaml` — adds a `kusto-grant` ARM step.
- `dev-infrastructure/dev-ci/cluster/opstool-kusto-grant.bicep` and `opstool-kusto-grant-assignments.bicep` + a `.tmpl.bicepparam` — new; scoped to the Kusto cluster's resource group. The assignments are declared here rather than through the shared `grant-access.bicep` module because that module names every assignment after the principal and database only, so one principal holding both `Ingestor` and `Viewer` would produce two resources with the same name.

The grant step runs under the `DevCI.Unprivileged` entrypoint, i.e. the unattended dev-ci postsubmit. It is a Kusto `principalAssignments` resource rather than an Azure RBAC role assignment, so it should not hit the postsubmit bot's ABAC condition — but it does need write access on the Kusto cluster in `hcp-kusto-us`, outside the opstool resource group. **If that fails in postsubmit, the grant belongs on the `DevCI.Privileged` entrypoint instead.**

## Dependency

One new dependency, `azkustoingest`, pinned to **v1.2.1** to match the `azkustodata` already in the workspace. Only `tooling/tenant-quota/go.mod` and `go.sum` change; no other module is touched.

Taking the newer `azkustoingest` v1.2.2 instead would have dragged `azkustodata` from v1.2.1 to v1.2.2 across `test`, `hcpctl`, `kustoctl` and `aro-hcp-exporter` — which neither version of `azkustoingest` actually requires, since both declare `azkustodata v1.2.1`.

## Verification

- `go build`, `go vet`, `go test`, `gofmt` — clean
- `make verify-schema` — pass
- `make verify-materialize` — "No changes detected"
- `az bicep build` on both changed templates — clean
- Table created and seeded manually on `hcp-dev-us-2` for validation — 245 rows across all four releases, tokens present on presubmits only and absent on the promoted environments. The query results above were measured against it. That table has since been dropped, so the DDL in this PR is what creates it.
- `make validate-config-pipelines-dev-ci` **fails on `main` already**, in `mock-identity-rbac*.tmpl.bicepparam` (`range loop not allowed`). Unrelated to this change and none of the files here appear in its output.

## Screenshots

Not applicable — no dashboard, UI, or alerting-rule changes. The table above is the observable behaviour change and is reproducible with the query in the table's own docstring.


dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
```diff
@@ -0,0 +1,56 @@
+// Outcomes of CI jobs, one row per Prow run.
+//
+// Unlike the other tables in this database, this one is not fed from Event Hub:
+// the release dashboard writes it directly, so it needs no raw staging table and
+// no update policy.
```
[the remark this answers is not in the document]
galchammat: Fixed in 8595a48. Corrected both the header comment and the `docstring` to name the opstool tenant-quota collector, with a pointer to `tooling/tenant-quota/pkg/cijoboutcomes` so the writer is findable from the table definition. The "release dashboard" wording was left over from an earlier plan where the dashboard was going to own this — you're right that it would have sent operators to the wrong place.

config/config-dev-ci.schema.json
```diff
@@ -274,6 +274,7 @@
               "properties": {
                 "image": { "$ref": "#/$defs/Image" },
                 "prow": { "$ref": "#/$defs/ProwMetricsConfig" },
+        "ciJobOutcomes": { "$ref": "#/$defs/CIJobOutcomesConfig" },
```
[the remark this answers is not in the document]
galchammat: Fixed in 6552a7b. Was indented 8 spaces where its siblings use 16 — an artefact of scripting the edit. Re-done in place preserving the file's compact inline style, so the change is now +26/-0 rather than reflowing the file.

tooling/tenant-quota/pkg/cijoboutcomes/sippy.go
```diff
@@ -0,0 +1,217 @@
...
+//
+// Presubmits is the population that provisions its own clusters, so it is the
+// only one whose runs can be joined to the alerts and logs in this cluster. The
+// promoted environments are recorded for their job and test outcomes alone.
+var sippyReleases = []string{
+	presubmitsRelease,
+	"aro-integration",
+	"aro-stage",
+	"aro-production",
+}
```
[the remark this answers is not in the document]
galchammat: Fixed in 6552a7b. Removed — config is the single source of truth for releases. It was left behind when the release list moved into `ciJobOutcomes.releases`; `golangci-lint` flagged it as unused in the same run.

dev-infrastructure/dev-ci/cluster/opstool-kusto-grant.bicep
```diff
@@ -0,0 +1,33 @@
...
+}
+
+module grantIngest '../../modules/logs/kusto/grant-access.bicep' = {
+  name: 'grantKusto-${databaseName}-${uniqueString(managedIdentityName, databaseName)}'
+  scope: resourceGroup(kustoResourceGroup)
+  params: {
+    ingestAccessPrincipalIds: [collectorIdentity.properties.principalId]
+    databaseName: databaseName
+    kustoName: kustoName
+  }
```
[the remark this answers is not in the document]
galchammat: Good catch — this was a real blocker, fixed in 6552a7b. `Ingestor` cannot query, so `watermarks()` would have failed on the first pass and nothing would ever have ingested. It slipped through because I seeded the table with my own `AllDatabasesAdmin` credentials, where the read always succeeds; the commit message even claimed Ingestor "can add rows but neither read them back", which contradicted the code.

The grant now covers both `Ingestor` and `Viewer`. Keeping the watermark in the table itself is deliberate: it means the resume point survives a restart, whereas holding it in memory would re-ingest the seed window every time the collector restarts.

Worth noting a second-order problem this surfaced — `modules/logs/kusto/grant-access.bicep` names every assignment `grant-\${guid(id, databaseName)}` regardless of role, so passing one principal as both an ingestor and a reader would have produced two resources with the same name and failed at deploy. The assignments are therefore declared in `opstool-kusto-grant-assignments.bicep` with role-distinct names rather than through that module; changing the module's naming would have renamed — and so recreated — the existing arobit and admin-api assignments.

tooling/tenant-quota/pkg/config/config.go
```diff
@@ -42,11 +48,78 @@ type Config struct {
...
+	} {
+		if strings.TrimSpace(value) == "" {
+			return fmt.Errorf("%s is required", name)
+		}
+	}
+
+	parsedURL, err := url.Parse(strings.TrimSpace(c.SippyURI))
+	if err != nil || parsedURL.Host == "" || (parsedURL.Scheme != "http" && parsedURL.Scheme != "https") {
+		return fmt.Errorf("ciJobOutcomes.sippyURI must be a valid HTTP or HTTPS URL")
+	}
```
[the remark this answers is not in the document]
galchammat: Fixed in 6552a7b. Both are now validated as HTTP/HTTPS at config load via a shared `validateHTTPURL`, which `sippyURI` also uses so the three endpoints are treated identically. Added `TestCIJobOutcomesConfigValidatesEndpoints` covering exactly the missing-scheme case you describe, plus empty values and the disabled-config path.

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
```diff
@@ -0,0 +1,57 @@
...
+//     | extend clusterToken = extract(@'(j[0-9a-z]{7})', 1, cluster)
+//     | where isnotempty(clusterToken)
+//     | join kind=inner ciJobOutcomes on clusterToken
+//
+// attributes a firing to a run. The token is stored rather than the cluster
+// names because one run creates several clusters (-svc, -mgmt-1, -mgmt-2) and
+// their number varies per run, so reconstructing names would drop rows.
+.create-merge table ciJobOutcomes (
+    buildId: string,
+    clusterToken: string,
```
stevekuznetsov: what is this?

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
```diff
@@ -0,0 +1,57 @@
...
+//     database('MonitoringEvents').alertEvents
+//     | extend cluster = tostring(alertContext.labels.cluster)
+//     | extend clusterToken = extract(@'(j[0-9a-z]{7})', 1, cluster)
+//     | where isnotempty(clusterToken)
+//     | join kind=inner ciJobOutcomes on clusterToken
+//
+// attributes a firing to a run. The token is stored rather than the cluster
+// names because one run creates several clusters (-svc, -mgmt-1, -mgmt-2) and
+// their number varies per run, so reconstructing names would drop rows.
+.create-merge table ciJobOutcomes (
```
stevekuznetsov, line 30: you likely want to record the mgmt & service clusters that were part of this test, so it's easy to cross-reference

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
```diff
@@ -0,0 +1,57 @@
...
+    buildId: string,
+    clusterToken: string,
+    jobName: string,
+    prowUrl: string,
+    sippyRelease: string,
+    family: string,
+    overallResult: string,
+    failed: bool,
+    testFailures: int,
+    failedTests: dynamic,
```
stevekuznetsov: you likely want a table for test names (and some stable id) and a join table for (these tests failed on this run) instead of dynamic. use a real schema so queries are fast

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
```diff
@@ -0,0 +1,57 @@
...
+    startedAt: datetime
```
stevekuznetsov: finishedAt?

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
stevekuznetsov: per-test rg/started/test-done/cleanup-started/all-finished would be good too

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
```diff
@@ -0,0 +1,57 @@
...
+// It exists to make the rest of the telemetry joinable to CI. Container logs,
+// Kubernetes events and alerts all record the cluster a run used, but nothing
+// records whether the run passed, which job it was, or which pull request it
+// tested - so questions like "which alerts fire on runs whose e2e tests passed"
+// cannot be asked. clusterToken bridges the two: it is the "j<7 chars>" segment
+// that appears in every CI cluster name, so
+//
+//     database('MonitoringEvents').alertEvents
+//     | extend cluster = tostring(alertContext.labels.cluster)
+//     | extend clusterToken = extract(@'(j[0-9a-z]{7})', 1, cluster)
```
stevekuznetsov: avoid any patterns where you have to do parsing, extracting, etc, as part of the query - just store the service and mgmt cluster names

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
stevekuznetsov: sorry, I did not RTFM

galchammat: Gone — replaced with `svcCluster` and `mgmtCluster`, read from the run's own `config.yaml` artifact rather than derived, so nothing has to be reconstructed.

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
galchammat: Done — both stored per run from the run's `config.yaml`; left empty for promoted environments, whose telemetry lives in their own Kusto anyway.

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
galchammat: Done — `ciTestNames` (name + hashed id) and `ciTestResults`; recording every executed test rather than only failures, so the per-test timings below aren't limited to the 0-3 failing rows.

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
galchammat: Added, from the run's `finished.json` — Sippy doesn't report it.

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
galchammat: All added, except test-done and cleanup-started are one column: they're the same instant by construction (`CleanupStartTime` is set to when the test body finished).

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
galchammat: Agreed — `extract()` is gone; consumers now join on the stored cluster names directly.

dev-infrastructure/modules/logs/kusto/tables/ciJobOutcomes.kql
galchammat: No worries — it's gone regardless.

dev-infrastructure/dev-ci/cluster/opstool-kusto-grant.bicep
```diff
@@ -0,0 +1,38 @@
...
+param databaseName string
+
+// The collector running on this cluster records CI job outcomes in the dev
+// Kusto, so that the alerts and logs already held there can be attributed to
+// the run that produced them.
+//
+// It needs both roles. Ingestor lets it write rows, and Viewer lets it read
+// back the newest row it already wrote, which is how each pass knows where to
+// resume. Ingestor alone cannot read, so the collector would fail on its first
+// query and never ingest anything.
```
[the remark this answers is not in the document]
galchammat: Correct — that rationale outlived the watermark it described. Reworded to say Viewer is needed to read back which runs and test names are already stored, which is what each pass subtracts.

tooling/tenant-quota/pkg/cijoboutcomes/writer.go
```diff
@@ -0,0 +1,472 @@
...
+	logger *slog.Logger
+	client *http.Client
+
+	// queued holds runs written but not yet visible to a query, keyed by build
+	// id and held with their start time so they can be dropped once the query
+	// covers them. artifactAttempts counts failed artifact reads per run, so a
+	// transient failure is retried rather than stored as a run with no tests.
+	// Both are guarded by queuedMu and are optimisations rather than
+	// correctness guarantees: ingestion is idempotent, so losing either to a
+	// restart costs repeated work and not duplicate rows.
```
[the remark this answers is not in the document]
galchammat: Right, the claim was too broad. Scoped it to a run's outcome and test rows, which are tagged with `IfNotExists`, and called out test names as the exception — they belong to no run, carry no tag, and can be written twice.

galchammat:
Both suppressed findings on `writer.go` were right and are fixed in `afdd815de`.

**Unbounded pass** — `writeOnce` ran on the loop's own context, so a hung Sippy, Kusto or GCS call would have stalled collection entirely instead of failing one tick. Now wrapped in `context.WithTimeout`, as `pkg/prow` and `pkg/subscriptionquota` already do. It takes its own timeout rather than the collector-wide 30s, which is sized for a single API call — a pass reads one set of artifacts per unseen run, so 30s would cut every pass short. Default is 10m, and cutting a pass short stays safe because stored runs are subtracted from the next one. A timed-out pass also no longer charges the run it was reading for the failure, which would otherwise have let three slow passes conclude a run had no tests.

**Unconstrained credential** — both clients used `WithDefaultAzureCredential`, which can fall back to a developer sign-in: fine on a laptop, broken in-cluster where this runs as a workload identity. Now one `DefaultAzureCredential` with `RequireAzureTokenCredentials`, matching `tooling/hcpctl/pkg/kusto`, built once per pass instead of twice.

On the **`hcpctl` dependency weight** — taking the cost knowingly. `cloud.google.com/go/storage`, `google.golang.org/api` and `openshift-tests-extension` are what read and parse the artifacts, so any implementation pulls them; only `hcpctl/pkg/kusto` and `azcore/arm` are true extras. Extracting a lightweight package is the better end state, but a copy would have to re-derive `ParseProwURL`, `findArtifactDir`, `fetchTestTimings` and `ExtractResourceGroup` — the parts most likely to drift when the artifact layout changes. `tooling/aro-hcp-exporter` already depends on `hcpctl` the same way.

raelga (approved):
/lgtm
/approve

galchammat:
@redhat-chai-bot e2e has been retried a handful of times. Can you determine whether the failures were flakes or a regression? https://cihealth.tools.hcpsvc.osadev.cloud/report?end_date=2026-09-10&mode=rolling&start_date=2026-09-04

PR #6685, opened 2026-08-25
fix: enable kubelet reservations on CI management clusters

geoberle:
## Summary
- Remove `applyKubeletFixes: false` overrides from CI environments (ci00, ci01) in `config/config.yaml`
- CI management clusters now inherit the default `true`, enabling kubelet resource reservations

## Why
CI management clusters have been hitting node NotReady conditions during heavy pod loads. Kubelet resource reservations prevent the kubelet from being starved of resources, keeping nodes healthy under pressure.

## Test plan
- [x] `config/rendered/` files updated via `make materialize` — ci00 and ci01 now show `applyKubeletFixes: true`
- [ ] CI pipelines apply the change after merge
- [ ] Monitor CI MC nodes for NotReady events

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1804,7 +1804,6 @@ clouds:
                 osDiskSizeGB: 64
             jaeger:
               deploy: false
-            applyKubeletFixes: false
           # SVC
           svc:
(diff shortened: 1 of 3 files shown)


raelga (approved):
/lgtm
/approve

PR #6687, opened 2026-08-26
Improve ARM error message when install deadline is exceeded

JakobGray:
https://issues.redhat.com/browse/ARO-28984

### What

When cluster/node pool create hits the RP cluster provisioning deadline and the cluster is still installing, ARM returns an error that gives the impression of a CS InternalServerError.

```
ERROR CODE: InternalServerError
"message": "[clusterServiceClusterStatus] <no_message>"
```

This changes keeps the create deadline sentence as the ARM error cause, and give in-progress Cluster Service create checks a real OperationState message so PickWorst does not fill in `<no_message>`.

| | Message |
|---|---|
| Before | `[clusterServiceClusterStatus] <no_message>` |
| After | `cluster creation did not complete before the deadline; [clusterServiceClusterStatus] cluster service is installing` |

### Why

ConvertClusterStatus maps to Provisioning with an empty message. PickWorstOperationState then substitutes <no_message>. The deadline path replaced the timeout sentence whenever PickWorst had any message, so the only text customers and e2e saw was [clusterServiceClusterStatus] <no_message> with InternalServerError. That hid the actual cause (create deadline) and blamed Cluster Service for cases where CS was healthy and still installing.

### Testing

Unit tests for new functions

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/backend/pkg/controllers/cluster/operations/operation_cluster_create.go
+++ b/backend/pkg/controllers/cluster/operations/operation_cluster_create.go
@@ -176,10 +176,10 @@ func (c *operationClusterCreate) SynchronizeOperation(ctx context.Context, key c
 		cluster.ServiceProviderProperties.CreateOperationCompletionDeadline != nil &&
 		c.clock.Now().After(cluster.ServiceProviderProperties.CreateOperationCompletionDeadline.Time) {
 
-		message := "cluster creation did not complete before the deadline"
-		if len(operationalState.Message) > 0 {
-			message = operationalState.Message
-		}
+		message := operationbase.DeadlineExceededMessage(
+			"cluster creation did not complete before the deadline",
+			operationalState.Message,
+		)
 		logger.Info("create operation deadline exceeded, marking as failed",
 			"deadline", cluster.ServiceProviderProperties.CreateOperationCompletionDeadline.Time,
 			"message", message)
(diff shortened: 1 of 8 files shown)


chetan-giradkar:
lgtm.
Are these the only places where we need to change how the ARM error is interpreted and passed along? 
Can be a followup to review in case it needs to be reviewed.

backend/pkg/controllers/cluster/operations/operation_cluster_create.go
mbarnes, line 188: We should define a specific error code for this:
```
coreapi.CloudErrorCodeDeadlineExceeded
```

`InternalServerError` was meant as a generic catch-all.

backend/pkg/controllers/cluster/operations/operation_cluster_create_test.go
```diff
@@ -240,6 +242,7 @@ func TestOperationClusterCreate_SynchronizeOperation(t *testing.T) {
 				assert.Equal(t, coreapi.ProvisioningStateFailed, op.Status)
 				require.NotNil(t, op.Error)
 				assert.Equal(t, coreapi.CloudErrorCodeInternalServerError, op.Error.Code)
+				assert.Contains(t, op.Error.Message, "cluster creation did not complete before the deadline")
```
mbarnes, line 245: Per my comment above, this is why we want a unique error code.  Checking the code alone should be sufficient, you wouldn't _need_ to compare the error message too.

mbarnes (changes requested):

JakobGray:
Only failure was around frontend latency stats, worth rerunning
/test e2e-parallel

backend/pkg/controllers/cluster/operations/operation_cluster_create.go
JakobGray: Done. Now sets CloudErrorCodeDeadlineExceeded. InternalServerError still applies for other unclassified failures

backend/pkg/controllers/cluster/operations/operation_cluster_create_test.go
JakobGray: Also done, now asserting CloudErrorCodeDeadlineExceeded

deads2k:
/hold

just blind from https://redhat-external.slack.com/archives/C075PHEFZKQ/p1790281147514139  if it's mistaken, my apologies, but I wanted batches to run overnight.

PR #6218, opened 2026-07-22
fix: include resource group name in orphaned MRG alerts

ziccardi:
Add subscription_id and resource_group labels to orphaned managed resource group metrics and alerts to make it clear which specific resource groups need investigation.

Changes:
- Add subscription_id and resource_group labels to both metrics:
  - aro_hcp_orphaned_managed_resource_groups_found_total
  - aro_hcp_orphaned_managed_resource_groups_deletion_failed_total
- Update OrphanedMRGDetected alert to show specific resource group and subscription in description
- Update OrphanedMRGDeletionFailing alert similarly
- Normalize label values to lowercase for consistency with existing backend metrics pattern
- Update alert tests to include new labels

Before: "Found X orphaned MRGs in location eastus" After: "Found orphaned MRG managed-rg-orphaned in subscription
       00000000-0000-0000-0000-000000000000, location eastus"

### Testing

Testing is required for feature completion and tests should be part of the pull
request along with the feature changes.

Describe the testing provided. If you did not add tests, provide a clear
justification.

- Unit tests
- [Integration tests](https://github.com/Azure/ARO-HCP/tree/main/test-integration)
- [E2E tests](https://github.com/Azure/ARO-HCP/blob/main/test/e2e/README.md)

### PR Checklist

- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
- [ ] Screenshots included (if graph/UI/metrics changes)
- [ ] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [ ] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

If E2E tests are included:

- [ ] E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md)
- [ ] If new E2E use case is covered (via a new test or new check/verifier),
  demonstrate that the test is able to detect a defect/error and fail with
  proper error message and logs which communicates nature of the problem.

PR #6688, opened 2026-08-26
feat: add orphaned managed resource group cleanup controller

ziccardi:
[ARO-27732](https://redhat.atlassian.net/browse/ARO-27732)

### What

Refactor the orphaned managed resource group (MRG) cleanup into a
  ManagedResourceGroupWatchingController + processor architecture, wire it
  through the config system, and add per-resource Prometheus alerting.

  Key changes:

  - **ManagedResourceGroupWatchingController** — subscription-informer-driven
    watcher that discovers Azure managed resource groups and enqueues them
    for processing (24-hour resync cadence).
  - **OrphanedManagedResourceGroupController** — processor that checks
    cluster existence (lister first, Cosmos fallback) and deletes orphaned
    MRGs when running in write mode.
  - **AFEC-based subscription ownership** — each environment only processes
    subscriptions it owns:
    - INT/STG: `--orphaned-mrg-cleanup-target-subscription-afec-flags`
      restricts cleanup to subscriptions with the matching AFEC.
    - PROD: `--orphaned-mrg-cleanup-excluded-subscription-afec-flags`
      excludes subscriptions owned by INT/STG.
    - DEV: both empty → controller disabled.
  - **Fail-closed** — nil `Properties` or `RegisteredFeatures` treats the
    subscription as not-owned, preventing accidental deletions during
    incomplete synchronization.
  - **Config system wiring** — schema, config.yaml defaults, values.yaml,
    and Helm deployment template for all three flags
    (`target-subscription-afec-flags`, `excluded-subscription-afec-flags`,
    `running-mode`).
  - **Detection-only rollout** — `orphanedMRGCleanupRunningMode` defaults
    to `"read"`. Only `"write"` enables actual deletion; any other value
    (including empty) defaults to read-only.
  - **Per-resource Prometheus labels** — `subscription_id` and
    `managed_resource_group_name` on `aro_hcp_orphaned_managed_resource_groups_found_total`
    and `aro_hcp_orphaned_managed_resource_groups_deletion_failed_total`,
    with `correlationId` annotations for distinct incident tracking.
  - **Cosmos data flow docs** — added controller entry documenting
    subscription and cluster reads.

### Why

  
  The controller runs in multiple environments (INT, STG, PROD), where a
  managed resource group belonging to one environment may appear orphaned
  from another environment's perspective. Without explicit ownership
  filtering, one environment could incorrectly delete resources belonging
  to another. Environment-aware AFEC gating ensures each controller
  instance only cleans up its own subscriptions.

  The detection-only rollout allows validating the controller's behavior
  in production before enabling actual deletions.

### Testing

justification.

- Unit tests

  - MSFT environment AFEC values (INT/STG/PROD) must be configured in the
    sdp-pipelines repo.
  - `orphanedMRGCleanupRunningMode` is intentionally `"read"` in all
    checked-in configs. Write mode will be enabled per-environment after
    detection is validated.

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


ziccardi:
@bennerv Mind a look please?

backend/pkg/controllers/cluster/deletion/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -46,6 +46,12 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
+const (
+	// AFEC feature flags that indicate a subscription should skip orphaned MRG deletion
```
stevekuznetsov, line 50: Why?

backend/pkg/controllers/cluster/deletion/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -46,6 +46,12 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
+const (
+	// AFEC feature flags for environment identification
+	afecFeatureINTApproved     = "Microsoft.RedHatOpenShift/INT-APPROVED"
```
bennerv: should we plumb these as config?  Right now, how are they getting set? Also use CEL rules to ensure that only one or the other can be set if you plumb config.

It may also be useful to validate these AFECs actually exist as registered features during CI check-in time.  That will prevent accidental spelling mistakes or otherwise.

backend/pkg/controllers/cluster/deletion/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -69,22 +75,51 @@ type cleanOrphanedClusterManagedResourceGroup struct {
 	cooldownChecker       controllerutil.CooldownChecker
 	resourcesDBClient     corecosmosstorage.ResourcesDBClient
 	azureFPAClientBuilder azureclient.FirstPartyApplicationClientBuilder
+	// myAFEC is the AFEC flag that identifies subscriptions owned by this environment (e.g., "Microsoft.RedHatOpenShift/STAGING-APPROVED")
+	// Only MRGs in subscriptions with this AFEC will be considered for cleanup. Empty in PROD.
+	myAFEC string
+	// otherAFECs is a set of AFEC flags that identify subscriptions owned by OTHER environments.
+	// MRGs in subscriptions with any of these AFECs will be skipped. Empty in INT/STAGING.
+	otherAFECs map[string]struct{}
```
bennerv: can/should we union these types.  They're mutually exclusive.  Naming things is hard, someone else probably has better descriptive names than below. 

```
type AFECCleanupType string
const (
  MySubscriptionAFECType = "MyAFEC"
  NotMySubscriptionAFECType = "NotMyAFECs"
  NoSubscriptionAFECType= "None"
) 

type afecCleanupConfig struct {
    Type  AFECCleanupType
    MySubscriptionAFEC string
    NotMySubscriptionAFEC []string
}
```

backend/pkg/controllers/cluster/deletion/clean_orphaned_cluster_managed_resource_group_controller.go
```diff
@@ -69,22 +75,51 @@ type cleanOrphanedClusterManagedResourceGroup struct {
...
 // that are not referenced by any HCPOpenShiftCluster in the database and cleans them up.
+//
+// Environment-based cleanup is controlled by these environment variables:
+// - MY_AFEC: Single AFEC flag identifying subscriptions owned by this environment (e.g., "Microsoft.RedHatOpenShift/STAGING-APPROVED")
+// - OTHER_AFECS: Comma-separated list of AFEC flags identifying subscriptions owned by OTHER environments
+//
+// PROD: Set OTHER_AFECS to skip INT/STAGING subscriptions
+// STAGING/INT: Set MY_AFEC to only clean subscriptions with that AFEC
+// DEV: Leave both empty to disable the controller
 func NewCleanOrphanedClusterManagedResourceGroupController(
```
bennerv: what do you think about doing something like this.  

- Implement a new Watching controller that's based on the Subscription watch but has managed resource groups as it's output. 
- Implement the OrphanedResourceGroupController as a waatching controller of the managed resource group.
- plumb the threadiness as a controller argument of 20
- have a needsWork function on the orphaned managed resource group controller that checks the subscription exists in the subscirption document and the AFECs are correctly configured

bennerv:
Left comments but summarize: 
- Instead of having an equivalent "threadiness" value in the controller, just create a new watching controller on managed resource groups and do work off of that in the orphaned resource group controller
- plumb the AFECs as `config` and set env vars in the deployment for usage.  Looks like this is missing right now in this PR
    - use CEL rules to make the argument mutually exclusive
- Refactor the AFEC types to be a union argument so one or the other gets set.

backend/pkg/controllers/cluster/deletion/managed_resource_group_watching_controller.go
```diff
@@ -0,0 +1,243 @@
+// Copyright 2026 Microsoft Corporation
```
bennerv, line 1: watching controllers typically live in `pkg/utils/controllerutils`

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,338 @@
...
+		for _, afec := range strings.Split(otherAFECsStr, ",") {
+			trimmed := strings.TrimSpace(afec)
+			if trimmed != "" {
+				otherAFECs[trimmed] = struct{}{}
+			}
+		}
+	}
+
+	// Determine mode
+	if myAFEC == "" && len(otherAFECs) == 0 {
```
bennerv, line 134: defense in depth, error out if we have both set.

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,338 @@
...
+// It returns true if the MRG is orphaned (cluster doesn't exist in Cosmos).
+func (c *orphanedManagedResourceGroupController) needsWork(ctx context.Context, key ManagedResourceGroupKey) (bool, *coreapi.Subscription, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	// Load subscription from Cosmos - this provides race protection.
+	// If subscription doesn't exist in Cosmos yet, we skip processing to avoid
+	// deleting MRGs before cluster data has been fully synced.
+	subscription, err := c.resourcesDBClient.Subscriptions().Get(ctx, key.SubscriptionID)
+	if cosmosstorageutils.IsNotFoundError(err) {
+		logger.Info("Subscription not found in Cosmos, skipping MRG (race protection)")
```
bennerv, line 183: technically since our managed resource group controller feeds off of a managed resource group watcher which feeds off of the subscription database, this should never happen but it's good protection and not bad to keep.

bennerv (commented):
This approach and refactor looks good to me.  Still missing plumbing config and the backend deployment. 

Good to keep the readonly mode on for first iteration and ensure that we audit everything before we start deleting just for safety.

backend/pkg/controllers/cluster/deletion/clean_orphaned_cluster_managed_resource_group_controller.go
ziccardi: Done

backend/pkg/utils/controllerutils/managed_resource_group_watching_controller.go
```diff
@@ -0,0 +1,243 @@
+// Copyright 2026 Microsoft Corporation
```
ziccardi: Moved

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
ziccardi: Done!

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,344 @@
...
+	logger.V(1).Info("Cluster exists - MRG is not orphaned")
+	return false, subscription, nil
+}
+
+// ProcessManagedResourceGroup implements ManagedResourceGroupProcessor.
+// It checks if the MRG is orphaned and deletes it if necessary.
+func (c *orphanedManagedResourceGroupController) ProcessManagedResourceGroup(ctx context.Context, key controllerutils.ManagedResourceGroupKey) error {
+	logger := utils.LoggerFromContext(ctx)
+
+	readOnly := c.readOnly
```
machi1990: hmm do we need to redeclare the variable?

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,344 @@
...
+
+// needsWork determines if a managed resource group should be deleted.
+// It returns true if the MRG is orphaned (cluster doesn't exist in Cosmos).
+func (c *orphanedManagedResourceGroupController) needsWork(ctx context.Context, key controllerutils.ManagedResourceGroupKey) (bool, *coreapi.Subscription, error) {
+	logger := utils.LoggerFromContext(ctx)
+
+	// Load subscription from Cosmos - this provides race protection.
+	// If subscription doesn't exist in Cosmos yet, we skip processing to avoid
+	// deleting MRGs before cluster data has been fully synced.
+	subscription, err := c.resourcesDBClient.Subscriptions().Get(ctx, key.SubscriptionID)
```
machi1990: We can get this from the lister

machi1990: No need to do direct read

backend/pkg/app/backend.go
```diff
@@ -1202,7 +1214,7 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
 				go createNodePoolScopedReadDesiresController.Run(ctx, 20)
 				go createServiceProviderClusterController.Run(ctx, 20)
 				go createServiceProviderNodePoolController.Run(ctx, 20)
-				go cleanOrphanedClusterManagedResourceGroupController.Run(ctx, 20)
+				go managedResourceGroupWatchingController.Run(ctx, 20)
```
machi1990: is 20 enough? :-)

backend/pkg/utils/controllerutils/managed_resource_group_watching_controller.go
```diff
@@ -0,0 +1,250 @@
...
+		"location", k.Location,
+	)
+}
+
+// ManagedResourceGroupWatchingControllerName is the name of the controller that
+// discovers managed resource groups in Azure by watching subscription changes.
+const ManagedResourceGroupWatchingControllerName = "ManagedResourceGroupWatching"
+
+// ManagedResourceGroupProcessor processes discovered managed resource groups.
+type ManagedResourceGroupProcessor interface {
```
machi1990: This is more like a https://github.com/ziccardi/ARO-HCP/blob/27d8fc2650cf7531f5085892a630a2c6538ecf90/internal/controllerutils/generic_watching_controller.go#L39 

Can we rename it as such i.e ManagedResourceGoupSyncer

and it having a SyncOnce method

backend/pkg/utils/controllerutils/managed_resource_group_watching_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	ProcessManagedResourceGroup(ctx context.Context, key ManagedResourceGroupKey) error
+}
+
+type managedResourceGroupWatchingController struct {
+	name                  string
+	location              string
+	processor             ManagedResourceGroupProcessor
```
machi1990: syncer ManagaedResourceGroupSyncer

backend/pkg/utils/controllerutils/managed_resource_group_watching_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	resourcesDBClient     corecosmosstorage.ResourcesDBClient
```
machi1990: I think we only need the subscription lister and no need to perform a live read from cosmos

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	// Parse the managedBy resource ID to extract cluster details
+	managedByID, err := azcorearm.ParseResourceID(key.ManagedBy)
+	if err != nil {
+		logger.Error(err, "Failed to parse managedBy resource ID")
+		return false, nil, nil // Invalid managedBy, skip
+	}
+
+	// Check if cluster exists in Cosmos
```
machi1990: Better to first to a simple cache check via the Lister. If we've a hit we know for sure the cluster exists and we avoid hitting cosmos. If the cache is cold, we double verify that the cluster doesn't exist via cosmo call

backend/pkg/app/backend.go
```diff
@@ -808,12 +811,21 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
...
+		b.options.OrphanedMRGOtherAFECs,
+		b.options.OrphanedMRGReadWrite,
+	)
+	managedResourceGroupWatchingController := controllerutils.NewManagedResourceGroupWatchingController(
+		b.options.AzureLocation,
+		orphanedManagedResourceGroupController,
 		b.options.ResourcesDBClient,
 		b.options.FPAClientBuilder,
 		backendInformers,
+		10*time.Minute,
```
machi1990: Do we need to resync every 10 minutes? I bet running this once a day is enough

backend/cmd/root.go
```diff
@@ -606,6 +616,9 @@ func NewBackendRootCmdFlags() *BackendRootCmdFlags {
 		ExitOnPanic:                                     true,
 		BackupScheduleCadence:                           string(backups.BackupCadenceProduction),
 		BackupScheduleState:                             string(coreapi.BackupScheduleStateEnabled),
+		OrphanedMRGMyAFEC:                               os.Getenv("MY_AFEC"),
+		OrphanedMRGOtherAFECs:                           os.Getenv("OTHER_AFECS"),
+		OrphanedMRGReadWrite:                            os.Getenv("CLEAN_ORPHANED_MANAGED_RESOURCE_GROUPS_MODE") == "readwrite",
```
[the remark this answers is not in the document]
machi1990: @ziccardi Yeap, we need to declare these in the deploy template and in the `config/` directory.

Also, is there a reason why you preferred env var over CLI flags just like other configs in the backend?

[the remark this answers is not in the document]
ziccardi: No reasons honestly. I already decided to switch to flags, but probably I lost that change. Doing now. Thanks!

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,359 @@
...
+		logger.Error(err, "Failed to get resource group state")
+		orphanedMRGsDeletionFailed.WithLabelValues(c.location).Inc()
+		return err
+	}
+
+	// Check if already being deleted
+	if rg.Properties != nil && rg.Properties.ProvisioningState != nil {
+		provisioningState := *rg.Properties.ProvisioningState
+		if provisioningState == "Deleting" {
+			logger.Info("Orphaned cluster managed resource group deletion already in progress",
```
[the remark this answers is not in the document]
machi1990: The managedby once set is immutable

backend/pkg/app/backend.go
```diff
@@ -808,12 +811,21 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
...
+		b.options.OrphanedMRGMyAFEC,
+		b.options.OrphanedMRGOtherAFECs,
+		b.options.OrphanedMRGReadWrite,
+	)
+	managedResourceGroupWatchingController := controllerutils.NewManagedResourceGroupWatchingController(
+		b.options.AzureLocation,
+		orphanedManagedResourceGroupController,
 		b.options.FPAClientBuilder,
 		backendInformers,
+		24*time.Hour,
```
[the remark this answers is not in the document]
machi1990: When a cluster is being deleted, the MRG should be deleted. The daily cadence is to catch and fix orphaned MRGs due to the race  in CS that was recreating the MRG after its deletion. The race has long been fixed and this cleanup is so that we fix those and have a cadence that runs at once to fix any future issues

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,359 @@
...
+	)
+
+	orphanedMRGsDeletionFailed = promauto.With(legacyregistry.Registerer()).NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "aro_hcp_orphaned_managed_resource_groups_deletion_failed_total",
+			Help: "Total number of orphaned cluster managed resource groups where deletion failed",
+		},
+		[]string{"location"},
+	)
+)
```
machi1990, line 60: We probably want to include the MRG as well as a label see https://github.com/Azure/ARO-HCP/pull/6218 ?

And in consequence modify the alert to also include the MRG same as https://github.com/Azure/ARO-HCP/pull/6218

ziccardi: Done

config/config.msft.clouds-overlay.yaml
```diff
@@ -134,6 +140,9 @@ clouds:
       prod:
         # this is the MSFT PRODUCTION environment
         defaults:
+          # Backend
+          backend:
+            orphanedMRGOtherAFECs: "Microsoft.RedHatOpenShift/INT-APPROVED,Microsoft.RedHatOpenShift/STAGING-APPROVED"
```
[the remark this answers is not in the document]
ziccardi: This is intentional

config/config.yaml
```diff
@@ -837,6 +837,9 @@ defaults:
     backupScheduleState: Enabled
     exitOnPanic: true
     logVerbosity: 4
+    orphanedMRGMyAFEC: ""
+    orphanedMRGOtherAFECs: ""
+    orphanedMRGReadWrite: false
```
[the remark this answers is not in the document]
ziccardi: This is intentional. WE will first rollout readonly to ensure the controller works as expected.

config/config.msft.clouds-overlay.yaml
machi1990, line 1: The change is this file needs to be done via sd-pipeline repo and not this overlay file. The sd-pipeline repo is our source of truth for anything higher env. 
I think @Copilot flagged it without having the needed context.
@ziccardi

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil {
+		logger.Error(err, "Failed to parse managedBy resource ID")
+		return false, nil, nil // Invalid managedBy, skip
+	}
+
+	// Check if cluster exists in Cosmos
+	_, err = c.resourcesDBClient.HCPClusters(
+		managedByID.SubscriptionID,
+		managedByID.ResourceGroupName,
+	).Get(ctx, managedByID.Name)
```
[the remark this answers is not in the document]
machi1990: This makes sense

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,366 @@
...
+		afecOwnership:         afecOwnership,
+	}
+}
+
+// parseAFECOwnership constructs AFECOwnershipConfig from the provided AFEC configuration.
+// The caller must ensure that myAFEC and otherAFECsStr are not both set (mutually exclusive).
+func parseAFECOwnership(myAFEC, otherAFECsStr string) AFECOwnershipConfig {
+	// Parse otherAFECs
+	otherAFECs := make(map[string]struct{})
+	if otherAFECsStr != "" {
```
machi1990: len(otherAFECsStr) != 0

Use len(s) throughout whenever checking for emptiness.

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,366 @@
...
+// AFECOwnershipConfig configures environment-based subscription ownership for orphaned MRG cleanup.
+type AFECOwnershipConfig struct {
+	Mode AFECOwnershipMode
+	// MyAFEC is the AFEC flag that identifies subscriptions owned by this environment.
+	// Used when Mode is AFECOwnershipMyAFEC.
+	MyAFEC string
+	// NotOtherAFECs is a set of AFEC flags that identify subscriptions owned by OTHER environments.
+	// Used when Mode is AFECOwnershipNotOtherAFECs.
+	NotOtherAFECs map[string]struct{}
+}
```
machi1990, line 83: The struct and its fields can also be named to take the naming pattern of the CLI flags

e.g instead of MyAFEC ==> `TargetFooBar` (an example)
instead of NotOtherAFects ==> `ExcludedFooBar`

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	AFECOwnershipNone AFECOwnershipMode = "None"
+	// AFECOwnershipMyAFEC requires subscriptions to have a specific AFEC flag (INT/STAGING environment).
+	AFECOwnershipMyAFEC AFECOwnershipMode = "MyAFEC"
+	// AFECOwnershipNotOtherAFECs requires subscriptions to NOT have any of the specified AFEC flags (PROD environment).
+	AFECOwnershipNotOtherAFECs AFECOwnershipMode = "NotOtherAFECs"
+)
+
+// AFECOwnershipConfig configures environment-based subscription ownership for orphaned MRG cleanup.
+type AFECOwnershipConfig struct {
+	Mode AFECOwnershipMode
```
machi1990, line 76: is this redundant of `readOnly:              runningMode != "write",`?

machi1990 (commented):
I've left a couple small comments.
otherwise lgtm to me once those are addressed.
On top of those, I'd like @miguelsorianod and/or @geoberle to also have a look at this

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
ziccardi: No, they serve different purposes:                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                          
  - Mode (line 174) controls which subscriptions the controller processes: None disables the controller entirely, MyAFEC restricts to subscriptions with a specific AFEC, NotOtherAFECs excludes subscriptions belonging to other environments.                                           
  - readOnly (line 317) controls whether the controller deletes orphaned MRGs it finds, or just logs them.

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
```diff
@@ -0,0 +1,366 @@
...
+		for _, afec := range strings.Split(excludedAFECsStr, ",") {
+			trimmed := strings.TrimSpace(afec)
+			if len(trimmed) != 0 {
+				excludedAFECs[trimmed] = struct{}{}
+			}
+		}
+	}
+
+	// Determine mode
+	if targetAFECs == "" && len(excludedAFECs) == 0 {
```
machi1990:
```suggestion
	if len(targetAFECs) == 0 && len(excludedAFECs) == 0 {
```

Please verify all the other instances as well when checking for emptiness of a string and use `len(s)`

backend/cmd/root.go
```diff
@@ -235,6 +238,13 @@ func (f *BackendRootCmdFlags) AddFlags(cmd *cobra.Command) {
...
 	cmd.Flags().StringVar(&f.BackupScheduleState, "backup-schedule-state", f.BackupScheduleState,
 		fmt.Sprintf("Backup schedule state. Accepted values: %s, %s", coreapi.BackupScheduleStateEnabled, coreapi.BackupScheduleStateDisabled))
+	cmd.Flags().StringVar(&f.OrphanedMRGCleanupTargetSubscriptionAFECFlags, "orphaned-mrg-cleanup-target-subscription-afec-flags", f.OrphanedMRGCleanupTargetSubscriptionAFECFlags,
+		"AFEC flag identifying subscriptions owned by this environment for orphaned MRG cleanup (e.g. Microsoft.RedHatOpenShift/STAGING-APPROVED)")
+	cmd.Flags().StringVar(&f.OrphanedMRGCleanupExcludedSubscriptionAFECFlags, "orphaned-mrg-cleanup-excluded-subscription-afec-flags", f.OrphanedMRGCleanupExcludedSubscriptionAFECFlags,
+		"Comma-separated AFEC flags identifying subscriptions owned by OTHER environments for orphaned MRG cleanup")
+	cmd.Flags().StringVar(&f.OrphanedMRGCleanupRunningMode, "orphaned-mrg-cleanup-running-mode", f.OrphanedMRGCleanupRunningMode,
+		"Running mode for orphaned MRG cleanup: 'write' to actually delete, any other value defaults to read-only.")
```
machi1990: As its about delete, I'll have two modes:
- `delete`
- `readonly` or `dry_run` mode and validate that its set to either or.

config/config.schema.json
```diff
@@ -1476,6 +1476,18 @@
...
+          "description": "AFEC flag identifying subscriptions owned by this environment for orphaned MRG cleanup"
+        },
+        "orphanedMRGCleanupExcludedSubscriptionAFECFlags": {
+          "type": "string",
+          "description": "Comma-separated AFEC flags identifying subscriptions owned by OTHER environments for orphaned MRG cleanup"
+        },
+        "orphanedMRGCleanupRunningMode": {
+          "type": "string",
+          "description": "Running mode for orphaned MRG cleanup: 'write' to actually delete orphaned managed resource groups, any other value (including empty) defaults to read-only mode."
+        },
```
machi1990, line 1491: can be a enum type like https://github.com/ziccardi/ARO-HCP/blob/8a79a12a7517643489de8765e75a2ba51e3b6996/config/config.schema.json#L830

backend/pkg/controllers/cluster/deletion/orphaned_managed_resource_group_controller.go
ziccardi: Sorry, I missed that one. Fixed.

machi1990 (approved):
/lgtm 

@geoberle @miguelsorianod can you also have a look please?

@ziccardi once this is approved and merged, we'll need to followup with the sd-pipeline changes to set the config

PR #6689, opened 2026-08-26
fix(e2e): work around OCP 4.22 DNS NetworkPolicy blocking Cilium

miquelsi:
https://redhat.atlassian.net/browse/ARO-29177

### What

Starting in OCP 4.22 (NE-1476, cluster-dns-operator@60b42af5), the DNS operator creates NetworkPolicy objects in openshift-dns (a default-deny policy plus a CIDR-only egress allow rule). Cilium's CIDR-based policy selectors cannot match the "host"/"remote-node" identities that kube-proxy's iptables DNAT causes CoreDNS's apiserver traffic to resolve to (the kube-apiserver-proxy ClusterIP rewrite happens before Cilium's eBPF datapath evaluates policy). As a result, CoreDNS's egress to the apiserver is dropped and DNS pods never become fully Ready on 4.22, while 4.20/4.21 are unaffected since no such NetworkPolicy exists there.

Add EnsureDNSAllowHostAPIServerCiliumNetworkPolicy in test/util/framework/cilium_helper.go, which queries the guest cluster's OpenShift version and, only for OCP >= 4.22, creates a CiliumNetworkPolicy allowing openshift-dns pods egress to the host/remote-node entities on TCP 6443. The check is retried (wait.PollUntilContextTimeout) to tolerate the CiliumNetworkPolicy CRD only being registered once cilium-operator schedules on a node.

The fix is explicitly version-gated: applying it unconditionally on 4.20/4.21, where openshift-dns currently has no NetworkPolicy at all, would flip DNS pod egress to default-deny and could break other legitimate egress (e.g. upstream DNS resolution) not covered by this narrow allow rule.

Wire the helper into all three Cilium CNI e2e tests, called right after node pool creation and before the cilium/nodes-ready verification step:
- cluster_create_cni_cilium.go
- cluster_create_complex_cilium_kv.go
- cluster_create_feature_aggregation.go (after the first node pool)

Live-validated on OCP 4.22.10: DNS pods went from 1/2 Ready (stuck) to 2/2 Running within ~30s of the CiliumNetworkPolicy being created, with dns/network/ingress ClusterOperators becoming healthy. Also validated no regression on 4.20/4.22 clean re-runs of all three tests.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/e2e/cluster_create_cni_cilium.go
+++ b/test/e2e/cluster_create_cni_cilium.go
@@ -96,6 +96,8 @@ var _ = Describe("Customer", func() {
 				framework.GetAdminRESTConfigTimeout,
 			)
 			Expect(err).NotTo(HaveOccurred(), "failed to get admin REST config for cluster %q", customerClusterName)
+			adminRESTConfig.QPS = 50
+			adminRESTConfig.Burst = 100
 			Expect(verifiers.VerifyHCPCluster(ctx, adminRESTConfig)).To(Succeed(), "failed to verify HCP cluster %q is available", customerClusterName)
 
 			By("getting kubeconfig content for Helm")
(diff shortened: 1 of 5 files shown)


raelga (approved):
/lgtm
/approve

PR #6701, opened 2026-08-26
docs: harden stuck cluster deletion SOP with verified cleanup lessons

hbhushan3:
Resolves [AROSLSRE-1892](https://redhat.atlassian.net/browse/AROSLSRE-1892)

### What

Updates `docs/ops/cleanup-stuck-cluster-deletion.md` with lessons verified while manually recovering a stage HCP cluster whose deletion had been stuck for several days.

- Adds a **mandatory gate** requiring Clusters Service and Maestro verification before any resolution strategy patches or deletes management-cluster resources.
- Matches the cluster ID against the **whole Maestro ResourceBundle object** instead of only `.manifests`, and adds a label-inventory query for `api.openshift.com/id`.
- Adds two diagnostics: explicit deletion timestamps (the default `kubectl` table omits them) and `kubectl get deployment capi-provider cluster-api`.
- Reorders Strategy 4 so machine resources are cleared **before** their parents (AzureMachine → Machine/MachineSet/MachineDeployment → Deployments → CAPI `Cluster` → `HostedControlPlane` → NodePool → HostedCluster → ManifestWork).
- Converts the remaining finalizer patches to idempotent merge patches, consistent with the rationale already stated in the doc.
- Adds an **Azure resource safety** warning that clearing an `AzureMachine` finalizer removes Kubernetes bookkeeping only and can leak Azure resources.
- Rewrites Scenario 5 around the real failure mode: CAPI machine deletion that cannot finish.

### Why

A cleanup script deleted a cluster-scoped `AppliedManifestWork`, which propagated HostedCluster deletion while `capi-provider` and `cluster-api` were scaled to zero before machine teardown completed. CAPI/CAPZ finalizers were left stranded with Machines pinned at `WaitingForInfrastructureDeletion`, and the controllers sat at `0/1` for more than five days, so nothing could self-resolve.

Following the SOP as written was not sufficient in two ways that mattered:

1. The Maestro lookup filtered only on `.manifests`. ResourceBundle names are UUIDs and the cluster ID can appear in `metadata.labels`, so a matching bundle could be missed and downstream resources would be recreated after cleanup.
2. Strategy 4 cleared Deployment, CAPI `Cluster`, and `HostedControlPlane` finalizers before the machine resources, which hides the actual blockers and can leave cloud resources orphaned.

The reordering and the Azure warning came directly from the recovery: clearing `AzureMachine` finalizers unblocked Kubernetes but does **not** delete the underlying VM, NICs, or disks, so the managed resource group still needs an explicit leak check.

### Testing

Documentation-only change; no unit, integration, or E2E tests apply.

The updated commands were exercised during the live recovery of the affected stage cluster. After cleanup, Clusters Service and Maestro queries returned `[]`, both the HostedCluster and control-plane namespaces returned `NotFound`, no HostedCluster / NodePool / CAPI `Cluster` / Machine / AzureMachine / ManifestWork / AppliedManifestWork remained, and nothing was recreated on a follow-up check.

The reordering in Strategy 4 is the behavioral change worth the closest look — the previous ordering is what made the blockers hard to identify during the incident.

### PR Checklist

- [x] Screenshots included (if graph/UI/metrics changes) — N/A
- [x] Self-reviewed the diff
- [x] Draft PR used for WIP (if applicable) — N/A

### DIFF
--- a/docs/ops/cleanup-stuck-cluster-deletion.md
+++ b/docs/ops/cleanup-stuck-cluster-deletion.md
@@ -181,7 +181,7 @@ kubectl get <kind> -n <namespace> -o json | \
   jq -r '.items[] | "\(.metadata.name)\tdel=\(.metadata.deletionTimestamp)"'
 
 # Check whether CAPI/CAPZ can still reconcile machine deletion
-kubectl get deployment capi-provider cluster-api -n <control-plane-namespace>
+kubectl get deployment capi-provider cluster-api -n <cp-namespace>
 
 # Check Hypershift operator logs
 kubectl logs -n hypershift deployment/operator --tail=50 | grep <cluster-name-or-id>
@@ -248,17 +248,17 @@ kubectl exec -n clusters-service deployment/clusters-service -- \
 #
 kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
   "curl -s 'http://localhost:8000/api/maestro/v1/resource-bundles?size=2900'" | \
-  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(tostring | test($cid)) |
+  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(tostring | index($cid)) |
   {id, name, created_at, consumer_name, deleted_at,
    labels: .metadata.labels}]'
 ```
 
-ResourceBundle names are UUIDs. Match the cluster ID against the whole object because it may appear in `manifests` or `metadata.labels`, then confirm the label inventory when needed:
+ResourceBundle names are UUIDs. Match the cluster ID against the whole object because it may appear in `manifests` or `metadata.labels`; a bundle missed here is one that Maestro can later use to recreate everything you just deleted. `index` is a literal substring match, so the ID is never interpreted as a regex. Confirm the label inventory when a match looks unexpected:
 
 ```bash
 kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
   "curl -s 'http://localhost:8000/api/maestro/v1/resource-bundles?size=2900'" | \
-  jq -r '[.items[].metadata.labels["api.openshift.com/id"]] | unique'
+  jq -r '[.items[].metadata.labels["api.openshift.com/id"] // empty] | unique'
 ```
 
 ### Phase 3: Resolution
@@ -411,7 +411,11 @@ for kind in machines.cluster.x-k8s.io machinesets.cluster.x-k8s.io machinedeploy
 done
 
 # Deployments with hypershift.openshift.io/component-finalizer
-for name in $(kubectl get deployment -n <cp-namespace> -o jsonpath='{.items[*].metadata.name}'); do
+# Only the Deployments actually holding that finalizer -- a CP namespace has dozens of
+# Deployments and stripping finalizers from all of them is far broader than intended.
+for name in $(kubectl get deployment -n <cp-namespace> -o json | \
+  jq -r '.items[] | select((.metadata.finalizers // []) |
+    index("hypershift.openshift.io/component-finalizer")) | .metadata.name'); do
   kubectl patch deployment $name -n <cp-namespace> \
     --type=merge -p='{"metadata":{"finalizers":null}}'
 done
@@ -461,7 +465,7 @@ kubectl exec -n clusters-service deployment/clusters-service -- \
 # On the Service Cluster - verify no orphaned Maestro bundles
 kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
   "curl -s 'http://localhost:8000/api/maestro/v1/resource-bundles?size=2900'" | \
-  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(tostring | test($cid)) | {id, name, deleted_at}]'
+  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(tostring | index($cid)) | {id, name, deleted_at}]'
 ```
 
 > **Note**: If CS still has the cluster record in `uninstalling` state after management cluster cleanup, re-issue the ARO-HCP `DELETE` (Strategy 1). With Maestro and the management cluster cleared, the CS controller can now complete the deletion and subsequent `GET`s will return `404`.
@@ -505,7 +509,7 @@ kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
 
 ### Scenario 5: CAPI Machine Deletion Cannot Finish
 
-**Symptoms**: The CP namespace remains `Terminating`; Machine conditions remain at `WaitingForInfrastructureDeletion`; and `kubectl get deployment capi-provider cluster-api -n <cp-ns>` shows one or both controllers at `0/N` replicas. Namespace conditions may reference infrastructure or machine kinds, or may only show the parent `Cluster`, `HostedControlPlane`, and Deployments that are waiting on them.
+**Symptoms**: The CP namespace remains `Terminating`; Machine conditions remain at `WaitingForInfrastructureDeletion`; and `kubectl get deployment capi-provider cluster-api -n <cp-namespace>` shows one or both controllers at `0/N` replicas. Namespace conditions may reference infrastructure or machine kinds, or may only show the parent `Cluster`, `HostedControlPlane`, and Deployments that are waiting on them.
 
 **Cause**: CAPI machine resources hold their own finalizers, but CAPI/CAPZ is no longer running to process them. The stale `WaitingForInfrastructureDeletion` condition cannot self-resolve. Common blockers in the CP namespace:
 
@@ -558,7 +562,7 @@ curl -s 'http://localhost:8002/api/maestro/v1/resource-bundles?size=2900' | \
 
 # Search for a specific cluster's bundles
 curl -s 'http://localhost:8002/api/maestro/v1/resource-bundles?size=2900' | \
-  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(tostring | test($cid)) |
+  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(tostring | index($cid)) |
   {id, name, deleted_at, labels: .metadata.labels}]'
 ```
 
--- a/docs/ops/cleanup-stuck-cluster-deletion.md
+++ b/docs/ops/cleanup-stuck-cluster-deletion.md
@@ -239,7 +239,8 @@ kubectl exec -n clusters-service deployment/clusters-service -- \
 #### Step 6: Check Maestro Resource Bundles
 
 ```bash
-# Search for resource bundles by cluster ID (search in manifest content)
+# Search for resource bundles by cluster ID (matches anywhere in the bundle,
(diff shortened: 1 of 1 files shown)


docs/ops/cleanup-stuck-cluster-deletion.md
```diff
@@ -239,11 +248,19 @@ kubectl exec -n clusters-service deployment/clusters-service -- \
 #
 kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
   "curl -s 'http://localhost:8000/api/maestro/v1/resource-bundles?size=2900'" | \
-  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(.manifests | tostring | test($cid)) |
+  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(tostring | test($cid)) |
   {id, name, created_at, consumer_name, deleted_at,
```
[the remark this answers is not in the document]
hbhushan3: Taking the literal-matching half of this, but keeping the object-wide scope deliberately.

Switched `test($cid)` → `index($cid)` in all three places, so the cluster ID is a literal substring and is never interpreted as a regex. Verified:

```
$ jq --arg cid "a.c123" '[.items[] | select(tostring | index($cid)) | .id]' t.json
[]
```

On narrowing to `manifests` and `metadata.labels`: I'd rather not. The two failure modes here are not symmetric. A false positive shows the operator one extra bundle to eyeball, which costs seconds. A false negative means a bundle is missed, the operator proceeds to strip finalizers downstream, and Maestro recreates everything — which is exactly the trap this PR exists to close. The previous `.manifests`-only filter *was* that narrow, and it missed a bundle carrying the ID in `metadata.labels`.

Since I can't enumerate every field the ID may appear in, over-matching in a read-only discovery step is the safer default. Added a line to the doc stating that rationale and the literal-match guarantee.

docs/ops/cleanup-stuck-cluster-deletion.md
```diff
@@ -395,13 +410,19 @@ for kind in machines.cluster.x-k8s.io machinesets.cluster.x-k8s.io machinedeploy
   done
 done
 
+# Deployments with hypershift.openshift.io/component-finalizer
+for name in $(kubectl get deployment -n <cp-namespace> -o jsonpath='{.items[*].metadata.name}'); do
+  kubectl patch deployment $name -n <cp-namespace> \
+    --type=merge -p='{"metadata":{"finalizers":null}}'
+done
```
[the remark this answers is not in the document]
hbhushan3: Good catch — this one was a regression I introduced in this PR. The original had a single `kubectl patch deployment <name>`; my loop widened it to every Deployment in the CP namespace, which strips finalizers from dozens of unrelated Deployments. Fixed in e201d690a to select only those holding the finalizer:

```
$ jq -r '.items[] | select((.metadata.finalizers // []) | index("hypershift.openshift.io/component-finalizer")) | .metadata.name' d.json
kube-apiserver
```

(`oauth-openshift` with no finalizers and `other` with a different finalizer are both correctly skipped.)

docs/ops/cleanup-stuck-cluster-deletion.md
```diff
@@ -176,6 +176,13 @@ kubectl get hostedcluster <name> -n <namespace> \
 # Check what's blocking a Terminating namespace
 kubectl get namespace <namespace> -o json | jq '.status'
 
+# Show deletion timestamps explicitly (the default kubectl table omits them)
+kubectl get <kind> -n <namespace> -o json | \
+  jq -r '.items[] | "\(.metadata.name)\tdel=\(.metadata.deletionTimestamp)"'
+
+# Check whether CAPI/CAPZ can still reconcile machine deletion
+kubectl get deployment capi-provider cluster-api -n <control-plane-namespace>
```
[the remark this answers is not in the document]
hbhushan3: Agreed — normalized to `<cp-namespace>` in e201d690a. There were actually three variants for the same thing (`<control-plane-namespace>` here, `<cp-ns>` in Scenario 5, `<cp-namespace>` in Strategy 4); all are now `<cp-namespace>`.

docs/ops/cleanup-stuck-cluster-deletion.md
```diff
@@ -239,11 +248,19 @@ kubectl exec -n clusters-service deployment/clusters-service -- \
...
    labels: .metadata.labels}]'
 ```
 
+ResourceBundle names are UUIDs. Match the cluster ID against the whole object because it may appear in `manifests` or `metadata.labels`, then confirm the label inventory when needed:
+
+```bash
+kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
+  "curl -s 'http://localhost:8000/api/maestro/v1/resource-bundles?size=2900'" | \
+  jq -r '[.items[].metadata.labels["api.openshift.com/id"]] | unique'
+```
```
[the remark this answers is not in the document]
hbhushan3: Agreed — fixed in e201d690a using `// empty` so nulls are dropped before `unique`:

```
jq -r '[.items[].metadata.labels["api.openshift.com/id"] // empty] | unique'
```

Returns `["abc123"]` on sample data containing bundles with no such label, instead of `[null, "abc123"]`.

docs/ops/cleanup-stuck-cluster-deletion.md
```diff
@@ -239,11 +248,19 @@ kubectl exec -n clusters-service deployment/clusters-service -- \
 #
 kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
   "curl -s 'http://localhost:8000/api/maestro/v1/resource-bundles?size=2900'" | \
-  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(.manifests | tostring | test($cid)) |
+  jq --arg cid "${CLUSTER_ID}" '[.items[] | select(tostring | index($cid)) |
```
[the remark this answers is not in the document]
hbhushan3: Correct, and it's a leftover from my own change — I broadened the filter but left the header comment describing the old `.manifests`-only behavior. Fixed in aae03e636:

```bash
# Search for resource bundles by cluster ID (matches anywhere in the bundle,
# including manifest content and metadata.labels)
```

Grepped the rest of the file for similar stale claims; this was the only one.

raelga (approved):
/lgtm
/approve

docs/ops/cleanup-stuck-cluster-deletion.md
stevekuznetsov, line 114: this should be in Kusto

stevekuznetsov, line 126: this should be in kusto

stevekuznetsov, line 155: this should be in kusto

stevekuznetsov, line 169: all of this should be in kusto

stevekuznetsov, line 225: this should be in kusto

docs/ops/cleanup-stuck-cluster-deletion.md
```diff
@@ -230,7 +239,8 @@ kubectl exec -n clusters-service deployment/clusters-service -- \
 #### Step 6: Check Maestro Resource Bundles
```
stevekuznetsov, line 239: this should be in Kusto

docs/ops/cleanup-stuck-cluster-deletion.md
```diff
@@ -239,11 +249,19 @@ kubectl exec -n clusters-service deployment/clusters-service -- \
...
+
+```bash
+kubectl exec -n maestro deployment/maestro -c maestro-server -- sh -c \
+  "curl -s 'http://localhost:8000/api/maestro/v1/resource-bundles?size=2900'" | \
+  jq -r '[.items[].metadata.labels["api.openshift.com/id"] // empty] | unique'
+```
+
 ### Phase 3: Resolution
 
 #### Strategy 1: Delete via Clusters Service API (Preferred)
```
stevekuznetsov, line 267: never. use frontend api first

docs/ops/cleanup-stuck-cluster-deletion.md
stevekuznetsov, line 289: no

stevekuznetsov, line 321: absolutely not

stevekuznetsov, line 377: every command after this terrifies me

stevekuznetsov, line 451: this should be in kusto

stevekuznetsov, line 482: do we have an upstream bug or PR to hypershift to fix this?

stevekuznetsov, line 528: do we have an upstream bug or PR to fix this?

stevekuznetsov, line 532: do we have an upstream bug or PR to fix this?

stevekuznetsov, line 474: a scenario is only allowed to exist in this list if we have clear bug/PR upstream to fix the offending software, and the moment that bug is gone, the scenario is gone

stevekuznetsov, line 586: kusto

hbhushan3:
@stevekuznetsov — worth flagging up front: all 16 comments land on lines this PR doesn't touch.

The diff here is +43/−17 against a 600-line document that already existed. It adds a CS/Maestro verification gate *before* the destructive steps, fixes a Maestro lookup that was missing bundles carrying the cluster ID in `metadata.labels`, reorders Strategy 4 so machine resources are cleared before their parents, and warns that clearing an `AzureMachine` finalizer leaks the Azure VM/NIC/disks. It doesn't add a kubectl-discovery step or a new scenario — every line you commented on is pre-existing.

I'm not arguing the direction. Kusto for discovery, frontend API as the only sanctioned entry point, and "no scenario without an upstream bug" are all better than what's there. But that's a rewrite of the document, not a change to this delta. So:

1. Land this as-is.
2. The restructure is a follow-up card: **AROSLSRE-1921**.

I'd lean that way because this came out of a live stage recovery where the existing text actively misled me (the `.manifests`-only Maestro filter missed a bundle, and clearing parent finalizers first hid the real blockers). Leaving that in place while the rewrite lands seems worse than an intermediate state.

The card carries your four asks plus the open questions I need answered to make it real — chiefly which Kusto cluster/DB to target for stg/prod, whether Kusto also replaces the pre-flight check immediately before a destructive action (ingestion lag vs. live object state), whether there's an admin/force delete on the frontend for the CS-stuck-in-`uninstalling` case, and whether upstream trackers already exist for Scenarios 5/6/7. If the answer on Strategies 2/3/4 is that they shouldn't be documented at all, tell me what the sanctioned path is instead and I'll replace all three sections with it.

stevekuznetsov:
Actively misleading people into using something other than Kusto is also not great. I know it's pre-existing. Up to you how you split it up. Let's get it fixed O(now).

PR #6706, opened 2026-08-27
fix(cleanup): protect identity pools from expiry (AROSLSRE-1916)

raelga:
[AROSLSRE-1916](https://redhat.atlassian.net/browse/AROSLSRE-1916)

### What

Excludes slot-managed identity-container resource groups from
`cleanup-sweeper` discovery before the generic 15-day `persist=true` expiry
rule.

Adds regression coverage against the repository policy and documents why these
long-lived E2E pools are protected.

### Why

The restored DEV shard0 pool has 300 resource groups with `persist=true`, a
newly stamped `createdAt`, and no `managedBy` parent. Without this exclusion,
the four-hour cleanup sweeper would select the entire pool after 15 days and
reintroduce the `ResourceGroupNotFound` failures recovered under AROSLSRE-1895.

### Testing

- `cd tooling/cleanup-sweeper && go test ./...`
- `GOTOOLCHAIN=go1.25.7 make verify-yamlfmt`
- The regression test loads `resourcegroups.policy.yaml` and verifies:
  - an identity-container group older than 15 days is skipped
  - matching is case-insensitive
  - an unrelated `persist=true` group still expires after 15 days

### Special notes for your reviewer

This changes the policy consumed by the periodic `sweeper-rg-ordered` CI job.
The exclusion is intentionally narrow to resource-group names beginning with
`aro-hcp-msi-container-`.

### PR Checklist

- [x] PR is scoped to a single task (no mixed concerns)
- [x] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [x] Summary explains the "Why" behind the change
- [x] Linked to relevant ticket/issue
- [x] Screenshots included (if graph/UI/metrics changes) - not applicable
- [x] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [x] Draft PR used for WIP (if applicable) - not applicable
- [x] Commit history is clean (rebased/squashed)
- [x] Tricky code blocks are commented
- [x] Specific reviewers tagged
- [x] All comment threads resolved before merge

If E2E tests are included:

- [x] E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md) - not applicable
- [x] If new E2E use case is covered (via a new test or new check/verifier),
  demonstrate that the test is able to detect a defect/error and fail with
  proper error message and logs which communicates nature of the problem. - not applicable

PR #6708, opened 2026-08-27
fix(infra): retry cluster ARM step on Key Vault RBAC propagation (AROSLSRE-1918)

raelga:
Jira: [AROSLSRE-1918](https://redhat.atlassian.net/browse/AROSLSRE-1918)

### What

Add `ForbiddenByRbac` to the existing `automatedRetry.errorContainsAny` list on the `cluster` ARM step in both `dev-infrastructure/svc-pipeline.yaml` and `dev-infrastructure/mgmt-pipeline.yaml`.

### Why

During a prod Canary GlobalBuildout in eastus2euap, the `Service.Infra` cluster step failed with `ForbiddenByRbac` on `Microsoft.KeyVault/vaults/keys/write` for the `aks-etcd-encryption` key. Creating that key via bicep is a control-plane write, so the deployment identity needs a Key Vault role that grants `keys/write`. On a fresh region the role assignment can lag propagation, and the whole rollout step fails. Both cluster steps deploy `aks-cluster-base.bicep` and create that key, so both can hit the same race. The retry lets the step recover once the assignment lands.

This only covers the propagation race. A genuinely missing grant still fails after the retries, same as today, just a few minutes later.

### Testing

`make validate-config-pipelines` passes. This is a pipeline config change with no code path to unit test; the retry itself is exercised by EV2 at rollout time.

`errorContainsAny` is an OR match and case-insensitive, so I matched the stable error code `ForbiddenByRbac` rather than a longer phrase. Rollout to int/stg/prod happens via sdp-pipelines after this merges.

### PR Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/dev-infrastructure/mgmt-pipeline.yaml
+++ b/dev-infrastructure/mgmt-pipeline.yaml
@@ -425,6 +425,10 @@ resourceGroups:
       - "Resource is in Updating state"
       - "There was a problem communicating with the identity service"
       - "DeploymentScriptExceededMaxAllowedTime"
+      # Transient RBAC propagation lag, mostly on fresh region buildout: the
+      # deployment identity's Key Vault role assignment (needed to create the
+      # aks-etcd-encryption key) may not have propagated yet. Retry once it lands.
+      - "ForbiddenByRbac"
       maximumRetryCount: 3
       durationBetweenRetries: 2m
   - name: svc-mgmt-aks-permissions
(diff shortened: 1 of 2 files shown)


raelga:
/test e2e-parallel

Retesting: cluster-candidate-5-0 provisioning failed with:

  InternalServerError: [hypershiftHostedCluster] hosted cluster has no installed version; hosted cluster degraded: UnavailableReplicas: router deployment has 1 unavailable replicas

Same exact signature also just hit #6706 (unrelated docs/cleanup-sweeper PR) in the same time window, confirming this is a shared-CI data-plane provisioning episode, not caused by this PR's ARM-retry diff.

PR #6716, opened 2026-08-27
fix(e2e): retry transient 401 during ARM deployment LRO polling

hbhushan3:
[AROSLSRE-1920](https://redhat.atlassian.net/browse/AROSLSRE-1920)

### What

Extends the existing `operationStatuses` poller policy so it retries a transient `401 Unauthorized` once, alongside the `404 DeploymentNotFound` handling it already does.

The two transients get **separate attempt budgets**, so a 401 cannot consume the `DeploymentNotFound` allowance. Path scope is unchanged — GET requests to:

```
/providers/Microsoft.Resources/deployments/{name}/operationStatuses/{id}
```

Renamed `lroPollerRetryDeploymentNotFoundPolicy` → `lroPollerRetryPolicy`, since the name no longer described what it does.

### Why

In [prow run 2092949454679183360](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/branch-ci-Azure-ARO-HCP-main-e2e-prod-e2e-parallel/2092949454679183360) (switzerlandnorth, 2026-08-27 13:28), `Customer should be able to update node pool labels and taints` failed in `creating customer resources`. The sequence:

1. Credential acquisition — OK
2. Resource group creation — OK
3. `PUT` of deployment `customer-infra-cluster-np-labels-taints-2l24k9` — OK
4. Early poll of `operationStatuses/...` — `200 OK`
5. Later poll of the **same** endpoint — `401 Unauthorized`, `WWW-Authenticate: ... error="invalid_token"`

That aborted `PollUntilDone`. The spec never reached the RP — no `frontendLogs` rows exist for the resource group — so this is not a node-pool defect. An otherwise-healthy run is lost before it starts doing useful work.

**azcore will not recover from this on its own.** In `runtime.BearerTokenPolicy.handleChallenge` (azcore v1.21.1):

- CAE auto-retry is gated on `c.params["error"] == "insufficient_claims"` with a `claims` value. `error="invalid_token"` does not match, so `parseCAEChallenge` returns nil.
- ARM's `arm/runtime.NewBearerTokenPolicy` sets only `AuthorizationHandler{OnRequest: ...}`; `OnChallenge` is nil.
- Both branches miss, control reaches `default:`, and the 401 is returned to the pipeline untouched.

Crucially, it calls `b.mainResource.Expire()` on **every** 401 *before* deciding not to handle it. The cached token is therefore already invalidated by the time the error reaches a PerCall policy, so simply re-issuing the request acquires a fresh token. That is what makes a bounded retry here sufficient rather than a no-op replay of the same bad token.

Adding `401` to `policy.RetryOptions.StatusCodes` was rejected as an alternative: azcore matches on status code only and cannot scope to an endpoint, so it would slow every genuine authz failure across the suite.

### Testing

Unit tests, added alongside the existing ones in `azure_client_policies_test.go`:

| Test | Asserts |
| --- | --- |
| `retries 401 Unauthorized then succeeds` | one retry, 2 transport calls |
| `exhausts auth retries on persistent 401` | stops at the budget, 2 calls, surfaces the error |
| `401 does not consume the DeploymentNotFound budget` | 401 → 404 → 404 → OK all succeed with `MaxRetries: 2` |
| `does not retry 401 on non-matching paths` | a 401 outside the poller path is untouched, 1 call |

```
$ golangci-lint-v2.5.0 run --build-tags='E2Etests' ./util/framework/...
0 issues.

$ go test ./util/framework/ -count=1
ok      github.com/Azure/ARO-HCP/test/util/framework    0.106s
```

`gofmt -l` and `go vet` both clean. No E2E changes.

- **Frequency is 1 occurrence.** This was the "Related but distinct — not actioned" item in [AROSLSRE-1920](https://redhat.atlassian.net/browse/AROSLSRE-1920), recorded but deliberately not fixed at the time. The argument for doing it now is that the policy and its test scaffolding already exist, so the marginal cost is small. Happy to hold if you'd rather wait for a recurrence.
- **`MaxAuthRetries` defaults to 1**, deliberately. A broad 401 retry would turn a genuine RBAC misconfiguration into a slow failure instead of a fast one; one attempt covers the token-refresh case without meaningfully delaying real authz errors.
- **Any 401 on the guarded path is retried**, not just `invalid_token`. ARM surfaces this as `AuthenticationFailed` in some paths and `invalid_token` in others; filtering on the error code risks missing variants, and the path scope already keeps the blast radius tiny.
- `tooling/templatize/pkg/pipeline/arm_retry_policy.go` carries an independent copy of the 404-only policy. It is a separate package serving the deployment tool rather than the e2e harness, so I left it alone — flagging it in case you want it aligned.

### DIFF
--- a/test/util/framework/azure_client_policies.go
+++ b/test/util/framework/azure_client_policies.go
@@ -162,35 +162,56 @@ func (p *requestIDPolicy) Do(req *policy.Request) (*http.Response, error) {
 	return req.Next()
 }
 
-// lroPollerRetryDeploymentNotFoundPolicy is a pipeline policy that retries transient
-// 404 DeploymentNotFound errors during ARM deployment LRO polling.
+// lroPollerRetryPolicy is a pipeline policy that retries transient errors during
+// ARM deployment LRO polling.
 //
-// This addresses Azure Resource Manager's eventual consistency behavior where
-// the operationStatuses endpoint may return 404 immediately after a deployment
-// is created, even though the deployment will eventually succeed.
+// It handles two distinct transients:
+//
+// 404 DeploymentNotFound addresses Azure Resource Manager's eventual consistency
+// behavior where the operationStatuses endpoint may return 404 immediately after
+// a deployment is created, even though the deployment will eventually succeed.
+//
+// 401 Unauthorized addresses ARM rejecting a previously-valid bearer token
+// mid-poll (observed as WWW-Authenticate error="invalid_token"). azcore's
+// BearerTokenPolicy only auto-recovers from CAE challenges carrying
+// error="insufficient_claims", and ARM installs no OnChallenge handler, so every
+// other 401 is surfaced unhandled. It does however call Expire() on the cached
+// token before giving up, so simply re-issuing the request acquires a fresh one.
 //
 // The policy only activates for GET requests to URLs matching:
 //
 //	/providers/Microsoft.Resources/deployments/{name}/operationStatuses/{id}
 //
 // All other requests pass through unmodified.
-type lroPollerRetryDeploymentNotFoundPolicy struct {
+type lroPollerRetryPolicy struct {
 	MaxRetries     int
+	MaxAuthRetries int
 	BaseBackoff    time.Duration
 	MaxBackoff     time.Duration
 	MaxRetryWindow time.Duration
 }
 
-func NewLROPollerRetryDeploymentNotFoundPolicy() *lroPollerRetryDeploymentNotFoundPolicy {
-	return &lroPollerRetryDeploymentNotFoundPolicy{
+func NewLROPollerRetryPolicy() *lroPollerRetryPolicy {
+	return &lroPollerRetryPolicy{
 		MaxRetries:     5,
+		MaxAuthRetries: 1,
 		BaseBackoff:    2 * time.Second,
 		MaxBackoff:     10 * time.Second,
 		MaxRetryWindow: 90 * time.Second,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


hbhushan3:
/test lint

Retrying — the `lint` job has now failed twice on transient Go module proxy errors from the CI cluster, without ever reaching the linter.

Run [2093044040860700672](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6716/pull-ci-Azure-ARO-HCP-main-lint/2093044040860700672) (on `d45a9e8dc`) died fetching golangci-lint's own `.mod`, before a single file was linted:

```
+ make lint
(re)installing /go/bin/golangci-lint-v2.5.0
go: downloading github.com/golangci/golangci-lint/v2 v2.5.0
github.com/golangci/golangci-lint/v2@v2.5.0: read
  "https://proxy.golang.org/github.com/golangci/golangci-lint/v2/@v/v2.5.0.mod":
  stream error: stream ID 3; INTERNAL_ERROR; received from peer
make: *** [.bingo/Variables.mk:72: /go/bin/golangci-lint-v2.5.0] Error 1
```

The earlier run failed the same way on `quasilyte/regex/syntax` and `AdminBenni/iota-mixing` — also golangci-lint's own dependencies, also `INTERNAL_ERROR`. Zero `file.go:N:M:` findings in either log.

For what it's worth, `proxy.golang.org` was serving fine from outside CI at the same time, so this looks like cluster egress rather than the public proxy.

Lint is clean locally against the same pinned version:

```
$ golangci-lint-v2.5.0 run --build-tags='E2Etests' ./util/framework/...
0 issues.
```

raelga (approved):
/lgtm
/approve

PR #6729, opened 2026-08-28
test: guard nodeVersionInMinor against unmapped OCP major/minor (AROSLSRE-1939)

raelga:
Jira: [AROSLSRE-1939](https://redhat.atlassian.net/browse/AROSLSRE-1939)

### What

Replaces the `ocpToK8sMinorOffset` arithmetic constant in `nodeVersionInMinor` (added in #6727) with an explicit `ocpToK8sMinor` map, and adds guards that return a clear error for any OCP major other than 4, or any OCP minor not yet in the map.

### Why

`nodeVersionInMinor` computed the expected Kubernetes minor as `expectedSemver.Minor + 13`. That holds for OCP 4.14-4.22, but nothing stops it from computing a plausible-looking but wrong value once we test against an OCP major other than 4 (the e2e suite already has an OCP 5.0 candidate channel test) or a minor we haven't verified yet. An explicit map fails loudly instead of guessing.

### Testing

`go test ./test/util/verifiers/...` passes, including new cases in `TestNodeVersionInMinor` for an unmapped OCP minor and an unmapped OCP major, both of which now return a "no known Kubernetes minor mapping" error instead of a wrong number. `TestOcpToK8sMinorOffset` was replaced with `TestOcpToK8sMinor`, which checks the map's known-good entries directly.

This branches off #6727 (still open), since it builds on that PR's `nodeVersionInMinor`/`KubeletVersion` change. The diff will include all of #6727's changes until that PR merges; the actual delta for this PR is just the `ocpToK8sMinorOffset` -> `ocpToK8sMinor` map switch and the two new guard checks.

### PR Checklist
- [x] Tests added/updated
- [x] Documentation not applicable (internal test helper)

### DIFF
--- a/test/util/verifiers/nodes.go
+++ b/test/util/verifiers/nodes.go
@@ -329,14 +329,27 @@ func nodeReadyAndSchedulable(node *corev1.Node) bool {
 	return nodeReady(node) && !node.Spec.Unschedulable
 }
 
-// ocpToK8sMinorOffset is the well-known offset between OCP minor versions and
-// Kubernetes minor versions: OCP 4.X ships Kubernetes 1.(X+13).
-// For example, OCP 4.14 = k8s 1.27, OCP 4.15 = k8s 1.28, ..., OCP 4.22 = k8s 1.35.
-const ocpToK8sMinorOffset uint64 = 13
+// ocpToK8sMinor maps known OCP 4.x minor versions to the Kubernetes minor version they ship.
+// This is an explicit allowlist rather than an arithmetic offset: the OCP-to-Kubernetes minor
+// mapping has held steady at +13 across 4.14-4.22, but that's not a guarantee that holds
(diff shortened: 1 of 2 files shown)


raelga:
/test e2e-parallel

Retesting: job pull-ci-Azure-ARO-HCP-main-e2e-parallel/2093422811048579072 had 2 failing specs, both unrelated to this PR's diff (test/util/verifiers only, nodeVersionInMinor):

- cluster_create_hypershift_presubmit.go:82: `hosted cluster degraded: UnavailableReplicas: router deployment has 1 unavailable replicas`
- cluster_delayed_role_assignments.go:227: cluster entered terminal Failed state, same router-unavailable-replica pattern in the underlying provisioning

This is the same cluster-creation infra flake (router deployment unavailable replicas) seen intermittently on the INT e2e-integration job before opening this PR, not caused by this change.

PR #6737, opened 2026-08-29
fix(backend): omit no-message sources from operation status (AROSLSRE-1943)

raelga:
[AROSLSRE-1943](https://redhat.atlassian.net/browse/AROSLSRE-1943)

### What

`PickWorstOperationState` (`backend/pkg/utils/operationutils/operation_state.go`) now skips sources with an empty `Message` when building the merged operation status string, instead of rendering them as a `[source] <no_message>` placeholder.

### Why

While investigating a recurring INT `e2e-integration-e2e-parallel` failure, the operation status showed `[clusterServiceClusterStatus] <no_message>`, which reads like an unreported error. Tracing the code showed it isn't one: any source tied for the current worst provisioning state gets merged into the message, even ones with nothing to report (just still reconciling). During normal `Provisioning` this shows up for basically every cluster create, making it impossible to tell a genuine unreported error apart from routine in-progress noise.

### Testing

```
cd backend && go build ./... && go vet ./pkg/utils/operationutils/...
cd backend && go test ./pkg/utils/operationutils/... -run TestPickWorstOperationState -v
```

All existing and updated unit tests pass. Updated the "empty message uses placeholder" test case in `operation_state_test.go` to reflect the new omit-instead-of-placeholder behavior, and added a case covering an empty-message source alongside one with a real message.

Presentation-only change to the merged status string, no change to reconciliation/provisioning logic. Found while triaging INT capacity/failures (see [AROSLSRE-1942](https://redhat.atlassian.net/browse/AROSLSRE-1942) for the related, separate mgmt-capacity fix in sdp-pipelines).

### DIFF
--- a/backend/pkg/utils/operationutils/operation_state.go
+++ b/backend/pkg/utils/operationutils/operation_state.go
@@ -79,6 +79,11 @@ func CompareOperationState(lhs, rhs *OperationState) int {
 }
 
 // PickWorstOperationState expects states pre-sorted and returns the worst state with merged messages.
+//
+// Only sources that report an actual message are included in the merged message: a source tied
+// for the worst provisioning state with no message of its own has nothing blocking to report (e.g.
+// it is simply still in progress), so it is omitted rather than rendered as a confusing "<no_message>"
+// placeholder that reads like an error.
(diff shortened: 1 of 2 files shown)


gmfrasca (commented):
/lgtm

PR #6754, opened 2026-08-31
fix: improve error messages for nodepool upgrades and cluster create timeouts

avollmer-redhat:
## Why

Two separate backend log/error messages are improved to make operational debugging easier:

1. **Nodepool upgrade timeouts**: The `VerifyNodePoolUpgrade` failure message previously dumped 50+ release image names per node (thousands of characters), making it nearly impossible to read. Now shows a concise summary with image count and kubelet/CRI-O versions.

2. **Cluster create timeouts**: When cluster create operations approach the 19-minute deadline waiting for HyperShift control plane convergence, the backend logged `"hosted cluster has no installed version"` with the raw status struct. Now logs a human-readable message like `"hosted cluster control plane version not yet completed: version 4.23.0 is Partial (want Completed), started 5m0s ago"`, making it clear what the control plane is doing.

## What

- `test/e2e/`: Replace full release image list in nodepool upgrade errors with concise summary (Fixes [AROSLSRE-1938](https://redhat.atlassian.net/browse/AROSLSRE-1938))
- `backend/pkg/controllers/operationcontrollers/`: Add `describeVersionHistory()` helper that formats `ControlPlaneUpdateHistory` entries with version, state, and elapsed time; update `hostedClusterOperationStatus()` to use it (Relates-to [AROSLSRE-1888](https://redhat.atlassian.net/browse/AROSLSRE-1888))
- Tests added for `describeVersionHistory()` covering empty history, single/multiple partial versions, and elapsed time formatting

## Checklist
- [x] Descriptive title in Conventional Commits format
- [x] Tests pass
- [x] Self-reviewed


JakobGray (commented):
The branch is 3538 commits behind main. The change looks like an improvement, but it might have to be wired differently following a rebase.

backend/pkg/controllers/cluster/operations/operation_cluster_create.go
```diff
@@ -449,3 +450,27 @@ func withDegradedSuffix(message string, hostedCluster *v1beta1.HostedCluster) st
...
+		if !entry.StartedTime.IsZero() && entry.State != configv1.CompletedUpdate {
+			elapsed := time.Since(entry.StartedTime.Time).Truncate(time.Second)
+			if elapsed < 0 {
+				elapsed = 0
+			}
+			desc += fmt.Sprintf(", started %s ago", elapsed)
+		}
+		descriptions = append(descriptions, desc)
+	}
+	return fmt.Sprintf("hosted cluster control plane version not yet completed: %s", strings.Join(descriptions, "; "))
```
raelga, line 478: this always prefixes with "not yet completed" even if an entry in the list is actually Completed. right now it's only ever called from the branch where none of them are, so it's fine, but if this ever gets reused elsewhere the wording would be misleading. maybe worth a comment on the function saying it assumes the caller already checked that

raelga: actually let's go with skipping the Completed entries instead of just documenting it, keeps the wording honest without relying on callers to remember the precondition:

```go
func describeVersionHistory(history []v1beta1.ControlPlaneUpdateHistory) string {
	if len(history) == 0 {
		return "hosted cluster has no version history entries"
	}
	descriptions := make([]string, 0, len(history))
	for _, entry := range history {
		if entry.State == configv1.CompletedUpdate {
			continue // only describe versions still in flight
		}
		desc := fmt.Sprintf("version %s is %s (want %s)", entry.Version, entry.State, configv1.CompletedUpdate)
		if !entry.StartedTime.IsZero() {
			elapsed := time.Since(entry.StartedTime.Time).Truncate(time.Second)
			if elapsed < 0 {
				elapsed = 0
			}
			desc += fmt.Sprintf(", started %s ago", elapsed)
		}
		descriptions = append(descriptions, desc)
	}
	return fmt.Sprintf("hosted cluster control plane version not yet completed: %s", strings.Join(descriptions, "; "))
}
```

avollmer-redhat: Applied, thanks! Skipping CompletedUpdate entries entirely is cleaner than documenting the precondition. Updated the test to match.

backend/pkg/controllers/cluster/operations/operation_cluster_create.go
```diff
@@ -449,3 +450,30 @@ func withDegradedSuffix(message string, hostedCluster *v1beta1.HostedCluster) st
...
+}
```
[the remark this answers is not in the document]
raelga: @copilot good catch, that's a real gap in the fix I suggested. Since the only caller guards on `!anyVersionInstalled` this can't happen today, but I'd rather not leave a function that can silently produce a malformed string if it's ever reused. Adding an explicit check for the case:

```go
	if len(descriptions) == 0 {
		return "hosted cluster control plane version history has no in-flight entries"
	}
	return fmt.Sprintf("hosted cluster control plane version not yet completed: %s", strings.Join(descriptions, "; "))
```

and a test case with an all-Completed history. No further action needed from you once this lands.

raelga (approved):
/lgtm
/approve

PR #6756, opened 2026-09-01
fix(test): remove private KAS e2e DNS workaround and add regression check

zgalor:
## Summary

- Remove the DNS bypass hack in the private KAS e2e test that was hiding [ARO-29270](https://redhat.atlassian.net/browse/ARO-29270)
- The test previously looked up the internal LB IP directly (`GetPrivateKASInternalIP`) and overwrote `adminRESTConfig.Host`, bypassing DNS entirely
- Now that CS MR !434 fixes the public DNS A record to point to the private IP, the test uses the kubeconfig as-is (DNS hostname → private IP → internal LB → KAS)
- Added an explicit DNS resolution check (`net.LookupHost` + `IsPrivate()`) to catch future regressions where DNS might point to a public IP again
- Deleted `resource_group_helper.go` — `GetPrivateKASInternalIP` was its only function and is now unused

## Why

The original test was designed around a known DNS limitation that turned out to be a customer-facing bug. The hack meant the e2e test could pass even though customers couldn't actually connect to their private KAS clusters via the documented API URL.

## Test plan

- [ ] `go build -tags E2Etests ./test/e2e/` — compiles
- [ ] `golangci-lint run --build-tags E2Etests ./e2e/` — no lint issues
- [ ] CI e2e: private KAS test passes with DNS-based connectivity
- [ ] DNS resolution check asserts the resolved IP is RFC 1918 private


zgalor:
@bennerv @machi1990 PTAL

zgalor:
@machi1990 @miquelsi PTAL

test/e2e/cluster_create_private_kas.go
```diff
@@ -107,6 +109,24 @@ var _ = Describe("Customer", func() {
 			apiURL := *cluster.Properties.API.URL
 			GinkgoLogr.Info("Cluster created with private KAS", "clusterName", customerClusterName, "apiURL", apiURL)
 
+			By("verifying API hostname DNS resolves to a private IP")
+			// The public DNS A record for a private KAS cluster must point
+			// to the private IP of the internal load balancer — not the
```
miguelsorianod, line 113: Why not then check that the DNS value resolves to the LB private IP using GetPrivateKASInternalIP? It seems it would be a more robust test

zgalor: Good point, and I considered this. The concern with comparing against `GetPrivateKASInternalIP` is that it re-introduces the Azure ARM LB lookup we just removed, creating two sources of truth in the test. If either the DNS record or the LB API behaves unexpectedly, it becomes harder to tell which side is wrong.

The check as written asserts the invariant that matters: the public DNS record must resolve to a private (RFC 1918) IP. That's exactly what was broken in [ARO-29270](https://redhat.atlassian.net/browse/ARO-29270) - the record pointed to a public IP. `IsPrivate()` catches that whole class of regression without coupling to the specific IP value, which could legitimately change if the LB is recreated.

The functional kubectl connectivity check from the VM is then the stronger end-to-end proof: it validates that the IP DNS resolved to is *actually reachable* as the KAS endpoint from inside the VNet. A correct DNS record pointing to a wrong IP would still fail here.

That said, if you feel the explicit IP comparison adds value, we could do it without re-adding `GetPrivateKASInternalIP` - by parsing the host from `adminRESTConfig.Host` after the fact and comparing. But given that the connectivity check already catches incorrect routing, I lean toward keeping it simple. Happy to change if there's consensus.

zgalor: Done. I've updated the DNS check to compare against the specific LB IP fetched via `GetPrivateKASInternalIP` rather than just asserting `IsPrivate()`. @machi1990  raised the same point in Slack, and I agree it makes the assertion more explicit and robust.

machi1990: Thank you

machi1990 (approved):
/lgtm

zgalor:
Also need a maintainer's approval @miguelsorianod @miquelsi

test/util/framework/resource_group_helper.go
miguelsorianod, line 51: How do we know that it's the LB associated to the Private KAS? this just seems to be listing LBs within the MRG and returning the first one that has a private ip

miguelsorianod: it would be good to narrow it to the expected LB name. Can we get it / calculate it somehow?

zgalor: Good question. I looked into both options for narrowing the lookup:

- **By LB name**: the LB is named `int-<clusterID>` where `clusterID` is HyperShift's internal ID, not the ARM cluster name - it's not accessible from the test without querying the management cluster.
- **By rule name**: the private KAS Kubernetes Service (`kube-apiserver-private`) has no named port, so Azure CCM derives the rule name from the service name and port number, making it cluster-specific rather than a stable constant we can rely on.

However, this test covers private KAS only (no private ingress), meaning there is exactly one internal LB in the managed RG - so the lookup is already unambiguous. I've updated the comment to make that explicit. Would you agree that's sufficient for this PR?

miguelsorianod: It would be preferable to narrow it so it's resilient to changes over time. Is `clusterID` Hypershift's internalID or is it the CS cluster id?

zgalor: After checking against a real cluster, I confirmed that HyperShift names the LB rule `kube-apiserver` on the internal KAS LB. That's now the single discriminator used in `GetPrivateKASInternalIP` - it's unambiguous even for fully-private clusters that have two internal LBs. Updated in the latest commit.

miguelsorianod: It's better than the previous approach although it's not the best one.
In theory there can still be collisions because two LBs can have the same rule name or rule name suffix. The LB name itself within a RG is collision-free so that would be the best approach.

I am not going to block it as it's an improvement compared to before and it's less likely to collide but using a collision-free approach would be the way to go.

miguelsorianod:
/lgtm

This will need approval from an owner of the `test` directory

test/util/framework/resource_group_helper.go
zgalor: We don't know the exact cluster ID because it's not exposed through the ARM API,
Would you prefer that we check for the LB name `int-` prefix in addition to the `kube-apiserver` LB rule suffix?

miguelsorianod: > We don't know the exact cluster ID because it's not exposed through the ARM API,
ah, I see...

>  Would you prefer that we check for the LB name int- prefix in addition to the kube-apiserver LB rule suffix?

That would be even more robust. Although again technically not collision-free. I leave it up to you at this point.

zgalor: Done. If you could take a final look 🙏

machi1990 (approved):
/lgtm

zgalor:
@patriksuba PTAL

PR #6769, opened 2026-09-01
fix(mgmt-agent): fire node-health on the wedge shape prod produces (AROSLSRE-1973)

raelga:
[AROSLSRE-1973](https://redhat.atlassian.net/browse/AROSLSRE-1973)

### What

Lowers `failuresFloor` on the `swift-vf-teardown` detector from 3 to 2, and pins the pod count against the captured incident with a production-evidence test.

### Why

The detector has never fired in production. It is on by default and running in every region, but a check across eastus2, uksouth and australiaeast found only controller lifecycle logs and zero `detector fired; node labeled wedged`.

It is not starved of input. eastus2 alone saw 2124 `mtpnc is not ready` events across 21 nodes in 14 days.

`failuresFloor` counts distinct pods stuck past the dwell, and a hard wedge produces very few of them. Once a node cannot build a sandbox, almost nothing new reaches a running state there, so a couple of pods retry indefinitely rather than a crowd of pods each failing once. Counting distinct pods is close to backwards for this case, because being wedged is what stops new pods arriving to be counted.

The captured uksouth node, counting distinct pods rather than events:

```
node                                distinct pods   events   span
aks-userswft2-40171262-vmss000001               2     1193   58.3h
```

One sampled hour inside that wedge, 2026-07-27 09:00 to 10:00, gave 21 failure events, 2 distinct failing pods and 0 fresh sandbox successes. `requireZeroSuccess` was satisfied, so the floor of 3 was the only thing holding it back.

Flaps are a different population and are already handled. Over 21 days australiaeast never exceeded 1.8 retries per pod, against 596 on the wedged node. A flapping pod gets its sandbox on retry, so it never reaches the dwell, and its node has other successes anyway. The false-positive protection comes from dwell and `requireZeroSuccess`, not from the floor.

### Testing

- Unit tests

`TestProductionHardWedgeShapeFires` replays the captured 2 pod, 0 success shape and asserts it fires. It is a real regression guard, not a tautology: it passes at `failuresFloor: 2` and fails at 3 with a message naming the cause. I verified both directions by flipping the constant.

The existing below-the-floor case in `decide_test.go` moves from 2 pods to 1, so it keeps testing the same property against the new floor.

`go test ./...` and `go vet` pass across `mgmt-agent`.

No new firings on current traffic: measured across every eastus2 window with 2 or more failing pods over 3 days, 0 of 19 would fire, all still blocked by a fresh success.

```
pods in window   windows   would fire at floor 2   blocked by success
2                     10                       0                  10
3                      5                       0                   5
4 or more              4                       0                   4
```

Behaviour is unchanged for everything except the 2 pod case, and that case is exactly the captured wedge.

This does not touch remediation. The controller still only labels and annotates, and its RBAC on nodes is get/list/watch/patch with no delete or eviction, so cordon, taint, evict and delete remain owned by a controller that does not exist yet.

The signatures are untouched. They are already pinned against this incident and match it correctly.

### PR Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/mgmt-agent/pkg/controller/nodehealth/detectors/decide_test.go
+++ b/mgmt-agent/pkg/controller/nodehealth/detectors/decide_test.go
@@ -155,11 +155,11 @@ func TestDecide(t *testing.T) {
 			name: "below the stuck-pod floor is not wedged",
 			node: testNode(true, true),
 			events: func() []*corev1.Event {
-				e, _ := stuckFailing(2, ago(15*time.Minute))
+				e, _ := stuckFailing(1, ago(15*time.Minute))
 				return e
 			}(),
 			pods: func() []*corev1.Pod {
-				_, p := stuckFailing(2, ago(15*time.Minute))
+				_, p := stuckFailing(1, ago(15*time.Minute))
 				return p
 			}(),
 			want: DecisionUnknown,
--- a/mgmt-agent/pkg/controller/nodehealth/detectors/production_evidence_test.go
+++ b/mgmt-agent/pkg/controller/nodehealth/detectors/production_evidence_test.go
@@ -266,3 +266,69 @@ func TestFutureDatedSuccessDoesNotSuppressDetection(t *testing.T) {
 		t.Errorf("a future-dated success suppressed detection of a real wedge: Decide = %v, want %v", got, DecisionWedged)
 	}
 }
+
+// TestProductionHardWedgeShapeFires pins the failure floor against the pod count
+// the captured incident actually produced.
+//
+// The other tests in this file replay the wedge with three stuck pods, which is
+// comfortably over any floor and so cannot detect a floor set too high. The
+// distinguishing property of a hard wedge is the opposite of a storm: once the
+// node cannot build a sandbox, the scheduler stops getting new pods to a running
+// state there, so a very small number of pods retry indefinitely instead of many
+// pods failing once each.
+//
+// Kusto for the captured uksouth node over the incident, counting distinct pods
+// rather than events, shows exactly that:
+//
+//	node                                distinct pods   events   span
+//	aks-userswft2-40171262-vmss000001               2     1193   58.3h
+//
+// One sampled hour inside the wedge, 2026-07-27 09:00 to 10:00, is the shape
+// replayed below: 2 distinct failing pods, no fresh sandbox success at all.
+//
+// The same query over flapping (not wedged) nodes returns 30 to 70 distinct pods
+// per node, with successes throughout, so the flap population is ruled out by
+// dwell and requireZeroSuccess rather than by the floor. That is why the floor
+// can sit at 2 without letting flaps through: a flapping pod gets its sandbox on
+// retry, so it never stays stuck for the dwell.
+func TestProductionHardWedgeShapeFires(t *testing.T) {
+	now := time.Date(2026, 7, 27, 9, 30, 0, 0, time.UTC)
+	const ns = "openshift-monitoring"
+	const host = "aks-userswift0-00000000-vmss000000"
+
+	var pods []*corev1.Pod
+	var events []*corev1.Event
+	for i, name := range []string{"ovnkube-node-x1", "prometheus-k8s-0"} {
+		uid := types.UID("uid-" + name)
+		pods = append(pods, &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name, UID: uid},
+			Status: corev1.PodStatus{
+				Phase: corev1.PodPending,
+				Conditions: []corev1.PodCondition{{
+					Type:               corev1.PodReadyToStartContainers,
+					Status:             corev1.ConditionFalse,
+					LastTransitionTime: metav1.NewTime(now.Add(-45 * time.Minute)),
+				}},
+			},
+		})
+		events = append(events, &corev1.Event{
+			ObjectMeta:     metav1.ObjectMeta{Namespace: ns, Name: name + ".evt"},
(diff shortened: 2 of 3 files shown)


raelga:
Cross-linking [#6776](https://github.com/Azure/ARO-HCP/pull/6776) ([AROSLSRE-1979](https://redhat.atlassian.net/browse/AROSLSRE-1979)), which changes the same detector and interacts with the reasoning here.

That PR scopes `requireZeroSuccess` so only pods requesting `aro.openshift.io/swift-nic` count as success, because on a partial wedge the overlay keeps working and its pod starts mask a dead delegated-NIC path.

This PR argues the lower floor is safe partly because "the node's other successes trip requireZeroSuccess anyway", and [#6776](https://github.com/Azure/ARO-HCP/pull/6776) narrows exactly that guard. So I re-measured the false-positive check in prod eastus2 over 3 days, comparing both success rules:

```text
windows with >=2 failing pods            41
would fire under the broad success rule   0
would fire under the scoped success rule  0
newly firing due to scoping               0
```

The differential is zero, so the safety argument here still holds once [#6776](https://github.com/Azure/ARO-HCP/pull/6776) lands. Method caveat: the success proxy was "a NIC pod observed Running on the node in the window", which over-counts relative to the real `SuccessAt`, so both columns move the same way.

The branches auto-merge with no conflicts and the full `nodehealth` suite passes on the combined tree. The two fix different wedge shapes, this one the 2-pod hard wedge, [#6776](https://github.com/Azure/ARO-HCP/pull/6776) the partial wedge where the overlay stays up, so both are needed.

raelga:
/unhold
/test e2e-parallel

Unholding: the bot held this after 3 retests on 8c7b7dd, but the e2e-parallel failures are not caused by this PR.

Last run ([2095100185347624960](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6769/pull-ci-Azure-ARO-HCP-main-e2e-parallel/2095100185347624960)) failed 20 specs, all of them timing out at HCP cluster creation after ~19 minutes, across completely unrelated areas: `admin_api`, `kusto_logs_present`, `oidc_issuer_workload_identity`, `hcp_backups`, `arm64_nodepool`, `kms_key_rotation`, `idms_lifecycle` and more. Example, `cluster_create_private_ingress.go:103`, creation starts 11:33:31 and the spec fails at 11:52:46.

This PR cannot produce that. The diff is two test files plus one constant in `swift_vf.go` (`failuresFloor` 3 to 2), and the only effect of a detector firing is a label. The controller does not cordon, taint, evict or delete:

```go
// Mitigation of a labeled node (cordon, taint, evict, delete) is owned
// by a separate controller; this controller stops at labeling.
```

Nothing in that path can stop a guest cluster from provisioning.

Scope check backs it up: 5 of the 12 most recent open PRs are failing e2e-parallel right now, including [#6775](https://github.com/Azure/ARO-HCP/pull/6775), which changes a single shell script (`hack/ci/grafana-preview.sh`) and cannot touch provisioning either. Mass cluster-create timeouts across many specs and many PRs is the shared-CI episode signature, not a regression here.

The last run was over an hour ago, so I am kicking off a fresh one rather than retesting back to back.

raelga:
/retest

The run never reached the tests. It failed at `cloning_source:pod_pending`: the `src` build pod sat unschedulable for an hour while the CI autoscaler could not scale up (`0/44 nodes are available`, all taints and affinity). Pure CI capacity, nothing to do with this change.

raelga:
/test e2e-parallel

Retesting: 2nd occurrence, same signature as the previous failure - cluster teardown timeout in canadacentral, not a new test failure. The v20260901preview cluster-health spec passed but cleanup hit:

  failed to cleanup resource group: at least one hcp cluster failed to delete ... RESPONSE 200 OK, ERROR CODE: InternalServerError

This is an RP-side delete/cleanup issue unrelated to this PR's mgmt-agent node-health diff. Other unrelated open PRs (#6792, #6791) are also currently failing e2e-parallel, consistent with a shared-CI infra episode rather than something introduced here.

raelga:
Not retesting immediately: 3rd occurrence, and now confirmed as a shared CI-subscription RoleAssignmentLimitExceeded quota-exhaustion episode affecting multiple concurrent PRs (#6769, #6776, #6786, #6781 all hit it in the same window during environment provisioning, no test suite ran). This is unrelated to this PR's diff. Backing off to periodic retest until the quota frees up.

raelga:
/hold cancel
/test e2e-parallel

Investigated the RoleAssignmentLimitExceeded failures across #6769/#6776/#6786/#6781 further. Checked the actual e2e-parallel job config in openshift/release (ALLOWED_SUBSCRIPTIONS): only 4 subscriptions are used (ARO HCP E2E Hosted Clusters (EA Subscription), ...2 (EA Subscription), Dev-02, Dev-03). None of them are currently anywhere near the 8000 role-assignment cap:

  EA Subscription:  1669/8000 (was 1720 an hour ago)
  EA Subscription 2: 622/8000 (was 907)
  Dev-02:           1134/8000 (was 1926)
  Dev-03:            399/8000 (was 571)

All four dropped substantially in the last hour (total -1300), consistent with a backlog of role assignments from earlier interrupted test batches finally being cleaned up by deprovisioning, not a permanently exhausted subscription. The failure itself happens in the shared environment-setup step (RG hcp-underlay-ci01-*), which only creates ~22 role assignments per run, so a single run's cost can't explain hitting the cap; whichever subscription got the failing lease must have had a transient backlog at that moment. This is infra-side and unrelated to this PR's diff. Cancelling the auto-hold and retesting once more given the numbers are recovering.

PR #6772, opened 2026-09-01
fix: handle empty timestamps in JUnit report generation

hlipsig:
When generating JUnit reports, skip steps that were never started or finished (empty StartedAt/FinishedAt timestamps) instead of attempting to parse them. This prevents "parsing time \"\" as \"2006-01-02T15:04:05Z07:00\"" errors for steps that were queued but not executed due to earlier failures or preemptions.

Steps are initialized with empty ExecutionInfo{} structs, and if a step is queued but never starts (or starts but is preempted before finishing), the timestamp fields remain empty strings. Previously this caused time.Parse() to fail with a cryptic error referencing Go's reference date format.

### What

Makes the source of the failure clear.

### Why

To reduce time to troubleshoot Prow failures. 

### Testing

This is a fix to a test, test ran locally without issues.

Doesn't fix an underlying CI issue, but will make it easier to spot the real problem.

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/tooling/templatize/pkg/pipeline/run.go
+++ b/tooling/templatize/pkg/pipeline/run.go
@@ -551,6 +551,16 @@ func runGraph(ctx context.Context, logger logr.Logger, executionGraph *graph.Gra
 			suite := suites.Suites[0]
 			for id, info := range timing {
 				thisLogger := logger.WithValues("id", id.String())
+				// Skip steps that were never started or finished (empty timestamps).
+				// This typically happens when a step was queued but not executed due to
+				// earlier failures or context cancellation. Without this check, time.Parse()
+				// would fail with: parsing time "" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "2006"
+				if info.StartedAt == "" || info.FinishedAt == "" {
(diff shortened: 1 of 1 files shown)


tooling/templatize/pkg/pipeline/run.go
```diff
@@ -551,6 +551,16 @@ func runGraph(ctx context.Context, logger logr.Logger, executionGraph *graph.Gra
 			suite := suites.Suites[0]
 			for id, info := range timing {
 				thisLogger := logger.WithValues("id", id.String())
+				// Skip steps that were never started or finished (empty timestamps).
+				// This typically happens when a step was queued but not executed due to
+				// earlier failures or context cancellation. Without this check, time.Parse()
+				// would fail with: parsing time "" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "2006"
+				if info.StartedAt == "" || info.FinishedAt == "" {
```
twolff-gh, line 558: What happens with startedAt set and finishedAt empty?

PR #6773, opened 2026-09-01
feat(hcpctl): add Kusto-backed oc-adm-inspect and wire it into must-g…

deads2k:
…ather

Adds `hcpctl oc-adm-inspect`, a Kusto-backed equivalent of `oc adm inspect` that reconstructs a namespace's state at a past point in time instead of querying a live cluster:

- Resources from kubernetesResourceSnapshots (latest snapshot per object at or before the timestamp), excluding objects that were actually deleted by then (a Delete event, or deletionTimestamp + deletionGracePeriodSeconds elapsed).
- Kubernetes events from kubernetesEvents for the namespace up to the timestamp.
- Container logs from containerLogs up to the timestamp, querying both the ServiceLogs database (management-cluster namespaces) and the HostedControlPlaneLogs database (hosted control plane pods) and merging them.

Behavior:
- Accepts namespaces via -n or positional ns/<name> args, a --timestamp, a --lookback window, and --management-cluster/--service-cluster (the telemetry `cluster` name). Namespaces like kube-system require specifying the cluster.
- If no cluster is given, discovers the clusters active in the window and lists them (split into management vs service) so the user can re-run.
- Inspecting a hosted-cluster namespace automatically adds the paired control-plane namespace and vice versa, derived from HostedCluster snapshots.
- Output mirrors oc adm inspect: namespaces/<ns>/<group>/<kind>.yaml, pods/<pod>/<pod>.yaml, pods/<pod>/<container>/<container>/logs/current.log, and core/events.yaml.

must-gather `query` now runs oc-adm-inspect for every HCP cluster discovered in the resource group, inspecting each cluster's hosted-cluster and control-plane namespaces on the management cluster hosting it, under <output>/oc-adm-inspect (skippable with --skip-oc-adm-inspect).

/hold  for me to read and review.

### DIFF
--- a/tooling/hcpctl/cmd/must-gather/query_cmd.go
+++ b/tooling/hcpctl/cmd/must-gather/query_cmd.go
@@ -16,11 +16,15 @@ package mustgather
 
 import (
 	"context"
+	"errors"
+	"path/filepath"
 
 	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 
+	"github.com/Azure/ARO-HCP/tooling/hcpctl/pkg/kusto"
 	"github.com/Azure/ARO-HCP/tooling/hcpctl/pkg/mustgather"
+	"github.com/Azure/ARO-HCP/tooling/hcpctl/pkg/ocadminspect"
 )
 
 func newQueryCommand() (*cobra.Command, error) {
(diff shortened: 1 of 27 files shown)


tooling/hcpctl/cmd/oc-adm-inspect/options.go
```diff
@@ -0,0 +1,225 @@
...
+
+// RawOptions is the unvalidated CLI configuration for oc-adm-inspect.
+type RawOptions struct {
+	Kusto             string
+	Region            string
+	SubscriptionID    string
+	ResourceGroup     string
+	Namespaces        []string
+	Timestamp         string
+	Lookback          time.Duration
```
deads2k: remove.  match the must-gather arguments exactly.

tooling/hcpctl/pkg/ocadminspect/writer.go
```diff
@@ -0,0 +1,232 @@
...
+			}
+		}
+		group := apiGroup(resource.APIVersion)
+		fileKey := filepath.Join(group, resourcePlural(resource.Kind)+".yaml")
+		byFile[fileKey] = append(byFile[fileKey], resource)
+	}
+
+	for fileKey, group := range byFile {
+		sort.Slice(group, func(a, b int) bool { return group[a].Name < group[b].Name })
+		if err := w.writeListFile(filepath.Join(w.namespaceDir(namespace), fileKey), group); err != nil {
```
rogbas, line 105: A malicious or accidental --namespace ../../directory could eventually write outside of the job folder with this approach. It's probably better to validate the namespace to match to a stricter regexp.

tooling/hcpctl/cmd/must-gather/query_cmd.go
```diff
@@ -64,5 +68,32 @@ func (opts *CompletedQueryOptions) RunQuery(ctx context.Context) error {
...
+
+	// For every HCP cluster discovered in the resource group, oc-adm-inspect the
+	// hosted-cluster namespace (which pulls in the paired control-plane namespace)
+	// on the management cluster that hosts it.
+	var inspectErr error
+	if !opts.SkipOCAdmInspect {
+		inspectErr = opts.runOCAdmInspect(ctx)
+	}
+
+	logger.Info("must-gather complete; content written", "output", opts.OutputPath)
```
rogbas, line 86: This log indicating success if being sent even when there are errors.

tooling/hcpctl/cmd/oc-adm-inspect/cmd.go
```diff
@@ -0,0 +1,138 @@
...
+
+	var b strings.Builder
+	fmt.Fprintf(&b, "clusters active around %s:\n", timestamp.Format(time.RFC3339))
+	if len(management) > 0 {
+		b.WriteString("\n  management clusters (pass --management-cluster <name>):\n")
+		for _, cluster := range management {
+			fmt.Fprintf(&b, "    %s\n", cluster)
+		}
+	}
+	if len(service) > 0 {
```
rogbas, line 131: What is the target use case for having this `--service-cluster` functionality?

I ask because there's no data collected from the Service Cluster in the default query mode of `oc-adm-inspect`.

deads2k: > What is the target use case for having this `--service-cluster` functionality?
> 
> I ask because there's no data collected from the Service Cluster in the default query mode of `oc-adm-inspect`.

didn't realize it was missing at the start.  I started this because I needed to fix a service cluster pod failing.

stevekuznetsov (commented):
/lgtm
/approve

PR #6779, opened 2026-09-02
feat: ARO-29395 - alert when validation controller reconciliation fails

patilsuraj767:
## Why

`BackendControllerRetryHotLoop` uses `workqueue_retries_total / workqueue_adds_total` as a proxy for controller failure. That ratio is wrong for validation controllers: they call `EnqueueAfter` for Failed/Unknown outcomes (which inflates **adds**), while a Failed validation returns `nil` and is **not** a reconcile error. Those controllers were excluded from the hot-loop alert ([commit](https://github.com/Azure/ARO-HCP/commit/2e60c8d4768bdef66f294deba3182a010f99c2f5)) until we had a signal that tracks actual `SyncOnce` errors.

## Summary

- Add `backend_controller_reconcile_errors_total` (incremented only when `SyncOnce` returns an error) and `backend_controller_reconcile_cooldown_skips_total` (incremented on validation cooldown no-ops so those retries don’t hide real failures).
- Add `BackendControllerReconcileErrorRate`: `errors / (reconciles − skips) > 0.5` for 10m, covering all backend controllers including validation.

## Alert PromQL (`BackendControllerReconcileErrorRate`)

The alert fires when more than half of **real** reconciles fail, sustained for 10 minutes:

```promql
errors / max(reconciles − cooldown_skips, 0) > 0.5
```

## Test plan

- [x] Promtool: fires at 80% error ratio; does not fire at 20%; does not fire when reconciles are high and errors are 0 (Failed + EnqueueAfter); fires when skips would otherwise hide a high error ratio
- [x] Go: reconcile errors increment only on SyncOnce error; cooldown skip increments skip counter


backend/alerts/backend-retryhotloop-prometheusRule.yaml
```diff
@@ -43,3 +45,52 @@ spec:
...
+                  sum by (controller, cluster) (
+                    max without(prometheus_replica) (
+                      rate(backend_controller_reconcile_total{namespace="aro-hcp"}[10m])
+                    )
+                  ) * 0
+                )
+              )
+            )
+          , 0)
+        ) > 0.5
```
patilsuraj767, line 88: ## Alert PromQL (`BackendControllerReconcileErrorRate`)
The alert fires when more than half of **real** reconciles fail, sustained for 10 minutes:
```promql
errors / max(reconciles − cooldown_skips, 0) > 0.5
```

### Numerator

```promql
sum by (controller, cluster) (
  max without(prometheus_replica) (
    rate(backend_controller_reconcile_errors_total{namespace="aro-hcp"}[10m])
  )
)
```

Rate of reconciles where `SyncOnce` returned a non-nil error (Cosmos Get/Replace failures, Unknown+`ReportError`, etc.). Failed/Passed/Skipped validations return nil and are not counted.

`max without(prometheus_replica)` dedups HA Prometheus replicas so the same scrape is not counted twice.

### Denominator

```promql
clamp_min(reconcile_total − (skips or reconcile_total * 0), 0)
```

1. **`rate(backend_controller_reconcile_total)`** — every `processNextWorkItem`, including validation cooldown no-ops.
2. **Subtract `rate(...cooldown_skips_total)`** — after a real error, `AddRateLimited` requeues quickly; those hits return nil on the cooldown gate. Without subtracting them, `errors/reconciles` sits around ~1/60 and would never exceed 0.5 even when every real attempt is failing.
3. **`or (rate(reconcile_total) * 0)`** — only validation controllers emit the skip metric. Prometheus `-` drops a series if the right-hand side is missing for that label set. A bare `0` / `vector(0)` has no `controller`/`cluster` labels, so it would not fill per-controller zeros. `reconcile_total * 0` copies those labels with value 0. `or` then means: use real skips if present, otherwise 0. Operation controllers become `errors / reconciles`; validation becomes `errors / (reconciles − skips)`.
4. **`clamp_min(..., 0)`** — if skip rate ever exceeds reconcile rate (reset/scrape noise), do not produce a negative denominator.

### Threshold

`> 0.5` for `10m` matches RetryHotLoop: majority of actual reconcile attempts failing, not a one-off blip.

backend/alerts/backend-retryhotloop-prometheusRule.yaml
miguelsorianod, line 30: This seems to keep the exclusions of those alerts for the validation controllers.
Isn't the idea to replace that across so we don't need exclusions particular to sets of controllers?

geoberle: what if we change the alert as this

from:

`workqueue_retries_total / workqueue_adds_total`

to:

`(workqueue_retries_total - backend_controller_reconcile_cooldown_skips_total) / workqueue_adds_total`

it lowers the "retries" by the amount of letimiate adds to the workqueue. this should workd for all controllers without the need of a name exclusion

patilsuraj767: Subtracting cooldown skips does not let us drop the name exclusion. The two counters measure different events, so a validation controller that is retrying Failed/Unknown outcomes as designed still looks like a hot loop.

`workqueue_retries_total` increments on every workqueue AddAfter and AddRateLimited. Validation controllers use EnqueueAfter (AddAfter) for Failed/Unknown results, so those designed ~60s requeues already raise `workqueue_retries_total`. 

Cooldown skips are separate. They increment only when a key is processed while the retry gate is still closed, that path then calls `EnqueueAfter` again, which adds another retry. Subtracting `backend_controller_reconcile_cooldown_skips_total` removes only those arrived too early extras. The designed Failed/Unknown requeues stay in the numerator, so the alert still fires and the name exclusion is still required.

BackendControllerRetryHotLoop is meant to catch error-driven AddRateLimited retries, not scheduled validation requeues.

Let me try to explain this with a example, Suppose a user creates an HCP cluster and a validation controller checks that some Azure resource exists. While that resource is missing, the validation returns Failed and calls EnqueueAfter 60s. In this case SyncOnce returns nil, **not a controller error**. 
That delayed requeue repeats until the user creates the resource and the check passes.
That is not a hot loop from controller failure. It is the intended wait-and-retry. 
So even if we subtract the cooldown skips there is chance of triggering the hotloop alert.

geoberle: i see.
we can merge the PR like this but we need to resolve the situation here. in a followup we can remove the alert and go for the new one but we need to make sure all controllers expose the required metrics. not everything is a genericwatchingcontroller.

PR #6780, opened 2026-09-02
Chore trigger e2e with CS local image that contains readdesire watching on hosted cluster done inside clusters service

machi1990:
NOT FOR MERGE; this is an attempt to run all e2e against the CS local changes.


machi1990:
Okay to close. It has done it's job and good progress on the cluster front.  A new one for node pool will be opened.

PR #6782, opened 2026-09-02
feat(mgmt-agent): detect nodes that never reached Ready (AROSLSRE-1983)

raelga:
Jira: [AROSLSRE-1983](https://redhat.atlassian.net/browse/AROSLSRE-1983)

### What

Adds a `never-ready` detector for a node that registers and never reaches Ready.

Its evidence is the Node object alone: the Ready condition has never been True, and its first transition coincides with the node's creation. A node that reached Ready at any point has a strictly later transition, which is what separates "never started" from "started and later failed".

Three supporting changes:

- An optional `nodeDetector` sub-interface, so a node-state detector fits without widening `Evaluate` for every existing detector.
- `Decide` now runs node-state detectors on the NotReady path and the pod/event detectors on the Ready path. The Ready path is unchanged.
- `Snapshot.Detail`, an escape hatch for the reason annotation.

### Why

`Decide` skips any NotReady node:

```go
// Node-Ready precondition: a NotReady node is left to node lifecycle.
if !isNodeReady(node) {
    return DecisionUnknown, Snapshot{}
}
```

Correct for a node that was Ready and dropped out. Wrong for one that never got there, because nothing in node lifecycle rescues it.

INT node `aks-userswft3-65765115-vmss00000g` (int-westus3, [AROSLSRE-1976](https://redhat.atlassian.net/browse/AROSLSRE-1976)) failed provisioning with a terminal `OSProvisioningClientError`, registered, never came up, and sat NotReady for 22 hours until a human cordoned it. From its captured `node.yaml`:

```text
creationTimestamp:          2026-08-31T21:32:14Z
Ready.lastTransitionTime:   2026-08-31T21:32:14Z
Ready.status:               False
Ready.reason:               KubeletNotReady
```

The condition was written once at registration and never moved, so the two timestamps are identical.

**The dwell is 30 minutes, set from data.** Across 1,119 dev nodes over 14 days, time from creation to first `Ready=True` was p50 1.0m, p90 1.3m, p99 3.1m, max 3.4m. Every node that was going to come up did so inside 3.4 minutes, so 30 minutes is about nine times the worst legitimate bootstrap, and still 44 times faster than the 22 hours we actually took.

**Why `Snapshot.Detail` is needed:** `labeler.go` fills the reason annotation from `ReasonString()`, which renders pod counts. A born-broken node would otherwise be annotated "0 pods stuck past dwell (0 stuck total), no recent success in 0s window". Detectors that leave `Detail` empty render byte-identically to before, which is pinned by a test.

**Why generic rather than IMDS-specific:** dead IMDS, a failed disk, a kubelet certificate problem and a failed image pull are all born broken and all remedied by reimaging, and one detector is one fault with one remedy. The cause goes in the signature annotation instead, taken from the Ready condition's reason. A cause-specific detector would also have to read azure-cns container logs, which this controller never does.

### Testing

`go build`, `go vet`, `gofmt` and `go test ./pkg/controller/nodehealth/...` all clean. Every pre-existing test passes untouched.

New tests cover the dwell boundary (including exactly at the dwell), the born-broken discriminator, a node that was Ready and dropped out, the registration tolerance either side, unusable evidence (no Ready condition, missing timestamps, clock skew), an empty Ready reason, ownership of a non-SWIFT node, and that the Ready path still belongs to `swift-vf-teardown`.

`TestProductionBornBrokenNodeFires` replays the captured INT node with its verbatim timestamps and asserts it fires at the dwell and not a second earlier.

I also mutation-checked the tests: removing the born-broken discriminator fails 4 of them and zeroing the dwell fails 2, so they are not passing vacuously.

**This cannot cause disruptive remediation.** The controller stops at labelling; mitigation of a labelled node is owned by a separate controller that is not shipped.

**Ownership is deliberately narrower than the fault.** The detector reuses `isSwiftV2Node` even though a born-broken node is not a SWIFT problem. `AnyApplies` runs ahead of the Ready gate and decides which nodes we hold labels on at all, so a detector applying to every node would make `DecisionNotApplicable` unreachable, and that is the only thing that retires a stale label. I verified the captured INT node did carry `kubernetes.azure.com/podnetwork-swiftv2-enabled=true`, so the scope is correct rather than lucky. Widening it is a separate decision.

**A silent node still gets evaluated:** a born-broken node produces no pods or events, but `resyncAll` re-enqueues every candidate node every 30 seconds, so it fires at the dwell rather than waiting for unrelated traffic.

Two open questions I would rather surface than bury:

1. Should AKS node auto-repair have owned this? It targets nodes that were healthy and went NotReady, so a never-provisioned instance plausibly falls outside it, which is the gap this fills. If it should have caught it, the better fix is an AKS bug and this becomes a stopgap.
2. The dwell rests on dev data, because prod does not snapshot Node objects today. Worth re-checking once it does.

This is independent of [#6769](https://github.com/Azure/ARO-HCP/pull/6769) and [#6776](https://github.com/Azure/ARO-HCP/pull/6776), which change detector thresholds rather than the Ready gate. I trial-merged all three: they auto-merge with no conflicts and the full suite passes on the combined tree.

### PR Checklist

- [x] Tests added, including a production-evidence test
- [x] `go build`, `go vet`, `gofmt` clean
- [x] Jira issue linked

### DIFF
--- a/mgmt-agent/pkg/controller/nodehealth/controller.go
+++ b/mgmt-agent/pkg/controller/nodehealth/controller.go
@@ -389,10 +389,15 @@ func (c *Controller) syncHandler(ctx context.Context, name string) error {
 		}
 		if changed {
 			detectionsTotal.WithLabelValues(snap.DetectorName, snap.MatchedSignature).Inc()
-			logger.Info("detector fired; node labeled wedged",
-				"detector", snap.DetectorName,
-				"signature", snap.MatchedSignature,
-				"evidence", snap.ReasonString())
+			// Pod evidence is logged as structured fields so it stays filterable.
+			// A node detector has no pod counts, so it logs its summary instead.
+			kv := []any{"detector", snap.DetectorName, "signature", snap.MatchedSignature}
+			if snap.Pods != nil {
+				kv = append(kv, "failures", snap.Pods.FailureCount, "recentSuccess", snap.Pods.RecentSuccess)
+			} else {
+				kv = append(kv, "evidence", snap.Detail)
+			}
+			logger.Info("detector fired; node labeled wedged", kv...)
 		}
 		// Mitigation of a labeled node (cordon, taint, evict, delete) is owned
 		// by a separate controller; this controller stops at labeling.
--- a/mgmt-agent/pkg/controller/nodehealth/controller.go
+++ b/mgmt-agent/pkg/controller/nodehealth/controller.go
@@ -392,8 +392,7 @@ func (c *Controller) syncHandler(ctx context.Context, name string) error {
 			logger.Info("detector fired; node labeled wedged",
 				"detector", snap.DetectorName,
 				"signature", snap.MatchedSignature,
-				"failures", snap.FailureCount,
-				"recentSuccess", snap.RecentSuccess)
+				"evidence", snap.ReasonString())
 		}
 		// Mitigation of a labeled node (cordon, taint, evict, delete) is owned
 		// by a separate controller; this controller stops at labeling.
--- a/mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
+++ b/mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
@@ -95,13 +95,9 @@ type Detector interface {
 }
 
 // nodeDetector is a Detector whose evidence is the Node object itself rather
-// than the Pods and Events on it. Decide evaluates these ahead of the Ready
-// gate, because their whole subject is a node that never became Ready and so
-// has no meaningful pod population to read.
-//
-// It is an optional extension rather than a change to Detector: a pod/event
-// detector has no node-only evaluation to offer, and widening Evaluate for the
-// sake of one family would touch every existing detector and its tests.
+// than the Pods and Events on it. Decide evaluates these on the NotReady path,
+// because their whole subject is a node that never became Ready and so has no
+// meaningful pod population to read.
 type nodeDetector interface {
 	Detector
 	// EvaluateNode reads the detector's signals from the Node alone. It performs
@@ -169,8 +165,8 @@ type Snapshot struct {
 	// Detail, when set, replaces the pod-centric summary ReasonString renders. A
 	// detector whose evidence is not pod counts sets this so the reason
 	// annotation describes what it actually observed, instead of reporting zero
-	// pods stuck in a zero-length window. Leaving it empty keeps the pod-centric
-	// summary, so every existing detector is unaffected.
+	// pods stuck in a zero-length window. When it is empty, ReasonString renders
+	// the pod-centric summary.
 	Detail string
 }
 
@@ -205,8 +201,8 @@ func Decide(node *corev1.Node, events []*corev1.Event, pods []*corev1.Pod, now t
 		return DecisionNotApplicable, Snapshot{}
 	}
 	// Node-Ready precondition. A node that was Ready and dropped out (reboot,
-	// upgrade, drain) is left to node lifecycle, which is what the original gate
-	// was for. A node that never reached Ready is a different case: nothing in
+	// upgrade, drain) is left to node lifecycle, which rescues it. A node that
+	// never reached Ready is a different case: nothing in
 	// node lifecycle rescues it, so it is ours. Only node-state detectors run on
 	// this path, because a node that never started has no pod population worth
 	// reading, and each of them rejects a node that was once Ready on its own
--- a/mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
+++ b/mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
@@ -63,11 +63,11 @@ func (d Decision) String() string {
 	}
 }
 
-// Detector is one fault family. Each concrete detector lives in its own file and
-// contributes only its specifics (applicability, signals, thresholds); the
-// evaluation primitives are shared through a common base (see signatureDetector).
-// Decide iterates the registry and never depends on a concrete type, so adding a
-// family is adding a Detector, not editing Decide.
+// Detector is what every fault family has in common: an identity and a scope.
+// It carries no evaluation method, because the two kinds of detector do not read
+// the same evidence and cannot share one. A concrete detector implements this
+// plus exactly one of PodDetector or NodeDetector, which is what puts it on the
+// matching path in Decide.
 type Detector interface {
 	// Name is a stable identifier used in logs, events, metrics, and the
 	// detector annotation.
@@ -82,6 +82,12 @@ type Detector interface {
 	// detector, independent of any node, so callers that only need the window do
 	// not have to evaluate the detector to learn it.
 	Window() time.Duration
+}
+
+// PodDetector reads a node's health from the Pods and Events held for it. It
+// runs only on a Ready node, where a pod population exists to be read.
+type PodDetector interface {
+	Detector
 	// Evaluate reads the detector's signals for the node from the Events and Pods
 	// currently held for it. Both the failure and the success signal come out of
 	// the Pods passed in, so the result depends on nothing but what a LIST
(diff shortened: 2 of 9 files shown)


raelga:
Force-pushed `e7454fa7e` to `fd875758b`. Commit message only, the tree is byte-identical: I dropped a `Co-authored-by: Copilot` trailer I should not have added, since I do not use that trailer.

No code change, so nothing about the review changes. Sorry for the CI restart.

```text
git diff e7454fa7e fd875758b   # empty
```

mgmt-agent/pkg/controller/nodehealth/detectors/never_ready.go
```diff
@@ -0,0 +1,176 @@
...
+		// The node was Ready at some point and lost it later. Node lifecycle owns
+		// that case, exactly as it did before this detector existed.
+		return snap
+	}
+
+	snap.StuckSince = created
+	// The node itself is the single subject, so both counts are one. They are
+	// reported so the metric and the logs read consistently with other detectors.
+	snap.FailureCount = 1
+	snap.SustainedCount = 1
```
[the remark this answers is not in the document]
raelga: @copilot Agreed, changed in [6b83915e8](https://github.com/Azure/ARO-HCP/commit/6b83915e8). Both counts stay zero now and I added an assertion in `TestProductionBornBrokenNodeFires` so it can't drift back. `MeetsThreshold` only keys on `StuckSince`, so no decision changes, and I verified that by setting `FailureCount = 1` again and watching the new assertion fail.

mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
```diff
@@ -151,10 +166,19 @@ type Snapshot struct {
 	// on DetectorName, not on this: one detector is one fault with one remedy,
 	// and branching on the signature would require knowing detector internals.
 	MatchedSignature string
+	// Detail, when set, replaces the pod-centric summary ReasonString renders. A
+	// detector whose evidence is not pod counts sets this so the reason
+	// annotation describes what it actually observed, instead of reporting zero
```
stevekuznetsov: You should make sure comments are evergreen. AI loves to make comments comparing to previous state, or designs considered but not implemented. None of those will be live in the code, so referring to them is nonsense.

mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
```diff
@@ -180,8 +204,27 @@ func Decide(node *corev1.Node, events []*corev1.Event, pods []*corev1.Pod, now t
 	if !AnyApplies(node) {
 		return DecisionNotApplicable, Snapshot{}
 	}
-	// Node-Ready precondition: a NotReady node is left to node lifecycle.
+	// Node-Ready precondition. A node that was Ready and dropped out (reboot,
```
stevekuznetsov, line 231: again a comment for 'original gate', delete

mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
```diff
@@ -180,8 +204,27 @@ func Decide(node *corev1.Node, events []*corev1.Event, pods []*corev1.Pod, now t
...
+	// upgrade, drain) is left to node lifecycle, which is what the original gate
+	// was for. A node that never reached Ready is a different case: nothing in
+	// node lifecycle rescues it, so it is ours. Only node-state detectors run on
```
stevekuznetsov: do not force two distinct and non-overlapping bits of logic through one code path, making claims about 'only ... run on this path' and hoping that you get the intent you want. it *looks* like you need to break the Detector abstraction or find a way to have it return the decision itself, etc. clearly the two we have are not isomorphic so hiding them behind one abstraction is wrong

mgmt-agent/pkg/controller/nodehealth/detectors/never_ready.go
```diff
@@ -0,0 +1,177 @@
...
+	if cond == nil || cond.Status == corev1.ConditionTrue {
+		return snap
+	}
+
+	created := node.CreationTimestamp.Time
+	transitioned := cond.LastTransitionTime.Time
+	if created.IsZero() || transitioned.IsZero() {
+		// Without both timestamps the discriminator cannot be evaluated, and
+		// guessing would risk labeling a node that simply came up before this
+		// controller started watching.
```
stevekuznetsov: I don't think this is true, since transitions before the controller watched would still have their info in the Node's database representation. Returning here if the timestamps are not both set is sound, though.

mgmt-agent/pkg/controller/nodehealth/detectors/never_ready.go
```diff
@@ -0,0 +1,177 @@
...
+		return snap
+	}
+	if transitioned.After(created.Add(neverReadyTolerance)) {
+		// The node was Ready at some point and lost it later. Node lifecycle owns
+		// that case, exactly as it did before this detector existed.
+		return snap
+	}
+
+	snap.StuckSince = created
+	// FailureCount and SustainedCount stay zero: they are defined as counts of
```
stevekuznetsov: change your data type

mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
raelga: Totally agree :) these slipped through my own review. And there were four of them, not two: same thing in the `Snapshot.Detail` doc, the tolerance branch in `never_ready.go`, and the `nodeDetector` doc. All described the diff instead of the code.

Rewritten, comments only. Fixed in 5689b57f7.

mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
raelga: Rewrote it rather than deleting it, since the split between the two paths still needs explaining. Happy to drop it if you'd rather.

Fixed in 5689b57f7.

mgmt-agent/pkg/controller/nodehealth/detectors/never_ready.go
raelga: Yeah, true. The timestamps are on the Node object, so even if we start watching late we still see them. This one is really just about missing data, so I kept the early return and reworded the comment.

Fixed in cd0471c39.

mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
raelga: Split it. `Detector` is now just identity and scope, and the evaluation lives in two interfaces: `PodDetector` reads the Pods and Events, `NodeDetector` reads the Node. `EvaluateNode` returns the Decision itself, so there is no snapshot lying around that the other path has to be trusted not to act on.

The registry is split the same way, `podRegistry` and `nodeRegistry`, so a detector can only run on the path its evidence exists on. That is a compile-time property now instead of a comment. Added a test that fails if the node detector ever grows an `Evaluate` method.

Fixed in c7a63cf0a.

mgmt-agent/pkg/controller/nodehealth/detectors/never_ready.go
raelga: Done. The pod counts moved into a `PodEvidence` struct and `Snapshot.Pods` is a pointer, so a detector that reads no pods leaves it nil instead of reporting zeroes that look like "no pods were stuck". `ReasonString` branches on nil now.

Fixed in c7a63cf0a.

mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
```diff
@@ -151,10 +162,19 @@ type Snapshot struct {
 	// on DetectorName, not on this: one detector is one fault with one remedy,
 	// and branching on the signature would require knowing detector internals.
 	MatchedSignature string
+	// Detail, when set, replaces the pod-centric summary ReasonString renders. A
+	// detector whose evidence is not pod counts sets this so the reason
+	// annotation describes what it actually observed, instead of reporting zero
+	// pods stuck in a zero-length window. When it is empty, ReasonString renders
+	// the pod-centric summary.
```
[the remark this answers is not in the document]
raelga: @copilot Rewrote the `StuckSince` doc to cover both cases: for a pod detector it is the oldest stuck-pod transition, for a node detector it is when the node's own state started. Both come from durable state.

Fixed in c7a63cf0a. No further action or commit needed from you, this is resolved.

mgmt-agent/pkg/controller/nodehealth/detectors/never_ready.go
```diff
@@ -0,0 +1,177 @@
...
+	if transitioned.Before(created) {
+		// A Ready transition predating the object cannot happen; treat it as bad
+		// data rather than evidence.
+		return snap
+	}
+	if transitioned.After(created.Add(neverReadyTolerance)) {
+		// The node was Ready at some point and lost it later. Node lifecycle owns
+		// that case.
+		return snap
+	}
```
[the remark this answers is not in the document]
raelga: @copilot That's true, the branch only proves the Ready condition changed after creation, not that the node was ever Ready. Unknown to False would land here too. Reworded it to say this is not the born-broken shape, without claiming the node was once Ready.

Fixed in c7a63cf0a. No further action or commit needed from you, this is resolved.

mgmt-agent/pkg/controller/nodehealth/controller.go
```diff
@@ -392,8 +392,7 @@ func (c *Controller) syncHandler(ctx context.Context, name string) error {
 			logger.Info("detector fired; node labeled wedged",
 				"detector", snap.DetectorName,
 				"signature", snap.MatchedSignature,
-				"failures", snap.FailureCount,
-				"recentSuccess", snap.RecentSuccess)
+				"evidence", snap.ReasonString())
```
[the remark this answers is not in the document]
raelga: @copilot That's fair, I dropped the filterable fields. Put them back: when `Pods` is set the log carries `failures` and `recentSuccess` as structured fields, and a node detector logs `evidence` with its detail string instead.

Fixed in 106f38b36. No further action or commit needed from you, this is resolved.

mgmt-agent/pkg/controller/nodehealth/detectors/decide.go
raelga: Dropped it. With the registry split the comment was explaining something the types now enforce, so there was nothing left worth saying. What is left is just why the gate exists at all.

Fixed in 5b50dc744.

raelga:
Rebased on main now that [#6769](https://github.com/Azure/ARO-HCP/pull/6769) and [#6776](https://github.com/Azure/ARO-HCP/pull/6776) merged, which dropped the lgtm.

The only conflict was in `signature_detector.go`, where [#6776](https://github.com/Azure/ARO-HCP/pull/6776) added the `successScope` check on the same lines this PR moved `RecentSuccess` under `Snapshot.Pods`. I kept both: the scope check is unchanged, it just writes to `snap.Pods.RecentSuccess`. The follow-up commit is the same rename in two test files that were still asserting on the old field path. No behaviour change on top of what you already reviewed.

@inbharajmani mind re-applying the `/lgtm` once CI is green?

PR #6798, opened 2026-09-02
test: add e2e coverage for HostedCluster deletion after etcd KMS Key Vault removal

cssjr:
[AROSLSRE-1709](https://redhat.atlassian.net/browse/AROSLSRE-1709)

### What

Adds a new permanent e2e regression test, `test/e2e/cluster_delete_kms_vault.go`:
creates an HCP cluster + node pool, deletes **only the etcd KMS Key Vault**
(not the customer resource group), then deletes the HCP cluster and asserts
it completes automatically within `HCPClusterDeletionTimeout` (25 min)
without manual finalizer removal.

Also adds a `GetARMKeyVaultClientFactory`/`OrDie` accessor to the test
framework (`test/util/framework/per_test_framework.go`), following the
existing ARM client-factory pattern (mirrors `GetARMComputeClientFactory`),
since no management-plane Key Vault client was previously exposed on the test
context. The existing `purgeDeletedKeyVaultsInResourceGroup` helper is
refactored to reuse this new accessor instead of constructing its own client
inline.

`nonlocal-e2e-specs.txt` regenerated via `make record-nonlocal-e2e` to include
the new spec.

### Why

AROSLSRE-1167's RCA confirmed a real `HostedCluster` deletion deadlock on
`prod-uksouth-mgmt-1`: the existing `cluster_delete_cx_rg` test deleted a
customer resource group (as designed) which also removed the etcd KMS Key
Vault living inside it. With the guest API permanently unavailable, the CAPZ
`availability-prober` blocked the `capi-provider` manager from ever starting,
so `AzureMachine` finalizers could never clear and deletion hung for weeks
until an SRE manually removed finalizers on 2026-06-11.

`AROSLSRE-1709` tracks the preventive HyperShift fix
(`openshift/hypershift#9403`, `DeleteOrphanedMachines`/
`capiProviderUnavailable`) and its acceptance criteria explicitly require
verifying this scenario recovers automatically without manual finalizer
removal, **while keeping `cluster_delete_cx_rg`'s RG-deletion behavior
intact** (it must keep validating its own, different intended behavior).
This PR closes that specific coverage gap by isolating just the confirmed
etcd KMS Key Vault trigger, rather than the whole customer RG.

### Testing

- [x] E2E tests
- Build/`go vet`/`golangci-lint` all pass locally for the changed packages.
- **Not yet verified end-to-end against the fix.** I ran this twice against a
  personal dev env with the patched `openshift/hypershift#9403` HyperShift
  operator image, but both runs failed during ordinary cluster creation
  (before ever reaching the Key Vault deletion step) due to the shared EA
  subscription's public IPv4 quota being fully exhausted (1000/1000) by
  concurrent usage from other teams — an unrelated, external environment
  constraint, not a defect in this test or the fix. Opening as **Draft**
  specifically to verify via `e2e-parallel` in CI, which should confirm:
  1. the test passes with the fix present, and
  2. it fails clearly (hangs to the 25-min `HCPClusterDeletionTimeout`, per
     the fix's own doc comment about what it prevents) without it — happy to
     paste CI logs here once available, including the operator's
     `"orphaning azuremachine stuck in deletion"` / `"capi-provider
     unavailable"` log line as direct proof the new code path fires.

- The fix intentionally leaves real Azure VM/NIC infrastructure in the
  managed resource group intact when it orphans a stuck `AzureMachine` (see
  the fix's own doc comment) — the standard per-test cleanup won't
  force-delete a deny-assignment-locked managed RG full of live VMs, so this
  test explicitly force-deletes the managed RG itself in its final step
  rather than relying on `DeferCleanup`.
- Per [AROSLSRE-1709](https://redhat.atlassian.net/browse/AROSLSRE-1709)'s explicit non-goals, `cluster_delete_cx_rg.go` is
  untouched — this is additive coverage, not a replacement.

### DIFF
--- a/test/e2e/cluster_delete_kms_vault.go
+++ b/test/e2e/cluster_delete_kms_vault.go
@@ -1,0 +1,164 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+	"github.com/Azure/ARO-HCP/test/util/verifiers"
+)
+
+// This test is a regression test for AROSLSRE-1167 / AROSLSRE-1709: a
+// HostedCluster deletion deadlocked on prod-uksouth-mgmt-1 after the
+// cluster_delete_cx_rg test (see cluster_delete_cx_rg.go) deleted the
+// customer resource group, which also removed the etcd KMS Key Vault living
+// inside it. With the guest API permanently unavailable, the CAPZ
+// availability-prober blocked the capi-provider manager from ever starting,
+// so AzureMachine finalizers could never clear and deletion hung until an SRE
+// manually removed the finalizers. cluster_delete_cx_rg must keep validating
+// the intended RG-deletion behavior, so this test isolates just the
+// confirmed root cause (the KMS Key Vault disappearing) without touching the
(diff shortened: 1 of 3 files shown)


test/e2e/cluster_delete_kms_vault.go
```diff
@@ -0,0 +1,176 @@
...
+		// (in dev) selects on ARO-HCP-RP-API-Compatible || Development-Only. This
+		// test uses ARM templates directly (like cluster_delete_cx_rg.go), so it
+		// is not RP-API-compatible; Development-Only is added solely to get a CI
+		// run on this draft PR since local personal-dev-env validation hit a
+		// shared-subscription public IP quota exhaustion. Removing this label
+		// restores eligibility for the broader integration/stage/prod-e2e-parallel
+		// suites (which exclude Development-Only), matching cluster_delete_cx_rg.go
+		// and cluster_delete_missing_identities.go's unlabeled state — the intended
+		// long-term home for this test, since the original incident was in prod.
+		labels.DevelopmentOnly,
```
cssjr, line 62: Remove before merging (used for testing only)

PR #6803, opened 2026-09-02
fix(validation): Validate an exact cluster version is specified when passing nightly channels

bennerv:
### What

Requires an exact cluster version when nightly channel is specified for clusters.

### Why

https://github.com/Azure/ARO-HCP/pull/6538/changes#diff-c85735cdffdb78eb1070b2d4cf9b78a882cda824695fc27719cd73a04b8ecf46R209-R213 results in a failure to resolve the control plane version if the nightly channel is used and we don't specify the exact version we want to install.  

### Testing

Unit tests and integration tests. 

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/internal/admission/admit_cluster.go
+++ b/internal/admission/admit_cluster.go
@@ -687,7 +687,7 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
 	}
 	// on nightlies and candidates, we don't wait for the target version to show up. We trust these people to
 	// know what they're doing.
-	if newObj.ChannelGroup == "nightly" || newObj.ChannelGroup == "candidate" {
+	if newObj.ChannelGroup == metadataapi.ChannelGroupNightly || newObj.ChannelGroup == metadataapi.ChannelGroupCandidate {
 		return nil
 	}
 
--- a/internal/api/metadataapi/utils.go
+++ b/internal/api/metadataapi/utils.go
@@ -53,11 +53,24 @@ var AllowControlPlaneNodePoolMajorVersionSkew = map[string][]string{
 	"4.23": {"5.1", "5.2"},
 }
 
+// OpenShift version update channel groups.
+const (
+	ChannelGroupStable    = "stable"
+	ChannelGroupFast      = "fast"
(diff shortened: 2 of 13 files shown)


machi1990 (approved):
/lgtm

internal/admission/admit_cluster.go
```diff
@@ -182,6 +190,26 @@ func mutateClusterControlPlaneExactVersion(_ context.Context, admissionContext *
...
+	// before validation and has already reduced a full version.id to its
+	// major.minor release line by this point, so static validation can no longer
+	// tell whether the customer supplied a full version. Reject unless the logic
+	// above established an exact pin (from a full version.id or the exact-version
+	// tag). versionID is the original, pre-reduction value.
+	if newObj.CustomerProperties.Version.ChannelGroup == metadataapi.ChannelGroupNightly &&
+		newObj.ServiceProviderProperties.ExperimentalFeatures.ControlPlaneExactVersion == nil {
+		return field.ErrorList{field.Invalid(
+			field.NewPath("customerProperties", "version", "id"), versionID,
+			"must be specified as MAJOR.MINOR.PATCH (optionally with a pre-release, e.g. a nightly build suffix) when channelGroup is \"nightly\"",
```
ahitacat: CS only fetches nightly images with the pre-release, the builds that only have X.Y.Z are ommited as they will be in candidate (and we can just test the latest pre-release). I will rephrase this and ask for the pre-release so we don't induce to use X.Y.Z version and then fail in CS step.

deads2k: Cluster service semantics no longer drive decisions here.

bennerv:
/hold 

discussion on admission vs static validation

bennerv:
/hold cancel

Moved to static validation after discussion.  Reason is we reuse it for cosmosdb writes in the future.

bennerv:
/hold cancel

follow-up for only supporting exact version via AFEC to come.  This can merge as is though with the caveot that if a customer PATCHES a nightly without the exact version underneath `version.id`, then it'll fail on the version static validation.

PR #6804, opened 2026-09-02
Revert: role-assignment cluster-create gate (AROSLSRE-1996)

raelga:
Jira: [AROSLSRE-1996](https://redhat.atlassian.net/browse/AROSLSRE-1996)

### What
Reverts #6681, which added an unbounded (no-timeout) gate in `operation_cluster_create.go` blocking cluster-create completion until Azure role assignments for the cluster's managed identities are observed as present.

### Why
The gate has no timeout, so any RBAC-visibility delay blocks cluster-create indefinitely instead of failing with a clear, timed-out error. This isn't hypothetical: Key Vault RBAC data-plane propagation lag ([AROSLSRE-1997](https://redhat.atlassian.net/browse/AROSLSRE-1997)) already causes this class of delay in roughly 8% of clusters even on healthy INT e2e runs. The original trigger for adding this gate's blocking behavior to fail badly (AROSLSRE-1994, recreated INT service principals with zero role assignments) is already fixed, but the unbounded-wait design remains a standing risk for any future RBAC-visibility delay.

### Testing
- `go build ./backend/...` passes clean on the reverted code.
- `go vet ./backend/...` passes clean.
- `go test ./backend/pkg/controllers/cluster/operations/...` passes.

This is a straight revert (`git revert -m 1`) of the merge commit, no other changes. Please weigh whether reverting outright vs. adding a timeout/clearer error to the existing gate is the better fix, happy to change approach.

### PR Checklist
- [x] Tests pass
- [x] Docs updated (n/a, revert)


gmfrasca (commented):
/lgtm

raelga:
Closing without merging. Dug further and found the premise doesn't hold: the role-assignments gate already shares the same umbrella `CreateOperationCompletionDeadline` (default 60m, e2e tests use 30m via the `ClusterCreationTimeout` tag) as every other check in `determineOperationState` (hostedCluster, clusterService, servingCABundle). None of those have individual timeouts either, they all rely on this same deadline. So this gate isn't an indefinite/unbounded block, it behaves exactly like the rest.

There's also a dedicated e2e test, `cluster_delayed_role_assignments.go`, that deliberately exercises this exact scenario (role assignments added after cluster creation) with an extended timeout, because a delay of up to ~15m is expected, known behavior.

The original [AROSLSRE-1994](https://redhat.atlassian.net/browse/AROSLSRE-1994) SP-recreation RBAC gap that triggered this is already fixed. So there's nothing currently broken for this PR to fix, closing it rather than merging a revert that would remove a working safety check.

PR #6807, opened 2026-09-03
test: add private/public visibility E2E tests for ARO-HCP clusters

jiezhao16:
## Summary
- Add 3 new E2E tests covering ARO-HCP cluster visibility combinations that were previously untested:
  1. **Fully private cluster** (`api.visibility=Private`, `ingress.type=Private`) — verifies KAS and ingress reachable only from within VNet
  2. **PublicAndPrivate KAS** (default Public visibility) — verifies both public (shared ingress) and private (Swift) KAS access paths
  3. **Fully private with KeyVault** (all three dimensions Private) — verifies no interaction effects between private settings

- Tests 1, 3 use `v20260630preview` API; Test 2 uses `v20251223preview`

Ref: [CNTRLPLANE-4047](https://redhat.atlassian.net/browse/CNTRLPLANE-4047)

### Verification approaches

**KAS reachability (private clusters):** Verified using `kubectl version` from a VM inside the customer VNet via the internal LB. This uses the unauthenticated `/version` endpoint, proving the private KAS network path (VM → customer subnet → internal LB → Swift → KAS pods) is functional without depending on authenticated access.

**KAS reachability (public clusters):** Verified using `TestHTTPSConnectivity` against the `/healthz` endpoint from outside the VNet, proving the shared ingress path is functional.

**kubelet→KAS path:** Verified by `kubectl get nodes` from the VM, checking that all worker nodes are Ready. Worker nodes can only reach Ready by registering with KAS over Swift, so Ready nodes prove this direction of the Swift path works.

**KAS→kubelet path:** Verified by `kubectl logs` on a running `openshift-dns` pod from the VM. kubectl logs proxies the request through KAS to the kubelet — a separate Swift connection from the kubelet→KAS registration path — proving full bidirectional Swift coverage. openshift-dns DaemonSet pods are used because they run on every worker node and are guaranteed present once nodes are Ready.

**Ingress (private clusters):** Verified by deploying a sample app via `kubectl apply` from the VM, then curling the route host from the VM (should succeed) and from outside the VNet (should fail). This tests the full IngressController/router-default path with internal load balancer scope.

**Ingress (public clusters):** Verified by checking the OpenShift console URL from outside the VNet. The console is deployed by default and served through the default ingress router, so reachability proves public ingress is operational.

**Non-reachability:** Private KAS and private ingress are verified to be NOT reachable from outside the VNet using `TestHTTPSConnectivity` (expects connection failure).

## Test plan

- [x] Verify compilation: `go build -tags E2Etests ./test/e2e/`
- [x] `make record-nonlocal-e2e` is a no-op (all tests are `AroRpApiCompatible`)
- [x] Test passes locally against pers env
- [x] CI e2e pass

### DIFF
--- a/test/e2e/cluster_create_fully_private.go
+++ b/test/e2e/cluster_create_fully_private.go
@@ -169,7 +169,7 @@ var _ = Describe("Customer", func() {
 				g.Expect(runErr).NotTo(HaveOccurred(), "RunVMCommand failed for kubectl version (output: %s)", versionOutput)
 				g.Expect(versionOutput).To(ContainSubstring("Server Version"),
 					"kubectl version should show Server Version, proving KAS is reachable from VM (output: %s)", versionOutput)
-			}, 60*time.Minute, 15*time.Second).Should(Succeed(), "KAS should be reachable from VM via internal LB")
+			}, 15*time.Minute, 15*time.Second).Should(Succeed(), "KAS should be reachable from VM via internal LB")
 			GinkgoLogr.Info("KAS is reachable from VM inside VNet")
 
 			By("verifying KAS is NOT reachable from outside the VNet")
--- a/test/e2e/cluster_create_fully_private.go
+++ b/test/e2e/cluster_create_fully_private.go
@@ -1,0 +1,243 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	hcpsdk20260630preview "github.com/Azure/ARO-HCP/test/sdk/v20260630preview/resourcemanager/redhatopenshifthcp/armredhatopenshifthcp"
+	"github.com/Azure/ARO-HCP/test/util/framework"
+	"github.com/Azure/ARO-HCP/test/util/labels"
+)
+
+// This test creates a fully private cluster (api.visibility: Private AND
+// ingress.type: Private) and verifies that:
+//   - Both KAS and ingress are only reachable from within the VNet
+//   - Neither KAS nor ingress is reachable from outside the VNet
+//   - The cluster is healthy and workloads are operational
+var _ = Describe("Customer", func() {
+	It("should create a fully private cluster with private KAS and private ingress and verify both are only reachable from VNet",
+		labels.RequireNothing,
+		labels.Critical,
+		labels.Positive,
+		labels.AroRpApiCompatible,
+		labels.CreateCluster,
+		labels.MIContainers(1),
+		func(ctx context.Context) {
+			const (
+				customerClusterName  = "fully-private"
+				customerNodePoolName = "np-1"
+			)
+
+			tc := framework.NewTestContext()
+
+			if tc.UsePooledIdentities() {
+				err := tc.AssignIdentityContainers(ctx, 1, framework.IdentityContainerAssignmentRetryInterval)
+				Expect(err).NotTo(HaveOccurred(), "failed to assign pooled identity containers")
+			}
+
+			By("creating a resource group")
+			resourceGroup, err := tc.NewResourceGroup(ctx, "fully-private", tc.Location())
+			Expect(err).NotTo(HaveOccurred(), "failed to create resource group for fully private test")
+
+			By("creating cluster parameters with private KAS and private ingress")
+			clusterParams := framework.NewDefaultClusterParams20260630()
+			clusterParams.ClusterName = customerClusterName
+			clusterParams.ManagedResourceGroupName = framework.SuffixName(*resourceGroup.Name, "-managed", 64)
+			clusterParams.APIVisibility = "Private"
+			clusterParams.IngressType = "Private"
+			clusterParams.OpenshiftVersionId = "4.22"
+
+			By("creating customer resources (infrastructure and managed identities)")
+			clusterParams, err = tc.CreateClusterCustomerResources20260630(ctx,
+				resourceGroup,
+				clusterParams,
+				map[string]interface{}{},
+				TestArtifactsFS,
+				framework.RBACScopeResourceGroup,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create customer resources for fully private cluster")
+
+			By("deploying test VM in customer VNet")
+			vmName, _, err := tc.DeployTestVM(ctx, TestArtifactsFS, *resourceGroup.Name, customerClusterName, clusterParams.VnetName, clusterParams.SubnetName)
+			Expect(err).NotTo(HaveOccurred(), "failed to deploy test VM for fully private verification")
+
+			By("creating the HCP cluster with private KAS and private ingress")
+			err = tc.CreateHCPClusterFromParam20260630(ctx,
+				GinkgoLogr,
+				*resourceGroup.Name,
+				clusterParams,
+				nil,
+				framework.ClusterCreationTimeout,
+			)
+			if isAPINotDeployedError(err) {
+				if time.Now().Before(framework.V20260630PreviewDeploymentDeadline) {
+					Skip(fmt.Sprintf("v20260630preview API not yet deployed; skipping until %s", framework.V20260630PreviewDeploymentDeadline.Format(time.RFC3339)))
+				}
+				Fail(fmt.Sprintf("v20260630preview API still not deployed as of %s deadline", framework.V20260630PreviewDeploymentDeadline.Format(time.RFC3339)))
+			}
+			Expect(err).NotTo(HaveOccurred(), "failed to create fully private HCP cluster %q", customerClusterName)
+
+			By("verifying cluster API visibility and ingress type are Private via ARM GET")
+			hcpClient := tc.Get20260630ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient()
+			cluster, err := hcpClient.Get(ctx, *resourceGroup.Name, customerClusterName, nil)
+			Expect(err).NotTo(HaveOccurred(), "failed to get cluster %q to verify fully private config", customerClusterName)
+			Expect(cluster.Properties).ToNot(BeNil(), "cluster %q Properties was nil", customerClusterName)
+			Expect(cluster.Properties.API).ToNot(BeNil(), "cluster %q Properties.API was nil", customerClusterName)
+			Expect(cluster.Properties.API.Visibility).ToNot(BeNil(), "cluster %q Properties.API.Visibility was nil", customerClusterName)
+			Expect(*cluster.Properties.API.Visibility).To(Equal(hcpsdk20260630preview.VisibilityPrivate),
+				"cluster %q API visibility should be Private", customerClusterName)
+			Expect(cluster.Properties.Ingress).ToNot(BeNil(), "cluster %q Properties.Ingress was nil", customerClusterName)
+			Expect(cluster.Properties.Ingress.Type).ToNot(BeNil(), "cluster %q Properties.Ingress.Type was nil", customerClusterName)
+			Expect(*cluster.Properties.Ingress.Type).To(Equal(hcpsdk20260630preview.IngressTypePrivate),
+				"cluster %q ingress type should be Private", customerClusterName)
+			Expect(cluster.Properties.API.URL).ToNot(BeNil(), "cluster %q Properties.API.URL was nil", customerClusterName)
+			apiURL := *cluster.Properties.API.URL
+			GinkgoLogr.Info("Cluster created fully private", "clusterName", customerClusterName, "apiURL", apiURL)
+
+			By("creating the node pool")
+			nodePoolParams := framework.NewDefaultNodePoolParams20260630()
+			nodePoolParams.ClusterName = customerClusterName
+			nodePoolParams.NodePoolName = customerNodePoolName
+			nodePoolParams.Replicas = int32(2)
+
+			err = tc.CreateNodePoolFromParam20260630(ctx,
+				GinkgoLogr,
+				*resourceGroup.Name,
+				clusterParams.ManagedResourceGroupName,
+				customerClusterName,
+				nodePoolParams,
+				framework.NodePoolCreationTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create node pool %q for fully private cluster %q",
+				customerNodePoolName, customerClusterName)
+
+			By("getting admin credentials for the cluster")
+			adminRESTConfig, err := tc.GetAdminRESTConfigForHCPCluster20240610(
+				ctx,
+				tc.Get20240610ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient(),
+				*resourceGroup.Name,
+				customerClusterName,
+				framework.GetAdminRESTConfigTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to get admin REST config for fully private cluster %q", customerClusterName)
+
+			By("verifying KAS is reachable from VM inside the VNet")
+			internalIP, err := framework.GetPrivateKASInternalIP(ctx, tc, clusterParams.ManagedResourceGroupName)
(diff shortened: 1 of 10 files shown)


test/e2e/cluster_create_fully_private.go
```diff
@@ -0,0 +1,243 @@
...
+				tc.Get20240610ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient(),
+				*resourceGroup.Name,
+				customerClusterName,
+				framework.GetAdminRESTConfigTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to get admin REST config for fully private cluster %q", customerClusterName)
+
+			By("verifying KAS is reachable from VM inside the VNet")
+			internalIP, err := framework.GetPrivateKASInternalIP(ctx, tc, clusterParams.ManagedResourceGroupName)
+			Expect(err).NotTo(HaveOccurred(), "failed to find private KAS internal LB IP in managed resource group %q", clusterParams.ManagedResourceGroupName)
```
zgalor, line 146: This is the same DNS workaround that was removed from `cluster_create_private_kas.go` in PR #6756 (ARO-29270). CS MR !434 fixes the public DNS A record to point to the KAS internal LB private IP for **all** private KAS clusters, including fully-private ones.

The hostname in the kubeconfig (`api.<id>.<region>.aroapp-hcp.io`) resolves uniquely to the KAS private IP regardless of how many internal LBs exist in the managed RG — so the updated `GetPrivateKASInternalIP` (with the `kube-apiserver` rule check) is also not needed.

Please apply the same fix as #6756:
1. Remove `GetPrivateKASInternalIP`, the host override, and `Insecure: true`
2. Use the kubeconfig as-is — the VM resolves DNS hostname → private IP → internal LB → KAS
3. Add the `net.LookupHost` + `IsPrivate()` DNS regression check that #6756 introduces

test/e2e/cluster_create_fully_private.go
```diff
@@ -0,0 +1,243 @@
...
+			kubeconfigB64 := base64.StdEncoding.EncodeToString([]byte(kubeconfig))
+
+			Eventually(func(g Gomega) {
+				versionCmd := fmt.Sprintf(
+					"KUBECONFIG=$(mktemp) && "+
+						"trap 'rm -f $KUBECONFIG' EXIT && "+
+						"echo '%s' | base64 -d > $KUBECONFIG && "+
+						"chmod 600 $KUBECONFIG && "+
+						"kubectl --kubeconfig=$KUBECONFIG version 2>&1",
+					kubeconfigB64,
```
zgalor, line 166: Two issues here:

1. **`2>&1` is no longer needed** — with PR #6429 merged, `RunVMCommand` now uses managed run commands and enforces non-zero exit codes. Stderr is no longer the failure signal; exit code is. The `2>&1` redirect no longer helps.

2. **`ContainSubstring("Server Version")` is fragile** — kubectl output format could change.

Use `RunKubectlOnVM` instead — it handles stderr correctly (`2>/dev/null`) and a plain `err == nil` check is sufficient:
```go
versionOutput, err := framework.RunKubectlOnVM(ctx, tc, *resourceGroup.Name, vmName, kubeconfigB64, "version", 2*time.Minute)
Expect(err).NotTo(HaveOccurred(), "kubectl version should succeed from VM (output: %s)", versionOutput)
```

test/e2e/cluster_create_fully_private_with_kv.go
```diff
@@ -0,0 +1,269 @@
...
+				tc.Get20240610ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient(),
+				*resourceGroup.Name,
+				customerClusterName,
+				framework.GetAdminRESTConfigTimeout,
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to get admin REST config for fully private cluster with KV %q", customerClusterName)
+
+			By("verifying KAS is reachable from VM inside the VNet")
+			internalIP, err := framework.GetPrivateKASInternalIP(ctx, tc, clusterParams.ManagedResourceGroupName)
+			Expect(err).NotTo(HaveOccurred(), "failed to find private KAS internal LB IP in managed resource group %q", clusterParams.ManagedResourceGroupName)
```
zgalor, line 177: Same DNS workaround issue as in `cluster_create_fully_private.go` — see comment there. CS MR !434 applies to fully-private clusters too; `GetPrivateKASInternalIP` and the host override should be removed. Please apply the same fix as PR #6756.

test/e2e/cluster_create_fully_private_with_kv.go
```diff
@@ -0,0 +1,269 @@
...
+			kubeconfigB64 := base64.StdEncoding.EncodeToString([]byte(kubeconfig))
+
+			Eventually(func(g Gomega) {
+				versionCmd := fmt.Sprintf(
+					"KUBECONFIG=$(mktemp) && "+
+						"trap 'rm -f $KUBECONFIG' EXIT && "+
+						"echo '%s' | base64 -d > $KUBECONFIG && "+
+						"chmod 600 $KUBECONFIG && "+
+						"kubectl --kubeconfig=$KUBECONFIG version 2>&1",
+					kubeconfigB64,
```
zgalor, line 197: Same issue as in `cluster_create_fully_private.go` — `2>&1` is no longer needed after PR #6429, and `ContainSubstring("Server Version")` is fragile. Use `RunKubectlOnVM` instead (see comment in the other file).

test/util/framework/resource_group_helper.go
```diff
@@ -59,9 +61,16 @@ func GetPrivateKASInternalIP(ctx context.Context, tc interface {
...
-				// Internal LBs have a private IP and no public IP
+				// Internal LBs have a private IP and no public IP.
+				// Fully-private clusters have two internal LBs (KAS and ingress),
+				// so also check for the kube-apiserver load balancing rule to
+				// identify the KAS LB specifically.
 				if fip.Properties.PrivateIPAddress != nil && fip.Properties.PublicIPAddress == nil {
-					return *fip.Properties.PrivateIPAddress, nil
+					for _, rule := range fip.Properties.LoadBalancingRules {
+						if rule.ID != nil && strings.HasSuffix(*rule.ID, "/kube-apiserver") {
+							return *fip.Properties.PrivateIPAddress, nil
```
zgalor, line 71: This distinction is clever, but the whole function becomes unnecessary once DNS is used — the API hostname resolves uniquely to the KAS IP.

Note also that this file was **deleted** in PR #6756 as part of the ARO-29270 DNS fix. These two PRs conflict. Applying the DNS fix to the fully-private tests will resolve both issues at once.

zgalor (commented):
## Review

### Coverage — no duplication ✅

The three new tests cover genuinely new combinations that don't duplicate existing tests:
- **`cluster_create_fully_private.go`** — private KAS + private ingress (interaction effects)
- **`cluster_create_fully_private_with_kv.go`** — all three private (KAS + ingress + KV)
- **`cluster_create_public_and_private_kas.go`** — default public cluster with explicit Swift path verification

These complement `cluster_create_private_kas.go` (private KAS / public ingress) and `cluster_create_private_ingress.go` (public KAS / private ingress) without overlap.

`cluster_create_public_and_private_kas.go` looks good — using `VerifyNodesReady` + `VerifyGetDeploymentLogs` to prove Swift networking is clean and requires no VM. No issues there ✅

test/e2e/cluster_create_fully_private.go
jiezhao16: Done

test/e2e/cluster_create_fully_private.go
jiezhao16: Done

test/e2e/cluster_create_fully_private_with_kv.go
jiezhao16: Done

test/e2e/cluster_create_fully_private_with_kv.go
jiezhao16: Done

test/util/framework/resource_group_helper.go
jiezhao16: I've removed the GetPrivateKASInternalIP usage from test files. I kept resource_group_helper.go for now since PR#6756 still references it, that'll resolve cleanly once your PR #6756 merges

jiezhao16: I've removed resource_group_helper.go from this PR's diff by restoring it to the main branch version

test/e2e/cluster_create_fully_private.go
```diff
@@ -0,0 +1,246 @@
...
+			Expect(err).NotTo(HaveOccurred(), "failed to parse API URL %q", apiURL)
+			apiHost := parsedURL.Hostname()
+			ips, err := net.LookupHost(apiHost)
+			Expect(err).NotTo(HaveOccurred(), "DNS lookup failed for API hostname %q", apiHost)
+			Expect(ips).NotTo(BeEmpty(), "DNS lookup returned no IPs for API hostname %q", apiHost)
+			resolvedIP := net.ParseIP(ips[0])
+			Expect(resolvedIP).NotTo(BeNil(), "DNS returned unparseable IP %q for %q", ips[0], apiHost)
+			Expect(resolvedIP.IsPrivate()).To(BeTrue(),
+				fmt.Sprintf("API hostname %q resolved to public IP %s — expected a private IP (RFC 1918) pointing to the internal LB", apiHost, resolvedIP))
+			GinkgoLogr.Info("API hostname DNS resolves to private IP", "hostname", apiHost, "ip", resolvedIP.String())
```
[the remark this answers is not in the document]
jiezhao16: This check mirrors the pattern from PR #6756 which checks ips[0]. For private KAS cluster, there is a single DNS A record pointing to the internal LB, so checking all records is not necessary here.

test/e2e/cluster_create_fully_private_with_kv.go
```diff
@@ -0,0 +1,272 @@
...
+			apiHost := parsedURL.Hostname()
+			ips, err := net.LookupHost(apiHost)
+			Expect(err).NotTo(HaveOccurred(), "DNS lookup failed for API hostname %q", apiHost)
+			Expect(ips).NotTo(BeEmpty(), "DNS lookup returned no IPs for API hostname %q", apiHost)
+			resolvedIP := net.ParseIP(ips[0])
+			Expect(resolvedIP).NotTo(BeNil(), "DNS returned unparseable IP %q for %q", ips[0], apiHost)
+			Expect(resolvedIP.IsPrivate()).To(BeTrue(),
+				fmt.Sprintf("API hostname %q resolved to public IP %s — expected a private IP (RFC 1918) pointing to the internal LB", apiHost, resolvedIP))
+			GinkgoLogr.Info("API hostname DNS resolves to private IP", "hostname", apiHost, "ip", resolvedIP.String())
+
```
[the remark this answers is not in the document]
jiezhao16: This check mirrors the pattern from PR #6756 which checks ips[0]. For private KAS cluster, there is a single DNS A record pointing to the internal LB, so checking all records is not necessary here.

test/e2e/cluster_create_fully_private.go
```diff
@@ -0,0 +1,246 @@
...
+
+			By("verifying API hostname DNS resolves to a private IP")
+			// The public DNS A record for a private KAS cluster must point to the
+			// private IP of the internal load balancer — not the shared ingress
+			// public IP. Test this explicitly to catch regressions like ARO-29270.
+			parsedURL, err := url.Parse(apiURL)
+			Expect(err).NotTo(HaveOccurred(), "failed to parse API URL %q", apiURL)
+			apiHost := parsedURL.Hostname()
+			ips, err := net.LookupHost(apiHost)
+			Expect(err).NotTo(HaveOccurred(), "DNS lookup failed for API hostname %q", apiHost)
```
zgalor, line 127: Two issues here (same applies to `cluster_create_fully_private_with_kv.go`):

1. **Context-unaware lookup, order-dependent check**: `net.LookupHost` has no context for cancellation, and checking `ips[0]` is order-dependent — DNS can return multiple records in any order. Use `net.DefaultResolver.LookupHost(ctx, apiHost)` and assert against the full set.

2. **Exact IP match vs `IsPrivate()`**: Our original review of this PR suggested removing `GetPrivateKASInternalIP` entirely. After review discussions on PR #6756 (Manyanda and Miguel's feedback), we revised that position — `GetPrivateKASInternalIP` was kept specifically for the DNS assertion, comparing the resolved IPs against the known LB IP for exactness. For consistency, this check should do the same: call `GetPrivateKASInternalIP` and use `ContainElement(lbIP)` instead of `IsPrivate()`.

   Important caveat: fully-private clusters have **two** internal LBs (KAS and ingress). The simple "first internal LB with private IP" lookup could return the ingress LB. `GetPrivateKASInternalIP` needs to distinguish them — the original version of this PR had a `kube-apiserver` load balancing rule check for this; that check should be included.

test/e2e/cluster_create_fully_private.go
```diff
@@ -0,0 +1,246 @@
...
+			// connects through the private KAS endpoint without any host override.
+			kubeconfig, err := framework.GenerateKubeconfig(adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "failed to generate kubeconfig from admin REST config")
+			kubeconfigB64 := base64.StdEncoding.EncodeToString([]byte(kubeconfig))
+
+			// kubectl version hits /version through the DNS-resolved private IP →
+			// internal LB → Swift → KAS pods, proving the private KAS network path
+			// is functional end-to-end.
+			versionOutput, err := framework.RunKubectlOnVM(ctx, tc, *resourceGroup.Name, vmName, kubeconfigB64, "version", 2*time.Minute)
+			Expect(err).NotTo(HaveOccurred(),
```
zgalor, line 174: Single-shot call. PR #6756 wraps `RunKubectlOnVM` in an `Eventually` retry loop (5 min / 15 s) because the internal LB backend pool can take time to become healthy after node pool creation. Same risk applies here — please add a retry loop. Same applies to `cluster_create_fully_private_with_kv.go`.

test/e2e/cluster_create_fully_private_with_kv.go
```diff
@@ -0,0 +1,272 @@
...
+			By("verifying API hostname DNS resolves to a private IP")
+			// The public DNS A record for a private KAS cluster must point to the
+			// private IP of the internal load balancer — not the shared ingress
+			// public IP. Test this explicitly to catch regressions like ARO-29270.
+			parsedURL, err := url.Parse(apiURL)
+			Expect(err).NotTo(HaveOccurred(), "failed to parse API URL %q", apiURL)
+			apiHost := parsedURL.Hostname()
+			ips, err := net.LookupHost(apiHost)
+			Expect(err).NotTo(HaveOccurred(), "DNS lookup failed for API hostname %q", apiHost)
```
zgalor, line 158: Same issues as in `cluster_create_fully_private.go` — use `net.DefaultResolver.LookupHost(ctx, ...)` + `ContainElement(lbIP)` with a `kube-apiserver`-aware `GetPrivateKASInternalIP`. See comment there.

test/e2e/cluster_create_fully_private_with_kv.go
```diff
@@ -0,0 +1,272 @@
...
+			// connects through the private KAS endpoint without any host override.
+			kubeconfig, err := framework.GenerateKubeconfig(adminRESTConfig)
+			Expect(err).NotTo(HaveOccurred(), "failed to generate kubeconfig from admin REST config")
+			kubeconfigB64 := base64.StdEncoding.EncodeToString([]byte(kubeconfig))
+
+			// kubectl version hits /version through the DNS-resolved private IP →
+			// internal LB → Swift → KAS pods, proving the private KAS network path
+			// is functional end-to-end.
+			versionOutput, err := framework.RunKubectlOnVM(ctx, tc, *resourceGroup.Name, vmName, kubeconfigB64, "version", 2*time.Minute)
+			Expect(err).NotTo(HaveOccurred(),
```
zgalor, line 205: Same as `cluster_create_fully_private.go` — wrap in an `Eventually` retry loop for LB backend readiness.

test/util/framework/resource_group_helper.go
```diff
@@ -52,20 +57,24 @@ func GetPrivateKASInternalIP(ctx context.Context, tc interface {
...
+				if fip.Properties == nil || fip.Properties.PrivateIPAddress == nil || fip.Properties.PublicIPAddress != nil {
 					continue
 				}
-				// Internal LBs have a private IP and no public IP
-				if fip.Properties.PrivateIPAddress != nil && fip.Properties.PublicIPAddress == nil {
-					return *fip.Properties.PrivateIPAddress, nil
+				for _, rule := range fip.Properties.LoadBalancingRules {
+					if rule.ID != nil && strings.HasSuffix(*rule.ID, "/kube-apiserver") {
+						return *fip.Properties.PrivateIPAddress, nil
+					}
```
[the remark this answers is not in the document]
jiezhao16: The current check works correctly — a load balancing rule's ARM resource ID ends with `/loadBalancingRules/kube-apiserver`, which is a valid match for `strings.HasSuffix(*rule.ID, "/kube-apiserver")`. This is the same pattern used in PR #6756.

This file (`resource_group_helper.go`) is a temporary addition to this PR. Once PR #6756 merges, it will be removed from this PR's diff and the version from PR #6756 will be used instead.

PR #6809, opened 2026-09-03
fix(int): update aro-hcp-int-msi-mock identity to new principal

venkateshsredhat:
The aro-hcp-int-msi-mock Entra app was recreated, changing its App ID
  and Principal ID. e2e-subscription-rbac-assignment-subscription.bicep
  reads miMockPrincipalId from config to grant Key Vault Crypto User on
  the E2E subscription — it never granted that role to the new principal,
  causing ValidAzureKMSConfig=False across 18 INT clusters from 2026-09-02.

  Old: miMockClientId=e8723db7  miMockPrincipalId=d6b62dfa
  New: miMockClientId=f2e4769e  miMockPrincipalId=6096f642


raelga:
`backend/values.yaml` and `cluster-service/values.yaml` also template `miMockClientId`/`miMockPrincipalId` straight from this same config value into the RP backend and cluster-service deployments, separate from the RBAC bicep path. Any environment stamping run against the current config bakes the old identity into those components too.

raelga (approved):
Confirmed: this matches the recreated aro-hcp-int-msi-mock identity (6096f642) and fixes the config side of the RBAC gap I already remediated live in INT.

PR #6680, opened 2026-08-25
Add deny assignment management

deads2k:
/assign @machi1990

PR #6816, opened 2026-09-03
fix(backend): wait for managed resource group before deny-assignment writes (AROSLSRE-2012)

raelga:
## What

The `ClusterDenyAssignment` controller (added in #6680) writes deny assignments scoped to the cluster's managed resource group without first confirming that resource group exists in Azure. This causes repeated `ResourceGroupNotFound` failures against ARM until the cluster-create deadline expires. See [AROSLSRE-2012](https://redhat.atlassian.net/browse/AROSLSRE-2012) and the evidence linked from [#6804](https://github.com/Azure/ARO-HCP/pull/6804#issuecomment-5527582776).

## Why

The deny-assignment controller can run before the managed resource group exists because it gates on `ClusterServiceIDForCluster`, which resolves via `PendingClusterServiceID`, set before the real CS cluster object (and its managed resource group) exists.

There is also a separate, deeper circular dependency: `createPreconditionDenyAssignmentsCreated` in `cluster_cluster_service_create_controller.go` blocks CS cluster object creation until deny assignments succeed, but Cluster Service only provisions the managed resource group after the CS cluster object exists, so deny assignments can never succeed on their own. That loop is not fixed by this PR, deads2k is pulling out that initialization logic separately to break it.

## Fix

Before issuing deny-assignment writes, check whether the managed resource group exists, mirroring the same check already used in `managed_resource_group_controller.go`. On `ResourceGroupNotFound`, log and wait for a later reconcile instead of hard-failing and retrying to failure.

This is a stopgap: it stops the ARM hammering, it does not break the circular dependency described above.

## Testing

- `go build ./pkg/azure/... ./pkg/controllers/cluster/denyassignments/...`
- `go test ./pkg/controllers/cluster/denyassignments/... ./pkg/azure/azuremockclient/... ./pkg/controllers/cluster/azureresources/...` all pass
- Added a `ResourceGroupsClientFunc` test fake (mirroring the existing `GenericResourcesClientFunc`/`DenyAssignmentsClientFunc` fakes), defaulting to "resource group exists" so existing tests didn't need changes

Refs: [AROSLSRE-2012](https://redhat.atlassian.net/browse/AROSLSRE-2012), #6680, #6804

### DIFF
--- a/backend/pkg/azure/azuremockclient/mock_clients.go
+++ b/backend/pkg/azure/azuremockclient/mock_clients.go
@@ -87,12 +87,43 @@ func (m *GenericResourcesClientFunc) BeginDeleteByID(ctx context.Context, resour
 	return nil, fmt.Errorf("GenericResourcesClientFunc: set DeleteErr to control this path; PollUntilDone cannot be called on a nil poller")
 }
 
+// ResourceGroupsClientFunc adapts a function to the ResourceGroupsClient.Get interface.
+// Tests set GetFunc to control the response per call; unset GetFunc reports the resource
+// group as existing (an empty successful response), which is the common case for tests that
+// don't care about resource-group existence.
+type ResourceGroupsClientFunc struct {
+	GetFunc func(ctx context.Context, resourceGroupName string, options *armresources.ResourceGroupsClientGetOptions) (armresources.ResourceGroupsClientGetResponse, error)
+}
+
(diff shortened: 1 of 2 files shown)


raelga:
Superseded by #6817, which disables the ClusterDenyAssignment controller entirely rather than just fixing the precondition. Closing since this patches code that won't run, and the re-enable will likely need reworked initialization logic anyway.

PR #6817, opened 2026-09-03
revert the deny assignment logic that needs managed resource group first

deads2k:


raelga (approved):
/lgtm
/approve

raelga:
For anyone landing here later: this revert breaks a circular dependency introduced by #6269 (`ClusterDenyAssignment` controller).

**The deadlock:** `createPreconditionDenyAssignmentsCreated` in `cluster_cluster_service_create_controller.go` blocked CS cluster object creation until deny assignments succeeded. But Cluster Service only provisions the managed resource group (MRG) *after* the CS cluster object exists, and deny assignments are written scoped to that MRG. So deny assignments could never succeed before the MRG existed, and the CS cluster object could never be created until deny assignments succeeded, an unbreakable wait with no forcing function on either side.

In practice this surfaced as repeated `ResourceGroupNotFound` failures against ARM until the cluster-create deadline expired (see [AROSLSRE-2012](https://redhat.atlassian.net/browse/AROSLSRE-2012), evidence in [#6804](https://github.com/Azure/ARO-HCP/pull/6804#issuecomment-5527582776)).

#6816 was a stopgap that made the deny-assignment writer check for MRG existence before writing (stops the ARM hammering), but its own description called out that it doesn't fix the deeper loop, `deads2k is pulling out that initialization logic separately to break it`, that's this PR. Closing #6816 in favor of this revert removes the blocking precondition entirely rather than trying to sequence around it.

PR #6818, opened 2026-09-03
test: prototype DEV mock-MSI ACL isolation (ARO-29287)

mgahagan73:
https://issues.redhat.com/browse/ARO-29287

Related: [ARO-29290](https://issues.redhat.com/browse/ARO-29290), [ARO-29288](https://issues.redhat.com/browse/ARO-29288), [ARO-29289](https://issues.redhat.com/browse/ARO-29289)

### What

Adds a DEV-only e2e spec and dedicated suite `development/mock-msi-acl` (`Parallelism` is the literal `1`) that strips, then restores, the **job-leased** mock-MSI subscription grants so cluster delete can be shown to work when the identity used for cluster ops has no ACL.

- New spec: `test/e2e/cluster_delete_missing_identities_dev_mock_msi.go` (`Mock-MSI-ACL`, `MIContainers:0`, dedicated UAMIs).
- Helpers: resolve lease from `LEASED_MSI_MOCK_SP` / `ARO_HCP_MSI_MOCK_PRINCIPAL_ID`, list with `assignedTo('{id}')`, strip, restore the same assignment GUIDs. Refuses `aro-dev-msi-mock2`.
- Excluded from every existing parallel suite. `test/e2e/cluster_delete_missing_identities.go` stays `StageAndProdOnly`.
- Docs: `docs/ci/mock-msi-acl-isolation.md` (false-pass if Helm drifted, personal-dev overlay).

Does **not** wire `openshift/release` ([ARO-29289](https://redhat.atlassian.net/browse/ARO-29289)) or change the hardcoded MI dataplane client.

### Why

In DEV every operator authenticates as one mock SP. Deleting customer UAMIs does not remove that SP's subscription `dev-msi-mock` / Key Vault Crypto User grants, so the STG/PROD missing-identity delete spec is vacuous if run in DEV. The high-value [ARO-29096](https://redhat.atlassian.net/browse/ARO-29096) signal is that delete still works when that impersonated identity has **no ACL**.

Ginkgo `Serial` does not serialize across OTE workers, and the default DEV suite is `Parallelism: 24`, so stripping the job's mock SP in-place would fail sibling specs. Isolation is "this invocation owns the leased pool SP exclusively", not "mutate a shared grant while 24 specs are using it". Shared personal-dev `aro-dev-msi-mock2` must never be stripped.

### Testing

- Unit tests: `test/util/framework/mock_msi_permissions_test.go` — lease resolve (mock2 refuse, pool lookup, env mismatch), `assignedTo` list filter, strip/restore (`RoleAssignmentExists` is success on restore).
- Integration tests: none. This is e2e + suite wiring; no RP/Cosmos artifact change.
- E2E: personal-dev soak on pool member 19 (`aro-hcp-msi-mock-cs-sp-dev-19`). Create + viability, strip both subscription assignments, delete customer UAMIs, cluster delete to ARM 404 and managed RG 404, `DeferCleanup` restored both GUIDs. `SUCCESS! -- 1 Passed | 0 Failed`.

  Failure modes seen while bringing pers up (spec/helpers did fail closed as intended): ARM `400 UnsupportedQuery` on `principalId eq` (filter is `assignedTo`); `AADSTS700027` when Helm `clientId` and mounted cert disagree; spec refuses to strip mock2. A 4.21 complete-create sanity created a viable cluster first; that spec later failed on a local `:8443` port-forward drop during node pool wait — this PR spec does not create a node pool.

  Not in the default DEV parallel job. `LEASED_MSI_MOCK_SP` is still provision-only until [ARO-29290](https://redhat.atlassian.net/browse/ARO-29290).

- `Parallelism: 1` must stay a literal. `ARO_HCP_SUITE_PARALLELISM` must not raise it.
- Env export is not enough: if CS/backend Helm still impersonate `aro-dev-msi-mock2` while the test strips a pool member, delete can false-pass. See the false-pass section in the isolation doc; [ARO-29290](https://redhat.atlassian.net/browse/ARO-29290) should smoke **running** pods, not only rendered config.
- Restore is RBAC only. Personal-dev Helm `miMock*` must be put back on mock2 after overlay testing.
- Suggested reviewers: @bvesel @stevekuznetsov

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/ci/README.md
+++ b/docs/ci/README.md
@@ -95,6 +95,12 @@ ARO HCP CI is split across this repository and the OpenShift CI configuration in
 - [Lease Configuration](identity-leasing.md#lease-configuration)
 - [Where To Look](identity-leasing.md#where-to-look)
 
+### [DEV Mock-MSI ACL Isolation](mock-msi-acl-isolation.md)
+
+- How to strip (then restore) the job-leased mock MSI for one DEV e2e spec without breaking sibling clusters
+- [Helm must match the lease (false pass)](mock-msi-acl-isolation.md#helm-must-match-the-lease-false-pass)
+- [Personal-dev overlay](mock-msi-acl-isolation.md#personal-dev-overlay)
+
 ### [DEV CI Monitoring and Alert Response](dev-ci-monitoring.md)
 
 - [At A Glance](dev-ci-monitoring.md#at-a-glance)
--- a/docs/ci/identity-leasing.md
+++ b/docs/ci/identity-leasing.md
@@ -333,6 +333,8 @@ MSI_MOCK_CERT_NAME=$(yq ".miMockPool.\"${LEASED_MSI_MOCK_SP}\".certName" dev-inf
 
 Jobs only consume the Boskos key and the static `msi-mock-pool.yaml` catalog at runtime. They do not query Entra or the `dev-ci` rollout directly during provisioning.
 
+The mock SP is process-wide for the job. Stripping its subscription grants while other specs in the same OTE invocation are still creating or deleting clusters would fail those siblings. The ARM-helper pool already isolates CheckAccess per job; MSI mock uses the same job-level lease. Per-spec isolation is a dedicated `development/mock-msi-acl` suite (`Parallelism: 1`) that strips then restores only a **pooled** mock principal — never `aro-dev-msi-mock2`. The test process must receive the same lease Helm applied, and live Cluster Service / backend `miMock*` must match it, or the spec can false-pass. See [DEV Mock-MSI ACL Isolation](mock-msi-acl-isolation.md) and [Helm must match the lease](mock-msi-acl-isolation.md#helm-must-match-the-lease-false-pass).
+
 ## ARM Helper Service Principal Pool
 
 The DEV ARM helper pool prevents concurrent E2E backends from sharing the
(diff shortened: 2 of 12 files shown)


patilsuraj767:
@mgahagan73  I took a look at the proposed changes, and they look good to me. Adding a dedicated suite with Parallelism: 1 should help ensure the role assignment is removed safely without impacting other specs running in the same E2E suite.

I don’t have enough insight into how the mock-MSI leasing works, so I think someone from the service lifecycle team should review this as well.

My only concern is that the e2e-parallel and the new mock-msi-acl suites should not get the same mock MSI identity when they are running in parallel.

mgahagan73:
Hi @patilsuraj767, thanks for taking a look. 

Boskos already hands out one mock SP per job (LEASED_MSI_MOCK_SP). development/mock-msi-acl being its own suite with Parallelism: 1 only keeps sibling specs in that invocation from racing the strip. It does not stop e2e-parallel from using the same SP if both suites run in the same ci-operator job (same provision, same lease). In that shape they would share the impersonated identity, and stripping it would take down the parallel suite.

We would create this as  a separate job that leases a different SP — and, from SLT feedback, a dedicated mock pool (1–2 members) so a crash or a Privileged RBAC re-apply would not break the shared aro-hcp-msi-mock-cs-sp-dev-* pool that every other DEV e2e job uses. ([ARO-29289](https://redhat.atlassian.net/browse/ARO-29289)) 

@stevekuznetsov Does that match what you want for how we would run this (separate job + tiny ACL-only pool, periodic first)?

PR #6746, opened 2026-08-31
POC: feat: add ExternalAuthAvailableController to surface O…

shitaljante:
…IDC client availability

https://redhat.atlassian.net/browse/ARO-29079

### What

Adds a new ExternalAuthAvailableController that reads HostedCluster OIDC client status conditions from the ReadDesire cache and maps them onto ExternalAuth.Status.UserFacingConditions as an Available condition. This decouples "the RP accepted the request" (provisioningState: Succeeded) from "the resource is fully ready" (Available: True/False).

- Reads HostedCluster.Status.Configuration.Authentication.OIDCClients[].Conditions from the ReadDesire cache
- Matches OIDC client statuses to ExternalAuth clients by component name/namespace (case-insensitive)
- Aggregates across multiple matched clients using worst-condition-wins semantics
- Maps HC conditions to a user-facing Available condition:
  - HC Available: True + OIDCConfigAvailable → Available: True, Reason: OIDCConfigAvailable
  - HC Degraded: True + OIDCClientSecretGet → Available: False, Reason: AwaitingSecret
  - HC Degraded: True + other → Available: False, forwarding HC reason/message
  - No HC data → Available: False/Unknown, Reason: HostedClusterNotReady
- No changes to the frontend, ARM API conversion, or provisioning state logic -- the existing newExternalAuthResourceStatus() already surfaces UserFacingConditions via the ARM API

### Why

The current external auth flow lacks a way for users to know whether their OIDC configuration is actually operational on the hosted cluster or waiting for the secrets from user.  

### Testing

- Unit tests
Output will be in the format - 
`        "status": {
            "conditions": [
                {
                    "lastTransitionTime": "2026-09-17T07:24:50Z",
                    "message": "The external auth provider is waiting for the client secret to be created in the openshift-config namespace",
                    "reason": "AwaitingSecret",
                    "status": "False",
                    "type": "ConsoleAvailable"
                },
                {
                    "lastTransitionTime": "2026-09-17T07:24:50Z",
                    "message": "OIDC client conditions do not indicate readiness",
                    "reason": "HostedClusterNotReady",
                    "status": "False",
                    "type": "CliAvailable"
                }
            ]
        }
        `

### PR Checklist

- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
[...]

PR #6823, opened 2026-09-04
Feat : ARO-29371 | Improve CA Certificate Validation in RP for ExternalAuth

m-mandhare:
[Improve CA Certificate Validation in RP for ExternalAuth](https://redhat.atlassian.net/browse/ARO-29371)

### What

RP External Auth now rejects issuer CAs that are 

- not PEM certificates, 
- are not CA:TRUE, 
- are expired 
-  include private keys or other PEM types
-  have trailing non-PEM data. 

If ca cert expires after creation or not yet valid, then it shows error in user facing condition

### Why

ValidatePEM previously accepted any parseable cert bundle, so a leaf, expired, or key-mixed CA could be stored and later break OIDC discovery against a private IdP. Stricter checks fail those requests at admission with a clear properties.issuer.ca error 

### Testing

justification.

- Unit tests
- Manual tests
 1. Not a pem certificate. Some blabber in ca feild.
image

2. Passed private key instead of ca cert
image

3. Passed csr
image

4. Empty certificate block
image

5. Valid ca + private key
image

6. Trailing junk after valid certificate
image

7. Invalid CA cert 
* A true CA certificate must explicitly state that it is allowed to sign other certificates. Looking at the decoded extension data inside your certificate text:
* Basic Constraints: CA:FALSE
image

8. Root CA is valid | leaf CA is invalid
image

9. Expired CA
image

10. CA cert not valid yet - It gives user facing condition / No error while put/patch 
image

image

11. If Ca cert expired after creating external auth -
image

Positive Scenarios - 
1. Valid CA cert
image

2. Valid CA bundle
image

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
+++ b/backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
@@ -39,6 +39,11 @@ const (
 	// CACertificateExpiryConditionType is the user-facing condition type set when the issuer
 	// CA certificate has expired. Written to Status.UserFacingConditions only while expired;
 	CACertificateExpiryConditionType = "CACertificateExpiry"
+
+	// CACertificateNotYetValidConditionType is the user-facing condition type set when the
+	// issuer CA certificate has a NotBefore in the future. A future NotBefore is accepted at
+	// admission time but surfaced here so operators are aware.
+	CACertificateNotYetValidConditionType = "CACertificateNotYetValid"
 )
 
 // externalAuthDegradedAggregator rolls per-controller Degraded conditions
@@ -123,10 +128,17 @@ func (c *externalAuthDegradedAggregator) SyncOnce(ctx context.Context, key contr
 	apimeta.SetStatusCondition(&replacement.Status.Conditions, aggregated)
 
 	// CACertificateExpiry is user-facing only when the CA has expired.
-	if caCond := caCertExpiryCondition(existing.Properties.Issuer.CA, c.clock.Now()); caCond != nil {
-		apimeta.SetStatusCondition(&replacement.Status.UserFacingConditions, *caCond)
+	// CACertificateNotYetValid is user-facing only when the CA has a start date in the future.
+	expiryCond, notYetValidCond := caCertValidityConditions(existing.Properties.Issuer.CA, c.clock.Now())
+	if expiryCond != nil {
+		apimeta.SetStatusCondition(&replacement.Status.UserFacingConditions, *expiryCond)
 	} else {
 		apimeta.RemoveStatusCondition(&replacement.Status.UserFacingConditions, CACertificateExpiryConditionType)
+	}
+	if notYetValidCond != nil {
+		apimeta.SetStatusCondition(&replacement.Status.UserFacingConditions, *notYetValidCond)
+	} else {
+		apimeta.RemoveStatusCondition(&replacement.Status.UserFacingConditions, CACertificateNotYetValidConditionType)
 	}
 
 	conditionsChanged := !equality.Semantic.DeepEqual(existing.Status.Conditions, replacement.Status.Conditions)
@@ -149,18 +161,23 @@ func (c *externalAuthDegradedAggregator) SyncOnce(ctx context.Context, key contr
 	return nil
 }
 
-// caCertExpiryCondition inspects all CERTIFICATE PEM blocks in ca and returns a
-// CACertificateExpiry condition only when the CA has expired. Healthy, missing, or
-// unparseable CAs return nil so the condition is omitted from user-facing status.
-// When a bundle contains multiple certs the cert with the earliest NotAfter
-// drives the condition.
-func caCertExpiryCondition(ca string, now time.Time) *metav1.Condition {
+// caCertValidityConditions inspects all CERTIFICATE PEM blocks in ca and returns
+// two conditions: expiryCond (CACertificateExpiry) and notYetValidCond (CACertificateNotYetValid).
+// Each is nil when not applicable. Both are derived from a single pass over the PEM bundle:
+//   - expiryCond is set when any cert has expired; the cert with the earliest NotAfter drives
+//     the condition message.
+//   - notYetValidCond is set when any cert has a start date in the future;
+//
+// Missing, empty, or entirely unparseable input causes both to return nil.
+func caCertValidityConditions(ca string, now time.Time) (expiryCond, notYetValidCond *metav1.Condition) {
 	if ca == "" {
-		return nil
+		return nil, nil
 	}
 
-	// Walk every PEM block; only CERTIFICATE blocks are relevant.
-	var earliest *x509.Certificate
+	// Single pass: track earliest-expiring cert and start date is in the future cert.
+	var earliestExpired *x509.Certificate     // drives CACertificateExpiry
+	var earliestNotYetValid *x509.Certificate // drives CACertificateNotYetValid
+
 	rest := []byte(ca)
 	for {
 		var block *pem.Block
@@ -175,25 +192,39 @@ func caCertExpiryCondition(ca string, now time.Time) *metav1.Condition {
 		if err != nil {
 			continue
 		}
-		if earliest == nil || cert.NotAfter.Before(earliest.NotAfter) {
-			earliest = cert
+		// Expiry: keep the cert with the earliest NotAfter.
+		if now.After(cert.NotAfter) {
+			if earliestExpired == nil || cert.NotAfter.Before(earliestExpired.NotAfter) {
+				earliestExpired = cert
+			}
+		}
+		// Not-yet-valid: keep the cert with the NotBefore that is start date still in the future.
+		if now.Before(cert.NotBefore) {
+			if earliestNotYetValid == nil || cert.NotBefore.After(earliestNotYetValid.NotBefore) {
+				earliestNotYetValid = cert
+			}
 		}
 	}
 
-	if earliest == nil {
-		return nil
+	if earliestExpired != nil {
+		notAfter := earliestExpired.NotAfter.UTC().Format(time.RFC3339)
+		expiryCond = &metav1.Condition{
+			Type:    CACertificateExpiryConditionType,
+			Status:  metav1.ConditionTrue,
+			Reason:  "Expired",
+			Message: fmt.Sprintf("issuer CA certificate (CN=%s) expired at %s", earliestExpired.Subject.CommonName, notAfter),
+		}
 	}
 
-	// If the CA is not expired, return nil.
-	if !now.After(earliest.NotAfter) {
-		return nil
+	if earliestNotYetValid != nil {
+		notBefore := earliestNotYetValid.NotBefore.UTC().Format(time.RFC3339)
+		notYetValidCond = &metav1.Condition{
+			Type:    CACertificateNotYetValidConditionType,
+			Status:  metav1.ConditionTrue,
+			Reason:  "NotYetValid",
+			Message: fmt.Sprintf("issuer CA certificate (CN=%s) is not yet valid (NotBefore %s)", earliestNotYetValid.Subject.CommonName, notBefore),
+		}
 	}
 
-	notAfter := earliest.NotAfter.UTC().Format(time.RFC3339)
-	return &metav1.Condition{
-		Type:    CACertificateExpiryConditionType,
-		Status:  metav1.ConditionTrue,
-		Reason:  "Expired",
-		Message: fmt.Sprintf("issuer CA certificate (CN=%s) expired at %s", earliest.Subject.CommonName, notAfter),
-	}
+	return expiryCond, notYetValidCond
 }
--- a/backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
+++ b/backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
@@ -16,6 +16,8 @@ package status
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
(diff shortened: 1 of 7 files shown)


m-mandhare:
ok-to-test

internal/validation/validators.go
```diff
@@ -1047,3 +1054,106 @@ func ValidateNodePoolVersionChange(desiredVersion semver.Version, activeVersions
...
+
+		if now.Before(cert.NotBefore) {
+			notBefore := cert.NotBefore.UTC().Format(time.RFC3339)
+			errs = append(errs, field.Invalid(
+				fldPath,
+				diagnostic,
+				pemBlockDetail(blockIndex, subject, fmt.Sprintf("is not yet valid (NotBefore %s)", notBefore)),
+			))
+		}
+		if now.After(cert.NotAfter) {
```
machi1990, line 1139: I wonder if this should be more a validation that runs continuously akin to cluster validation?
for example, what if it hasn't expired now but it expires a month later after we've accepted it.

internal/validation/validators.go
```diff
@@ -1047,3 +1054,106 @@ func ValidateNodePoolVersionChange(desiredVersion semver.Version, activeVersions
...
+
+		if !cert.IsCA {
+			errs = append(errs, field.Invalid(
+				fldPath,
+				diagnostic,
+				pemBlockDetail(blockIndex, subject, "must be a CA certificate"),
+			))
+		}
+
+		if now.Before(cert.NotBefore) {
```
machi1990, line 1131: If a cert isn't usable until a certain time in the future, does OCP fails because of it?

machi1990: And the failure is a static validation or more a user facing condition if so, can we rely on surfacing it or do something similar to https://github.com/Azure/ARO-HCP/pull/6823/changes#r3949225272

internal/validation/validators.go
```diff
@@ -846,6 +849,10 @@ func ValidatePEM(ctx context.Context, op operation.Operation, fldPath *field.Pat
 		return field.ErrorList{field.Invalid(fldPath, *value, "not a valid PEM")}
 	}
```
JameelB: Probably don't need to iterate through the given pem twice? validateCACertificatePEM can probably also check if a cert is actually found and return an error if not?

JameelB, line 850: Probably don't need to iterate through the given pem twice? validateCACertificatePEM can probably also check if a cert is actually found and return an error if not?

internal/validation/validators.go
shitaljante: I think we have addressed this TODO. The comment should either be removed or updated to if anything still missing.

internal/validation/validators.go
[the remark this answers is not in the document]
shitaljante: This make sense, When a user submits a private key as the CA, This value can end up in ARM API responses, logs, and traces.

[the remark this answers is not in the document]
m-mandhare: Added in new changes. Now it gives error message with certificate CN or block no -        
 "message": "Invalid value: \"block 1\": PEM block 1 has type \"EC PRIVATE KEY\"; must be CERTIFICATE",

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -130,3 +148,52 @@ func (c *externalAuthDegradedAggregator) SyncOnce(ctx context.Context, key contr
...
+			break
+		}
+		if block.Type != "CERTIFICATE" {
+			continue
+		}
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			continue
+		}
+		if earliest == nil || cert.NotAfter.Before(earliest.NotAfter) {
```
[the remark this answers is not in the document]
m-mandhare: We are already checking this condition at admission level. Hence no need to check it here.

internal/validation/validators.go
m-mandhare: Added continuous validation with user facing condition to track expired CA after acceptance

internal/validation/validators.go
```diff
@@ -1047,3 +1045,105 @@ func ValidateNodePoolVersionChange(desiredVersion semver.Version, activeVersions
...
+				pemBlockDetail(blockIndex, subject, fmt.Sprintf("has expired (NotAfter %s)", notAfter)),
+			))
+		}
+	}
+
+	if !foundBlock {
+		return field.ErrorList{field.Invalid(fldPath, "", "not a valid PEM")}
+	}
+
+	if len(bytes.TrimSpace(rest)) > 0 {
```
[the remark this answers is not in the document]
m-mandhare: not adding as rest can have private key or sensitive data

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -33,6 +35,17 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
+const (
+	// CACertificateExpiryConditionType is the user-facing condition type set when the issuer
+	// CA certificate has expired. Written to Status.UserFacingConditions only while expired;
+	CACertificateExpiryConditionType = "CACertificateExpiry"
```
JameelB: Might be better to used 'Expired' here as that indicates the current state of the ca. Expiry to me refers to a description of the date/time when the ca is/will expire.

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -134,3 +160,71 @@ func (c *externalAuthDegradedAggregator) SyncOnce(ctx context.Context, key contr
...
+
+// caCertValidityConditions inspects all CERTIFICATE PEM blocks in ca and returns
+// two conditions: expiryCond (CACertificateExpiry) and notYetValidCond (CACertificateNotYetValid).
+// Each is nil when not applicable. Both are derived from a single pass over the PEM bundle:
+//   - expiryCond is set when any cert has expired; the cert with the earliest NotAfter drives
+//     the condition message.
+//   - notYetValidCond is set when any cert has a start date in the future;
+//
+// Missing, empty, or entirely unparseable input causes both to return nil.
+func caCertValidityConditions(ca string, now time.Time) (expiryCond, notYetValidCond *metav1.Condition) {
```
JameelB: nit: this reads like a property, i think it would read better as if it was an action since it's performing something, in this case a check. Something like isCaCertValid(), getCaCertValidityConditions() would read better imo. 

Also why pass the time?

JameelB: expiryCond -> expiredCondition

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -134,3 +160,71 @@ func (c *externalAuthDegradedAggregator) SyncOnce(ctx context.Context, key contr
...
+		}
+	}
+
+	if earliestExpired != nil {
+		notAfter := earliestExpired.NotAfter.UTC().Format(time.RFC3339)
+		expiryCond = &metav1.Condition{
+			Type:    CACertificateExpiryConditionType,
+			Status:  metav1.ConditionTrue,
+			Reason:  "Expired",
+			Message: fmt.Sprintf("issuer CA certificate (CN=%s) expired at %s", earliestExpired.Subject.CommonName, notAfter),
```
JameelB:
```suggestion
			Message: fmt.Sprintf("CA certificate (CN=%s) expired at %s", earliestExpired.Subject.CommonName, notAfter),
```

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -134,3 +160,71 @@ func (c *externalAuthDegradedAggregator) SyncOnce(ctx context.Context, key contr
...
+		}
+	}
+
+	if earliestNotYetValid != nil {
+		notBefore := earliestNotYetValid.NotBefore.UTC().Format(time.RFC3339)
+		notYetValidCond = &metav1.Condition{
+			Type:    CACertificateNotYetValidConditionType,
+			Status:  metav1.ConditionTrue,
+			Reason:  "NotYetValid",
+			Message: fmt.Sprintf("issuer CA certificate (CN=%s) is not yet valid (NotBefore %s)", earliestNotYetValid.Subject.CommonName, notBefore),
```
JameelB:
```suggestion
			Message: fmt.Sprintf("CA certificate (CN=%s) is not yet valid (NotBefore %s)", earliestNotYetValid.Subject.CommonName, notBefore),
```

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -33,6 +35,17 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
+const (
+	// CACertificateExpiryConditionType is the user-facing condition type set when the issuer
```
JameelB:
```suggestion
	// CACertificateExpiryConditionType is the user-facing condition type set when the
```

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -33,6 +35,17 @@ import (
...
+	// CA certificate has expired. Written to Status.UserFacingConditions only while expired;
+	CACertificateExpiryConditionType = "CACertificateExpiry"
+
+	// CACertificateNotYetValidConditionType is the user-facing condition type set when the
+	// issuer CA certificate has a NotBefore in the future. A future NotBefore is accepted at
```
JameelB:
```suggestion
	// CA certificate has a NotBefore in the future. A future NotBefore is accepted at
```

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -134,3 +160,71 @@ func (c *externalAuthDegradedAggregator) SyncOnce(ctx context.Context, key contr
...
+		}
+		// Not-yet-valid: keep the cert with the NotBefore that is start date still in the future.
+		if now.Before(cert.NotBefore) {
+			if earliestNotYetValid == nil || cert.NotBefore.After(earliestNotYetValid.NotBefore) {
+				earliestNotYetValid = cert
+			}
+		}
+	}
+
+	if earliestExpired != nil {
```
JameelB: is there a way to surface all the expired/not yet valid certs in the chain? specifying only one may lead to a user thinking that's the only one that is expired/not valid. Once that specific one is fixed, then a new error could surface. It doesn't really matter which ones the earliest to expire/latest to be valid, once one of them isn't valid the whole chain breaks. 

If not, a generic message saying the ca cert is expired/not valid would probably do? Something like 'a certificate in the chain is expired'/'a certificate in the chain is not yet valid', wdyt?

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -130,3 +161,78 @@ func (c *externalAuthDegradedAggregator) SyncOnce(ctx context.Context, key contr
...
+
+	return expiryCond, notYetValidCond
+}
+
+func formatCACertExpiredMessage(certs []*x509.Certificate) string {
+	parts := make([]string, len(certs))
+	for i, cert := range certs {
+		parts[i] = fmt.Sprintf("CN=%s (NotAfter %s)", cert.Subject.CommonName, cert.NotAfter.UTC().Format(time.RFC3339))
+	}
+	return fmt.Sprintf("%d CA certificate(s) in the issuer bundle have expired: %s", len(certs), strings.Join(parts, "; "))
```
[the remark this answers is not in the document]
m-mandhare: In Error message only CN name and date is added. Hence it will not full the capacity

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
```diff
@@ -113,7 +127,24 @@ func (c *externalAuthDegradedAggregator) SyncOnce(ctx context.Context, key contr
 
 	replacement := existing.DeepCopy()
 	apimeta.SetStatusCondition(&replacement.Status.Conditions, aggregated)
-	if equality.Semantic.DeepEqual(existing.Status.Conditions, replacement.Status.Conditions) {
+
+	// CACertificateExpiry is user-facing only when the CA has expired.
+	// CACertificateNotYetValid is user-facing only when the CA has a start date in the future.
+	expiryCond, notYetValidCond := c.getCaCertValidityConditions(existing.Properties.Issuer.CA)
```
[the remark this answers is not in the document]
m-mandhare: Updated doc - cosmos-data-flow.md

backend/pkg/controllers/externalauth/status/externalauth_degraded_aggregator.go
machi1990, line 1: The aggregator shouldn't be modified; this inspect all controllers and writes the conditions onto the `HCPOpenShiftClusterExternalAuth.Status.Conditions` 
 
We need a different aggregator: see https://github.com/shitaljante/ARO-HCP/blob/a29501510499ea976e8117ecc5a97648985458ef/backend/pkg/controllers/externalauth/status/externalauth_userfacing_aggregator.go that @shitaljante is introducing in https://github.com/Azure/ARO-HCP/pull/6746 and add this logic there.

It's critical because we don't want two controllers writing on a single field; potentially racing each other.

machi1990 (changes requested):

PR #6826, opened 2026-09-04
fix: align alertEvents KQL column order with deployed Kusto schema

inbharajmani:
https://redhat.atlassian.net/browse/AROSLSRE-2028

## Summary
- The `alertEvents` table was originally created with `alertContext` as the last column. When `region` was added in [6fbf78db5](https://github.com/Azure/ARO-HCP/commit/6fbf78db5), it was placed **before** `alertContext` in the KQL source, but `.create-merge table` only appends new columns — Kusto added `region` **after** `alertContext` in the deployed schema.
- This column order mismatch causes the update policy to fail with `ScriptExecutionFailed: Query schema does not match table schema`, breaking the `global-pipeline-postsubmit` job ([failing run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/branch-ci-Azure-ARO-HCP-main-global-pipeline-postsubmit/2095742533232496640)).
- Swaps `alertContext` and `region` in both the table definition and the transform function to match the actual deployed order (confirmed via Kusto schema query against dev).

## Test plan
- [x] Verify `make verify-kql` passes
- [ ] `global-pipeline-postsubmit` job succeeds after merge

### DIFF
--- a/dev-infrastructure/modules/logs/kusto/tables/alertEvents.kql
+++ b/dev-infrastructure/modules/logs/kusto/tables/alertEvents.kql
@@ -25,8 +25,8 @@ with (docstring = "Staging table for metric alert events ingested from Event Hub
     resolvedDateTime: datetime,
     targetResourceId: string,
     targetResourceType: string,
-    region: string,
-    alertContext: dynamic
+    alertContext: dynamic,
+    region: string
 )
 with (docstring = "Alert events with top-level Common Alert Schema fields extracted")
 
@@ -46,8 +46,8 @@ rawAlertEventsTransform() {
         resolvedDateTime = todatetime(essentials.resolvedDateTime),
         targetResourceId = tostring(essentials.alertTargetIDs[0]),
         targetResourceType = tostring(essentials.targetResourceType),
-        region = tostring(raw.data.alertContext.labels.region),
-        alertContext = raw.data.alertContext
+        alertContext = raw.data.alertContext,
+        region = tostring(raw.data.alertContext.labels.region)
 }
 
 .alter table alertEvents policy update @'[{"IsEnabled": true, "Source": "rawAlertEvents", "Query": "rawAlertEventsTransform()", "IsTransactional": true, "PropagateIngestionProperties": false}]'


raelga (approved):
/lgtm
/approve

raelga:
**Force-merge justification**

Note up front: this is a forward fix, not a revert. A revert is not the correct remedy here (see criterion 2), so this is the green-but-blocked case where the only thing keeping a demonstrably-green, minimal fix out of `main` is unrelated `e2e-parallel` churn in tide. It meets all four criteria from the alert-latch thread:

**1. INT and the global deployment pipeline are blocked by the problem.**
The INT Kusto deploy is failing outright. The `kusto-westus3` deployment to `hcp-kusto-int-us` (cluster `hcp-int-us`, database `MonitoringEvents`) terminates in `ScriptExecutionFailed` on the `alertEvents` update policy:

> Reason: Error during execution of a policy operation: Query schema does not match table schema. QuerySchema=('...,string,string,string,dynamic'), TableSchema=('...,string,string,dynamic,string').

The same failure reddens `global-pipeline-postsubmit` on `main`: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/branch-ci-Azure-ARO-HCP-main-global-pipeline-postsubmit/2095742533232496640 . INT observability stays broken and every postsubmit stays red until this lands.

**2. High confidence the fix resolves the problem, and a revert would not.**
Root cause is visible directly in the schema mismatch above: the deployed table ends `...alertContext (dynamic), region (string)`, while the KQL source query produces `...region (string), alertContext (dynamic)`. When `region` was added in commit 6fbf78db5 it was placed before `alertContext` in the source, but Kusto's `.create-merge table` only appends new columns, so the live schema has `region` last. This PR swaps the two columns in both the table definition and the transform function to match the deployed order (confirmed via a Kusto schema query against dev). A revert is not viable: the live schema already has `region` appended, so reverting the source would not restore the deployed column order; only a forward alignment fixes it.

**3. The fix builds and passes all fast testing.**
`make verify-kql` passes and all fast checks are green (verify, lint, bicep-lint, integration, test-unit, images, e2e-images, config-change-detection, mega-linter). The only non-green contexts are `e2e-parallel` (pending) and `tide`. The change is a 4-line column swap in a single file (`dev-infrastructure/modules/logs/kusto/tables/alertEvents.kql`) with no relationship to e2e, so the E2E churn is provably unrelated to this change.

**4. The change is clean.**
Minimal, single-file, +4/-4 column-order swap; lgtm + approved. No mixed concerns.

The durable follow-up (making the KQL source and `.create-merge` ordering drift-proof) can go through the queue normally.

PR #6837, opened 2026-09-04
fix(e2e test): make >= version selection compatible with nightly runs

mbukatov:
https://redhat.atlassian.net/browse/ARO-29462

This is just a quick draft how a fix could work, so that I can revisit it with more care next week.

### What

Make sure E2E tests which require a particular minimal version (eg. OCP >= 4.22) are either skipped or uses a latest version when executed in a nightly run job.

So in this PR, we:

- define new `ErrNightlyVersionTooOld` error and related `IsIncompatibleNightlyVersionError` function 
- define new funtion `PickAtLeastOpenshiftVersionId(defaultVersion, minimalVersion string) (string, error)` which returns default version if it satisfies the minimal version, otherwise a minimal is returned, with one exception: for nightly builds, `ErrNightlyVersionTooOld` is returned instead
- implement unit tests for `PickAtLeastOpenshiftVersionId`
- update E2E tests which defines particular version:
    - cluster_create_private_kas.go
    - kms_key_rotation.go
 

### Why

Current design has a bug:

- First, a test case asks for defaults via `NewDefaultClusterParams*` functions, eg. `clusterParams := framework.NewDefaultClusterParams20251223()` so that the test has this version available in `clusterParams.OpenshiftVersionId`.
- If it's a nightly run (env variable `ARO_HCP_OPENSHIFT_CHANNEL_GROUP` is `nightly`), `DefaultOpenshiftControlPlaneVersionId` function will fill in a proper nightly version id
- But if an E2E test case then redefines `clusterParams.OpenshiftVersionId` (because it needs at least 4.22 and sets it to 4.22 because of it), we have 2 problems: we don't use nightly build anymore, and we are passing only x.y version when using nightly channel. So that this can fail either because of wrong version passed, or because of missing version.

### Testing

justification.

- [X] Unit tests: unit tests created for new `PickAtLeastOpenshiftVersionId` function
- [Integration tests](https://github.com/Azure/ARO-HCP/tree/main/test-integration): NA
- [E2E tests](https://github.com/Azure/ARO-HCP/blob/main/test/e2e/README.md):
     - [X] rerun of pre check e2e run in dev (which is mandatory)
     - [X] run of a nightly build with smaller default OCP version than the KAS test requires: [pull-ci-Azure-ARO-HCP-main-prod-e2e-parallel-ocp-nightly #2098771064220815360](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6837/pull-ci-Azure-ARO-HCP-main-prod-e2e-parallel-ocp-nightly/2098771064220815360) (test case skipped with a clear reason when run on nightly `4.21.0-0.nightly-multi-2026-09-12-094059` while requiring at least 4.22)

### PR Checklist

- [X] PR is scoped to a single task (no mixed concerns)
- [X] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [X] Summary explains the "Why" behind the change
- [X] Linked to relevant ticket/issue
- [x] Screenshots included (if dashboards or other UI changes): NA
- [X] Self-reviewed the diff
- [X] Draft PR used for WIP (if applicable)
- [X] Tricky code blocks are commented

If E2E tests are included:

- [X] E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md)
- [X] If new E2E use case is covered (via a new test or new check/verifier),
  proper error message and logs which communicates nature of the problem:
  - [pull-ci-Azure-ARO-HCP-main-prod-e2e-parallel-ocp-nightly/2098541426177478656](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6837/pull-ci-Azure-ARO-HCP-main-prod-e2e-parallel-ocp-nightly/2098541426177478656)
  - [pull-ci-Azure-ARO-HCP-main-prod-e2e-parallel-ocp-nightly #2098771064220815360](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6837/pull-ci-Azure-ARO-HCP-main-prod-e2e-parallel-ocp-nightly/2098771064220815360) (test case skipped with a clear reason when run on nightly `4.21.0-0.nightly-multi-2026-09-12-094059` while requiring at least 4.22)

### DIFF
--- a/test/e2e/cluster_create_private_kas.go
+++ b/test/e2e/cluster_create_private_kas.go
@@ -75,10 +75,6 @@ var _ = Describe("Customer", func() {
 			)
 			Expect(err).NotTo(HaveOccurred(), "failed to create customer resources for private KAS cluster")
 
-			By("deploying test VM in customer VNet")
-			vmName, _, err := tc.DeployTestVM(ctx, TestArtifactsFS, *resourceGroup.Name, customerClusterName, clusterParams.VnetName, clusterParams.SubnetName)
-			Expect(err).NotTo(HaveOccurred(), "failed to deploy test VM for private KAS verification")
-
 			By("creating the HCP cluster with private KAS")
 			err = tc.CreateHCPClusterFromParam20251223(ctx,
 				GinkgoLogr,
@@ -88,6 +84,10 @@ var _ = Describe("Customer", func() {
 				framework.ClusterCreationTimeout,
 			)
 			Expect(err).NotTo(HaveOccurred(), "failed to create HCP cluster %q with private KAS", customerClusterName)
+
+			By("deploying test VM in customer VNet")
+			vmName, _, err := tc.DeployTestVM(ctx, TestArtifactsFS, *resourceGroup.Name, customerClusterName, clusterParams.VnetName, clusterParams.SubnetName)
+			Expect(err).NotTo(HaveOccurred(), "failed to deploy test VM for private KAS verification")
 
 			By("verifying cluster API visibility is Private via ARM GET")
 			hcpClient := tc.Get20251223ClientFactoryOrDie(ctx).NewHcpOpenShiftClustersClient()
--- a/test/e2e/cluster_create_private_kas.go
+++ b/test/e2e/cluster_create_private_kas.go
@@ -65,9 +65,13 @@ var _ = Describe("Customer", func() {
 
 			// Private KAS requires OCP >= 4.22 (CS validation rejects lower versions)
 			openshiftVersionID, err := framework.PickAtLeastOpenshiftVersionId(clusterParams.OpenshiftVersionId, "4.22")
-			// If the default is nightly which isn't >= 4.22, skip this test with an explanation
+			// If the default is nightly which isn't >= 4.22, skip this test with an explanation.
+			// Log the reason before calling Skip: Skip() stores its message internally and the
+			// Ginkgo reporter only emits file:line in verbose mode, never the message text itself.
 			if framework.IsIncompatibleNightlyVersionError(err) {
-				Skip(fmt.Sprintf("private KAS requires OCP >= 4.22, but default version %q does not satisfy it: %v", clusterParams.OpenshiftVersionId, err))
+				skipMsg := fmt.Sprintf("private KAS requires OCP >= 4.22, but default version %q does not satisfy it: %v", clusterParams.OpenshiftVersionId, err)
+				GinkgoLogr.Info(skipMsg)
+				Skip(skipMsg)
 			}
 			Expect(err).NotTo(HaveOccurred(), "failed to select OpenShift version >= 4.22 for private KAS test (default version: %q)", clusterParams.OpenshiftVersionId)
 			// Otherwise, just use the selected version
--- a/test/e2e/cluster_create_private_kas.go
+++ b/test/e2e/cluster_create_private_kas.go
@@ -62,8 +62,16 @@ var _ = Describe("Customer", func() {
 			clusterParams.ClusterName = customerClusterName
 			clusterParams.ManagedResourceGroupName = framework.SuffixName(*resourceGroup.Name, "-managed", 64)
 			clusterParams.APIVisibility = "Private"
+
 			// Private KAS requires OCP >= 4.22 (CS validation rejects lower versions)
-			clusterParams.OpenshiftVersionId = "4.22"
+			openshiftVersionID, err := framework.PickAtLeastOpenshiftVersionId(clusterParams.OpenshiftVersionId, "4.22")
+			// If the default is nightly which isn't >= 4.22, skip this test with an explanation
+			if framework.IsIncompatibleNightlyVersionError(err) {
+				Skip(fmt.Sprintf("private KAS requires OCP >= 4.22, but default version %q does not satisfy it: %v", clusterParams.OpenshiftVersionId, err))
+			}
+			Expect(err).NotTo(HaveOccurred(), "failed to select OpenShift version >= 4.22 for private KAS test (default version: %q)", clusterParams.OpenshiftVersionId)
+			// Otherwise, just use the selected version
+			clusterParams.OpenshiftVersionId = openshiftVersionID
 
 			By("creating customer resources (infrastructure and managed identities)")
 			clusterParams, err = tc.CreateClusterCustomerResources20251223(ctx,
--- a/test/util/framework/version_helper.go
+++ b/test/util/framework/version_helper.go
@@ -189,7 +189,7 @@ func getLatestInstallVersionForNightlyChannel(ctx context.Context, version strin
 // default and given minimal version constraint arguments.
 //   - If defaultVersion already satisfies the minimalVersion, the default is
 //     returned unchanged.
-//   - If it doesn't, the mimimal version is returned instead, with one
+//   - If it doesn't, the minimal version is returned instead, with one
 //     exception: For nightly builds, an error is returned instead (since
 //     nightly versions cannot be bumped to a different minor version).
 //
--- a/test/util/framework/version_helper.go
+++ b/test/util/framework/version_helper.go
@@ -37,6 +37,7 @@ var (
 	ErrNightlyReleaseStreamNotFound = errors.New("nightly release stream not found")
 	ErrNoAcceptedNightlyTags        = errors.New("no accepted nightly tags found")
 	ErrNoParseableNightlyTags       = errors.New("no parseable nightly tags found")
+	ErrNightlyVersionTooOld         = errors.New("nightly version is too old")
 )
 
 const (
@@ -66,6 +67,12 @@ func retryOnTransientError[T any](ctx context.Context, f func() (T, error)) (T, 
 		}
 	}
 	return zero, fmt.Errorf("after %d attempts: %w", versionFetchMaxRetries+1, lastErr)
+}
+
+// IsIncompatibleNightlyVersionError returns true if the error indicates that
+// a nightly version doesn't satisfy given constraints.
+func IsIncompatibleNightlyVersionError(err error) bool {
+	return errors.Is(err, ErrNightlyVersionTooOld)
 }
 
 // IsVersionNotFoundError returns true if the error indicates a version was not found,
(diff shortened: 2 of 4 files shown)


test/e2e/cluster_create_private_kas.go
```diff
@@ -62,8 +62,17 @@ var _ = Describe("Customer", func() {
 			clusterParams.ClusterName = customerClusterName
 			clusterParams.ManagedResourceGroupName = framework.SuffixName(*resourceGroup.Name, "-managed", 64)
 			clusterParams.APIVisibility = "Private"
+
 			// Private KAS requires OCP >= 4.22 (CS validation rejects lower versions)
-			clusterParams.OpenshiftVersionId = "4.22"
+			openshiftVersionID, err := framework.PickLatestOpenshiftVersionId(clusterParams.OpenshiftVersionId, "4.22")
```
miquelsi: This should apply also to node pool version selection

mbukatov: Good point.

mbukatov: I was updating the affected E2E tests, and noticed that the tests didn't touched the node pool version at all. So I didn't changed that.

test/util/framework/version_helper.go
```diff
@@ -177,3 +184,46 @@ func getLatestInstallVersionForNightlyChannel(ctx context.Context, version strin
...
+	defaultSemver, err := semver.ParseTolerant(defaultVersion)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse default version %q: %w", defaultVersion, err)
+	}
+	minimalSemver, err := semver.ParseTolerant(minimalVersion)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse minimal version %q: %w", minimalVersion, err)
+	}
+
+	if strings.Contains(defaultVersion, "nightly") {
```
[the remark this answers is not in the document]
mbukatov: Nightly version should be recognizable from the version itself. No need to take the channel into account from this perspective.

test/util/framework/version_helper.go
```diff
@@ -177,3 +184,46 @@ func getLatestInstallVersionForNightlyChannel(ctx context.Context, version strin
 
 	return latestTagName, nil
 }
+
+// PickAtLeastOpenshiftVersionId selects latest version based on a predefined
+// default and given minimal version constraint arguments.
+//   - If defaultVersion already satisfies the minimalVersion, the default is
+//     returned unchanged.
+//   - If it doesn't, the mimimal version is returned instead, with one
```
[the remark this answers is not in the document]
mbukatov: fixed in 89088ae3d

mbukatov:
So, looking at the results of [prod-e2e-parallel-ocp-nightly](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6837/pull-ci-Azure-ARO-HCP-main-prod-e2e-parallel-ocp-nightly/2098541426177478656), I see:

Test case *Customer should create a cluster with private KAS and verify API server is only reachable from VNet while ingress remains public* was skipped:

```
  STEP: creating cluster parameters with private API visibility @ 09/11/26 22:54:48.138
  [SKIPPED] in [It] - /opt/app-root/src/github.com/Azure/ARO-HCP/test/e2e/cluster_create_private_kas.go:70 @ 09/11/26 22:54:48.182
"ts"="2026-09-11 22:54:48.182641" "level"=0 "msg"="===== TEST CASE ENDED: SUCCESS ====="
```

While the skip message doesn't provide enough detail I hoped, it works as expected since this nightly run is using `4.21.0-0.nightly-multi-2026-09-11-175229` as can be seen in other runs, eg.:

```
"ts"="2026-09-11 22:57:29.553673" "level"=0 "msg"="Starting HCP cluster creation (v20251223preview)" "clusterName"="cilium-cluster" "resourceGroup"="complex-cilium-kv-6l7dfsdwsr8l" "version"="4.21.0-0.nightly-multi-2026-09-11-175229" "channelGroup"="nightly"
```

So we can see, that when a test case require OCP 4.22 via `openshiftVersionID, err := framework.PickAtLeastOpenshiftVersionId(clusterParams.OpenshiftVersionId, "4.22")`, error is returned so that `framework.IsIncompatibleNightlyVersionError(err)` is true and the test case is indeed skipped.

Fixed in [1dc8c59](https://github.com/Azure/ARO-HCP/pull/6837/commits/1dc8c593a6ee7eaba9fe2924e2b09dc6476e485d).

mbukatov:
Good, in run [2098771064220815360](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/Azure_ARO-HCP/6837/pull-ci-Azure-ARO-HCP-main-prod-e2e-parallel-ocp-nightly/2098771064220815360) I see that the feature works as intended.

When I look up *Customer should create a cluster with private KAS and verify API server is only reachable from VNet while ingress remains public* among skipped cases, in the log I see:

```
"ts"="2026-09-12 14:07:05.960056" "level"=0 "msg"="private KAS requires OCP >= 4.22, but default version \"4.21.0-0.nightly-multi-2026-09-12-094059\" does not satisfy it: nightly version is too old: nightly build 4.21.0-0.nightly-multi-2026-09-12-094059 does not satisfy minimum 4.22"
  [SKIPPED] in [It] - /opt/app-root/src/github.com/Azure/ARO-HCP/test/e2e/cluster_create_private_kas.go:74 @ 09/12/26 14:07:05.96
```

test/util/framework/version_helper.go
```diff
@@ -177,3 +184,46 @@ func getLatestInstallVersionForNightlyChannel(ctx context.Context, version strin
...
+	defaultSemver, err := semver.ParseTolerant(defaultVersion)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse default version %q: %w", defaultVersion, err)
+	}
+	minimalSemver, err := semver.ParseTolerant(minimalVersion)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse minimal version %q: %w", minimalVersion, err)
+	}
+
+	if strings.Contains(defaultVersion, "nightly") {
```
machi1990, line 209: candidate channel also can have pre-release suffix e.g `rc` or `ec`.

Instead of doing it like this, you probably want `len(defaultVersion.Pre) > 0 `

test/e2e/cluster_create_private_kas.go
```diff
@@ -62,8 +62,20 @@ var _ = Describe("Customer", func() {
 			clusterParams.ClusterName = customerClusterName
 			clusterParams.ManagedResourceGroupName = framework.SuffixName(*resourceGroup.Name, "-managed", 64)
 			clusterParams.APIVisibility = "Private"
+
 			// Private KAS requires OCP >= 4.22 (CS validation rejects lower versions)
-			clusterParams.OpenshiftVersionId = "4.22"
+			openshiftVersionID, err := framework.PickAtLeastOpenshiftVersionId(clusterParams.OpenshiftVersionId, "4.22")
```
machi1990, line 67: If the pipeline is configured to run with 4.20 version, do we want to run tests that can only be run if the version is higher than the one the pipeline was configured with?

mbukatov: If the pipeline is configured to run with 4.20 version, this PR makes sure that the test which requires at least 4.22 (via `framework.PickAtLeastOpenshiftVersionId`) will get skipped, with an explanation logged.

In the same way, as it happens when you ]run nightly job with `4.21.0-0.nightly-multi-2026-09-12-094059`](https://github.com/Azure/ARO-HCP/pull/6837#issuecomment-5647255345).

test/util/framework/version_helper.go
mbukatov: Good point. Will check what other pre release versions we can run with.

test/e2e/cluster_create_private_kas.go
```diff
@@ -75,10 +87,6 @@ var _ = Describe("Customer", func() {
 			)
 			Expect(err).NotTo(HaveOccurred(), "failed to create customer resources for private KAS cluster")
 
-			By("deploying test VM in customer VNet")
```
patriksuba: Why deploy the test VM after the cluster is created?

test/e2e/cluster_create_private_kas.go
machi1990:
```
	if framework.IsIncompatibleNightlyVersionError(err) {
				skipMsg := fmt.Sprintf("private KAS requires OCP >= 4.22, but default version %q does not satisfy it: %v", clusterParams.OpenshiftVersionId, err)
				GinkgoLogr.Info(skipMsg)
				Skip(skipMsg)
			}
```

This is done for nightly, my question was more a general one, especially considering that we've a code path 
```
	if defaultSemver.GTE(minimalSemver) {
		return defaultVersion, nil
	}
	return minimalVersion, nil
```

That returns a minimal version.

test/e2e/cluster_create_private_kas.go
patriksuba, line 78: Why deploy the test VM after the cluster is created?

mbukatov: It's mentioned in the commit message. In case the cluster fails to deploy or we skip the test case, we are not wasting time and resources on a test VM.

test/e2e/cluster_create_private_kas.go
mbukatov: Oh, you are right, I was confused a bit.

The current behaviour drafted in the PR is intentional. When the test requires at least 4.22 but we run with 4.20 by default, we really want the test to run with 4.22, assuming it's available.

This is because we want to preserve the original behaviour, when a test case specified version as 4.22 directly because the default was <=4.21, and the test required 4.22.

We don't currently have a standardized way to skip a test case when OCP version it relies on is not yet available. That is something I find useful, but it would require more thoughts and would be out of scope of the original fix. Eg. if one wrote such a test and then the test were always skipped, there would be no way for us to see the results until we create a separate run with a new version.

test/util/framework/version_helper.go
mbukatov: Addressed in 56cbfad27de35af6dc91d7b62fb2fefb9ef6068f

test/e2e/cluster_create_private_kas.go
```diff
@@ -62,8 +62,20 @@ var _ = Describe("Customer", func() {
...
 			clusterParams.ManagedResourceGroupName = framework.SuffixName(*resourceGroup.Name, "-managed", 64)
 			clusterParams.APIVisibility = "Private"
+
 			// Private KAS requires OCP >= 4.22 (CS validation rejects lower versions)
-			clusterParams.OpenshiftVersionId = "4.22"
+			openshiftVersionID, err := framework.PickAtLeastOpenshiftVersionId(clusterParams.OpenshiftVersionId, "4.22")
+			// If the default is nightly which isn't >= 4.22, skip this test with an explanation.
+			// Log the reason before calling Skip: Skip() stores its message internally and the
+			// Ginkgo reporter only emits file:line in verbose mode, never the message text itself.
+			if framework.IsIncompatibleNightlyVersionError(err) {
```
mgahagan73, line 71: By the time we get to this version check, we have already created several resources (MI, Resource group) which we have to clean up if we skip the test because we cannot find a good version. Could you move this check so it runs before we create any resources so there is no risk of leaving orphan resources as well as slightly improve execution time in cases where the test skips?

mbukatov: That makes sense, will do that.

test/util/framework/version_helper.go
```diff
@@ -177,3 +184,46 @@ func getLatestInstallVersionForNightlyChannel(ctx context.Context, version strin
...
+	if err != nil {
+		return "", fmt.Errorf("failed to parse minimal version %q: %w", minimalVersion, err)
+	}
+
+	if len(defaultSemver.Pre) > 0  {
+		// For nightly builds, ignore the pre-release suffix and compare Major.Minor.Patch
+		// directly. Bare semver ordering would rank "4.19.0-0.nightly-multi-..." below
+		// "4.19.0", producing false negatives for patch-zero minimums. Nightly builds are
+		// always patch 0 (X.Y.0-0.nightly-multi-...), so a patch-qualified minimum such
+		// as "4.22.1" correctly rejects a 4.22.0 nightly.
```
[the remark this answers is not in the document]
mbukatov: This is intended behaviour, when we run with nightlies, we want to run with nightlies only.

test/e2e/cluster_create_private_kas.go
mbukatov: When I think about it, I don't think we need to do that, the resource group itself we would spare is empty container at this point, and MI's are created already, the test will just return it to a pool quickly. I don't think we need to change the tests in this way.

mgahagan73: At any rate, changing the ordering would only make a small difference and I think we can address it in a follow-up

mgahagan73:
Since this PR should improve CI reliability and the identified issues can be addressed in follow-up I'm ok with approving it at this point.

/approve

mvacula02 (approved):
/lgtm

PR #6838, opened 2026-09-04
fix(test-integration): honor Retry-After HTTP headers on GET requests.  

bennerv:
### What

Fixes flaky integration admin tests. 
### Why

The breakglass admin integration test was flaky under CPU load. kubernetesApply writes to the fake object tracker synchronously, but the handler reads the session through an informer lister cache updated asynchronously by the watch goroutine. An immediate read-after-write in the next step could observe the stale (not-ready) session, so step 6 got "Session is not ready" instead of the kubeconfig. The handler only falls back to the live client on NotFound, and the fake tracker does not bump resourceVersion, so nothing bridged the gap.

Similar to our GAs, let's honor the retry-after header if the session isn't immediately ready to allow time for the lister cache to get updated by the informer.  

### Testing
Used the script to replicate, no longer happening after the change.  

```bash
#!/bin/bash

NCPU=$(nproc); echo "saturating $NCPU cores"
for c in $(seq 1 $NCPU); do ( while :; do :; done ) & done
HOGS=$(jobs -p)

# background churn: keep re-running the mock suite to stir the scheduler
( for r in $(seq 1 60); do go test ./test-integration/admin/ -run 'TestAdminCRUD/WithMock' -count=1 >/dev/null 2>&1; done ) &
CHURN=$!

for i in $(seq 1 25); do
  if ! go test ./test-integration/admin/ -run 'TestAdminCRUD/WithMock/HCP/breakglass$' -count=4 -race >/tmp/bg_$i.log 2>&1; then
    echo "=== FAILED on iteration $i ==="
    grep -E "FAIL|Session is not ready|--- FAIL" /tmp/bg_$i.log | head -15
    break
  else
    echo "iter $i: PASS"
  fi
done

# ALWAYS clean up the busy-loops afterwards:
kill $HOGS 2>/dev/null; kill $CHURN 2>/dev/null; wait 2>/dev/null
```

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


test-integration/utils/databasemutationhelpers/http_test_accessor.go
```diff
@@ -30,7 +30,18 @@ import (
 	"github.com/Azure/ARO-HCP/internal/utils"
 )
 
+// GetResponse is the concrete value returned (as any) by HTTPTestAccessor.Get:
+// the decoded response body plus the response headers, so callers can honor
+// server hints such as Retry-After. Implementations that cannot surface headers
+// (e.g. SDK-backed clients) leave Header nil.
+type GetResponse struct {
```
bennerv, line 37: This kind of sucks, but it allows us to maintain the `any` response from the Get interface.  We could update the Get to return a raw `http.Response` and do the handling caller-side but this is less intrusive and feedback is relatively immediate.  

Open to changing it though.

PR #6840, opened 2026-09-04
chore: remove one-shot legacy kusto-monitoring detach step (ARO-28726)

cssjr:
## Why

Follow-up cleanup for [ARO-28726](https://issues.redhat.com/browse/ARO-28726).

PR #5967 moved the `kusto-monitoring` ARMStack from Monitoring (region) scope to Geography scope. To avoid `DeploymentStackResourceManagedByAnotherStack` during that migration, it added a **one-time** migration step, `detach-legacy-monitoring-stack`, that detaches the legacy `Monitoring.kusto.kusto-monitoring` deployment stack (`az stack group delete --action-on-unmanage detachAll`) so the new Geography-scope stack can adopt its resources. The step is idempotent and, once the legacy stack no longer exists, a permanent no-op.

That migration has now rolled all the way through to **[PROD](https://releases.dev.aro.azure-test.net/releases/versions?mode=pr&repo=aro-hcp&pr=5967)** (INT → STG → PROD), so the legacy stack is detached everywhere and the step is dead weight. This removes it so it does not linger in the pipeline.

## What

- Remove the `detach-legacy-monitoring-stack` Shell step from `geography-pipeline.yaml` and `geography-pipeline-stg.yaml`.
- Remove the now-unnecessary `dependsOn` on that step from the `kusto-monitoring` step in both files.
- Delete `dev-infrastructure/scripts/detach-legacy-monitoring-stack.sh`.

## Testing

- `make validate-config-pipelines` passes — both pipelines still parse and the dependency graph is consistent (`kusto-monitoring` now depends only on `kusto-lookup` and `action-groups`).
- No remaining references to the step or script anywhere in the repo.

No test is added: this removes a one-shot migration step that is already a permanent no-op, and a test asserting its absence would itself be dead weight to remove later.


cssjr:
@roivaz had to rebase which reset your `lgtm`... could I get another one from you?  Thanks!

PR #6848, opened 2026-09-07
fix(backend): don't gate node pool CS delete on ApplyDesire purge

nimrodshn:
### What

Don't gate node pool CS delete on ApplyDesire purge; Dispatch `CS.DeleteNodePool` immediately on `DeletionTimestamp`. 

### Why

NodePoolClusterServiceDeleteDispatch refused to call CS.DeleteNodePool
until every kube-applier ApplyDesire tagged for the NodePool had been
purged from Cosmos. Those two conditions cannot both be satisfied.

On a per-node-pool ARM DELETE, ClusterResourcesController drops the
NodePool from its desired set and flips the ApplyDesire to Type=Delete.
kube-applier deletes the HyperShift NodePool CR, but Cluster Service
still owns the ManifestWork that materializes it, so the work-agent
re-applies the CR a couple of minutes later. kube-applier goes back to
SuccessfullyDeleted=False (WaitingForDeletion), the desire never
converges, and so it is never purged -- meaning the CS delete that
would have removed the ManifestWork is never dispatched. Observed as
the same NodePool CR recreated under four UIDs over 25 minutes, broken
only by the cluster teardown at the end of the E2E run.

Remove the gate: dispatch now fires as soon as DeletionTimestamp is
set. Teardown ordering is still enforced downstream --
NodePoolChildResourcesCleanupController removes the node-pool-scoped
desires only once Cluster Service confirms the NodePool is gone, and
NodePoolDeletionController will not delete the Cosmos NodePool document
while any ApplyDesire remains.

Also adds coverage for the node-pool branch of processClusterResources,
which had none: a NodePool marked for deletion is reaped even while
Cluster Service is still reporting the CR. That reap is now safe
precisely because dispatch no longer waits on it.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/backend/pkg/controllers/nodepool/deletion/node_pool_child_resources_cleanup_controller.go
+++ b/backend/pkg/controllers/nodepool/deletion/node_pool_child_resources_cleanup_controller.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/Azure/ARO-HCP/backend/pkg/utils/controllerutils"
 	"github.com/Azure/ARO-HCP/internal/api/coreapi"
+	"github.com/Azure/ARO-HCP/internal/api/kubeapplierapi"
 	"github.com/Azure/ARO-HCP/internal/database/cosmosstorage/corecosmosstorage"
 	"github.com/Azure/ARO-HCP/internal/database/cosmosstorage/cosmosstorageutils"
 	"github.com/Azure/ARO-HCP/internal/database/cosmosstorage/kubeappliercosmosstorage"
@@ -38,8 +39,10 @@ import (
 // the NodePool is marked for deletion and Cluster Service has confirmed the
 // delete on its side. Controller status documents (NodePoolControllerResourceType)
 // are left alone. Nodepool-scoped kube-applier *Desires are deleted here using
-// the parent cluster's ServiceProviderCluster placement. The orphan scraper
-// handles controller status after the NodePool document itself is removed.
+// the parent cluster's ServiceProviderCluster placement, except for ApplyDesires
+// that name an owning controller - those are that controller's to tear down; see
+// extraDeleteGateShouldDeleteApplyDesire. The orphan scraper handles controller
+// status after the NodePool document itself is removed.
 type nodePoolChildResourcesCleanupController struct {
 	nodePoolLister       corelisters.NodePoolLister
 	resourcesDBClient    corecosmosstorage.ResourcesDBClient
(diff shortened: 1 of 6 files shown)


backend/pkg/controllers/nodepool/deletion/node_pool_cluster_service_delete_dispatch_controller.go
```diff
@@ -160,16 +151,6 @@ func (c *nodePoolClusterServiceDeleteDispatchSyncer) SyncOnce(ctx context.Contex
 		return nil
 	}
 
-	// Wait until ClusterResourcesController has deleted all its tagged ApplyDesires for this NodePool.
-	hasTaggedDesires, err := c.hasNodePoolApplyDesires(ctx, key)
```
machi1990, line 164: Why is this removed completely?

nimrodshn: So that the dispatcher does not hold from dispatching on `hasNodePoolApplyDesire`. This hold the deletion of the MW which will cause a tug of war between kube-applier and the ACM work-agent instead of both agreeing on the delete. This is the fix.

machi1990: I believe this completely bypasses the intent of https://github.com/nimrodshn/ARO-HCP/blob/34fa833729c033a7b16cd0e6d0c93de1e1e7ceae/backend/pkg/controllers/clusterresources/cluster_resources_controller.go#L318 as it opens a gate for https://github.com/nimrodshn/ARO-HCP/blob/34fa833729c033a7b16cd0e6d0c93de1e1e7ceae/backend/pkg/controllers/nodepool/deletion/node_pool_child_resources_cleanup_controller.go#L245 to be exercised 

Let's continue the discussion on Slack about how we want to handle deletion and we can come back here

backend/pkg/controllers/nodepool/deletion/node_pool_child_resources_cleanup_controller.go
```diff
@@ -314,7 +320,66 @@ func (c *nodePoolChildResourcesCleanupController) ensureNodePoolScopedKubeApplie
 		return utils.TrackError(fmt.Errorf("error iterating nodepool-scoped kube-applier resources: %w", err))
 	}
 
+	if retained > 0 {
```
machi1990: Was this added for debugging?

I don't find it yourself as it just contained the counter of what hasn't been deleted rather than what's the actual resourceId that hasn't been delete. I'd remove the log.

Otherwise, if we care about such a log, I see  a debug log in the `if !shouldDelete { ` branch in line 308 that contains the resourceID of what has been skipped for deletion.

backend/pkg/controllers/nodepool/deletion/node_pool_child_resources_cleanup_controller.go
```diff
@@ -314,7 +320,66 @@ func (c *nodePoolChildResourcesCleanupController) ensureNodePoolScopedKubeApplie
...
+			clusterResourceID.SubscriptionID, clusterResourceID.ResourceGroupName, clusterResourceID.Name, nodePoolResourceID.Name)
+		if err != nil {
+			return false, utils.TrackError(fmt.Errorf("failed to create nodepool-scoped ApplyDesire CRUD: %w", err))
+		}
+
+		applyDesire, err := applyDesireCRUD.Get(ctx, strings.ToLower(applyDesireResourceID.Name))
+		if cosmosstorageutils.IsNotFoundError(err) {
+			// Raced with the owner purging it; nothing left to delete.
+			return false, nil
+		}
```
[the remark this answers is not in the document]
machi1990: No, the owning controller should take care of this. No need to change this. 
If the resource is no longer there, there is no need to issue another deletion

backend/pkg/controllers/nodepool/deletion/node_pool_child_resources_cleanup_controller.go
nimrodshn: Let me remove it

machi1990 (approved):
/lgtm

Get @miguelsorianod and/or @geoberle to review and approve this

stevekuznetsov:
/hold

must have e2e which fails without the patch but succeeds with it, running in our dev jobs

nimrodshn:
closing in favor of #6855

PR #6856, opened 2026-09-07
feat(fleet): replace mgmt-cluster Bicep with aks-cluster-create Go tool

geoberle:
## Summary

Move management-cluster provisioning from Bicep into a Go tool as the first step toward environment-derived node pool management. This stage uses static `config.yaml` `mgmt.aks.{systemAgentPool,userAgentPool,infraAgentPool}` configuration and shares the resource-modeling code with fleet controllers. ARM still handles resource dependencies and long-running operations, rather than the tool independently polling every pool.

Tracking: this PR serves as the implementation and review record for the static-configuration stage; no separate tracking ticket is linked. Follow-up PRs will add desired-state allocation (PR2) and the day-2 node pool reconciliation controller (PR3).

## Provisioning behavior

- **One incremental ARM deployment per cluster:** `dev-infrastructure/scripts/aks-cluster-create` builds the template from SDK cluster and agent-pool specifications. The bootstrap system pool is inline only for initial cluster creation; configured child pools are deployed declaratively. Unconfigured pools are left untouched, and Kubernetes versions are omitted on updates to avoid implicit upgrades.
- **Avoid no-op cluster reconciliation:** omit the cluster resource when its owned configuration already matches and provisioning has succeeded, because even an identical AKS PUT starts reconciliation. Include it for configuration drift or recovery from unsuccessful provisioning. Log changed cluster field paths, not configuration values.
- **Restart and recovery:** use the stable `aks-cluster-create` deployment name, wait for previous deployment and cluster operations, then observe live configuration before building a fresh template. A previous successful deployment does not replace the live drift check. Log the previous Azure deployment error before recovering Failed/Canceled work.
- **Resource diagnostics without continuous polling:** collect deployment operations in one paginated pass after completion or failure. Log resource identity, provisioning state and operation, timestamp, duration, status code, and nested provider errors. Collection is best-effort, bounded to 30 seconds even after cancellation, and never replaces the deployment result. Request and response payloads are excluded from operation logs.
- **Provisioning handover:** stamp new clusters with `aro-hcp-provisioning`. Deployment success is the handover barrier: refresh the cluster and its ETag, then conditionally remove the marker while applying configured tags and preserving external tags. Failed deployments leave the marker intact; finalization does not independently poll every pool.

## Shared packages and pipeline changes

- Share compute types, zone/capacity helpers, SKU metadata caching, agent-pool helpers, and SDK-based pool specification construction through `fleet/pkg/compute` and `fleet/pkg/azure/{skucache,agentpools,agentpoolspec}`. Adapt the capacity reporting controller to these packages and `armcontainerservice` v8, and align affected module dependencies.
- **Deployment pipeline change:** `dev-infrastructure/mgmt-pipeline.yaml` pre-builds the binary and runs `cluster-create` as a Shell step between `infra` and `cluster`.
- Move AKS prerequisites—managed identity, subnets, NSG, outbound IP, and KMS Key Vault—from `mgmt-cluster.bicep` to `mgmt-infra.bicep` so they exist before the tool runs. Keep `mgmt-cluster.bicep` for post-cluster wiring and extract ACR pull, workload identity, and maintenance configuration into `aks-cluster-post.bicep`.

## Test plan

Verified locally for the updated AKS tool:

- [x] `cd dev-infrastructure/scripts/aks-cluster-create && GOWORK=off go test -race ./...`
- [x] `cd dev-infrastructure/scripts/aks-cluster-create && GOWORK=off go build -o aks-cluster-create .`
- [x] Local SDK-transport smoke run: previous deployment error before recovery, multi-page operation collection, nested provider errors and timings, diagnostics after cancellation, exclusion of request/response payloads, and best-effort handling of unavailable diagnostics. No Azure deployments created; temporary smoke files removed.

Broader validation remains to be completed or confirmed for this revision:

- [ ] `go test ./fleet/pkg/azure/... ./fleet/pkg/compute/... ./fleet/pkg/controllers/capacityreporting/...`
- [ ] `make validate-config-pipelines`
- [ ] `make verify-materialize` after `cd config && make materialize`
- [ ] Deploy to a personal development environment and verify cluster and pool creation
- [ ] All required non-Tide CI checks pass (new checks are pending after the push)


config/config.yaml
```diff
@@ -28,7 +28,7 @@ defaults:
   region: "{{ .ctx.region }}"
   regionRG: "{{ .ctx.region }}-shared-resources"
   cloud: "{{ .ev2.cloudName }}"
-  azureRegionAvailabilityZoneCount: {{ .ev2.availabilityZoneCount }}
+  azureRegionAvailabilityZoneCount: {{ .ev2.availabilityZoneLiveCount }}
```
geoberle, line 31: FYI this is the real active and usable AZs in a region

config/config.yaml
```diff
@@ -1605,25 +1605,6 @@ clouds:
...
-                  zones: '1'
-            mgmt:
-              aks:
-                userAgentPool:
-                  zones: '1'
-                infraAgentPool:
-                  zones: '1'
-                  vmSize: 'Standard_D2s_v3'
-                systemAgentPool:
-                  zones: '1'
```
geoberle, line 1636: leftover cleanup

dev-infrastructure/scripts/aks-cluster-create/cluster_spec.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+	if len(o.metricAnnotationsAllowlist) == 0 && (liveMetrics == nil || liveMetrics.MetricAnnotationsAllowList == nil) {
+		metrics.MetricAnnotationsAllowList = nil
+	}
+	if metrics.MetricLabelsAllowlist == nil && metrics.MetricAnnotationsAllowList == nil {
+		desired.Properties.AzureMonitorProfile.Metrics.KubeStateMetrics = nil
+	}
+
+	live := *existing
+	properties := *existing.Properties
```
bennerv, line 197: i don't think there's a state where properties can be nil, might be good to check before derefrencing.

dev-infrastructure/scripts/aks-cluster-create/create.go
```diff
@@ -0,0 +1,467 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v8"
+
+	"github.com/Azure/ARO-HCP/fleet/pkg/azure/agentpoolspec"
+	"github.com/Azure/ARO-HCP/fleet/pkg/azure/skucache"
+	"github.com/Azure/ARO-HCP/fleet/pkg/compute"
+)
+
+// The Azure SDK exposes cluster and agent pool provisioning states as strings.
+const (
+	provisioningStateCreating  = "Creating"
```
bennerv: aren't these enums based on the api version of the aks cluster?

dev-infrastructure/scripts/aks-cluster-create/create.go
```diff
@@ -0,0 +1,467 @@
...
+	})
+	if err != nil && lastConflict != nil && wait.Interrupted(err) {
+		return fmt.Errorf("cluster reconciliation retries interrupted: %w", errors.Join(err, lastConflict))
+	}
+	return err
+}
+
+// reconcileCluster drives one attempt of the cluster lifecycle in three phases:
+//
+//  1. Ensure cluster — reates a new cluster with an inline bootstrap system pool,
```
bennerv: nit: 
```suggestion
//  1. Ensure cluster — creates a new cluster with an inline bootstrap system pool,

File: dev-infrastructure/scripts/aks-cluster-create/create.go [hunk 6]
```diff
@@ -0,0 +1,467 @@
...
+		return err
+	}
+
+	// Phase 3: finalize tags.
+	return o.reconcileClusterTags(ctx, cluster)
+}
+
+// ensurePools ensures all configured pools exist, match their desired
+// configuration, and are healthy. Each pool's own ETag is used for updates
+// and recovery of failed pools.
```
bennerv: this flags for any pool that doesn't match.  do we want to only atomically update a single pool rather than them all if one doesn't match?  

I assume that the situation where one pool doesn't match desired state but the others do is rare.

dev-infrastructure/scripts/aks-cluster-create/main.go
```diff
@@ -0,0 +1,110 @@
...
+func main() {
+	verbosity := 0
+	if v := os.Getenv("LOG_VERBOSITY"); len(v) > 0 {
+		if n, err := strconv.Atoi(v); err == nil {
+			verbosity = n
+		}
+	}
+	logger := funcr.NewJSON(func(obj string) {
+		fmt.Fprintln(os.Stderr, obj)
+	}, funcr.Options{Verbosity: verbosity})
```
bennerv: I think we use a different logging pattern elsewhre no?

dev-infrastructure/scripts/aks-cluster-create/main.go
```diff
@@ -0,0 +1,110 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
```
bennerv, line 99: how'd you get to 30 minutes here?  Is that what our bicep template uses today already?

dev-infrastructure/scripts/aks-cluster-create/options.go
```diff
@@ -0,0 +1,406 @@
...
+// Complete builds the Azure credential and clients used to create the cluster
+// and its pools.
+func (o *validatedOptions) Complete() (*completedOptions, error) {
+	cred, err := azidentity.NewDefaultAzureCredential(&azidentity.DefaultAzureCredentialOptions{RequireAzureTokenCredentials: true})
+	if err != nil {
+		return nil, fmt.Errorf("azidentity: %w", err)
+	}
+
+	policyClientOptions := &policy.ClientOptions{}
+	armClientOptions := &azcorearm.ClientOptions{ClientOptions: *policyClientOptions}
```
bennerv: if these are empty, why can't we just use nil for the options or an empty options?

bennerv: It might make sense to set the user agent header here using the internal package + a git rev that is plumbed into makefile so that we can tell who is making the AKS updates here if you want to set options.

dev-infrastructure/scripts/aks-cluster-create/cluster_spec.go
geoberle: the func rejects clusters with nil properties right at the beginning

dev-infrastructure/scripts/aks-cluster-create/create.go
bennerv: Nope they're string pointers.

dev-infrastructure/scripts/aks-cluster-create/create.go
```diff
@@ -0,0 +1,467 @@
...
+				logger.Info("waiting for pool provisioning", "pool", pool.Name, "provisioningState", state)
+				return false, nil
+			case provisioningStateFailed:
+				if existing.Properties.ETag == nil || len(*existing.Properties.ETag) == 0 {
+					return false, fmt.Errorf("pool %q has no ETag for update", pool.Name)
+				}
+				options.IfMatch = existing.Properties.ETag
+				logger.Info("recovering failed pool", "pool", pool.Name)
+			default:
+				return false, fmt.Errorf("pool %q has unsupported provisioning state %q", pool.Name, state)
```
bennerv: I wouldn't state that it's an unsupported provisioning state.  There's also a canceled provisioning state.  I don't know if we're accounting for all of them from AKS side.  

Do we care to "recover" from the default, or implement a canceled one?

bennerv: Resource providers define their own non-terminal states.  For example, here are ours: https://github.com/Azure/ARO-HCP/blob/076bb8fbe62e65e30d9039e33c1d3926b52c6e14/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpopenshiftclusters/preview/2026-09-01-preview/openapi.json#L2105-L2112

The [terminal ones though ](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/async-api-reference.md#azure-asyncoperation-resource-format)are Succeeded, Failed, Canceled

dev-infrastructure/scripts/aks-cluster-create/create.go
geoberle: only the pools that don't match are updated

dev-infrastructure/scripts/aks-cluster-create/main.go
geoberle: shell step don't have a controllable timeout yet
i added a comment and will work on adding controllable timeouts

dev-infrastructure/scripts/aks-cluster-create/options.go
geoberle: i added only the agent header for now. git rev is complex because of the way we build such scripts

dev-infrastructure/scripts/aks-cluster-create/main.go
geoberle: adapted to slog

dev-infrastructure/scripts/aks-cluster-create/create.go
```diff
@@ -0,0 +1,486 @@
...
+	for _, pool := range pools {
+		g.Go(func() error {
+			defer utilruntime.HandleCrash()
+			changed, err := ensurePool(gctx, o.poolsClient, o.resourceGroup, o.clusterName, pool, networkConfig, logger)
+			if changed {
+				poolsChanged.Store(true)
+			}
+			return err
+		})
+	}
```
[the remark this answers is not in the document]
bennerv: I think this was fixed in go1.22.

dev-infrastructure/scripts/aks-cluster-create/create.go
```diff
@@ -0,0 +1,486 @@
...
+	refreshCluster := false
+	err := wait.PollUntilContextCancel(ctx, 10*time.Second, true, func(ctx context.Context) (bool, error) {
+		existing, err := client.Get(ctx, resourceGroup, clusterName, pool.Name, nil)
+		options := &armcontainerservice.AgentPoolsClientBeginCreateOrUpdateOptions{}
+		if err == nil {
+			state := ""
+			if existing.Properties != nil {
+				state = ptr.Deref(existing.Properties.ProvisioningState, "")
+			}
+			switch state {
```
bennerv:
```
switch state {

case provisioningStateSucceeded: 
  // validate pool is up-to-date

case provisioningStateFailed, provisioningStateCanceled: 
  // attempt to recover nodepool

default: 
  // non-terminal provisioning state, track elapsed time and output provisioning state

File: dev-infrastructure/scripts/aks-cluster-create/create.go (hunk 11 above)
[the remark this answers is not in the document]
Comment (geoberle) [reply dev-infrastructure/scripts/aks-cluster-create/create.go]: we have modern go. not an issue

File: dev-infrastructure/scripts/aks-cluster-create/reconcile.go [hunk 13]
```diff
@@ -0,0 +1,66 @@
...
+			logger.Info("cluster configuration is up to date")
+			return existing, nil
+		}
+		if existing.ETag == nil || len(*existing.ETag) == 0 {
+			return nil, fmt.Errorf("cluster ETag is missing")
+		}
+		options.IfMatch = existing.ETag
+		logger.Info("updating cluster configuration", "fields", changed)
+	}
+	poller, err := o.clustersClient.BeginCreateOrUpdate(ctx, o.resourceGroup, o.clusterName, spec, options)
```
raelga: Shouldn't we wait here, like we do for agent pools? If I rerun the pipeline after an interruption and AKS is still creating or updating the cluster, we send another PUT even if the config already matches. That can hit an operation-in-progress conflict and use up all six retries in about 30 seconds, while AKS can take several minutes to finish. I'd wait for that operation, fetch the cluster again, and then check if anything still needs updating.

fleet/pkg/azure/skucache/skucache.go
```diff
@@ -140,60 +189,179 @@ func (c *SKUCache) fetchSKUResources(ctx context.Context, subscriptionID string)
...
+
 	if vcpus, ok := lookupInt64(sku, capabilityNameVCPUs); ok {
-		resources[corev1.ResourceCPU] = *resource.NewQuantity(vcpus, resource.DecimalSI)
+		meta.VCPUs = vcpus
+		if available, ok := lookupInt64(sku, capabilityNameVCPUsAvailable); ok && available < vcpus {
+			meta.ConstrainedVCPUs = true
+		}
+	}
+	if memoryGB, ok := lookupFloat64(sku, capabilityNameMemoryGB); ok {
+		meta.MemoryGB = int64(memoryGB)
```
raelga: We're losing the decimals here, so `3.5 GiB` becomes `3 GiB`. We use this value for the scale ceiling when there are no ready-node samples, and the missing memory gets multiplied by the pool's max count. For a pool with a max of 10 nodes, we'd report 30 GiB instead of 35 GiB. Can we keep it in bytes or use a `resource.Quantity` so we don't lose that precision?

raelga (approved):
Thanks for the follow up changes @geoberle !

/lgtm
/approve

geoberle:
/hold
need to reevaluate because of subscription throttling

dev-infrastructure/scripts/aks-cluster-create/cluster_spec.go
```diff
@@ -0,0 +1,301 @@
...
+
+	desired := armcontainerservice.ManagedCluster{
+		SKU: &armcontainerservice.ManagedClusterSKU{
+			Name: ptr.To(armcontainerservice.ManagedClusterSKUNameBase),
+			Tier: ptr.To(armcontainerservice.ManagedClusterSKUTierStandard),
+		},
+		Identity: &armcontainerservice.ManagedClusterIdentity{
+			Type: ptr.To(armcontainerservice.ResourceIdentityTypeUserAssigned),
+			UserAssignedIdentities: map[string]*armcontainerservice.ManagedServiceIdentityUserAssignedIdentitiesValue{
+				o.managedIdentityID: {},
```
bennerv, line 61: we don't assign an identity to the cluster?  I thought we did for ACR pull?

bennerv: Discussed, that IS the identity we asssign.  Server is responsible for populating clientID prinicpalID.

raelga (approved):
/lgtm
/approve

PR #6857, opened 2026-09-07
refactor(fleet): add Controller interface, WriteController

geoberle:
### What

Port controller mechanics from the backend pattern: a common Controller interface (Run, QueueForInformers, EnqueueAfter), WriteController for persisting controller Degraded conditions to Cosmos, and a ManagementClusterWatchingController base that automatically writes controller documents after each sync. All existing stamp-watching controllers now return the Controller interface.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/fleet/pkg/controllers/base/management_cluster_watching_controller.go
+++ b/fleet/pkg/controllers/base/management_cluster_watching_controller.go
@@ -1,0 +1,143 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package base
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/go-logr/logr"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/cache"
+
+	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+
+	"github.com/Azure/ARO-HCP/internal/api/coreapi"
+	"github.com/Azure/ARO-HCP/internal/api/fleetapi"
+	"github.com/Azure/ARO-HCP/internal/api/metadataapi"
+	"github.com/Azure/ARO-HCP/internal/controllerutils"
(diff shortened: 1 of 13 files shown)


fleet/pkg/controllers/base/utils.go
```diff
@@ -0,0 +1,163 @@
...
+
+	existingController, err = controllerCRUD.Get(ctx, controllerName)
+	switch {
+	case err == nil:
+		return existingController, nil
+	case cosmosstorageutils.IsNotFoundError(err):
+		if secondAttempt {
+			return nil, utils.TrackError(fmt.Errorf("second NotFound, Conflict, NotFound error: %w", err))
+		}
+		timer := time.NewTimer((cosmosstorageutils.SoftDeleteTTLSeconds + 1) * time.Second)
```
raelga, line 152: this blocks the worker goroutine for the full soft-delete TTL on a conflict-then-notfound race. it's rare and only stalls this one key, but might be cleaner to just return a retryable error here and let the workqueue backoff handle the wait instead of sleeping inside the sync call

fleet/pkg/controllers/base/utils.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+// ControllerMutationFunc mutates a controller document in place. It should only
+// perform short calls, not long lookups. It must not fail.
+type ControllerMutationFunc func(controller *coreapi.Controller)
+
+// ReportSyncError returns a mutation that sets the Degraded condition based on
+// whether a sync error occurred.
+func ReportSyncError(syncErr error) ControllerMutationFunc {
```
raelga, line 38: since WriteController only writes when the message actually changes, if any syncer ever wraps an error with something non-deterministic (timestamp, request id) we'd get a cosmos write on every single reconcile even though nothing meaningful changed. worth double checking the errors coming out of the ported syncers are stable strings

fleet/pkg/controllers/base/management_cluster_watching_controller.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (k ManagementClusterKey) AddLoggerValues(logger logr.Logger) logr.Logger {
+	return logger.WithValues(
+		utils.LogValues{}.
+			AddLogValuesForResourceID(k.GetResourceID())...)
+}
+
+func (k ManagementClusterKey) InitialController(controllerName string) *coreapi.Controller {
+	resourceID := metadataapi.Must(azcorearm.ParseResourceID(
```
raelga, line 56: building the child resource id by string concat here, might be worth using an arm resource id builder instead so it doesn't break if controllerName ever needs escaping. not blocking, just a nit

fleet/pkg/controllers/base/utils.go
geoberle: existing code.

fleet/pkg/controllers/base/management_cluster_watching_controller.go
geoberle: also existing code

fleet/pkg/controllers/base/utils.go
geoberle: that's on purpose. this is not like conditions but an error tracker. exactly the same in backend

raelga (approved):

PR #6865, opened 2026-09-08
bring back deny assignment creation

deads2k:
unreverts https://github.com/Azure/ARO-HCP/pull/6817, built on top of https://github.com/Azure/ARO-HCP/pull/6795 so that the managed resourcegroup will exist.


machi1990 (changes requested):
We need to also update https://github.com/deads2k/ARO-HCP/blob/80581afc64e43646738ab9d000b41a6a44382820/backend/pkg/controllers/cluster/denyassignments/deny_assignment_controller.go#L112 so that we gate NeedsWork on 
`Status.AzureResources.ManagedResourceGroup.AzureResource`

We can do it here since;
- we long had an observer that fills the MRG resource ID onto status
- and we'll soon have the creation owned by the backend; which we've in this code base.

PR #6869, opened 2026-09-09
ARO-29051- KAS API SLO dashboard references non-existent metrics

LiniSusan:
https://issues.redhat.com/browse/ARO-29051
### What
Add compatibility Prometheus recording rules so the existing KAS API SLO dashboard keeps working without changing its queries.
The dashboard (`observability/grafana-dashboards/kas-monitor/kas_monitor_api_slo.json`) still queries the old blackbox-probe names:
- `probe_success` (1-day panel)
- `probe_availability:ratio_avg_7d`
- `probe_availability:ratio_avg_30d`
Those series no longer exist. This PR records them from the live KSM signal `hostedClusterAPI_kubeapiserver_available{status="True"}` (HostedCluster `KubeAPIServerAvailable` condition).
Files:
- `observability/alerts/HCPkasProbeAvailabilityRecord-prometheusRule.yaml`
- `observability/alerts/HCPkasProbeAvailabilityRecord-prometheusRule_test.yaml`
- registered in `observability/recording-rules-hcps.yaml`
- generated `dev-infrastructure/modules/metrics/rules/generatedHCPRecordingRules.bicep` (Azure resource `hcp-kas-probe-availability-recording-rules`)
Existing KSM recording rules (`hostedClusterAPI_kubeapiserver_available:ratio_avg_*`) are unchanged. Dashboard JSON is unchanged.

### Why
The KAS API SLO dashboard panels were empty because they still queried retired probe metrics. Availability is already measured from the HostedCluster kube-apiserver Available condition. Recreating the old series from that condition restores the dashboard without a query migration and without rewriting the KSM rules other alerts already depend on.
Rules are gated on a currently reported KSM series so deleted clusters do not leave stale probe values.

### DIFF
--- a/observability/alerts/HCPkasProbeAvailabilityRecord-prometheusRule.yaml
+++ b/observability/alerts/HCPkasProbeAvailabilityRecord-prometheusRule.yaml
@@ -1,0 +1,33 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: hcp-kas-probe-availability-recording-rules
+  namespace: prometheus
+spec:
+  groups:
+  - name: hcp-kas-probe-availability-recording-rules
+    interval: 1m
+    rules:
+    # Compatibility series for the kas-monitor API SLO dashboard.
+    # The dashboard still queries the historical blackbox-probe names;
+    # these rules record them from the HostedCluster kube-apiserver
+    # Available condition without changing existing KSM recording rules.
+    #
+    # All rules are gated on the raw KSM metric currently being reported.
+    # This prevents stale data from deleted clusters.
+
+    - record: probe_success
+      expr: |
+        avg by (name, namespace, _id, resource_id, cluster) (hostedClusterAPI_kubeapiserver_available{status="True"})
+        and on (name, namespace, _id, resource_id, cluster)
+        count by (name, namespace, _id, resource_id, cluster) (hostedClusterAPI_kubeapiserver_available)
+    - record: probe_availability:ratio_avg_7d
+      expr: |
+        avg by (name, namespace, _id, resource_id, cluster) (avg_over_time(hostedClusterAPI_kubeapiserver_available{status="True"}[7d]))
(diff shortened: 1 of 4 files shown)


observability/alerts/HCPkasProbeAvailabilityRecord-prometheusRule.yaml
```diff
@@ -0,0 +1,33 @@
...
+    rules:
+    # Compatibility series for the kas-monitor API SLO dashboard.
+    # The dashboard still queries the historical blackbox-probe names;
+    # these rules record them from the HostedCluster kube-apiserver
+    # Available condition without changing existing KSM recording rules.
+    #
+    # All rules are gated on the raw KSM metric currently being reported.
+    # This prevents stale data from deleted clusters.
+
+    - record: probe_success
```
swiencki, line 19: Recording this as the bare `probe_success` name (the standard blackbox-exporter series) risks colliding with real probe metrics on the shared workspace, so please use a scoped record name such as `kubeapiserver:probe_success` to keep the KAS SLO backfill isolated.

observability/alerts/HCPkasProbeAvailabilityRecord-prometheusRule.yaml
```diff
@@ -0,0 +1,36 @@
...
+    # kube-apiserver Available condition without changing existing KSM
+    # recording rules.
+    #
+    # probe_success is recorded as kubeapiserver:probe_success so it does
+    # not collide with the standard blackbox-exporter series on the shared
+    # Azure Monitor workspace.
+    #
+    # All rules are gated on the raw KSM metric currently being reported.
+    # This prevents stale data from deleted clusters.
+    - record: kubeapiserver:probe_success
```
swiencki: Recording this as kubeapiserver:probe_success (namespaced) instead of a bare probe_success cleanly resolves the collision risk with the blackbox-exporter series on the shared workspace, and the presence gate on the raw metric correctly drops deleted clusters.

PR #6873, opened 2026-09-09
poc db design identity replacement storing all identity metadata sources in same identity mapping with oidc federation example

miguelsorianod:


backend/cmd/root.go
```diff
@@ -405,7 +406,7 @@ func (f *BackendRootCmdFlags) ToBackendOptions(ctx context.Context, cmd *cobra.C
 		)
 
 		// In ARO-HCP environments where we don't have a real FPA, we use the HardcodedIdentityFPAMIDataplaneClientBuilder to create the FPA MI dataplane client builder
-		fpaMIDataplaneClientBuilder, err = newHardcodedIdentityFPAMIDataplaneClientBuilder(
+		fpaMIDataplaneClientBuilder, hardcodedIdentity, err = newHardcodedIdentityFPAMIDataplaneClientBuilder(
```
miguelsorianod, line 409: we now return the hardcodedIdentity too because some controllers persist it as the identity metadata information. An alternative would be to still keep using the mock FPAMIdataplaneClient that would return the mock info but then it would not be accurate and in recent discussions we said we don't want to "hide" the reality

backend/pkg/app/backend.go
```diff
@@ -1070,13 +1082,42 @@ func (b *Backend) runBackendControllersUnderLeaderElection(ctx context.Context,
 		b.options.FPAMIDataplaneClientBuilder,
 	)
 
+	var fetchManagedIdentitiesInfoDataplaneBuilder azureclient.FPAMIDataplaneClientBuilder
+	if b.options.HardcodedIdentity == nil {
+		fetchManagedIdentitiesInfoDataplaneBuilder = b.options.FPAMIDataplaneClientBuilder
+	}
+	fetchManagedIdentitiesInfoController := clusteridentity.NewFetchManagedIdentitiesInfoController(
```
miguelsorianod, line 1089: For this specific controller we either set hardcodedidentity or fpamidataplaneclientbuilder, but not both. We don't use the mock client implementation of the midataplane service, instead we rely on hardcodedidentity. Same reasoning as https://github.com/Azure/ARO-HCP/pull/6873#discussion_r3968390997

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -300,6 +303,150 @@ type ServiceProviderClusterStatus struct {
...
+	// FetchManagedIdentitiesInfo should re-query identity metadata sources.
+	// Nil means recheck immediately.
+	// This allows the controller to avoid repeatedly hitting Azure APIs to
+	// recheck that the desired state is true.
+	// Controllers should set this field with substantial jitter: without another
+	// concern, jitter of 50% is considered normal so that any storms are quickly
+	// dissipated. Additionally, long recheck times are recommended for resources
+	// outside of their active phases. Order of at least six hours is, with durations up to 24 hours considered normal.
+	// Written by: FetchManagedIdentitiesInfo
+	ManagedIdentitiesEarliestRecheckTime *metav1.Time `json:"managedIdentitiesEarliestRecheckTime,omitempty"`
```
miguelsorianod, line 326: This one could be moved to EarliestRecheckTimesByController

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -300,6 +303,150 @@ type ServiceProviderClusterStatus struct {
...
+type ManagedIdentityDataplaneOIDCFederationStatus struct {
+	// EarliestRecheckTime is the earliest time at which the controller should
+	// re-query Azure for the managed identity's OIDC Federation status. Nil means recheck immediately.
+	// This allows the controller to avoid repeatedly hitting an Azure API to
+	// recheck that the desired state is true.
+	// Controllers should set this field with substantial jitter: without another
+	// concern, jitter of 50% is considered normal so that any storms are quickly
+	// dissipated. Additionally, long recheck times are recommended for resources outside of their active phases. Order of at least six hours is, with durations up to 24 hours considered normal.
+	// Written by: DataPlaneOIDCFederation
+	EarliestRecheckTime *metav1.Time `json:"earliestRecheckTime,omitempty"`
```
miguelsorianod, line 424: If we want to keep the earliestrecheck per oidc federated identity we can't move this to EarliestRecheckTimesByController, because it's not at the controller level but at the identity level within the controller

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -598,3 +745,81 @@ const (
...
+	// Written by: FetchManagedIdentitiesInfo
+	TenantID *string `json:"tenantId,omitempty"`
+	// RetrievalError, when non-nil, is the error (truncated to the first 1024
+	// characters) from the most recent attempt to retrieve this identity's
+	// metadata from this source. When set, ClientID, PrincipalID, and TenantID
+	// are nil because the last retrieval attempt failed, and any previously
+	// resolved values are no longer trustworthy. It is nil when the last
+	// retrieval succeeded.
+	// Written by: FetchManagedIdentitiesInfo
+	RetrievalError *string `json:"retrievalError,omitempty"`
```
miguelsorianod, line 808: RetrievalError is not being set for all identity metadata sources but allows us to have the same data type for all of them. An alternative is to have dffernt data types for each source so we don't define the RetrievalError for the ones we don't use it (yet?)

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -598,3 +745,81 @@ const (
 	// the cluster-autoscaler ControlPlaneComponent on the management cluster control plane namespace.
 	ReadonlyHypershiftControlPlaneComponentClusterAutoscaler MaestroBundleInternalName = "readonlyHypershiftControlPlaneComponentClusterAutoscaler"
 )
+
+// ManagedIdentityMetadata holds the metadata retrieved for a single managed
+// identity from each source that applies to it. Sources that do not apply
+// are left nil. ARM Get failures still set MetadataFromARMUserAssignedIdentitiesAPI
+// with RetrievalError.
+type ManagedIdentityMetadata struct {
```
miguelsorianod, line 753: As discussed, we define the three potential sources of metadata information for an azure resource id in aro-hcp. This has the implication that now the consumers need to be aware of the environment they run so they know from what source(s) is the information expected

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -598,3 +745,81 @@ const (
...
+	MetadataFromManagedIdentitiesDataplaneService *IdentityMetadataValue `json:"metadataFromManagedIdentitiesDataplaneService,omitempty"`
+	// MetadataFromHardcodedIdentity is the metadata for the identity retrieved from the Hardcoded Identity as the source.
+	// Set only in environments where the real Managed Identities Data Plane
+	// service is not available, and only for identities registered with the
+	// dataplane (control-plane operator identities and the service managed identity).
+	// When the source applies, the pointer is set even if this pass did not
+	// fill ClientID/PrincipalID/TenantID: an empty IdentityMetadataValue means
+	// not resolved yet or could not be resolved in this pass.
+	// Written by: FetchManagedIdentitiesInfo
+	MetadataFromHardcodedIdentity *IdentityMetadataValue `json:"metadataFromHardcodedIdentity,omitempty"`
```
miguelsorianod, line 784: In the environments where the MIDataplane service doesn't exist this would remain nil

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -598,3 +745,81 @@ const (
...
+	// MetadataFromManagedIdentitiesDataplaneService is the metadata for the identity retrieved from the **real**
+	// Managed Identities Dataplane Service as the source. Nil when the environment
+	// uses the hardcoded identity instead, or when this identity is not registered
+	// with the dataplane (it is not a control-plane operator identity or the
+	// service managed identity). When the source applies, the pointer is set
+	// even if this pass did not fill ClientID/PrincipalID/TenantID: an empty
+	// IdentityMetadataValue means not resolved yet or could not be resolved
+	// in this pass.
+	// Written by: FetchManagedIdentitiesInfo
+	MetadataFromManagedIdentitiesDataplaneService *IdentityMetadataValue `json:"metadataFromManagedIdentitiesDataplaneService,omitempty"`
```
miguelsorianod, line 775: In the environments where no real MIDataplane service is available this would remain nil.

In the environments where a real MIDataplane service exists, this would not be nil, but the contents within might be unset or become unset in different scenarios:
* Azure fetch failure
* The resourceid is not set in any control plane operator nor the service managed identity (this changes over time)
* The identity doesn't actually exist in azure or stops existing in azure. The MIDataplane service returns the information but with nil clientid,principalid,...

Identities that are set only for data plane operators will have this attribute set to nil. This is because it's not possible to retrieve data plane identities information using the MIDataplane service (only the ones associaated to ocntrol plane operators and/or smi are configured with the MIDataplane integration)

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -598,3 +745,81 @@ const (
...
+	ResourceID *azcorearm.ResourceID `json:"resourceId,omitempty"`
+
+	// MetadataFromARMUserAssignedIdentitiesAPI is the metadata for the identity retrieved from
+	// the ARM User Assigned Identities API (https://learn.microsoft.com/en-us/rest/api/managedidentity/user-assigned-identities)
+	// as the source. Nil when ARM is not queried for this identity (the cluster's
+	// service managed identity cannot Get itself). When ARM is queried, this
+	// pointer is set even if the Get failed: ClientID, PrincipalID, and TenantID
+	// are then nil and RetrievalError records why.
+	// Written by: FetchManagedIdentitiesInfo
+	MetadataFromARMUserAssignedIdentitiesAPI *IdentityMetadataValue `json:"metadataFromARMUserAssignedIdentitiesAPI,omitempty"`
```
miguelsorianod, line 765: If an identity is only used as SMI then this will never be set. This is because the SMI can't be used to retrieve its own information using ARM UserAssignedIdentities API

miguelsorianod: If an identity is set as a control plane operator one or as a dataplane operator one this would be set. For control plane opreator ones this will have duplicated information with MetadataFromManagedIdentitiesDataplaneService as it's possible to retrieve from both.
It's unclear if we want to do this.

backend/pkg/controllers/cluster/dataplaneworkloads/oidc_federation_intent_controller.go
```diff
@@ -0,0 +1,363 @@
...
+		// The ServiceProviderCluster has not been created yet. The dedicated
+		// CreateServiceProviderCluster controller creates it; we pick it up on a
+		// later requeue once it exists.
+		return nil
+	}
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get ServiceProviderCluster: %w", err))
+	}
+
+	dataPlaneOperators := existingCluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.DataPlaneOperators
```
miguelsorianod, line 127: This controller now needs to become aware of both the identity mappings map and the last desired dataplane operators information. This is so it only configures for intent of oidc federation the dataplane ones as well as knowing their information and the information they need to gather from.

In other controllers that deal with control plane operaotr identities or the service managed identity it gets complicated further. Aside from that information, they also need to become aware on whether the RP is running in an environment that has the MIDataplane service available or not, because they need to be able to know whether it should pull the metadata from the hardcodedidentity source or from the real midataplane source in the identity mappings map

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -300,6 +303,150 @@ type ServiceProviderClusterStatus struct {
...
+	// (control-plane operators and the ServiceManagedIdentity) are ignored, even
+	// when the same UAMI is still used as CP or SMI.
+	// DataPlaneOIDCFederationIntent adds those identities as PendingConfigure, and
+	// marks identities that have left that data-plane set as PendingDeconfigure
+	// (stamping DeconfigureTimestamp when that was requested so Azure FIC deletes
+	// wait 24 hours on a live cluster). DataPlaneOIDCFederation then creates or
+	// deletes federated identity credentials in Azure (one per data-plane operator
+	// service account) and advances the phase to Configured or Deconfigured.
+	// Written by: DataPlaneOIDCFederationIntent, DataPlaneOIDCFederation
+	ManagedIdentitiesWithDataPlaneWorkloadsOIDCFederation map[ManagedIdentityDataplaneOIDCFederationKey]*ManagedIdentityDataplaneOIDCFederationStatus `json:"managedIdentitiesWithDataPlaneWorkloadsOIDCFederation,omitempty"`
```
miguelsorianod, line 346: In this map as the key we have ManagedIdentityDataplaneOIDCFederationStatus which contains the corresponding identity metadata source instead of the three of them. That means however that in Cosmos at that level you can't see where the source that the information was retrievedfrom.You just see that the resourceid resolved to a specific metadata values.

PR #6875, opened 2026-09-09
Fix: Cilium CNI conflist `cniVersion` mismatch with Multus

miquelsi:
## Summary

All ARO-HCP e2e tests that install Cilium via Helm (`cluster_create_cni_cilium.go`,
`cluster_create_feature_aggregation.go`, `cluster_create_complex_cilium_kv.go`) were
failing deterministically against recent OCP builds: worker nodes never became `Ready`
because Multus (the OCP CNI meta-plugin) crash-looped on every node.

The fix supplies Cilium with a custom CNI configuration list (conflist) via the
`cni.configMap` Helm value, pinning its `cniVersion` field to `"1.0.0"` instead of
letting Cilium generate its own (incompatible) default.

## What changed

- **`test/util/framework/cilium_helper.go`**
  - Added `EnsureCiliumCNIConfigMap(ctx, adminRESTConfig, namespace, name, conflistJSON)` —
    creates (or updates) a Kubernetes `ConfigMap` holding a full CNI conflist for Cilium
    to read and write to disk verbatim, bypassing Cilium's own built-in template
    rendering.
  - Added two conflist constants matching Cilium's own built-in templates, but with a
    fixed, Multus-compatible `cniVersion`:
    - `CiliumConflistNone` — for Cilium's default (unchained) mode.
    - `CiliumConflistPortmap` — for Cilium's `portmap` chaining mode.
  - Added `MultusCompatibleCNIVersion = "1.0.0"` documenting the chosen version and why.

- **`test/e2e/cluster_create_cni_cilium.go`**
  - Before installing the Cilium Helm chart, calls `EnsureCiliumCNIConfigMap` with
    `CiliumConflistPortmap` (this test uses `cni.chainingMode: "portmap"`).
  - Adds `"configMap": ciliumCNIConfigMapName` to the Helm values' `cni` block.

- **`test/e2e/cluster_create_feature_aggregation.go`**
  - Same pattern, using `CiliumConflistNone` (this test does not set `chainingMode`, so
    Cilium runs in its default "none" mode).

- **`test/e2e/cluster_create_complex_cilium_kv.go`**
  - Same pattern as `cluster_create_feature_aggregation.go`, using `CiliumConflistNone`.

### The failure

On affected clusters, worker nodes stayed `NotReady` indefinitely with kubelet
reporting:

```
KubeletNotReady: container runtime network not ready: NetworkReady=false
reason:NetworkPluginNotReady message:Network plugin returns error: no CNI
configuration file in /etc/kubernetes/cni/net.d/. Has your network provider
started?
```

The `openshift-multus` DaemonSet pods were in `CrashLoopBackOff`, and their logs
showed the root cause:

```
[error] failed to create the configuration manager for the primary CNI plugin:
failed to load the primary CNI configuration as a multus delegate with error
'delegate cni version is 0.3.1 while top level cni version is 1.1.0'
```

### Root cause

Two independent pieces changed relative to one another:

1. **Cilium (pinned at v1.19.x in these tests)** hard-codes `"cniVersion": "0.3.1"` in
   every one of its built-in conflist templates
   (`daemon/cmd/cni/config.go` in `cilium/cilium`), for every chaining mode ("none",
   "flannel", "portmap"). This is **not configurable via any Helm value** — there is no
   `cni.cniVersion` key in the chart. (Cilium only changed this default itself in
   v1.20.0, via commit `3cd4de85a0`, "cni: Bump default version to 1.0.0".)

2. **OCP's `cluster-network-operator`** ships the Multus manifest
   (`bindata/network/multus/multus.yaml`) with its own top-level `cniVersion`. This was
   bumped from `"0.3.1"` to `"1.1.0"` by commit `a5d3f195`
   ("[NVIDIA-616](https://redhat.atlassian.net/browse/NVIDIA-616): Bump Multus CNI API version to 1.1.0"). That commit is present on
   `master`, `release-4.23`, `release-5.0`, `release-5.1`, and `release-5.2` — i.e. it
   affects **every currently active OCP release line**, not something specific to OCP
   5.0. It is absent from `release-4.20`/`4.21`/`4.22`.

Multus's own delegate-version compatibility check
(`multus-cni`'s `pkg/server/config/generator.go`, `CheckVersionCompatibility`) is
one-directional:

```go
if multusCNIVersion.GTE(v040) {        // only runs if Multus's own version >= 0.4.0
    if delegateVersion.LT(v040) {      // ...and only fails if delegate is < 0.4.0
        return fmt.Errorf(versionFmt, delegateVersion, mc.CNIVersion)
    }
}
```

- On **older OCP builds** (Multus top-level `cniVersion` still `0.3.1`), this check is
  skipped entirely (`0.3.1 >= 0.4.0` is false) — a `0.3.1` Cilium delegate is silently
  accepted. This is why the bug was never observed there.
- On **OCP builds with the CNO bump** (Multus top-level `cniVersion` now `1.1.0`), the
  check activates, and Cilium's `0.3.1` delegate conflist now fails it — every time,
  deterministically.

### Why the fix works, and why it's safe on older OCP too

Since Cilium doesn't expose a way to change its own generated `cniVersion`, the fix uses
Cilium's documented "bring your own conflist" mechanism: setting the `cni.configMap`
Helm value points Cilium at a `ConfigMap` whose content it writes to
`/etc/cni/net.d/05-cilium.conflist` (via multus's confPath) unmodified, instead of
rendering its own hard-coded template.

The custom conflists used here are functionally identical to Cilium's own built-in
"none" and "portmap" templates, with only the `cniVersion` field changed from `0.3.1` to
`1.0.0`. `1.0.0` was chosen because:

- It satisfies Multus's `>= 0.4.0` check, and
- `cilium-cni` itself explicitly advertises support for CNI spec versions up to `1.1.0`
  (`plugins/cilium-cni/cmd/cmd.go`), so `1.0.0` is safely within what the binary
  understands — no functional regression.

Because Multus's check is one-directional (only fails when its own version is new *and*
the delegate's is old), always declaring a modern `1.0.0` delegate version is safe
regardless of which OCP build runs the test:

- **Pre-bump OCP (e.g. 4.21)**: Multus's own version is still `0.3.1`, so the check is
  skipped entirely — a `1.0.0` delegate is accepted the same as a `0.3.1` one would be.
- **Post-bump OCP (4.23+, 5.x)**: Multus's own version is `1.1.0`, so the check runs and
  `1.0.0 >= 0.4.0` passes.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/test/e2e/cluster_create_cni_cilium.go
+++ b/test/e2e/cluster_create_cni_cilium.go
@@ -104,6 +104,10 @@ var _ = Describe("Customer", func() {
 			kubeconfigContent, err := framework.GenerateKubeconfig(adminRESTConfig)
 			Expect(err).NotTo(HaveOccurred(), "failed to generate kubeconfig for cluster %q", customerClusterName)
 
+			By("providing a Multus-compatible custom CNI conflist for Cilium")
+			err = framework.EnsureCiliumCNIConfigMap(ctx, adminRESTConfig, ciliumNamespace, framework.CiliumCNIConfigMapName, framework.CiliumConflistPortmap)
+			Expect(err).NotTo(HaveOccurred(), "failed to create Cilium CNI conflist ConfigMap")
+
 			By("installing Cilium via Helm")
 			ciliumValues := map[string]any{
 				"debug": map[string]any{
@@ -138,6 +142,7 @@ var _ = Describe("Customer", func() {
 					"binPath":      "/var/lib/cni/bin",
 					"confPath":     "/var/run/multus/cni/net.d",
 					"chainingMode": "portmap",
+					"configMap":    framework.CiliumCNIConfigMapName,
 				},
 				"prometheus": map[string]any{
 					"serviceMonitor": map[string]any{
--- a/test/util/framework/cilium_helper.go
+++ b/test/util/framework/cilium_helper.go
@@ -16,6 +16,7 @@ package framework
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"time"
@@ -26,6 +27,7 @@ import (
 	"helm.sh/helm/v4/pkg/cli"
 	"helm.sh/helm/v4/pkg/kube"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -34,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/utils/ptr"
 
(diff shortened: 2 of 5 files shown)


test/util/framework/cilium_helper.go
```diff
@@ -102,6 +104,72 @@ func InstallCiliumChart(ctx context.Context, chartVersion string, values map[str
...
+      "type": "cilium-cni",
+      "enable-debug": false,
+      "log-file": "/var/run/cilium/cilium-cni.log"
+    },
+    {
+      "type": "portmap",
+      "capabilities": {"portMappings": true}
+    }
+  ]
+}`
```
mbukatov: Minor nitpick: I would rather stored these as a map structure, and then converted it to string using json module, since that will ensure that the json is valid and enable us to tweak it later if needed.

test/e2e/cluster_create_cni_cilium.go
```diff
@@ -104,6 +104,11 @@ var _ = Describe("Customer", func() {
 			kubeconfigContent, err := framework.GenerateKubeconfig(adminRESTConfig)
 			Expect(err).NotTo(HaveOccurred(), "failed to generate kubeconfig for cluster %q", customerClusterName)
 
+			By("providing a Multus-compatible custom CNI conflist for Cilium")
+			const ciliumCNIConfigMapName = "cilium-cni-conf"
```
mbukatov: This constant should be better place in the cilium helper file, isn't it?

mbukatov (commented):
two minor comments, otherwise /lgtm

mbukatov (commented):
thank you for implementing the changes

/lgtm

PR #6883, opened 2026-09-10
fix: exclude soft-deleted MI's from sweeper retention

sclarkso:
https://redhat.atlassian.net/browse/AROSLSRE-2071

### What

Exclude soft-deleted **managed identities** from the cleanup-sweeper's role-assignment retention. `newGraphDeletedPrincipalLookup` now routes through a new pure helper, `deletedPrincipalIsRetainable`, which reports a soft-deleted directory object as retainable *unless* it is a service principal whose `servicePrincipalType` is `ManagedIdentity`. All other soft-deleted principals (app registrations, users, groups, and service principals of any other type) are still retained, and anything that is not a service principal — or a service principal without a reported type — fails safe to retained.

Because both retention paths (the `Discover` deletedItems pass and the immediately-before-delete revalidation) go through that single lookup, this one change covers both. No behavior changes for non-managed-identity principals.

### Why
PR #6762 ([AROSLSRE-1969](https://redhat.atlassian.net/browse/AROSLSRE-1969)) made the sweeper retain the RBAC of *any* soft-deleted principal, to stop background cleanup from stripping access from a shared service principal that is still recoverable and about to be restored. That protection is correct for the incident it was written for: five **app-registration-backed** mock service principals were soft-deleted, restored via Graph, and needed their RBAC intact.

On teardown a managed identity's backing service principal is soft-deleted and sits in the recycle bin for ~30 days. Assignments the sweeper used to delete immediately are now retained for that whole window — even though a managed identity is never restored (a redeploy provisions a new principal ID with fresh assignments), so retaining its RBAC protects nothing and only consumes role-assignment quota.

This change narrows retention back to genuinely restorable principals (app registrations, users, groups), keeping the [AROSLSRE-1969](https://redhat.atlassian.net/browse/AROSLSRE-1969) protection while letting the sweeper reclaim managed-identity grants on its normal hourly run.

### Testing
• New unit test `TestDeletedPrincipalIsRetainable` covering: managed identity (not retained), case-insensitive type match, application SP (retained), SP with no reported type (fail-safe retained), and a non-SP directory object (retained). Existing retention tests (`TestPrincipalRequiresRoleAssignmentRetention`, `TestResolveSoftDeletedPrincipalIDs`, fail-closed variants) unchanged and pass.
• `cd tooling/cleanup-sweeper && go build ./... && go test ./...` — all pass.
• Validated against the live DEV subscription: of 2,881 orphaned assignments (principal absent from the active directory), classifying each principal via Graph `deletedItems` yielded 2,879 `ManagedIdentity` + 2 already-purged and **0** app registrations / users / groups — i.e. this change makes exactly the managed-identity backlog reclaimable and touches nothing restorable.
• Cross-checked the five [AROSLSRE-1969](https://redhat.atlassian.net/browse/AROSLSRE-1969) principals: none appear in the reclaimable set, and the ones spot-checked are currently active (restored) — confirming the incident's protected identities are unaffected.
### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/tooling/cleanup-sweeper/pkg/engine/steps/roleassignments/delete.go
+++ b/tooling/cleanup-sweeper/pkg/engine/steps/roleassignments/delete.go
@@ -26,6 +26,7 @@ import (
 	msgraphsdk "github.com/microsoftgraph/msgraph-sdk-go"
 	graphdirectoryobjects "github.com/microsoftgraph/msgraph-sdk-go/directoryobjects"
 	graphgroups "github.com/microsoftgraph/msgraph-sdk-go/groups"
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	graphodataerrors "github.com/microsoftgraph/msgraph-sdk-go/models/odataerrors"
 
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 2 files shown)


sclarkso:
superseded: https://github.com/Azure/ARO-HCP/pull/6876

PR #6855, opened 2026-09-07
[ARO-28373] Reintroduce ClusterResourcesController

nimrodshn:
https://redhat.atlassian.net/browse/ARO-28373

### What

Reintroduces the reverted #6070 with the addition of #6848 on top.

### Testing

Testing is required for feature completion and tests should be part of the pull
request along with the feature changes.

Describe the testing provided. If you did not add tests, provide a clear
justification.

- Unit tests
- [Integration tests](https://github.com/Azure/ARO-HCP/tree/main/test-integration)
- [E2E tests](https://github.com/Azure/ARO-HCP/blob/main/test/e2e/README.md)

### PR Checklist

- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
- [ ] Screenshots included (if dashboards or other UI changes)
- [ ] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [ ] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

If E2E tests are included:

- [ ] E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md)
- [ ] If new E2E use case is covered (via a new test or new check/verifier),
  demonstrate that the test is able to detect a defect/error and fail with
  proper error message and logs which communicates nature of the problem.

PR #6889, opened 2026-09-10
[ARP-29335] Initial implementation of cluster resources delete path in ClusterResourcesController

nimrodshn:
https://redhat.atlassian.net/browse/ARO-29335

### What

Follow up for #6855 .

1. Introduce a "removal chain" concept in the `ClusterResourcesController` ensuring that the `ApplyDesire` documents are deleted (with their underlying resources, ie `Type=Delete`) in the correct order.
2. Removing the gate that had blocked the cluster dispatch. Now both CS and Kube-Applier attempt to delete the underlying objects. Since CS and kube-applier are both is idempotent (that is no-op on missing resources) having both attempt to delete the resources is safe.

### Why
We introduce a deletion path in the cluster resources controller so that it can take ownership of cluster resources deletion. This is a preparation for moving `ManifestWorks` created by CS to `DeleteOptions=Orphan` so that Maestro/ACM path no longer deletes its managed objects.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


backend/pkg/controllers/clusterresources/apply_desire_destruct_chain.go
```diff
@@ -0,0 +1,182 @@
...
+// kube-applier: the HostedCluster CR would disappear along with the control
+// plane namespace its finalizers need, leaving HyperShift no chance to run its
+// own cleanup and leaving the remaining desires waiting on objects that were
+// removed by the namespace GC rather than by their own Type=Delete desire.
+//
+// The two steps split on targetsNamespace and its negation, so every owned
+// desire is claimed by exactly one of them. That coverage matters: a desire no
+// step claims would never be torn down.
+var applyDesireDestructChain = []applyDesireDestructStep{
+	clusterResourcesDestructStep{},
```
machi1990: At the moment; the PNI and PodNetwork CRs are deleted after the hosted cluster CR is deleted. 
One of the reason was to be sure that the HCP router pods using the PNI has been successfully deleted.
Did you evaluate if deleting it at the same time as the hostedcluster won't have any adverse effect?

backend/pkg/controllers/clusterresources/apply_desire_destruct_chain.go
```diff
@@ -0,0 +1,182 @@
...
+// applyDesireDestructChain, and a step that has not fully drained blocks every
+// later step until a subsequent resync.
+//
+// A step owns both halves of its job: which of the cluster's ApplyDesires it is
+// responsible for, and how those desires are torn down. Nothing outside the
+// step needs to know its selection rule, so a step whose teardown is more than
+// "flip every match to Type=Delete" is free to implement destruct differently.
+//
+// The steps themselves live in apply_desire_destruct_steps.go.
+type applyDesireDestructStep interface {
```
machi1990: the `destruct` terminology isn't something used in the repo; perhaps we want something like


```suggestion
type applyDesireRemovalStep interface {

 remove(
		ctx context.Context,
		kubeApplierDBClient kubeappliercosmosstorage.KubeApplierDBClient,
		owned []*kubeapplierapi.ApplyDesire,
	) (canContinue bool, err error)
```

Can you see how this sounds to you and apply throughout?

backend/pkg/controllers/clusterresources/apply_desire_destruct_chain.go
nimrodshn: Yup this is a major concern.
I was trying to bring it up with @miguelsorianod - 

The way I see it there are two options:
1. Wait in between steps here so that CS can remove the PNI PodNetwork etc.
2. We can record these resources in CS (currently not recorded) and share the deletion chain order in here so that both properly delete. Partially duplicating the logic. (CS is to be removed so this is temporary). 

@deads2k What do you think?

backend/pkg/controllers/clusterresources/apply_desire_destruct_chain.go
```diff
@@ -0,0 +1,182 @@
...
+// Deleting a Namespace tears down everything inside it out from under the
+// kube-applier: the HostedCluster CR would disappear along with the control
+// plane namespace its finalizers need, leaving HyperShift no chance to run its
+// own cleanup and leaving the remaining desires waiting on objects that were
+// removed by the namespace GC rather than by their own Type=Delete desire.
+//
+// The two steps split on targetsNamespace and its negation, so every owned
+// desire is claimed by exactly one of them. That coverage matters: a desire no
+// step claims would never be torn down.
+var applyDesireDestructChain = []applyDesireDestructStep{
```
deads2k: Why isn't the first applyDesireRemovalStep, actually deleting the ApplyDesires for items inside each of the namespaces that will be removed from the management cluster in the cascade.  For instance, any configmap or such that we create, we don't need to individually delete and probably don't actually want to delete it.

backend/pkg/controllers/clusterresources/apply_desire_destruct_chain.go
deads2k: Everything that we must delete, we need to represent here.

backend/pkg/controllers/clusterresources/apply_desire_destruct_chain.go
```diff
@@ -0,0 +1,182 @@
...
+	for _, desire := range existing {
+		if desire.Tags == nil ||
+			desire.Tags[kubeapplierapi.TagControllerName] != ClusterResourcesControllerName {
+			continue
+		}
+		owned = append(owned, desire)
+	}
+
+	for _, step := range applyDesireDestructChain {
+		canContinue, err := step.destruct(ctx, kubeApplierDBClient, owned)
```
deads2k:
```suggestion
		done, err := step.destruct(ctx, kubeApplierDBClient, owned)
```

backend/pkg/controllers/clusterresources/apply_desire_destruct_steps.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (clusterResourcesDestructStep) name() string { return "cluster-resources" }
+
+func (s clusterResourcesDestructStep) destruct(
+	ctx context.Context,
+	kubeApplierDBClient kubeappliercosmosstorage.KubeApplierDBClient,
+	owned []*kubeapplierapi.ApplyDesire,
+) (bool, error) {
+	return deleteMatchingApplyDesires(ctx, kubeApplierDBClient, s.name(), owned,
+		func(desire *kubeapplierapi.ApplyDesire) bool { return !targetsNamespace(desire) })
```
deads2k: Instead of this, we know exactly what the names are. Let's enumerate exactly which ones we want to remove.  I bet it's just the HostedCluster...

backend/pkg/controllers/clusterresources/apply_desire_destruct_steps.go
```diff
@@ -0,0 +1,70 @@
...
+type namespacesDestructStep struct{}
+
+func (namespacesDestructStep) name() string { return "namespaces" }
+
+func (s namespacesDestructStep) destruct(
+	ctx context.Context,
+	kubeApplierDBClient kubeappliercosmosstorage.KubeApplierDBClient,
+	owned []*kubeapplierapi.ApplyDesire,
+) (bool, error) {
+	return deleteMatchingApplyDesires(ctx, kubeApplierDBClient, s.name(), owned, targetsNamespace)
```
deads2k: Same here, we know exatly which name.

stevekuznetsov:
This is a kind of silly drive-by comment that I hope does not come off wrong - many parts of CS are serial for no reason other than they are. We should try to stay away from mental framing of things like 'chain' as I worry old habits kick in and we write more of this serial/pessimistic code rather than what would be truly performant. Thinking through the actual DAG of dependencies on destruction is important.

deads2k:
> This is a kind of silly drive-by comment that I hope does not come off wrong - many parts of CS are serial for no reason other than they are. We should try to stay away from mental framing of things like 'chain' as I worry old habits kick in and we write more of this serial/pessimistic code rather than what would be truly performant. Thinking through the actual DAG of dependencies on destruction is important.

I agree, but the simplest construction of this cannot be completed until cluster-service is no longer the source of content for these resources.

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,287 @@
...
+			logger.Error(
+				fmt.Errorf("ApplyDesire %q is not claimed by any teardown step", desire.ResourceID.Name),
+				"ApplyDesire will not be removed; assign its name to a step in applyDesireRemovalChain",
+				"resourceID", desire.ResourceID.String())
+		}
+		owned = append(owned, desire)
+	}
+
+	for _, step := range applyDesireRemovalChain {
+		done, err := step.destruct(ctx, kubeApplierDBClient, owned)
```
machi1990:
```suggestion
		done, err := step.remove(ctx, kubeApplierDBClient, owned)
```

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+		if !done {
+			return nil
+		}
+	}
+
+	return nil
+}
+
+// deleteMatchingApplyDesires is the teardown primitive the destruct steps share:
```
machi1990:
```suggestion
// deleteMatchingApplyDesires is the teardown primitive the removal steps share:
```

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,287 @@
...
+// it flips every desire in owned that matches selects to Type=Delete and purges
+// the ones the kube-applier reports as deleted. owned must already be filtered
+// to the desires this controller owns. canContinue is true only when nothing
+// matching selects is left.
+func deleteMatchingApplyDesires(
```
machi1990: ensureMatchingApplyDesiresRemoved

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,287 @@
...
+// Cosmos document for every desire in owned that matches selects, leaving
+// spec.targetItem alone. The kube-applier stops reconciling the object, and
+// deleting the namespace it lives in reclaims it.
+//
+// Use this only for objects the namespace cascade genuinely covers. For anything
+// with a finalizer that reaches outside its own namespace, dropping the document
+// strands the object and whatever it holds — use deleteMatchingApplyDesires.
+//
+// There is nothing to wait for, so canContinue is true unless a drop failed.
+func dropMatchingApplyDesires(
```
machi1990: `deleteMatchingApplyDesiresDocuments` 

To indicate that we are only perform the cosmo delete of the document

loggs and comments to be changed in according

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,287 @@
...
+// cascadeCoveredRemovalStep stopped reconciling.
+//
+// Steps claim desires by name, and the names come from classifyClusterResource,
+// which errors on anything it does not recognize. Between the two, the claim
+// sets are exhaustive; deleteAllOwnedApplyDesires warns if a desire nonetheless
+// shows up unclaimed, since nothing would tear it down.
+var applyDesireRemovalChain = []applyDesireRemovalStep{
+	cascadeCoveredRemovalStep{},
+	hostedClusterRemovalStep{},
+	swiftPodNetworkInstanceRemovalStep{},
```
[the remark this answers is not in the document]
machi1990: So the HC depends on the PNI, it makes sense for the HC to be deleted first before the PNI.
CS does it this way.

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,287 @@
...
+// finalizer, which deprovisions Azure infrastructure and tears down the control
+// plane while the namespaces it depends on are still intact. Deleting a
+// Namespace out from under it would garbage collect the HostedCluster CR before
+// that finalizer could complete, stranding Azure resources and, because the
+// namespace cascade honors finalizers, likely wedging the namespace in
+// Terminating.
+//
+// The SWIFT networking resources come after it, in reverse order of creation.
+// Both hold Azure networking state behind finalizers, and PodNetwork is
+// cluster-scoped besides, so neither can be left to the cascade. Cluster Service
```
machi1990: do we need to mention CS in the comment? In the world CS isn't there, we'll need to remember to update the comment which isn't great. We've the ordered deletion not because CS imposes it but due to the dependency of resources as they are plumbed in management cluster so mentioning CS isn't useful.

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,287 @@
...
+	// them out of owned — every ApplyDesire this controller owns for the
+	// cluster. canContinue reports whether the step is drained: false means it
+	// is still waiting on at least one desire and the chain must stop here
+	// until a later resync. A step that returns an error is never treated as
+	// drained.
+	destruct(
+		ctx context.Context,
+		kubeApplierDBClient kubeappliercosmosstorage.KubeApplierDBClient,
+		owned []*kubeapplierapi.ApplyDesire,
+	) (canContinue bool, err error)
```
machi1990:
```suggestion
	) (done bool, err error)
```

To be changed throughout

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,287 @@
...
+	// name identifies the step in log output.
+	name() string
+
+	// destruct tears down the desires this step is responsible for, picking
+	// them out of owned — every ApplyDesire this controller owns for the
+	// cluster. canContinue reports whether the step is drained: false means it
+	// is still waiting on at least one desire and the chain must stop here
+	// until a later resync. A step that returns an error is never treated as
+	// drained.
+	destruct(
```
machi1990:
```suggestion
	// remove tears down the desires this step is responsible for, picking
	// them out of owned — every ApplyDesire this controller owns for the
	// cluster. canContinue reports whether the step is drained: false means it
	// is still waiting on at least one desire and the chain must stop here
	// until a later resync. A step that returns an error is never treated as
	// drained.
    remove (
```

I propose usage of `remove` over destruct

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
```diff
@@ -0,0 +1,246 @@
...
+// namespace delete will garbage collect for us. They are inert — namespaced, and
+// with no finalizer reaching outside their own namespace — so deleting them
+// individually would be slower and would pull configuration out from under the
+// HostedCluster teardown that is still running. Dropping the document leaves the
+// object in place for that teardown to read; only the reconcile stops.
+//
+// NodePool is here because Cluster Service still dispatches node pool deletion
+// itself; see the comment in SyncOnce's node pool branch. Its Azure state is
+// deprovisioned by the HostedCluster finalizer, which hostedClusterRemovalStep
+// waits out, so nothing is stranded by letting the namespace reclaim the CR.
```
machi1990, line 128: @nimrodshn  have you verified that in the world where CS is no longer there it is safe to delete the hostedcluster cr directly without deleting or scaling down the node pool CR?

The part that I want us to verify is that the namespace deletion won't be stuck because of the finalisers because the CAPI pods etc have been removed by the deletion of the HostedCluster CR

nimrodshn: Good catch. Let me verify that.

machi1990: Thank you

nimrodshn: @machi1990 
Tested this branch against https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/387 and verified deletion works correctly.

The following e2e testing was performed:
1. Created a cluster.
2. Waited for it to move to 'ready' state.
3. Created a nodepool.
4. Waited for it to move to 'ready' state.
5. Deleted the **cluster**.
6. Verified deletion was succesful and hostedcluster resource is purged.
7. Verified deletion of all resources finished succesfuly.

machi1990: The NodePool CR has a finaliser
```
 finalizers:
    - hypershift.openshift.io/finalizer
```

From your testing, were you able to pinpoint which component was responsible for the removal of the finaliser and at exactly what moment was that component removed? 

I believe a more sound deletion flow that I'd like to see is a more deterministic deletion for NodePool when a cluster is being deleted and not rely on CS being there and CS doing half the work or the presence of this https://github.com/nimrodshn/ARO-HCP/blob/0aa5ba26cbc706ff38b5c664670648e4413b0e1a/backend/pkg/controllers/clusterresources/cluster_resources_controller.go#L279 which isn't called when the cluster is under deletion.

i.e don't include the NodePool in the CRs that can be deleted in cascade when the namespace are cleaned but rather have a step (could be the same step that deletes the HostedCluster CR) that ensures that NodePool CRs are gone.

nimrodshn: > i.e don't include the NodePool in the CRs that can be deleted in cascade when the namespace are cleaned but rather have a step (could be the same step that deletes the HostedCluster CR) that ensures that NodePool CRs are gone.

Makes sense, ty!.

nimrodshn: @machi1990 I've added a `nodePoolRemovalStep` ahead of `hostedClusterRemovalStep` to ensure proper deletion of NodePool (and underlying Azure resource).

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
```diff
@@ -0,0 +1,276 @@
...
+type namespacesRemovalStep struct{}
+
+func (namespacesRemovalStep) name() string { return "namespaces" }
+
+func (s namespacesRemovalStep) remove(
+	ctx context.Context,
+	kubeApplierDBClient kubeappliercosmosstorage.KubeApplierDBClient,
+	owned []*kubeapplierapi.ApplyDesire,
+) (bool, error) {
+	return ensureMatchingApplyDesiresRemoved(ctx, kubeApplierDBClient, s.name(), owned, namespaceDesireNames.selects)
```
deads2k, line 230: as you write this, every resource we pass to `ensureMatchingApplyDesiresRemoved`, be sure we have `ReadDesires` for them so we can easily see the state of these things to diagnose why something didn't get deleted.

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
```diff
@@ -0,0 +1,276 @@
...
+	_ applyDesireRemovalStep = swiftPodNetworkRemovalStep{}
+	_ applyDesireRemovalStep = namespacesRemovalStep{}
+)
+
+// The desire names classifyClusterResource produces. It returns an error for
+// any resource it does not recognize, so this enumeration is closed: a new
+// resource type cannot reach Cosmos without being added there first, and adding
+// it there without assigning it to one of the sets below trips the
+// unclaimed-desire warning in deleteAllOwnedApplyDesires.
+const (
```
deads2k: seems like these constants should be used when creating the names too, right?

deads2k:
Also, be sure to update the operation cluster delete controller so that it will show a status that indicates insight like htis message, only human readable (and don't share the actual management cluster resource name, only the logical name as done here):


```
{
	"time": "2026-09-14T16:45:29.7452209Z",
	"level": "INFO",
	"source": {
		"function": "github.com/Azure/ARO-HCP/backend/pkg/controllers/clusterresources.ensureMatchingApplyDesiresRemoved",
		"file": "/app/backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go",
		"line": 218
	},
	"msg": "ApplyDesire teardown in progress",
	"resourceType": "microsoft.redhatopenshift/hcpopenshiftclusters",
	"controller_name": "clusterresources",
	"subscription_id": "e8c5a115-842d-4d7e-98ad-cfb2c50b209e",
	"resource_group": "admin-api-breakglass-j2v7m9sjbzjc",
	"resource_name": "sre-hcp-cluster",
	"resource_id": "/subscriptions/e8c5a115-842d-4d7e-98ad-cfb2c50b209e/resourcegroups/admin-api-breakglass-j2v7m9sjbzjc/providers/microsoft.redhatopenshift/hcpopenshiftclusters/sre-hcp-cluster",
	"hcp_cluster_name": "/subscriptions/e8c5a115-842d-4d7e-98ad-cfb2c50b209e/resourcegroups/admin-api-breakglass-j2v7m9sjbzjc/providers/microsoft.redhatopenshift/hcpopenshiftclusters/sre-hcp-cluster",
	"cluster_id": "/subscriptions/e8c5a115-842d-4d7e-98ad-cfb2c50b209e/resourcegroups/admin-api-breakglass-j2v7m9sjbzjc/providers/microsoft.redhatopenshift/hcpopenshiftclusters/sre-hcp-cluster",
	"waitingFor": [
		"controlplanenamespace",
		"hostedclusternamespace"
	],
	"step": "namespaces"
}
```

That will make debugging much faster.

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
nimrodshn: Done.

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
nimrodshn: Done - in a separate commit.

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
```diff
@@ -534,9 +672,9 @@ func (c *clusterResourcesController) deleteStaleApplyDesires(
 			return err
 		}
 		if removed {
-			logger.Info("purged stale ApplyDesire", "desireName", desire.ResourceID.Name)
+			logger.Info("purged stale ApplyDesire", "DesireName", desire.ResourceID.Name)
```
machi1990: Here we need to delete the corresponding stale ReadDesire right?

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,301 @@
...
+		if !done {
+			return nil
+		}
+	}
+
+	// After all ApplyDesires are removed, clean up the ReadDesires that were
+	// created for observability during deletion. ReadDesires are purely
+	// observational and don't manage object lifecycle, so they can be removed
+	// once the ApplyDesire chain completes.
+	if err := c.deleteAllOwnedReadDesires(ctx, key, managementCluster); err != nil {
```
machi1990: We need to ensure that https://github.com/nimrodshn/ARO-HCP/blob/58f695ae1231458d2fb5e4c04ffa08bd2486cedf/backend/pkg/controllers/cluster/deletion/cluster_child_resources_cleanup_controller.go#L396 doesn't clean them too; let's add a gate there

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,301 @@
...
+		removed, err := kubeapplierhelpers.EnsureApplyDesireRemoved(ctx, desireName, crud)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if !removed {
+			waitingFor = append(waitingFor, desireName)
+			continue
+		}
+		logger.Info("deleted ApplyDesire", "step", stepName, "desireName", desireName)
```
machi1990, line 217: Would it have been lean if the deletion of the corresponding read desire is done here instead of line 172? 

i.e once the apply desire is removed, we delete the corresponding read desire.

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
nimrodshn: Yes I added it!

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
nimrodshn: Makes sense.

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
machi1990, line 1: @nimrodshn we probably need to get this controller and all siblings file under the https://github.com/nimrodshn/ARO-HCP/tree/58f695ae1231458d2fb5e4c04ffa08bd2486cedf/backend/pkg/controllers/cluster package. 

At the moment they reside in `backend/pkg/controllers/clusterresources/cluster_resources_controller.go`

@deads2k @nimrodshn thoughts?

deads2k: > David Eads (@deads2k) Nimrod Shneor (@nimrodshn) thoughts?

yeah. Let's get our next set of CI results first.

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
machi1990: Did you forget to push ? @nimrodshn I don't see the change, similar for the other comments

nimrodshn: Done, sorry.

backend/pkg/controllers/cluster/deletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -538,3 +553,49 @@ func (c *clusterChildResourcesCleanupController) extraDeleteGateShouldDeleteAppl
 		return true, nil
 	}
 }
+
+// extraDeleteGateShouldDeleteReadDesire reports whether a cluster-scoped
+// ReadDesire document may be removed here.
+//
+// A ReadDesire that records an owning controller in Tags[TagControllerName]
+// belongs to that controller's teardown: the owner removes it as part of its
+// cleanup sequence. Deleting the document here would interfere with the owner's
```
machi1990: The problem is not all controllers deletes their read desire e.g https://github.com/nimrodshn/ARO-HCP/blob/e090ad852823b5d4e829ba97591df7b261347a67/backend/pkg/controllers/cluster/readdesires/create_cluster_scoped_read_desires_controller.go and https://github.com/nimrodshn/ARO-HCP/blob/e090ad852823b5d4e829ba97591df7b261347a67/backend/pkg/controllers/nodepool/readdesires/create_nodepool_scoped_read_desires_controller.go doesn't seem to delete its own read desire.

The cheapest of the fix is to ensure that those deletes their own read desire or here we only gate against the read desire we want to be skipped i.e the ones owned by this resources controller

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
```diff
@@ -315,6 +283,53 @@ func (c *clusterResourcesController) processClusterResources(ctx context.Context
...
+					classified.desireName, managementCluster, target, tags,
+				)
+			}
+
+			if readErr != nil {
+				errs = append(errs, readErr)
+				continue
+			}
+
+			if err := kubeapplierhelpers.EnsureReadDesire(ctx, readCRUD, c.readDesireLister, readDesire); err != nil {
```
machi1990, line 330: A HostedCluster and NodePool CRs have already a read desire created.
its done by these controllers  https://github.com/nimrodshn/ARO-HCP/blob/e090ad852823b5d4e829ba97591df7b261347a67/backend/pkg/controllers/cluster/readdesires/create_cluster_scoped_read_desires_controller.go and https://github.com/nimrodshn/ARO-HCP/blob/e090ad852823b5d4e829ba97591df7b261347a67/backend/pkg/controllers/nodepool/readdesires/create_nodepool_scoped_read_desires_controller.go respectively; by additionally creating one we'll have 2 read desires created for these resources in the backend and an additional 1 read desire for each CR that CS creates; so a total of 3 for each CR. 
I wonder if the backend can simply create 1 read desire for each CR - or we are fine with this as its meant to be temporary situation?

backend/pkg/controllers/cluster/deletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -538,3 +553,49 @@ func (c *clusterChildResourcesCleanupController) extraDeleteGateShouldDeleteAppl
...
+			return false, nil
+		}
+		if err != nil {
+			return false, utils.TrackError(fmt.Errorf("failed to get ReadDesire %q: %w", readDesireResourceID.String(), err))
+		}
+
+		if owningController := readDesire.Tags[kubeapplierapi.TagControllerName]; len(owningController) > 0 {
+			logger.Info("waiting for owning controller to tear down cluster-scoped ReadDesire",
+				"readDesireResourceID", readDesireResourceID.String(), "owningController", owningController)
+			return false, nil
```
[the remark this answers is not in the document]
nimrodshn: @deads2k I dont have anything smart to say here. Theoretically, every document should have been created by some controller thus it should be deleted in _some_ path.

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
```diff
@@ -315,6 +283,53 @@ func (c *clusterResourcesController) processClusterResources(ctx context.Context
 		if err := kubeapplierhelpers.EnsureApplyDesire(ctx, crud, c.applyDesireLister, desire); err != nil {
 			errs = append(errs, err)
 		}
+
+		// Ensure ReadDesire exists for resources that go through
+		// ensureMatchingApplyDesiresRemoved during deletion, so operators can
+		// observe their state and debug deletion issues (e.g., stuck finalizers).
+		if shouldCreateReadDesireFor(classified.desireName) {
```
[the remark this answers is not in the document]
nimrodshn: Done.

backend/pkg/controllers/cluster/deletion/cluster_child_resources_cleanup_controller.go
nimrodshn: Yeah I was just wondering regarding what to do in such situations as copilot mentioned this as well..

nimrodshn: I implemented -
> here we only gate against the read desire we want to be skipped i.e the ones owned by this resources controller

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
nimrodshn: So `EnsureReadDesire` has a "create if not existing" semantics so if we keep the resource ID consistent there should be one document per CR, let me verify if that is the case.

backend/pkg/controllers/cluster/deletion/cluster_child_resources_cleanup_controller.go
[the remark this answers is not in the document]
nimrodshn: I chose to take @machi1990's offer in https://github.com/Azure/ARO-HCP/pull/6889#discussion_r4017286451

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
```diff
@@ -535,6 +622,22 @@ func (c *clusterResourcesController) deleteStaleApplyDesires(
...
+			// deleteStaleApplyDesires removes a resource no longer returned by Cluster
+			// Service. Without this cleanup, the orphaned ReadDesire would be preserved
+			// by the ownership gate during cluster deletion and never get cleaned up.
+			readDesireCRUD, readErr := kubeApplierDBClient.ReadDesiresFor(scope)
+			if readErr != nil {
+				logger.Error(readErr, "failed to get ReadDesire CRUD for stale cleanup", "desireName", desire.ResourceID.Name)
+			} else {
+				if delErr := readDesireCRUD.Delete(ctx, desire.ResourceID.Name); delErr != nil && !cosmosstorageutils.IsNotFoundError(delErr) {
+					logger.Error(delErr, "failed to delete stale ReadDesire", "desireName", desire.ResourceID.Name)
+				} else if delErr == nil {
```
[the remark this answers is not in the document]
machi1990: Ah damn; this is sadly true. I suggested this approach without taking in consideration failure scenarios. Apologies @nimrodshn

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
```diff
@@ -315,6 +283,53 @@ func (c *clusterResourcesController) processClusterResources(ctx context.Context
 		if err := kubeapplierhelpers.EnsureApplyDesire(ctx, crud, c.applyDesireLister, desire); err != nil {
 			errs = append(errs, err)
 		}
```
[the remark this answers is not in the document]
nimrodshn: Fixed the deletion of ReadDesire.

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,329 @@
...
+		// management cluster, delete the corresponding ReadDesire. It was only
+		// needed for observability during deletion; now that deletion is complete,
+		// it's no longer useful.
+		readDesireCRUD, readErr := readDesireCRUDFor(kubeApplierDBClient, desire)
+		if readErr != nil {
+			// Log but don't fail the step - ReadDesire cleanup is best-effort
+			logger.Error(readErr, "failed to get ReadDesire CRUD", "desireName", desireName)
+		} else {
+			if delErr := readDesireCRUD.Delete(ctx, desireName); delErr != nil && !cosmosstorageutils.IsNotFoundError(delErr) {
+				logger.Error(delErr, "failed to delete ReadDesire", "desireName", desireName)
```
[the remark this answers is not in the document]
nimrodshn: Fixed to retry this cleanup.

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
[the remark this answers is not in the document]
nimrodshn: Fixed such that the controller will retry to purge both.

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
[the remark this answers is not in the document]
nimrodshn: Fixed in the deletion path such that created read desires will be deleted and retried.

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
```diff
@@ -525,11 +612,27 @@ func (c *clusterResourcesController) deleteStaleApplyDesires(
...
+		// visit this resource again.
+		readDesireCRUD, err := kubeApplierDBClient.ReadDesiresFor(scope)
+		if err != nil {
+			return utils.TrackError(fmt.Errorf("get ReadDesire CRUD for stale cleanup of %s: %w", desire.ResourceID.Name, err))
+		}
+		if err := readDesireCRUD.Delete(ctx, desire.ResourceID.Name); err != nil && !cosmosstorageutils.IsNotFoundError(err) {
+			return utils.TrackError(fmt.Errorf("delete stale ReadDesire %s: %w", desire.ResourceID.Name, err))
+		}
+		logger.Info("purged stale ReadDesire", "desireName", desire.ResourceID.Name)
+
```
machi1990: The read desire are there to track and be able to debug unsuccessful delete of apply desire.
As such we want them to be deleted after the apply desires has been deleted.

The first suggestion in this comment https://github.com/Azure/ARO-HCP/pull/6889#discussion_r4024152664 

> the orphan can therefore block cluster cleanup permanently. Delete the ReadDesire before purging the ApplyDesire or add an independent retry/sweep instead of swallowing this error.

is wrong so, I suggest we take the second and more precisely what was there before this comment https://github.com/Azure/ARO-HCP/pull/6889#discussion_r4015447044

i.e purgeAll the read desires after all the apply desires are gone. 

@nimrodshn let me know if it makes sense, again sorry for the back & forth on this one after the misleading comment I left earlier on.

nimrodshn: Fixed ty!

backend/pkg/controllers/cluster/deletion/cluster_child_resources_cleanup_controller.go
```diff
@@ -389,11 +390,26 @@ func (c *clusterChildResourcesCleanupController) ensureClusterScopedKubeApplierR
 		return skipBackupDesires(ctx, resourceID)
 	}
 
+	// Combined gate for ClusterScopedReadDesire: first check ownership (owned ReadDesires
+	// must be torn down by their controller), then check if it's a backup desire.
+	readDesireGate := func(ctx context.Context, resourceID *azcorearm.ResourceID) (bool, error) {
```
machi1990, line 395: A similar gate is needed for the nodepool's corresponding controller https://github.com/nimrodshn/ARO-HCP/blob/c9c409f71f8ce26543b21bfb0b42bcdf5dd5edd5/backend/pkg/controllers/nodepool/deletion/node_pool_child_resources_cleanup_controller.go#L284

machi1990 (commented):
@nimrodshn we also need to improve https://github.com/nimrodshn/ARO-HCP/blob/e62b21b8caecd8a5b7828215bc27371c35daa9b8/backend/pkg/controllers/clusterresources/cluster_resources_controller.go#L584 to ensure that stale read desire are also removed post the removal of stale apply desire

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
```diff
@@ -315,6 +283,53 @@ func (c *clusterResourcesController) processClusterResources(ctx context.Context
...
+		// ensureMatchingApplyDesiresRemoved during deletion, so operators can
+		// observe their state and debug deletion issues (e.g., stuck finalizers).
+		if shouldCreateReadDesireFor(classified.desireName) {
+			var readDesire *kubeapplierapi.ReadDesire
+			var readErr error
+			var readCRUD cosmosstorageutils.ResourceCRUD[kubeapplierapi.ReadDesire, *kubeapplierapi.ReadDesire]
+
+			if len(classified.nodePoolName) != 0 {
+				readCRUD, readErr = kubeApplierDBClient.ReadDesiresForNodePool(
+					key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName, classified.nodePoolName,
```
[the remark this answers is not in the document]
nimrodshn: Fixed in the backend/pkg/controllers/nodepool/deletion/node_pool_child_resources_cleanup_controller.go

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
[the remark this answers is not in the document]
machi1990: +1, this needs to be addressed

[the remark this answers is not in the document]
machi1990: +1 as something that needs to be addressed

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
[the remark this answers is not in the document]
machi1990: +1 as something to be addressed

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
```diff
@@ -525,16 +615,31 @@ func (c *clusterResourcesController) deleteStaleApplyDesires(
...
+			// controller outlive their ApplyDesire when deleteStaleApplyDesires removes
+			// a resource no longer returned by Cluster Service. Without this cleanup, the
+			// orphaned ReadDesire would be preserved by the ownership gate during cluster
+			// deletion and never get cleaned up.
+			readDesireCRUD, readErr := kubeApplierDBClient.ReadDesiresFor(scope)
+			if readErr != nil {
+				return utils.TrackError(fmt.Errorf("get ReadDesire CRUD for stale cleanup of %s: %w", desire.ResourceID.Name, readErr))
+			}
+			if delErr := readDesireCRUD.Delete(ctx, desire.ResourceID.Name); delErr != nil && !cosmosstorageutils.IsNotFoundError(delErr) {
+				return utils.TrackError(fmt.Errorf("delete stale ReadDesire %s: %w", desire.ResourceID.Name, delErr))
```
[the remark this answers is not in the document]
machi1990: +1 this needs to be addressed.

We talked about in https://github.com/Azure/ARO-HCP/pull/6889#discussion_r4025562252

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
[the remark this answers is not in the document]
machi1990: +1 this needs to be fixed.

I see it being addressed similarly to comment as https://github.com/Azure/ARO-HCP/pull/6889#discussion_r4025562252

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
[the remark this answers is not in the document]
nimrodshn: Done.

backend/pkg/controllers/clusterresources/cluster_resources_controller.go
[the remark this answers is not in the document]
nimrodshn: Done.

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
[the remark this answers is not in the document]
nimrodshn: Done same as https://github.com/Azure/ARO-HCP/pull/6889#discussion_r4028141750

config/config.yaml
```diff
@@ -1113,25 +1108,25 @@ defaults:
...
-        digest: sha256:a7027158c0b762347e55d5c4212b6745b21df83421b874f37d02b1b87e4be016 # v2.17.3-497 (2026-09-22 21:10)
+        digest: sha256:da8ebe662611d03e32b5b524837c20e8eb1fadfd9e845ef313912b7fea0731b3 # v2.17.3-495 (2026-09-21 14:20)
   # Cluster Service
   clustersService:
     image:
       registry: quay.io
       repository: app-sre/aro-hcp-clusters-service
-      digest: sha256:ea75b70cae009f8d77d8d0857f7bdb011ff55c9f867c121962849dec25da3867 # 2c53dc63606ae24e7765a34b5579ca3fc8f3c92d (2026-09-21 17:44)
+      digest: sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6 # Testing CS PR 387
     # controls how many HCPs can be placed on an MC, sadly not configurable per MC
```
machi1990: You forgot to bump the registry and repository 

```
quay.io/app-sre/aro-hcp-clusters-service@sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6: not found\nCommand failed (attempt 1/5). Retrying in 2s...\nError response from registry: failed to resolve sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6: quay.io/app-sre/aro-hcp-clusters-service@sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6: not found\nCommand failed (attempt 2/5). Retrying in 4s...\nError response from registry: failed to resolve sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6: quay.io/app-sre/aro-hcp-clusters-service@sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6: not found\nCommand failed (attempt 3/5). Retrying in 8s...\nError response from registry: failed to resolve sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6: quay.io/app-sre/aro-hcp-clusters-service@sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6: not found\nCommand failed (attempt 4/5). Retrying in 16s...\nError response from registry: failed to resolve sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6: quay.io/app-sre/aro-hcp-clusters-service@sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6: not found\nCommand failed after 5 attempts: oras cp quay.io/app-sre/aro-hcp-clusters-service@sha256:299c08f05b5b37da0e9c8b727bd0f57e519da49d0bb6daf96b57d0744493e5a6 arohcpsvcdev.azurecr.io/app-sre/aro-hcp-clusters-
```

nimrodshn: Done, ty!

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
```diff
@@ -0,0 +1,269 @@
...
+// order they run in live in apply_desire_removal_chain.go.
+
+var (
+	_ applyDesireRemovalStep = cascadeCoveredRemovalStep{}
+	_ applyDesireRemovalStep = nodePoolRemovalStep{}
+	_ applyDesireRemovalStep = hostedClusterRemovalStep{}
+	_ applyDesireRemovalStep = swiftPodNetworkInstanceRemovalStep{}
+	_ applyDesireRemovalStep = swiftPodNetworkRemovalStep{}
+	_ applyDesireRemovalStep = namespacesRemovalStep{}
+)
```
machi1990:
```suggestion
var (
	_applyDesireRemovalStep = managedClusterRemovalStep{}
	_ applyDesireRemovalStep = cascadeCoveredRemovalStep{}
	_ applyDesireRemovalStep = nodePoolRemovalStep{}
	_ applyDesireRemovalStep = hostedClusterRemovalStep{}
	_ applyDesireRemovalStep = swiftPodNetworkInstanceRemovalStep{}
	_ applyDesireRemovalStep = swiftPodNetworkRemovalStep{}
	_ applyDesireRemovalStep = namespacesRemovalStep{}
)
```
np

backend/pkg/controllers/clusterresources/apply_desire_removal_chain.go
```diff
@@ -0,0 +1,405 @@
...
+// cascadeCoveredRemovalStep stopped reconciling.
+//
+// Steps claim desires by name, and the names come from classifyClusterResource,
+// which errors on anything it does not recognize. Between the two, the claim
+// sets are exhaustive; deleteAllOwnedApplyDesires warns if a desire nonetheless
+// shows up unclaimed, since nothing would tear it down.
+var applyDesireRemovalChain = []applyDesireRemovalStep{
+	cascadeCoveredRemovalStep{},
+	nodePoolRemovalStep{},
+	managedClusterRemovalStep{},
```
machi1990:
```suggestion
	managedClusterRemovalStep{},
	nodePoolRemovalStep{},
```

nimrodshn: Done,

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
nimrodshn: Done.

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
```diff
@@ -0,0 +1,270 @@
...
+// HostedCluster teardown that is still running. Dropping the document leaves the
+// object in place for that teardown to read; only the reconcile stops.
+//
+// Nothing with a finalizer belongs here. NodePool in particular does not: see
+// nodePoolDesireNames.
+var cascadeCoveredDesireNames = newDesireNameSet(
+	DesireNameDefaultIngressConfigMap,
+	DesireNameOCPPullSecret,
+	DesireNameBoundServiceAccountSigningKeySecretSync,
+	DesireNameDefaultIngressWildcardCertSecretSync,
```
machi1990: These don't live in neither the HCP Namespace nor the HC namespace.
They live in the `open-cluster-management-policies` (see https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/pkg/clusterprovisioner/acm/aro/hosted_cluster_provision_step.go#L68 and https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/pkg/clusterprovisioner/acm/aro/hosted_cluster_provision_step.go#L307) which isn't deleted as part of cluster deletion. 
We need to delete them properly and ensure their removal as part of cluster deletion

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
```diff
@@ -0,0 +1,270 @@
...
+	DesireNameKubeAPIServerServingCertSecretSync,
+	DesireNameBoundServiceAccountSigningKeySecretProviderClass,
+	DesireNameDefaultIngressWildcardCertSecretProviderClass,
```
machi1990: Same comment as https://github.com/Azure/ARO-HCP/pull/6889#discussion_r4098937165

backend/pkg/controllers/clusterresources/apply_desire_removal_steps.go
```diff
@@ -0,0 +1,270 @@
...
+// namespace delete will garbage collect for us. They are inert — namespaced, and
+// with no finalizer reaching outside their own namespace — so deleting them
+// individually would be slower and would pull configuration out from under the
+// HostedCluster teardown that is still running. Dropping the document leaves the
+// object in place for that teardown to read; only the reconcile stops.
+//
+// Nothing with a finalizer belongs here. NodePool in particular does not: see
+// nodePoolDesireNames.
+var cascadeCoveredDesireNames = newDesireNameSet(
+	DesireNameDefaultIngressConfigMap,
```
machi1990: same comment as https://github.com/Azure/ARO-HCP/pull/6889#discussion_r4098937165

machi1990: For this one specifically see https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/pkg/clusterprovisioner/acm/aro/hosted_cluster_provision_step.go#L339

PR #6900, opened 2026-09-10
gate v5 upgrades without data-plane mirror

YuriyNik:
(https://redhat.atlassian.net/browse/ARO-29503)

### What
Adds the v5 data-plane mirror upgrade gate. The backend mirrors whether the HostedCluster contains the ocp-v5.0-art-dev image source, and frontend admission rejects 4.x → 5.x upgrades when the mirror is explicitly absent.

  ### Why

  OpenShift 5.x upgrades require the v5 data-plane mirror. This validation prevents unsupported upgrades while preserving fail-open behavior for unknown mirror state.

### Testing

justification.

- Unit tests

• ### Special notes for your reviewer

  The gate applies only to 4.x → 5.x upgrades when the mirror state is explicitly false. Missing or nil state fails open. Same-major upgrades and 5.x z-stream upgrades are unaffected. Tests cover mirror presence, absence, nil state,
  missing SPCs, and cross-major upgrade scenarios.

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go
```diff
@@ -211,3 +215,21 @@ func getHostedClusterDesiredVersionChannels(hostedCluster *hsv1beta1.HostedClust
 	}
 	return hostedCluster.Status.Version.Desired.Channels
 }
+
+func getHostedClusterV5MirrorPresent(hostedCluster *hsv1beta1.HostedCluster) *bool {
```
stevekuznetsov: You never return `nil`, so why is this `*bool`?

YuriyNik: good catch Steve! will fix this !

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -204,6 +204,13 @@ type ServiceProviderClusterStatus struct {
 	// Written by: ControlPlaneActiveVersions
 	DesiredVersionChannels []string `json:"desiredVersionChannels,omitempty"`
 
+	// DataPlaneV5MirrorPresent reports whether the observed HostedCluster carries the
+	// ocp-v5.0-art-dev data-plane image mirror in Spec.ImageContentSources. Mirrored by the
+	// backend so DB-free admission can block a 4->5 upgrade when the mirror is missing.
+	// nil means "not yet observed" (admission treats it as allow / fail-open).
+	// Written by: ControlPlaneActiveVersions
+	DataPlaneV5MirrorPresent *bool `json:"dataPlaneV5MirrorPresent,omitempty"`
```
machi1990: I been thinking about this and I am not a big fan of this addition and the way that the active version controller has been modified in consequence. So I've a different suggestion, let me know what do you  @YuriyNik @stevekuznetsov @deads2k think about it;

The suggestion is to leverage the existing `DesiredHostedCluster` https://github.com/YuriyNik/ARO-HCP/blob/36dee781c5cbe786715b73b209f6f7dd3f624e08/internal/api/coreapi/types_serviceprovider_cluster.go#L105 field

> // DesiredHostedCluster is the HostedCluster that we want to exist on the management cluster.
	// We will only explicitly set the fields we care about, but serialization may store additional empty fields.
	// Once this contains the critical values, we will create it on management clusters.
	// We may or may not choose to store the actual state in status.  We may choose to store the actual state independently.
	DesiredHostedCluster *v1beta1.HostedCluster `json:"desiredHostedCluster,omitempty"`

At present, this isn't filled by any controller. But we've some elements already that allows us to create a new dedicate controller that backfills the field by reading the `/resources` endpoint from CS similar to https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/clusterresources/cluster_resources_controller.go 

From the returned list; the resource that corresponds to the HostedCluster will be the one that is named with `kind.namespace.name`; see the naming pattern of such resources here https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/pkg/clusterprovisioner/resource_naming.go#L40 

In this proposal: 
- the backfilling controller will reside up until when the backend knows how to construct the HostedCluster and not read it from CS (work ongoing on that).
- The backfilling controller will only needs to sync if the cluster isn't being deleted
- The backfilling controller needs to sync only when the corresponding CS cluster has been created
- The backfilling controller should be designed in way that its aware that during early cluster provisioning phases, the /resources endpoint might not contain the HostedCluster resources un until when https://github.com/openshift-online/aro-hcp-clusters-service/blob/master/pkg/clusterprovisioner/acm/aro/hosted_cluster_provision_step.go step is reached in CS. So its important that it doesn't send an error in such cases.
- We'll need to gate cluster creation operation https://github.com/Azure/ARO-HCP/blob/main/backend/pkg/controllers/cluster/operations/operation_cluster_create.go on the DesiredHostedCluster being written so that we are sure that the controller does its job

internal/admission/admit_cluster.go
```diff
@@ -728,6 +729,38 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
...
+	}
+	// This backstop intentionally covers only upgrades into OpenShift 5.x; a
+	// future major version needs its own mirror status and admission rule.
+	if newVersion.Major != 5 || oldVersion.Major >= newVersion.Major {
+		return nil
+	}
+	if admissionContext.ServiceProviderCluster == nil {
+		return nil
+	}
+	present := admissionContext.ServiceProviderCluster.Status.DataPlaneV5MirrorPresent
```
machi1990: In light to https://github.com/Azure/ARO-HCP/pull/6900/#discussion_r4016499618 (assuming we've agreement on that suggestion) this logic will need to change so that it reads from `Spec.DesiredHostedCluster.Spec.ImageContentSources` and inspect the entry that we want to be contained.

internal/api/coreapi/types_serviceprovider_cluster.go
machi1990: @YuriyNik following discussion on chat, my suggestion above has received refinement thoughts and we want the following.

> The suggestion is to leverage the existing DesiredHostedCluster https://github.com/YuriyNik/ARO-HCP/blob/36dee781c5cbe786715b73b209f6f7dd3f624e08/internal/api/coreapi/types_serviceprovider_cluster.go#L105 field

Instead of this, let's add a new field in https://github.com/YuriyNik/ARO-HCP/blob/36dee781c5cbe786715b73b209f6f7dd3f624e08/internal/api/coreapi/types_serviceprovider_cluster.go Status named `ActualHostedCluster *v1beta1.HostedCluster json:"actualHostedCluster,omitempty"`

And a new controller that fills writes this value; The controller shape will be a bit similar to backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go but simply write the observed HostedClusterCR.

In the frontend admission checks; https://github.com/Azure/ARO-HCP/pull/6900#discussion_r4016516001  it now opens the path to perform the check on `.Status.ActualHostedCluster.Spec.ImageContentSources`.

Doing it this way; removes the need to declare a `bool` type field that doesn't cater for future needs e.g what do we do when v6 is there?

And a dedicated controller keeps each controller single responsible for things its supposed to do; the active version controller remains strictly related to version actuation.

The new controller is responsible for writes of the observed HostedCluster CR.

The frontend now has all the data it needs from the HostedCluster CR to perform the now and future validations as they are needed.

In the followup future we could potentially, remove https://github.com/YuriyNik/ARO-HCP/blob/36dee781c5cbe786715b73b209f6f7dd3f624e08/internal/api/coreapi/types_serviceprovider_cluster.go#L205 `DesiredVersionChannels` as well and perform the corresponding observation from the `.Status.ActualHostedCluster.Status..`

internal/admission/admit_cluster.go
machi1990: See updated/revisited suggestion https://github.com/Azure/ARO-HCP/pull/6900#discussion_r4017866333

machi1990 (changes requested):
Requested changes on how to structure the serviceprovidercluster type to cater for the now and future needs, plus a dedicated controller that goes with it.

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -204,6 +204,33 @@ type ServiceProviderClusterStatus struct {
...
+	// The frontend is deliberately denied any path to a management cluster (no
+	// kube-applier container access, no ReadDesireLister, no Maestro), so a
+	// frontend compromise cannot create arbitrary resources on a management
+	// cluster. Anything admission needs to know about the real HostedCluster has
+	// to travel through this field; see internal/admission/CLAUDE.md.
+	//
+	// Mirrored verbatim except for server-side bookkeeping that carries no
+	// meaning for consumers and would otherwise force a Cosmos write on every
+	// observed revision: metadata.managedFields, metadata.resourceVersion and the
+	// kubectl last-applied-configuration annotation are cleared before storing.
```
machi1990: I am thinking we can actually do a mirror of the stanza as is.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
machi1990, line 1:
```
backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
```

have `hostedcluster` as the package name

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -308,6 +335,12 @@ type ServiceProviderClusterStatus struct {
 	KeyRotationBackupFingerprint string `json:"keyRotationBackupFingerprint,omitempty"`
 }
 
+// OcpV5ArtDevMirrorSource is the platform-managed image source that OpenShift
+// 5.x data-plane releases are published under. A cluster whose HostedCluster
+// spec.imageContentSources lacks this entry cannot pull 5.x data-plane images,
+// so an upgrade into 5.x would strand its nodes.
+const OcpV5ArtDevMirrorSource = "quay.io/openshift-release-dev/ocp-v5.0-art-dev"
```
machi1990: Ths can probably go in https://github.com/YuriyNik/ARO-HCP/tree/6f78e7eb00a7a1808a38dc2aeb66bfa9e9dbe3b1/internal/utils/apihelpers and named as

internal/admission/admit_cluster.go
```diff
@@ -728,6 +733,83 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
...
+func admitClusterV5DataPlaneMirror(_ context.Context, admissionContext *ClusterAdmissionContext, op operation.Operation, fldPath *field.Path, newObj, oldObj *coreapi.VersionProfile) field.ErrorList {
+	// Only enforce on UPDATE, and only when the customer is actually changing
+	// version.id. On CREATE there is no prior version to cross a major boundary
+	// from, and an unchanged version is not an upgrade.
+	if op.Type != operation.Update || oldObj == nil {
+		return nil
+	}
+	if len(newObj.ID) == 0 || oldObj.ID == newObj.ID {
+		return nil
+	}
```
machi1990: Already enforced by the caller here https://github.com/YuriyNik/ARO-HCP/blob/6f78e7eb00a7a1808a38dc2aeb66bfa9e9dbe3b1/internal/admission/admit_cluster.go#L578 so we can drop this

internal/admission/admit_cluster.go
```diff
@@ -728,6 +733,83 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
...
+			return nil
+		}
+	}
+
+	versionPath := fldPath.Child("id")
+	return field.ErrorList{field.Invalid(
+		versionPath,
+		newObj.ID,
+		fmt.Sprintf("cannot upgrade to OpenShift %d.x: this cluster is missing the required data-plane image "+
+			"mirror %q, so its nodes would be unable to pull %d.x images; contact support to have the mirror "+
```
machi1990:
```suggestion
			"mirror %q, so its nodes would be unable to pull %d.y images; contact support to have the mirror "+
```

internal/admission/admit_cluster.go
```diff
@@ -728,6 +733,83 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
...
+		if imageContentSource.Source == coreapi.OcpV5ArtDevMirrorSource {
+			return nil
+		}
+	}
+
+	versionPath := fldPath.Child("id")
+	return field.ErrorList{field.Invalid(
+		versionPath,
+		newObj.ID,
+		fmt.Sprintf("cannot upgrade to OpenShift %d.x: this cluster is missing the required data-plane image "+
```
machi1990:
```suggestion
		fmt.Sprintf("cannot upgrade to OpenShift %d.y: this cluster is missing the required data-plane image "+
```

internal/admission/admit_cluster.go
```diff
@@ -728,6 +733,83 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
...
+	if oldErr != nil || newErr != nil {
+		// admitClusterVersionProfile already reports parse failures; do not
+		// duplicate them here.
+		return nil
+	}
+	// Scoped to upgrades that cross into 5.x. A future major version ships its
+	// own mirror and needs its own rule rather than a widened one here.
+	if newVersion.Major != 5 || oldVersion.Major >= newVersion.Major {
+		return nil
+	}
```
machi1990:
```suggestion
	// Scoped to upgrades that cross into 5.y. A future major version ships its
	// own mirror and needs its own rule rather than a widened one here.
	if newVersion.Major != 5 && oldVersion.Major != 4 {
		return nil
	}
```

I think we probably want this explicit guard.

internal/admission/admit_cluster.go
```diff
@@ -728,6 +733,83 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
...
+	actualHostedCluster := admissionContext.ServiceProviderCluster.Status.ActualHostedCluster
+	if actualHostedCluster == nil {
+		return nil
+	}
+
+	for _, imageContentSource := range actualHostedCluster.Spec.ImageContentSources {
+		if imageContentSource.Source == coreapi.OcpV5ArtDevMirrorSource {
+			return nil
+		}
+	}
```
machi1990: np: Will this read nicesly if we use: https://pkg.go.dev/slices#ContainsFunc ?

backend/pkg/controllers/cluster/actualhostedcluster/actual_hosted_cluster_controller.go
```diff
@@ -0,0 +1,170 @@
...
+		kubeApplierInformers,
+		5*time.Minute,
+		syncer,
+	)
+}
+
+// SyncOnce mirrors the observed HostedCluster onto the ServiceProviderCluster,
+// writing only when the sanitized object differs from what is already stored.
+func (c *actualHostedClusterSyncer) SyncOnce(ctx context.Context, key controllerutils.HCPClusterKey) error {
+	existingCluster, err := c.resourcesDBClient.HCPClusters(key.SubscriptionID, key.ResourceGroupName).Get(ctx, key.HCPClusterName)
```
machi1990: read from the lister just like https://github.com/YuriyNik/ARO-HCP/blob/6f78e7eb00a7a1808a38dc2aeb66bfa9e9dbe3b1/backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go#L148

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
YuriyNik: Done. Moved the controller to backend/pkg/controllers/cluster/hostedcluster with package name hostedcluster.

internal/api/coreapi/types_serviceprovider_cluster.go
YuriyNik: Done. Added apihelpers.OcpV5ArtDevMirrorSource and HostedClusterHasImageContentSource under internal/utils/apihelpers.

internal/admission/admit_cluster.go
YuriyNik: Done. Removed the redundant update/precondition guard; yes, the caller already enforces it.

internal/admission/admit_cluster.go
YuriyNik: Done. Updated the error wording to use %d.y.

internal/admission/admit_cluster.go
YuriyNik: Done. The error now formats the mirror source with %q.y

internal/admission/admit_cluster.go
YuriyNik: Used the correct De Morgan equivalent: oldVersion.Major != 4 || newVersion.Major != 5. This gates only 4.y → 5.y; 4.y → 4.y and 4.y → 6.y remain ungated.

internal/admission/admit_cluster.go
YuriyNik: Done. The check now uses apihelpers.HostedClusterHasImageContentSource, backed by slices.ContainsFunc.

backend/pkg/controllers/cluster/actualhostedcluster/actual_hosted_cluster_controller.go
YuriyNik: Done. The controller now receives and uses clusterLister, following the existing controller pattern.

internal/api/coreapi/types_serviceprovider_cluster.go
YuriyNik: Done. The controller now mirrors the HostedCluster verbatim. This preserves the complete observed state for DB-free admission, but HostedCluster status changes may trigger an SPC Replace and changefeed
  notification for each revision. This increases Cosmos RU usage and wakes SPC watchers; the behavior is described in docs/cosmos-data-flow.md.

internal/admission/admit_cluster.go
```diff
@@ -728,6 +733,76 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
...
+	actualHostedCluster := admissionContext.ServiceProviderCluster.Status.ActualHostedCluster
+	if actualHostedCluster == nil {
+		return nil
+	}
+
+	if apihelpers.HostedClusterHasImageContentSource(actualHostedCluster, apihelpers.OcpV5ArtDevMirrorSource) {
+		return nil
+	}
+
+	versionPath := fldPath.Child("id")
```
machi1990, line 796: np: can be inlined

YuriyNik: ok, sure

internal/api/coreapi/types_serviceprovider_cluster.go
YuriyNik: Implemented the revised approach: added Status.ActualHostedCluster and a dedicated controller to mirror the observed HostedCluster. Removed DataPlaneV5MirrorPresent and the active-version controller changes. Spec.DesiredHostedCluster remains unchanged.

internal/admission/admit_cluster.go
YuriyNik: Updated. Admission now reads Status.ActualHostedCluster.Spec.ImageContentSources and checks for the required mirror. A nil actual HostedCluster still fails open.

internal/admission/admit_cluster.go
```diff
@@ -718,16 +723,87 @@ func admitClusterVersionID(_ context.Context, admissionContext *ClusterAdmission
...
+
+	// Not observed yet: fail closed (see the doc comment above). Admission must
+	// have an observed HostedCluster before allowing a 4.y -> 5.y upgrade.
+	if admissionContext.ServiceProviderCluster == nil {
+		return missingV5DataPlaneMirrorError(fldPath, newObj.ID, newVersion.Major)
+	}
+	actualHostedCluster := admissionContext.ServiceProviderCluster.Status.ActualHostedCluster
+	if actualHostedCluster == nil {
+		return missingV5DataPlaneMirrorError(fldPath, newObj.ID, newVersion.Major)
+	}
```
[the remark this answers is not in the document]
machi1990: @YuriyNik you probably want to update the PR's description

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
```diff
@@ -0,0 +1,149 @@
...
+
+	hostedCluster, err := kubeapplierhelpers.GetCachedHostedClusterForCluster(ctx, c.readDesireLister, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get HostedCluster from ReadDesire: %w", err))
+	}
+	if hostedCluster == nil {
+		// ReadDesire absent, or the kube-applier has not observed the
+		// HostedCluster yet. We are re-enqueued when it writes status, and the
+		// stored value stays nil ("unknown") until then.
+		return nil
```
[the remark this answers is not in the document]
machi1990: This is unlikely to occur up only maybe during deletion. 
I am meaning toward not rewritting and leave the previously observed read desire. 
Perhaps what we can do here is addign logging @YuriyNik @Copilot

machi1990 (approved):
/lgtm

There are some comments from Copilot about stale docs, please review and update them.

Also ping @deads2k for a final review and approval as it looks good to me.

backend/pkg/kubeapplierhelpers/clusters.go
```diff
@@ -59,13 +67,71 @@ func GetCachedHostedClusterForCluster(
...
+// stops that requirement from silently retightening every other caller — most
+// notably operationClusterDelete, which reads nil as "the HostedCluster is
+// gone" and would otherwise finish a delete on a transient observation failure.
+//
+// Returns (nil, nil) when the ReadDesire is absent (NotFound), its Successful
+// condition is not true, or it carries no kubeContent.
+//
+// Returns a non-nil error only for hard failures: a non-NotFound lister error,
+// or unmarshal failure.
+func GetObservedHostedClusterForCluster(
```
machi1990: @YuriyNik I think the bot's comment made you add this.

In my opinion the pattern prior was okay using `kubeapplierhelpers.GetCachedHostedClusterForCluster` and leaving the content as last previously observed if hosted cluster is nil / an error occurred

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
```diff
@@ -0,0 +1,208 @@
...
+		// retract is reported alongside it rather than hiding the root cause.
+		return utils.TrackError(errors.Join(
+			fmt.Errorf("failed to get HostedCluster from ReadDesire: %w", err),
+			c.clearMirror(ctx, key, existing),
+		))
+	}
+	if hostedCluster == nil {
+		// ReadDesire absent, its last observation failed, or the kube-applier
+		// has not observed the HostedCluster yet.
+		return c.clearMirror(ctx, key, existing)
```
machi1990: We might be hitting this path quite a lot especially when the readdesire lister cache is cold e.g during startup.
In my opinion, let's not get fancy especially with @Copilot review and follow the same pattern as  https://github.com/YuriyNik/ARO-HCP/blob/3e95ed9ff5e45dc55bd5826f736736e54e4ba393/backend/pkg/controllers/cluster/version/control_plane_active_version_controller.go

i.e read from cache and parse using the existing utility (GetCachedHostedClusterForCluster).
if nil or error, return and leave previous observed.
if successfully parsed, check if there are diffs and update it if so. Otherwise don't update.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
```diff
@@ -0,0 +1,208 @@
...
+		"hostedClusterNamespace", existing.Status.ActualHostedCluster.Namespace,
+		"hostedClusterName", existing.Status.ActualHostedCluster.Name)
+	return nil
+}
+
+// sanitizeHostedCluster removes server-side metadata that changes independently
+// of the HostedCluster state admission consumes. Keeping it out of the mirror
+// prevents unrelated Kubernetes updates from causing ServiceProviderCluster
+// replacements and changefeed fan-out.
+func sanitizeHostedCluster(hostedCluster *hsv1beta1.HostedCluster) *hsv1beta1.HostedCluster {
```
machi1990: No need to do this. Store as is. I had left an initial comment about it, it was removed and now its back; not sure why - perhaps after @copilot review?

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
```diff
@@ -0,0 +1,208 @@
...
+		))
+	}
+	if hostedCluster == nil {
+		// ReadDesire absent, its last observation failed, or the kube-applier
+		// has not observed the HostedCluster yet.
+		return c.clearMirror(ctx, key, existing)
+	}
+
+	replacement := existing.DeepCopy()
+	replacement.Status.ActualHostedCluster = sanitizeHostedCluster(hostedCluster)
```
machi1990: No need to sanitize.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
YuriyNik: Agreed, and the cold-cache point is the decisive one — on startup this would have cleared every mirror before the ReadDesire informer synced, which is exactly the churn we don't want. Now follows the control_plane_active_version_controller.go pattern: read from cache, on nil or error return and leave the previously observed value, otherwise diff and only write when it changed.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
YuriyNik: Sorry about the round-trip — it came back via the Copilot review. sanitizeHostedCluster is removed for good; the HostedCluster is stored as observed, and MirrorsVerbatim is back to asserting that resourceVersion and managedFields survive.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
YuriyNik: Done — the call site is back to replacement.Status.ActualHostedCluster = hostedCluster.

backend/pkg/kubeapplierhelpers/clusters.go
YuriyNik: You're right, that was the bot's doing. Reverted — GetObservedHostedClusterForCluster and its test are gone, and clusters.go is now byte-identical to main again, so kubeapplierhelpers no longer appears in this PR at all.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	hostedCluster, err := kubeapplierhelpers.GetCachedHostedClusterForCluster(ctx, c.readDesireLister, key.SubscriptionID, key.ResourceGroupName, key.HCPClusterName)
+	if err != nil {
+		return utils.TrackError(fmt.Errorf("failed to get HostedCluster from ReadDesire: %w", err))
+	}
+	if hostedCluster == nil {
+		// ReadDesire absent, or the kube-applier has not observed the
+		// HostedCluster yet. We are re-enqueued when it writes status, and the
+		// stored value stays nil ("unknown") until then.
```
machi1990, line 113: it won't necessarily stay nil, it'll stay to the previously observed value

YuriyNik: can i clarify - are you talking about that 'hostedCluster == nil' is not reachable here ? i suppose this branch can happens.

YuriyNik: maybe leave this check but update  the comment

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
deads2k: This decision  means that when the HostedCluster is deleted, the ServiceProviderCluster will continue to have a previously set value.  Is that truly your intent?

deads2k (commented):
I doubt you really want this behavior, but if you do I'm willing to approve

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
YuriyNik: Discussed with @machi1990 in Slack and we agreed to keep it — but the reason is worth recording here rather than just deferring to the thread.
Retracting on nil looked right to me at first too, but nil at this point isn't only "the HostedCluster is gone". GetCachedHostedClusterForCluster is an in-memory GetByKey on the union ReadDesire indexer, GenericWatchingController.Run starts workers with no WaitForCacheSync, and an empty union reports itself synced (UnionDesireInformer.HasSynced: "An empty union is considered synced (vacuously true); callers that need a populated union should gate on their MC count" — this controller doesn't). So clearing on nil would wipe and then rewrite every mirror on each backend restart: 2×N Replaces, a changefeed wake per cluster, and a window where every 4→5 PATCH is rejected. The early return is load-bearing.
 The deletion case is covered without it: SyncOnce skips on DeletionTimestamp != nil, and clusterChildResourcesCleanupController deletes the whole ServiceProviderCluster document, so a stale mirror never outlives its cluster.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
YuriyNik: i suggest change comment to this 		// No observed HostedCluster right now. Leave the stored value as it is:
		// nil if we have never observed this cluster, otherwise the last value
		// we mirrored.
		//
		// Clearing it here would be wrong. Nil also covers "the union ReadDesire
		// informer has not filled in yet" — workers start before the
		// per-management-cluster sub-informers register, and an empty union
		// reports itself synced — so every backend restart would wipe and then
		// rewrite every mirror. A HostedCluster that is genuinely gone belongs to
		// a cluster that is itself being deleted, which we skipped above, and the
		// whole ServiceProviderCluster document is removed along with it.
		//
		// We are re-enqueued when the kube-applier writes status.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
machi1990: > The deletion case is covered without it: SyncOnce skips on DeletionTimestamp != nil, and clusterChildResourcesCleanupController deletes the whole ServiceProviderCluster document, so a stale mirror never outlives its cluster.

@YuriyNik On top of this, let's set the .ActualHostedCluster to nil during during deletion if it sees that the HC is nil.

YuriyNik: Done. It ended up folded into David's point — the controller now distinguishes "no ReadDesire / read hasn't succeeded" from "read succeeded and found nothing", and only the last one clears the mirror. That covers your case automatically: when the HC is gone during deletion, the read comes back successful-and-empty and .ActualHostedCluster is set to nil.

backend/pkg/kubeapplierhelpers/clusters.go
```diff
@@ -42,15 +44,59 @@ import (
...
-// GetCachedHostedClusterForCluster reads the HostedCluster mirror from the
-// per-cluster ReadDesire. The ReadDesire's Status.KubeContent.Raw carries
-// the observed HostedCluster JSON; we decode it directly and return the
-// typed object.
+// HostedClusterObservation distinguishes an unavailable ReadDesire from a
+// successful observation that found no HostedCluster.
+type HostedClusterObservation struct {
+	HostedCluster *v1beta1.HostedCluster
+	Observed      bool
+}
```
machi1990: @YuriyNik is this needed?

We could simply return `HostedCluster, bool, error` ? What are your thoughts?

backend/pkg/kubeapplierhelpers/clusters.go
```diff
@@ -42,15 +44,59 @@ import (
...
+
+// GetHostedClusterObservationForCluster reads the HostedCluster mirror from
+// the per-cluster ReadDesire. Observed is false when the ReadDesire is absent,
+// has not completed successfully, or contains no current observation. An
+// observed ReadDesire with empty KubeContent has Observed=true and a nil
+// HostedCluster: the management cluster was successfully checked and the
+// HostedCluster is absent.
+func GetHostedClusterObservationForCluster(
```
machi1990: This also gets it from the cache.

So its more like:

GetCachedHostedClusterObservationForCluster.

But in general, I am wondering if we couldn't simply re-use `GetCachedHostedClusterForCluster` for what we want to do in the `ActualHostedCluster` controller over adding another abstraction;

In that controller
```
hc, err = GetCachedHostedClusterForCluster()
if err return err
if hc != nil || (hc == nil and cluster is deleting) 
- we set scpc.status.actualhosted = hc
- update the spc 

return
```

@YuriyNik what are your thoughts?

machi1990: > But in general, I am wondering if we couldn't simply re-use GetCachedHostedClusterForCluster for what we want to do in the ActualHostedCluster controller over adding another abstraction;


Hmm what I said doesn't work best when the cache is cold; it'll will prematurely unclear the hosted cluster while in actual sense it might still be existing and in process of being deleted. Ignore the above quoted part of the comment and change the method to `GetCachedHostedClusterObservationForCluster` only and have `GetCachedHostedClusterForCluster` re-use it.

backend/pkg/kubeapplierhelpers/clusters.go
YuriyNik: Good point. The struct is unnecessary for this two-state result. I’ll simplify it to return (*v1beta1.HostedCluster, bool, error), where the boolean indicates whether the ReadDesire observation completed successfully. nil, true will represent a successful empty observation and will clear the mirror; nil, false will retain the existing mirror for unavailable/failed observations.

backend/pkg/kubeapplierhelpers/clusters.go
YuriyNik: ok, agree. i will  rename it to GetCachedHostedClusterObservationForCluster and GetCachedHostedClusterForCluster reuses it.

machi1990 (approved):
/lgtm

backend/pkg/kubeapplierhelpers/clusters.go
```diff
@@ -42,38 +44,53 @@ import (
...
+// Returns (nil, nil) when the ReadDesire has not been created yet (NotFound) or
+// carries no kubeContent. It deliberately does not inspect the "Successful"
+// condition, so a ReadDesire whose latest read failed still yields the content
+// from the last read that worked. Callers that must tell "we have not looked"
+// apart from "we looked and found nothing" want
+// GetCachedHostedClusterObservationForCluster instead.
+//
+// Returns a non-nil error only for hard failures: a non-NotFound lister
+// error, or unmarshal failure.
+func GetCachedHostedClusterForCluster(
```
deads2k: No. don't make two methods.  Just add the `present bool` to the return of the original method and update all callers.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
```diff
@@ -0,0 +1,172 @@
...
+		// Clearing here would be wrong. "Not observed" also covers a cold union
+		// ReadDesire informer — workers start before the per-management-cluster
+		// sub-informers register, and an empty union reports itself synced — so
+		// clearing would wipe and then rewrite every mirror on each backend
+		// restart.
+		//
+		// We are re-enqueued when the kube-applier writes status.
+		return nil
+
+	case hostedCluster != nil && isDeleting:
```
deads2k: this case is wrong. When the cluster isn't nil and it isn't deleting, we *still* want the data to be accurate.

We always want the data accurate, you shouldn't be creating cases that make the data inaccurate.

backend/pkg/kubeapplierhelpers/clusters.go
YuriyNik: Added the bool to the original method, removed the wrapper, and updated all callers.

backend/pkg/controllers/cluster/hostedcluster/actual_hosted_cluster_controller.go
YuriyNik: Removed the deletion exception; successful observations now update or clear the mirror during deletion too.

PR #6912, opened 2026-09-11
feat: support RHCOS marketplace image override for nodepools via experimental tag

bennerv:
## Why

Testing preview/unreleased RHCOS builds on ARO-HCP node pools currently requires a custom Cluster Service deployment. This adds a supported path for pointing a node pool at a specific Azure Marketplace image, so preview marketplace offers can be validated end-to-end before they ship.

The override is deliberately narrow: it is gated behind the `ExperimentalReleaseFeatures` AFEC, exposed only via an experimental ARM tag rather than the public API surface, and immutable after node pool creation.

No tracking ticket exists for this yet (CONTRIBUTING.md rule 4) — this is exploratory enablement work. I'll link one before moving out of draft.

## What

- `chore:` bumps ocm-sdk-go to v0.1.510 / ocm-api-model to v0.0.466, picking up upstream `AzureNodePoolImage` and replacing the previously-needed SDK forks.
- `feat:` adds the `aro-hcp.experimental.nodepool.marketplace-image` tag taking a `publisher:offer:sku:version` URN, a `MarketplaceImage` field on node pool service-provider properties, an admission mutator that validates the URN on CREATE only, and plumbing to Cluster Service in `BuildCSNodePool`.
- `test(e2e):` adds a marketplace-image node pool test against a 5.0 cluster, plus node verifiers.
- Docs at `docs/testing-preview-marketplace-offer.md`.

## :warning: Draft — contains a DROP commit

The tip commit pins Cluster Service to a personal ACR build (`arohcpsvcdev.azurecr.io/bvesel-test/aro-hcp-clusters-service`) so dev environments get a CS that understands `AzureNodePoolImage`. **It must not merge.** Upstream CS has since moved to `quay.io/app-sre/aro-hcp-clusters-service`; once that image carries the support, this commit gets dropped.

`internal/ocm/convert.go` imports `ocm-api-model/clientapi/arohcp/v1alpha1` directly for `AzureNodePoolImage`, since ocm-sdk-go hasn't generated an alias yet. Commented inline; remove once the SDK re-exports it.

## Testing

Verified locally: `internal`, `backend`, `frontend` build clean, and `go vet -tags E2Etests` passes on `test/e2e` and `test/util`. Full `make verify` / `make test` not run locally — relying on CI.


bennerv:
TODO: 
- we can remove some validations from CS and put them here in a controller for validating the gen2 capability, the image actually existing in ARM (unless it's preview) and the instance type.

admin/server/go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6 v6.2.0
 	github.com/go-logr/logr v1.4.3
 	github.com/microsoft/go-otel-audit v0.2.2
-	github.com/openshift-online/ocm-sdk-go v0.1.509
+	github.com/openshift-online/ocm-sdk-go v0.1.510
```
JameelB: does this need updating to [v0.1.511](https://github.com/openshift-online/ocm-sdk-go/pull/1203)?

internal/ocm/convert.go
```diff
@@ -26,6 +26,10 @@ import (
 
 	azcorearm "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 
+	// csarhcpv1alpha1 is the model package that arohcpv1alpha1 aliases. It is imported
+	// directly only for AzureNodePoolImage, for which ocm-sdk-go has not yet generated
+	// an alias; drop this import once the SDK re-exports it.
+	csarhcpv1alpha1 "github.com/openshift-online/ocm-api-model/clientapi/arohcp/v1alpha1"
```
bennerv: TODO: bump

PR #6920, opened 2026-09-11
fix(validation): validate that activeKey.version is not a URL string

cadenmarchese:
### What

- Validate that activeKey.version is a 32 char hex before proceeding

### Why

- Today, you can pass a full Key URL here and the install will fail late. We need to be able to fail early if full Key URL is passed to a field that doesn't support it. 32-key hex is the correct format for a key version (see https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/resource-manager/Microsoft.KeyVault/KeyVault/stable/2026-02-01/openapi.json#L2141-L2146)
- This mitigates incidents where users are passing full Key URL before it's supported by our APIs. This fix is separate from the effort to support full Key URL in our APIs.

### Testing

justification.

- Unit tests

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:


internal/validation/validators_test.go
```diff
@@ -451,3 +451,93 @@ func TestValidateNodePoolVersionChange(t *testing.T) {
 		})
 	}
 }
+
+func TestAzureKeyVaultKeyVersion(t *testing.T) {
+	ctx := context.Background()
+	op := operation.Operation{Type: operation.Create}
+	fldPath := field.NewPath("version")
+
+	tests := []struct {
```
rajdeepc2792, line 460: Is there any other special character, that isn't allowed? Can't we fit in the string check in a strict regex?

cadenmarchese: Not explicitly. I'm hoping to keep the validation pretty loose and only validate that it isn't a URL, since we can't guarantee that the key version string created by Azure won't change in the future. We can at least assert that it shouldn't be the full URL right now, for this specific field.

hlipsig (approved):
LGTM

internal/validation/validators.go
```diff
@@ -398,6 +398,19 @@ func ValidateUUID(_ context.Context, _ operation.Operation, fldPath *field.Path,
 	return nil
 }
 
+func AzureKeyVaultKeyVersion(_ context.Context, _ operation.Operation, fldPath *field.Path, value, _ *string) field.ErrorList {
```
bennerv, line 406: is there not a regex we can use?

bennerv: > object-version 	An object-version is a system-generated, 32 character string identifier that is optionally used to address a unique version of an object.

https://learn.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates

internal/validation/validate_cluster.go
```diff
@@ -1062,6 +1062,7 @@ func validateKmsKey(ctx context.Context, op operation.Operation, fldPath *field.
 	}
 	errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("version"), &newObj.Version, nil)...)
```
bennerv, line 1063: Does this make it a required value?  If so, why are we returning nil in the `AzureKeyVaultKeyVersion` function if `len() == 0`?

bennerv (commented):
It's possible during an update operation if the value doesn't change we just return early.  This would protect us against values that previously were valid but no longer are.  i don't anticipate we ever get into that situation, so it'd be more a nit that can be ignored unless someone feels very strongly about it. 

The cases i think it would be useful is if we have backend writers to cosmosdb that go through the validation but don't mutate this field, and the field is already broken, we don't want the backend controller to fail because the customer provided bad input at some point in time it was invalid and the backend can't fix it becuase it doesn't have the proper information.  

:shrug:

internal/validation/validators.go
cadenmarchese: That works. The key object's swagger has it too: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/resource-manager/Microsoft.KeyVault/KeyVault/stable/2026-02-01/openapi.json#L2141-L2146

I'll make the change.

internal/validation/validators.go
```diff
@@ -370,6 +370,11 @@ var (
 	diskEncryptionSetName            = `^[a-zA-Z0-9_-]+$`
 	diskEncryptionSetNameRegex       = regexp.MustCompile(diskEncryptionSetName)
 	diskEncryptionSetNameErrorString = `(must contain only alphanumeric characters, underscores, and hyphens)`
+
+	// keyVaultKeyVersion See https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/resource-manager/Microsoft.KeyVault/KeyVault/stable/2026-02-01/openapi.json#L2141-L2146
+	keyVaultKeyVersion            = `^[0-9a-fA-F]{32}$`
```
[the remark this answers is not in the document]
cadenmarchese: I fixed the PR description.

internal/validation/validate_cluster.go
```diff
@@ -1061,7 +1061,7 @@ func validateKmsKey(ctx context.Context, op operation.Operation, fldPath *field.
 		errs = append(errs, immutableByCompare(ctx, op, fldPath.Child("version"), &newObj.Version, safe.Field(oldObj, toKmsKeyVersion))...)
 	}
 	errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("version"), &newObj.Version, nil)...)
-	errs = append(errs, MaxLen(ctx, op, fldPath.Child("version"), &newObj.Version, nil, 255)...)
+	errs = append(errs, AzureKeyVaultKeyVersion(ctx, op, fldPath.Child("version"), &newObj.Version, nil)...)
```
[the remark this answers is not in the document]
cadenmarchese: TypeSpec / API changes are immutable once the API is released.

test-integration/frontend/artifacts/FrontendCRUD/Cluster/read-old-data/01-load-old-data/create-with-tags.json
```diff
@@ -34,7 +34,7 @@
 							"activeKey": {
 								"name": "encryptionKeyName",
 								"vaultName": "keyVaultName",
-								"version": "2024-12-01-preview"
+								"version": "4e832b5c8f1e4e3d9c6b2a1f3e7d9c5b"
```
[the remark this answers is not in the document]
cadenmarchese: Existing clusters will have 32-char hex per the KeyVault Key API spec. Retroactively introducing this validation on existing clusters should be safe. I'll defer to human review on if this is worth adjusting.

cadenmarchese:
e2e passed before the rebase:
Screenshot 2026-09-23 at 9 27 58 AM

PR #6921, opened 2026-09-11
[wip] Persist HostedCluster identity before cluster creation

deads2k:
I need to review first.

### DIFF
--- a/backend/pkg/controllers/cluster/creation/hosted_cluster_identity_controller.go
+++ b/backend/pkg/controllers/cluster/creation/hosted_cluster_identity_controller.go
@@ -1,0 +1,185 @@
+// Copyright 2026 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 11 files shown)


backend/pkg/controllers/cluster/creation/hosted_cluster_identity_controller.go
```diff
@@ -0,0 +1,185 @@
...
+			return "", err
+		}
+		if unique {
+			return candidate, nil
+		}
+	}
+	return "", fmt.Errorf("unable to generate a unique HostedCluster name for cluster %s", clusterName)
+}
+
+func (c *hostedClusterIdentitySyncer) isHostedClusterNameUnique(ctx context.Context, candidate, subscriptionID, resourceGroupName string) (bool, error) {
```
machi1990, line 141: Let's try to achieve this without reaching to CS.

deads2k: lol.  I did note in teh PR description that I hadn't reviewed yet.  I told it to copy cluster-service code.  Is there any reason to avoid just generating a UID for the namespace and the name and calling it a day?  Maybe with a prefix of some kind?

PR #6923, opened 2026-09-11
fix: lowercase name in node pool scoped readdesire

placecol:
https://redhat.atlassian.net/browse/ARO-29572

### What

CS lowercases the ARM node pool name for both its internal node pool ID and the AzureNodePool.ResourceName it derives the Hypershift NodePool object's name from (see internal/ocm/convert.go's BuildCSNodePool). The target name used in the node pool scoped readdesire did not do the same, which resulted in no status returned for node pools with uppercase characters in their name.

### Why

Without this fix node pool update operations will perpetually be in an `Updating` state until they timeout.

### Testing

Unit test was added for the readdesire controller.

E2e test `nodepool_update_nodes.go` was updated to use upper case in the node pool name:
```
const (
    customerClusterName  = "np-update-nodes-hcp-cluster"
    customerNodePoolName = "np-update-NODES"
    oneNodePoolName      = "np-one-node"
)
```

Without the fix the test fails due to the update operation timing out, backend logs show the readdesire never observed the node pool:
```
{
...
	"msg": "picked node pool update operation status",
...
	"resource_name": "np-update-nodes",
...
	"message": "[hypershiftNodePool] Hypershift NodePool has not been observed yet",
	"operation_id": "df6c683b-f079-46ba-9586-e3040d2473d9",
	"provisioningState": "Updating"
}
```

With the fix the test succeeds, backend logs show the appropriate `Updating` hypershift status:
```
{
...
	"msg": "picked node pool update operation status",
...
	"provisioningState": "Updating",
	"message": "[hypershiftNodePool] hypershift NodePool status replicas is 2, want 3"
}
```

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:

### DIFF
--- a/backend/pkg/controllers/nodepool/readdesires/create_nodepool_scoped_read_desires_controller.go
+++ b/backend/pkg/controllers/nodepool/readdesires/create_nodepool_scoped_read_desires_controller.go
@@ -142,7 +142,10 @@ func (c *createNodePoolScopedReadDesiresSyncer) SyncOnce(ctx context.Context, ke
 	}
 	csClusterID := existingNodePool.ServiceProviderProperties.ClusterServiceID.ClusterID()
 
-	target := nodePoolTarget(c.hostedClusterNamespaceEnvIdentifier, csClusterID, csClusterDomainPrefix, existingNodePool.ID.Name)
+	// CS lowercases the ARM node pool name for both its internal node pool ID and the
+	// AzureNodePool.ResourceName it derives the Hypershift NodePool object's name from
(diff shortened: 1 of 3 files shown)


raelga (approved):
Fix makes sense, CS lowercases the node pool name when deriving the Hypershift NodePool object name, but the readdesire target wasn't doing the same, so uppercase-named node pools never got observed and updates would hang. One line fix, good test coverage for both cases, CI green.

/lgtm
/approve

stevekuznetsov (approved):
/lgtm
/approve

PR #6931, opened 2026-09-13
fix(dev-infrastructure): unblock mgmt cluster creation from svc cluster

geoberle:
### What

Moves the per-management-cluster kube-applier CosmosDB container + role assignments (kube-applier MI and Clusters Service MI) out of `mgmt-cluster.bicep` into a new dedicated `kube-applier-cosmos` step that runs after the `cluster` step instead of gating it.

### Why

#6709 threaded `csManagedIdentityPrincipalId` into the mgmt `cluster` ARM step, sourced from the `service` resource group's `output` step. That `output` step `externalDependsOn` `Service.Infra`'s `service/cluster` step (the service AKS cluster build, ~10 min). As a result the mgmt cluster ARM deployment serialized behind the full service cluster build instead of running in parallel with it, adding ~15-17 minutes to rollout time.

The new `kube-applier-cosmos` step depends on `management/cluster` and pulls `csManagedIdentityPrincipalId` from `service/output` the same way `svc-mgmt-aks-permissions` already does, so it adds no new stall — it just moves the wait to after `cluster`, not before it.

### Testing

- `bicep build` on all touched/new templates
- `make validate-config-pipelines` (full templatize pipeline validation)


raelga (approved):
/lgtm
/approve

PR #6944, opened 2026-09-14
feat(backend): add controller that fetches and persists identity metadata from different sources

miguelsorianod:
Introduce a general identity-metadata controller that resolves ClientID, PrincipalID, and TenantID for every cluster managed identity in CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities (control-plane operators, data-plane operators, and the service managed identity) from the sources that apply to that identity.
Results are written onto ServiceProviderCluster.Status.ManagedIdentityDetails as last-pass-only values, keyed by lowercased resource ID:
- ARM User Assigned Identities API for control-plane and data-plane operators. SMI-only identities are skipped because end-users are not asked to grant the SMI read on itself.
- Managed Identities Data Plane when that service is available, for dataplane-registered identities (MSI-based identities)
- Hardcoded identity in environments without the real dataplane

Retrieval failures persist RetrievalError and clear previously resolved IDs so readers never treat stale values as current. Recheck timing is honored only while the stored identity set still matches CustomerProperties.

The motivation of this design is that we persist all the information that we can from the sources where it applies and we can retrieve from. This is opposed as the approach followed until now where we were attempting to make this transparent to the controllers, trying to hide the information behind a fake MIDataplane client. This has the implication that the controllers will need to be aware from which one of the sources they need to obtain the metadata from, which also means they will become aware on whether the MIDataplane service is available or not, and have branching code that implements the behavior in the different scenarios. This decision is intentional, and the reason is that trying to hide/make transparent the information makes the code harder to understand, debug and extend so we are prioritizing being explicit with the tradeoff of the code needing to become aware of some of the details explained above.

Additionally, wire HardcodedIdentity through BackendOptions so the controller can choose dataplane vs hardcoded without passing the full MI dataplane client builder into callers that do not need it.

### Testing

justification.

- Unit tests

### PR Checklist

- [ ] Self-reviewed the diff

If E2E tests are included:


backend/pkg/controllers/cluster/identity/fetch_managed_identities_info.go
```diff
@@ -0,0 +1,679 @@
...
+	if err != nil {
+		return nil, []error{utils.TrackError(fmt.Errorf("failed to get User Assigned Identities Client: %w", err))}
+	}
+
+	metadata := make(map[string]*coreapi.IdentityMetadataValue)
+	var errs []error
+	for _, identity := range armIdentities {
+		resourceIDKey := strings.ToLower(identity.resourceID.String())
+		// If the Service Managed Identity does not have permissions to do a GET over the identity resource ID it is querying, this call
+		// fails and the controller accumulates the error.
```
miguelsorianod, line 605: The code currently intentionally lets the controller error if the SMI doesn't exist or doesn't have permissions. The alternative would be to detect whether it exists but that would require an additional MIDataplane call. For now I decided to let it error

backend/pkg/controllers/cluster/identity/fetch_managed_identities_info.go
```diff
@@ -0,0 +1,679 @@
...
+	"github.com/Azure/ARO-HCP/internal/database/informers/coreinformers"
+	"github.com/Azure/ARO-HCP/internal/database/listers/corelisters"
+	"github.com/Azure/ARO-HCP/internal/utils"
+)
+
+const (
+	// FetchManagedIdentitiesInfoControllerName is the single source of truth
+	// for this controller's name. It is used for the workqueue name (a Prometheus
+	// label), context/logger controller name, and log fields.
+	FetchManagedIdentitiesInfoControllerName = "FetchManagedIdentitiesInfo"
```
miguelsorianod, line 47: This controller is persisting a subset of information that is already being persisted by the FetchMSIIdentitiesInfo controller and by the FetchDataPlaneOperatorsManagedIdentitiesInfo, in different cosmos attributes.
The idea is to let both run, and eventually move the existing controllers and code that rely on `MSIManagedIdentities` and `DataPlaneOperatorsManagedIdentities` to use `ManagedIdentityDetails` instead. Preferably when we work on updating those controllers to support identities replacement.

The tradeoff is that it will put a bit more load on the APIs of ARM UserAssignedIdentities and MIDataplane service until we remove the old ones, as we will have both sets running.

backend/pkg/controllers/cluster/identity/fetch_managed_identities_info.go
```diff
@@ -0,0 +1,679 @@
...
+
+	// managedIdentitiesRecheckInterval is the base interval before re-querying
+	// identity metadata sources when ClientID/PrincipalID/TenantID are already
+	// resolved. Combined with managedIdentitiesRecheckJitter via wait.Jitter.
+	managedIdentitiesRecheckInterval = 1 * time.Hour
```
miguelsorianod, line 52: I set it to 1 hour. Unsure what value we want

backend/pkg/controllers/cluster/identity/fetch_managed_identities_info.go
miguelsorianod, line 47: From this PR alone it might be a bit hard to get a good idea on how will this be used. For that, you can take a look at a PoC PR that showcases implementing data plane identities oidc federation with support for identities replacement https://github.com/Azure/ARO-HCP/pull/6893, as well as another PoC with denyassignment https://github.com/Azure/ARO-HCP/pull/6899 that involves smi-based identities.
Those can give a better idea on how the new ManagedIdentityDetails is used, how are the controllers that rely on it now need to extract from the appropriate metadata sources and need to become aware on whether the MI Dataplane service is available or not, have to have different code branching with different implementations depending on that and so on

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -599,3 +613,93 @@ const (
...
+type ManagedIdentityMetadata struct {
+	// ResourceID is the Azure Resource ID of the managed identity.
+	// Written by: FetchManagedIdentitiesInfo
+	ResourceID *azcorearm.ResourceID `json:"resourceId,omitempty"`
+
+	// MetadataFromARMUserAssignedIdentitiesAPI is the metadata for the identity retrieved from
+	// the ARM User Assigned Identities API (https://learn.microsoft.com/en-us/rest/api/managedidentity/user-assigned-identities)
+	// as the source. Nil when ARM does not apply: ARM is queried for control-plane and data-plane operator
+	// identities. It is not queried for an identity used only as the service managed identity, because
+	// end-users are not asked to grant the Service Managed Identity read on itself. If the same resource ID is also a
```
miguelsorianod, line 630: > It is not queried for an identity used only as the service managed identity, because end-users are not asked to grant the Service Managed Identity read on itself.

Unfortunately we have to do it this way because we are currently not asking end-users to grant the Service Managed Identity permissions to do a Get on itself (https://learn.microsoft.com/en-us/rest/api/managedidentity/user-assigned-identities/get?view=rest-managedidentity-2024-11-30&tabs=HTTP).
I wonder if we would like to start requiring that. The issue however is that that would be a breaking change because pre-existing clusters/identities don't have that, so I don't know if we have much margin here.

If we were requesting the end-users to give permissions to the SMI to do a GET UserAssignedIdentities on itself technically we wouldn't need to query the metadata (clientid, principalid, tenantid) from the MIDataplane service at all. However, we should also be aware that maybe we still would like to have that information retrieved from both metadata sources because if the end-users do not configure the SMI permissions correctly it wouldn't be able to fetch it. It depends on what do we want to store. As of now the information of some identities is being fetched and stored in both MetadataFromARMUserAssignedIdentitiesAPI and MetadataFromManagedIdentitiesDataplaneService, and the information might not be 1:1 between them as they are fetched from different services non atomically. The controllers also need to decide from which one they prefer to fetch in the cases where they have the information available in both

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -599,3 +613,93 @@ const (
...
+	RetrievalError *string `json:"retrievalError,omitempty"`
+}
+
+// TODO here or as a function outside of this type/package?
+// HasResolvedIdentityInformation reports whether value has non-empty ClientID,
+// PrincipalID, and TenantID. A non-nil RetrievalError is unresolved. An empty
+// value (all fields nil) is also unresolved: that source applies but has not
+// been resolved yet or could not be resolved in this pass. The receiver must
+// be non-nil.
+func (v *IdentityMetadataValue) HasResolvedIdentityInformation() bool {
```
miguelsorianod, line 697: This will be used by other controllers. In general we don't put logic in the internal/api/coreapi types but this seemed recurrent enough and simple enough to be worth it having associated to the type

backend/cmd/root.go
```diff
@@ -491,14 +492,18 @@ func (f *BackendRootCmdFlags) ToBackendOptions(ctx context.Context, cmd *cobra.C
 			azureclient.CheckAccessV2InsecureARMPermissionsManagerRateLimiterBurst,
 		)
 
-		// In ARO-HCP environments where we don't have a real FPA, we use the HardcodedIdentityFPAMIDataplaneClientBuilder to create the FPA MI dataplane client builder
-		fpaMIDataplaneClientBuilder, err = newHardcodedIdentityFPAMIDataplaneClientBuilder(
+		// In ARO-HCP environments where we don't have a real FPA, we load the hardcoded identity
+		// and use it both as the mock MI dataplane source and as the identity metadata source.
+		hardcodedIdentity, err = newHardcodedIdentity(
```
miguelsorianod, line 497: We intentionally separate now the construction of the hardcoded identity itself and pass it to the hardcoded fpa midataplaneclientbuilder. This is because some controllers will only need the hardcodedidentity without the midataplane client.

This means that at some point we will only use the mock MIDataplane client when needing to retrieve short-lived credentials, or not even that, just having it in the hardcoded type and have branching code that uses the real midataplane client when the service is available or the hardcodedidentity when the real service is not available

backend/pkg/controllers/cluster/identity/fetch_managed_identities_info.go
```diff
@@ -0,0 +1,679 @@
...
+			if c.hardcodedIdentity != nil {
+				replacementManagedIdentityDetails[resourceIDKey].MetadataFromHardcodedIdentity = &coreapi.IdentityMetadataValue{}
+			} else {
+				replacementManagedIdentityDetails[resourceIDKey].MetadataFromManagedIdentitiesDataplaneService = &coreapi.IdentityMetadataValue{}
+			}
+		}
+	}
+
+	var errs []error
+	if c.hardcodedIdentity != nil {
```
miguelsorianod, line 297: This shows how now the controllers become aware on whether the real midataplane service is available or not. In this one, having hardcodedIdentity non nil signals that the real midataplane service is not available, so in that case we use the hardcodedidentity and we do not execute the path that relies on the midataplane client / service

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -599,3 +613,93 @@ const (
...
+	// the ARM User Assigned Identities API (https://learn.microsoft.com/en-us/rest/api/managedidentity/user-assigned-identities)
+	// as the source. Nil when ARM does not apply: ARM is queried for control-plane and data-plane operator
+	// identities. It is not queried for an identity used only as the service managed identity, because
+	// end-users are not asked to grant the Service Managed Identity read on itself. If the same resource ID is also a
+	// control-plane or data-plane operator identity, ARM still applies.
+	// When this metadata source applies, a failure to retrieve this metadata using the API results in the
+	// ClientID, PrincipalID, and TenantID attributes being set to nil and RetrievalError records what went wrong. This includes the case
+	// where the identity does not exist in Azure.
+	// Written by: FetchManagedIdentitiesInfo
+	MetadataFromARMUserAssignedIdentitiesAPI *IdentityMetadataValue `json:"metadataFromARMUserAssignedIdentitiesAPI,omitempty"`
```
miguelsorianod, line 636: All metadatasources have the type `*IdentityMetadataValue`. The RetrievalError attribute is not leveraged in some sources (for example the hardcoded identity one), but I use the same `IdentityMetadataValue` in all of them for simplicity/consistence and to keep a single type for them. If we want to be more strict then we would need to define different types for them, and it would also impact code that reads them as the methods/functions would need to deal with different types.

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -599,3 +613,93 @@ const (
...
+// retrieval for this source. They are not sticky. A later pass that fails,
+// including an intermittent API error, persists them as nil and may set
+// RetrievalError. A later successful pass can populate them again. Readers
+// must not assume that once-resolved values remain set.
+type IdentityMetadataValue struct {
+	// ClientID is the Client ID of the managed identity as returned by the
+	// source on the most recent retrieval. It may be nil even after a previous
+	// successful retrieval. See IdentityMetadataValue.
+	// Written by: FetchManagedIdentitiesInfo
+	ClientID *string `json:"clientId,omitempty"`
```
miguelsorianod, line 669: Notice how ClientID, PrincipalID, TenantID are intentionally pointers. They can become nil when there are transient errors when fetching from the APIs, or when the underlying identity does not exist (for example the MIDataplane service API currently returns them as nil in that case).

As another note, in the case of the hardcodedidentity metadata source they should never be nil (unless there's a bug in our code) when the metadatasource is set but similar to https://github.com/Azure/ARO-HCP/pull/6944/changes#r4008406313 we reuse the same type for consistency/simplicity.

backend/pkg/controllers/cluster/identity/fetch_managed_identities_info.go
```diff
@@ -0,0 +1,678 @@
...
+	}
+
+	fpaMIDataplaneClient, err := c.fpaMIdataplaneClientBuilder.ManagedIdentitiesDataplane(cluster.ServiceProviderProperties.ManagedIdentitiesDataPlaneIdentityURL)
+	if err != nil {
+		return nil, []error{utils.TrackError(fmt.Errorf("failed to get Managed Identities Data Plane Client: %w", err))}
+	}
+
+	fpaMIDataplaneCredentials, err := fpaMIDataplaneClient.GetUserAssignedIdentitiesCredentials(ctx, dataplane.UserAssignedIdentitiesRequest{IdentityIDs: identityResourceIDStrs})
+	if err != nil {
+		return nil, []error{utils.TrackError(fmt.Errorf("failed to get Managed Identities Data Plane Credentials: %w", err))}
```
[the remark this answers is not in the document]
miguelsorianod: I am not sure this is fully accurate. If that fails even though we don't persist retrievalerror we are unsetting them

backend/pkg/controllers/cluster/identity/fetch_managed_identities_info.go
```diff
@@ -0,0 +1,678 @@
...
+	smiResourceID := cluster.CustomerProperties.Platform.OperatorsAuthentication.UserAssignedIdentities.ServiceManagedIdentity
+	if smiResourceID == nil {
+		return nil, []error{utils.TrackError(fmt.Errorf("cluster ServiceManagedIdentity is nil. Cannot resolve ARM User Assigned Identities metadata"))}
+	}
+
+	// If the Service Managed Identity does not exist, this call fails and the controller will accumulate the error.
+	userAssignedIdentitiesClient, err := c.smiClientBuilder.UserAssignedIdentitiesClient(ctx, cluster.ServiceProviderProperties.ManagedIdentitiesDataPlaneIdentityURL, smiResourceID, cluster.ID.SubscriptionID)
+	if err != nil {
+		return nil, []error{utils.TrackError(fmt.Errorf("failed to get User Assigned Identities Client: %w", err))}
+	}
```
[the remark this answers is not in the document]
miguelsorianod: Similar comment to https://github.com/Azure/ARO-HCP/pull/6944#discussion_r4008469825

internal/api/coreapi/types_serviceprovider_cluster.go
```diff
@@ -599,3 +613,93 @@ const (
...
+// been resolved yet or could not be resolved in this pass. The receiver must
+// be non-nil.
+func (v *IdentityMetadataValue) HasResolvedIdentityInformation() bool {
+	if v.RetrievalError != nil {
+		return false
+	}
+
+	return len(ptr.Deref(v.ClientID, "")) > 0 &&
+		len(ptr.Deref(v.PrincipalID, "")) > 0 &&
+		len(ptr.Deref(v.TenantID, "")) > 0
```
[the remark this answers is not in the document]
miguelsorianod: This will be covered further when adding the controllers that rely on this information

PR #6952, opened 2026-09-15
chore: unpin eastus2 e2e from test tenant

roivaz:
## Why

Azure has approved compatible Dsv5 capacity for the Red Hat PROD E2E subscriptions in East US 2. Keeping the temporary Test-Tenant override would unnecessarily prevent the slot manager from leasing the normal `Prod - 00` and `Prod - 01` pools for that region.

## What changed

- Remove the `eastus2` `allowedSubscriptions` override so it inherits the normal PROD subscription pool selection.
- Keep `eastus2euap` pinned to `ARO HCP E2E`, since that remains the only subscription onboarded for the canary region.

## Validation

- `make -C config materialize`
- `make -C config validate`

Tracking: [ARO-29058](https://issues.redhat.com/browse/ARO-29058)

### DIFF
--- a/config/config.msft.clouds-overlay.yaml
+++ b/config/config.msft.clouds-overlay.yaml
@@ -150,25 +150,11 @@ clouds:
             audit:
               connectSocket: true
         regions:
-          eastus2:
-            e2e:
-              regionTest:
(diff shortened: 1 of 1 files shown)


raelga (approved):
/lgtm
/approve

PR #6956, opened 2026-09-15
feat: expose Copilot retry-after metadata

mmazur:
https://redhat.atlassian.net/browse/AROSLSRE-2011

Follow-up to #6862.

Codex session: `01a0a54d-32f7-7543-873c-e262e36ba719`

### What

- Add optional `RetryAfterSeconds` metadata to `agent.CopilotSessionError`.
- Correlate the Copilot SDK follow-up `auto_mode_switch.requested` event with the active serialized send.
- Bound the follow-up wait so a missing runtime event cannot indefinitely delay returning the original session error.
- Preserve `nil` when Copilot does not provide a retry/reset duration.

### Why

Copilot reports rate-limit classification in `SessionErrorData`, but reports the optional reset duration separately on the immediately following auto-mode-switch event. Consumers can currently identify a 429 but cannot schedule a retry using the provider-supplied duration.

Returning the duration on the existing typed error lets callers use `errors.As` and inspect `RetryAfterSeconds` without parsing messages or changing the `LLMSession` interface.

### Testing

- `cd tooling/hcpctl && go test ./... -count=1`
- `cd tooling/hcpctl && go test -race ./pkg/agent/... -count=1`


mmazur:
Closing. We could use a generic RetryAfter indicator, but this isn't that, since it's tied to EligibleForAutoSwitch flag.

PR #6577, opened 2026-08-17
[wip] Progressive version rollout 01 initial try

deads2k:
### Testing

Testing is required for feature completion and tests should be part of the pull
request along with the feature changes.

Describe the testing provided. If you did not add tests, provide a clear
justification.

- Unit tests
- [Integration tests](https://github.com/Azure/ARO-HCP/tree/main/test-integration)
- [E2E tests](https://github.com/Azure/ARO-HCP/blob/main/test/e2e/README.md)

### PR Checklist

- [ ] PR is scoped to a single task (no mixed concerns)
- [ ] Title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [ ] Summary explains the "Why" behind the change
- [ ] Linked to relevant ticket/issue
- [ ] Screenshots included (if graph/UI/metrics changes)
- [ ] Self-reviewed the diff
- [ ] CI/CD checks are passing (ignore Tide)
- [ ] Draft PR used for WIP (if applicable)
- [ ] Commit history is clean (rebased/squashed)
- [ ] Tricky code blocks are commented
- [ ] Specific reviewers tagged
- [ ] All comment threads resolved before merge

If E2E tests are included:

- [ ] E2E tests follow [Principles of Good E2E Test Case Design](https://github.com/Azure/ARO-HCP/blob/main/test/AGENTS.md)
- [ ] If new E2E use case is covered (via a new test or new check/verifier),
  demonstrate that the test is able to detect a defect/error and fail with
  proper error message and logs which communicates nature of the problem.

PR #6968, opened 2026-09-15
fix(validation): pin exact OpenShift versions through the exact-version tag only

tony-schndr:
https://redhat.atlassian.net/browse/ARO-29608

### What

Makes the `control-plane-exact-version` ARM resource tag the **only** way to pin an exact OpenShift build. `version.id` now always carries the bare `MAJOR.MINOR` release line, and a patch-bearing `version.id` is rejected.

Version pins are only allowed through:

```yaml
version:
  id: 5.0
  channel: nightly
tags:
  "aro-hcp.experimental.cluster.control-plane-exact-version": "5.0.0-0.nightly-2026-09-03-214615"
```

Instead of how it was done previously:

```yaml
version:
  id: 5.0.0-0.nightly-2026-09-03-214615
  channel: nightly
```

Concretely:
- Admission no longer derives a pin from `version.id`, and no longer rewrites `version.id` at all. The exact-version tag is translated in the same pass as every other experimental tag, into `ServiceProviderProperties.ExperimentalFeatures.ControlPlaneExactVersion`.
- The tag is now **required** when `channelGroup` is `nightly`.
- The pin must be a build of the release line requested in `version.id`
  (pin `4.21.3` with `version.id: 4.20` is rejected).

### Why

Fixes the cluster update regression introduced by https://github.com/Azure/ARO-HCP/pull/6803, which has kept the `periodic-ci-Azure-ARO-HCP-main-periodic-stage-e2e-parallel-ocp-nightly` job red since it merged. `version.id` cannot round-trip an exact build: admission moved it to the internal-only `ControlPlaneExactVersion` (on no versioned ARM type) and truncated
`version.id` to its release line, so a GET returned only `4.21`. Sending that back on any PATCH or PUT looked like the customer removing the pin, so the mutation cleared it, and #6803's validator then rejected the bare `version.id` with a 400 -- meaning a nightly cluster could not be updated at all after create. An ARM tag does round-trip through GET, so moving the pin there removes the wedge instead of papering over it. This is the approach agreed in the #6803 review thread, which merged with the regression as a known caveat and no follow-up owner.

### Testing

1. `fix(validation):` -- the behaviour change, plus unit tests for the new admission and validation rules.
2. `test(test-integration):` -- replaces #6803's create-only artifact trees with trees for the new shape, and adds a PATCH round trip: create a nightly cluster with the tag, PATCH an unrelated tag, GET and assert the pin survived. That is the call that 400s on main today.
3. `test(e2e):` -- moves every exact build out of `OpenshiftVersionId` and into the tag, via two new framework helpers for create and upgrade paths.

- [x] Unit tests
- [x] [Integration tests](https://github.com/Azure/ARO-HCP/tree/main/test-integration)
- [x] [E2E tests](https://github.com/Azure/ARO-HCP/blob/main/test/e2e/README.md)
 
**Behaviour change beyond nightly.** A patch-bearing `version.id` was previously accepted for AFEC subscriptions on any channel group. It is now rejected everywhere, with an error pointing at the tag. The tag-only pin itself still works on any channel group; only the *requirement* is nightly-specific.

### PR Checklist

- [x] Self-reviewed the diff

If E2E tests are included:


internal/validation/validate_cluster.go
```diff
@@ -158,52 +159,63 @@ func validatePrivateKASRequiresMinimumVersion(_ context.Context, _ operation.Ope
...
-	effectiveVersionID := newCluster.CustomerProperties.Version.ID
-	if exact := newCluster.ServiceProviderProperties.ExperimentalFeatures.ControlPlaneExactVersion; exact != nil {
-		effectiveVersionID = exact.String()
+	tagPath := field.NewPath("tags").Key(metadataapi.TagClusterControlPlaneExactVersion)
+	exact := newCluster.ServiceProviderProperties.ExperimentalFeatures.ControlPlaneExactVersion
+
+	if newCluster.CustomerProperties.Version.ChannelGroup == metadataapi.ChannelGroupNightly && exact == nil {
+		return field.ErrorList{field.Required(tagPath,
+			"nightly builds are not published to the update graph, so the exact build must be pinned with this tag "+
+				"when channelGroup is \"nightly\" (e.g. \"4.21.0-0.nightly-2026-08-05-123456\")")}
```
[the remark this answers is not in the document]
tony-schndr: As I understand the only "customers" pinning version ID are e2e tests, this should be acceptable.

internal/admission/admit_cluster.go
```diff
@@ -419,9 +345,93 @@ func AdmitCluster(ctx context.Context, admissionContext *ClusterAdmissionContext
...
+	desiredVersion := admissionContext.ServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion
+	if desiredVersion == nil {
+		return nil
+	}
+
+	if newExact.LT(*desiredVersion) {
+		return field.ErrorList{field.Invalid(
+			field.NewPath("tags").Key(metadataapi.TagClusterControlPlaneExactVersion),
+			newExact.String(),
+			fmt.Sprintf("may not decrease the desired control plane version from %s", desiredVersion),
```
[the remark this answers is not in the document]
tony-schndr: @machi1990 I'm curious on your thoughts on this comment from copilot and since you implemented the controller I thought I'd check with you.

My thoughts are that unless we wrote an e2e test that creates a cluster without a version pin, then updates with a version pin, we would never hit this scenario.  And since this feature is internal use only it is not likely to happen unless we wrote a bad e2e test.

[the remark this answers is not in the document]
machi1990: We'll at some point support rollbacks (downgrades but it'll SREs initiated rollbacks this will be done via https://github.com/Azure/ARO-HCP/pull/6577).
When that feature is in place, it makes sense to not have another form of it via a customer facing API so locking out downgrades via validation makes sense. 

On top of this change, we need to also;
- get the latest active version in the admission check and validate that the pinned exact version doesn't attempt to downgrade
- enhance control_plane_desired_version_controller.go so that it performs the downgrade check in https://github.com/tony-schndr/ARO-HCP/blob/6e7413f9261ae3f89f1d8b9b4817769e713e32da/backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go#L183 otherwise reject it 
- enhance the operation controller so that when exact version is supplied we also validate the desired version matches that exact version https://github.com/tony-schndr/ARO-HCP/blob/6e7413f9261ae3f89f1d8b9b4817769e713e32da/backend/pkg/controllers/cluster/operations/operation_cluster_update.go#L250

internal/admission/admit_cluster.go
```diff
@@ -87,104 +87,9 @@ func MutateCluster(ctx context.Context, admissionContext *ClusterAdmissionContex
...
-		exact := parsedVersionID
-		newObj.ServiceProviderProperties.ExperimentalFeatures.ControlPlaneExactVersion = &exact
-		newObj.CustomerProperties.Version.ID = fmt.Sprintf("%d.%d", parsedVersionID.Major, parsedVersionID.Minor)
-	default:
-		// Neither the tag nor a patch-bearing version.id is present. If the old
-		// cluster carried an exact pin, the customer is removing it, so clear it.
-		if oldObj != nil && oldObj.ServiceProviderProperties.ExperimentalFeatures.ControlPlaneExactVersion != nil {
-			newObj.ServiceProviderProperties.ExperimentalFeatures.ControlPlaneExactVersion = nil
-		}
-	}
```
machi1990, line 184: Why did we loose this?

tony-schndr: With this change pinning an exact version is no longer allowed through `version.id` therefore this admission is removed.

The new path for mutating `.ServiceProviderProperties.ExperimentalFeatures.ControlPlaneExactVersion` happens in `mutateClusterExperimentalFeatures`

machi1990: Let me ask differently;

During an update of the cluster, what happens to an existing cluster that had an exact version supplied via the tag that's no longer carrying the exact version tag?

tony-schndr: PUT without the tag clears the tag.
PATCH without the tag doesn't clear the tag.

tony-schndr: And there is a guard to block clearing the tag if the channel is set to nightly

machi1990: Thank you. That covers it all.

backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go
```diff
@@ -183,6 +185,30 @@ func (c *controlPlaneVersionSyncer) SyncOnce(ctx context.Context, key controller
...
+		// (pinned before this guard existed, or a control plane that advanced out of band) must
+		// keep reconciling — re-litigating it would latch IntentFailed on a cluster that needs no
+		// write at all, which is the same wedge this change set exists to remove.
+		previousDesired := cachedServiceProviderCluster.Spec.ControlPlaneVersion.DesiredVersion
+		if previousDesired == nil || !exact.EQ(*previousDesired) {
+			// Unlike the monotonic guard on the resolved path below, which silently leaves desired
+			// unchanged when the graph moves on its own, this rejects: there the customer did
+			// nothing, here they explicitly asked for a lower build.
+			floor, floorKind := apihelpers.ControlPlaneVersionFloor(cachedServiceProviderCluster)
+			if floor != nil && exact.LT(*floor) {
```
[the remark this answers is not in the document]
tony-schndr: The equality shortcut can only skip the floor check in cases where the resulting write is a no-op, so no pin below the floor is ever persisted here.

In the scenario described — unpinned cluster with `DesiredVersion` 4.21.5, first pin of 4.21.5, active advances to 4.21.9 before this controller runs — `exact.EQ(*previousDesired)` is true, so `replacement` is a deep copy of the cached document with `DesiredVersion` set to a fresh pointer holding the value it already had. `equality.Semantic.DeepEqual(cachedServiceProviderCluster, replacement)` is therefore true and the `Replace` is skipped. `DesiredVersion` stays at exactly the value it held before the pin was added; nothing is lowered and no downgrade is triggered.

The only behavioural difference from re-checking the floor in that window is that no error is surfaced — and the thing it would reject is a write that does not happen. That is deliberate, for the reason in the comment above: an already-applied pin sitting below a floor that advanced out of band must keep reconciling, and latching `IntentFailed` on a cluster that needs no write is the wedge this change set exists to remove. The unpinned resolved path below behaves identically: its monotonic guard silently leaves `DesiredVersion` unchanged when active has moved past it.

Worth noting the first-pin case is not unguarded — admission distinguishes a first pin from an unchanged pin, because unlike this controller it can see the old pin:

https://github.com/Azure/ARO-HCP/blob/fd78b8e58b/internal/admission/admit_cluster.go#L408-L416

and it is covered by the `update-reject-first-exact-version-pin-below-active` integration artifact tree. This controller only sees current state, so `previousDesired` is the closest available proxy; where that proxy is imprecise, the write is a no-op anyway.

test-integration/frontend/artifacts/FrontendCRUD/Cluster/update-reject-lowered-exact-version-pin/01-httpCreate-cluster/cluster.json
```diff
@@ -0,0 +1,53 @@
...
+		"platform": {
+			"managedResourceGroup": "managed-resource-group-name",
+			"networkSecurityGroupId": "/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.Network/networkSecurityGroups/nsg",
+			"outboundType": "LoadBalancer",
+			"subnetId": "/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet"
+		},
+		"version": {
+			"channelGroup": "nightly",
+			"id": "4.20"
+		}
```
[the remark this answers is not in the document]
tony-schndr: Copilot is wrong, tag is set and the version is 4.20 ....

internal/admission/admit_cluster.go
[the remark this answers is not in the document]
tony-schndr: Added each change requested

internal/utils/apihelpers/version.go
```diff
@@ -35,6 +35,34 @@ func FindLowestAndHighestClusterVersion(activeVersions []coreapi.ServiceProvider
...
+// installing — there is nothing to decrease from.
+//
+// The floor is the greater of the desired version the desired-version controller resolved and the
+// highest version the control plane has actually run. Desired alone is not enough: it is unset until
+// the controller first resolves, and it can lag a control plane that advanced out of band.
+//
+// NOTE: blang semver compares pre-release identifiers lexically, not chronologically, so every
+// "4.x.0-0.nightly-multi-*" sorts above every "4.x.0-0.nightly-2026-*" regardless of build date.
+// Re-pinning a nightly cluster across those two streams is rejected by this floor.
+func ControlPlaneVersionFloor(serviceProviderCluster *coreapi.ServiceProviderCluster) (*semver.Version, string) {
```
machi1990: I am not entirely sure why this is needed.


I basically meant;
From the `serviceProviderCluster.Status.ControlPlaneVersion.ActiveVersions` we take the latest element; just like https://github.com/tony-schndr/ARO-HCP/blob/466e2e9244431d18141c0e26224633aa9f8122d0/backend/pkg/controllers/cluster/version/control_plane_desired_version_controller.go#L296 

And use the desired version to check for any downgrade (exact vs both of those versions)  in both the controller and the admission check

PR #6969, opened 2026-09-15
fix: require all operator identities on cluster create

n-gangrade:
https://redhat.atlassian.net/browse/ARO-29253

> **Reviewers:** 138 files, but only one changes behaviour —
> `internal/validation/validate_cluster.go`, which adds two checks. The other 137 are tests
> and fixtures.

### What

Cluster create is now rejected synchronously if it omits any operator identity the cluster
cannot come up without: 8 always-required control plane operators, 3 always-required data
plane operators, and `kms` when etcd encryption is `CustomerManaged`. Requirements are read
from `azure.ClusterScopedIdentitiesConfig`, which already describes them.

Operator names must also match a recognized operator exactly. A mis-cased or misspelled name
was previously accepted and then failed asynchronously.

Create-only, since `operatorsAuthentication` is immutable and running this on update would
permanently block clusters created before it existed.

### Why

A customer created a cluster with customer-managed etcd encryption but omitted the `kms`
identity. ARM returned `201 Accepted`, the create stalled indefinitely, and it took a Sev2 to
diagnose.

The existing cross-check only compares the two customer-supplied identity lists *against each
other* — omit an operator from both and they stay consistent, so it passes. Nothing checked
completeness. Several downstream consumers assume the full set, including the role-assignment
and identity-sync controllers, and the failure surfaces as a stalled create rather than an
error.

Same request: `upstream/main` returns `201 Accepted` plus an `Azure-AsyncOperation` header and
polls forever. This branch returns `400 InvalidRequestContent` naming the missing identity.

### Testing

**Unit** — `internal/validation` drops each of the 12 required identities in turn and asserts
the create is rejected for exactly that operator. Verified by mutation: neutering either
requirement loop fails these tests and nothing else.

**Contract test** — `internal/azure/cluster_scoped_identities_config_test.go` pins the expected
operator sets with hardcoded names, since validation and its fixtures both derive from the same
config and would otherwise only prove they agree with each other.

**Integration** — two new cases, `create-without-kms-identity` and
`create-without-ingress-identity`. Both were run against `upstream/main` with only the test
directory copied in; they fail there with `An error is expected but got nil` and pass here,
confirming they detect the defect rather than just agreeing with the new code.

No new E2E tests — the E2E framework already supplies all 12 identities.

### Special notes for reviewer

**Names are rejected, not normalized.** `"KMS"` could be lowercased so it just works, but
that's a silent mutation — you PUT `"KMS"` and GET `"kms"` back. Strict-to-lenient is safe to
loosen later; the reverse isn't. Happy to switch if you'd prefer.

**CI is red for infrastructure reasons, not this change.** All Prow jobs are failing to
schedule with `nonretryable error: no build client found for cluster "build11"/"build13"`.  Will retest once CI is healthy.

Two gaps left out of scope: `serviceManagedIdentity` has no requirement metadata to
validate against, and the `MaximumSet_Gen.json` examples (10 files across 5 API versions)
show empty operator maps, so they now document a request the RP rejects.

### PR Checklist

- [ ] Screenshots included (if dashboards or other UI changes) — N/A
- [x] Self-reviewed the diff


n-gangrade:

test-integration/frontend/artifacts/FrontendCRUD/Cluster/create-without-ingress-identity/01-httpCreate-cluster/cluster.json
```diff
@@ -0,0 +1,78 @@
...
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/cloud-network-config-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/cluster-api-azure-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/control-plane-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/disk-csi-driver-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/file-csi-driver-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/image-registry-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/kms-identity": {}
```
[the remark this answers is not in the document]
n-gangrade: Both of these are based on a misreading — the omitted identity is removed from the top-level identity.userAssignedIdentities map as well, in both fixtures.

create-without-kms-identity has 8 entries in the identity map and 8 in controlPlaneOperators, with kms-identity and kms absent from each. create-without-ingress-identity likewise omits ingress-identity and ingress from both; the kms-identity entry visible in that diff is correct, since kms isn't the omitted one there.

Because both maps stay consistent, the cross-field validator reports nothing, which is exactly why these cases fail on upstream/main with An error is expected but got nil and pass here with a single error.

Applying the suggested changeset to the ingress fixture would remove kms-identity while kms remains in controlPlaneOperators, producing identity is not assigned to this resource and defeating the point of the test.

test-integration/frontend/artifacts/FrontendCRUD/Cluster/create-without-kms-identity/01-httpCreate-cluster/cluster.json
```diff
@@ -0,0 +1,78 @@
...
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/cloud-network-config-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/cluster-api-azure-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/control-plane-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/disk-csi-driver-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/file-csi-driver-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/image-registry-identity": {},
+			"/subscriptions/6b690bec-0c16-4ecb-8f67-781caf40bba7/resourceGroups/bar/providers/Microsoft.ManagedIdentity/userAssignedIdentities/ingress-identity": {}
```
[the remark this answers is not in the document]
n-gangrade: Both of these are based on a misreading — the omitted identity is removed from the top-level identity.userAssignedIdentities map as well, in both fixtures.

create-without-kms-identity has 8 entries in the identity map and 8 in controlPlaneOperators, with kms-identity and kms absent from each. create-without-ingress-identity likewise omits ingress-identity and ingress from both; the kms-identity entry visible in that diff is correct, since kms isn't the omitted one there.

Because both maps stay consistent, the cross-field validator reports nothing, which is exactly why these cases fail on upstream/main with An error is expected but got nil and pass here with a single error.

Applying the suggested changeset to the ingress fixture would remove kms-identity while kms remains in controlPlaneOperators, producing identity is not assigned to this resource and defeating the point of the test.

test-integration/frontend/artifacts/FrontendCRUD/Cluster/create-without-ingress-identity/01-httpCreate-cluster/cluster.json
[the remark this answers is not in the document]
n-gangrade: Rechecked this fixture on the current branch. create-without-ingress-identity uses no-ingress-cluster in both cluster.json and 00-key.json. It omits ingress from controlPlaneOperators and ingress-identity from identity.userAssignedIdentities, while retaining kms and kms-identity.

The fixture is intentionally testing the missing-ingress case and is internally consistent. The comment appears to be based on a misread of the diff. I’m resolving this thread.

test-integration/frontend/artifacts/FrontendCRUD/Cluster/create-without-kms-identity/01-httpCreate-cluster/cluster.json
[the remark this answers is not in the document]
n-gangrade: Rechecked this fixture on the current branch. create-without-kms-identity uses no-kms-cluster in both cluster.json and 00-key.json. It omits kms from controlPlaneOperators and kms-identity from identity.userAssignedIdentities, while retaining ingress and ingress-identity.

The fixture is intentionally testing the missing-KMS case and is internally consistent. The comment appears to be based on a misread of the diff. I’m resolving this thread.

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,181 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
 	return nil
 }
 
+// clusterScopedIdentities is the source of truth for which operator identities a cluster needs.
+// The role definition config set only affects the Azure role IDs attached to each identity, which
+// are irrelevant here, so either set yields the same requirements.
+var clusterScopedIdentities = azure.NewClusterScopedIdentitiesConfig(azure.RoleDefinitionConfigSetNameDev)
```
miguelsorianod, line 220: We need to improve on this a bit:
This is initializing the configuration here as a package-level variable and hardcoding the usage of `azure.RoleDefinitionConfigSetNameDev`. That can change per environment. What we should do is:
1. Add a `--azure-cluster-scoped-identities-role-set-name` CLI flag to the RP Frontend binary, similarly to how it's done for RP Backend
2. Instantiate the config in the top-level layer of the application and pass it as a Frontend option (like for example the CS (Clusters Service) client
3. Pass the config to where we want to perform the validation. In this specific case, I am unsure about here (ValidateCluster) is the appropriate place, as we don't currently pass dependencies (unless passed as an option?). Maybe the cluster admission layer would be more appropriate? something I'd like to discuss a bit first with the group

miguelsorianod:
This is great :+1:, we had the task of moving these validations from CS to the RP pending so we returned errors synchronously at request-response time instead of failing silently asynchronously, so this should cover it once ready.

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,181 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+		},
+		enabledBy: "properties.etcd.dataEncryption.keyManagementMode is CustomerManaged",
+	},
+}
+
+// validateRequiredOperatorIdentities rejects a create that omits an operator identity the cluster
+// cannot come up without. The existing cross-check only compares the two customer-supplied identity
+// lists against each other, so omitting an operator from both leaves them consistent and passes.
+//
+// Create-only: operatorsAuthentication is immutable, so running this on update would permanently
```
miguelsorianod, line 253: This should not be an issue: The validation still occurs on CS side asynchronously which means that those clusters always fail on creation. So if there are some existing ones, they are in a failed state anyway. I think we can safely introduce logic without worrying about pre-existing resources.

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,181 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+	{
+		operatorName: string(azure.ClusterOperatorIdentifierKMS),
+		isEnabled: func(newCluster *coreapi.HCPOpenShiftCluster) bool {
+			return newCluster.CustomerProperties.Etcd.DataEncryption.KeyManagementMode == metadataapi.EtcdDataEncryptionKeyManagementModeTypeCustomerManaged
+		},
+		enabledBy: "properties.etcd.dataEncryption.keyManagementMode is CustomerManaged",
+	},
+}
+
+// validateRequiredOperatorIdentities rejects a create that omits an operator identity the cluster
```
miguelsorianod, line 249: > validateRequiredOperatorIdentities rejects a create that omits an operator identity the cluster
> cannot come up without

This is good: Conceptually, this is checking the operators that are considered as required for the cluster. Of those, there are two types:
* Unconditionally required
* Required on enablement of a specific functionality (it depends on the shape of the cluster). I see this is attempted to be covered with `conditionallyRequiredControlPlaneOperator`. Conceptually, all kinds of operators (not only controlplane operators) can have conditionally required identities, so maybe we could generalize this a bit more so we can support both controlplane and dataplane operators (I call this generalized concept "Cluster Scoped Operator Identity"

Another relevant note: Operators can exist or not exist depending on the OpenShift version that we are dealing with. If the operator does not exist for the openshift version then we should error if it's been specified, even if the operator name is a recognized operator name. The `IsSupportedForOpenShiftVersion` method of the config type should help to detect that. I see that for controlplaneoperators this is being leveraged in some way, but I don't see it for the data plane operators. We should improve on this.

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,181 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+
+// validateOperatorIdentityNames rejects operator names the service does not recognize.
+//
+// The backend indexes operators by exact name, so a mis-cased or misspelled key is accepted at
+// create time and then fails asynchronously -- the same silent failure this validation exists to
+// prevent. Matching is therefore exact, unlike the requirement check, which stays case-insensitive
+// so that a mis-cased key reports one clear error rather than two.
+//
+// Create-only, for the same reason as validateRequiredOperatorIdentities.
+func validateOperatorIdentityNames(_ context.Context, op operation.Operation, newCluster, _ *coreapi.HCPOpenShiftCluster) field.ErrorList {
```
miguelsorianod, line 326: This is good too: We should error when an operator identifier is not defined in the cluster-scoped identities configuration, independently of the OCP version being specified

internal/validation/validate_cluster.go
n-gangrade: You're right.  If CS rejects incomplete identity sets asynchronously, then no healthy cluster exists without them, so there's nothing to lock out.  I corrected the comment.

internal/validation/validate_cluster.go
n-gangrade: I have added both of these now (9b08a4fb2).

The conditional table now carries a plane, so data plane operators can be conditionally required.  Name validation now also rejects recognized operators that don't exist for the cluster's version - layered separately from name recognition, since an unparseable version shouldn't disable the latter.

Worth noting: every operator currently has `MinVersionInclusive: 4.19` and no max, and the validation floor is 4.20, so the version check can't fire with today's config. The tests pin an operator's max version artificially to exercise it, which means swapping the package-level `clusterScopedIdentities` — one more argument for making it injectable as you suggested.

internal/validation/validate_cluster.go
n-gangrade: Agreed, and that's now enforced separately from the version check.  Unknown identifiers are rejected unconditionally, version support only when the version parses. There's a test pinning it, since the obvious implementation would have coupled the two.

internal/validation/validate_cluster.go
n-gangrade: I've done 1 and 2 in 9304618de — the frontend now takes `--azure-cluster-scoped-identities-role-set-name` (dev/public, defaulting to dev, same shape and error message as the backend), builds the config in `Run()`, and holds it on `Frontend`.

Left 3 alone: the validation still reads the package-level config, since wiring the Frontend's copy through depends on whether the check stays in `ValidateCluster` or moves to admission. Has the group had a chance to discuss it?

internal/validation/validate_cluster.go
n-gangrade: You're right. If CS rejects incomplete identity sets asynchronously, no healthy cluster exists without them, so there's nothing to lock out. Corrected the comment.

frontend/cmd/cmd.go
```diff
@@ -113,7 +123,8 @@ func NewRootCmd() *cobra.Command {
 
 func NewFrontendOpts() *FrontendOpts {
 	return &FrontendOpts{
-		exitOnPanic: true,
+		exitOnPanic:                             true,
+		azureClusterScopedIdentitiesRoleSetName: string(internalazure.RoleDefinitionConfigSetNameDev),
```
miguelsorianod: Don't set any default. It's a required parameter

frontend/pkg/frontend/frontend.go
```diff
@@ -77,6 +78,11 @@ type Frontend struct {
 
 	apiRegistry coreapi.APIRegistry
 
+	// clusterScopedIdentities describes which operator identities a cluster requires. The role
+	// definition config set it is built from varies per environment, so it is supplied by the
+	// caller rather than chosen here.
+	clusterScopedIdentities *azure.ClusterScopedIdentitiesConfig
```
miguelsorianod: nitpick: name it clusterScopedIdentitiesConfig

frontend/cmd/cmd.go
miguelsorianod: This also means we need to set it in the deployments, and make the deployment configurable, similar to how it's done in backend

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,258 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+
+// conditionallyRequiredOperatorIdentity names an operator identity that is only required once the
+// cluster enables the feature that uses it, mirroring azure.IdentityRequirementTypeOnEnablement.
+// Operators marked azure.IdentityRequirementTypeAlways are read from clusterScopedIdentities
+// instead.
+//
+// Note that CustomerManaged is currently the only accepted etcd key management mode, so the
+// kms entry below is required on every create in practice.
+type conditionallyRequiredOperatorIdentity struct {
+	plane        operatorPlane
```
miguelsorianod: I am not sure that I like the `operatorPlane` concept. Why not having conditionallyRequiredControlPlaneOperatorIdentity, conditionallyRequiredDataPlaneOperatorIdentity?

Keep in mind that the same operatorName can be in both controlplane and dataplane side. An example is the `image-registry` one

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,258 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+	}
+	operatorConfig, ok := clusterScopedIdentities.ControlPlaneOperatorsIdentities[identifier]
+	return !ok || operatorConfig.IsSupportedForOpenshiftVersion(version)
+}
+
+// validateRequiredOperatorIdentities rejects a create that omits an operator identity the cluster
+// cannot come up without. The existing cross-check only compares the two customer-supplied identity
+// lists against each other, so omitting an operator from both leaves them consistent and passes.
+//
+// Create-only: operatorsAuthentication is immutable, so an update cannot change the identity set.
```
miguelsorianod: It doesn't matter that it's immutable. We shouldn't differentiate between operations. If it's immutable then this should work correctly, or the immutable part will fail. The same applies to the other function.

This can also for example start to fail if the operators identities config itself changes and an update comes even if no changes on the identities occur

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,258 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+	for _, operatorName := range unrecognized {
+		errs = append(errs, field.Invalid(fldPath.Key(operatorName), operatorName,
+			fmt.Sprintf("unrecognized operator name; supported values: %s", strings.Join(recognized, ", "))))
+	}
+	return errs
+}
+
+// operatorIdentitiesByLowercaseName rekeys operator identities so lookups are case-insensitive,
+// consistent with the resource ID comparisons in validateOperatorAuthenticationAgainstIdentities.
+func operatorIdentitiesByLowercaseName(operators map[string]*azcorearm.ResourceID) map[string]*azcorearm.ResourceID {
```
miguelsorianod: I am not sure I understand this. Operator names shouldn't be lowercased as they are not case-insensitive. Case-insensitivity matters for Azure ResourceIDs but not for operator names. If an operator name doesn't exactly match the definition in cluster-scoped identities config it is not recognized

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,258 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+		if existing, ok := byLowercaseName[lowercaseName]; ok && existing != nil {
+			continue
+		}
+		byLowercaseName[lowercaseName] = identity
+	}
+	return byLowercaseName
+}
+
+func operatorIdentitySupplied(operators map[string]*azcorearm.ResourceID, operatorName string) bool {
+	identity, ok := operators[strings.ToLower(operatorName)]
```
miguelsorianod: Same comment as https://github.com/Azure/ARO-HCP/pull/6969/changes#r4065624401 about operator names

frontend/cmd/cmd.go
n-gangrade: Removed the default, so validate() rejects the empty value like the backend does

frontend/pkg/frontend/frontend.go
n-gangrade: Done.

internal/validation/validate_cluster.go
n-gangrade: Dropped OperatorPlane, it's conditionallyRequiredControlPlaneOperatorIdentity now with plane-specific helpers.  Didn't add the data plane type since kms is the only conditional entry and it's control plane — can add it when something actually needs it.

internal/validation/validate_cluster.go
n-gangrade: done, dropped the gate from both functions.

internal/validation/validate_cluster.go
n-gangrade: Agreed, matching is exact now.  The remaining ToLower calls in the file are all on resource ids.

internal/validation/validate_cluster.go
n-gangrade: Same fix.  operatorIdentitySupplied does an exact lookup now.

frontend/Makefile
```diff
@@ -36,7 +36,8 @@ run: $(BINARY)
 	./$(BINARY) --location ${LOCATION} \
 		--clusters-service-url http://localhost:8000 \
 		--cosmos-name ${DB_NAME} \
-		--cosmos-url ${DB_URL}
+		--cosmos-url ${DB_URL} \
+		--azure-cluster-scoped-identities-role-set-name dev
```
miguelsorianod: If we want to set it in the Makefile let's not hardcode it, let's retrieve it dynamically like DB_URL and DB_NAME. See Env.mk in the `frontend` directory.

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,211 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+
+	errs := field.ErrorList{}
+
+	controlPlaneRequired := make([]string, 0, len(clusterScopedIdentities.ControlPlaneOperatorsIdentities))
+	for operatorName := range clusterScopedIdentities.AlwaysRequiredControlPlaneOperators(&version) {
+		controlPlaneRequired = append(controlPlaneRequired, string(operatorName))
+	}
+	// Sorted so error ordering does not depend on map iteration order.
+	slices.Sort(controlPlaneRequired)
+	for _, operatorName := range controlPlaneRequired {
```
miguelsorianod: The body of the loop can be flipped and simply be:

if !operatorIdentitySupplied(controlPlaneSupplied, operatorName) {
		errs = append(errs, field.Required(controlPlanePath.Key(operatorName), fmt.Sprintf("a user-assigned identity for the %q control plane operator is required", operatorName)))}
}

Similar approach can be taken for the datapPlaneRequired loop

internal/validation/validate_cluster.go
```diff
@@ -101,6 +102,12 @@ func ValidateCluster(ctx context.Context, op operation.Operation, newCluster, ol
 	// version profile and the service-provider exact pin, so it lives at cluster level.
 	errs = append(errs, validateNightlyChannelRequiresFullVersion(ctx, op, newCluster, oldCluster)...)
 
+	// some operator identities only become required once the feature that uses them is enabled
+	errs = append(errs, validateRequiredOperatorIdentities(ctx, op, newCluster, oldCluster)...)
```
miguelsorianod: If we end up having the logic in validate_cluster (pending to discuss), we should move this inside and the other function call inside `validateUserAssignedIdentitiesProfile`. The validate_cluster design follows nested calls that follow the json hierarchy (with the exception of checks that are cross-fields)

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,211 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+	for operatorName := range clusterScopedIdentities.DataPlaneOperatorsIdentities {
+		names = append(names, string(operatorName))
+	}
+	slices.Sort(names)
+	return names
+}
+
+// controlPlaneOperatorSupportedForVersion reports whether the control plane operator exists at all
+// for the given OpenShift version. Unrecognized names are reported separately, so they count as
+// supported here to keep a single key from producing two errors.
```
miguelsorianod: > Unrecognized names are reported separately, so they count as
// supported here to keep a single key from producing two errors.

No let's not do that. Returning `controlPlaneOperatorSupportedForVersion` true when the operator is not even recognized is not correct.

It's not an issue to return that it's both non recognized and no supported (two error messages).
Same applies to dataPlaneOperatorSupportedForVersion.

Document the behavior for the functions (an unrecognized one is also returned as false here)

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,211 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+// dataPlaneOperatorSupportedForVersion is the data plane counterpart of
+// controlPlaneOperatorSupportedForVersion.
+func dataPlaneOperatorSupportedForVersion(operatorName string, version *semver.Version) bool {
+	operatorConfig, ok := clusterScopedIdentities.DataPlaneOperatorsIdentities[azure.ClusterOperatorIdentifier(operatorName)]
+	return !ok || operatorConfig.IsSupportedForOpenshiftVersion(version)
+}
+
+// validateRequiredOperatorIdentities rejects a cluster that omits an operator identity the cluster
+// cannot come up without. The existing cross-check only compares the two customer-supplied identity
+// lists against each other, so omitting an operator from both leaves them consistent and passes.
```
miguelsorianod: > The existing cross-check only compares the two customer-supplied identity lists against each other, so omitting an operator from both leaves them consistent and passes.

Do you mean `.customerPropreties.Platform.OperatorsAuthentication.UserAssignedIdentities.ControlPlaneOperators and  `.identity` maps? if so, remove this comment as it doesn't provide value.
Document what this function actually does

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,211 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+			continue
+		}
+		errs = append(errs, field.Required(dataPlanePath.Key(operatorName), fmt.Sprintf("a user-assigned identity for the %q data plane operator is required", operatorName)))
+	}
+
+	for _, operator := range conditionallyRequiredControlPlaneOperatorIdentities {
+		if !operator.isEnabled(newCluster) {
+			continue
+		}
+		if !controlPlaneOperatorSupportedForVersion(operator.operatorName, &version) {
```
miguelsorianod: Document why the continue here and in operatorIdentitySupplied

internal/validation/validate_cluster.go
```diff
@@ -207,6 +214,211 @@ func validateNightlyChannelRequiresFullVersion(_ context.Context, op operation.O
...
+		errs = append(errs, field.Required(controlPlanePath.Key(operator.operatorName), fmt.Sprintf("a user-assigned identity for the %q control plane operator is required when %s", operator.operatorName, operator.enabledBy)))
+	}
+
+	return errs
+}
+
+// validateOperatorIdentityNames rejects operator names the service does not recognize.
+//
+// The backend indexes operators by exact name, so a mis-cased or misspelled key is accepted at
+// create time and then fails asynchronously -- the same silent failure this validation exists to
```
miguelsorianod: > The backend indexes operators by exact name, so a mis-cased or misspelled key is accepted at
// create time and then fails asynchronously

I am not sure I understand this correctly. A mis-cased or misspelled key shouldn't be accepted at create time.

Is it just explaining that this catches that? if so, just document that name checking is case sensitive

internal/validation/validate_cluster.go
miguelsorianod: >     In this specific case, I am unsure about here (ValidateCluster) is the appropriate place, as we don't currently pass dependencies (unless passed as an option?). Maybe the cluster admission layer would be more appropriate? something I'd like to discuss a bit first with the group

After discussing this with the wider group, it matches with my original thinking. This should be moved to the admission validation and not a cluster validation. The reason is because it depends on an external instantiated dependency.

n-gangrade: All three are done.  Part 3 in c568b4fc.  Both checks moved into admission and take the config from ClusterAdmissionContext, which the frontend populates.  The package-level variable is gone.

frontend/Makefile
n-gangrade: Done in 58a5ce4ef. Added CLUSTER_SCOPED_IDENTITIES_ROLE_SET_NAME to Env.mk, sourced from .frontend.clusterScopedIdentitiesConfig.roleSetName, so make run picks up the same value the deployment uses instead of a separate hardcoded copy.

internal/validation/validate_cluster.go
n-gangrade: Fixed both loops.  They live in admit_cluster.go now after the move to admission, but the shape is as you described.

internal/validation/validate_cluster.go
n-gangrade: Moot now, both moved to admission in c568b4fc2.

internal/validation/validate_cluster.go
n-gangrade: Done in 715ccb137. Both return false for an unrecognized operator now, and the doc comments say so. A misspelled key reports both that the name is unrecognized and that it does not exist for the version.

internal/validation/validate_cluster.go
n-gangrade: Done in f2ce10833. You read it right, it meant the operator maps vs .identity. I replaced it with what the function actually checks: the operators the config marks as always required for the cluster's version, plus the ones that become required when the feature using them is enabled.

internal/validation/validate_cluster.go
n-gangrade: Done in 3867a8c5f, one line on each continue.

internal/validation/validate_cluster.go
n-gangrade: Done in cf5edcbd5. It now just says name matching is case-sensitive and why.

test-integration/frontend/artifacts/FrontendCRUD/Cluster/peer-field-validation/01-httpCreate-cluster/expected-error.txt
```diff
@@ -43,6 +43,21 @@
...
+      {
+        "code": "InvalidRequestContent",
+        "message": "Invalid value: \"operator2\": unrecognized operator name; supported values: cloud-controller-manager, cloud-network-config, cluster-api-azure, control-plane, disk-csi-driver, file-csi-driver, image-registry, ingress, kms",
+        "target": "properties.platform.operatorsAuthentication.userAssignedIdentities.controlPlaneOperators[operator2]"
+      }
+      {
+        "code": "InvalidRequestContent",
+        "message": "Invalid value: \"dataplane-operator\": unrecognized operator name; supported values: disk-csi-driver, file-csi-driver, image-registry",
+        "target": "properties.platform.operatorsAuthentication.userAssignedIdentities.dataPlaneOperators[dataplane-operator]"
+      }
```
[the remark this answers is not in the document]
n-gangrade: The test passes. Expected errors are matched per object as substrings, so additional errors in the response do not fail the fixture.

internal/admission/admit_cluster.go
```diff
@@ -431,9 +438,215 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
 	errs = append(errs, admitClusterEtcdKmsKeyVersionChange(ctx, admissionContext, op, fldPath.Child("etcd", "dataEncryption", "customerManaged", "kms", "activeKey", "version"), newObj, oldObj)...)
 	errs = append(errs, admitClusterPlatform(ctx, admissionContext, op, fldPath.Child("platform"), &newObj.Platform)...)
 
+	identitiesPath := fldPath.Child("platform", "operatorsAuthentication", "userAssignedIdentities")
+	errs = append(errs, admitRequiredOperatorIdentities(admissionContext, identitiesPath, newObj)...)
+	errs = append(errs, admitOperatorIdentityNames(admissionContext, identitiesPath, newObj)...)
```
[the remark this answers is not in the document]
n-gangrade: Not applying. This was removed deliberately in review: operatorsAuthentication is immutable, but the required set comes from config and can change independently, so gating on operation type gives no real guarantee. Update tests cover both a complete set passing and a dropped identity failing.

frontend/cmd/cmd.go
```diff
@@ -100,6 +103,13 @@ func NewRootCmd() *cobra.Command {
 	rootCmd.Flags().StringVar(&opts.clustersServiceURL, "clusters-service-url", "https://api.openshift.com", "URL of the OCM API gateway.")
 	rootCmd.Flags().BoolVar(&opts.insecure, "insecure", false, "Skip validating TLS for clusters-service.")
 
+	rootCmd.Flags().StringVar(
+		&opts.azureClusterScopedIdentitiesRoleSetName,
+		"azure-cluster-scoped-identities-role-set-name",
+		opts.azureClusterScopedIdentitiesRoleSetName,
```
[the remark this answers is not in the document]
n-gangrade: Not applying. The default was removed at review request since the flag is required. The documented local path is make run, which passes it from Env.mk, and there is no documented raw binary invocation.

frontend/pkg/frontend/cluster.go
```diff
@@ -273,9 +273,10 @@ func (f *Frontend) newClusterAdmissionContext(ctx context.Context, op operation.
 	}
 
 	admissionContext := &admission.ClusterAdmissionContext{
-		Clock:           f.clock,
-		Subscription:    subscription,
-		OriginalCluster: originalCluster.DeepCopy(),
+		Clock:                   f.clock,
+		Subscription:            subscription,
+		OriginalCluster:         originalCluster.DeepCopy(),
+		ClusterScopedIdentities: f.clusterScopedIdentitiesConfig,
```
miguelsorianod, line 279: nitpick: name it ClusterScopedIdentitiesConfig

internal/admission/admit_cluster.go
```diff
@@ -431,9 +438,215 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
 	errs = append(errs, admitClusterEtcdKmsKeyVersionChange(ctx, admissionContext, op, fldPath.Child("etcd", "dataEncryption", "customerManaged", "kms", "activeKey", "version"), newObj, oldObj)...)
 	errs = append(errs, admitClusterPlatform(ctx, admissionContext, op, fldPath.Child("platform"), &newObj.Platform)...)
 
+	identitiesPath := fldPath.Child("platform", "operatorsAuthentication", "userAssignedIdentities")
+	errs = append(errs, admitRequiredOperatorIdentities(admissionContext, identitiesPath, newObj)...)
```
miguelsorianod, line 442: These calls have to be inside admitClusterPlatform

miguelsorianod: Even better, ideally have an admitOperatorsAuthentication within admitClusterPlatform, then an admitUserAssignedIdentities within it and then within that one call the two methods. In that way the function call hierarchy follows the api structure hierarchy

internal/admission/admit_cluster.go
```diff
@@ -431,9 +438,215 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
...
+	return ok && operatorConfig.IsSupportedForOpenshiftVersion(version)
+}
+
+// unsupportedOperatorNameErrors rejects operator names that do not exist for the cluster's
+// OpenShift version. A name the service does not recognize is reported here as well as by
+// unrecognizedOperatorNameErrors, since it is both unknown and unavailable for the version.
+func unsupportedOperatorNameErrors(supplied map[string]*azcorearm.ResourceID, fldPath *field.Path, planeLabel string, version *semver.Version, isSupported func(operatorName string, version *semver.Version) bool) field.ErrorList {
+	unsupported := make([]string, 0, len(supplied))
+	for operatorName := range supplied {
+		if operatorName == "" || isSupported(operatorName, version) {
```
miguelsorianod: add comment:
		// An empty name is already reported by the static identity validation.

miguelsorianod:
I left a few more minor comments. At this point the PR is in a quite advanced state.
It also needs a rebase solving conflicts from main

internal/validation/validate_cluster.go
```diff
@@ -101,6 +101,9 @@ func ValidateCluster(ctx context.Context, op operation.Operation, newCluster, ol
 	// version profile and the service-provider exact pin, so it lives at cluster level.
 	errs = append(errs, validateNightlyChannelRequiresFullVersion(ctx, op, newCluster, oldCluster)...)
 
+	// some operator identities only become required once the feature that uses them is enabled,
```
miguelsorianod: I guess leftover comment here

frontend/pkg/frontend/cluster.go
n-gangrade: Done in 1191c3070. Renamed to ClusterScopedIdentitiesConfig.

internal/admission/admit_cluster.go
n-gangrade: Done in 1191c3070. Nested it as you described. admitRequiredOperatorIdentities and admitOperatorIdentityNames also take customerProperties, since they need the cluster version and etcd mode.

internal/admission/admit_cluster.go
n-gangrade: Done in 1191c3070.

internal/validation/validate_cluster.go
n-gangrade: Done in 1191c3070.

internal/admission/admit_cluster.go
```diff
@@ -431,9 +438,214 @@ func admitClusterCustomerProperties(ctx context.Context, admissionContext *Clust
 	errs = append(errs, admitClusterEtcdKmsKeyVersionChange(ctx, admissionContext, op, fldPath.Child("etcd", "dataEncryption", "customerManaged", "kms", "activeKey", "version"), newObj, oldObj)...)
 	errs = append(errs, admitClusterPlatform(ctx, admissionContext, op, fldPath.Child("platform"), &newObj.Platform)...)
 
+	identitiesPath := fldPath.Child("platform", "operatorsAuthentication", "userAssignedIdentities")
+	errs = append(errs, admitRequiredOperatorIdentities(admissionContext, identitiesPath, newObj)...)
+	errs = append(errs, admitOperatorIdentityNames(admissionContext, identitiesPath, newObj)...)
```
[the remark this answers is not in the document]
n-gangrade: Not applying. The create-only gate was removed deliberately in review: operatorsAuthentication is immutable, but the required set comes from config and can change independently of the cluster, so gating on operation type doesn't give the guarantee it appears to. Update tests cover both a complete set passing and a dropped identity failing.

frontend/pkg/frontend/frontend.go
```diff
@@ -117,8 +124,9 @@ func NewFrontend(
 				return utils.ContextWithLogger(context.Background(), logger)
 			},
 		},
-		auditClient:       auditClient,
-		resourcesDBClient: resourcesDBClient,
+		auditClient:             auditClient,
+		resourcesDBClient:       resourcesDBClient,
+		clusterScopedIdentities: clusterScopedIdentities,
```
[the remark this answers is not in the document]
n-gangrade: Fixed. The config is threaded through to admission now via ClusterAdmissionContext, the package-level variable is gone, and the deployment passes the flag.

test-integration/frontend/create_readonly_test.go
```diff
@@ -153,6 +160,13 @@ func TestCreateIgnoresReadOnlyFields(t *testing.T) {
 					}
 					body, err := json.Marshal(payload)
 					require.NoError(t, err)
+					if tc.kind == "Cluster" {
+						// The operatorsAuthentication block above replaces the one the payload
+						// ships with, so put the required operator identities back. Re-read the
+						// result so the assertions below compare against what was actually sent.
+						body = withRequiredOperatorIdentities(body, subscriptionID)
+						require.NoError(t, json.Unmarshal(body, &payload))
+					}
```
machi1990: This change is confusing;

The overall intent of this test is to ensure that readonly fields are are ignored when the end user sets them.

It seems that once https://github.com/Azure/ARO-HCP/pull/7061 was merged, the addition in this PR didn't build nicely on top of the existing changes.

Notably, these https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L119 and https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L128 which were there before because the initial payload wasn't shipping the operators identity section and the .identity section. 
However, following the change in this PR, we are always setting these sections https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/cross_version_roundtrip_test.go#L254

My proposal is thus;
- We drop https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L119  and https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L128 and L134 lines 
- Enhance https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/cross_version_roundtrip_test.go#L254 so that it also write kms identity when the corresponding etcd profile is set.
- In the `inject` part https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L152 we change the `identityID` in use so that it injects to some spoofedIDs to all the identities (at the end we do a get we'll verify that they are dropped indeed). Also https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L150 only has to be done if the kind is a cluster. Notice how it only does it if the kind isn't externalauth. In reality only the cluster at the moment can have assigned identities. So worth moving that to https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L154 block. Similary this block https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L190 only has to be exercised if the kind is a Cluster
- This https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L163 won't be needed anymore as the payload already contains the required identities
- This https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L230 also is modified to write a clientID/PrincipalID for all the identities and then https://github.com/n-gangrade/ARO-HCP/blob/1191c30708aa874c2ef1a55f27fe3330c8bcc8bf/test-integration/frontend/create_readonly_test.go#L241 assertion is also changed as such to verify that each served identity has the values written

machi1990 (changes requested):

test-integration/frontend/create_readonly_test.go
n-gangrade: Done in 47b70933d. Dropped the overrides and the re-apply after marshalling, and withRequiredOperatorIdentities now adds kms when the payload sets customer-managed etcd, so the templates no longer hardcode it. Spoofing and both assertions cover every identity now.

One deviation on point 3: I kept the identity block for node pools. NodePool does have an Identity field and nodepools_methods.go:49 clears it, so Cluster-only leaves that path uncovered. Happy to switch if you'd rather.

PR #6972, opened 2026-09-16
test: remove AFEC-gated API version pre-check from E2E tests (AROSLSRE-2118)

raelga:
Jira: [AROSLSRE-2118](https://redhat.atlassian.net/browse/AROSLSRE-2118)

### What

Removes the `tc.IsHCPAPIVersionAvailable` pre-check from the two remaining E2E tests that use it (`cluster_create_v20260901preview.go`, `cluster_nodepool_active_versions.go`) and deletes the now-unused `IsHCPAPIVersionAvailable` function and `requiredResourceTypesForAPIVersion` var from `test/util/framework/hcp_helper.go`.

### Why

`IsHCPAPIVersionAvailable` queries `Providers.Get("Microsoft.RedHatOpenShift")` via the standard ARM SDK. The API versions listed in that response are gated by the `HcpPrivatePreview` AFEC feature, which is registered per subscription and does not reflect whether the RP actually serves the version. The Integration and Stage test subscriptions lack this AFEC flag, so the pre-check reports API versions as unavailable even though the RP genuinely serves them, since real cluster-create calls are routed separately through `INT-APPROVED`/`STAGING-APPROVED` test endpoints unaffected by this restriction. This caused false failures like `API version 2026-06-30-preview should be fully available by ...` in Integration and Stage even when the version was fully deployed.

The `isAPINotDeployedError` post-check, which inspects the real error returned by the actual cluster-create call, is unaffected by this AFEC visibility gap and remains the sole "not yet deployed" signal paired with each test's `timeBombDeadline`.

### Testing

`cd test && CGO_ENABLED=0 go build ./...` and `CGO_ENABLED=0 go vet ./e2e/... ./util/framework/...` both pass clean. No golden fixture regeneration needed since no `Describe`/`Context`/`It` strings changed.

Follow-up agreed in the ARO-HCP first-party Slack thread on the `2026-06-30-preview` E2E failures: migrate off the retired-looking version first (#6961), then remove the unreliable pre-check once green.

### PR Checklist

- [x] Tests pass locally
- [x] No unrelated changes bundled


mbarnes (changes requested):
Changes here look good, but since this is a follow-up I hope you'll forgive some scope creep.

I noticed `isAPINotDeployedError` is defined in `/test/e2e/nodepool_ephemeral_osdisk.go`.  If we're going to rely more heavily on this post-check pattern, that function should be moved to the `/test/util/framework` package.

raelga:
@mbarnes moved `isAPINotDeployedError` to `test/util/framework` as exported `IsAPINotDeployedError` in [fa71aed](https://github.com/Azure/ARO-HCP/commit/fa71aed8b), updated both callers to use it. Turns out `nodepool_ephemeral_osdisk.go` defined it but never called it itself, so I dropped the now-unused imports there too.

mbarnes (approved):
/lgtm

PR #6973, opened 2026-09-16
chore: bump cihealth image

roivaz:
## Why

The Test Platform moved new job artifacts to the public GCS bucket after the credential-leak cutover. DEV CIHealth must run the build from roivaz/ARO-HCP-CIHealth#12 before its database is reset and ingestion resumes.

The controller reconciliation horizon is reduced from four weeks to one week so the fresh database does not unnecessarily backfill a month of pre-cutover run metadata. The app keeps its four-week analysis horizon and will accumulate that history naturally.

## Image update

| Name | Old Digest | New Digest | Tag | Date | Status |
| --- | --- | --- | --- | --- | --- |
| cihealth | bf999ec87b4b… | f3fd1b3ecd57… | 4d50a6e | 2026-09-16 07:24 | updated |

## Validation

- `make -C config materialize`
- rendered `config/config-dev-ci.yaml` successfully with `templatize configuration validate`
- Helm chart renders app history at 4 weeks and controller history at 1 week

The full `validate-config-pipelines-dev-ci` target remains blocked by pre-existing restricted-template errors in unrelated `.tmpl.bicepparam` files.


roivaz:
This PR only affects CIHealth deployment which is not exercised at all by e2e-parallel. All other checks are passing. Reason for the force merge: CIHealth is currently broken due to the Test Platform team having moved the GCS bucket where test artifacts are stored.

